PR #1, opened 2019-04-04
Add cross-platform CI

matchai:


matchai:
Rustup is the unified way to manage Rust versions on all platforms, so we can use similar installation methods on all OS's we're testing (Linux, Mac, Windows). We can also use it to change the Rust support channel, so we can additionally test Rust beta and nightly, allowing us to catch deprectations in advance.

Clippy is Rust's linter, so it will spot non-ideomatic code and suggest fixes. 😊

PR #5, opened 2019-04-15
Add README.md badges

matchai:
Let's make that README a little more colorful! 🌈  

### Added
- Discord badge
- Azure pipelines badge

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 <h3 align="center">Starship ✨🚀</h3>
 <p align="center">The cross-shell prompt for astronauts.</p>
-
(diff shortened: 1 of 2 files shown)


Snuggle:
Is there any difference between https://badgen.net/ and https://shields.io/?

matchai:
ShieldsIO was recent for a few years, causing badges to be broken on repos. I don't know if they've made any changes to their infra, but Badgen is lightweight (2K lines of code), really fast, and hasn't broken any badges that I've seen. Just anecdotal experience though. 🤷‍♂️

PR #6, opened 2019-04-15
Add integration tests for directory segment

matchai:
### Changed
- Added current_dir param to segments to make them more testable
- Moved all existing integration tests to a `tests/` dir

### Added
- A whole bunch of new integration tests

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -3,7 +3,6 @@ use ansi_term::Color;
 use clap::ArgMatches;
 use dirs;
 use git2::Repository;
-use std::env;
 use std::path::Path;
 
 /// Creates a segment with the current directory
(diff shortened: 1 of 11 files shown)


matchai:
The heck?? The tests pass in CI?
The git dir sections are failing for me locally 🤔 

Going to investigate a little more before it's ready for review.

src/modules/directory.rs
```diff
@@ -15,28 +14,25 @@ use std::path::Path;
...
-
     let dir_string;
-    if let Ok(repo) = git2::Repository::discover(&current_path) {
+    if let Ok(repo) = git2::Repository::discover(current_dir) {
         // Contract the path to the git repo root
         let repo_root = get_repo_root(&repo);
         let repo_folder_name = repo_root.file_name().unwrap().to_str().unwrap();
 
-        dir_string = contract_path(&current_path, repo_root, repo_folder_name);
+        dir_string = contract_path(current_dir, repo_root, repo_folder_name);
```
Snuggle: What do the &variables do and why did you remove these?

tests/directory.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    Ok(())
+}
+
+#[test]
+fn truncated_directory_in_git_repo() -> io::Result<()> {
+    let temp_dir = TempDir::new("starship")?;
+    let repo_dir = temp_dir.path().join("rocket-controls");
+    let dir = repo_dir.join("src/meters/fuel-gauge");
+    fs::create_dir_all(&dir)?;
```
Snuggle: This is super odd syntax! What does it do? `?;`

src/modules/directory.rs
matchai: `&variables` are references to an object, rather than objects themselves.
It's good practice to use instances where possible as it's lighter in memory to pass a reference around than an instance.

https://github.com/starship/starship/pull/6/files/daeb7cf6bfbd170f149ea98a73bfa5d5a5b67a07#diff-95036686adb679a23cd72cd4cca8cbc9R17
In this case (`current_dir: &Path`) `current_dir` is a reference to a `Path`, so I don't need to add the ampersand. If instead, the function param accepted `current_dir: Path`, I would refer to it with `&current_dir` so that I was using the reference.

Is that clear? 🙂

tests/directory.rs
matchai: `let temp_dir = TempDir::new("starship")?;` will compile into:
```rust
match TempDir::new("starship") {
    Ok(dir)  => dir,
    Err(err) => return Err(err),
}
```

It's a quick way to return an `Error` if the function around it is a `Result` (which will allow returning `Error`s).

PR #14, opened 2019-04-18
Use "context" to contain run details

matchai:
I noticed we were passing around a whole lot of arguments with the details of the particular run. I did a little refactoring to contain that data in a `context` struct.

In the future, this will contain other details such as:
- A list of files in `PWD`
- A list of binaries in `PATH`
- Configuration parsed from the config file
- Maybe cached values from previous runs? 🤔

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,0 +1,27 @@
+use clap::ArgMatches;
+use std::env;
+use std::path::PathBuf;
+
+pub struct Context<'a> {
+    pub current_dir: PathBuf,
+    pub arguments: ArgMatches<'a>,
+}
+
+impl<'a> Context<'a> {
+    pub fn new(arguments: ArgMatches) -> Context {
+        // TODO: Currently gets the physical directory. Get the logical directory.
+        let current_dir = env::current_dir().expect("Unable to identify current directory.");
+
+        Context {
+            current_dir,
+            arguments,
+        }
+    }
+
+    pub fn new_with_dir<T>(arguments: ArgMatches, dir: T) -> Context where T: Into<PathBuf> {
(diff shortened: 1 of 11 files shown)


src/modules/line_break.rs
```diff
@@ -1,9 +1,8 @@
 use super::Segment;
-use clap::ArgMatches;
-use std::path::Path;
+use crate::context::Context;
 
 /// Creates a segment for the line break
-pub fn segment(_current_dir: &Path, _args: &ArgMatches) -> Option<Segment> {
+pub fn segment(_context: &Context) -> Option<Segment> {
```
Multimo: What do you think of removing `Context` struct as a argument here? It will make it easier to refactor in the future if we can easily see what does and does not depend on `Context`?

src/context.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub struct Context<'a> {
+    pub current_dir: PathBuf,
+    pub arguments: ArgMatches<'a>,
+}
+
+impl<'a> Context<'a> {
+    pub fn new(arguments: ArgMatches) -> Context {
+        // TODO: Currently gets the physical directory. Get the logical directory.
+        let current_dir = env::current_dir().expect("Unable to identify current directory.");
```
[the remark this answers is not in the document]
matchai: Yep for sure. 👌 
In the old prompts we would scan the current directory once per segment, so we'll do what we can to reused that same data in the shared context.

src/modules/line_break.rs
matchai: I was thinking that the `_` prefix would denote that the argument is unused.

PR #20, opened 2019-04-24
Add a Contributing section to README.md

Multimo:
We could defs add more here but I think its a start.

@matchai any thoughts on what else to add? Maybe a discord link?

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,3 +99,10 @@ I'm very new to Rust, so any help is appreciated when it comes to improving deve
     ```bash
     ZSH_THEME="starship"
     ```
+## Contributing
+
+To test locally run the below command:
+
+```bash
+cargo run -- $status
+```


matchai:
I definitely agree that we should better document the contribution process. Usually, the place to put that would be in `CONTRIBUTING.md`. When the file is in a repo, GitHub will let contributors know when creating PRs:

image


If you want to help us get started with some basics, here are some good examples:
https://github.com/opengovernment/opengovernment/blob/4dc9499f735b680c124f04100b1ab19146242a13/CONTRIBUTING.md
https://github.com/rails/rails/blob/81b8edc31dd11ae2bbff52bdd1fa730c024405d8/CONTRIBUTING.md
https://github.com/prettier/prettier/blob/ed15b6d04e60a75d81c406ee899a86690187d44e/CONTRIBUTING.md

PR #23, opened 2019-04-24
Add Python version segment

### LINKED ISSUE #22: Implement the prompt segment for Python version (🐍)
### Feature Details

Implement the prompt segment containing the currently running Python version.

Should show if:

- python is available as a command
- The current directory contains
   - .python-version
   - requirements.txt
   - pyproject.toml
   - *.py

### Implementation

The implmentation done for spaceship can be referenced here: https://github.com/denysdovhan/spaceship-prompt/blob/cd4bd15b8bada850d8f68c664e669b844aec0519/sections/pyenv.zsh

johnletey:
This is a quick implementation ... would love feedback!

Closes #22

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -8,8 +8,9 @@ use std::process::Command;
 ///
 /// Will display the Python version if any of the following criteria are met:
 ///     - Current directory contains a `.py` file
-///     - Current directory contains a `requirements.txt` file
 ///     - Current directory contains a `.python-version` file
+///     - Current directory contains a `requirements.txt` file
+///     - Current directory contains a `pyproject.toml` file
 pub fn segment(context: &Context) -> Option<Segment> {
     let is_py_project = context.dir_files.iter().any(has_py_files);
     if !is_py_project {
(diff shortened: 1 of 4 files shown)


src/modules/python.rs
```diff
@@ -0,0 +1,76 @@
...
+    };
+
+    is_py_file(&dir_entry)
+        || is_python_version(&dir_entry)
+        || is_requirements_txt(&dir_entry)
+        || is_py_project(&dir_entry)
+}
+
+fn get_python_version() -> Option<String> {
+    match Command::new("python").arg("-V").output() {
```
matchai: Total nitpick, but I'd personally use `--version` over `-V` to make it explicit. `-V` can often be the flag for "verbose".

```suggestion
    match Command::new("python").arg("--version").output() {
```

src/modules/python.rs
```diff
@@ -0,0 +1,76 @@
...
+        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
+        Err(_) => None,
+    }
+}
+
+fn format_python_version(python_stdout: String) -> String {
+    format!("v{}", python_stdout.replace("Python ", "").trim())
```
matchai: Another total nitpick. I'd use `trim_start_matches` over `replace` for a little more explicitness.
```suggestion
    format!("v{}", python_stdout.trim_start_matches("Python ").trim())
```

matchai:
Looks great! 😄 
Thank you for your contribution @johnletey!

PR #25, opened 2019-04-25
Add package version segment

### LINKED ISSUE #17: Implement the prompt segment for package version (📦)
### Feature Details
Implement the prompt segment containing the package version of the package present in the current directory.

#### Node.js
Should show package versions from Node.js projects from the `package.json` file, in the `version` key.

#### Rust
Should show package versions from Rust projects from the `Cargo.toml` file, in the `package.version` key.

### Implementation
The original spaceship implementation showing `⚠️` when the package is missing a version was the source of a lot of confusion, so I don't personally think it should be added. I'm open to discussing the implementation.

The implmentation done for spaceship can be referenced here: https://github.com/denysdovhan/spaceship-prompt/blob/d9f25e14e7bec4bef223fd8a9151d40b8aa868fa/sections/package.zsh

johnletey:
Would love some feedback!

Closes #17

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -47,8 +47,9 @@ fn has_js_files(dir_entry: &PathBuf) -> bool {
 fn get_package_version(context: &Context) -> Option<String> {
     let is_rs_project = context.dir_files.iter().any(has_rs_files);
     if !is_rs_project {
+        // TODO: Implement for nodejs projects
         let is_js_project = context.dir_files.iter().any(has_js_files);
-        // TODO: Implement for nodejs
+        return None;
     } else {
         match Command::new("cargo").arg("pkgid").output() {
             Ok(output) => Some(format_rust_version(
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -26,51 +26,59 @@ pub fn segment(context: &Context) -> Option<Segment> {
     }
 }
 
+// TODO: Combine into one function and just call for different file names!
 fn has_rs_files(dir_entry: &PathBuf) -> bool {
-    let is_rs_file =
-        |d: &PathBuf| -> bool { d.is_file() && d.extension().unwrap_or_default() == "rs" };
     let is_cargo_toml =
         |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "Cargo.toml" };
 
-    is_rs_file(&dir_entry) || is_cargo_toml(&dir_entry)
+    is_cargo_toml(&dir_entry)
 }
 
 fn has_js_files(dir_entry: &PathBuf) -> bool {
-    let is_js_file =
-        |d: &PathBuf| -> bool { d.is_file() && d.extension().unwrap_or_default() == "js" };
-    let is_node_modules =
-        |d: &PathBuf| -> bool { d.is_dir() && d.file_name().unwrap_or_default() == "node_modules" };
     let is_package_json = |d: &PathBuf| -> bool {
         d.is_file() && d.file_name().unwrap_or_default() == "package.json"
     };
 
-    is_js_file(&dir_entry) || is_node_modules(&dir_entry) || is_package_json(&dir_entry)
+    is_package_json(&dir_entry)
+}
+
+// TODO: Move to `utils.rs` file and import
+fn read_file(file_name: String) -> String {
+    let mut file = File::open(file_name).unwrap();
+    let mut data = String::new();
+    file.read_to_string(&mut data).unwrap();
+
+    data
+}
+
+fn extract_cargo_version() -> String {
+    let data = read_file("Cargo.toml".to_string());
+
+    let toml = data.parse::<toml::Value>().unwrap();
+    let version = format_version(toml["package"]["version"].to_string());
+    return version;
+}
+
+fn extract_package_version() -> String {
+    let data = read_file("package.json".to_string());
+
+    let json: serde_json::Value = serde_json::from_str(&data).unwrap();
+    let version = format_version(json["version"].to_string());
+    return version;
 }
 
 fn get_package_version(context: &Context) -> Option<String> {
     let is_rs_project = context.dir_files.iter().any(has_rs_files);
-    if !is_rs_project {
-        let is_js_project = context.dir_files.iter().any(has_js_files);
-        if !is_js_project {
-            return None;
-        } else {
-            let mut file = File::open("package.json").unwrap();
-            let mut data = String::new();
-            file.read_to_string(&mut data).unwrap();
-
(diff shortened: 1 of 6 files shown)


src/modules/package.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+fn has_rs_files(dir_entry: &PathBuf) -> bool {
+    let is_rs_file =
+        |d: &PathBuf| -> bool { d.is_file() && d.extension().unwrap_or_default() == "rs" };
+    let is_cargo_toml =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "Cargo.toml" };
+
+    is_rs_file(&dir_entry) || is_cargo_toml(&dir_entry)
```
matchai: Should it matter if there are `*.rs` files but no `Cargo.toml`?
I don't know that we should be checking for anything but `Cargo.toml` here.

src/modules/package.rs
```diff
@@ -0,0 +1,89 @@
...
+fn has_js_files(dir_entry: &PathBuf) -> bool {
+    let is_js_file =
+        |d: &PathBuf| -> bool { d.is_file() && d.extension().unwrap_or_default() == "js" };
+    let is_node_modules =
+        |d: &PathBuf| -> bool { d.is_dir() && d.file_name().unwrap_or_default() == "node_modules" };
+    let is_package_json = |d: &PathBuf| -> bool {
+        d.is_file() && d.file_name().unwrap_or_default() == "package.json"
+    };
+
+    is_js_file(&dir_entry) || is_node_modules(&dir_entry) || is_package_json(&dir_entry)
```
matchai: Same here. It doesn't seem very actionable if we have a `*.js` file or `node_modules` folder without a `package.json` file.

src/modules/package.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+fn get_package_version(context: &Context) -> Option<String> {
```
matchai: I think we should try to avoid the nesting the if statements here if possible.

Maybe the following would work:
```rust
let is_rs_project = ...;
if is_rs_project {
	return extract_cargo_version();
}

let is_js_project = ...;
if is_js_project {
	return extract_package_version();
}

None
```

src/modules/package.rs
```diff
@@ -0,0 +1,89 @@
...
+    let is_rs_project = context.dir_files.iter().any(has_rs_files);
+    if !is_rs_project {
+        let is_js_project = context.dir_files.iter().any(has_js_files);
+        if !is_js_project {
+            return None;
+        } else {
+            let mut file = File::open("package.json").unwrap();
```
matchai: This could possibly throw if you don't have read permissions on `package.json`.
`unwrap()` would cause the prompt to panic in that case.

src/modules/package.rs
```diff
@@ -0,0 +1,89 @@
...
+            let mut data = String::new();
+            file.read_to_string(&mut data).unwrap();
+
+            let json: serde_json::Value = serde_json::from_str(&data).unwrap();
+            let version = format_version(json["version"].to_string());
```
matchai: What happens here if `package.json` doesn't have a version?

src/modules/package.rs
```diff
@@ -0,0 +1,89 @@
...
+    match get_package_version(context) {
+        Some(package_version) => {
+            const RUST_CHAR: &str = "📦";
+            const SEGMENT_COLOR: Color = Color::Red;
+
+            let mut segment = Segment::new("package");
+            // TODO: Should have prefix "is "
+            segment.set_style(SEGMENT_COLOR);
+
+            segment.set_value(format!("{} {}", RUST_CHAR, package_version));
```
sirMerr: You can chain these, i.e. `segment.set_style(...).set_value(...)`

src/modules/package.rs
```diff
@@ -0,0 +1,89 @@
...
+            file.read_to_string(&mut data).unwrap();
+
+            let json: serde_json::Value = serde_json::from_str(&data).unwrap();
+            let version = format_version(json["version"].to_string());
+            return Some(version);
+        }
+    } else {
+        let mut file = File::open("Cargo.toml").unwrap();
+        let mut data = String::new();
+        file.read_to_string(&mut data).unwrap();
```
sirMerr: You could abstract this logic since it's pretty similar to what you do on https://github.com/starship/starship/pull/25/files#diff-fbc7ade66816e21bb775d1b968edfa64R59. It might be handy to have it across many of our segments eventually.

johnletey: @sirMerr I agree.

src/modules/package.rs
johnletey: @matchai So remove the `.unwrap()`?

src/modules/package.rs
johnletey: I get a `vnull` ... which means that the version is `null` and then I appended a `v`

Shot 2019-04-26 at 08 06 12

src/modules/package.rs
johnletey: @sirMerr This would over complicate the line ... I'm leaving it as is!

src/modules/package.rs
johnletey: I'm planning on just making 1 function (`has_file`) that checks to see if there is a certain `file_name` in the current directory. Then I just call `has_file("Cargo.toml")` and `has_file("package.json")`

src/modules/package.rs
matchai: We should handle the error instead of letting it panic.
I'd take a look at Rust error handling solutions.

src/modules/package.rs
Snuggle: What would be the ideal way to handle this scenario?

If `package.json`  doesn't have a version, should we hide the section? If we did, that might be confusing for users, _why does my prompt not work?!_

Or should it say something like ` 📦 NO_VER` or `📦 vN/A` or...?

johnletey: @Snuggle @matchai I'd personally vote for `📦 NO_VER`. What do you guys think?

matchai: In spaceship it has always been `📦⚠️`, which has lead to _a lot_ of confusion. I'd be fine with changing it here.

I'd personally prefer not to show the version section at all, to be honest. 🤔
`📦 NO_VER` is fine with me if that's what most people prefer, but I'm afraid it may just be noisy if people are intentionally omitting a version.

johnletey: > but I'm afraid it may just be noisy if people are intentionally omitting a version.

@matchai Exactly what I was thinking ... let's leave it out!

johnletey:
@matchai @sirMerr @Snuggle I'm going to close this issue and reopen it (with all the `master` branch changes added. Stay tuned ...

PR #26, opened 2019-04-25
Add docker for integration tests

matchai:
### Details
Experimenting with how we could implement integration tests, I've happened upon this solution:
- All tests that are dependant on external binaries or that make filesystem modifications outside of a tempdir have the `#[ignore]` attribute
	- Allows us to avoid having new contributors accidentally create files all over their filesystem if they haven't read the `CONTRIBUTING.md` docs (when we get to it)
- An `integration_test` bash script is created that does the following:
	- Build image with a fixed version of binaries that we depend on
	- Caches dependencies in a dummy project for quicker builds
	- Runs `cargo test -- --ignored` in the docker instance, running _all_ tests

This allows for an easily reproducible integration test run that can include filesystem operations anywhere and can reliably have all required dependencies at the expected versions.
Unit tests can still be run as expected with `cargo test`.

### Drawbacks
- ~~As it is, the docker image does not rebuild when the Dockerfile is modified~~
- ~~The docker image does not rebuild if dependencies change~~
- This will make testing on Linux easier, but we may have to do something similar in our Azure pipelines to allow us to run integration tests on macOS and Windows

Questions, opinions or suggestions are appreciated. 😄 
I am a total Docker newbie, so there might be a more correct way of doing this.


johnletey:
@matchai I'm very new to docker, but after quickly creating an account/installing on my computer and running `./integration_test`, everything worked

so it's definitely idiot proof 👏

johnletey:
@matchai I'd also add a test that shows what the prompt looks like. So just run `cargo run -- $status` in the Dockerfile.

johnletey:
@matchai The only problem with this is that it would put a constraint on developers ... that is, that they have to have Docker installed on their machine

matchai:
> I'd also add a test that shows what the prompt looks like. So just run cargo run -- $status in the Dockerfile.

I don't know that the Dockerfile is the right place to put that. 🤔
Maybe if we had a Makefile with common scripts, we could put it there.

> The only problem with this is that it would put a constraint on developers ... that is, that they have to have Docker installed on their machine

Correct. I've updated the script to warn the user if they don't have Docker installed.
Contributors don't _have_ to have docker. These tests will also be run as part of CI once they are merged in, so they'll see the logs in Azure Pipelines.

johnletey:
> Maybe if we had a Makefile with common scripts, we could put it there.

Sounds good

> Correct. I've updated the script to warn the user if they don't have Docker installed.

Wonderful!

tests/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM rust:latest
+
+# Install Node.js
+ENV NODE_VERSION 12.0.0
+RUN curl https://raw.githubusercontent.com/creationix/nvm/master/install.sh | bash \
```
Multimo: I think we might have to include all the different languages ( python, Go etc.) here? Could get pretty intense?

tests/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# We want dependencies cached, so copy those first
+COPY ./Cargo.lock ./Cargo.lock 
+COPY ./Cargo.toml ./Cargo.toml 
+
+# Cargo.toml will fail to parse without my_benchmark
+RUN mkdir benches
+RUN touch benches/my_benchmark.rs
+
+# This is a dummy build to get dependencies cached
+RUN cargo build --release
```
Multimo: If we wanna go this route it could potentially useful to do multi stage builds or alternatively just build the base image with all the environments and then push that image to docker-hub or the azure pipeline image repo. 

We could then reuse that image, could make things a lot faster since its reusing cached images. 
Especially since they gonna pretty big with all the environments needed to run the tests.

tests/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+  && . $HOME/.nvm/nvm.sh \
+  && nvm install $NODE_VERSION \
+  && nvm alias default $NODE_VERSION \
+  && nvm use default
+ENV PATH /root/.nvm/versions/node/v$NODE_VERSION/bin:$PATH
+# Check that Node.js was correctly installed
+RUN node --version 
+
+# Create blank project
+RUN USER=root cargo new --bin starship
```
Multimo: Its cleaner to just `COPY . .` instead. Would also put it after the `COPY Cargo.toml` command.

tests/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+WORKDIR /starship
+
+# We want dependencies cached, so copy those first
+COPY ./Cargo.lock ./Cargo.lock 
+COPY ./Cargo.toml ./Cargo.toml
```
Multimo: `The docker image does not rebuild if dependencies change` 

Each command here will run and create a layer / image. Everytime the docker image rebuild it will check the hash and either use the cache or re run the command. So in this case it will hash the contents of the `Cargo.toml` and use that as the hash. So if the contents change it will rerun this command and all below it.

Multimo:
""I'd also add a test that shows what the prompt looks like. So just run cargo run -- $status in the Dockerfile."""

Should be a test inside the project no?

"""The only problem with this is that it would put a constraint on developers ... that is, that they have to have Docker installed on their machine"""

I think this is a fair enough request if someone wants to run the tests locally. I think Docker is pretty standard nowadays because its easy to use and takes away the issue of "Its worked on my machine" :)

tests/Dockerfile
matchai: Correct. I have intentionally made the `integration_test` script skip `docker build` if the image already exists because of the time required to send the build context to the docker daemon when `target/` is built. It's usually around 1GB, so it can take a while.

It looks like adding `.dockerignore` to the project root with `target/` solves it. 😄

tests/Dockerfile
matchai: > it could potentially useful to do multi stage builds

I haven't done multi-stage builds before, but I'm open to it.

> alternatively just build the base image with all the environments and then push that image to docker-hub or the azure pipeline image repo.

Yep, that's what I had in mind.
For the time being the base image will be changing _very_ often, so I wasn't sure how we would want to manage it so that changes could be made to the image while making local features testable. Would we keep the `Dockerfile` in this repo? Can we default to using a local image over the remote one? I haven't done a whole lot of Docker, so there's probably a good answer. 😄

tests/Dockerfile
matchai: It definitely can get hairy.
I'd be curious to know if we can find a good polyglot image with fixed versions, but also the ability to install stable Rust. 🤔

tests/Dockerfile
matchai: We're intentionally creating an empty project so that we can download and cache dependencies in the snapshot. It will make consecutive builds much faster since `src` is often changing.

matchai:
>Should be a test inside the project no? Just running the main function, could be useful even on local. Not sure how you would asset the test tho, didnt panic?

I think a full-prompt test would be most useful in the integration tests, so that we can verify it completely with the fixed language versions. Definitely something we will be adding soon.

matchai:
I updated the Azure Pipelines tests to perform the full integration tests on all OSes, and removed the benchmark, as it will soon be replaced with a benchmarking GitHub Action.

PR #32, opened 2019-04-26
Add the Git branch segment

matchai:
I really didn't realize how much I depended on the git branch section of the prompt. With this added feature, you should now be pretty comfortable using this as your daily driver. 🚗  

Partly implements #11.

### Added
- `Repository` to Context for reuse in `directory` and `git_branch`
- `git_branch` to prompt

### Changed
- Made segments **bold** to match spaceship

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -7,6 +7,7 @@ pub struct Context<'a> {
     pub current_dir: PathBuf,
     pub dir_files: Vec<PathBuf>,
     pub arguments: ArgMatches<'a>,
+    pub repository: Option<git2::Repository>,
 }
 
 impl<'a> Context<'a> {
@@ -34,10 +35,13 @@ impl<'a> Context<'a> {
             .map(|entry| entry.path())
             .collect::<Vec<PathBuf>>();
 
+        let repository: Option<git2::Repository> = git2::Repository::discover(&current_dir).ok();
+
         Context {
-            current_dir,
             arguments,
+            current_dir,
             dir_files,
+            repository,
         }
     }
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,9 +1,11 @@
 use clap::ArgMatches;
 use std::env;
+use std::fs::{self, ReadDir};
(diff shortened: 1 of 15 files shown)


src/context.rs
```diff
@@ -34,10 +36,13 @@ impl<'a> Context<'a> {
             .map(|entry| entry.path())
             .collect::<Vec<PathBuf>>();
 
+        let repository: Option<Repository> = Repository::discover(&current_dir).ok();
```
Multimo: It looks like we are running `Context::new()` every time?
Perhaps this is premature but it could be a good idea to chuck the Context struct on the heap and return a reference instead? try to only it initialize once and then reuse across re-renders since i'm pretty sure its read-only anyway? 

Could make it easy to do the caching you were talking of but then we have to think about invalidation which is a whole thing. 

Anyway just a thought, something we could look into later.

matchai: > It looks like we are running Context::new() every time?
> Perhaps this is premature but it could be a good idea to chuck the Context struct on the heap and return a reference instead?

I'm not sure I understand. `Context::new()` should only be running once in `print.rs` every time the whole prompt is being rendered. The only time context is used is by the reference passed here to generate the modules:
https://github.com/starship/starship/blob/master/src/print.rs#L30

I was hoping to eventually implement some form of caching that would persist between calls to `starship`, but yeah, that will end up being complicated.

src/context.rs
```diff
@@ -34,10 +36,13 @@ impl<'a> Context<'a> {
...
+
         Context {
-            current_dir,
             arguments,
+            current_dir,
             dir_files,
+            repository,
```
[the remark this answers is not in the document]
matchai: I'm kind of just adding things that would reduce redundant operations across segments. 😅 

I know that we'll end up using `Context` to share user configuration across segments, and maybe a list of binaries accessible via `$PATH` (so that we don't have to call `Command::new()` when it doesn't make sense to). I don't foresee much else being added after this point.

Can you think of a better way to organize `Context`?

PR #34, opened 2019-04-27
Implement directory scanner

Multimo:
#### Description

Refactored all the functions that handle looking at an dir path and check to see if it matches some criteria, has '.js' extension etc.

#### Motivation and Context

Closes #

We had a bit of repeated logic to see whether a language matches a criteria, `is_js_project` etc. Moved it all into a series of functions so its easier to reuse.

+ not super happy with the struct as an argument approach thinking just arguments to a function is better? could make it easier to handle optional arguments. Was also thinking about making it a method of context? What do you guys think? 

+ had trouble testing the (`is_dir()` code path]("https://play.rust-lang.org/?version=nightly&mode=debug&edition=2015"). I think there is a bug in rust itself here :(
Will investigate more than potentially raise an issue in the language.

+ `find_file` is a bad name but wasnt sure what to name it? Did we want to do a utils mod by chance?

#### Types of changes

- [x] Refactor (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:


src/find_file.rs
```diff
@@ -0,0 +1,109 @@
...
+            == **file
+    });
+
+    match found_file {
+        Some(file) => !file.is_empty(),
+        None => false,
+    }
+}
+
+pub fn has_files_with_extension(dir_entry: &PathBuf, extension: &String) -> bool {
```
matchai: It looks to me like this is all the function is doing right now.

```suggestion
pub fn has_extension(dir_entry: &PathBuf, extension: &String) -> bool {
```

src/modules/nodejs.rs
```diff
@@ -12,7 +12,13 @@ use crate::context::Context;
 ///     - Current directory contains a `package.json` file
 ///     - Current directory contains a `node_modules` directory
 pub fn segment(context: &Context) -> Option<Segment> {
-    let is_js_project = context.dir_files.iter().any(has_js_files);
+    let js_criteria = find_file::Criteria {
```
matchai: It looks to me like we'll eventually need all three `find_file::Criteria` to be vectors.
We should probably plan ahead for that. 🤔

src/modules/python.rs
```diff
@@ -1,18 +1,24 @@
...
-///     - Current directory contains a `.python-version` file
+///     - Current directory contains a `.pyproject.toml` file
 ///     - Current directory contains a `requirements.txt` file
 ///     - Current directory contains a `pyproject.toml` file
 pub fn segment(context: &Context) -> Option<Segment> {
-    let is_py_project = context.dir_files.iter().any(has_py_files);
+    let python_criteria = find_file::Criteria {
+        files: vec!["requirements.txt, pyproject.toml", ".pyproject.toml"],
+        extension: "py".to_string(),
+        folder: "".to_string(),
```
matchai: Since we won't always need all three criteria, maybe this would be best done as a builder pattern?

matchai:
> not super happy with the struct as an argument approach thinking just arguments to a function is better? could make it easier to handle optional arguments. Was also thinking about making it a method of context? What do you guys think?

I agree. Since we won't need all arguments all the time, I think a builder pattern, like for `Segment` would work best here.

> find_file is a bad name but wasnt sure what to name it? Did we want to do a utils mod by chance?

I'd personally name it something like `project_detection.rs`, or something along those lines.
I'd be good with adding a utils mod.

src/find_file.rs
```diff
@@ -0,0 +1,109 @@
...
+    let found_file = files.into_iter().find(|file| {
+        dir_entry
+            .file_name()
+            .and_then(OsStr::to_str)
+            .unwrap_or_default()
+            == **file
+    });
+
+    match found_file {
+        Some(file) => !file.is_empty(),
```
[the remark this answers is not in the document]
Multimo: This is bad naming again on my part this is actually a `str`. Updated the name to `file_name`, hopefully thats a little clearer?

src/find_file.rs
Multimo: 👍

src/modules/nodejs.rs
Multimo: 👍

matchai:
From an ergonomics standpoint, I would like it if we could trim away some of the boilerplate.

How would you feel about an API like this?

```rust
let is_js_project = context::scan_dir::new()
    .has_files(vec!["package.json"])
    .has_folders(vec!["node_modules"])
    .has_extensions(vec!["js"])
    .scan();
```

Context doesn't need to be provided, no need for additional imports, and the output can be gotten with just another chained function.

Multimo:
I can defs agree that the ergonomics could be better. What do you think of the example below?

```rs
    let is_rs_project = project_detector::Criteria::new()
        .set_files(vec!["Cargo.toml"])
        .set_extensions(vec!["rs"])
        .scan(&context.dir_files);
    if !is_rs_project {
        return None;
    }
```

We are passing context around as an immutable reference to all modules. It would mean every time we call scan we have to clean up the `scan_dir` fields in context? I am not sure the extra complexity and state would be worth the ergonomics.

What do you think?

matchai:
> We are passing context around as an immutable reference to all modules. It would mean every time we call scan we have to clean up the scan_dir fields in context?

I'm not sure I totally understand. In the example I provided, `context::scan_dir::new()` would return a new instance of `scan_dir`, which should no longer be associated with immutable `context`. That `scan_dir` instance is now owned where the constructor was called, so there should be no need to clean its fields.

If what I'm suggesting doesn't make sense, I'm also fine with the syntax you proposed, where `scan()` is provided with `&context.dir_files`. 😄

src/modules/python.rs
```diff
@@ -8,11 +8,15 @@ use super::{Context, Module};
 ///
 /// Will display the Python version if any of the following criteria are met:
 ///     - Current directory contains a `.py` file
-///     - Current directory contains a `.python-version` file
```
matchai: I'm pretty sure `.python-version` is used by pyenv. 🤔

src/modules/python.rs
```diff
@@ -35,25 +39,6 @@ pub fn segment(context: &Context) -> Option<Module> {
     }
 }
 
-fn has_py_files(dir_entry: &PathBuf) -> bool {
```
matchai: Feels good to strip this away 🤤

src/modules/python.rs
Multimo: whoops!

src/modules/python.rs
Multimo: image

Multimo:
So i have been trying to implement what you described but I am having a lot of trouble implementing it.

```
let is_js_project = context::scan_dir::new()
    .has_files(vec!["package.json"])
    .has_folders(vec!["node_modules"])
    .has_extensions(vec!["js"])
    .scan();
```

The problem I am mainly having is around calling it and lifetimes. I must be missing something crucial with how traits work as i have hit a wall with it.

If you want to have a go i would love to see what you come up with?

matchai:
I'll give it a shot. 💪

matchai:
@Multimo 
Here's my version of the latter API, which worked locally: https://gist.github.com/matchai/4043e7cbe11a9b6236c0cc13957ad032

src/utils/project_detector.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+impl<'a> Criteria<'a> {
+    pub fn new() -> Criteria<'a> {
+        Criteria {
+            ..Default::default()
+        }
+    }
+
+    pub fn set_files(mut self, files: Vec<&'static str>) -> Self {
```
matchai: I think accepting an array may be the more idiomatic and performant way to do this.
That's what's done in the `std::process::Command` args API: https://doc.rust-lang.org/src/std/process.rs.html#490-516

Multimo: Yup your right there is less overhead and we know it's static, good thinking!

Multimo:
@matchai Thanks for showing me 👍  I have finished up on the implementation, let me know if you see anything else that could be done?

src/context.rs
```diff
@@ -56,4 +57,138 @@ impl<'a> Context<'a> {
...
+
+        buf.set_file_name("some-file.rs");
+        assert_eq!(has_extension(&buf, &extensions), false);
+
+        buf.set_file_name("some-file.js");
+        assert_eq!(has_extension(&buf, &extensions), true)
+    }
+
+    #[test]
+    fn test_criteria_scan() {
```
matchai: 👍

src/context.rs
```diff
@@ -56,4 +57,138 @@ impl<'a> Context<'a> {
...
+        ScanDir {
+            dir_files: self.dir_files.as_ref(),
+            files: &[],
+            folders: &[],
+            extensions: &[],
+        }
+    }
+}
+
+// A Scruct of Criteria which will be used to verify current PathBuf is
```
matchai:
```suggestion
// A struct of Criteria which will be used to verify current PathBuf is
```

src/context.rs
```diff
@@ -56,4 +57,138 @@ impl<'a> Context<'a> {
...
+// of X language, criteriass can be set via the builder pattern
```
matchai:
```suggestion
// of X language, criteria can be set via the builder pattern
```

PR #35, opened 2019-04-28
Add the package version segment

### LINKED ISSUE #17: Implement the prompt segment for package version (📦)

johnletey:
#### Description

A continuation of #25, which I closed because there were some major changes to `master` and merging would have been difficult. I've tried to incorporate everything from the reviews given in the previous PR.

#### Motivation and Context

Closes #17

#### Types of changes

Shot 2019-04-28 at 15 59 54

#### Checklist:


src/modules/package.rs
```diff
@@ -0,0 +1,108 @@
...
+    let is_package_json = |d: &PathBuf| -> bool {
+        d.is_file() && d.file_name().unwrap_or_default() == "package.json"
+    };
+
+    is_package_json(&dir_entry)
+}
+
+// TODO: Move to `utils.rs` file and import
+fn read_file(file_name: String) -> String {
+    let mut file = File::open(file_name).unwrap();
```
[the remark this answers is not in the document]
johnletey: @matchai Is this resolved?

src/modules/package.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+fn is_package_json(dir_entry: &PathBuf) -> bool {
+    dir_entry.is_file() && dir_entry.file_name().unwrap_or_default() == "package.json"
+}
+
+// TODO: Move to `utils.rs` file and import
+fn read_file(file_name: &str) -> String {
+    let mut file = File::open(file_name).unwrap_or_else(|error| {
+        panic!("There was a problem opening the file: {:?}", error);
```
matchai: The issue with having unwrap before was that it would panic.
Unfortunately the current solution still panics, but now with a descriptive error.

We want to make sure that the prompt never panics since it'll be hard to use terminal to solve the problem. The ideal would be that in these instances the prompt will simply early return `None`, or something along those lines.

johnletey: @matchai Resolved?

src/modules/package.rs
```diff
@@ -0,0 +1,125 @@
...
+fn is_cargo_toml(dir_entry: &PathBuf) -> bool {
+    dir_entry.is_file() && dir_entry.file_name().unwrap_or_default() == "Cargo.toml"
+}
+
+fn is_package_json(dir_entry: &PathBuf) -> bool {
+    dir_entry.is_file() && dir_entry.file_name().unwrap_or_default() == "package.json"
+}
+
+// TODO: Move to `utils.rs` file and import
+fn read_file(file_name: &str) -> Result<String, io::Error> {
```
matchai: I recommend using the io Result here.
```suggestion
fn read_file(file_name: &str) -> io::Result<String> {
```

src/modules/package.rs
```diff
@@ -0,0 +1,125 @@
...
+    let mut file = File::open(file_name)?;
+    let mut data = String::new();
+
+    file.read_to_string(&mut data)?;
+
+    Ok(data)
+}
+
+fn extract_cargo_version() -> Option<String> {
```
matchai: The two extract functions next really deep because we keep using `match` for flow control.
I would really prefer it if we made more use of early returns than nesting deeper into another `match`.

A really good read on the topic: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

src/modules/package.rs
```diff
@@ -0,0 +1,129 @@
...
+fn read_file(file_name: &str) -> io::Result<String> {
+    let mut file = File::open(file_name)?;
+    let mut data = String::new();
+
+    file.read_to_string(&mut data)?;
+
+    Ok(data)
+}
+
+fn extract_cargo_version() -> Option<String> {
```
matchai: In order to make `extract_cargo_version()` and `extract_cargo_version()` more testable, I think it'd be best if we had it accept a string argument with the contents of the file. That way we could unit test this function by passing some example contents of a `Cargo.toml` and `package.json`. 😄

matchai: I think something along these lines might be a little more concise:
```rust
fn extract_cargo_version() -> Option<String> {
    let file_contents = read_file("Cargo.toml").ok()?;
    let cargo_toml = file_contents.parse::<toml::Value>().ok()?;
    
    match cargo_toml["package"]["version"].as_str() {
        Some(raw_version) => {
            let version = format_version(raw_version.to_string());
            Some(version)
        }
        None => None
    }
}
```

src/modules/package.rs
```diff
@@ -0,0 +1,108 @@
...
+        return extract_cargo_version(file_contents);
+    }
+
+    let has_package_json = context.dir_files.iter().any(is_package_json);
+    if has_package_json {
+        let file_contents = read_file("package.json").ok()?;
+        return extract_package_version(file_contents);
+    }
+
+    return None;
```
matchai:
```suggestion
   None
```

src/modules/package.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+        None => None,
+    }
+}
+
+fn extract_package_version(file_contents: String) -> Option<String> {
+    let json: Option<serde_json::Value> = serde_json::from_str(&file_contents).ok()?;
+
+    match json {
+        None => None,
```
matchai: nitpick: I would put `None` at the end of the match statement to match the other matches.

PR #39, opened 2019-04-30
Install a fixed version of rust for CI integration tests

matchai:
#### Description

Make `rustup` install a specific version of Rust along with the required Rust toolchain to run the build and tests.

- Split out the Rust installation logic from within `azure-pipelines.yml`
- Define a fixed version to be installed within Azure Pipelines
- Run tests with `rustup run` using the required toolkit
- Make the same changes in the Dockerfile

#### Motivation and Context

In order to be able to have integration tests that run `rust --version`, we need the locally used Rust to be a fixed version, while running tests and builds with the current stable/beta/nightly.

#### Types of changes

- [x] Maintenance

#### Checklist:

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:latest
+FROM rust:1.34.0
 
 # Install Node.js
 ENV NODE_VERSION 12.0.0
@@ -24,7 +24,8 @@ RUN mkdir benches
 RUN touch benches/my_benchmark.rs
 
 # This is a dummy build to get dependencies cached
-RUN cargo build --release
+RUN rustup install stable
+RUN rustup run stable cargo build --release
 
 # Delete the dummy build
 RUN rm -rf /starship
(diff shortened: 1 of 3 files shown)


matchai:
@Multimo Do you think I'm overcomplicating this, or do you think this complexity makes sense?

tests/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:latest
+FROM rust:1.34.0
```
Multimo: This makes sense. `:latest` will always grab the latest version and would invalidate the tests so this is legit 👍

tests/Dockerfile
```diff
@@ -24,7 +24,8 @@ RUN mkdir benches
 RUN touch benches/my_benchmark.rs
 
 # This is a dummy build to get dependencies cached
-RUN cargo build --release
+RUN rustup install stable
+RUN rustup run stable cargo build --release
```
Multimo: I am honestly not sure about the significance of this? Could you expand a little why you made this change? Wouldn't the version of rust specified in the docker image default to stable?

PR #40, opened 2019-04-30
Refactor segments into modules

### LINKED ISSUE #18: RFC: Rethinking the approach to sections/segments
## Motivation
Thinking over the implementation of spaceship, the "section" architecture has made sense for a long time, but as the complexity of sections grew with feature requests, I think we should rethink the approach to how sections of the prompt are divided.

The `kubectl` sections is a good example of where the "section" has been outgrown:
```
[kubectl prefix] [kubectl symbol] [kubectl context] ([kubectl namespace])
      at                ☸️               aks1            kube-system
```
image

This model only allows for prefix, value, and suffix. The only way to compose the above section is by giving individual components to a section multiple jobs, which will make configuration more difficult down the line.

## Design
What I have in mind is the following breakdown:
```
Prompt
└── Module
    ├── Prefix
    ├── Suffix
    └── Segment
        └── Value
```

**Prompt** - The whole visible prompt

**Module** - A collection of segments. In the above example, `kubectl` would be the module, with a prefix of `at `. The module could contain the default styling for each segment within it. A module could have one or many segments.

**Segment** - A segment is an individually configurable component within a module. In the above example, `kubectl namespace` would be a segment, which would inherit the styling from its module, but could have its styling overwritten through configuration or could be disabled altogether.

## Drawbacks
- Clarity when it comes to documenting the project structure to future contributors
- Overhead involved with creating simpler modules (e.g. `char`) 

## Questions
- Would we want prefixes/suffixes for segments as well as for modules?

Taking a look at `kubectl namespace` in the above example, would the parentheses around the namespace be simply a part of the segment value? It would be a shame to not support configurability with regard to the parentheses being there, but maybe that level of configuration is too granular? 🤔

I have done _very little_ OOP, so I am very open to suggestions to feedback!

matchai:
Heeeeere's a big one! 🎣 
This PR does _a lot_! I'm really sorry for the amount that there is to review. 😅 

Many of the lines are just tests being updated. Don't worry too much about those.

This API is _not_ set in stone. All feedback is appreciated.
I don't mind redoing this if it means we're redoing it correctly! 😄 

I'll be doing another pass over these files tomorrow to add documentation where I might've forgotten it.

#### Description

Breaks down what were previously "segments" into "modules" and "segments".
For a quick idea, I suggest taking a look at `nodejs.rs` in this PR.

- Created the `Module` struct and builder functions
- Retrofit all existing segments to be modules

Other stuff that was fixed in the process:
- Skip printing the first module's prefix
- No longer using constructors in tests. Turned assertions into plain string comparisons.

#### Motivation and Context

Closes #18

By breaking elements of the prompt into smaller named portions, we can allow for more granular configuration of how the prompt looks and works. This will make something like disabling all prompt symbols doable.

#### Types of changes

- [x] Refactor

Everything works as it always has 🙌 

image

#### Checklist:

### DIFF
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -19,17 +19,14 @@ pub fn segment(context: &Context) -> Option<Module> {
     match get_node_version() {
         Some(node_version) => {
             const NODE_CHAR: &str = "⬢";
-            const MODULE_COLOR: Color = Color::Green;
+            let module_color = Color::Green.bold();
 
             let mut module = Module::new("node");
-            module.set_style(MODULE_COLOR.bold());
+            module.set_style(module_color);
 
-            let symbol = module.new_segment("symbol");
-            symbol.set_value(NODE_CHAR);
-
-            let version = module.new_segment("version");
             let formatted_version = node_version.trim();
-            version.set_value(formatted_version);
+            module.new_segment("symbol", NODE_CHAR);
+            module.new_segment("version", formatted_version);
 
             Some(module)
         }
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -2,8 +2,7 @@ use ansi_term::Color;
 use std::path::PathBuf;
 use std::process::Command;
 
-use super::{Segment, PromptComponent};
-use crate::context::Context;
+use super::{Context, Module};
 
 /// Creates a segment with the current Node.js version
 ///
@@ -11,7 +10,7 @@ use crate::context::Context;
 ///     - Current directory contains a `.js` file
 ///     - Current directory contains a `package.json` file
 ///     - Current directory contains a `node_modules` directory
-pub fn segment(context: &Context) -> PromptComponent {
+pub fn segment(context: &Context) -> Option<Module> {
     let is_js_project = context.dir_files.iter().any(has_js_files);
     if !is_js_project {
         return None;
@@ -20,15 +19,19 @@ pub fn segment(context: &Context) -> PromptComponent {
     match get_node_version() {
         Some(node_version) => {
             const NODE_CHAR: &str = "⬢";
-            const SEGMENT_COLOR: Color = Color::Green;
+            const MODULE_COLOR: Color = Color::Green;
 
-            let mut segment = Segment::new("node");
-            segment.set_style(SEGMENT_COLOR.bold());
+            let mut module = Module::new("node");
+            module.set_style(MODULE_COLOR.bold());
 
+            let symbol = module.new_segment("symbol");
+            symbol.set_value(NODE_CHAR);
+
+            let  version = module.new_segment("version");
             let formatted_version = node_version.trim();
-            segment.set_value(format!("{} {}", NODE_CHAR, formatted_version));
+            version.set_value(formatted_version);
 
-            Some(Box::new(segment))
+            Some(module)
         }
         None => None,
     }
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -18,7 +18,7 @@ pub fn segment(context: &Context) -> Option<Module> {
 
     match get_node_version() {
         Some(node_version) => {
-            const NODE_CHAR: &str = "⬢";
+            const NODE_CHAR: &str = "⬢ ";
             let module_color = Color::Green.bold();
 
             let mut module = Module::new("node");
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -27,7 +27,7 @@ pub fn segment(context: &Context) -> Option<Module> {
             let symbol = module.new_segment("symbol");
             symbol.set_value(NODE_CHAR);
 
-            let  version = module.new_segment("version");
+            let version = module.new_segment("version");
             let formatted_version = node_version.trim();
             version.set_value(formatted_version);
 
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -2,7 +2,7 @@ use ansi_term::Color;
 use std::path::PathBuf;
 use std::process::Command;
 
-use super::Segment;
+use super::{Segment, PromptComponent};
 use crate::context::Context;
 
 /// Creates a segment with the current Node.js version
@@ -11,7 +11,7 @@ use crate::context::Context;
 ///     - Current directory contains a `.js` file
 ///     - Current directory contains a `package.json` file
 ///     - Current directory contains a `node_modules` directory
-pub fn segment(context: &Context) -> Option<Segment> {
+pub fn segment(context: &Context) -> PromptComponent {
     let is_js_project = context.dir_files.iter().any(has_js_files);
     if !is_js_project {
         return None;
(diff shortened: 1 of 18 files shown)


sirMerr:
Looks awesome!

src/segment.rs
```diff
@@ -61,77 +53,70 @@ impl Segment {
...
+            Some(style) => style.paint(&self.value),
+            None => ANSIString::from(&self.value),
         }
-
-        segment_string
     }
 }
 
-type BoxedSegment = Option<Box<Segment>>;
+impl fmt::Display for SegmentAffix {
```
Snuggle: Could we please have a quick comment here for this section?

src/segment.rs
```diff
@@ -61,77 +53,70 @@ impl Segment {
...
-            suffix,
-        } = self;
-
-        let mut segment_string = String::new();
-
-        // Skip the prefix for the first segment
-        if let Some(prefix) = prefix {
-            segment_string += &prefix.output()
-        }
+impl fmt::Display for Segment {
```
Snuggle: What does this do?

matchai: [`std::fmt::Display`](https://doc.rust-lang.org/std/fmt/trait.Display.html) is the trait required to make a type printable. By defining it, we can tell Rust how to handle formatting the segment when trying to print it.

matchai:
> Is there a way to reduce duplication here or would that not make sense in this case?

The implementation of this trait is very slightly different between `Segment` and `SegmentAffix`, along with `Module` and `ModuleAffix`:
https://github.com/starship/starship/blob/ebd4338a43416d1524f5a1af51229d7e142320f6/src/module.rs#L89-L94
https://github.com/starship/starship/blob/ebd4338a43416d1524f5a1af51229d7e142320f6/src/module.rs#L151-L155

That being said, I don't believe it would be possible to reduce the duplication without creating a new shared trait between `Module` and `Segment`. I don't think that is worthwhile yet the with the little duplication.

src/segment.rs
matchai: This particular snippet is mostly boilerplate that is used when defining the `std::fmt::Display` implementation. The part that isn't boilerplate is just `self.ansi_string()`. 😄

Snuggle:
Okiedokie, yup!
There are a couple other sections that have a similar structure but a couple little differences, it probably isn't worthwhile. 😊

tests/directory.rs
```diff
@@ -142,11 +120,11 @@ fn truncated_directory_in_git_repo() -> io::Result<()> {
...
     Repository::init(&repo_dir).unwrap();
 
-    let expected = Segment::new("dir")
-        .set_value("src/meters/fuel-gauge")
-        .set_style(Color::Cyan.bold())
-        .output();
-    let actual = common::render_segment("dir", &dir);
+    let expected = format!("via {} ", Color::Cyan
+        .bold()
+        .paint("src/meters/fuel-gauge")
```
Multimo: digging the file paths haha 🚀

src/modules/nodejs.rs
```diff
@@ -20,15 +19,19 @@ pub fn segment(context: &Context) -> Option<Segment> {
...
-            const SEGMENT_COLOR: Color = Color::Green;
+            const MODULE_COLOR: Color = Color::Green;
 
-            let mut segment = Segment::new("node");
-            segment.set_style(SEGMENT_COLOR.bold());
+            let mut module = Module::new("node");
+            module.set_style(MODULE_COLOR.bold());
 
+            let symbol = module.new_segment("symbol");
+            symbol.set_value(NODE_CHAR);
```
Multimo: Since we are returning `self` with these methods we can use chaining to get rid of the intermediate vars. I think they distract a away from the builder pattern. we are really just mutating the module instance 👍 

```rust
module.new_segment("symbol").set_value(NODE_CHAR);
```

matchai: I was thinking the same thing. That would mean that `new_segment()` would have to return a `&mut Segment` instead of a `Segment`, but we eventually need to return an owned `Segment`.

In the end it would look like this:
```rust
let module = module.new_segment("symbol").set_value(NODE_CHAR).build();

Some(module)
```

`.build()` would be returning the instance of `Segment`.

matchai: On second thought, I could have the methods all pass around `mut Segment` instead of `&mut Segment`. 🤔
I wonder what the performance costs are.

matchai: ...But then we can't work with segments that were already added to `Module`s 
since we're no longer operating with references. 😖

matchai: No... Those remain references so it's fine.

...

I'll do a little thinking without typing to GitHub

Multimo: hmmm yeah i misread i understand now. I was thinking we were passing the module reference around instead of pointers to its segments. 

I guess another approach is you could have add a second argument to new_segment method or have the setters on module rather than the module affix. Not sure how much value there is in doing this tho? 

``` rust
new_segment(name, value)
update_segment(name, value, style)

// which could eventually lead to something like:
let module = Module::new("rust")
                .set_style(MODULE_COLOR.bold())
                .new_segment("symbol", RUST_CHAR)
                .new_segment("version", format_rustc_version(rust_version)));
```
or we could return a pointer to the segment with `new_segment` and return a pointer to the module with the `set_value` but im not sure thats the best approach.

matchai: Oh! You were talking specifically about the `new_segment()` methods. Got it! 😂

Multimo: pointers will probably best fast than passing structs around the call stack. My guess is that the differences are probably so minimal we wouldn't notice since they are not that big
.

Multimo: honestly i am fine with the current approach, just thought it could an interesting avenue to explore :

matchai: There will be cases where segments will themselves have prefixes and suffixes, which is why we're setting the value separately.

Do you see this as an improvement?
```rust
const NODE_CHAR: &str = "⬢";
let module_color = Color::Green.bold();

let mut module = Module::new("node");
module.set_style(module_color);

let formatted_version = node_version.trim();
module.new_segment("symbol").set_value(NODE_CHAR);
module.new_segment("version").set_value(formatted_version);

Some(module)
```

Multimo: > There will be cases where segments will themselves have prefixes and suffixes, which is why we're setting the value separately.

I don't think i have a very clear picture of all the use cases but personally i feel its cleaner.

matchai: We could certainly make `new_segment_with_value()` or something along those lines.

matchai: > I don't think i have a very clear picture of all the use cases but personally i feel its cleaner.

To give context, here's an example of a module that will have a segment with a prefix and suffix:

```
[kubectl prefix] [kubectl symbol] [kubectl context] 
       at               ☸️                aks1

[kubectl namespace prefix] [kubectl namespace] [kubectl namespace suffix]
             (                 kube-system                  )
```
image

To be honest, I may be creating this API a little too defensively. 😅

matchai: I guess there won't ever be a situation where a segment _won't_ have a value.
What you're suggesting definitely seems like a better approach. 😄
I'll go ahead and add that.

PR #44, opened 2019-05-02
Add Go version module

### LINKED ISSUE #43: Implement the prompt module for Go version (🐹)
## Feature Request
Implement the prompt segment containing the currently running Go version.
Use the hamster emoji (🐹) as the symbol for the module, and have a cyan module color.

Should show if:
- The current directory contains
	- go.mod
	- Godeps
	- glide.yaml
	- *.go
	- Gopkg.yml
	- Gopkg.lock
- The current directory is $GOPATH

The implementation used in spaceship is a good reference to have:
https://github.com/denysdovhan/spaceship-prompt/blob/d9f25e14e7bec4bef223fd8a9151d40b8aa868fa/sections/golang.zsh

johnletey:
#### Motivation and Context

Closes #43 

#### Types of changes

Shot 2019-05-02 at 09 42 21

#### Checklist:

### DIFF
--- a/src/modules/go.rs
+++ b/src/modules/go.rs
@@ -9,7 +9,8 @@ use super::{Context, Module};
 /// Will display the Python version if any of the following criteria are met:
 ///     - Current directory contains a `.go` file
 ///     - Current directory contains a `go.mod` file
-///     - Current directory contains a `Godeps` file
+///     - Current directory contains a `go.sum` file
+///     - Current directory contains a `Godeps` directory
 ///     - Current directory contains a `glide.yaml` file
 ///     - Current directory contains a `Gopkg.yml` file
 ///     - Current directory contains a `Gopkg.lock` file
--- a/src/modules/go.rs
+++ b/src/modules/go.rs
@@ -1,0 +1,91 @@
+use ansi_term::Color;
+use std::path::PathBuf;
+use std::process::Command;
+
+use super::{Context, Module};
+
+/// Creates a segment with the current Python version
+///
+/// Will display the Python version if any of the following criteria are met:
+///     - Current directory contains a `.go` file
+///     - Current directory contains a `go.mod` file
+///     - Current directory contains a `Godeps` file
+///     - Current directory contains a `glide.yaml` file
+///     - Current directory contains a `Gopkg.yml` file
+///     - Current directory contains a `Gopkg.lock` file
+///     - Current directory is $GOPATH
+pub fn segment(context: &Context) -> Option<Module> {
+    let is_go_project = context.dir_files.iter().any(has_go_files);
+    if !is_go_project {
+        return None;
+    }
+
+    match get_go_version() {
+        Some(go_version) => {
+            const GO_CHAR: &str = "🐹 ";
+            let module_color = Color::Cyan.bold();
+
+            let mut module = Module::new("go");
+            module.set_style(module_color);
+
+            let formatted_version = format_go_version(go_version);
+            module.new_segment("symbol", GO_CHAR);
+            module.new_segment("version", formatted_version);
+
+            Some(module)
+        }
+        None => None,
+    }
+}
+
+fn has_go_files(dir_entry: &PathBuf) -> bool {
+    // TODO: Check to see if the current directory is $GOPATH
+    let is_go_mod =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "go.mod" };
+    let is_godeps =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "Godeps" };
+    let is_glide_yaml =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "glide.yaml" };
+    let is_go_file =
+        |d: &PathBuf| -> bool { d.is_file() && d.extension().unwrap_or_default() == "go" };
+    let is_gopkg_yml =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "Gopkg.yml" };
+    let is_gopkg_lock =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "Gopkg.lock" };
+
+    is_go_mod(&dir_entry)
+        || is_godeps(&dir_entry)
+        || is_glide_yaml(&dir_entry)
+        || is_go_file(&dir_entry)
+        || is_gopkg_yml(&dir_entry)
+        || is_gopkg_lock(&dir_entry)
+}
+
+fn get_go_version() -> Option<String> {
+    match Command::new("go").arg("version").output() {
+        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
+        Err(_) => None,
+    }
+}
+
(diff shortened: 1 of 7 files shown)


src/modules/go.rs
```diff
@@ -0,0 +1,91 @@
...
+        None => None,
+    }
+}
+
+fn has_go_files(dir_entry: &PathBuf) -> bool {
+    // TODO: Check to see if the current directory is $GOPATH
+    let is_go_mod =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "go.mod" };
+    let is_godeps =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "Godeps" };
```
Multimo: I think Godeps might be a folder? https://github.com/tools/godep

src/modules/go.rs
```diff
@@ -0,0 +1,91 @@
...
+            Some(module)
+        }
+        None => None,
+    }
+}
+
+fn has_go_files(dir_entry: &PathBuf) -> bool {
+    // TODO: Check to see if the current directory is $GOPATH
+    let is_go_mod =
+        |d: &PathBuf| -> bool { d.is_file() && d.file_name().unwrap_or_default() == "go.mod" };
```
Multimo: we could maybe add `go.sum` for good measure :)

src/modules/go.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+}
+
+fn format_go_version(go_stdout: String) -> String {
+    let mut trimmed_version = go_stdout
+        .trim_start_matches("go version go")
+        .trim()
+        .to_string();
+    let offset = &trimmed_version.find(' ').unwrap();
+    let formatted_version: String = trimmed_version.drain(..offset).collect();
```
Multimo: Rust has made me question whether strings are just hard to manipulate or if its a rust thing haha 👍

src/modules/mod.rs
```diff
@@ -17,6 +18,7 @@ pub fn handle(module: &str, context: &Context) -> Option<Module> {
         "node" | "nodejs" => nodejs::segment(context),
         "rust" | "rustlang" => rust::segment(context),
         "python" => python::segment(context),
+        "go" => go::segment(context),
```
Multimo: could maybe add "golang" as well?

johnletey:
@Multimo Thanks for the review ... I still need to check whether or not the current directory is inside `$GOPATH` ... but I'll definitely make your suggestions

src/modules/go.rs
johnletey: @Multimo Do I need to address anything here?

johnletey:
@matchai On it

tests/Dockerfile
```diff
@@ -9,15 +9,31 @@ RUN curl https://raw.githubusercontent.com/creationix/nvm/master/install.sh | ba
   && nvm use default
 ENV PATH /root/.nvm/versions/node/v$NODE_VERSION/bin:$PATH
 # Check that Node.js was correctly installed
-RUN node --version 
+RUN node --version
+
+# Install Go
```
matchai: Heck! That's a long installation process.
I wonder if we can trim this down some.

As I understand it https://github.com/moovweb/gvm is a bit more of a staple in the Go ecosystem than goenv. It may take considerably less to install. 🤔

tests/go.rs
```diff
@@ -0,0 +1,134 @@
+use ansi_term::Color;
```
matchai: This is more a note to myself. The way that we integration test, we don't test for a case where the module _shouldn't_ appear. As it exists, `common:render_module()` wouldn't let us check that, but it'd be worth making it work that way.

@johnletey is this something you want to tackle, or should I address this another time?

johnletey: @matchai If you don't mind, I'd prefer if you addressed this another time

src/modules/go.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+fn get_go_version() -> Option<String> {
+    match Command::new("go").arg("version").output() {
+        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
+        Err(_) => None,
+    }
+}
+
+fn format_go_version(go_stdout: String) -> String {
```
mehcode: Disclaimer: The comment from @Multimo is what had me stop and offer help here. Plus it was a fun distraction.

---

Let's call the `format_go_version` currently in the patch, `format_go_version_1`. Running it through `cargo bench` on my box produces: 177 ns.

---

First let's remove an unnecessary allocation. Instead of creating a temporary `String` at line 77 after `trim()` we can leave it as a slice until the end:

```rust
let mut trimmed_version = go_stdout
    .trim_start_matches("go version go")
    .trim();

let offset = trimmed_version.find(' ').unwrap();
let formatted_version: String = trimmed_version[..offset].into();
```

This cleans this up a bit and improves performance to 131 ns.

---

Next, let's remove the panic by assuming if we can't find `' '` the string cuts off there.

```rust
let offset = trimmed_version.find(' ').unwrap_or_else(|| trimmed_version.len());
```

This decreases the performance to 133 ns but panics are bad.

---

Next let's try changing this up a bit with `split` 

```rust
go_stdout
    // split into ["", "1.12.4 linux/amd64"]
    .splitn(2, "go version go")
    // return "1.12.4 linux/amd64"
    .nth(1)?
    // split into ["1.12.4", "linux/amd64"]
    .split_whitespace()
    // return "1.12.4"
    .next()
    // format as "v1.12.4"
    .map(|version| format!("v{}", version))
```

This brings us down to **105 ns**. You'll notice that now we are also accounting for the failure case of a `go version` returning garbage output. This function would return `None` in that case.

---

We can still squeeze a bit more performance out of this. `format!` is currently relatively slow in Rust for such simple tasks as string concatenation.

```rust
// Use with_capacity to pre-allocate the string so we can guarantee there is only 1 allocation
let mut buf = String::with_capacity(version.len() + 1);
buf.push('v');
buf.push_str(version);
buf
```

This increases the performance to **77 ns** or **2.3x** faster than what we started. 

That was a fun exercise.

src/modules/go.rs
```diff
@@ -0,0 +1,93 @@
...
+    is_go_mod(&dir_entry)
+        || is_go_sum(&dir_entry)
+        || is_godeps(&dir_entry)
+        || is_glide_yaml(&dir_entry)
+        || is_go_file(&dir_entry)
+        || is_gopkg_yml(&dir_entry)
+        || is_gopkg_lock(&dir_entry)
+}
+
+fn get_go_version() -> Option<String> {
```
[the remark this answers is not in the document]
johnletey: @mehcode Thank you for the review, I have made your changes

src/modules/go.rs
```diff
@@ -0,0 +1,93 @@
...
+///     - Current directory contains a `glide.yaml` file
+///     - Current directory contains a `Gopkg.yml` file
+///     - Current directory contains a `Gopkg.lock` file
+pub fn segment(context: &Context) -> Option<Module> {
+    let is_go_project = context.dir_files.iter().any(has_go_files);
+    if !is_go_project {
+        return None;
+    }
+
+    match get_go_version() {
```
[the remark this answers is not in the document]
johnletey: @mehcode Thank you for the review, I have made your changes

PR #45, opened 2019-05-03
Implement the git status module

### LINKED ISSUE #11: Implement the prompt segment for git branch name and status
Port the functionality present in spaceship for the `git_branch` and `git_status` segments.

image

The following files can be used as reference for the spaceship implementation:
https://github.com/denysdovhan/spaceship-prompt/blob/master/sections/git.zsh
https://github.com/denysdovhan/spaceship-prompt/blob/master/sections/git_branch.zsh
https://github.com/denysdovhan/spaceship-prompt/blob/master/sections/git_status.zsh

To retreive repo information, we will be using [`git2`](https://github.com/rust-lang/git2-rs), which is already currently used in the `directory` section.

matchai:
#### Description

- Add `branch_name` to `Context`
- Add logger and debug logging for the repo state
- Add git status segment

#### Motivation and Context

Closes #11 

After doing some benchmarking of the git_status module, it appeared that the performance difference between using `git2-rs` and the locally installed version of Git was within a margin of error, so I opted to use `git2-rs` to remove the need for git as a dependency.

#### Types of changes

image

#### Checklist:


src/lib.rs
```diff
@@ -1,4 +1,7 @@
 // Lib is present to allow for benchmarking
+#[macro_use]
+extern crate log;
```
mehcode: As this is edition 2018, you can instead use `log::info!("Hello")` (or import `use log::info`) everywhere and not worry about `#[macro_use]` here. Personally I find I like seeing `log::error!("danger")` instead of `error!("danger")` anyway.

src/main.rs
```diff
@@ -1,6 +1,10 @@
 #[macro_use]
 extern crate clap;
 
+#[macro_use]
+extern crate log;
+
+extern crate pretty_env_logger;
```
mehcode: Likewise you can delete all these extern crate declarations.

src/modules/git_status.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+fn get_repo_status(repository: &Repository) -> Option<Status> {
+    let mut status_options = git2::StatusOptions::new();
+    status_options.include_untracked(true);
+
+    let repo_file_statuses = repository.statuses(Some(&mut status_options)).ok()?;
+
+    // Statuses are stored as bitflags, so use BitOr to join them all into a single value
+    let repo_status = repo_file_statuses.iter().fold(Status::empty(), |acc, x| acc | x.status());
```
mehcode: This operation is actually implemented in bitflags for you. Just `repo_file_statuses.iter().collect()`. The `.collect()` iterator adapter is controlled through `FromIterator` trait (which bitflags implements for all of the flag types it generates).

src/lib.rs
matchai: Thanks for the heads up @mehcode! 😄

src/modules/git_status.rs
mehcode: Hm. I see. I feel `Statuses::status` method should probably exist but I'm not the `git2` guys.

---

You can adapt my snippet like this for it to work:

```rust
let repo_status: Status = repo_file_statuses.iter().map(|e| e.status()).collect();
```

You're not saving much on code complexity but its declarative so I think its still a small improvement.

mehcode:
More of a general comment but `libgit2` is slow and a bit bloated. I'd recommend shelling out to `git` directly. Furthermore you're locked in to a very single-threaded execution model as `libgit2` is not thread safe. 

You probably want to change `prompt_order.iter()` to `prompt_order.par_iter()` (in `print.rs`) with rayon to calculate the prompt in parallel but that's not possible with a `libgit2` type on the `Context` object.

src/modules/git_status.rs
matchai: Excellent. Sounds good 👍

matchai:
> You probably want to change prompt_order.iter() to prompt_order.par_iter() (in print.rs) with rayon to calculate the prompt in parallel but that's not possible with a libgit2 type on the Context object.

Thanks for the heads up @mehcode! The plan was to eventually swap in `par_iter()` in `print.rs`. I had no clue `libgit2` would make that impossible though. 😧

To provide context for others reading this, here is where it is detailed that `libgit2` is not thread safe: https://github.com/libgit2/libgit2/blob/master/docs/threading.md

And here's the error surfaced by Rust's compiler when Rayon is eventually added:
image

I guess it's back to the drawing board with this prompt segment. 😅

src/modules/git_status.rs
```diff
@@ -0,0 +1,122 @@
...
+    module.get_prefix().set_value("[").set_style(module_style);
+    module.get_suffix().set_value("] ").set_style(module_style);
+    module.set_style(module_style);
+
+    let ahead_behind = get_ahead_behind(&repository, &branch_name);
+    log::debug!("Repo ahead/behind: {:?}", ahead_behind);
+    let stash_object = repository.revparse_single("refs/stash");
+    log::debug!("Stash object: {:?}", stash_object);
+    let repo_status = get_repo_status(&repository);
+    log::debug!("Repo status: {:?}", repo_status);
```
Multimo: This is cool will make debugging a lot easier 👍 Just want to double check if its still in there on purpose :)

src/context.rs
```diff
@@ -36,15 +37,20 @@ impl<'a> Context<'a> {
...
 
-        let repo_root: Option<PathBuf> = Repository::discover(&current_dir)
-            .ok()
+        let repository = Repository::discover(&current_dir).ok();
+        let repo_root = repository
+            .as_ref()
             .and_then(|repo| repo.workdir().map(|repo| repo.to_path_buf()));
+        let branch_name = repository
+            .as_ref()
+            .and_then(|repo| get_current_branch(&repo));
```
Multimo: `as_ref` is neat, didn't know you could do that 👍

src/modules/git_status.rs
matchai: Yep, it's in there on purpose. 😄
I'd like to eventually go back and add detailed trace and debug logs to make it clear what the prompt is doing under the hood.

PR #55, opened 2019-05-15
Fix CI to work with changes to Azure Pipelines build agent

matchai:
#### Description

so... many... commits... 😴 

- Break up CI process into smaller template files
- Update `$PATH` to include the new cargo installation directory
- Change Python version to use a version supported by Azure Pipelines

Fixing tests:
- Windows filepaths would alternate between `/` and `\`. Now using `/` for all path separation in the prompt.
- Tempfile was not creating tempdirs in a folder where we have write permissions, so now we create tempdirs in `$HOME`, where we do have permissions

#### Motivation and Context

CI is failing 😢 

#### Types of changes

#### Checklist:


ci/azure-setup-test-env.yml
```diff
@@ -0,0 +1,16 @@
...
+    displayName: "Install a fixed version of Node"
+  # Install Go
+  - task: GoTool@0
+    inputs:
+      versionSpec: "1.10"
+    displayName: "Install a fixed version of Go"
+  # Install Python
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: "3.6.8"
```
johnletey: Was there something wrong with Python version 3.7.2? The Rust was failing, not the Python

tests/directory.rs
```diff
@@ -98,7 +98,7 @@ fn truncated_directory_in_root() -> io::Result<()> {
 #[test]
 #[ignore]
 fn git_repo_root() -> io::Result<()> {
-    let tmp_dir = TempDir::new()?;
+    let tmp_dir = TempDir::new_in(home_dir().unwrap())?;
     let repo_dir = tmp_dir.path().join("rocket-controls");
```
johnletey: Instead of creating `tmp_dir`, can't we just say that `let repo_dir = Path::new("$HOME/rocket-controls")`?

tests/directory.rs
```diff
@@ -117,7 +117,7 @@ fn git_repo_root() -> io::Result<()> {
 #[test]
 #[ignore]
 fn directory_in_git_repo() -> io::Result<()> {
-    let tmp_dir = TempDir::new()?;
+    let tmp_dir = TempDir::new_in(home_dir().unwrap())?;
     let repo_dir = tmp_dir.path().join("rocket-controls");
```
johnletey: Same here

ci/azure-setup-test-env.yml
youssefhabri: I think some tests where failing because Python v3.7.2 is not available.
```
Version spec 3.7.2 for architecture x64 did not match any version in Agent.ToolsDirectory.
Versions in /opt/hostedtoolcache:

2.7.16 (x64)
3.4.8 (x64)
3.5.7 (x64)
3.6.8 (x64)
3.7.3 (x64)
```

johnletey: Got it ... didn't know what was going on. Thanks for explaining @youssefhabri!

tests/directory.rs
matchai: To make the tests atomic, we use unique paths whenever we're performing filesystem operations that create or modify something.
Otherwise, having the directory already exist could cause collisions. For example, because we initialize the git repo, using the same directory to initialize a git repo again throws an error.

johnletey: Understand ... thanks for explaining. Seems fine to me then!

PR #56, opened 2019-05-15
feat: Implement the prompt module for username

### LINKED ISSUE #48: Implement the prompt module for username
## Feature Request

Implement the prompt module containing the name of the current user.

Should show if:
- The current user isn't the same as the one that is logged in (`$LOGNAME != $USER`)
- The current user is root (`UID = 0`)
- The user is currently connected as an SSH session (`$SSH_CONNECTION`)

The color of the module should be yellow, unless the user is root. In that case, the module color should be red.

#### Example

image

The implementation used in spaceship is a good reference to have:
https://github.com/denysdovhan/spaceship-prompt/blob/master/sections/user.zsh

youssefhabri:
#### Description
Implement the prompt module for username, which should show the username if:

- The current user isn't the same as the one that is logged in (`$LOGNAME != $USER`)
- The current user is root (`UID = 0`)
- The user is currently connected as an SSH session (`$SSH_CONNECTION`)

#### Motivation and Context

Closes #48

#### Types of changes

#### Checklist:


src/modules/username.rs
```diff
@@ -0,0 +1,49 @@
...
+///
+/// Will display the Node.js version if any of the following criteria are met:
+///     - The current user isn't the same as the one that is logged in ($LOGNAME != $USER)
+///     - The current user is root (UID = 0)
+///     - The user is currently connected as an SSH session ($SSH_CONNECTION)
+pub fn segment(_context: &Context) -> Option<Module> {
+    let user = env::var("USER").unwrap_or("".to_string());
+    let logname = env::var("LOGNAME").unwrap_or("".to_string());
+    let ssh_connection = env::var("SSH_CONNECTION").unwrap_or("".to_string());
+    let uid = get_uid().unwrap_or(1000);
```
youssefhabri: I defaulted to uid(1000) instead of just `unwrap`, since I'm not what to do instead of just using uid(1000) when it fails to get the uid.

youssefhabri: I defaulted to uid(1000) instead of just unwrap, since I'm not what to do instead of just using uid(1000) when it fails to get the uid.

youssefhabri: also, is it ok to default the other `env::var` to empty strings?

src/modules/directory.rs
```diff
@@ -39,6 +39,8 @@ pub fn segment(context: &Context) -> Option<Module> {
     let truncated_dir_string = truncate(dir_string, DIR_TRUNCATION_LENGTH);
     module.new_segment("path", truncated_dir_string);
 
+    module.get_prefix().set_value("");
```
youssefhabri: @matchai This change causes the directory.rs tests to fail. was the `via` prefix needed, or was it just the default?

matchai: `via ` was the default, but I believe it should be `in `.
We may need to add some logic to skip the prefix of the first segment that isn't `None`. At the moment, I believe it skips the first prefix, regardless of whether it's a `None` or a `Some`.

src/modules/username.rs
matchai: I think in this situation, it may make more sense to have them as `Option<String>` values rather than defaulting to empty strings.

```rust
// This will turn a `Result` into an `Option`
let user = env::var("USER").ok();
```

src/modules/username.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+fn get_uid() -> Option<u32> {
+    match Command::new("id").arg("-u").output() {
+        Ok(output) => {
+            let uid = String::from_utf8(output.stdout)
+                .unwrap()
+                .trim()
+                .parse::<u32>()
+                .unwrap();
```
youssefhabri: How about this? I'm always worried about using `unwrap`. But I'm not sure what's the best way to handle this rn

matchai: I'd take a look at using `try!()` or ts shorthand `?` if you want it to return with a value of `None` if the value doesn't exist. If instead, you want to operate on the `Option` without unwrapping it, take a look at [`map()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map) or [`and_then()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then).

src/modules/directory.rs
matchai: Looking it over, it seems [we strip out `None` values before skipping the first prefix](https://github.com/starship/starship/blob/master/src/print.rs#L38-L46), so no additional work should be needed here.

youssefhabri: Just to clarify, what I meant is that the `directory` segment defaults to the `via ` prefix and causes this to happen:
Screenshot from 2019-05-16 06-16-29
So, should I keep the change I made and have the `username` segment add the `in ` suffix, or should the `directory` segment be the one adding the prefix `in `? which one would make more sense?

matchai: Directory should be the one with the `in ` prefix since it would make sense with other prompt orders and configurations.

src/modules/username.rs
```diff
@@ -0,0 +1,43 @@
...
+///     - The current user isn't the same as the one that is logged in ($LOGNAME != $USER)
+///     - The current user is root (UID = 0)
+///     - The user is currently connected as an SSH session ($SSH_CONNECTION)
+pub fn segment(_context: &Context) -> Option<Module> {
+    let user = env::var("USER").ok();
+    let logname = env::var("LOGNAME").ok();
+    let ssh_connection = env::var("SSH_CONNECTION").ok();
+    let uid = get_uid();
+
+    if user != logname || uid == Some(0) || ssh_connection.is_some() {
```
matchai: Because `get_uid()` spawns a new process, let's try and make that the last check in this series of checks.

```suggestion
    if user != logname || ssh_connection.is_some() || is_root() {
```

In this example, `is_root()` would perform the full uid check so that we don't need to `get_uid()` if either of the other two are successful. You can maybe also change the module color within that function.

src/modules/username.rs
```diff
@@ -0,0 +1,49 @@
...
+        module.new_segment("username", user?);
+
+        return Some(module);
+    }
+
+    None
+}
+
+fn get_uid() -> Option<u32> {
+    match Command::new("id").arg("-u").output() {
```
matchai: It appears that `$UID` is an environment variable in bash and ZSH. Could we perhaps check for the environment variable and use `id -u` has a fallback if it doesn't exist?

youssefhabri: Ok, I'll try that and see how it goes

matchai:
@all-contributors please add @youssefhabri for code

PR #62, opened 2019-05-21
Add config

### LINKED ISSUE #10: RFC: Prompt configuration
## Summary

Using a TOML configuration file `~/spaceship.toml` instead of how environment variables were used in spaceship.

## Motivation

I would like to reconsider how we should be doing prompt configuration. Historically we have had all configuration done with environment variables:

```bash
SPACESHIP_PROMPT_ORDER=(
  dir,
  ruby
)

# DIR
SPACESHIP_DIR_PREFIX='' # disable directory prefix, cause it's not the first section
SPACESHIP_DIR_TRUNC='1' # show only last directory

# RUBY
SPACESHIP_RUBY_PREFIX="ruby:("
SPACESHIP_RUBY_SUFFIX=") "
SPACESHIP_RUBY_SYMBOL=""
```

This has lead to a few issues:
- Increasing granularity of configuration meant longer and longer variable names
- Unable to use more expressive data types in configuration
- Bloating of environment variables list (spacefish is >50% of mine)

Here's an example of that same configuration in TOML:
```toml
segment_order = [
  "dir",
  "ruby"
]

[dir]
prefix = ""
truncation = 1

[ruby]
prefix = "ruby:("
suffix = ") "
symbol = ""
```

It didn't feel like we had enough nesting to make the complexity of JSON worthwhile as a config file format:
<details>
<summary>JSON Example</summary>

```json
{
	"segmentOrder": [
		"dir",
		"ruby"
	],

	"dir": {
		"prefix": "",
		"truncation": 1
	},
	"ruby": {
		"prefix": "ruby:(",
		"suffix": ") ",
		"symbol": ""
	}
}
```

</details>

YAML also looks like a very promising option. It is arguably easier to read than TOML, but begins to be difficult to work with when you have any nesting, as it's completely indentation-based.
<details>
<summary>YAML Example</summary>

```yaml
segment_order: 
  - dir
  - ruby

dir:
  prefix: ""
  truncation: 1

ruby:
  prefix: "ruby:("
  suffix: ") "
  symbol: ""
```

</details>

Where I think light-nesting will come to be useful is if we were to support individual styling of prefixes and suffixes, like in this example:

```toml
[ruby]
symbol = ""
color = "cyan"

  [ruby.prefix]
  color = "magenta"
  text = "ruby:("

  [ruby.suffix]
  color = "magenta"
  text = ") "
```

The same thing in YAML becomes difficult to parse and write:

```yaml
ruby:
  symbol: ""
  color: "cyan"
  prefix:
    color: "magenta"
    text: "ruby:("
  suffix:
    color: "magenta"
    text: ") "
```

## Drawbacks

- TOML is not the most well-recognized of configuration file formats, so there will need to be some clear documentation written around config files.
- Error messaging will need to be added in the event that the TOML file is unable to be parsed. This was not an issue in the past when working with environment variables.

### LINKED ISSUE #27: Add starship prompt configuration file
Implement starship prompt configuration based on the conversation that took place in #10. 

### Feature Design
- Read a `starship.toml` file placed in `$XDG_CONFIG_HOME`
	- If `$XDG_CONFIG_HOME` is not set, read `$HOME/starship.toml`
- User configuration is stored in the `Context` struct for use across segments
- Use `Segment.name` as TOML table keys

### PR should include
- Demonstrate configuration by having `char` symbol be replaceable
- Add integration test for char being replaced

matchai:
#### Description

This is going to remain a draft PR until I add more tests. ✍️ 

##### Relevant to the PR:
- Create `Config` struct that is added to `Context` when initialized
- `Context` now has a method for creating modules. This allows us to provide modules with a reference to the configuration specific to that module
  - Because there's now a reference in `Module`, everything that returns a module needs lifetimes 😓

##### Drive-by improvements:
- Remove `SegmentAffix`. A segment prefix and suffix can just themselves be segments. Was complicated for little reason 😅 
- Refactor the package module to be simpler
- Split out `read_file` into `utils.rs

#### Motivation and Context

Closes #27 and closes #10 

#### Types of changes

#### Checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,25 @@
+use crate::utils;
+
+// use std::collections::HashMap;
+use dirs::home_dir;
+
+static config: toml::value::Table = toml::value::Table::new();
+// static config: HashMap<String, ModuleConfig> = HashMap::new();
+
+// enum ModuleConfig {
+//     ConfigValue(HashMap<String, String>),
+//     SegmentConfig(HashMap<String, HashMap<String, String>>)
+// }
+
+pub fn initialize() -> Option<()> {
+    let file_path = home_dir()?.join(".config/starship.toml");
+    let toml_content = utils::read_file(&file_path.to_str()?).ok()?;
+
+    config = toml::from_str(&toml_content).ok()?;
+
+    Some(())
+}
+
+pub fn get_config(config_value: &str, default: &str) -> String {
+    default.to_string()
+}
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,25 +1,33 @@
 use crate::utils;
 
-// use std::collections::HashMap;
 use dirs::home_dir;
 
-static config: toml::value::Table = toml::value::Table::new();
-// static config: HashMap<String, ModuleConfig> = HashMap::new();
-
-// enum ModuleConfig {
-//     ConfigValue(HashMap<String, String>),
-//     SegmentConfig(HashMap<String, HashMap<String, String>>)
-// }
-
-pub fn initialize() -> Option<()> {
-    let file_path = home_dir()?.join(".config/starship.toml");
-    let toml_content = utils::read_file(&file_path.to_str()?).ok()?;
-
-    config = toml::from_str(&toml_content).ok()?;
-
-    Some(())
+pub struct Config {
+    data: toml::value::Table,
 }
 
-pub fn get_config(config_value: &str, default: &str) -> String {
-    default.to_string()
+impl Config {
+    pub fn initialize() -> Config {
+        if let Some(file_data) = Config::config_from_file() {
+            return Config {
+                data: file_data
+            };
+        }
+
+        Config {
+            data: toml::value::Table::new(),
+        }
+    }
+
+    fn config_from_file() -> Option<toml::value::Table> {
+        let file_path = home_dir()?.join(".config/starship.toml");
+        let toml_content = utils::read_file(&file_path.to_str()?).ok()?;
+
+        let config = toml::from_str(&toml_content).ok()?;
+        Some(config)
+    }
+
+    pub fn get_config(config_value: &str, default: &str) -> String {
+        default.to_string()
+    }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -7,6 +7,7 @@ pub struct Config {
 }
 
 impl Config {
+    /// Initialize the Config struct
     pub fn initialize() -> Config {
         if let Some(file_data) = Config::config_from_file() {
             return Config {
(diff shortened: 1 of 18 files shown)


src/config.rs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        Config {
+            data: toml::value::Table::new(),
+        }
+    }
+
+    /// Create a config from a starship configuration file
+    fn config_from_file() -> Option<toml::value::Table> {
+        let file_path = home_dir()?.join(".config/starship.toml");
```
Multimo: How does one config a config :P Do you think we there is a case for having the config somewhere other than the home dir?

src/context.rs
```diff
@@ -64,6 +71,10 @@ impl<'a> Context<'a> {
         dir
     }
 
+    pub fn new_module(&self, name: &str) -> Module {
+        Module::new(name, self.config.get_module_config(name))
```
Multimo: 👍

src/modules/python.rs
```diff
@@ -10,15 +10,10 @@ use super::{Context, Module};
...
 ///     - Current directory contains a `requirements.txt` file
 ///     - Current directory contains a `pyproject.toml` file
-pub fn segment(context: &Context) -> Option<Module> {
+pub fn segment<'a>(context: &'a Context) -> Option<Module<'a>> {
     let is_py_project = context
         .new_scan_dir()
-        .set_files(&[
-            "requirements.txt",
-            ".python-version",
-            "pyproject.toml",
```
Multimo: 😬

src/config.rs
```diff
@@ -0,0 +1,42 @@
...
+        let toml_content = utils::read_file(&file_path.to_str()?).ok()?;
```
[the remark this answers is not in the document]
matchai: Good call! 👍 
Not something we have to address in this PR, but I think we may want to think about how we want to elegantly show warnings and alerts (configuration error, update available, etc). Because every prompt instance is stateless, it would print the error _every time_. 🤔 

What would be the ideal way to inform the user that something is wrong with their configuration without getting in their way as they address it?
IMO, the ideal solution would be to print a warning at the start of each session (3 hrs since the prompt was last used would be a fresh instance, for instance).

src/config.rs
matchai: This can be handled with an environment variable (`$STARSHIP_CONFIG`).
As was suggested by @Snuggle in https://github.com/starship/starship/issues/10#issuecomment-484329229, we should use `$XDG_CONFIG_HOME`, if available, before resorting to `~/.config`.
This was something I was thinking we would address in a future PR.

PR #66, opened 2019-05-26
Fix python version bug

### LINKED ISSUE #53: Python version not showing on python versions < 3.4
#### Current Behavior
Starship seems to detect a python project, however in my case, it seems to not be able to detect the version, even tho when I run `python --version` I get `Python 2.7.15rc1`

#### Expected Behavior
It should show the Python version in the prompt

Screenshot from 2019-05-14 21-03-16

#### Environment
- Starship version: 0.1.0
- Shell type: fish
- Shell version: fish, version 2.7.1
- Shell plugin manager: oh-my-fish
- Terminal emulator: Tilix
- Operating system: elementary OS 5.0

```bash
set -U fish_user_abbreviations

export EDITOR="nvim"

export BAT_THEME="TwoDark"
export FLUTTER_HOME=/opt/flutter

export WINEARCH=win32
export WINEPREFIX=$HOME/.wine_x86

export GOPATH=$HOME/Development/Golang

# Spacefish configuration
export SPACEFISH_DIR_TRUNC=0

# PATH configuration
set -x PATH $PATH $FLUTTER_HOME/bin
set -x PATH $PATH $HOME/.config/composer/vendor/bin
set -x PATH $PATH $HOME/.yarn/bin
set -x PATH $PATH $HOME/.cargo/bin

# Loading autojump if installed
if test -f $HOME/.autojump/share/autojump/autojump.fish
    source $HOME/.autojump/share/autojump/autojump.fish
end

# Loading abbreviations if existing
if test -f $HOME/.config/fish/abbreviations.fish
    source "$HOME/.config/fish/abbreviations.fish"
end

```

#### Starship Configuration
```bash
// Doesn't exists
```

johnletey:
#### Description

The Python version wasn't showing for versions < 3.4 (reported by @youssefhabri). The problem was that `python --version` for < 3.4 is outputting to stderr, but we were only checking stdout since for versions >= 3.4, that's where it outputs. I have fixed this bug.

#### Motivation and Context

Closes #53 

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -46,7 +46,14 @@ pub fn segment(context: &Context) -> Option<Module> {
 
 fn get_python_version() -> Option<String> {
     match Command::new("python").arg("--version").output() {
-        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
+        Ok(output) => {
+            let temp_version = String::from_utf8(output.stdout).unwrap();
+            if temp_version == "" {
+                Some(String::from_utf8(output.stderr).unwrap())
+            } else {
+                Some(temp_version)
+            }
+        }
         Err(_) => None,
     }
 }
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -47,11 +47,15 @@ pub fn segment(context: &Context) -> Option<Module> {
 fn get_python_version() -> Option<String> {
     match Command::new("python").arg("--version").output() {
         Ok(output) => {
-            let temp_version = String::from_utf8(output.stdout).unwrap();
-            if temp_version == "" {
-                Some(String::from_utf8(output.stderr).unwrap())
+            // We have to check both stdout and stderr since for Python versions
+            // < 3.4, Python reports to stderr and for Python version >= 3.5,
+            // Python reports to stdout
+            if output.stdout.is_empty() {
+                let stderr_string = String::from_utf8(output.stderr).unwrap();
+                Some(stderr_string)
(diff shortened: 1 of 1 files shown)


src/modules/python.rs
```diff
@@ -46,7 +46,14 @@ pub fn segment(context: &Context) -> Option<Module> {
 
 fn get_python_version() -> Option<String> {
     match Command::new("python").arg("--version").output() {
-        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
+        Ok(output) => {
```
matchai: I think we should try to avoid the `String::from_utf8()` if we don't need it.
`output.stdout` is a `Vec<u8>`, so we should be able to use `output.stdout.is_empty()`.


```rust
if stdout.is_empty() {
	let stdout_string = String::from_utf8(stdout).unwrap();
	return Some(stdout_string);
} else {
	let stderr_string = String::from_utf8(output.stderr).unwrap();
	return Some(stderr_string);
}
```

matchai: I'd also suggest adding a comment explaining why we are checking both stdout and stderr. 😄

johnletey: @matchai Done!

rahulrajpl:
I have also faced this issue. I permanently changed the 'python' command using 
~~~
sudo update-alternatives --install /usr/bin/python python /usr/bin/python3 10
~~~
and it worked 

before:
image

after:
image

PR #68, opened 2019-05-27
Use tarpaulin and Codecov for adding code coverage to project

matchai:
#### Description

Our code coverage is slipping because I've been a little lazy lately. I'd be nice to have visibility over where we are most lacking in coverage. 😄 

#### Motivation and Context

Closes #

#### Types of changes

#### Checklist:

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -34,7 +34,7 @@ RUN wget -O python.tar.xz "https://www.python.org/ftp/python/${PYTHON_VERSION%%[
 	\
 	&& cd /usr/src/python \
 	&& gnuArch="$(dpkg-architecture --query DEB_BUILD_GNU_TYPE)" \
-	&& make -j8 build_all \
+	&& ./configure \
 	&& make -j8 altinstall \
 	\
 	&& find /usr/local -depth \
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -34,7 +34,7 @@ RUN wget -O python.tar.xz "https://www.python.org/ftp/python/${PYTHON_VERSION%%[
 	\
 	&& cd /usr/src/python \
 	&& gnuArch="$(dpkg-architecture --query DEB_BUILD_GNU_TYPE)" \
-	&& ./configure \
+	&& ./configure --build="$gnuArch" \
 	&& make -j8 altinstall \
 	\
 	&& find /usr/local -depth \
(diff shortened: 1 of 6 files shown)


matchai:
I'm going to try to get this working in the starship Docker build since the process of installing `tarpaulin` increases linux test time from 4 mins to 12 mins.
I'll first see if I can get the Docker tests to build and publish the image for use in future tests. Hopefully that will really reduce the time needed in CI.

matchai:
#70 addresses the Docker stuff mentioned above.
I'll update the `Dockerfile` once that is merged so that coverage can be done in the docker tests.

tests/Dockerfile
```diff
@@ -1,38 +1,32 @@
-FROM rust:latest
+FROM rust
+
+# Install Tarpaulin (code coverage tool)
+# https://github.com/xd009642/tarpaulin
+COPY --from=xd009642/tarpaulin /usr/local/cargo/bin/cargo-tarpaulin /usr/local/cargo/bin/cargo-tarpaulin
 
 # Install Node.js
 ENV NODE_VERSION 12.0.0
-RUN curl https://raw.githubusercontent.com/creationix/nvm/master/install.sh | bash \
```
matchai: It looked like having `NODE_VERSION` set automatically caused `nvm` to install that version and set it to default. If you check the logs of older runs, we were installing Node multiple times. 😛

tests/Dockerfile
```diff
@@ -1,38 +1,32 @@
-FROM rust:latest
+FROM rust
+
+# Install Tarpaulin (code coverage tool)
+# https://github.com/xd009642/tarpaulin
+COPY --from=xd009642/tarpaulin /usr/local/cargo/bin/cargo-tarpaulin /usr/local/cargo/bin/cargo-tarpaulin
```
matchai: We copy in the binary directly from the built docker image of tarpaulin.
This seems much easier and faster than building it manually. 😄

tests/Dockerfile
```diff
@@ -1,38 +1,32 @@
...
-# RUN git clone https://github.com/wfarr/goenv.git $GOENV_ROOT
-RUN git clone https://github.com/syndbg/goenv.git $GOENV_ROOT
 ENV PATH $GOENV_ROOT/bin:$GOENV_ROOT/shims:$PATH
-RUN eval "$(goenv init -)"
-RUN mkdir -p $GO_ROOT
-ENV GOPATH $GO_ROOT
-ENV PATH $GO_ROOT/bin:$PATH
-RUN goenv install $GO_VERSION
-RUN goenv global $GO_VERSION
+RUN git clone https://github.com/syndbg/goenv.git $GOENV_ROOT \
```
matchai: Creating less docker layers by bundling these operations into one command makes for a smaller image.

PR #70, opened 2019-05-28
Use starship-test docker image for running integration tests

matchai:
#### Description

- Publish docker image used in testing
- Update `integration_test` to pull the image and build, using the image as its cache
- Publish the built image after successful test runs on master

Side note: I'm no Docker wizard, so our Docker image is pretty huge, which still makes tests still take a while. If anyone wants to try slimming it down, please feel free! 😄 

#### Motivation and Context

Speed up docker tests in CI by 3x! 🚗💨 

#### Types of changes

#### Checklist:


ci/azure-test-docker.yml
```diff
@@ -0,0 +1,16 @@
...
+  steps:
+  - script: ./integration_test
+    displayName: Run integration test within Docker
+  
+  - script: |
+      docker login -u $(dockerUsername) -p $(dockerPassword)
+      docker push starshipcommand/starship-test
+    # Only push new image if on master and build is passing
+    condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
+    displayName: Push image to dockerhub
```
johnletey: It's just a tiny nitpick ... but can you indent these lines to match up with the above lines 😄.

matchai: Yep, that indentation is needed.
YAML can be weird-looking. 😛

matchai:
Looks like there's a small issue here.
Because the build process is being done on an `amd64` architecture, the docker image has to be rebuilt on all other architectures.

Builds may be faster on CI, but may not be as fast on your own computer. 😕
I'll see if it'd be easy to add multi-architecture builds, but I guess that can be addressed at a later time.

ci/azure-test-docker.yml
johnletey: Understand

PR #71, opened 2019-05-30
Refactor integration tests

matchai:
#### Description

Don't be scared off by the number of files changed. I simply moved the test files, and GitHub doesn't play nicely. 😅 

- Create subcommands to be able to print modules independently
	- `starship prompt` will print the full prompt
	- `starship module <MODULE_NAME>` will print a specific module
		e.g. `starship module python`
	- Added `--path` flag to print the prompt or modules without being in a specific directory
	- Added `--status` flag to provide the status of the last command, instead of requiring it as an argument
- Refactored integration tests to be end-to-end tests, since there was no way in integration tests to set the environment variables for a specific command, which was required for the `username` module
- Moved e2e tests to `tests/testsuite` to allow for a single binary to be built
	- Tests will build/run faster
	- No more false positives for unused functions
- Added tests for `username`
- Removed codecov + tarpaulin 😢 

**Removing Codecov + tarpaulin**
The decision was split between more thorough end-to-end testing or having code coverage tooling. For the time being, I think the confidence provided by e2e tests outweighs the value of having code coverage.

This may change if we find a reliable mocking solution that allows most of our tests to be integration tests.
	

#### Motivation and Context

Adding integration tests was difficult. Hopefully, this will be easier to maintain.

#### Types of changes

#### Checklist:


matchai:
By updating tests to use the binary directly, I inadvertently made coverage instrumentation unable to catch our test runs. 🤦‍♂

tests/testsuite/directory.rs
```diff
@@ -0,0 +1,158 @@
...
+        Color::Cyan.bold().paint("starship/thrusters/rocket")
+    );
+    assert_eq!(expected, actual);
+    Ok(())
+}
+
+#[test]
+#[ignore]
+fn git_repo_root() -> io::Result<()> {
+    // TODO: Investigate why git repo related tests fail when the tempdir is within /tmp/...
```
Snuggle: I've noticed this with random software so often. If I put a file in `/tmp`, it just fails in some unexpected way. It can't be a permissions issue, I'm really not sure what causes this weirdness.

Sometimes I've stored images in /tmp and been unable to select them in applications. I'd love an explanation. :woman_shrugging:

PR #85, opened 2019-06-26
feat: Implement the prompt module for jobs

### LINKED ISSUE #80: Implement the prompt module for jobs (✦)
## Feature Request
Implement the prompt segment indicating that there are currently running jobs.
The `✦` symbol has been the default symbol for the module, with a blue module color. If there is more than 1 job running, a numbered indicator is shown with the number of jobs running.

It appears that fish has its own `jobs` binary, so testing will be needed on fish, zsh, and bash to ensure that the implementation is consistent across the shells.

The implementation used in spaceship is a good reference:
https://github.com/denysdovhan/spaceship-prompt/blob/5f0a747171cd171c578f9359da517545f60cc1da/sections/jobs.zsh

johnletey:
#### Description

I have implemented the prompt module that shows a ✦ when you have a job running

#### Motivation and Context

Closes #80

#### Types of changes

Shot 2019-08-12 at 17 45 39

#### Checklist:


matchai:
Hey @johnletey. How are you doing with this module? Do you need a hand getting this working?

johnletey:
> Hey @johnletey. How are you doing with this module? Do you need a hand getting this working?

Thanks so much @matchai ... yes, as a matter of fact, I would love some help getting this working!

matchai:
What exactly do you need help with? What does the PR currently do?

johnletey:
@matchai Currently the code doesn't show anything (even when I have it print when there is no job currently running)

matchai:
Running `jobs` with a spawned command might only show the jobs that are running in bash. I would recommend passing the ongoing jobs as a flag in `starship init`.

johnletey:
@matchai Do you mean `starship init` or `starship prompt`?

matchai:
I mean `starship init`. We would pass the output of a function in the user's shell, much like how `$status` is passed in as a flag because it isn't exported by its shell: https://github.com/starship/starship/blob/master/src/init.rs#L30

johnletey:
@matchai Did I do it right? I still need to display something for different values of the `--job` parameter, but did I at least instantiate and set everything up correctly?

matchai:
Yep! That's what I had in mind there.

johnletey:
@matchai Wonderful ... sorry, but ran into some problems running locally on my computer ... is the newly added code correct?

src/modules/jobs.rs
```diff
@@ -1,35 +1,21 @@
...
-            const JOB_CHAR: &str = "✦ ";
-            let module_color = Color::Blue.bold();
+    let mut module = context.new_module("jobs");
+    module.set_style(module_color);
 
-            let mut module = context.new_module("jobs");
-            module.set_style(module_color);
+    let arguments = &context.arguments;
+    if arguments.value_of("jobs").unwrap_or("0") == "0" {
+        module.new_segment("symbol", "");
```
matchai: At this point, I'd probably suggest returning `None`.

johnletey: Sounds good @matchai!

johnletey:
Updated @matchai!

src/modules/jobs.rs
```diff
@@ -0,0 +1,25 @@
...
+    let module_color = Color::Blue.bold();
+
+    let mut module = context.new_module("jobs");
+    module.set_style(module_color);
+
+    let arguments = &context.arguments;
+    let num_of_jobs = arguments.value_of("jobs").unwrap_or("0");
+    if num_of_jobs == "0" {
+        return None;
+    } else {
```
matchai: I would suggest removing this `else {`.
The code should behave the same but will involve less nesting.

johnletey: So something like this?

```rust
if num_of_jobs != "0" {
  module.new_segment("symbol", PROMPT_CHAR);
  if num_of_jobs != "1" {
    module.new_segment("number", num_of_jobs);
  }
};
```

src/main.rs
```diff
@@ -22,6 +22,13 @@ fn main() {
         .help("The status code of the previously run command")
         .takes_value(true);
 
+    let job_arg = Arg::with_name("job")
```
matchai: I would consider naming this flag `jobs`, as we're expecting a count of potentially multiple jobs.

johnletey: will do

src/main.rs
```diff
@@ -22,6 +22,13 @@ fn main() {
...
+        .short("j")
+        .long("job")
+        .value_name("JOB")
+        .help("The number of currently running jobs")
+        .takes_value(0);
```
matchai: Why the `.takes_value(0)`? 🤔 
afaik, `takes_value()` accepts a boolean value.

src/main.rs
johnletey: done @matchai!

src/main.rs
johnletey: You are right ... setting to `false` by default

src/modules/jobs.rs
```diff
@@ -0,0 +1,25 @@
+use ansi_term::Color;
+
+use super::{Context, Module};
+
+/// Creates a segment to show if there are any active jobs running
+pub fn segment<'a>(context: &'a Context) -> Option<Module<'a>> {
```
matchai: It looks like unit tests are failing because of this:
```suggestion
pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
```

I would take the time to review our continuous integration.

src/modules/jobs.rs
johnletey: nvm @matchai ... figured it out

matchai:
This feature is almost at the point of being mergable, but it seems CI is still failing. You can take a look at the results here: https://dev.azure.com/starship-control/starship/_build/results?buildId=1067&view=results

johnletey:
@matchai Tests are now passing!

matchai:
Excellent!
The last things that remain are documentation and acceptance tests:
- Acceptance tests would be added in a `jobs.rs` file in `tests/testsuite`
- A new heading would be made in `docs/config/README.md`
- `README.md` would be updated to no longer say that this is a planned feature

johnletey:
@matchai For the docs, should I update the code so that users can set a custom symbol for jobs?

matchai:
Yeah, I think that would be some great configuration to include!
Another possible configuration option is the threshold above which the job count is shown (`1` by default).

johnletey:
@matchai I've added tests + configurations to this module!

The tests don't seem to be passing because my module is always returning nothing. I have a suspicion that it has to do with the fact that I'm typecasting quite a lot, but would love your opinion on it!

Also, I have also updated the docs, but I don't like the wording, so would also love to have you review that as well!

johnletey:
@matchai All the tests are passing! Would love a review now 😄

src/modules/mod.rs
```diff
@@ -30,6 +31,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
         "username" => username::module(context),
         "battery" => battery::module(context),
         "cmd_duration" => cmd_duration::module(context),
+        "job" | "jobs" => jobs::module(context),
```
matchai: We're soon going to be making all modules only render with a single specific name, so let's to that in advance with this module.
```suggestion
        "jobs" => jobs::module(context),
```

src/print.rs
```diff
@@ -19,6 +19,7 @@ const PROMPT_ORDER: &[&str] = &[
     "python",
     "go",
     "cmd_duration",
+    "jobs",
```
matchai: Much like in spaceship, I think this would be better placed after `line_break`, so it would appear beside the character.

tests/testsuite/jobs.rs
```diff
@@ -0,0 +1,85 @@
+use ansi_term::Color;
```
matchai: Tests look great. 🎉

src/init.rs
```diff
@@ -108,6 +108,6 @@ function fish_prompt;
     set -l exit_code $status;
     set -l CMD_DURATION "$CMD_DURATION$cmd_duration";
     set -l starship_duration (math --scale=0 "$CMD_DURATION / 1000");
-    starship prompt --status=$exit_code --cmd-duration=$starship_duration;
+    starship prompt --status=$exit_code --cmd-duration=$starship_duration --jobs=$(count (jobs -p));
```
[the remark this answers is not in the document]
johnletey: Got it @matchai ... I have no clue how Fish works, so I just pushed your suggestion 😄

src/main.rs
```diff
@@ -43,6 +43,13 @@ fn main() {
         .help("The execution duration of the last command, in seconds")
         .takes_value(true);
 
+    let jobs_arg = Arg::with_name("jobs")
+        .short("j")
+        .long("jobs")
+        .value_name("JOB")
```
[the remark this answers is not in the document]
johnletey: Thanks!

src/modules/jobs.rs
```diff
@@ -0,0 +1,29 @@
...
+    let threshold = module.config_value_i64("threshold").unwrap_or(1);
+
+    const JOB_CHAR: &str = "✦ ";
+    let module_color = Color::Blue.bold();
+
+    module.set_style(module_color);
+
+    let arguments = &context.arguments;
+    let num_of_jobs = arguments.value_of("jobs").unwrap_or("0");
+    let num_of_jobs: i64 = num_of_jobs.parse().unwrap();
```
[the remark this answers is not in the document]
johnletey: Fixed

src/print.rs
johnletey: @matchai I've made this change, but in my opinion, it looks kind of weird:

Shot 2019-08-12 at 08 47 50

matchai: Here's what the implementation from spaceship looks like, which I think is a little closer to what I had in mind.

image

Perhaps we can strip away the spacing after the `✦` and the prefix of the module.

src/init.rs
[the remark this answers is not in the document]
matchai: No worries! I would suggest installing fish for testing these init functions in these situations. 👍

src/print.rs
johnletey: Understand @matchai ... I like that a lot better ... I have updated the code

PR #90, opened 2019-07-02
feat: Implement simplified prompt setup process

### LINKED ISSUE #47: Discussion: Installation process
As we get closer and closer to a primitive MVP of starship, we should start discussing our approach to installing the prompt. Here are a few possible implementations that I've been thinking about. I'd be open to hearing how you would suggest going about installing starship.

### What is needed

- `starship` is added to `$PATH`
- The following is executed as the prompt: `starship $status`

#### Summary
Installing `starship` via `apt-get` or `brew` runs a post-install script that appends the shell config file with a script that overwrites the prompt.

#### Details
- Add the following to your config file:

  ```bash
  # ~/.config/config.fish
  eval (starship init)

  # ~/.bashrc or ~/.zshrc
  eval "$(starship init)"
  ```

- `starship init` will output the following based on the shell being used:
  ```fish
  # fish
  function fish_prompt; starship $status; end

  # bash
  PS1="$(starship $?)"

  # zsh
  PROMPT="$(starship $?)"
  ```

#### Pros
- Easy for users to install manually
- Easy to automate
- Easy to uninstall
- Easy to troubleshoot

#### Cons
- Will overwrite the prompt set by a shell package manager or framework

#### Conclusion

| Metric            | Rating |
| ----------------- | ------ |
| Easy to implement | ★★★★★  |
| Easy to install   | ★★★★★  |
| Works as expected | ★★☆☆☆  |

---

#### Summary
Installing `starship` via `apt-get` or `brew` runs a post-install script that identifies the shell package managers or frameworks installed, and will install starship with the API of those respective package managers and frameworks.

#### Details
- Identify the prompt used by the user
- Identify which shell package manager or framework is being used (if any)
- Run necessary commands to install the prompt

```bash
## Fish is the primary prompt

# fisher is installed
fisher add starship/fish

# oh-my-fish is installed
omf install starship/fish
omf theme starship
```

```bash
## ZSH is the primary prompt

# oh-my-zsh is installed
git clone https://github.com/starship/zsh.git "$ZSH_CUSTOM/themes/starship"
ln -s "$ZSH_CUSTOM/themes/starship/starship.zsh-theme" "$ZSH_CUSTOM/themes/starship.zsh-theme"

# Antigen is installed
antigen theme starship/zsh

# Antibody is installed
antibody bundle starship/zsh

# zgen is installed
zgen load starship/zsh starship
```

#### Pros
- The most convenient solution for users
- Compatible with all supported package managers and frameworks

#### Cons
- Predicting which shell and package manager is being used by the user could be error-prone
- Will require implementation for each package manager
- Makes our installation process reliant on an API contract with each plugin manager and framework

#### Conclusion

| Metric            | Rating |
| ----------------- | ------ |
| Easy to implement | ★☆☆☆☆  |
| Easy to install   | ★★★★★  |
| Works as expected | ★★★★★  |

---

#### Summary
Installing `starship` via `apt-get` or `brew` runs a post-install script that
asks the user a series of questions to determine the correct way to install the prompt.

Similar to how `rustup` is installed.

<details><summary>Rustup Installation Screenshot</summary>

image

</details>

#### Details
- Ask the user which shell they use
- Ask the user which package manager or framework they use
- Install the correct prompt based on the user's input

#### Pros
- Won't be installed using the incorrect method
- Less magic makes it easier to troubleshoot
- Clear to the user
- Compatible with all supported package managers and frameworks

#### Cons
- Requires user intervention
- Won't be able to install silently by default
- Will require implementation for each package manager
- Makes our installation process reliant on an API contract with each plugin manager and framework

#### Conclusion

| Metric            | Rating |
| ----------------- | ------ |
| Easy to implement | ★★☆☆☆  |
| Easy to install   | ★★★★☆  |
| Works as expected | ★★★★★  |

---

I think that a mix of the last two may be the way to go. We attempt to identify the shell and package manager being used by the user, we then verify with the user (much like `rustup`), and install it accordingly. The user would be able to customize the installation and override the identified shell and package manager.

It would also support a `--silent` flag to simply go with the predicted method, or an installation method could be specifically chosen with flags (e.g. `--shell=zsh --install-with=antigen`).

matchai:
#### Description

- Add `starship init` which prints the shell function used to execute starship
- Document the new setup process using `starship init`
- Remove benchmarks for now (WIP replacement benchmarks in https://github.com/starship/starship/tree/benchmarking)

**[📝 Click here to view the updated README.md](https://github.com/starship/starship/blob/init-command/README.md)**

#### Motivation and Context

Closes #47 

#### Types of changes

image

image

#### Checklist:


src/init.rs
```diff
@@ -0,0 +1,40 @@
...
+            let script = "PROMPT=\"$(starship prompt --status=$?)\"";
+            Some(script)
+        }
+        "fish" => {
+            let script = "function fish_prompt; starship prompt --status=$status; end";
+            Some(script)
+        }
+        _ => {
+            println!(
+                "printf \"\\n{0} is not yet supported by starship.\\n\
```
johnletey: Is there supposed to be a `printf` inside the `println!`?

src/main.rs
```diff
@@ -28,13 +29,26 @@ fn main() {
...
+        )
+        .required(true);
+
+    let matches = App::new("starship")
         .about("The cross-shell prompt for astronauts. ✨🚀")
         // pull the version number from Cargo.toml
         .version(crate_version!())
         // pull the authors from Cargo.toml
         .author(crate_authors!())
         .after_help("https://github.com/matchai/starship")
```
johnletey:
```suggestion
        .after_help("https://github.com/starship/starship")
```

src/main.rs
```diff
@@ -28,13 +29,26 @@ fn main() {
...
-    let matches = App::new("Starship")
+    let shell_arg = Arg::with_name("shell")
+        .value_name("SHELL")
+        .help(
+            "The name of the currently running shell\nCurrently supported options: bash, zsh, fish",
+        )
+        .required(true);
+
+    let matches = App::new("starship")
         .about("The cross-shell prompt for astronauts. ✨🚀")
```
johnletey:
```suggestion
        .about("☄🌌️ The cross-shell prompt for astronauts.")
```

src/main.rs
```diff
@@ -28,13 +29,26 @@ fn main() {
         .help("The path that the prompt should render for")
         .takes_value(true);
 
-    let matches = App::new("Starship")
+    let shell_arg = Arg::with_name("shell")
+        .value_name("SHELL")
+        .help(
+            "The name of the currently running shell\nCurrently supported options: bash, zsh, fish",
```
johnletey:
```suggestion
            "The name of the current running shell\nCurrently supported options: bash, zsh, fish",
```

johnletey: Also, do you want to display the currently running shell? There is text there ... but nothing else

On second thought though ... Grammarly is telling me that "current running shell" is incorrect and it should be "currently running shell" ... so whichever one sounds better to you!

src/init.rs
matchai: Yes. What is within that `println()` is wrapped in an `eval()`, so the printf is needed to have it print the newlines as expected.

johnletey: Understand now @matchai!

PR #94, opened 2019-07-12
chore: Refactor getting string values from config

matchai:
#### Description

- A small refactor to what will soon be a common process; getting config that are of type string

#### Motivation and Context

Closes #

#### Types of changes

#### Checklist:


johnletey:
@matchai Just curious, how are you reading in a custom prompt character if you are converting the configuration file contents to a boolean?

matchai:
The config value is only retrieved as a boolean if `get_as_bool()` is used. This is meant for boolean configuration values, like for "disabled". 
In the case of setting the prompt character `get_as_str()` is used.

PR #106, opened 2019-07-20
fix: handle path to shell in starship init

### LINKED ISSUE #100: Starship doesn't support /usr/local/bin/zsh
#### Current Behavior
When running zsh under the name `/usr/local/bin/zsh` on MacOS, there is an error:

```
/usr/local/bin/zsh is not yet supported by starship.
For the time being, we support bash, zsh, and fish.
Please open an issue in the starship repo if you would like to see support for /usr/local/bin/zsh:
```
and I'm dumped back in to my default zprezto prompt.

#### Expected Behavior
Things should work fine: `/usr/local/bin/zsh` is decidedly a version of `zsh`.

Easy workaround for me: just make iTerm2 start the profile under the name `zsh` instead of `/usr/local/bin/zsh`. However, there may be certain edge cases where this isn't possible (e.g. you need a specific version of zsh for some reason or another and have to specify the absolute path).

Other potentially useful info:
- This isn't unique to iTerm2: running `/usr/local/bin/zsh` from MacOS's default Terminal will also trigger this error.
- This error does not seem to affect my Linux desktop (running same versions of zsh/starship).
- This does not affect fish, even when it's invoked as `/usr/local/bin/fish`, on any of the systems/terminal emulators I tried.

#### Environment
- Starship version: starship 0.2.0
- Shell type: zsh
- Shell version: zsh 5.7.1-dev-0
- Shell plugin manager: ZPrezto
- Terminal emulator: iTerm2
- Operating system: MacOS 10.14.5

```bash
# Source Prezto.
if [[ -s "${ZDOTDIR:-$HOME}/.zprezto/init.zsh" ]]; then
  source "${ZDOTDIR:-$HOME}/.zprezto/init.zsh"
fi

#Load theme
autoload -Uz promptinit
promptinit
prompt minimal

eval "$(starship init $0)"
```

#### Starship Configuration

I can't seem to find it :(

Even running a `find` over my entire home directory pulls up nothing.

#### Possible Solution

While I don't have a solid fix in mind, I'm making it a goal to learn a little Rust right now and I'd be happy to try to work on this (unless you think it would be easier to fix yourself). I tried implementing a quick fix using `std::path::Path`, but I couldn't quite get it working before some other work dragged me away (you have to use `OsString`/`OsStr` to use `Path`, and there's a whole lotta Options going on when trying to match.)

chipbuster:
This adds support for qualified paths (e.g. using
`/usr/local/bin/zsh` instead of `zsh`) to init.rs.

#### Description

`init` now converts the shell name into an OsStr, then to a Path, then gets the file stem, and unwraps back into a str. While this process can fail (yielding a None), it's highly unlikely to unless the user has messed with their shells or there's an issue in Starship--therefore, the failure message in this case simply asks the user to file a bug report.

I'm a little torn about the design, even though it's probably not terribly consequential either way (just me getting my newbie jitters out, I guess). The alternative would be to simply `unwrap()` the file stem and let it panic if it does--the conversion should almost never fail unless the user has done something bizarre to their shell or there's a bug in the Rust standard libraries.

#### Motivation and Context

Closes #100 

#### Types of changes

I've tested `starship init` on Linux and MacOS using names both with and without leading path components (as well as unsupported names), and everything seems to work as expected.

I can't actually get the `None` case to trigger, since I don't know how to cause a failure in `OsStr::to_str()`, but I've verified that evaluating the printed text in that case prints the intended message.

On MacOS and Linux, using bash/zsh/fish or their absolute paths:
```
"eval $(./starship init $0)"
```

To test for error cases:
```
"eval $(./starship init silly)"
```

#### Checklist:

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -4,9 +4,9 @@ use std::path::Path;
 pub fn init(shell_name: &str) {
     log::debug!("Shell name: {}", shell_name);
 
-    let shell_basename = Path::new(OsStr::new(shell_name))
+    let shell_basename = Path::new(shell_name)
         .file_stem()
-        .and_then(|path| path.to_str());
+        .and_then(OsStr::to_str);
 
     let setup_script = match shell_basename {
         Some("bash") => {
@@ -22,11 +22,10 @@ pub fn init(shell_name: &str) {
             Some(script)
         }
         None => {
-            println!(
-                "printf \"Starship cannot understand the name: {0}\\n\
-                 If you have not made weird changes to your shell, this\\n\
-                 probably indicates a bug in Starship. Please file a bug at\\n\
-                 https://github.com/starship/starship/issues/new\\n\"",
+            println!("Invalid shell name provided: {}\\n\
+                    If this issue persists, please open an \
+                    issue in the starship repo: \\n\
+                    https://github.com/starship/starship/issues/new\\n\"",
                 shell_name
             );
             None
--- a/src/init.rs
+++ b/src/init.rs
@@ -1,25 +1,45 @@
+use std::ffi::OsStr;
+use std::path::Path;
+
 pub fn init(shell_name: &str) {
     log::debug!("Shell name: {}", shell_name);
-    let setup_script = match shell_name {
-        "bash" => {
+
+    let shell_basename = Path::new(OsStr::new(shell_name))
+        .file_stem()
+        .and_then(|path| path.to_str());
+
+    let setup_script = match shell_basename {
+        Some("bash") => {
             let script = "PS1=\"$(starship prompt --status=$?)\"";
             Some(script)
         }
-        "zsh" => {
+        Some("zsh") => {
             let script = "PROMPT=\"$(starship prompt --status=$?)\"";
             Some(script)
         }
-        "fish" => {
+        Some("fish") => {
             let script = "function fish_prompt; starship prompt --status=$status; end";
             Some(script)
         }
+        None => {
+            println!(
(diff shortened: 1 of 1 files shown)


src/init.rs
```diff
@@ -1,25 +1,45 @@
...
             Some(script)
         }
-        "fish" => {
+        Some("fish") => {
             let script = "function fish_prompt; starship prompt --status=$status; end";
             Some(script)
         }
+        None => {
+            println!(
+                "printf \"Starship cannot understand the name: {0}\\n\
```
matchai: I would personally opt for a simpler error message here:
```
"Invalid shell name provided: {}\\n\
If this issue persists, please open an issue in the starship repo: \\n\
https://github.com/starship/starship/issues/new\\n\"
```

chipbuster:
Pushed a commit implementing the suggestions. 

Is there a preferred way to do it on GitHub? I'm used to working on a pretty old version of GitLab, so I'm not sure what the standard technique is for GH.

matchai:
>Is there a preferred way to do it on GitHub?

You did it correctly. 🙂 

Right now CI is failing because [the formatter failed](https://dev.azure.com/starship-control/starship/_build/results?buildId=990&view=logs&jobId=f14073de-e1a4-5ba6-5dc1-4c667fb92951), so the last thing left to do is for you to run [`rustfmt`](https://github.com/rust-lang/rustfmt) and commit the formatted changes. 😄

matchai:
Thank you for taking the time to write a PR! 😄 

@all-contributors please add @chipbuster for bug and code.

PR #118, opened 2019-07-27
feat: implement timer module

### LINKED ISSUE #104: Implement the prompt module for execution time
## Feature Request
Implement the prompt segment showing the execution time of the last command. By default, the module will only be shown if the execution time of the last command if it took more than 2 seconds to execute.

Much like the `jobs` module, each shell will likely have its own say to track and retrieve execution time, so information about the last command's execution may need to be passed in as a flag during `starship init`.

The implementation used in spaceship is a good reference for this module:
https://github.com/denysdovhan/spaceship-prompt/blob/5f0a747171cd171c578f9359da517545f60cc1da/sections/exec_time.zsh

chipbuster:
This PR adds a module in starship that adds a job timer.

#### Description

Implement a timer module that takes a commandline argument, the number of seconds the last job took to complete, and displays it if appropriate.

Alters shell initialization files to compute this number using `date +%s` where needed.

Adds a config section to configure minimum amount of time before timer is shown (default is 2s)

#### Motivation and Context

Closes #104

#### Types of changes

#### 
Screenshots (if appropriate):

Top-to-bottom: bash, zsh, fish (zsh is slightly doctored because my old shell config kept throwing some lines in there)

image

Tested by using in all three environments. Bash and Zsh proved surprisingly tricky to do correctly, so their shell configs are now fairly large.

Behavior is not 100% consistent between shells: when spamming empty commands, fish will continue to display the last command time, while bash and zsh will not.

Passes all integration tests.

#### Checklist:

(But only in-module tests--don't know if there are tests elsewhere)

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -373,21 +373,23 @@ The `cmd_duration` module shows how long the last command took to execute.
 The module will be shown only if the command took longer than two seconds, or
 the `min_time` config value, if it exists.
 
-NOTE: Command duration is currently **not supported in `bash`**. See
+::: warning
+Command duration is currently not supported in `bash`. See
 [this issue](https://github.com/starship/starship/issues/124) for more details.
+:::
 
 ### Options
 
 | Variable   | Default | Description                         |
 | ---------- | ------- | ----------------------------------- |
-| `disabled` | `false` | Disables the `cmd_duration` module. |
 | `min_time` | `2`     | Shortest duration to show time for. |
+| `disabled` | `false` | Disables the `cmd_duration` module. |
 
 ### Example
 
 ```toml
 # ~/.config/starship.toml
 
-[username]
+[cmd_duration]
 min_time = 4
 ```
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -373,7 +373,7 @@ The `cmd_duration` module shows how long the last command took to execute.
 The module will be shown only if the command took longer than two seconds, or
 the `min_time` config value, if it exists.
 
-::: warning
+::: warning NOTE
 Command duration is currently not supported in `bash`. See
 [this issue](https://github.com/starship/starship/issues/124) for more details.
 :::
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -109,6 +109,33 @@ command had an unsuccessful status code (non-zero).
 symbol = "❯"
 ```
 
+## Command Duration
+
+The `cmd_duration` module shows how long the last command took to execute.
+The module will be shown only if the command took longer than two seconds, or
+the `min_time` config value, if it exists.
+
+::: warning NOTE
+Command duration is currently not supported in `bash`. See
+[this issue](https://github.com/starship/starship/issues/124) for more details.
+:::
+
+### Options
+
+| Variable   | Default | Description                         |
+| ---------- | ------- | ----------------------------------- |
+| `min_time` | `2`     | Shortest duration to show time for. |
+| `disabled` | `false` | Disables the `cmd_duration` module. |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[cmd_duration]
+min_time = 4
+```
+
 ## Directory
 
 The `directory` module shows the path to your current directory, truncated to
@@ -367,29 +394,3 @@ The module will be shown if any of the following conditions are met:
 disabled = true
 ```
 
-## Command Duration
-
-The `cmd_duration` module shows how long the last command took to execute.
-The module will be shown only if the command took longer than two seconds, or
-the `min_time` config value, if it exists.
-
-::: warning NOTE
-Command duration is currently not supported in `bash`. See
-[this issue](https://github.com/starship/starship/issues/124) for more details.
-:::
-
-### Options
-
-| Variable   | Default | Description                         |
-| ---------- | ------- | ----------------------------------- |
-| `min_time` | `2`     | Shortest duration to show time for. |
-| `disabled` | `false` | Disables the `cmd_duration` module. |
-
-### Example
-
-```toml
-# ~/.config/starship.toml
-
-[cmd_duration]
-min_time = 4
-```
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -366,3 +366,28 @@ The module will be shown if any of the following conditions are met:
 [username]
 disabled = true
 ```
+
+## Command Duration
+
+The `cmd_duration` module shows how long the last command took to execute.
+The module will be shown only if the command took longer than two seconds, or
+the `min_time` config value, if it exists.
+
+NOTE: Command duration is currently **not supported in `bash`**. See
+[this issue](https://github.com/starship/starship/issues/124) for more details.
+
+### Options
+
+| Variable   | Default | Description                         |
+| ---------- | ------- | ----------------------------------- |
+| `disabled` | `false` | Disables the `cmd_duration` module. |
+| `min_time` | `2`     | Shortest duration to show time for. |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[username]
+min_time = 4
+```
--- a/src/init.rs
+++ b/src/init.rs
@@ -66,6 +66,10 @@ pub fn init(shell_name: &str) {
    we only start the timer if no timer has been started since the last prompt draw,
    tracked by the variable PREEXEC_READY. Similarly, only draw timing info if
    STARSHIP_START_TIME is defined, in case preexec was interrupted.
+
+   Finally, note that `eval` will evaluate this thing as a single line, so things
+   like comments (#) should be avoided and semicolons should be used on most 
+   statements.
 */
 
 const BASH_INIT: &str = r##"
@@ -83,17 +87,15 @@ starship_precmd() {
         unset STARSHIP_START_TIME;
     else
         PS1="$(starship prompt --status=$?)";
-    fi
+    fi;
     PREEXEC_READY=true;
 };
-# Check for preexec/precmd frameworks
 if [[ $preexec_functions ]]; then
     preexec_functions+=(starship_preexec);
     precmd_functions+=(starship_precmd);
     STARSHIP_START_TIME=$(date +%s);
-fi
-# Check if it is safe to trap DEBUG: get name of function trapping DEBUG, if any
-dbg_trap="$(trap -p DEBUG | cut -d' ' -f3 | tr -d \')"  
+fi;
+dbg_trap="$(trap -p DEBUG | cut -d' ' -f3 | tr -d \')";
 if [[ -z $dbg_trap || "$dbg_trap" = "starship_preexec" ]]; then
     trap starship_preexec DEBUG;
     PROMPT_COMMAND=starship_precmd;
--- a/src/init.rs
+++ b/src/init.rs
@@ -1,57 +1,26 @@
 use std::ffi::OsStr;
 use std::path::Path;
 
+/* We need to send execution time to the prompt for the timer module. For fish,
+this is fairly straightforward. For bash and zsh, we'll need to use several
+shell utilities to get the time, as well as render the prompt */
+
 pub fn init(shell_name: &str) {
     log::debug!("Shell name: {}", shell_name);
 
     let shell_basename = Path::new(shell_name).file_stem().and_then(OsStr::to_str);
 
     let setup_script = match shell_basename {
-        // The contents of `PROMPT_COMMAND` are executed as a regular Bash command
-        // just before Bash displays a prompt.
         Some("bash") => {
-            let script = r##"
-            starship_preexec() {
-                if [ "$PREEXEC_READY" = "true" ]; then
-                    PREEXEC_READY=false;
-                    STARSHIP_START_TIME=$(date +%s);
-                fi
-            };
-
-            starship_prompt() {
-                PREEXEC_READY=true;
-                STARSHIP_END_TIME=$(date +%s);
-                ELAPSED=$((STARSHIP_END_TIME - STARSHIP_START_TIME));
-                PS1="$(starship prompt --status=$? --elapsed=$ELAPSED)";
-            };
- 
-            trap starship_preexec DEBUG;
-            PROMPT_COMMAND=starship_prompt;
-            "##;
+            let script = BASH_INIT;
             Some(script)
         }
-        /* `precmd` executes a command before the zsh prompt is displayed.
-        We need to set STARSHIP_START_TIME as part of the init commmand or
-        else the first prompt will display a very large time */
         Some("zsh") => {
-            let script = r##"
-            precmd() {
-                STARSHIP_END_TIME="$(date +%s)"
-                ELAPSED=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
-                PROMPT="$(starship prompt --status=$? --elapsed=$ELAPSED)"
-            };
-            preexec(){
-                STARSHIP_START_TIME="$(date +%s)"
-            };
-            STARSHIP_START_TIME="$(date +%s)"
-            "##;
+            let script = ZSH_INIT;
             Some(script)
         }
         Some("fish") => {
-            let script = r##"function fish_prompt;
-            set -l elapsed (math --scale=0 "$CMD_DURATION / 1000");
-            starship prompt --status=$status --elapsed=$elapsed;
-            end;"##;
+            let script = FISH_INIT;
             Some(script)
         }
         None => {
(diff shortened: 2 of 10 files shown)


chipbuster:
A couple of notes about this: first, I've utterly goobered the git stuff--it's showing 25 commits to merge in, some of which I'm pretty sure were already merged. If anyone knows how to fix this, please let me know.

Second, I'm not sure if the design of the rendering function is sane--I think it's nice, but it also has a `(&u64, &&str)` as the argument, which seems a little ugly.

Finally, there are two places where the things can silently fail:
 - If a negative or very large number is passed as a minimum for displaying time, it'll simply not display the time.
- If a bash shell already has `DEBUG` trapped, re-trapping it will break whatever was depending on it (and there's no way to un-trap that function). Since we can't do timing without the DEBUG trap, the timer is just silently disabled. This has already bitten me once in testing (I forgot I had a primitive timer running already), but I'm not sure if we want to display a warning on every launch.

src/init.rs
```diff
@@ -59,3 +52,97 @@ pub fn init(shell_name: &str) {
...
+
+/* For zsh: preexec_functions and precmd_functions provide preexec/precmd in a
+   way that lets us avoid clobbering them.
+
+   Zsh quirk: preexec() is only fired if a command is actually run (unlike in
+   bash, where spamming empty commands still triggers DEBUG). This means a user
+   spamming ENTER at an empty command line will see increasing runtime (since
+   preexec never actually fires to reset the start time).
+
+   To fix this, only pass the time if STARSHIP_START_TIME is defined, and unset
```
matchai: Nice 👍

src/modules/timer.rs
```diff
@@ -0,0 +1,83 @@
+use crate::config::TableExt;
```
matchai: In #116, I unify `crate::config::TableExt` and `crate::config::Config`.
Maybe we should try merging that in first to avoid creating conflicts.

src/modules/timer.rs
```diff
@@ -0,0 +1,83 @@
...
+    let elapsed = arguments
+        .value_of("elapsed_time")
+        .unwrap_or("x") // not a time
+        .parse::<u64>()
+        .ok()?;
+
+    // Question: Is this a good series of calls to use? Is there a cleaner way to set types?
+    // Question: If the input config is negative, should I warn the user?
+    let config_min = context
+        .config
```
matchai: The module already contains its config, so you can access it via `module.config`.
https://github.com/starship/starship/blob/master/src/context.rs#L102

src/modules/timer.rs
```diff
@@ -0,0 +1,83 @@
...
+        .get_module_config("timer")
+        .and_then(|table| table.get_config("min_time"))
+        .and_then(|value| value.as_integer())
```
matchai: Ideally, I think we should have a `get_as_int` accessor with the other config accessors:
https://github.com/starship/starship/blob/af1b66708abc462fe0db34083c3f6c1d790f61a3/src/config.rs#L12-L14 (this is on the branch where `TableExt` is unified with `Config`)

src/modules/timer.rs
```diff
@@ -0,0 +1,83 @@
...
+        n => format!("{}{}", n, suffix),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_10s() {
```
matchai: 👍

src/modules/character.rs
```diff
@@ -1,4 +1,5 @@
 use super::{Context, Module};
+use crate::config::{Config, TableExt};
```
matchai: Is this addition needed? 🤔

chipbuster: Whoa, how did that get there?

I might have fumbled a key combo in my editor, but that is definitely not needed.

src/init.rs
```diff
@@ -59,3 +52,97 @@ pub fn init(shell_name: &str) {
...
+};
+precmd_functions+=(starship_precmd);
+preexec_functions+=(starship_preexec);
+STARSHIP_START_TIME="$(date +%s)";
+"##;
+
+/* Fish setup is simple because they give us CMD_DURATION. Hooray! */
+const FISH_INIT: &str = r##"
+function fish_prompt;
+    set -l elapsed (math --scale=0 "$CMD_DURATION / 1000");
```
matchai: I would suggest adding:
```fish
set -l command_duration "$CMD_DURATION$cmd_duration"
```

Between fish 2.7 and 3.0, the casing for the variable was changed.

src/init.rs
```diff
@@ -59,3 +52,97 @@ pub fn init(shell_name: &str) {
         print!("{}", script);
     }
 }
+
+/* For bash: we need to manually hook functions ourself: PROMPT_COMMAND will exec
+   right before the prompt is drawn, and any function trapped by DEBUG will exec
+   before a command is run. There is no way to avoid clobbering these functions,
+   so we will not hook them if they are already defined.
+
+   There is a preexec/precmd framework for bash out there: if we find the
```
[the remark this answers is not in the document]
chipbuster: That's correct. Actually, even if `bash-preexec` is installed on its own, it won't do anything because apparently you're supposed to source it as the very last thing in your shell config. Presumably if you were *using* the functionality, you'd have those arrays defined.

Honestly I don't know what the best way to do this is. I really want to avoid clobbering any pre-existing functionality, but there's no way to do that in native bash while also installing our timing functions.

matchai:
> I've utterly goobered the git stuff--it's showing 25 commits to merge in, some of which I'm pretty sure were already merged. If anyone knows how to fix this, please let me know.

This has happened to me more times than I can count.  😅
Don't worry about it. We'll be squash-merging your changes into a single commit.

src/modules/mod.rs
```diff
@@ -28,6 +29,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
         "git_status" => git_status::module(context),
         "username" => username::module(context),
         "battery" => battery::module(context),
+        "timer" => timer::module(context),
```
matchai: Thoughts on naming this module `command_duration`?
I don't think timer is very clear.

johnletey:
> it's showing 25 commits to merge in, some of which I'm pretty sure were already merged.

yes @chipbuster @matchai ... this happens when you're on a branch (not the master branch) that you have previously opened a PR for (that is now merged) ... and then you create a new branch off of this non-master branch

so your new branch has all of your old commits (which have now been merged into master), and all of your new commits

src/modules/timer.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    module.set_style(module_color);
+    module.new_segment("timer", &format!("took {}", render_time(elapsed)));
+    module.get_prefix().set_value("");
+
+    Some(module)
+}
+
+/// Render the time into a nice human-readable string
+fn render_time(raw_seconds: u64) -> String {
```
matchai: We already have [`uom`](https://crates.io/crates/uom) as a sub-dependency of `battery`.
Maybe we should use that for zero-cost conversions here?

chipbuster:
@johnletey I think I accidentally merged starship/master into my branch, instead of rebasing my branch onto it, and they diverged some time ago.

If we don't squash on merge, do you know how the process is handled? Will Github detect that some of the commits are duplicate and skip over them?

matchai:
If we don't squash we'll get repeat commits. 🤷‍♂️

src/modules/mod.rs
chipbuster: Agreed on `timer`. Would `cmd_duration` also work?

matchai: Yep, that should be fine.

src/modules/timer.rs
chipbuster: Hmm, I can't seem to access the config from within the module-generation function though, since it's private to Module.

```
field `config` of struct `module::Module` is private
```

Would adding an accessor to `Module` work? Or should I just make the config `pub`?

src/modules/timer.rs
chipbuster: Implemented. I've called it get_as_i64 since it wasn't clear to me initially what type of int would be returned.

src/modules/timer.rs
chipbuster: @matchai My impression after reading the `uom` doc is that it's great if we want to convert seconds into hours, or minutes into days, but if we have a number of seconds that we want to display as DD:HH:MM:SS, I don't see a way to do it aside from writing a custom `Formatter`, which would basically replicate what we're doing now, but in the library.

Is there something I missed while skimming the docs? I'm still not quite used to reading rust docs yet.

johnletey:
> If we don't squash on merge, do you know how the process is handled? Will Github detect that some of the commits are duplicate and skip over them?

Oh boy ... not quite sure ... that's as far as my Github expertise goes!

Zegnat:
> _[…]_ I've utterly goobered the git stuff--it's showing 25 commits to merge in, some of which I'm pretty sure were already merged. If anyone knows how to fix this, please let me know.

Assuming you have at least one branch that is up-to-date with starship/master, you can rebase against that one. I like to [setup an upstream remote](https://help.github.com/en/articles/configuring-a-remote-for-a-fork) for that, then you should be able to do something like:

```bash
git rebase -i upstream/master
```

With this git should already try and deduplicate commits, and you will be given the option to `drop` any commits you recognise as unnecessary.

src/modules/timer.rs
matchai: Looks like you're right. 👍

src/init.rs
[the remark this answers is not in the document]
matchai: Would it be worthwhile for us to include and source `bash-preexec` if it hasn't already been?

src/modules/cmd_duration.rs
```diff
@@ -0,0 +1,83 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("cmd_duration")?;
+    let arguments = &context.arguments;
+    let elapsed = arguments
+        .value_of("elapsed_time")
+        .unwrap_or("x") // not a time
+        .parse::<u64>()
+        .ok()?;
+
+    // Question: If the input config is negative, should I warn the user?
```
matchai: This can be a TODO for when error handling is implemented.
For the time being, I'd add a debug log for it and return `None`.

src/init.rs
```diff
@@ -59,3 +52,101 @@ pub fn init(shell_name: &str) {
...
+precmd_functions+=(starship_precmd);
+preexec_functions+=(starship_preexec);
+STARSHIP_START_TIME="$(date +%s)";
+"##;
+
+/* Fish setup is simple because they give us CMD_DURATION. Just account for name
+changes between 2.7/3.0 and do some math to convert ms->s and we can use it */
+const FISH_INIT: &str = r##"
+function fish_prompt;
+    let -l CMD_DURATION "$CMD_DURATION$cmd_duration"
```
matchai:
```suggestion
    set -l CMD_DURATION "$CMD_DURATION$cmd_duration"
```

src/main.rs
```diff
@@ -36,6 +36,13 @@ fn main() {
         )
         .required(true);
 
+    let cmd_duration_arg = Arg::with_name("elapsed_time")
```
matchai: I think we should stick to `cmd_duration` here too.
As I understand it, this is a measure of the duration of the last command's execution time, not the time elapsed since the start of the last command; or am I misunderstanding how it is implemented?

chipbuster: Yes, that's correct. I was just mentally doing a manual string replace and missed this part.

src/init.rs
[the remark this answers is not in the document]
chipbuster: Hmmm...now here's an interesting idea.

The big problem I see is that bash-preexec has to be [the absolute last thing sourced in the shell startup](https://github.com/rcaloras/bash-preexec/issues/97), or else it breaks. This would, in turn, require that we have `starship init` be the last line in the user's shell, and even then, `bash-preexec` will [*still* clobber any existing DEBUG traps](https://github.com/rcaloras/bash-preexec/issues/39).

Fundamentally, I don't think there's a way to avoid clobbering `DEBUG`/`PROMPT_COMMAND` in the bash startup because of the limitations of the shell. We can try to take the edge off of it, but there's simply no way to make it foolproof. In light of that, I think we have two options: we can make a best-effort attempt to compensate ourselves and (possibly noisily) disable modules if that fails, or we can try to provide a framework where the user can provide their own initializations.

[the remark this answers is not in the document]
chipbuster: I expect this to be  a pretty rare occurrence--the two Bash config frameworks I'm able to find are [Oh-My-Bash](https://github.com/ohmybash/oh-my-bash) and [Bash-it](https://github.com/Bash-it/bash-it). Oh-my-bash doesn't hook `DEBUG` at all, while `Bash-it` only hooks it specifically to provide a `preexec`-like function (sadly, it uses different function names for that functionality than `bash-preexec`).

What I'm taking from above is that it's going to be pretty rare for a user to hook the DEBUG trap without being aware that they've done it, so if we warn them about it, they should usually know how to fix it. We could enhance this by adding something in our documentation if needed.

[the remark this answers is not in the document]
matchai: At this point, I'm thinking we should probably remove Bash support for this feature and circle back to it once we have a robust solution figured out. The rest of the feature seems to work well. 🙂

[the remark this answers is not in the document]
chipbuster: Seems like a plan.

I'll put in some code to have starship log a message when calling init bash and then strip out the init code.

Since we're going to be squash-committing, should I reserve the code somewhere in a comment? Or are we okay with losing the bash init code for now?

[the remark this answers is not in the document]
matchai: Could you open an issue to add command duration support in Bash? We could keep the code for it there. 
I'm not at my computer right now, otherwise I'd take care of it.

src/modules/cmd_duration.rs
```diff
@@ -0,0 +1,93 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("cmd_duration")?;
+    let arguments = &context.arguments;
+    let elapsed = arguments
+        .value_of("cmd_duration")
+        .unwrap_or("invalid_time")
+        .parse::<u64>()
+        .ok()?;
+
+    let signed_config_min = context
```
matchai: The `config_value_X` convenience functions should work here:

```rust
let signed_config_min = module
	.config_value_i64("min_time")
	.unwrap_or(2);
```

src/main.rs
```diff
@@ -36,6 +36,13 @@ fn main() {
         )
         .required(true);
 
+    let cmd_duration_arg = Arg::with_name("cmd_duration")
+        .short("d")
+        .long("duration")
+        .value_name("DURATION")
+        .help("The amount of time since the start of the last command (s)")
```
matchai:
```suggestion
        .help("The execution duration the last command, in seconds")
```

src/main.rs
```diff
@@ -36,6 +36,13 @@ fn main() {
         )
         .required(true);
 
+    let cmd_duration_arg = Arg::with_name("cmd_duration")
+        .short("d")
+        .long("duration")
+        .value_name("DURATION")
```
matchai:
```suggestion
        .value_name("CMD_DURATION")
```

src/main.rs
```diff
@@ -36,6 +36,13 @@ fn main() {
         )
         .required(true);
 
+    let cmd_duration_arg = Arg::with_name("cmd_duration")
+        .short("d")
+        .long("duration")
```
matchai:
```suggestion
        .long("cmd-duration")
```

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "starship"
-version = "0.4.0"
+version = "0.3.2"
```
matchai: Whoops! Looks like the version in `Cargo.toml` and `Cargo.lock` got changed.

src/modules/cmd_duration.rs
chipbuster: Oh yeah, that's much better.

Was this added recently or did I miss it when I was writing this?

Cargo.toml
chipbuster: Ah, sorry. I had a bad feeling about that change while I was rebasing.

src/modules/cmd_duration.rs
matchai: Yeah, it was a recent addition in #120.

src/modules/cmd_duration.rs
```diff
@@ -0,0 +1,89 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("cmd_duration")?;
+    let arguments = &context.arguments;
+    let elapsed = arguments
+        .value_of("cmd_duration")
+        .unwrap_or("invalid_time")
+        .parse::<u64>()
+        .ok()?;
+
+    let signed_config_min = module.config_value_i64("min time").unwrap_or(2);
```
matchai: No spaces allowed in TOML keys 😛 

```suggestion
    let signed_config_min = module.config_value_i64("min_time").unwrap_or(2);
```

matchai:
I totally forgot that we also need to add some integration tests and documentation for this new module.
Would you like me to take care of it after this PR, or would you rather tackle it yourself?

The following would need doing:
- Integration tests would be added in a `cmd_duration.rs` file in `tests/testsuite`
- A new heading would be made in `docs/config/README.md`
- `README.md` would be updated to no longer say that this is a planned feature

chipbuster:
I'd be happy to do it, if you're willing to deal with a couple more reviews. I'm also starting to think I should hold off on working on this until this conference is over (Friday) and I have a chance to make up some sleep--I'm made some awfully silly mistakes the last few days.

What would need to be added for integration tests? Would it just be a bunch of tests using `render_module` with args for various `cmd-duration`s?

matchai:
Haha, no rush. I don't mind going through some more reviews. Enjoy your conference and we can take care of this later. 😄 

> What would need to be added for integration tests? Would it just be a bunch of tests using render_module with args for various cmd-durations?

Yep, that should be fine. This is what I was thinking:
- Testing the output at > and < the default cmd-duration criteria
- The output at > and < a cmd-duration criteria set by configuration

The recent `directory` tests are a good reference:
https://github.com/starship/starship/pull/120/files#diff-901574afaba53d9b8e5b1e81fe984069

docs/config/README.md
```diff
@@ -366,3 +366,28 @@ The module will be shown if any of the following conditions are met:
...
 disabled = true
 ```
+
+## Command Duration
+
+The `cmd_duration` module shows how long the last command took to execute.
+The module will be shown only if the command took longer than two seconds, or
+the `min_time` config value, if it exists.
+
+NOTE: Command duration is currently **not supported in `bash`**. See
```
matchai: As we're using Vuepress for our configuration, this may be a good place to put a "custom container: https://vuepress.vuejs.org/guide/markdown.html#custom-containers

I was thinking either a tip or a warning.

```
::: tip NOTE
Command duration is currently **not supported in `bash`**. See
[this issue](https://github.com/starship/starship/issues/124) for more details.
:::
```

docs/config/README.md
```diff
@@ -366,3 +366,28 @@ The module will be shown if any of the following conditions are met:
...
+| ---------- | ------- | ----------------------------------- |
+| `disabled` | `false` | Disables the `cmd_duration` module. |
+| `min_time` | `2`     | Shortest duration to show time for. |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[username]
```
matchai: Whoops!

```suggestion
[cmd_duration]
```

docs/config/README.md
```diff
@@ -366,3 +366,28 @@ The module will be shown if any of the following conditions are met:
...
+the `min_time` config value, if it exists.
+
+NOTE: Command duration is currently **not supported in `bash`**. See
+[this issue](https://github.com/starship/starship/issues/124) for more details.
+
+### Options
+
+| Variable   | Default | Description                         |
+| ---------- | ------- | ----------------------------------- |
+| `disabled` | `false` | Disables the `cmd_duration` module. |
```
matchai: Total nitpick: can we move `disabled` to the bottom of the table so it is consistent with the other modules?

src/modules/cmd_duration.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use super::{Context, Module};
+
+/// Outputs the time it took the last command to execute
+///
+/// Will only print if last command took more than a certain amount of time to
+/// execute. Default is two seconds, but can be set by config option `min_time`.
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("cmd_duration")?;
+    if module.config_value_bool("disabled").unwrap_or(false) {
```
matchai: This logic is already present in `context.new_module()` for all modules. 😜
https://github.com/starship/starship/blob/f40f73dc8e26e6d8730344f4c6cbb47ee3d48342/src/context.rs#L104-L109

tests/testsuite/cmd_duration.rs
```diff
@@ -0,0 +1,83 @@
...
+        .output()?;
+    let actual = String::from_utf8(output.stdout).unwrap();
+
+    let expected = format!("{} ", Color::Yellow.bold().paint("took 5s"));
+    assert_eq!(expected, actual);
+    Ok(())
+}
+
+#[test]
+#[ignore]
```
matchai: Because the config files that are generated are tempfiles created in a temp directory, there's no harm with running these tests on all machines. 🙂

```suggestion
```

tests/testsuite/cmd_duration.rs
```diff
@@ -0,0 +1,83 @@
...
+        .output()?;
+    let actual = String::from_utf8(output.stdout).unwrap();
+
+    let expected = "";
+    assert_eq!(expected, actual);
+    Ok(())
+}
+
+#[test]
+#[ignore]
```
matchai:
```suggestion
```

tests/testsuite/cmd_duration.rs
```diff
@@ -0,0 +1,83 @@
...
+        .output()?;
+    let actual = String::from_utf8(output.stdout).unwrap();
+
+    let expected = format!("{} ", Color::Yellow.bold().paint("took 10s"));
+    assert_eq!(expected, actual);
+    Ok(())
+}
+
+#[test]
+#[ignore]
```
matchai:
```suggestion
```

tests/testsuite/cmd_duration.rs
```diff
@@ -0,0 +1,83 @@
+use ansi_term::Color;
```
matchai: The tests look great! 👍

docs/config/README.md
chipbuster: Great idea. I had initially bolded that text only because I wanted it to be attention-grabbing and bolding the whole thing seemed obnoxious, but a warning works much better for that.

src/modules/cmd_duration.rs
chipbuster: d'oh

tests/testsuite/cmd_duration.rs
chipbuster: So the `#[ignore]` is usually used for tests that might (significantly) alter the test computer's state?

matchai: Yep. Ignored tests are run in Docker with `./integration_test`, but are not run with `cargo test`.

docs/config/README.md
```diff
@@ -366,3 +366,30 @@ The module will be shown if any of the following conditions are met:
 [username]
 disabled = true
 ```
+
+## Command Duration
```
matchai: Nitpick: The configuration option headings are listed in alphabetical order

src/init.rs
```diff
@@ -58,3 +51,62 @@ pub fn init(shell_name: &str) {
...
+STARSHIP_START_TIME="$(date +%s)";
+"##;
+
+/* Fish setup is simple because they give us CMD_DURATION. Just account for name
+changes between 2.7/3.0 and do some math to convert ms->s and we can use it */
+const FISH_INIT: &str = r##"
+function fish_prompt;
+    set -l CMD_DURATION "$CMD_DURATION$cmd_duration";
+    set -l starship_duration (math --scale=0 "$CMD_DURATION / 1000");
+    starship prompt --status=$status --cmd-duration=$starship_duration;
```
matchai: Unfortunately, it looks like a bug snuck in here. 🐛 
Because we are running the `set` commands before the prompt, we lose the old `$status` to the one set by `set`.

It seems we may need to set a variable at the start of the init script to hold onto the `$status` at the time:
```fish
set -l exit_code $status
```

chipbuster: Funnily enough, I fixed that for bash and zsh. I'm not sure whether I missed the fact that it was occurring for fish or just forgot to make the edits.

Is there a way we could do automated tests in an actual shell to catch these sort of regressions? I was kind of doing them manually, but even with the basic combinations of `true`/`false` and `sleep` and multiple shells, it was getting a little difficult to figure out if the behavior I was seeing was correct.

matchai: Agreed. We should introduce full end-to-end tests to catch these bugs. 👍

matchai:
@all-contributors please add @chipbuster for docs and test.

PR #120, opened 2019-07-28
feat: Allow directory truncation length to be configured

andrewda:
#### Description

This allows the directory truncation length to be configured. Previously, it was hard-coded to truncate to `3` parent directories.

#### Types of changes

Tested locally on Windows and using Docker.

#### Checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,6 +12,7 @@ pub trait Config {
     // Config accessor methods
     fn get_as_bool(&self, key: &str) -> Option<bool>;
     fn get_as_str(&self, key: &str) -> Option<&str>;
+    fn get_as_i64(&self, key: &str) -> Option<i64>;
 
     // Internal implementation for accessors
     fn get_config(&self, key: &str) -> Option<&toml::value::Value>;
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -129,6 +130,23 @@ impl Config for Table {
 
         str_value
     }
+
+    /// Get a key from a module's configuration as a string
```
matchai:
```suggestion
    /// Get a key from a module's configuration as an integer
```

tests/testsuite/directory.rs
```diff
@@ -95,6 +95,30 @@ fn truncated_directory_in_root() -> io::Result<()> {
     Ok(())
 }
 
+#[test]
+#[ignore]
+fn truncated_directory_config() -> io::Result<()> {
```
matchai: Looks good! 👍 
This covers the case where the directory isn't truncated because its length is so long.
Could we add one extra test here to ensure that truncation does happen correctly after it has been changed? (e.g. `directory.truncation_length = 2`)

src/module.rs
```diff
@@ -96,9 +96,19 @@ impl<'a> Module<'a> {
     }
 
     /// Get a module's config value as a string
-    fn config_value(&self, key: &str) -> Option<&str> {
+    pub fn config_value(&self, key: &str) -> Option<&str> {
```
matchai: I don't think we need this as `pub` now that we have the accessors beneath it.

src/config.rs
```diff
@@ -165,4 +183,20 @@ mod tests {
         table.insert(String::from("boolean"), toml::value::Value::Boolean(true));
         assert_eq!(table.get_as_str("boolean"), None);
     }
+
+    #[test]
+    fn table_get_as_i64() {
```
matchai: 👍

src/config.rs
```diff
@@ -129,6 +130,23 @@ impl Config for Table {
 
         str_value
     }
+
+    /// Get a key from a module's configuration as a string
+    fn get_as_i64(&self, key: &str) -> Option<i64> {
```
matchai: 👍

src/config.rs
andrewda: Whoops, good catch 👍

PR #131, opened 2019-08-10
feat: implement the ruby module

AZanellato:
#### Motivation and Context

There isn't a ruby prompt, which is one more language for the prompt.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/ruby.rs
+++ b/src/modules/ruby.rs
@@ -8,11 +8,10 @@ use super::{Context, Module};
 /// Will display the Ruby version if any of the following criteria are met:
 ///     - Current directory contains a `.rb` file
 ///     - Current directory contains a `Gemfile` file
-///     - Current directory contains a `.ruby-version` file
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let is_rb_project = context
         .new_scan_dir()
-        .set_files(&["Gemfile", ".ruby-version"])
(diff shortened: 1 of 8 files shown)


src/modules/ruby.rs
```diff
@@ -0,0 +1,60 @@
...
+use std::process::Command;
+
+use super::{Context, Module};
+
+/// Creates a module with the current Ruby version
+///
+/// Will display the Ruby version if any of the following criteria are met:
+///     - Current directory contains a `.rb` file
+///     - Current directory contains a `Gemfile` file
+///     - Current directory contains a `.ruby-version` file
```
matchai: I would drop support here for `.ruby-version` until we specifically support RVM.
```suggestion
```

src/modules/ruby.rs
```diff
@@ -0,0 +1,60 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let is_rb_project = context
+        .new_scan_dir()
+        .set_files(&["Gemfile", ".ruby-version"])
```
matchai:
```suggestion
        .set_files(&["Gemfile"])
```

src/modules/ruby.rs
AZanellato: Done! :smile:

matchai:
@all-contributors please add @AZanellato for code, docs and tests.

PR #133, opened 2019-08-10
feat: Add configuration to change the character for non-zero statuses

### LINKED ISSUE #130: Add alternative Character module configuration for the color-blind
## Feature Request
Add the ability to add a character to the prompt in the case of a non-zero exit code. Other prompts use things like 🚫, or 💥, or a red X. Having it configurable would be best, of course.

As someone who is red-green colorblind, it is very difficult to distinguish between the green and red arrows. Watching the intro animated GIF, I didn't even realize that the prompt changed after running `false` until I was reading through the documentation.

Add a configurable character to the prompt in the case of the previous command exiting non-zero.

Possibly *replacing* the character prompt with something else? So instead of ➜, maybe an X? I'm not sure where adding or replacing would be best, and would be very happy with either option.

chipbuster:
#### Description

Prompt can now switch characters in addition to switching character color. Add configuration options in so that users can do either, both, or neither.

#### Motivation and Context

Closes #130 

#### Types of changes

image

image

#### Checklist:


src/modules/character.rs
```diff
@@ -9,22 +9,39 @@ use ansi_term::Color;
 /// (green by default)
 /// - If the exit-code was anything else, the arrow will be formatted with
 /// `COLOR_FAILURE` (red by default)
+#[allow(clippy::collapsible_if)]
```
chipbuster: That shouldn't be there.

matchai:
I'm not sure I see the value in the `use_color_for_status` config. I had assumed that we wouldn't be stripping color from the prompt, rather just changing the character being used.

I had assumed we would simply be adding an `error_symbol` configuration value, which when set, would replace the `symbol` value in the case of a non-zero status. Now seeing it in action, I do think `use_symbol_for_status` makes sense to have so we have a good default for `error_symbol`.

chipbuster:
The reason I originally added the option (it's not 100% related to the character-switching issue) is that #110 is still open, and the only workaround I've found for it is to remove colors from the character. Even if that issue is patched by upstream one day, it's not hard to imagine us eventually expanding to support some shell where a similar issue could arise and it can't/won't be patched by the shell (or someone working on some ancient system where the patch hasn't hit the shell yet--the MacOS bash shell is from 2007!)

Thinking about it a little more closely now, it seems #128 is essentially a much finer-grained, more complete version of this, so I think I'll go ahead and drop the `use_color_for_status` in favor of the proposed changes there.

chipbuster:
I swear, someday I will figure out how to get autoformat-on-save to work correctly in VSCode...

PR #136, opened 2019-08-10
Update installation docs

matchai:
#### Description

Start listing installation methods as we gradually add starship to package repos.


matchai:
Yep, you're right.
Updated the PR to mention a version requirement.

PR #138, opened 2019-08-11
feat: Implement the prompt module for time

### LINKED ISSUE #135: Add current time to prompt
## Feature Request

It'd be great if the current time was displayed in the prompt line.

Something like: 
image

johnletey:
#### Description

As requested by @alichtman, I've taken time to work on implementing the prompt module for time the current time. It's not quite ready yet, as I'm working on fixing a bug ... but I'll let you know when I'm done and it's all working!

#### Motivation and Context

Closes #135

#### Types of changes

Shot 2019-08-16 at 11 33 07

> 24hr time

Shot 2019-08-16 at 11 33 21

> 12hr time

#### Checklist:


johnletey:
@matchai I've marked this a ready for review, as the code is working

Still have to write tests for the module, and it's going to be tricky to test given the fact that we'll have to set the local time to a specific value ... and I'm not entirely quite sure how to do that

Also, I'm still working on the docs

docs/config/README.md
```diff
@@ -409,6 +409,31 @@ The module will be shown if any of the following conditions are met:
...
+For the `format` configuration value, take a look [here](https://docs.rs/chrono/0.4.7/chrono/format/strftime/index.html) to see what you can pass in.
+:::
+
+### Options
+
+| Variable   | Default | Description                                  |
+| ---------- | ------- | -------------------------------------------- |
+| `12hr`     | `false` | Enables 12 hour formatting                   |
+| `format`   | `''`    | Enables the user to set a custom time format |
+| `disabled` | `false` | Disables the `username` module.              |
```
LinusU: Should this say "Disables the `time` module."?

```suggestion
| `disabled` | `false` | Disables the `time` module.              |
```

johnletey: It totally should @LinusU ... thanks for reviewing!

src/modules/time.rs
```diff
@@ -0,0 +1,30 @@
...
+        .unwrap_or("%H:%M")
+        .to_owned();
+
+    let is_12hr = module.config_value_bool("12hr").unwrap_or(false);
+    if is_12hr {
+        time_format = module
+            .config_value_str("format")
+            .unwrap_or("%I:%M %p")
+            .to_owned();
+    }
```
chipbuster: Personally, I would style the above block as 

```rust
let is_12hr = module.config_value_bool("12hr").unwrap_or(false);
let default_format = if is_12hr { "%I:%M:%p" } else { "%H:%M" };
let time_format = module. 
            .config_value_str("format")
            .unwrap_or(default_format)
            .to_owned();
```
(Or something like that, I haven't actually tested yet).

Not a huge deal, but it does avoid repeating the lookup into the module.

johnletey: Thank you so much @chipbuster!

src/modules/time.rs
```diff
@@ -0,0 +1,29 @@
+use ansi_term::Color;
+use chrono::{DateTime, Local};
+
+use super::{Context, Module};
+
+/// Outputs the current time
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
```
pickfire: Is it possible to elide the `'a` lifetime here?

src/modules/time.rs
```diff
@@ -0,0 +1,29 @@
...
+    let default_format = if is_12hr { "%I:%M%p" } else { "%H:%M" };
+    let time_format = module
+        .config_value_str("format")
+        .unwrap_or(default_format)
+        .to_owned();
+
+    let local: DateTime<Local> = Local::now();
+
+    module.set_style(Color::Yellow.bold());
+    module.new_segment("time", &format!("[{}]", local.format(&time_format)));
```
matchai: I don't know that we should have the brackets on either side of the module's output.
Perhaps we could have a `prefix` and `suffix` segment that are empty by default, but could be used to add these brackets?

```toml
[time]
prefix = "["
suffix = "]"

File: src/modules/time.rs [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("time")?;
+
+    // Remove when logic for disabled by default exists
+    if module.config_value_bool("disabled").unwrap_or(true) {
+        return None;
+    }
+
+    let is_12hr = module.config_value_bool("12hr").unwrap_or(false);
+    let default_format = if is_12hr { "%I:%M%p" } else { "%H:%M" };
```
matchai: Since users usually use the time in their prompt for second-by-second logging, I think having seconds by default would be of value. Thoughts?

matchai: It looks like there are premade format strings for exactly that:
```suggestion
    let default_format = if is_12hr { "%r" } else { "%T" };
```

docs/config/README.md
```diff
@@ -437,6 +437,31 @@ The module will be shown if any of the following conditions are met:
...
+::: tip
+For the `format` configuration value, take a look [here](https://docs.rs/chrono/0.4.7/chrono/format/strftime/index.html) to see what you can pass in.
+:::
+
+### Options
+
+| Variable   | Default | Description                                  |
+| ---------- | ------- | -------------------------------------------- |
+| `12hr`     | `false` | Enables 12 hour formatting                   |
+| `format`   | `''`    | Enables the user to set a custom time format |
```
matchai:
```suggestion
| `format`   | `''`    | Accepts a [chrono format string](https://docs.rs/chrono/0.4.7/chrono/format/strftime/index.html) to customize the time formatting |
```

docs/config/README.md
```diff
@@ -437,6 +437,31 @@ The module will be shown if any of the following conditions are met:
 symbol = "⚙️ "
 ```
 
+## Time
+
+The `time` module shows the current **local** time.
+
+::: tip
```
matchai: I would personally suggest putting the details about formatting in the description above. I think the tip box would be better suited to remind the user that this module is "disabled by default".

src/modules/time.rs
```diff
@@ -0,0 +1,29 @@
...
+    let time_format = module
+        .config_value_str("format")
+        .unwrap_or(default_format)
+        .to_owned();
+
+    let local: DateTime<Local> = Local::now();
+
+    module.set_style(Color::Yellow.bold());
+    module.new_segment("time", &format!("[{}]", local.format(&time_format)));
+    module.get_prefix().set_value("");
```
matchai: I think the module prefix should be `at ` by default.
If this module is at the end of the prompt order, that would make sense as a prefix.

That said, we should probably get around to implementing reordering of the prompt 😅

src/modules/time.rs
chipbuster: Seconded.

src/modules/time.rs
chipbuster: Don't suffixes only trigger if there's a module after the current one? If time is the last thing displayed on a line, the prompt will look like

```
v1.3.6 at [10:35
```
which I think looks unbalanced.

matchai: Only prefixes behave that way, not suffixes.
That said, these would be segments titled "prefix" and "suffix". They would not be instances of the `ModuleAffix` class, as those are usually colorless.

Sorry if that's confusing. 😅

chipbuster: Oh, I see. This would allow the user to configure the symbols they want used to wrap the time using `starship.toml`

Makes sense!

tests/testsuite/time.rs
```diff
@@ -0,0 +1,32 @@
...
+fn config_enabled() -> io::Result<()> {
+    let output = common::render_module("time")
+        .use_config(toml::toml! {
+            [time]
+            disabled = false
+        })
+        .output()?;
+    let actual = String::from_utf8(output.stdout).unwrap();
+
+    let expected = "";
```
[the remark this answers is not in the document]
johnletey: Yes @matchai ... I'm working on it!

src/modules/time.rs
johnletey: So ... I have made this change (https://github.com/starship/starship/pull/138/commits/f966b3dd1d28c0d9c0be65e6afa78b5ceb979b9b) ... but I just have one question, which is, should I still have this line?

```rust
module.get_prefix().set_value("");
```

or should I change it to

```rust
module.get_prefix().set_value("at");
```

johnletey: Nevermind ... just saw @matchai's review further on addressing my question 😄

src/modules/time.rs
johnletey: Done! (https://github.com/starship/starship/pull/138/commits/6d9f8eaface17cf71901a74ab0679cb719ff98f7)

src/modules/time.rs
johnletey: Done! (https://github.com/starship/starship/pull/138/commits/6d9f8eaface17cf71901a74ab0679cb719ff98f7)

johnletey: Done! (https://github.com/starship/starship/pull/138/commits/6d9f8eaface17cf71901a74ab0679cb719ff98f7)

src/print.rs
```diff
@@ -20,6 +20,7 @@ const PROMPT_ORDER: &[&str] = &[
     "cmd_duration",
     "line_break",
     "jobs",
+    "time",
```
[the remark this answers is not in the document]
johnletey: Done! (https://github.com/starship/starship/pull/138/commits/2543fe5a40af476765459a9e20549b508958ac77)

docs/config/README.md
johnletey: Done! (https://github.com/starship/starship/pull/138/commits/7727b90ddad41784aa6ea39421bc28a578bd3bdc)

matchai:
Hey @johnletey! 
I would love to get this feature merged in as soon as possible.
Do you need a hand to get this one finished up? 🙂

johnletey:
@matchai I would love some help finishing this up! So sorry for my inactivity with the project!

chipbuster:
I've rebased the changes over onto the current master branch [in my fork](https://github.com/chipbuster/starship/tree/add-time). @johnletey If you don't have any major work sitting unpushed, I can go ahead and push to this branch as well so that we're using the latest GH Actions framework?

chipbuster:
A quick thought on tests for this module:

It might be the case that this is a module for which it makes sense to have unit tests but no integration tests, because of the problems associated with freezing the time on the test system.

Even if we could do that, all it does is tell us that `Local::now()` is working, which, really, you would hope is the case.

The other option I see is to create an trait for dependency injection on `Local::now()`, but that's essentially doing the same thing: presenting the system with an appropriate (fake) time, seeing that it works with that fake time, and then hoping that `Local::now()` works when in production.

johnletey:
@chipbuster I don't have any urgent local changes at the moment, so if you want to start working off of this branch, go for it!

chipbuster:
@matchai @johnletey A quick summary of the major changes since we last looked at this file:

- I implemented the styling change for the time--you can now specify custom styles, falling back to a default `bold yellow` if none is found.
- I removed the `prefix` and `suffix` variables since they can be implemented with the `format` string--see the config page for an example.
- Formatting of the time has been separated out into a function which does not rely on the module or the current system time. While the current implementation is a bit...barebones, we can add functionality to it later if we want.
- Unit tests have been added using various times and format strings.
- Integration tests have been added that check the output of the `time` module *except* for the actual time, i.e. that the color codes are correct, the prefix is what we expect, and any decorations (e.g. extra brackets) on the format string are obeyed.

While I'm not 100% happy with this level of test coverage, I think it's definitely good enough to go into production (assuming that the testsuite passes, of course...)

chipbuster:
One thing to note is that the current location for the time can look a bit funky depending on the prefix and styling.

image

chipbuster:
Okay, I'm getting minorly trolled by clippy here. Anyone know what's causing this? It's showing errors on the CI even though it looks fine locally (and I'm using the same rustup toolchain version as the CI, 1.37 stable).

nickwb:
> Okay, I'm getting minorly trolled by clippy here. Anyone know what's causing this? It's showing errors on the CI even though it looks fine locally (and I'm using the same rustup toolchain version as the CI, 1.37 stable).

```
     assert!(actual.len() > 0);
                ^^^^^^^^^^^^^^^^ help: using `is_empty` is clearer and more explicit: `!actual.is_empty()`
```

johnletey:
Thanks for working on this @chipbuster! Really appreciate it!

I have gone ahead and fixed the clippy error (thanks @nickwb for pointing it out)

chipbuster:
@nickwb There was a more interesting clippy error happening earlier where it was complaining about a trait not being in scope and suggesting an import statement...but once I include the import statement, it fires a warning about unused imports.

Not a huge deal, just annoying that it was both doing that *and* that I couldn't get it to replicate on my local system.

docs/config/README.md
```diff
@@ -603,6 +604,38 @@ The module will be shown if any of the following conditions are met:
 symbol = "⚙️ "
 ```
 
+## Time
+
+The `time` module shows the current **local** time.
+The `format` configuration value is used by the `chrono` crate to control how the time is displayed. Take a look [at the chrono strftime docs](https://docs.rs/chrono/0.4.7/chrono/format/strftime/index.html) to see what options are available.
```
matchai: I'd suggest adding a link here:

```suggestion
The `format` configuration value is used by the [`chrono`](https://crates.io/crates/chrono) crate to control how the time is displayed. Take a look [at the chrono strftime docs](https://docs.rs/chrono/0.4.7/chrono/format/strftime/index.html) to see what options are available.
```

tests/testsuite/time.rs
```diff
@@ -0,0 +1,65 @@
...
+/* Note: tests in this crate cannot rely on the actual time displayed by
+the module, since that is dependent on the time inside the test environment,
+which we cannot control.
+
+However, we *can* test certain things here, such as the fact that the module
+should not display when disabled, should display *something* when enabled,
+and should have the correct prefixes and suffixes in a given config */
+
+#[test]
+#[ignore]
```
matchai: I think these `#[ignore]`s should be safe to remove.
These tests should have no side-effects.

PR #140, opened 2019-08-12
feat: Implement Python version from pyenv

MaT1g3R:
#### Description

Implement pyenv version for the Python module, it looks like this:
`foo on  master via 🐍 pyenv system`

This behavior can be enabled via setting `use_pyenv` to true.
The "pyenv" prefix before the version name can be configured using `pyenv_prefix`

#### Motivation and Context

pyenv provides named Python versions, users might want to display that instead of `python --version`

#### Types of changes

#### Checklist:

### DIFF
--- a/ci/setup-test-env.yml
+++ b/ci/setup-test-env.yml
@@ -10,27 +10,16 @@ steps:
       versionSpec: "1.10"
     displayName: "Install a fixed version of Go"
 
-  # Because the Pipelines agent updates are out of sync, we can't install 3.6.9
-  # with PythonTool on macOS
-
+  # We are using pyenv to install Python for integration tests
   # Install Python
-  - task: UsePythonVersion@0
-    inputs:
-      versionSpec: "3.6.9"
-    displayName: "Install a fixed version of Python"
-    condition: not(eq(variables['Agent.OS'], 'Darwin'))
-  # Install Python (macOS)
   - script: |
       echo "##vso[task.setvariable variable=PYTHON_VERSION;]3.6.9"
       echo "##vso[task.setvariable variable=PYENV_ROOT;]$HOME/.pyenv"
-    condition: eq(variables['Agent.OS'], 'Darwin')
   - script: |
       curl https://pyenv.run | bash
       echo "##vso[task.setvariable variable=PATH;]$PYENV_ROOT/bin:$PYENV_ROOT/shims:$PATH"
-    condition: eq(variables['Agent.OS'], 'Darwin')
   - script: |
       eval "$(pyenv init -)"
       pyenv install $PYTHON_VERSION
       pyenv global $PYTHON_VERSION
-    condition: eq(variables['Agent.OS'], 'Darwin')
     displayName: "Install a fixed version of Python"
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -328,7 +328,7 @@ symbol = "🎁 "
 
 The `python` module shows the currently installed version of Python.
 
-If `use_pyenv` is set to `true`, it will display the pyenv version name.
+If `pyenv_version_name` is set to `true`, it will display the pyenv version name.
 
 Otherwise, it will display the version number from `python --version`
 and show the current Python virtual environment if one is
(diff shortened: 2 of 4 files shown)


matchai:
Thank you for opening a PR @MaT1g3R!
I think the ability to check the version of pyenv is a valuable addition, but I wish we didn't need to toggle a configuration option to know the version of Python being used if pyenv is active. The average user wouldn't read the config page, so it would be great if we could support `pyenv` version detection without it.

Do you think that running `pyenv version` in addition to `python --version` would be too costly?

I was thinking for a while that we might want a `bins` value in `Context` that would contain a list of all binaries present in `$PATH`. This could be a way to avoid spawning a process unnecessarily when a binary (`pyenv` in this case) isn't installed.

MaT1g3R:
We could do something like `run pyenv, if no output fallback to python --version`, but I think that is a bit too magical. I think giving the user the choice to use pyenv instead of Python is good, because they might not be interested in the version name even if they had pyenv installed and want to see the version number instead.

matchai:
> they might not be interested in the version name even if they had pyenv installed and want to see the version number instead.
Oh, my bad! I forgot it was displaying the version name. I fell into a rabbit hole thinking it was displaying the pyenv version number instead of the Python version number. Yes, having that behind configuration makes sense. 👍

Do you know whether the prompt would currently respect the Python version set by pyenv if it was set locally? I know we wouldn't have issues with Python versions set globally.

MaT1g3R:
It works for both `pyenv shell foo` and `pyenv local foo`

matchai:
> It works for both pyenv shell foo and pyenv local foo

Perfect. 👍

docs/config/README.md
```diff
@@ -342,6 +347,9 @@ The module will be shown if any of the following conditions are met:
 | ---------- | ------- | -------------------------------------------------------- |
 | `symbol`   | `"🐍 "` | The symbol used before displaying the version of Python. |
 | `disabled` | `false` | Disables the `python` module.                            |
+| `use_pyenv` | `false` | Use pyenv to get Python version                            |
```
matchai: I think a more appropriate name might be `pyenv_version_name` since the default implementation does identify the python version used by `pyenv`. The difference here would be that it would show the version name extracted by pyenv.

Thoughts?

MaT1g3R: The default implementation doesn't use `pyenv` to get the Python version, but I agree `pyenv_version_name` is more clear

matchai: Right, the default implementation doesn't use `pyenv`. What I mean to say is that assuming you have installed everything correctly, the version of Python being shimmed by `pyenv` will output the version that is defined by `pyenv` globally or locally. 

We should be ready to merge as soon as the config value's name is updated. 🙂

src/modules/python.rs
```diff
@@ -26,19 +26,19 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let use_pyenv = context
+    let pyenv_version_name = context
```
matchai: You should be able to do the following once the module has been created to make things slightly simpler.

```rust
let pyenv_version_name = module
        .config_value_bool("pyenv_version_name")
        .unwrap_or(false);
```

MaT1g3R: I think my idea was not making the module until `select_python_version` returns `Some`, I don't know if we want to keep that behavior or not

matchai: I think it's best if we perform that check after the module is created. That way we can skip the check if the module is disabled (which is checked during module creation).

ci/setup-test-env.yml
```diff
@@ -10,27 +10,16 @@ steps:
       versionSpec: "1.10"
     displayName: "Install a fixed version of Go"
 
-  # Because the Pipelines agent updates are out of sync, we can't install 3.6.9
-  # with PythonTool on macOS
-
+  # We are using pyenv to install Python for integration tests
```
matchai: 👍

src/modules/python.rs
```diff
@@ -24,26 +26,52 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    match get_python_version() {
-        Some(python_version) => {
-            const PYTHON_CHAR: &str = "🐍 ";
-            let module_color = Color::Yellow.bold();
+    let mut module = context.new_module("python")?;
+    let pyenv_version_name = module
```
matchai: 👍

tests/testsuite/python.rs
```diff
@@ -93,3 +93,42 @@ fn with_virtual_env() -> io::Result<()> {
     assert_eq!(expected, actual);
     Ok(())
 }
+
+#[test]
```
matchai: 👍

PR #143, opened 2019-08-12
ci: Automate the bumping of the starship Brew formula

matchai:
#### Description

Automate the process of updating the Homebrew formula, using `brew bump-formula-pr`.

### DIFF
--- a/ci/install-cross-rust.yml
+++ b/ci/install-cross-rust.yml
@@ -37,7 +37,7 @@ steps:
       git checkout 718a19c
       git merge -m "No pseudo tty" pitkley/docker-no-pseudo-tty
       cargo install --force --path .
-    displayName: Instaling cross supprot
+    displayName: Install cross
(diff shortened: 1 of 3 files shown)


ci/install-cross-rust.yml
```diff
@@ -37,7 +37,7 @@ steps:
       git checkout 718a19c
       git merge -m "No pseudo tty" pitkley/docker-no-pseudo-tty
       cargo install --force --path .
-    displayName: Instaling cross supprot
+    displayName: Install cross
```
johnletey:
```suggestion
    displayName: Install cross-shell support
```

matchai: This installs the build tool called "cross": https://github.com/rust-embedded/cross

johnletey: 😆 That's hilarious ... then nvm

PR #147, opened 2019-08-13
fix: Fix issues with the init script for login shells

### LINKED ISSUE #141: Compatibility with login shell
The doc told me to add this line to my `~/.zshrc`

> eval "$(starship init $0)"

However, `$0` is `-zsh` when I use zsh as my login shell. The command above would throw me  an error

> error: Found argument '-z' which wasn't expected, or isn't valid in this context

Maybe using string substitution to remove the prefix '-' is better?

> eval "$(starship init ${0#'-'})"

matchai:
#### Description

The current installation instructions users of Bash and Zsh to use `$0` to expand into a path to their shell. This can occasionally be `-zsh` if Zsh is their login shell. Updated the installation instructions to be explicit about what shell is being used.

An alternative solution would be to strip symbols prefixing a path to a binary. What are your thoughts?

#### Motivation and Context

Closes #141 

#### Types of changes

#### Checklist:


chipbuster:
@matchai I thought about this a little more carefully last night and I think this is the only way forward, because clap seems to treat anything leading with a `-` as an argument flag. That means that we'd need to slightly break/modify clap in order to even get the chance to try to strip leading symbols off the path...and that seems a *lot* harder than just modifying the instructions.

PR #153, opened 2019-08-14
feat: Display Python package version from poetry

klane:
#### Description

Adds package version support for Python packages build with [`poetry`](https://poetry.eustace.io/). The version number is extracted from the `tool.poetry` section of the `pyproject.toml` file in the current directory, similar to the support for `cargo` packages. The prompt in `poetry` package directories now looks similar to:

`foo on  master is 📦 v0.1.0 via 🐍 v3.7.4`

#### Motivation and Context

Starship does not currently display Python package version numbers.

#### Types of changes

#### Checklist:


src/modules/package.rs
```diff
@@ -57,6 +65,11 @@ fn get_package_version() -> Option<String> {
         return extract_package_version(&package_json);
     }
 
+    let poetry_toml = utils::read_file("pyproject.toml");
+    if let Ok(poetry_toml) = poetry_toml {
+        return extract_poetry_version(&poetry_toml);
+    }
+
```
pickfire: Would it be better if
```rust
    if let Ok(cargo_toml) = utils::read_file("Cargo.toml") {
        extract_cargo_version(&cargo_toml)
    } else if let Ok(package_json) = utils::read_file("package.json") {
        extract_package_version(&package_json)
    } else if let Ok(poetry_toml) = utils::read_file("pyproject.toml") {
        extract_poetry_version(&poetry_toml)
    } else {
        None
    }
```

matchai: I personally prefer how it's read without the `else if`s, but it's functionally the same. It's up to @LinusU.  😄

matchai:
Looks great! Thank you for taking the time to add this feature @klane! 😄 

It looks like CI is failing on `rustfmt`, so I'd suggest running `cargo fmt` and committing those changes.

Before we merge it in, can we please have a unit test added for the new `extract_poetry_version` function like we have for the other two `extract_X_version` functions?

src/modules/package.rs
klane: I committed the suggested change. It is slightly shorter now, which will be beneficial if more package types are added.

klane:
> It looks like CI is failing on `rustfmt`, so I'd suggest running `cargo fmt` and committing those changes.

This is my first exposure to Rust and `cargo fmt` is unavailable on my system. It looks like I need to first install `rustup` and then install `rustfmt` with `rustup component add rustfmt`. Is that the correct process?

> Before we merge it in, can we please have a unit test added for the new `extract_poetry_version` function like we have for the other two `extract_X_version` functions?

Of course. I managed to completely miss the existing tests. I saw the empty package test file and assumed they had not yet been implemented.

matchai:
> This is my first exposure to Rust and cargo fmt is unavailable on my system. It looks like I need to first install rustup and then install rustfmt with rustup component add rustfmt. Is that the correct process?

Yes, that is correct. Sorry, I had assumed that `rustfmt` came included with Rust 😅

> Of course. I managed to completely miss the existing tests. I saw the empty package test file and assumed they had not yet been implemented.

Yeah, I was planning on writing integration tests for the `package` module, but forgot about it, leaving an empty file. Unit tests should do for now 👍

klane:
> Yes, that is correct. Sorry, I had assumed that `rustfmt` came included with Rust 😅

It does not appear to when you `brew install rust` and installing `rustfmt` with `cargo` gave me an error regarding a deprecated version. Using `brew install rustup` seems to be the way to go.

I borrowed from `test_extract_cargo_version` to make a `poetry` unit test and fixed a long line to satisfy `rustfmt`. Let me know if you want any other changes @matchai.

PR #154, opened 2019-08-14
style: improve fish init readibility

pickfire:
- remove semi-colons
- remove first blank line

#### Description

Use fish conventional style when generating the `init`, better readability if people did not `eval`.

#### Motivation and Context

Just improved readability that the fish code have the same style as most of the other fish code.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -162,11 +162,10 @@ STARSHIP_START_TIME="$(date +%s)";
 
 /* Fish setup is simple because they give us CMD_DURATION. Just account for name
 changes between 2.7/3.0 and do some math to convert ms->s and we can use it */
-const FISH_INIT: &str = r##"
-function fish_prompt;
-    set -l exit_code $status;
-    set -l CMD_DURATION "$CMD_DURATION$cmd_duration";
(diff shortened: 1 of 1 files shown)


matchai:
Hey @pickfire! Thank you for opening a PR. 😄 
As much as I agree that this would make for more idiomatic fish code, I believe fish's `eval` does not preserve newlines, so the semicolons are necessary.

Could you please try running this locally to confirm?

chipbuster:
This breaks on my machine :( I think it has the right idea though.

Is there any way to get eval to evaluate these things as different lines (maybe if we tried to treat them as files by using [heredocs](https://en.wikipedia.org/wiki/Here_document) and then using `source` instead of `eval`?)

The added complexity of having everything on one line makes it hard to figure out whether something is going to work (and when it breaks, why it broke)

matchai:
I'm going to close this PR as it is for now.
If we find an alternate solution that works, we can open a fresh PR for it. 🙂

pickfire:
Oh, I did not know that this breaks though. Sorry about that.

PR #166, opened 2019-08-17
feat: Allow bash users to specify a precmd function

### LINKED ISSUE #161: Can starship change terminal title?
This might be a dumb FAQ, but I couldn't figure it out. I always like to have my prompts change the title of my terminals as I change directories. In the old days I added bits to `PS1` that the internet told me did this, and it did. However, since trying out Starship (which I like) my title isn't changing anymore.

Is there something I can add to the toml file that will enable this?

chipbuster:
#### Description

Allows bash users to run a function before the shell is drawn, by defining a function and setting its name equal to `starship_precmd_user_func`, e.g.

```
function derp(){
# do something
}
starship_precmd_user_func=derp
eval $(starship init bash)
```

will cause `derp` to be run before drawing the prompt.

This can be combined with other bash features to do things like, e.g. setting the terminal title.

#### Motivation and Context

Closes #161 (though somewhat indirectly)

The feature requested in #161 can already be done for zsh by adding a function to `precmd_functions` and is done by default in fish. This adds support to be able to do it in `bash` as well.

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -17,7 +17,7 @@ into the prompt-rendering procedure:
 
 - To run a custom function right before the prompt is drawn, define a new
   function and then assign its name to `starship_precmd_user_func`. For example,
-  to draw a rocket face before the prompt, you would do
+  to draw a rocket before the prompt, you would do
 
 ```bash
 function blastoff(){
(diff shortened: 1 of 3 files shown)


matchai:
LGTM 👍 
I think we should document this somewhere for Bash users to know.

chipbuster:
Agreed, but I'm not sure what a good place for it is.

Do we have a "shell-specific quirks" spot in our documentation anywhere?

matchai:
We could certainly add a new page or section to the "Configuration" page.
I was thinking perhaps a "Frequently Asked Questions" section, or "Advanced Usage" / "Advanced Configuration".

chipbuster:
I think both are good to have, but this sounds more like an "Advanced Configuration" kind of thing.

Do I just make a new directory in the docs folder and add a file called "README.md" within it?

matchai:
> Do I just make a new directory in the docs folder and add a file called "README.md" within it?

Yep. You should also update https://github.com/starship/starship/blob/master/docs/.vuepress/config.js if you want it to appear in the sidebar.

docs/advanced-config/README.md
```diff
@@ -0,0 +1,74 @@
+# Advanced Configuration
+
+While Starship is a versatile shell, sometimes you need to do more than edit
+`starship.toml` to get it to do certain things. This page logs some of the more
```
matchai:
```suggestion
`starship.toml` to get it to do certain things. This page details some of the more
```

PR #168, opened 2019-08-17
feat: Implement a two-phase init which allows us to write normal init scripts

chipbuster:
#### Description

Implement a two-phase init procedure in starship. The first phase causes the shell to source a subshell, while the second phase (in the subshell) prints the main init script.

This allows us to have nice init scripts with good styling, comments, and no pile of semicolons. Even better, it works as a drop-in replacement, so we don't need to update the docs.

#### Motivation and Context
One of the problems with using `eval` directly is that our scripts all end up on one line. This leads to several problems when it comes to script development:

 - You cannot comment the script, because `#` characters continue to the end of the line, which is the rest of the script when it's all on one line.
 - Debugging is a nightmare--all you know is that there's an error *somewhere* within the script
 - You're forced to place semicolons in places which are not intuitive, even for experienced script writers (you have to have semicolons after every conditional statement, but if you put it after a `then`, it breaks badly and I don't know why)

This can be solved by using subshells to capture output from `source`. However, this leads to an additional problem: if we change our command to `source <(starship init $0)`, the users will have to change their shell rcfiles, which is going to be a bit of a hassle.

Instead, this update allows us to get the benefits of sourcing configs without any of the downside. It does this with a two-phase algorithm (the code is demoed with zsh syntax, but the same idea holds for all shells):

**Main invocation**

The output of `starship init zsh` is `source <(starship init zsh --print-full-init)`. This causes zsh to run a subshell which runs `starship init zsh --print-full-init`, then feed that output to `source` as a file.

**Inner invocation**

The output of `starship init zsh --print-full-init` is the zsh init script (`ZSH_INIT` in the code), complete with whitespace and newlines.

Since this is passed as a file to `source`, whitespace is retained and we can do things like comments.

#### Types of changes

Ideally, nothing should change. However, error reporting and recovery is a lot better.

In the following screenshot, I put the same malformed zsh statement into the init script. On the right is what the error looks like before this PR, on the left is after.

image

Note that the left side (the "after" image) tells us the error is on line 20, which we can use to debug. Meanwhile, the right side (the "before") just tells us that something has gone wrong at one of the many conditional tokens in the init script.

So one major, major caveat here: this is definitely pushing a little into dark shell sorcery, and I'm not 100% confident that this solution is portable across all versions of all shells. I already hit one weird edge case due to the fact that MacOS's version of `bash` is ancient, and there may be more in an older version of zsh, or maybe even in fish.

An alternative is to simply use `source <(starship init <shell>)`, which I am fairly confident will almost always work, but requires users to change their rcfiles. On one hand, we're pre-1.0, so a little breakage is to be expected, but on the other, I don't like springing surprises on people.

#### Checklist:

- [ ] I have updated the documentation accordingly. (not needed)

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -112,10 +112,10 @@ const BASH_INIT: &str = r##"
 starship_preexec() {
     # Avoid restarting the timer for commands in the same pipeline
     if [ "$PREEXEC_READY" = "true" ]; then
-        PREEXEC_READY=false;
-        STARSHIP_START_TIME=$(date +%s);
+        PREEXEC_READY=false
+        STARSHIP_START_TIME=$(date +%s)
     fi
-};
+}
 # Will be run before the prompt is drawn
 starship_precmd() {
     # Save the status, because commands in this pipeline will change $?
@@ -123,7 +123,9 @@ starship_precmd() {
     export STARSHIP_SHELL="bash"
 
     # Run the bash precmd function, if relevant
-    "${starship_precmd_user_func-:}";
+    "${starship_precmd_user_func-:}"
+
+    # Prepare the timer data, if needed.
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME=$(date +%s)
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
@@ -133,7 +135,7 @@ starship_precmd() {
         PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)")"
     fi
     PREEXEC_READY=true;  # Signal that we can safely restart the timer
-};
+}
 
 # If the user appears to be using https://github.com/rcaloras/bash-preexec,
 # then hook our functions into their framework.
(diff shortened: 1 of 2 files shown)


chipbuster:
A system like this would have allowed #154 to be merged without breaking the init.

chipbuster:
Currently waiting on #183 to merge, then I'll rebase and un-draft-mode this thing.

src/init.rs
```diff
@@ -1,26 +1,39 @@
...
+pub fn init_stub(shell_name: &str) {
     log::debug!("Shell name: {}", shell_name);
 
     let shell_basename = Path::new(shell_name).file_stem().and_then(OsStr::to_str);
 
-    let setup_script = match shell_basename {
+    let setup_stub = match shell_basename {
         Some("bash") => {
-            let script = BASH_INIT;
+            /* This *should* look like the zsh function, but bash 3.2 (MacOS default shell)
```
matchai: 👍

src/init.rs
```diff
@@ -1,26 +1,39 @@
...
             Some(script)
         }
         Some("zsh") => {
-            let script = ZSH_INIT;
+            let script = "source <(starship init zsh --print-full-init)";
             Some(script)
         }
         Some("fish") => {
-            let script = FISH_INIT;
+            // Fish does process substitution with pipes and psub instead of bash syntax
```
matchai: Huh... Neat! 👍

src/init.rs
```diff
@@ -46,142 +59,179 @@ pub fn init(shell_name: &str) {
...
-   Finally, to work around existing DEBUG traps in the absence of a preexec-like,
-   we parse out the name of the old DEBUG hook, then make a new function which
-   calls both that function and our starship hooks. We don't do this for
-   PROMPT_COMMAND because that would probably result in two prompts.
-
-   We need to quote the output of `$(jobs -p | wc -l)` since MacOS `wc` leaves
-   giant spaces in front of the number (e.g. "    3"), which messes up the
-   word-splitting. Instead, quote the whole thing, then let Rust do the whitespace
-   trimming within the jobs module
+/* GENERAL INIT SCRIPT NOTES
```
matchai: 👍

PR #169, opened 2019-08-17
feat: Display Vi mode as PROMPT

### LINKED ISSUE #129: Implement module for Vi mode
I'm currently using spaceship prompt in zsh, which has vi mode indication for normal and insert mode (`spaceship_vi_mode_enable`). Does starship offers vi mode indication already?

If not, I'd like to leave as feature request.

I've seen it being implemented in two ways, with dual prompts, like `❯❯❯` or `❯` for insert mode and `❮❮❮` or  `❮` for normal mode (could be configurable), or with an additional section, like, `[I] ➜` and `[N] ➜`. sorin and pure prompts do the former, spaceship does the latter.

cappyzawa:
#### Description

I enable to change prompt if [zsh: 18 Zsh Line Editor](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Keymaps) is `vicmd`.

KEYMAP is `vicmd`.
Screen Shot 2019-08-17 at 20 27 15

KEYMAP is other.
Screen Shot 2019-08-17 at 20 27 03

`vicmd_symbol` is configurable.

#### Motivation and Context

Closes #129 

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -100,7 +100,7 @@ can do this in two ways: by changing color (red/green) or by changing its shape
 | `symbol`                | `"➜"`   | The symbol used before the text input in the prompt. |
 | `error_symbol`          | `"✖"`   | The symbol used before text input if the previous command failed. |
 | `use_symbol_for_status` | `false` | Indicate error status by changing the symbol.         |
-| `vicmd_symbol`          | `"❮"`   | The symbol used before the text input in the prompt if Zsh Line Editor keymap is `vicmd`. |
+| `vicmd_symbol`          | `"❮"`   | The symbol used before the text input in the prompt if zsh is in vim normal mode. |
 | `disabled`              | `false` | Disables the `character` module.                      |
 
 ### Example
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -100,6 +100,7 @@ can do this in two ways: by changing color (red/green) or by changing its shape
 | `symbol`                | `"➜"`   | The symbol used before the text input in the prompt. |
 | `error_symbol`          | `"✖"`   | The symbol used before text input if the previous command failed. |
 | `use_symbol_for_status` | `false` | Indicate error status by changing the symbol.         |
+| `vicmd_symbol`          | `"❮"`   | The symbol used before the text input in the prompt if Zsh Line Editor keymap is `vicmd`. |
 | `disabled`              | `false` | Disables the `character` module.                      |
 
 ### Example
--- a/src/init.rs
+++ b/src/init.rs
@@ -158,6 +158,12 @@ if [[ ${preexec_functions[(ie)starship_preexec]} -gt ${#preexec_functions} ]]; t
     preexec_functions+=(starship_preexec);
 fi;
 STARSHIP_START_TIME="$(date +%s)";
+function zle-keymap-select
+{
+    PROMPT=$(starship prompt --keymap=$KEYMAP)
+    zle reset-prompt
+}
+zle -N zle-keymap-select
 "##;
 
 /* Fish setup is simple because they give us CMD_DURATION. Just account for name
--- a/src/init.rs
+++ b/src/init.rs
@@ -160,10 +160,10 @@ fi;
 STARSHIP_START_TIME="$(date +%s)";
 function zle-keymap-select
 {
-    PROMPT=$(starship prompt --keymap=$KEYMAP)
-    zle reset-prompt
-}
-zle -N zle-keymap-select
+    PROMPT=$(starship prompt --keymap=$KEYMAP --jobs="$(jobs | wc -l)");
+    zle reset-prompt;
+};
+zle -N zle-keymap-select;
 "##;
 
 /* Fish setup is simple because they give us CMD_DURATION. Just account for name
--- a/tests/testsuite/character.rs
+++ b/tests/testsuite/character.rs
@@ -72,3 +72,37 @@ fn char_module_symbolyes_status() -> io::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn char_module_vicmd_keymap() -> io::Result<()> {
+    let expected_vicmd = format!("{} ", Color::Green.bold().paint("❮"));
+    let expected_specified = format!("{} ", Color::Green.bold().paint("N"));
+    let expected_other = format!("{} ", Color::Green.bold().paint("➜"));
+
+    // zle keymap is vicmd
+    let output = common::render_module("character")
+        .arg("--keymap=vicmd")
+        .output()?;
+    let actual = String::from_utf8(output.stdout).unwrap();
+    assert_eq!(expected_vicmd, actual);
+
+    // specified vicmd character
+    let output = common::render_module("character")
+        .use_config(toml::toml! {
+            [character]
+            vicmd_symbol = "N"
+        })
+        .arg("--keymap=vicmd")
+        .output()?;
+    let actual = String::from_utf8(output.stdout).unwrap();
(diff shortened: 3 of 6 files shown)


chipbuster:
Hi @cappyzawa 

Thanks for the PR!

When I test it on my MacBook Pro, I get this strange artifact where line history gets nuked when switching between normal mode and insert mode (shown below). Do you know what might be causing or how it could be fixed?

vitest

docs/config/README.md
```diff
@@ -100,6 +100,7 @@ can do this in two ways: by changing color (red/green) or by changing its shape
 | `symbol`                | `"➜"`   | The symbol used before the text input in the prompt. |
 | `error_symbol`          | `"✖"`   | The symbol used before text input if the previous command failed. |
 | `use_symbol_for_status` | `false` | Indicate error status by changing the symbol.         |
+| `vicmd_symbol`          | `"❮"`   | The symbol used before the text input in the prompt if Zsh Line Editor keymap is `vicmd`. |
```
chipbuster: As discussed on #130, this might not work well with our colorblind support. It might be better to add a new module to indicate vi-status.

@matchai I guess we have to decide now. Do we want to overwrite the prompt symbol or use a new module? FWIW, on testing this, it looks like after running a command, you will initially have the success/failure symbol, and it only changes to the vi symbol if you actually go into normal mode.

chipbuster:
```suggestion
| `vicmd_symbol`          | `"❮"`   | The symbol used before the text input in the prompt if zsh is in vim normal mode. |
```

src/init.rs
```diff
@@ -158,6 +158,12 @@ if [[ ${preexec_functions[(ie)starship_preexec]} -gt ${#preexec_functions} ]]; t
     preexec_functions+=(starship_preexec);
 fi;
 STARSHIP_START_TIME="$(date +%s)";
+function zle-keymap-select
+{
+    PROMPT=$(starship prompt --keymap=$KEYMAP)
+    zle reset-prompt
+}
+zle -N zle-keymap-select
```
chipbuster: Unfortunately, this is going to crash `zsh` when you try to eval, because the *entire* script is evaluated on one line.

There is a PR in progress to fix that, but until that's merged, we'll need to spam semicolons everywhere for this to successfully `eval`.

```
function zle-keymap-select
{
    PROMPT=$(starship prompt --keymap=$KEYMAP);
    zle reset-prompt;
};
zle -N zle-keymap-select;
```

matchai:
Would you be interested in adding vi-mode functionality for fish shell?
The `$fish_key_bindings` variable within fish serves the same purpose to `$KEYMAP`.
If not, this could be addressed in a separate PR.

tests/testsuite/character.rs
```diff
@@ -72,3 +72,37 @@ fn char_module_symbolyes_status() -> io::Result<()> {
 
     Ok(())
 }
+
```
chipbuster: :+1:

src/init.rs
chipbuster: Just so that I understand this correctly: if I'm using emacs mode, or vi mode without ever switching into normal mode, this function will never fire right? It'll be just like using starship before this change?

Or is there some situation in which the prompt might redraw in spite of no mode change occuring?

docs/config/README.md
matchai: I'm in favor of replacing the symbol. Perhaps we could have a colorblind mode that will prefix the character symbol with a "✖" on a non-zero status, and leave a space for it otherwise: ` ` `➜` or `✖➜`.

cappyzawa:
@chipbuster Thanks for your review!!

I tried https://github.com/starship/starship/pull/169#issuecomment-522250017. 
fooo

This behavior isn't reproduced in my environment. 😓

@matchai Hi.
I have never used fish shell.
> this could be addressed in a separate PR.

Please:bow:

docs/config/README.md
chipbuster: I think I'd be happy with leaving this as it is for now then. You'll lose return status info if you switch into normal mode, but I don't think that information is so crucial that it has to be preserved in the scrollback history.

chipbuster: If we need to, we can add a colorblind mode later, though I'd prefer to review the entire shell to make sure that everything is colorblind-friendly before committing to it as an option.

chipbuster:
@cappyzawa Okay, I'll do some digging around on my computer and get back to you.

Could you provide a few details about your system (OS, ZSH version) so I can try to locate the source of the issue?

cappyzawa:
@chipbuster My environment is below.
```bash
$ zsh --version
zsh 5.7.1 (x86_64-apple-darwin18.2.0)

$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.14.6
BuildVersion:   18G87

$ rustc --version
rustc 1.37.0 (eae3437df 2019-08-13)
```

Thanks!

src/init.rs
```diff
@@ -158,6 +158,12 @@ if [[ ${preexec_functions[(ie)starship_preexec]} -gt ${#preexec_functions} ]]; t
     preexec_functions+=(starship_preexec);
 fi;
 STARSHIP_START_TIME="$(date +%s)";
+function zle-keymap-select
+{
+    PROMPT=$(starship prompt --keymap=$KEYMAP);
+    zle reset-prompt;
+}
```
chipbuster: Whoops, need a semicolon after the function close brace.
```suggestion
};
```

(The rules are dumb, this is why we're trying to get rid of one-line eval)

chipbuster:
@cappyzawa I found it, it was my fault :P

I had the dev version of `zsh` installed from back when there was a bug in 5.1.1 (ironically, also having to do with lines erasing themselves from the shell).

I'm going to go ahead and assume that's a regression in `zsh`, and not some feature they plan on introducing in 5.7.2.

src/init.rs
cappyzawa: oh. sorry.
I'll squash this to the previous commit.

src/init.rs
```diff
@@ -158,6 +158,12 @@ if [[ ${preexec_functions[(ie)starship_preexec]} -gt ${#preexec_functions} ]]; t
     preexec_functions+=(starship_preexec);
 fi;
 STARSHIP_START_TIME="$(date +%s)";
+function zle-keymap-select
+{
+    PROMPT=$(starship prompt --keymap=$KEYMAP);
```
chipbuster: One last comment! (You actually pushed your commit right before I hit 'ENTER').

```suggestion
    PROMPT=$(starship prompt --keymap=$KEYMAP --jobs="$(jobs | wc -l)");
```

While there's no good way to keep the return status info while using character changing, we probably should keep the jobs info up.

cappyzawa: OK!

cappyzawa: I'll squash this changes to df8d5090bb8f718c7e92a8 too.

cappyzawa:
@chipbuster Thanks for your review!

chipbuster:
@cappyzawa Thanks for taking the time to make this PR! 

(and sorry about the false alarm with line deletion)

Your build failed just now, but it was because of a python error, which I'm pretty sure isn't your fault, (probably a temporary failure in the build system). I'll poke the pipeline until the build succeeds, and then we'll merge your changes in.

chipbuster:
It looks like you might have uncovered a genuine race condition within our test suite with your changes!

I've gone ahead and removed those tests for now just so that we can get this merged--I'll throw them back in some other time.

chipbuster:
@all-contributors please add @cappyzawa for code, docs and tests.

PR #170, opened 2019-08-17
feat: Add Crystal version module

### LINKED ISSUE #160: Add Crystal on starship
It is heavily inspired by Ruby' syntax, so you can simple adapt the current implementation. 
Website: https://crystal-lang.org/

youssefhabri:
#### Description
Add Crystal language version module

#### Motivation and Context

Closes #160

#### Types of changes

#### Checklist:


src/modules/crystal.rs
```diff
@@ -0,0 +1,58 @@
...
+        .set_extensions(&["cr"])
+        .scan();
+
+    if !is_crystal_project {
+        return None;
+    }
+
+    match get_crystal_version() {
+        Some(crystal_version) => {
+            const CRYSTAL_CHAR: &str = "💎 ";
```
chipbuster: Hmmm....we're already using bold red with this symbol for [ruby](https://github.com/starship/starship/blob/master/src/modules/ruby.rs#L24). Is that going to get confusing?

chipbuster:
Aw heck, I didn't mean to actually do a full review there, just to make a one-line comment.

src/modules/crystal.rs
youssefhabri: yeah, I thought so too, but I have no idea what to use instead.

matchai: Maybe the "crystal ball"? 🔮

tests/testsuite/crystal.rs
```diff
@@ -0,0 +1,54 @@
+use ansi_term::Color;
```
chipbuster: Just a heads-up: these tests aren't actually running on `cargo test` or in the integration script because you didn't add it to `test/testsuite/main.rs`.

But I think if you do that, you'll also have to modify the docker container/test environment to include an install of crystal.

youssefhabri: I think I'll just leave it off for now. I don't wanna mess with the Dockerfile, I'll just leave that up to @matchai

vitallium:
@youssefhabri, hi, do you have any plans on updating this pull request? I would like to get this merged.

youssefhabri:
@vitallium I've updated the module but I kept the tests ignored as that would require adding crystal to the actions workflow

tests/testsuite/crystal.rs
```diff
@@ -0,0 +1,54 @@
...
+use std::fs::{self, File};
+use std::io;
+
+use crate::common;
+
+#[test]
```
chipbuster: While we're going to be replacing the test framework with the new mocking framework soon, I'm fine with leaving these in place for now so that we don't have to come up with them again when we do the move.

matchai:
@all-contributors Please add @youssefhabri for doc and test. 🎉

PR #171, opened 2019-08-17
feat: Add configuration for reordering the prompt module and disabling default order

### LINKED ISSUE #158: Add configuration for reordering the prompt modules
#### 🌱 This issue is open for first-time contributors to starship!
With a little bit of familiarity with Rust, you should be able to pick up this issue.
Please feel free to take your time with this one and ask any starship maintainers for help. 🙏 
We're here to guide you on your way to mastery of the codebase! 😄 

So far we have enabled configuration for many of the prompt's modules, but we haven't yet enabled users to configure the order in which the modules are shown in the prompt.

This would involve adding a configuration option for `prompt_order` that would accept an array of modules by their names. This would override the [default prompt order](https://github.com/starship/starship/blob/master/src/print.rs#L10-L25).

Similarly to how we have implemented the `add_newline` configuration below it, we would be accessing a config value to change the default behavior.

N/A

iamsauravsharma:
#### Description

This PR adds functionality for reordering the prompt module it provide out ~~two~~ one configuration setting i.e ```prompt_order``` which can be set as array of string ~~and ```disable_default_order``` as Boolean~~.

#### Example
1. configuration
```
prompt_order=["rust","git_status","package","character]
```  
Screenshot_20190817_233555

2. configuration
```
prompt_order=["rust","git_status","package","line_break","character"] 
```  
Screenshot_20190817_233641

3. configuration
```
prompt_order=["rust","line_break","package","line_break","character"]
```
Screenshot_20190817_233718

#### Motivation and Context

Closes #158

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -49,7 +49,7 @@ This is the list of prompt-wide configuration options.
 | Variable       | Default | Description                                                        |
 | -------------- | ------- | ------------------------------------------------------------------ |
 | `add_newline`  | `true`  | Add a new line before the start of the prompt.                     |
-| `prompt_order` | `[]`    | Configure out custom prompt order overriding default prompt order. |
+| `prompt_order` | `[]`    | Configure out custom prompt order overriding [default prompt order](#default-prompt-order). |
 
 ### Example
 
@@ -58,8 +58,31 @@ This is the list of prompt-wide configuration options.
 
 # Disable the newline at the start of the prompt
 add_newline = false
+# Overwrite a default_prompt_order and  use custom prompt_order
 prompt_order=["rust","line_break","package","line_break","character"]
 ```
+
+### Default prompt order
+```default_prompt_order``` is used as ```prompt_order``` when config have no options named as ```prompt_order``` or when you have set ```prompt_order = []```
+```toml
+default_prompt_order = [
+    "username",
+    "directory",
+    "git_branch",
+    "git_status",
+    "package",
+    "nodejs",
+    "rust",
+    "python",
+    "golang",
+    "cmd_duration",
+    "line_break",
+    "jobs",
+    "battery",
+    "character",
+]
+```
+
 
 ## Battery
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -46,11 +46,10 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Variable                | Default | Description                                                          |
-| ----------------------- | ------- | -------------------------------------------------------------------- |
-| `add_newline`           | `true`  | Add a new line before the start of the prompt.                       |
-| `prompt_order`          | `[]`    | Configure out custom prompt order modifying default order            |
-| `disable_default_order` | `false` | Disable default config order and only enable out custom prompt order |
+| Variable       | Default | Description                                                        |
+| -------------- | ------- | ------------------------------------------------------------------ |
+| `add_newline`  | `true`  | Add a new line before the start of the prompt.                     |
+| `prompt_order` | `[]`    | Configure out custom prompt order overriding default prompt order. |
 
 ### Example
 
@@ -59,8 +58,7 @@ This is the list of prompt-wide configuration options.
 
 # Disable the newline at the start of the prompt
 add_newline = false
-prompt_order=["rust","line_break","package","line_break","character"] 
-disable_default_order=true
+prompt_order=["rust","line_break","package","line_break","character"]
 ```
 
 ## Battery
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -46,9 +46,11 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Variable      | Default | Description                                    |
-| ------------- | ------- | ---------------------------------------------- |
-| `add_newline` | `true`  | Add a new line before the start of the prompt. |
+| Variable                | Default | Description                                                          |
+| ----------------------- | ------- | -------------------------------------------------------------------- |
+| `add_newline`           | `true`  | Add a new line before the start of the prompt.                       |
+| `prompt_order`          | `[]`    | Configure out custom prompt order modifying default order            |
+| `disable_default_order` | `false` | Disable default config order and only enable out custom prompt order |
 
 ### Example
 
@@ -57,6 +59,8 @@ This is the list of prompt-wide configuration options.
 
 # Disable the newline at the start of the prompt
 add_newline = false
+prompt_order=["rust","line_break","package","line_break","character"] 
+disable_default_order=true
 ```
 
 ## Battery
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -49,7 +49,7 @@ This is the list of prompt-wide configuration options.
 | Variable       | Default | Description                                                        |
 | -------------- | ------- | ------------------------------------------------------------------ |
 | `add_newline`  | `true`  | Add a new line before the start of the prompt.                     |
-| `prompt_order` | `[]`    | Configure out custom prompt order overriding [default prompt order](#default-prompt-order). |
+| `prompt_order` | [link](#default-prompt-order) | Configure the order in which the prompt module occurs. |
 
 ### Example
 
@@ -63,8 +63,8 @@ prompt_order=["rust","line_break","package","line_break","character"]
 ```
 
 ### Default prompt order
-```default_prompt_order``` is used as ```prompt_order``` when config have no options named as ```prompt_order``` or when you have set ```prompt_order = []```
-```toml
+The ```default_prompt_order``` configuration option is used to define the order in which modules are shown in the prompt, if empty or no ```prompt_order``` is provided. The default is as shown:
+```
 default_prompt_order = [
     "username",
     "directory",
--- a/src/config.rs
+++ b/src/config.rs
@@ -13,6 +13,7 @@ pub trait Config {
     fn get_as_bool(&self, key: &str) -> Option<bool>;
     fn get_as_str(&self, key: &str) -> Option<&str>;
     fn get_as_i64(&self, key: &str) -> Option<i64>;
+    fn get_as_array(&self, key: &str) -> Option<&Vec<toml::value::Value>>;
 
     // Internal implementation for accessors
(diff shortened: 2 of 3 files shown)


docs/config/README.md
```diff
@@ -46,9 +46,11 @@ This is the list of prompt-wide configuration options.
...
 ### Options
 
-| Variable      | Default | Description                                    |
-| ------------- | ------- | ---------------------------------------------- |
-| `add_newline` | `true`  | Add a new line before the start of the prompt. |
+| Variable                | Default | Description                                                          |
+| ----------------------- | ------- | -------------------------------------------------------------------- |
+| `add_newline`           | `true`  | Add a new line before the start of the prompt.                       |
+| `prompt_order`          | `[]`    | Configure out custom prompt order modifying default order            |
+| `disable_default_order` | `false` | Disable default config order and only enable out custom prompt order |
```
matchai: I don't think we really need the `disable_default_order` configuration option.
The default value of `prompt_order` could be retrieved from [PROMPT_ORDER](https://github.com/starship/starship/blob/master/src/print.rs#L10-L25), and overriding it in configuration would disable the default prompt order.

Please let me know if you have other functionality in mind around that configuration. 🙂

src/config.rs
```diff
@@ -141,6 +142,21 @@ impl Config for Table {
 
         i64_value
     }
+
+    /// Get a key from a module's configuration as a vector
```
matchai: Looks good. 👍

src/print.rs
```diff
@@ -36,7 +37,59 @@ pub fn prompt(args: ArgMatches) {
...
+        "golang",
+        "cmd_duration",
+        "line_break",
+        "jobs",
+        "battery",
+        "character",
+    ];
+
+    // Write out a custom prompt order
+    if let Some(values) = config.get_as_array("prompt_order") {
```
matchai: To try to keep things semantic, I think it would be best if we replaced references to `values` with `modules`.

src/print.rs
```diff
@@ -7,7 +7,8 @@ use crate::context::Context;
 use crate::module::Module;
 use crate::modules;
 
-const PROMPT_ORDER: &[&str] = &[
+//List of all prompt order
+const ALL_PROMPT_LIST: &[&str] = &[
```
matchai: I think this might be a little more descriptive.
```suggestion
const MODULE_LIST: &[&str] = &[
```

src/print.rs
```diff
@@ -36,7 +37,59 @@ pub fn prompt(args: ArgMatches) {
         writeln!(handle).unwrap();
     }
 
-    let modules = PROMPT_ORDER
+    let mut prompt_order: Vec<&str> = Vec::new();
+    //List of default prompt order
+    let default_prompt_order: &[&str] = &[
```
matchai: I think we should be able to make `default_prompt_order` a `const` rather than a `let`, to avoid the variable assignment.

src/print.rs
```diff
@@ -36,7 +37,59 @@ pub fn prompt(args: ArgMatches) {
...
+                } else {
+                    log::debug!(
+                        "Expected prompt_order to contain value from {:?}. Instead received {}",
+                        ALL_PROMPT_LIST,
+                        value,
+                    );
+                }
+            } else {
+                log::debug!(
+                    "Expected prompt_order to be a array of string. Instead received {} of type {}",
```
matchai: Small nitpick
```suggestion
                    "Expected prompt_order to be an array of strings. Instead received {} of type {}",
```

docs/config/README.md
iamsauravsharma: I have nothing special functionality in mind. But what about adding some custom module `languages` which will add all supported languages instead of adding all languages in config file. So, starship can easily support new languages module which are added in future by default on new version of starship without modifying custom config file. User can also add one by one languages by themselves.

docs/config/README.md
```diff
@@ -57,6 +58,28 @@ This is the list of prompt-wide configuration options.
 
 # Disable the newline at the start of the prompt
 add_newline = false
+prompt_order=["rust","line_break","package","line_break","character"]
+```
+
+### default_prompt_order
```
chipbuster: I'm not 100% sure about exposing this in the docs right now: we currently have two other PRs in the pipeline to add new modules and several additional issues about modules beyond that. If we throw this in now, we'll have to make sure that this and `print.rs` stay in sync.

I do think it's a good idea, but we might want to save it for 1.0.

iamsauravsharma: I have removed expose of default_prompt_order since I also believe it is not good idea to add this currently with other PR open. Can implement this in near future after addition of all language module

docs/config/README.md
iamsauravsharma: If `language` custom module is good custom module for addition then it may be added in another PR separately rather then implementing in this PR.

docs/config/README.md
iamsauravsharma: But after merging of this PR other PR open needs to add out added module in `ALL_MODULE` const in `print.rs` to support out custom prompt order for those module but need not to update config README

src/print.rs
```diff
@@ -36,7 +56,41 @@ pub fn prompt(args: ArgMatches) {
...
+                            ALL_MODULES,
+                            value,
+                        );
+                    }
+                } else {
+                    log::debug!(
+                    "Expected prompt_order to be an array of strings. Instead received {} of type {}",
+                    module,
+                    module.type_str()
+                );
```
andrewda: The above 4 lines should be indented another level

src/print.rs
```diff
@@ -7,7 +7,27 @@ use crate::context::Context;
...
+    "line_break",
+    "nodejs",
+    "package",
+    "python",
+    "ruby",
+    "rust",
+    "username",
+];
+
+//List of default prompt order
```
andrewda:
```suggestion
// List of default prompt order
```

src/print.rs
```diff
@@ -7,7 +7,27 @@ use crate::context::Context;
 use crate::module::Module;
 use crate::modules;
 
-const PROMPT_ORDER: &[&str] = &[
+//List of all modules
```
andrewda:
```suggestion
// List of all modules
```

docs/config/README.md
```diff
@@ -46,9 +46,10 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Variable      | Default | Description                                    |
-| ------------- | ------- | ---------------------------------------------- |
-| `add_newline` | `true`  | Add a new line before the start of the prompt. |
+| Variable       | Default | Description                                                        |
+| -------------- | ------- | ------------------------------------------------------------------ |
+| `add_newline`  | `true`  | Add a new line before the start of the prompt.                     |
+| `prompt_order` | `[]`    | Configure out custom prompt order overriding default prompt order. |
```
matchai: I think we may want to add the default `prompt_order` here, but it may be too big a list for the table.
Perhaps we could add a separate subb-heading to the "Prompt" heading specifically detailing the default prompt order and how changing it would work.

chipbuster: @matchai That was originally in the PR, but I voiced concerns about having that exposed separately (it means that we'd have to mirror changes in this doc as well every time we make a change, and we've been making a lot recently, with two or three PRs in the last 24 hours that would have needed to remember to make changes there).

Would it be possible to insert something here that generates the order directly from the code, so that we don't have to keep them in sync?

chipbuster: Perhaps a really simple solution would be to include a link to the GitHub source code (with the correct line number) for anyone who wants to know what the default order is.

docs/config/README.md
chipbuster: I think it would be a great idea, but getting the implementation right would be a challenge--you're essentially talking about implementing an extensions client for starship. We probably won't be ready to do that for some time.

docs/config/README.md
matchai: > Perhaps a really simple solution would be to include a link to the GitHub source code (with the correct line number)

I'm in favor of this 👍

iamsauravsharma: Main issue for linking out line number is changing line number since above `DEFAULT_PROMPT_ORDER` we have `ALL_MODULE` const and import statement which needs to be updated out for each addition of module since each modules need to be added over `ALL_MODULE` for custom prompt order to work as well as we may add new imports. Maybe adding subheading as orginally in the PR and commenting `DEFAULT_PROMPT_ORDER` code with information that if someone change this value they need to also update config docs

matchai: I think that's reasonable. 👍

docs/config/README.md
```diff
@@ -46,9 +46,10 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Variable      | Default | Description                                    |
-| ------------- | ------- | ---------------------------------------------- |
-| `add_newline` | `true`  | Add a new line before the start of the prompt. |
+| Variable       | Default | Description                                                        |
+| -------------- | ------- | ------------------------------------------------------------------ |
+| `add_newline`  | `true`  | Add a new line before the start of the prompt.                     |
+| `prompt_order` | `[]`    | Configure out custom prompt order overriding [default prompt order](#default-prompt-order). |
```
matchai: I think this may be a slightly clearer way of explaining this option.

```suggestion
| `prompt_order` | [link](#default-prompt-order)    | Configure the order in which the prompt modules appear. |
```

docs/config/README.md
```diff
@@ -57,8 +58,32 @@ This is the list of prompt-wide configuration options.
 
 # Disable the newline at the start of the prompt
 add_newline = false
+# Overwrite a default_prompt_order and  use custom prompt_order
+prompt_order=["rust","line_break","package","line_break","character"]
 ```
 
+### Default prompt order
+```default_prompt_order``` is used as ```prompt_order``` when config have no options named as ```prompt_order``` or when you have set ```prompt_order = []```
```
matchai: A little bit of nitpicking around how this option is documented.

```suggestion
The ```default_prompt_order``` configuration option is used to define the order in which modules are shown in the prompt. If no ```default_prompt_order``` is provided, the default is as shown:
```

chipbuster:
@all-contributors please add @iamsauravsharma for docs.

PR #173, opened 2019-08-17
feat: Add nix-shell support

### LINKED ISSUE #172: Add nix-shell support
It's useful to know when you're inside a nix-shell environment (like a python virtualenv).

Determine it with the `IN_NIX_SHELL` env variable.

None. `IN_NIX_SHELL` is simple enough I think.

bbigras:
#### Description
The module determine if inside a nix-shell env using `$IN_NIX_SHELL`.

#### Motivation and Context
It's useful to know when you're inside a nix-shell environment (like a python virtualenv).

Closes #172

#### Types of changes

image

#### Checklist:

I didn't do the doc yet.

Is it a bad idea to show the content of `$IN_NIX_SHELL` in the prompt (for security maybe)? Maybe it could be safer if I validate that it's either "1", "pure" or "impure".

Also for some reason, [lorri](https://github.com/target/lorri) (a nix-shell tool) sets `IN_NIX_SHELL` to "1" instead of "pure" or "impure". Is it ok to show "nix-shell" instead of "1"? I can check with the devs if their env is considered "pure" or "impure". Or maybe they could add their own env var.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ The prompt shows information you need while you're working, while staying sleek 
 - Current Ruby version (`💎`)
 - Current Python version (`🐍`)
 - Current Go version (`🐹`)
+- Nix-shell environment detection
 - Current version of package in current directory (`📦`)
   - npm (Node.js)
   - cargo (Rust)
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -490,3 +490,29 @@ The module will be shown if any of the following conditions are met:
 [username]
 disabled = true
 ```
+
+## Nix-shell
+
+The `nix_shell` module shows the nix-shell environment.
+The module will be shown when inside a nix-shell environment.
+
+### Options
+
+| Variable     | Default  | Description                        |
+| ------------ | -------- | ---------------------------------- |
+| `disabled`   | `false`  | Disables the `username` module.    |
+| `use_name`   | `false`  | Display the name of the nix-shell. |
+| `impure_msg` | `impure` | Customize the "impure" msg.        |
+| `pure_msg`   | `pure`   | Customize the "pure" msg.          |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[nix_shell]
+disabled = true
+use_name = true
+impure_msg = "impure shell"
+pure_msg = "pure shell"
+```
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -8,6 +8,7 @@ mod git_status;
 mod golang;
 mod jobs;
 mod line_break;
+mod nix_shell;
 mod nodejs;
 mod package;
 mod python;
@@ -35,6 +36,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
         "battery" => battery::module(context),
         "cmd_duration" => cmd_duration::module(context),
         "jobs" => jobs::module(context),
+        "nix_shell" => nix_shell::module(context),
 
         _ => {
             eprintln!("Error: Unknown module {}. Use starship module --list to list out all supported modules.", module);
--- a/src/modules/nix_shell.rs
+++ b/src/modules/nix_shell.rs
@@ -1,0 +1,54 @@
+use ansi_term::Color;
+use std::env;
+
+use super::{Context, Module};
+
+// IN_NIX_SHELL should be "pure" or "impure" but lorri uses "1" for "impure"
+// https://github.com/target/lorri/issues/140
+
+/// Creates a module showing if inside a nix-shell
+///
+/// The module will use the `$IN_NIX_SHELL` and `$name` environment variable to
+/// determine if it's inside a nix-shell and the name of it.
+///
+/// The following options are availables:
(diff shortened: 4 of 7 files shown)


bbigras:
About the lorri problem, someone on irc told me to open [a bug report](https://github.com/target/lorri/issues/140) and that it should be "impure" instead of "1".

The module could show "impure" if "1" or show nothing at all.

lovesegfault:
Isn't it possible to get the shell name as well? I recall that being feasible somehow

bbigras:
> Isn't it possible to get the shell name as well? I recall that being feasible somehow

`$name` seems to have it. Except that when using lorri it's defined as `lorri-keep-env-hack-shell-name`. I wonder if it would be a good idea to use `$name` since maybe it could be overwritten by other programs.

Maybe there's another way or maybe we could make nix-shell define a new env var like `$NIX_SHELL_NAME`.

lovesegfault:
I don't think changing the var name is feasible without hacks in the `shell.nix`. What we can do is re-export `$name` as `NIX_SHELL_NAME` when entering the shell, and then if anything overrides the original `$name` it's still fine

bbigras:
> I don't think changing the var name is feasible without hacks in the shell.nix.

Just to be sure, I meant that maybe we could change the nix-shell tool to define NIX_SHELL_NAME too... one day.

> What we can do is re-export $name as NIX_SHELL_NAME when entering the shell

How could we do that?

lovesegfault:
@bbigras https://doc.rust-lang.org/std/env/fn.set_var.html

That should be part of the work, that and https://doc.rust-lang.org/std/env/fn.var.html.

bbigras:
Oh thanks and sorry I'm familiar with Rust. I meant to ask if you thought starship should do it and how. Like if starship should check after every command if NIX_SHELL_NAME changed and re-export $name.

bbigras:
We could have an option to use $name optionally.

**name option enabled with nix-shell impure ($name is "shell")**
image

**name option enabled with nix-shell pure**
image

**name option disabled with nix-shell impure**
image

**name option disabled with nix-shell pure**
image

What do you guys think?

Other ideas:
- use (p) and (i) instead of (pure) and (impure)
- use a star like "shell*" only when pure.
- use the color to show if pure or not

chipbuster:
@bbigras I think "pure" and "impure" are good starting points: it's pretty clear what they're supposed to mean. We can implement the others as config options, e.g. via options called `pure_display_string` and `impure_display_string` (names need to be changed--I haven't looked at our config scheme for a while)

bbigras:
> We can implement the others as config options, e.g. via options called pure_display_string and impure_display_string (names need to be changed--I haven't looked at our config scheme for a while)

For the other case it uses the `name` env variable. Do config scheme supports using env variables in string options?

chipbuster:
No, but you could set an option like `use_env_shell_name` as boolean or something to enable or disable that. We have a few modules where we combine several string options with boolean options to control the behavior (right now, character.rs is the one that springs to mind, but there might be others)

bbigras:
> No, but you could set an option like use_env_shell_name as boolean or something to enable or disable that. We have a few modules where we combine several string options with boolean options to control the behavior (right now, character.rs is the one that springs to mind, but there might be others)

Oh yeah that was my plan. I misunderstood what you said in your last message.

I'll update the PR soon.

bbigras:
> Would this module be present most of the time, or is it for short durations?

I don't know how other devs use nix-shell but when I'm working on a project I'm in the same nix-shell, in one terminal window, for the whole day.

Also emacs uses the nix-shell in my case, with direnv.

So it's for long period of time.

matchai:
I'm a little hesitant to add such a long module if it's going to be present for full shell sessions. Most modules are context or location-sensitive. 

I'd be interested in hearing the opinion of some @starship/astronauts.

bbigras:
> I'm a little hesitant to add such a long module if it's going to be present for full shell sessions. Most modules are context or location-sensitive.

I thought I knew what you mean but now I'm not sure. Could you explain what do you mean by "long module" and maybe provide an exemple of a long module, one context sensitive and one location-sensitive?

matchai:
> I thought I knew what you mean but now I'm not sure. Could you explain what do you mean by "long module" and maybe provide an exemple of a long module, one context sensitive and one location-sensitive?

By "long", I just mean the character length by default. Something shorter would be the `git_status` module, which would be context-dependent. Location dependent would be something like the `rust` module, which is disabled once you leave a directory.

bbigras:
> By "long", I just mean the character length by default.

Oh I thought you meant how long (in time) do I stay inside a nix-shell during the day.

When you use nix-shell with direnv like I do, you enter a nix-shell when you enter some directory and you exit it when you leave it.

I guess some people might invoke nix-shell manually and stay in it for a while but I think a nix-shell is most useful for per-project environments.

So in my case, this module would be used like the rust one.

chipbuster:
@matchai I'm not a fan of prompt length either, but this has increasingly become an issue as we've added more modules. I'm getting pretty close to running off the edge of my standard terminal windows.

I'm not 100% sure if this one module represents big enough of an issue to be worried about it on its own, but I also appreciate that lots of small compounding changes can be an issue...

PR #174, opened 2019-08-17
Implement basic mocking capability for Command

matchai:
#### Description

So far, we have had to install all the applications for which we call the binaries of int our Dockerfile and test environment.

This would allow us to define static outputs that are produced by each binary's version command; making most tests run on all computers, and run _much_ faster.

Right now it works in theory, but it seems the starship binary that is used for integration tests isn't actually built with the test config.

We may either have to update the integration tests to call the module print function directly or find another way to build starship with mocks enabled.

#### Motivation and Context

Closes #

#### Types of changes

#### Checklist:


chipbuster:
Poking around the interwebz has revealed several interesting things, though I'm not sure how relevant any of them to our case (I'm not used to testing programs that require interfacing with the environment):

- Rust will [not compile all code units in test mode[https://freyskeyd.fr/cfg-test-and-cargo-test-a-missing-information] even when running tests. Specifically,

> When using a separate tests folder, your crate (lib) is considered as a seperate crate and compile like any other people will use it. It means that every cfg(test) will not be compiled

So that's a bummer. I don't see a workaround mentioned there, which I assume means that it doesn't exist.

Apparently most StackOverflow users who write Rust answers [favor dependency injection[(https://stackoverflow.com/questions/50888225/is-there-a-way-of-detecting-whether-code-is-being-called-from-tests-in-rust) as a testing technique (see also some of the links in the answer). Unfortunately, this would involve rewriting a lot of our tests, but it seems we might have to do that anyways (and it might also be useful if we ever decide to start reading environment variables instead of arguments).

PR #183, opened 2019-08-18
fix: Solve bash & zsh cursor location confusion issue

### LINKED ISSUE #110: Starship Prompt confuses bash/zsh cursor location
## Bug Report

Sorry about all the bug reports I've been generating recently! On the bright side, this is probably the last thing that's been bothering me in day-to-day use. Unfortunately, it seems a tad tricky to fix...

#### Current Behavior

In ZSH, using tab-completion with starship confuses the prompt, and causes some characters to be duplicated if there's more than one completion option. Once the extra characters are printed, the "zero point" of the shell seems to shift over, and you can no longer overwrite them. It also affect character wrapping on the right edge of the screen.

[Similar behavior exists in bash](https://imgur.com/2FnGjOJ), but it's a lot subtler (possibly because bash does not redraw certain areas). You have to mess with history to get it to show up.

#### Expected Behavior

Prompt should not interfere with TAB-completion or anything involving cursor jumps.

#### Environment
- Starship version: 0.2.0
- Shell type: zsh
- Shell version: zsh 5.7.1-dev-0
- Shell plugin manager: antibody
- Terminal emulator: iTerm2
- Operating system: MacOS 10.14.5

Replicates in a shell where the only line in .zshrc is `eval $(starship init zsh)`.

#### Starship Configuration

None.

#### Possible Solution

I don't have an idea for the solution but I think I know why this happens.

If you capture the output of `starship prompt` and dump it into a file, you get the following:

```
^[[1;36m/private/tmp^[[0m
^[[1;32m➜^[[0m
```

There are 11 non-printable characters on the second line of that prompt, and the space that's left between the start-of-prompt and start-of-unerasable-characters is exactly that many characters wide.

If you disable the color setting in the characters module, both the tab-completion issue and the character wrapping issue go away:

This leads me to conclude that the shell is counting the color escape codes as "printed characters" and placing the cursor too far to the right as a result.

If we were setting `PROMPT`/`PS1` via a hand-coded shell sequence, the solution would be to escape characters to indicate to the shell that these sequences are zero length:

- `\[ ... \]` for bash [[1]](https://unix.stackexchange.com/questions/28827/why-is-my-bash-prompt-getting-bugged-when-i-browse-the-history)[[2]](https://unix.stackexchange.com/questions/105958/terminal-prompt-not-wrapping-correctly)
- `%{ %}` [for zsh](https://stackoverflow.com/questions/11916064/zsh-tab-completion-duplicating-command-name)

but we're not--the color codes are coming from the ANSI library, and I don't see a way to tell the library to wrap the escape codes in a shell-specific escape sequence.

iamsauravsharma:
#### Description

This PR solves issue of cursor location confusion on bash and zsh as discussed over #110 . It solves this issue by modifying out ANSIString and wrap non-printing character escape for all of ansi-term additional parts.  

#### Motivation and Context

Closes #110 

#### Types of changes

#### Checklist:

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -96,10 +96,10 @@ starship_precmd() {
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME=$(date +%s);
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME));
-        PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)";
+        PS1="$(STARSHIP_SHELL="bash" starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)";
         unset STARSHIP_START_TIME;
     else
-        PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)")";
+        PS1="$(STARSHIP_SHELL="bash" starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)")";
     fi;
     PREEXEC_READY=true;
 };
@@ -143,10 +143,10 @@ starship_precmd() {
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME="$(date +%s)";
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME));
-        PROMPT="$(starship prompt --status=$STATUS --cmd-duration=$STARSHIP_DURATION --jobs="$(jobs | wc -l)")";
+        PROMPT="$(STARSHIP_SHELL="zsh" starship prompt --status=$STATUS --cmd-duration=$STARSHIP_DURATION --jobs="$(jobs | wc -l)")";
         unset STARSHIP_START_TIME;
     else
-        PROMPT="$(starship prompt --status=$STATUS --jobs="$(jobs | wc -l)")";
+        PROMPT="$(STARSHIP_SHELL="zsh" starship prompt --status=$STATUS --jobs="$(jobs | wc -l)")";
     fi
 };
 starship_preexec(){
--- a/src/init.rs
+++ b/src/init.rs
@@ -92,14 +92,15 @@ starship_preexec() {
 };
 starship_precmd() {
     STATUS=$?;
+    export STARSHIP_SHELL="bash";
     "${starship_precmd_user_func-:}";
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME=$(date +%s);
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME));
-        PS1="$(STARSHIP_SHELL="bash" starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)";
+        PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)";
         unset STARSHIP_START_TIME;
     else
-        PS1="$(STARSHIP_SHELL="bash" starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)")";
+        PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)")";
     fi;
     PREEXEC_READY=true;
 };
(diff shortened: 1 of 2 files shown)


src/module.rs
```diff
@@ -79,10 +79,42 @@ impl<'a> Module<'a> {
     /// Returns a vector of colored ANSIString elements to be later used with
     /// `ANSIStrings()` to optimize ANSI codes
     pub fn ansi_strings(&self) -> Vec<ANSIString> {
+        let shell = std::env::var("STARSHIP_SHELL").unwrap_or_default();
         let mut ansi_strings = self
             .segments
             .iter()
             .map(Segment::ansi_string)
+            .map(|ansi| {
+                let mut replaced_initial = false;
```
andrewda: A more descriptive name for this variable might be `escaped` or `is_escaped`.

andrewda:
Also, it looks like tests will need to be updated accordingly to account for this change https://dev.azure.com/starship-control/starship/_build/results?buildId=1508&view=logs&j=d9d4c2cb-ba39-5e18-d2a1-2a7ddd88c390&t=836754d4-a7dc-5a8b-a4c0-cebadad6c826&l=417

iamsauravsharma:
@andrewda I have changed out logic of how non printing escape conversion works. Now all test are passing.

src/init.rs
```diff
@@ -96,10 +96,10 @@ starship_precmd() {
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME=$(date +%s);
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME));
-        PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)";
+        PS1="$(STARSHIP_SHELL="bash" starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)";
```
chipbuster: I think it would be easier and cleaner to set this in the start of the script with an export line, e.g.

```
export STARSHIP_SHELL="bash";
```

at the end of the init script. Just remember to put the semicolon in, or the shell will throw up a huge fuss.

src/init.rs
```diff
@@ -143,10 +143,10 @@ starship_precmd() {
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME="$(date +%s)";
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME));
-        PROMPT="$(starship prompt --status=$STATUS --cmd-duration=$STARSHIP_DURATION --jobs="$(jobs | wc -l)")";
+        PROMPT="$(STARSHIP_SHELL="zsh" starship prompt --status=$STATUS --cmd-duration=$STARSHIP_DURATION --jobs="$(jobs | wc -l)")";
```
chipbuster: Can move this to an export statement like with bash as well.

src/module.rs
```diff
@@ -118,6 +125,43 @@ impl<'a> fmt::Display for Module<'a> {
     }
 }
```
chipbuster:
```suggestion

/// Many shells cannot deal with raw unprintable characters (like ANSI escape sequences) and 
/// miscompute the cursor position as a result, leading to strange visual bugs. Here, we wrap these
/// characters in shell-specific escape codes to indicate to the shell that they are zero-length.
```

You can update the comment with something that makes more sense to you, but let's leave something for future code spelunkers to know why we did this ^^

src/module.rs
```diff
@@ -118,6 +125,43 @@ impl<'a> fmt::Display for Module<'a> {
...
+fn ansi_strings_modified(ansi_strings: Vec<ANSIString>, shell: String) -> Vec<ANSIString> {
+    ansi_strings
+        .iter()
+        .map(|ansi| {
+            let mut escaped = false;
+            let final_string: String = ansi
+                .to_string()
+                .chars()
+                .map(|x| match x {
+                    '\u{1b}' => {
```
chipbuster: My preference would be to have compile-time consts in the start of the function to avoid having magic constants here, so:

```
const ESCAPE_BEGIN : char = '\u{1b}';
const MAYBE_ESCAPE_END : char = 'm';
```
at the top of the function and then matching on `ESCAPE_BEGIN` and `MAYBE_ESCAPE_END` in the code. (Again, pick names that make sense to you).

src/module.rs
```diff
@@ -118,6 +125,43 @@ impl<'a> fmt::Display for Module<'a> {
...
+                        escaped = true;
+                        match shell.as_str() {
+                            "bash" => String::from("\u{5c}\u{5b}\u{1b}"), // => \[ESC
```
chipbuster: My initial thought was to have the `ansi_strings` function pass the escape sequences into this function, but after some thought, I think your method is much more extensible 👍 .

chipbuster:
Finally, I can stop manually patching my version of starship to not use color. Thanks again!

@all-contributors please add @iamsauravsharma for code.

PR #194, opened 2019-08-19
fix: Restore python tests deleted in d90c43b

chipbuster:
#### Description

We deleted some python tests in d90c43b due to race condition. This PR restores them.

#### Motivation and Context

Closes #

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -111,4 +111,13 @@ mod tests {
         let input = "Python 3.7.2";
         assert_eq!(format_python_version(input), "v3.7.2");
     }
+
+    #[test]
+    #[ignore]
+    fn testvirtual_env() {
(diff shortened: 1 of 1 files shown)


matchai:
I worry that the environment variable being set here may be polluting the environment for other parallel tests. I think this test makes most sense to move into being an integration test, where tests don't share environment variables.

chipbuster:
Unfortunately, these tests need access to the `get_python_virtual_env` function to operate, and that's private within this module so we [can't access it from integration tests](https://stackoverflow.com/questions/27978377/how-do-i-test-private-methods-in-rust).

We already have tests using empty envars and `my_env` in the integration tests, so we could just drop these tests and rely on those?

matchai:
We already have tests using empty envars and my_env in the integration tests, so we could just drop these tests and rely on those?

I think that would make sense.

PR #200, opened 2019-08-19
chore(deps): Update `crossbeam-epoch` dependency

saghm:
Hello! I ran [`cargo audit`](https://rustsec.org/) on the crate, and it reported that the `memoffset` transitive dependency (`rayon` -> `crossbeam-deque` -> `crossbeam-epoch` -> `memoffset`) has an advisory published for the version that the `Cargo.lock` is pinned to. While we can't update it directly due to `crossbeam-epoch` requiring an older version, `crossbeam-deque` is happy to take an updated `crossbeam-epoch`, so I ran `cargo update -p crossbeam-epoch`, which in turn updated `memoffset` to the latest version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,21 +183,21 @@ name = "crossbeam-deque"
 version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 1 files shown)


saghm:
Glad to help! As always, thanks for the extremely fast turnaround time!

PR #203, opened 2019-08-19
WIP: Add kubernetes hints: current-context and current namespace

### LINKED ISSUE #201: Kubernetes hints
## Feature Request

Is there any plan to implement kubernetes hints in the prompt?  I thought about starting on a PR, but want to check here first.

jrop:
#### Description

This module adds the k8s current-context and namespace, (if a kubectl binary is present).

#### Motivation and Context

Closes #201 

#### Types of changes

#### Checklist:


chipbuster:
Hey @jrop, are you a GitLab user? We do the WIP thing at my current job too, but unfortunately the only way to mark a PR as draft on GitHub is to explicitly do so at PR creation.

jrop:
@chipbuster Yeah, I use GitLab for anything I have the chance to.  Makes sense.  I marked it as WIP because I have only done preliminary testing, and have been mulling over how to add some automated tests to the code as well.

src/modules/kubernetes.rs
```diff
@@ -0,0 +1,54 @@
+use ansi_term::Color;
+use std::process::Command;
+
+use super::{Context, Module};
+
+fn ex(cmd: &str, args: &[&str]) -> Result<String, String> {
```
chipbuster: This is a nice helper--so nice, in fact, that I don't think it should stay limited to this scope. This seems like a pretty common operation 

I won't ask you to refactor the entire codebase, but we should probably look into moving it out of this at some point in the future.

(roping in @matchai for his opinion)

src/modules/kubernetes.rs
```diff
@@ -0,0 +1,54 @@
...
+    Ok(output)
+}
+
+/// Creates a module with the current k8s context/namespace
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let maybe_ctx: Option<String> = ex("kubectl", &["config", "current-context"])
+        .map(Some)
+        .or::<String>(Ok(None))
+        .unwrap();
+    let maybe_namespace: Option<String> = ex(
```
chipbuster: What's the runtime cost of running `kubectl`, especially if your caches are empty?

Most of our modules are gated behind the existence of files, which is a relatively quick check, then a more expensive call to runtimes if the appropriate files are found. Without that, we'd be forking a process for each module every time we ran the shell, which isn't exactly *slow* individually, but adds up.

Is there the presence of some sort of file/envar or other quick check we could use as a fast check before running kubectl?

jrop: @chipbuster Possibly.  I can look into checking for `$KUBECONFIG` (iirc) and then parsing that file.  I think all of the data is in there somewhere.

bijancn:
I am afraid https://github.com/starship/starship/pull/404 is superseding this change as it is only using the file system rather than calling `kubectl`, so I would suggest to close this one. Is that okay for you @jrop ?

jrop:
@bijancn Thanks for the update, and yes that is fine with me.  I have not made the time to work on this anyway, to bring it up-to-date.  I am excited for the possibility of k8s integration soon.  Thanks for the heads up

PR #215, opened 2019-08-21
fix: Only print shell escapes when in a prompt context

### LINKED ISSUE #211: starship prompt printing all escape character
#### Current Behavior
Currently starship prompt is printing all escape character which are used for support of bash/zsh cursor position issue solved earlier which cause starship prompt while run for testing print all of escape character outlines.

This is caused by setting STARSHIP_SHELL environment variable in init script

Screenshot_20190820_221333

#### Expected Behavior
starship prompt should not print out non printing escape character while running it manually.

#### Environment
- Starship version: 0.12.0
- Shell type: [fish, zsh]
- Shell version: 5.0.7
- Shell plugin manager: none
- Terminal emulator: All
- Operating system: Manjaro

#### Possible Solution
Do not export STARSHIP_SHELL init variable instead provide variable value while setting PS1 and PROMPT for bash and zsh respectively
OR
Do not export STARSHIP_SHELL init variable instead add new flag for prompt subcommand to set SHELL
OR
Leave STARSHIP_PROMPT as it is and add new flag may be test which change STARSHIP_SHELL value to test and run starship prompt

chipbuster:
#### Description

Add an additional variable to starship shell to let it know when it's printing in a shell prompt context. Zero-length wrappers are only printed when in a shell prompt context.

#### Motivation and Context

Closes #211 

#### Types of changes

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -739,7 +739,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "starship"
-version = "0.12.0"
+version = "0.12.1"
 dependencies = [
  "ansi_term 0.12.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "battery 0.7.4 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/src/module.rs
+++ b/src/module.rs
@@ -105,9 +105,15 @@ impl<'a> Module<'a> {
             .map(Segment::ansi_string)
             .collect::<Vec<ANSIString>>();
 
+        /* Only print escaped sequences if we are printing in a prompt (or
+        somewhere the user wants the shell-specific escapes), but not if e.g.
+        someone has called `starship prompt` from the command line */
+        let use_shell_escapes =
+            std::env::var("STARSHIP_PRINT_SHELL_ESCAPES").unwrap_or("false".to_string());
+        let use_shell_escapes = use_shell_escapes == "true" || use_shell_escapes == "yes";
         let mut ansi_strings = match shell.as_str() {
-            "bash" => ansi_strings_modified(ansi_strings, shell),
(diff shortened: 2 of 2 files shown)


chipbuster:
Oooh, nailed by clippy. Never gotten that before. How exciting!

After thinking about this a little more, I think I prefer the approach proposed by @iamsauravsharma to use an environmental variable to control the shell name and a binary flag to control whether escape sequences get printed or not. I'll leave the PR like this until I get a chance to correct it.

Rationale: The same two invocations of starship should print the same *information*. Putting shell escapes into an envar breaks this rule, by allowing the same command to print different pieces of information (shell-specific escapes) depending on if the envar is set or not.

However, controlling *how* this information is interpreted and displayed can be a shell-specific problem as well, and not one that we want to pass args for all the time. For instance, it looks like (#179) zsh will need to handle job counts differently than other shells. We could handle this by requiring a shell name when either `--jobs` or `--escapes`(hypothetical name) is used...but this gets into a whole can of worms that makes it very annoying for the user to figure out which flags they need to pass. Much simpler, in this case, is to just put it in the environment.

Another potential example of this is the `STATUS` variable in the shell inits--in principle, we could pass this as an environmental variable, but right now it's only needed in one argument, so we pass it there. in the future, if we ever have two independent flags that both need to know the return code, we could instead opt to take the return code by environment.

This is a (very tired) first draft musing, so sorry if it makes little sense. The long and short is that I think
  - Flags/arguments should be used in most cases
  - Flags/arguments should *always* be used to change which items get printed
  - envars should be used for shell-specific config that's needed in multiple places

chipbuster:
Unfortunately, my proposed solution cannot work in the current codebase, as there's no way to pass arguments to the printing functions without significant refactoring. We'll have to save my proposed change for the (eventual) refactor we're going to do either right before or right after 1.0.

PR #217, opened 2019-08-21
fix: Restore Ruby module

### LINKED ISSUE #202: Ruby module is missing from default prompt order
#### Current Behavior
Directories that should display the ruby version in the prompt do not unless a custom prompt_order is specified. 

#### Expected Behavior
The Ruby module should be included in the `DEFAULT_PROMPT_ORDER` constant so that it's inclusion in the prompt is determined by the `starship.toml` config file.

#### Environment
- Starship version: `starship 0.10.1`
- Shell type: `zsh`
- Shell version: `zsh 5.7.1 (x86_64-apple-darwin18.2.0)`
- Shell plugin manager: `antigen`
- Terminal emulator: `iTerm2`
- Operating system: OSX `10.14.6`

Not relevant

#### Starship Configuration

```bash
# DOES NOT WORK WITHOUT THE FOLLOWING DECLARATION
prompt_order = [
  "character",
  "username",
  "directory",
  "git_branch",
  "git_status",
  "package",
  "nodejs",
  "ruby",
  "rust",
  "python",
  "golang",
  "cmd_duration",
  "line_break",
  "jobs",
  "battery",
]

add_newline = false

# Replace the "➜" symbol in the prompt with "λ"
[character]      # The name of the module we are confguring is "character"
symbol = " λ"     # The "symbol" segment is being set to "❯"
error_symbol = " λ"
use_symbol_for_status = true

[directory]
truncation_length = 5

[line_break]
disabled = true

[nodejs]
disabled = true

[ruby]
disabled = false
```

#### Possible Solution
Add "ruby" to the `DEFAULT_PROMPT_ORDER` constant. Or, have that list "discoverable" via metadata for each module.

Current values in `master` branch:

```
const DEFAULT_PROMPT_ORDER: &[&str] = &[
    "username",
    "directory",
    "git_branch",
    "git_status",
    "package",
    "nodejs",
    "rust",
    "python",
    "golang",
    "cmd_duration",
    "line_break",
    "jobs",
    "battery",
    "character",
];
```

chipbuster:
#### Description

Adds ruby module to `DEFAULT_PROMPT_ORDER`

#### Motivation and Context

Closes #202 

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -46,9 +46,9 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Variable       | Default | Description                                                        |
-| -------------- | ------- | ------------------------------------------------------------------ |
-| `add_newline`  | `true`  | Add a new line before the start of the prompt.                     |
+| Variable       | Default                       | Description                                            |
+| -------------- | ----------------------------- | ------------------------------------------------------ |
+| `add_newline`  | `true`                        | Add a new line before the start of the prompt.         |
 | `prompt_order` | [link](#default-prompt-order) | Configure the order in which the prompt module occurs. |
 
 ### Example
@@ -72,6 +72,7 @@ default_prompt_order = [
     "git_status",
     "package",
     "nodejs",
+    "ruby",
     "rust",
     "python",
     "golang",
--- a/src/print.rs
+++ b/src/print.rs
@@ -18,6 +18,7 @@ const DEFAULT_PROMPT_ORDER: &[&str] = &[
     "git_status",
     "package",
     "nodejs",
+    "ruby",
     "rust",
     "python",
     "golang",


chipbuster:
Also, I'm apparently making some whitespacing changes. Thanks, VSCode!

(Can we implement a `rustfmt`-like check for Markdown?)

docs/config/README.md
```diff
@@ -46,9 +46,9 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Variable       | Default | Description                                                        |
-| -------------- | ------- | ------------------------------------------------------------------ |
-| `add_newline`  | `true`  | Add a new line before the start of the prompt.                     |
+| Variable       | Default                       | Description                                            |
+| -------------- | ----------------------------- | ------------------------------------------------------ |
+| `add_newline`  | `true`                        | Add a new line before the start of the prompt.         |
 | `prompt_order` | [link](#default-prompt-order) | Configure the order in which the prompt module occurs. |
```
iamsauravsharma: Recently `starship module --list` flag was added which list all supported modules. So it will be better if docs was updated with this information about where user can get list of all supported values.

chipbuster: Sorry, not 100% sure what you mean by that. The docs have the default listing (under default_prompt_order) and `ruby` is already printed by `starship module --list`

iamsauravsharma: Currenly user cannot find out what is accepted value for prompt_order through docs. So directing user to `starship module --list` will help user to see all accepted value for prompt_list.

chipbuster: Is that specific to the Ruby module, or a general note for the docs?

matchai: Since "Prompt Order" will (I believe) always have the full list of modules, maybe we don't need to discern between the accepted values and the default list. 🤔

chipbuster: Sorry, I'm still a little lost here.

If we're talking about changes to the overall documentation structure or code structure, I'm all for it (I *really* don't like having three copies of the same information floating around in the codebase), but that seems like a separate issue.

matchai: We could continue this discussion elsewhere. 😛
I'll merge this for now.

PR #224, opened 2019-08-22
fix: Update init to use full path to starship executable

### LINKED ISSUE #220: starship doesn't work if not in PATH (starship: command not found)
#### Current Behavior
```
$ which starship
which: no starship in [...]
$ eval "$(devstarship/target/debug/starship init bash)"
starship: command not found
```

#### Expected Behavior
It should work.

#### Environment
- Starship version: 88d65f1
- Shell type: bash
- Shell version: 4.4.23(1)-release
- Operating system: NixOS

#### Possible Solution
The problem is that `starship init` and `starship prompt` only use `starship` and not its full path.

I have a PR for this.

bbigras:
#### Description
- get the current exe path with env::current_exe() **(which might cause security problems)**
- put the path in doublequotes to support spaces in the path
- escape doublequotes in the path

I didn't want to touch the big BASH_INIT, ZSH_INIT.. variable so I renamed the `starship` to `STARSHIP_EXEC` and used a replace. I wonder if it's less optimal than using the format macro or something else.

#### Motivation and Context

Closes #220

#### Types of changes

#### Checklist:

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -1,5 +1,6 @@
 use std::ffi::OsStr;
 use std::path::Path;
+use std::{env, io};
 
 /* We use a two-phase init here: the first phase gives a simple command to the
 shell. This command evaluates a more complicated script using `source` and
@@ -11,29 +12,45 @@ rest of the script, and you have to spam semicolons everywhere. By using
 source and process substitutions, we make it possible to comment and debug
 the init scripts. */
 
+fn path_to_starship() -> io::Result<String> {
+    let current_exe = env::current_exe()?
+        .to_str()
+        .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
+        .to_string();
+    Ok(current_exe)
+}
+
 /* This prints the setup stub, the short piece of code which sets up the main
 init code. The stub produces the main init script, then evaluates it with
 `source` and process substitution */
-pub fn init_stub(shell_name: &str) {
+pub fn init_stub(shell_name: &str) -> io::Result<()> {
     log::debug!("Shell name: {}", shell_name);
 
     let shell_basename = Path::new(shell_name).file_stem().and_then(OsStr::to_str);
+
+    let starship = path_to_starship()?.replace("\"", "\"'\"'\"");
 
     let setup_stub = match shell_basename {
         Some("bash") => {
             /* This *should* look like the zsh function, but bash 3.2 (MacOS default shell)
             does not support using source with process substitution, so we use this
             workaround from https://stackoverflow.com/a/32596626 */
-            let script = "source /dev/stdin <<<\"$(starship init bash --print-full-init)\"";
+            let script = format!(
+                "source /dev/stdin <<<\"$(\"{}\" init bash --print-full-init)\"",
+                starship
+            );
             Some(script)
         }
         Some("zsh") => {
-            let script = "source <(starship init zsh --print-full-init)";
+            let script = format!("source <(\"{}\" init zsh --print-full-init)", starship);
             Some(script)
         }
         Some("fish") => {
             // Fish does process substitution with pipes and psub instead of bash syntax
-            let script = "source (starship init fish --print-full-init | psub)";
+            let script = format!(
+                "source (\"{}\" init fish --print-full-init | psub)",
+                starship
+            );
             Some(script)
         }
         None => {
@@ -62,15 +79,18 @@ pub fn init_stub(shell_name: &str) {
     if let Some(script) = setup_stub {
         print!("{}", script);
     };
+    Ok(())
 }
 
 /* This function (called when `--print-full-init` is passed to `starship init`)
 prints out the main initialization script */
-pub fn init_main(shell_name: &str) {
+pub fn init_main(shell_name: &str) -> io::Result<()> {
+    let starship_path = path_to_starship()?.replace("\"", "\"'\"'\"");
+
     let setup_script = match shell_name {
-        "bash" => Some(BASH_INIT),
-        "zsh" => Some(ZSH_INIT),
-        "fish" => Some(FISH_INIT),
+        "bash" => Some(BASH_INIT.replace("## STARSHIP ##", &format!("\"{}\"", starship_path))),
+        "zsh" => Some(ZSH_INIT.replace("## STARSHIP ##", &format!("\"{}\"", starship_path))),
+        "fish" => Some(FISH_INIT.replace("## STARSHIP ##", &format!("\"{}\"", starship_path))),
         _ => {
             println!(
                 "printf \"Shell name detection failed on phase two init.\\n\
@@ -83,6 +103,7 @@ pub fn init_main(shell_name: &str) {
     if let Some(script) = setup_script {
         print!("{}", script);
     };
+    Ok(())
 }
 
 /* GENERAL INIT SCRIPT NOTES
@@ -132,10 +153,10 @@ starship_precmd() {
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME=$(date +%s)
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
-        PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)"
+        PS1="$(## STARSHIP ## prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)"
         unset STARSHIP_START_TIME
     else
-        PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)")"
+        PS1="$(## STARSHIP ## prompt --status=$STATUS --jobs="$(jobs -p | wc -l)")"
     fi
     PREEXEC_READY=true;  # Signal that we can safely restart the timer
 }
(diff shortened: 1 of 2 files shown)


src/init.rs
```diff
@@ -11,29 +12,46 @@ rest of the script, and you have to spam semicolons everywhere. By using
...
 init code. The stub produces the main init script, then evaluates it with
 `source` and process substitution */
-pub fn init_stub(shell_name: &str) {
+pub fn init_stub(shell_name: &str) -> io::Result<()> {
     log::debug!("Shell name: {}", shell_name);
 
     let shell_basename = Path::new(shell_name).file_stem().and_then(OsStr::to_str);
 
+    // let starship = get_starship_path()?.replace("\"", "\"'\"'\"");
+    let starship = get_starship_path()?.replace("\"", "\"\\\"\"");
```
bbigras: shellcheck whines about this one but not `.replace("\"", "\"'\"'\"")`.

chipbuster:
Unfortunately, I think I see an attack path here which can be exploited. It's similar to the [pulseaudio one](https://vulners.com/securityvulns/SECURITYVULNS:DOC:22183?utm_source=securityvulns&utm_medium=redirect), except that our binary isn't suid, so you can't automatically gain root privileges, but it could still lead to nasty stuff. 

Just for an illustration, let's say that the user being attacked isn't root, but has some juicy information (e.g. passwords) stored in a directory in their `HOME`. The attack looks like this:

1. The attacker has regular user access on the target machine, but doesn't have access to the files.
2. The attacker links `starship` to a location of their choosing.
3. The victim starts up a shell.
4. The shell executes `eval $(starship init bash)`.
5. Because of the link, the `current_exe()` resolves to the `starship` under the attacker's control.
6. The attacker replaces their copy of starship with a binary that prints `chmod -R 777 ${HOME}`
7. The shell evaluates `source <( attacker_executable )` which runs `chmod -R 777 ${HOME}`.
8. Attacker waltzes into victim's home directory and steals all their ~bitcoin~passwords.

Note that as long as the attacker can replace the executable between steps 4 and 7, they can trick the victim into executing anything. If the victim is root, they can drop an executable that prints `usermod -a -G attacker_account sudo` and they'll get root.

Even if they miss their window between (4) and (7), the hardlink is going to persist, and the attacker's executable will now be executed by the victim's shell every time they draw a prompt (which is a fairly common operation). They can change the executable to read the victim's files, vandalize them, etc. etc.

Granted I don't 100% understand the pulseaudio attack (in particular, this seems like a common-enough operation that most "user-friendly" linuxes should provide some sort of protection against it), but if that attack works, I think this one will too.

If there are any security folk out there that wanna come in and tell me I've horribly misunderstood how this works, I'd be happy to hear it...

src/init.rs
```diff
@@ -11,29 +12,46 @@ rest of the script, and you have to spam semicolons everywhere. By using
 source and process substitutions, we make it possible to comment and debug
 the init scripts. */
 
+fn get_starship_path() -> io::Result<String> {
```
chipbuster: Left a long-winded comment about this and potential privilege escalation attacks on the PR. Long and short, I'm not sure this is secure and I don't think it's salvageable with `current_exe`, at least on Linux (where the implementation reads /proc/self/exe`.

Do you think it would be possible to do this by reading `argv[0]` and using `fs::canonicalize`? That should get us a pretty similar effect without worrying about hardlink trickery. The only problem is that I'm not sure how to read `argv[0]` using clap (there was an issue opened about it, but it seems to have died without resolution).

src/init.rs
chipbuster: Can we do this with raw string literals? Would make it a lot easier to read imo, especially if you have syntax highlighting.

```suggestion
    let starship = get_starship_path()?.replace(r#"""#,r#""\"""#);
```

...okay, maybe "easier" is the wrong word, but you don't have to work as hard parsing the backslashes. I'm not even sure if the suggestion is correct, as it stands right now.

src/init.rs
```diff
@@ -11,29 +12,46 @@ rest of the script, and you have to spam semicolons everywhere. By using
...
+        .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
+        .to_string();
+    Ok(current_exe)
+}
+
 /* This prints the setup stub, the short piece of code which sets up the main
 init code. The stub produces the main init script, then evaluates it with
 `source` and process substitution */
-pub fn init_stub(shell_name: &str) {
+pub fn init_stub(shell_name: &str) -> io::Result<()> {
```
chipbuster: A little confused by this change. Is it just stylistic, or does it enable something in the code that we didn't have before?

src/init.rs
```diff
@@ -62,15 +80,18 @@ pub fn init_stub(shell_name: &str) {
...
 prints out the main initialization script */
-pub fn init_main(shell_name: &str) {
+pub fn init_main(shell_name: &str) -> io::Result<()> {
+    let starship_path = get_starship_path()?.replace("\"", "\"\\\"\"");
+
     let setup_script = match shell_name {
-        "bash" => Some(BASH_INIT),
-        "zsh" => Some(ZSH_INIT),
-        "fish" => Some(FISH_INIT),
+        "bash" => Some(BASH_INIT.replace("STARSHIP_EXEC", &format!("\"{}\"", starship_path))),
```
chipbuster: Smart :clap: . I had it in my head for the longest time to use `lazy_static!` for this purpose, but this is clearly much cleaner.

src/init.rs
chipbuster: (and same question for `init_main`)

src/init.rs
```diff
@@ -132,10 +154,10 @@ starship_precmd() {
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME=$(date +%s)
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
-        PS1="$(starship prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)"
+        PS1="$(STARSHIP_EXEC prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)"
```
chipbuster: On first reading, I thought `STARSHIP_EXEC` was a shell variable. It wasn't until second reading that I realize that it's going to be nuked before it ever makes it to the shell.

Could we make this a string that's invalid as a shell variable, e.g. `## STARSHIP ##` or something so that there's no chance someone thinks it's executed as shell code?

src/init.rs
chipbuster: On second thought, I don't have a clear opinion on which one is better: one forces you to parse escape sequences in your head, one makes it impossible to visually separate adjacent double-quotes. Pick the one you think is better.

chipbuster: Just me being persnickety: could we change this name to something like `path_to_starship`?

bbigras: Yes, done.

src/init.rs
bbigras: Good idea. I renamed it to `## STARSHIP ##`.

src/init.rs
bbigras: Thanks ☺️

src/init.rs
bbigras: Normally I try to avoid using `unwrap()` an `expect()` and I prefer to use `Result` everywhere. But I could use `expect()` on `path_to_starship()`. I guess it shouldn't fail.

chipbuster: Oh, I see, this is for error handling outside the function.

Keep it then. We need to start removing places where things panic, and moving to `Result` and `Option` is a good step in that direction. I was just wondering if there was something else I was missing.

src/init.rs
bbigras: I did a quick test in `main()`.

```rust
use std::env;
println!("current_exe: {:?} {:?}", std::env::current_exe(), std::env::current_exe().map(|v| std::fs::canonicalize(v)));

for argument in env::args() {
    println!("ARG> {} {:?}", argument.clone(), std::fs::canonicalize(argument));
}
```

I created an hardl link (test_hardlink) pointing to starship and I got:
```
current_exe: Ok("/home/bbigras/test_hardlink") Ok(Ok("/home/bbigras/test_hardlink"))
ARG> ./test_hardlink Ok("/home/bbigras/test_hardlink")
thread 'main' panicked at 'not yet implemented', src/main.rs:28:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

src/init.rs
bbigras: Gotcha.

bbigras:
I wonder why direnv (a popular Go project) doesn't seem to care about it).

https://github.com/direnv/direnv/blob/54cb3c5a91edb377b4b7387e2cda43dc39064a4a/config.go#L54-L61

https://github.com/direnv/direnv/blob/9c4aba849ba4ffe841bba57b05237dc97cfc9fda/shell_bash.go#L9-L18

https://golang.org/pkg/os/#Executable

chipbuster:
> I wonder why direnv (a popular Go project) doesn't seem to care about it).

Looking at the [source of `executable`](https://github.com/golang/go/blob/835f983da6866deec1c6550555c96de729c8848c/src/os/executable_path.go) (which is the function underlying `os.Executable`), it looks like they do canonicalization of `argv[0]`.

Rust's `current_exe` uses (at least on linux) the `/proc/self/exe` mechanism which is apparently prone to spoofing.

src/init.rs
chipbuster: I've gotten a few opinions from people who are better at systems stuff than me. I think I'm okay with using current_exe for this--the danger comes when using an suid binary that relaunches itself.

bbigras: Excellent, thanks!

bbigras:
```rust
    // let starship = get_starship_path()?.replace("\"", "\"'\"'\"");
    let starship = get_starship_path()?.replace("\"", "\"\\\"\"");
```

@chipbuster it's me or I I saw you comment on those lines but I can't find it in all the resolved conversations. 

I'm thinking of using the first line to prevent shellcheck from whining.

chipbuster:
@bbigras I'm convinced that GitHub ate something on this thread last night, but it's possible I just fumbled the keystrokes .

What's shellcheck complaining about? Some of the message it throws out are for bad practices in handwritten shell scripts, but not necessarily in autogenerated ones.

(I think I did have a comment about potentially using raw-strings instead of escapes all over the place, but I think it's a sidegrade at best, as you end up trading "mentally parsing backslashes" for "trying to figure out how many double/single quotes there are here")

bbigras:
with `replace("\"", "\"\\\"\"")`:
```
In 1.sh line 2:
source /dev/stdin <<<"$("/home/bbigras2/a 'te"\""st/starship" init bash --print-full-init)"
       ^--------^ SC1091: Not following: /dev/stdin was not specified as input (see shellcheck -x).
                                              ^-- SC2140: Word is of the form "A"B"C" (B indicated). Did you mean "ABC" or "A\"B\"C"?
```

with `replace("\"", "\"'\"'\"")` it's fine.

chipbuster:
Looks like a shellcheck misfire to me (where the word `B` is `\"`), but if the other form works and keeps the linters happy, let's keep it!

chipbuster:
@all-contributors, let's add @bbigras for the excellent code. Please and thanks!

bbigras:
Thanks for the merge and the all-contributors PR @chipbuster.

PR #233, opened 2019-08-23
ci: Migrate CI from Azure Pipelines to GitHub Actions

matchai:
#### Description

Migrated CI from Azure Pipelines to GitHub Actions.
Until the release process is figured out in Actions, we'll stick to using Azure pipelines for releases.

#### Motivation and Context

Closes #

#### Types of changes

#### Checklist:


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,79 @@
...
+      - uses: actions/checkout@master
+      - run: cargo test  -- -Z unstable-options --include-ignored
+
+  # Run the tests in the Docker image
+  docker_test:
+    needs: [rustfmt, clippy, cargo_check]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - run: docker pull starshipcommand/starship-test
```
JasonEtco: Is this the same image as `tests/Dockerfile`? If so, maybe it makes more sense to build from there, rather than pulling the public one - that way changes to the Dockerfile are automatically tests. You _could_ do both even, but that feels redundant.

matchai: I just pull the public one so it'll run faster.
In the end `./integration_test` does build `tests/Dockerfile` using the cache from the pulled version.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,106 @@
...
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macOS-latest]
+    steps:
+      - uses: hecrj/setup-rust-action@master
+      - uses: actions/checkout@master
+
+      # Install cross
+      # Temporary fix until https://github.com/rust-embedded/cross/pull/169 is merged.
+      - run: git config --global user.email "not_necessery@dont.need"
```
JasonEtco: You can do a multi-line run:

```yaml
  - run: |
      echo one
      echo two
      echo "Hey Matan"
```

matchai: Omg this changes everything!
I was looking everywhere in the docs for an example of this. 🤯

tests/testsuite/python.rs
```diff
@@ -110,42 +109,3 @@ fn with_virtual_env() -> io::Result<()> {
     assert_eq!(expected, actual);
     Ok(())
 }
-
```
chipbuster: Why are we deleting this? Has the test been obsoleted by something?

matchai:
> Just to confirm: it looks like we still do release binaries via Azure? Does GH Actions not have support for that yet?
It does, but it's configured a fair bit differently.
I wanted to experiment first in a test repo before trying it out here.

PR #252, opened 2019-08-28
Change ZSH init to use jobstates for num jobs

### LINKED ISSUE #179: ZSH Jobs module has wrong count after `cd`
## Bug Report

Well this is a weird one.

#### Current Behavior

Open up `starship` with `zsh`. Then run the following commands:

```
sleep 5
^Z
cd <some other directory>
```

where `^Z` refers to putting `sleep` into the background with Ctrl+Z.

You should now see that the jobs counter shows two jobs instead of the expected 1. Strangely, running `jobs` at this point only shows one line. Even more strangely, going back to the original directory that `sleep` was run from makes the job counter appear correctly again.

This appears to be a zsh-only behavior--this could not be reproduced in bash or fish.

#### Expected Behavior

Number of jobs should not depend on the current working directory.

image

#### Environment
- Starship version: 0.10.1, from commit dfe2ae643af77b882ae1cfbe81fa74d62250eb93
- Shell type: zsh
- Shell version: zsh 5.7.1 (x86_64-apple-darwin18.2.0)
- Shell plugin manager: antigen + custom
- Terminal emulator: iTerm
- Operating system: MacOS 10.14.6

Reproduces in a shell with no config.

#### Starship Configuration
N/A

chipbuster:
#### Description
Change ZSH jobs to use `jobstates` array provided by zsh/parameters module rather than using `jobs | wc`, which can produce errors.

Marked as draft because it works on my machine, but I'd like to test several ZSH versions across multiple machines before committing it in to the main codebase.

#### Motivation and Context

Closes #179 

#### Types of changes

#### Checklist:

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -208,19 +208,24 @@ ever drawn once (for the prompt immediately after it is run).
 */
 
 const ZSH_INIT: &str = r##"
+zmodload zsh/parameter  # Needed to access jobstates variable for NUM_JOBS
+
 # Will be run before every prompt draw
 starship_precmd() {
     # Save the status, because commands in this pipeline will change $?
     STATUS=$?
 
+    # Use length of jobstates array as number of jobs. Expansion fails inside
+    # quotes so we set it here and then use the value later on.
+    NUM_JOBS=$#jobstates  
     # Compute cmd_duration, if we have a time to consume
     if [[ ! -z "${STARSHIP_START_TIME+1}" ]]; then
         STARSHIP_END_TIME="$(date +%s)"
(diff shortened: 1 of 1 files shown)


chipbuster:
Well, it works going back to zsh-5.0.8.

I get a message suggesting that it should work in zsh-4.2.7, but my shared objects are screwed up, so it can't actually load. ZSH-4.0.9 won't compile on my machine.

Given that the oldest version of ZSH I'm aware of being out in the wild is the 5.0.x line (which was released in 2012 and is still present on CentOS/RHEL 7), I'm going to call this test result "good enough for me." If anybody would like me to test the 2007 ZSHses more closely, please leave a comment saying so.

PR #264, opened 2019-08-30
fix: Fixes the `nix_shell` module, allowing it to be used with `starship prompt`

nickwb:
#### Description
Adds `nix_shell` to `module::ALL_MODULES`, so that it is not skipped when generating the user's prompt.

#### Motivation and Context
Currently including `nix_shell` in your `prompt_order` has no effect.

#### Types of changes

Not yet. I'm having some build issues with `git2`...

#### Checklist:

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -17,6 +17,7 @@ pub const ALL_MODULES: &[&str] = &[
     "jobs",
     "line_break",
     "nodejs",
+    "nix_shell",
     "package",
     "python",
     "ruby",


nickwb:
@bbigras 

> That's weird. The nix-shell module did work for me.

I think if you are using default prompt order, rather than an overridden prompt order, it would work fine without this change.

See here: https://github.com/starship/starship/blob/master/src/print.rs#L56 for where the check occurs, but the check only occurs when `config.get_as_array("prompt_order")` is `Some`.

bbigras:
Gotcha. Thanks!

PR #267, opened 2019-08-31
feat: clap -> structopt

bbigras:
#### Description
switch from clap to structopt (which uses clap).
[structopt 0.3 was just released](https://old.reddit.com/r/rust/comments/cxgw86/announcing_structopt_03/)

#### Motivation and Context
Structopt is pretty nice. It allows to specify a struct with all the command line options.
It also supports default values and parsing.

Using an enum for the subcommands is also nice to prevent bugs if we use a flag from another subcommand.

This is a simple PR just to see if you are interested. If this get merged maybe the next step could be to pass the subcommand's enum to the right functions instead of matching the enum everywhere like I did.

#### Types of changes

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -303,6 +303,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "heck"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "unicode-segmentation 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "humantime"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


chipbuster:
Interesting idea! I saw the structopt release announcement on /r/rust, but I didn't get a chance to follow up on it much.

I guess the questions I have are the typical big ones:

  - What do we gain by switching to structopt vs clap? Is it just being able to specify a struct of opts or is there another benefit? (Or perhaps I don't properly appreciate how big of a benefit that is).

  - By SemVer rules, clap should be API-stable while structopt is not. Do we know what this looks like in practice (i.e. has structopt historically broken API guarantees?).

  - I noticed that we no longer have to specify lifetimes for contexts (I'm guessing they used to hold borrowed strings and now hold an opt struct?). Does that mean that our modules no longer need lifetime specifications as well?

PR #276, opened 2019-09-03
feat: Adds Git State module for showing "REBASING 2/3", etc.

nickwb:
This PR adds a new module `git_state`. It will show in directories which are part of a git repository, and where there is an operation in progress, such as: rebasing, bisecting, etc.

See the updated docs for more details.

#### Description
- Adds the `git_state` module.
- Adds `git_state` to the default prompt order
- Updates the documentation to describe the `git_state` module

#### Motivation and Context
I noticed when rebasing that starship showed a rather unhelpful "HEAD" as the branch name.

#### Types of changes

image

#### Checklist:


src/modules/git_state.rs
```diff
@@ -0,0 +1,165 @@
+use ansi_term::Color;
+use git2::{Repository, RepositoryState};
+use std::path::Path;
+
+use super::{Context, Module};
+
+/// Creates a module with the state of the git repository at the current directory
+///
+/// During an git operation it will show: REBASING, BISECTING, MERGING, etc.
```
wyze: Typo, `an` -> `a`.

wyze:
I also prefer `Rebasing` over `REBASING` as the default, but maybe it is just me. I wouldn't have an issue using the configuration to change them though. Just seeing what others might think.

nickwb:
Thanks for the feedback @wyze!

I also don't feel strongly about the default text. The uppercase seemed the nicest to me aesthetically, but like you I wouldn't be worried if I just configured this myself.

Thanks for the tip on the integration tests.. I didn't event realise they were there 😅 
I'll have a go at writing some tomorrow.

Is there a reason that all of the git status tests are set to ignored right now..?

I've pushed a fix for the typo!

wyze:
> Is there a reason that all of the git status tests are set to ignored right now..?

See https://github.com/starship/starship/pull/247#issuecomment-525447155

chipbuster:
> Is there a reason that all of the git status tests are set to ignored right now..?

The short version is that we're using 'ignored' as a proxy for 'only run in Docker test container'. Those tests rely on certain software versions, certain files being present, etc, so it's not safe to run them outside a virtualized environment.

nickwb:
I've started putting together some integration tests. Interestingly, they are passing locally but not inside the docker image.

I'll investigate more tomorrow...

wyze:
> Interestingly, they are passing locally but not inside the docker image.

Seems like it is failing on the formatting check before it gets to the tests.

nickwb:
Ok, the build is happy again 😃 

I think this is ready to go now. Any further thoughts on this pull request?

nickwb:
Ah, before you merge, I just realised I lost the fix for the typo somewhere along the way..

nickwb:
Ok - I rebased to master and fixed that typo. Fingers crossed the build goes green and it's ready to go.

src/modules/git_state.rs
```diff
@@ -0,0 +1,165 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let repo_root = context.repo_root.as_ref()?;
+    let mut repository = Repository::open(repo_root).ok()?;
+    let state_description = get_state_description(&mut repository);
+
+    if let StateDescription::Clean = state_description {
+        return None;
+    }
+
+    let mut module = context.new_module("git_state")?;
```
chipbuster: Can we move this to the top of the function? A la #283, touching the filesystem is sometimes a very expensive operation, and we'd like to avoid doing it at all if the module is disabled.

docs/config/README.md
```diff
@@ -230,6 +232,36 @@ truncation_length = "4"
...
+| `merge`            | `"MERGING"`        | The text displayed when a `merge` is in progress.                                                                |
+| `revert`           | `"REVERTING"`      | The text displayed when a `revert` is in progress.                                                               |
+| `cherry_pick`      | `"CHERRY-PICKING"` | The text displayed when a `cherry-pick` is in progress.                                                          |
+| `bisect`           | `"BISECTING"`      | The text displayed when a `bisect` is in progress.                                                               |
+| `am`               | `"AM"`             | The text displayed when an `apply-mailbox` (`git am`) is in progress.                                            |
+| `am_or_rebase`     | `"AM/REBASE"`      | The text displayed when an ambiguous `apply-mailbox` or `rebase` is in progress.                                 |
+| `progress_divider` | `"/"`              | The symbol or text which will separate the current and total progress amounts. (e.g., `" of "`, for `"3 of 10"`) |
+
```
chipbuster: Does this module support a `disabled` option?

docs/config/README.md
```diff
@@ -230,6 +232,36 @@ truncation_length = "4"
 truncation_symbol = ""
 ```
 
+## Git State
+
+The `git_state` module will show in directories which are part of a git
```
chipbuster: Nice, clean explanation.

src/modules/git_state.rs
```diff
@@ -0,0 +1,165 @@
...
+    module.get_prefix().set_value("(");
+    module.get_suffix().set_value(") ");
```
chipbuster: Are the prefix and suffix set this way specifically because they're meant to be used right next to the other `git` modules? I think we usually go for something like " on " or " in " for defaults.

src/modules/git_state.rs
```diff
@@ -0,0 +1,165 @@
...
+    module.set_style(Color::Yellow.bold());
+
+    let label = match state_description {
+        StateDescription::Label(label) => label,
+        StateDescription::LabelAndProgress(label, _) => label,
+        // Should only be possible if you've added a new variant to StateDescription
```
chipbuster: I love seeing explanations for when panics occur or why things are safe. :+1:

src/modules/git_state.rs
```diff
@@ -0,0 +1,165 @@
...
+        _ => panic!("Expected to have a label at this point in the control flow."),
+    };
+
+    module.new_segment(label.segment_name, label.message_default);
+
+    if let StateDescription::LabelAndProgress(_, progress) = state_description {
+        module.new_segment("progress_current", &format!(" {}", progress.current));
+        module.new_segment("progress_divider", "/");
+        module.new_segment("progress_total", &format!("{}", progress.total));
```
chipbuster: IIRC, if there's a TOML option by the name, then the value is overridden with that value. Does that mean we've created (unintended, undocumented) TOML options here which could break the module if they were set?

src/modules/git_state.rs
```diff
@@ -0,0 +1,165 @@
...
+    let just_label = StateDescription::Label(&REBASE_LABEL);
+
+    let dot_git = repository
+        .workdir()
+        .and_then(|d| Some(d.join(Path::new(".git"))));
+
+    let dot_git = match dot_git {
+        None => {
+            // We didn't find the .git directory.
+            // Something very odd is going on. We'll just back away slowly.
```
chipbuster: :laughing:

tests/testsuite/git_state.rs
```diff
@@ -0,0 +1,177 @@
...
+    let status = command.status()?;
+
+    if expect_ok && !status.success() {
+        Err(Error::from(ErrorKind::Other))
+    } else {
+        Ok(())
+    }
+}
+
+fn create_repo_with_conflict() -> io::Result<tempfile::TempDir> {
```
chipbuster: Well. That is one monster of a function, and you've written it beautifully. 10/10, would read again.

src/modules/git_state.rs
chipbuster: It's not a big deal if the answer is "yes," it's just going to be a mental note in the 1.0 refactor to have ways to avoid this.

src/modules/git_state.rs
nickwb: Ah, good point. I actually missed that `new_module` checked the `disabled` setting. 👍 

I'll fix that shortly.

docs/config/README.md
nickwb: I guess all modules implicitly do. I will update the docs 😄

src/modules/git_state.rs
nickwb: Honestly, this choice was mostly to ensure that the prompt doesn't get too long. In the default prompt order, this comes directly after the branch name, so it usually makes sense it context.

That said, I'm happy to make changes if people think there are better options..?

src/modules/git_state.rs
nickwb: Yeah, the answer is yes. I think this issue affects most modules in some form or another.

src/modules/git_state.rs
chipbuster: I think this is a fine compromise to make for now. I think at some point in the future, we're going to have to start adding shell adapter code that tells starship what size the terminal is, and let starship decide what gets sacrificed if the terminal is too small to fit everything.

(I also seem to recall properly handling SIGWINCH being a PITA, so I'm really looking forward to that...)

nickwb:
@chipbuster - I've added the changes related to `disabled` as discussed

src/modules/git_state.rs
chipbuster: Alright, we'll kick that particular can down the road for now then.

chipbuster:
Hmmm....seems like I'm still being blocked from merging by @wyze's requested changes.

nickwb:
> Hmmm....seems like I'm still being blocked from merging by @wyze's requested changes.

Build is still running :smile:

wyze:
> Hmmm....seems like I'm still being blocked from merging by @wyze's requested changes.

Hah, I didn't think my reviews counted enough to block a merge!

src/modules/git_state.rs
matchai: Haha. You have uncovered my janky implementation of configuration. 🙃 
I would definitely like to rework it during the refactor.

matchai:
@all-contributors Please add @nickwb for code, docs and test. Also please add @wyze for review.

matchai:
@all-contributors Please add @nickwb for code, docs and test.

matchai:
@all-contributors Please add @wyze for review.

PR #241, opened 2019-08-26
Fix for Termux not having '/dev/stdin'

bbigras:
#### Description
Use `$BASH_VERSINFO` to determine if `source <(starship init bash --print-full-init)` can be used.

This can only work if bash has `$BASH_VERSINFO` in 3.2.

#### Motivation and Context
To make `eval "$(starship init bash)"` work on Termux.

Closes #235

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #278, opened 2019-09-03
fix: Be more restrictive with bash init fallback

nickwb:
Changes the bash init fallback to be `Major Version > 3` rather than `Major Version > 4`.

#### Description
The issue being worked around affects bash 3.2, so `Major Version > 3` is sufficient.
The fallback requires `/dev/stdin`, which may not exist on all systems - so it's preferable to use the normal bootstrap when possible.

This should fix starship for some Windows users who are using bash without full POSIX emulation.

For example: Git Bash, which is part of Git for Windows, does not recognise `/dev/stdin`, but does ship with bash 4.x.

#### Motivation and Context
Improve support for starship on Windows 😃 

Bash on Windows is a bit of a minefield, because there are various projects maintaining various POSIX emulation layers and various patch lists. This change will not fix starship for ALL bash users on windows, but it should fix it for a portion of them.

#### Types of changes

(I did specifically test this on bash 3.2 on MacOS).

#### Checklist:

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -32,12 +32,40 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
 
     let setup_stub = match shell_basename {
         Some("bash") => {
-            /* This *should* look like the zsh function, but bash 3.2 (MacOS default shell)
-            does not support using source with process substitution, so we use this
-            workaround from https://stackoverflow.com/a/32596626 */
+            /*
+             * The standard bash bootstrap is:
+             *      `source <(starship init bash --print-full-init)`
+             *
+             * Unfortunately there is an issue with bash 3.2 (the MacOS
+             * default) which prevents this from working. It does not support
+             * `source` with process substitution.
+             *
+             * There are more details here: https://stackoverflow.com/a/32596626
+             *
+             * The workaround for MacOS is to use the `/dev/stdin` trick you
+             * see below. However, there are some systems with emulated POSIX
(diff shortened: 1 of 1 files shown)


chipbuster:
@nickwb Do you know what minor version git bash ships with specifically? I ask because we discovered while testing #241 that for some reason, the standard init pattern doesn't work with bash-4.0. The earliest version I can confirm it working with is 4.1.

nickwb:
Git for Windows 2.22.0 and 2.23.0 (the latest) both ship with bash: `GNU bash, version 4.4.23(1)-release (x86_64-pc-msys)`.

Sadly there is some regression in 2.23.0, because `eval "$(starship init bash)"` is broken in 23, but works fine in 22.

Bash 4.4.23 seems to have been around since 2.22.0 (December 2018), and Bash 4.4.x for numerous years before that.

nickwb:
@chipbuster - Do you think we should take the `bash >= 4.1` logic that was proposed at one point for #241?

Based on the versions of Git Bash I mentioned in the previous comment, v4.1 as a minimum should be suitable for everyone who has installed/updated Git for Windows in the last few years.

chipbuster:
@nickwb I think that's probably the best way to go here. It won't be perfect, but we'll be supporting a lot more configurations than we are now.

nickwb:
@chipbuster - Ok, I've pushed an update as discussed.

chipbuster:
@nickwb Please leave a comment reminding me to add you to all-contributors once #276 is merged.

PR #281, opened 2019-09-03
feat: Add config for ahead/behind count of tracked branch

### LINKED ISSUE #273: Add configuration to toggle the ahead/behind/diverged count in the git_status module

I personally find the `git_status` module to be a little noisy and hard to parse with the new ahead/behind/diverged count and would like the option to toggle it.

A new configuration option along the lines of `git_status.show_diverged_count`.

As it stands, there is no alternative available.

---

@wyze: Would you be interested in tackling this ticket, since you're already familiar with the implementation?

### LINKED ISSUE #292: Git Status: Separate segments for ahead/behind counts, so that the symbol can be configured without losing the count
## Feature Request
Currently the "symbol" and the "count" for ahead and behind are merged in to a single segment. As a consequence, if you override the symbol, you will not be shown the count.

Could we separate these in to four segments:

- `ahead`
- `ahead_count`
- `behind`
- `behind_count`

This will allow people to override `ahead` and `behind`, and also see the counts.

wyze:
#### Description
Add a configuration option (`show_sync_count`) to the `git_status` module that will show/hide the counts ahead/behind of the tracked branch. Currently have this default to `false`, and would opt-in to show this information.

#### Motivation and Context
Closes #273 and Closes #292 

#### Types of changes

#### Checklist:


matchai:
Thank you for the quick PR, @wyze. 😄 
From what I can tell, this looks like it hides the ahead/behind count when the status is "diverged". I was thinking that we would similarly hide the ahead/behind count when also "ahead" or "behind". Sorry if I wasn't clear in the issue.

Let me know what you think 👍

wyze:
I reread the issue and I can see how it would be in all 3 situations. I think I was thrown off by the configuration option name proposed as it only mentioned `diverged`. Is that still a good option name? Or something like `show_ahead_behind_count`?

matchai:
Gotcha. `show_ahead_behind_count` or maybe `show_sync_count`?
I can't think of a better config name right now. 🤔

wyze:
I like sync count better than what I came up with.

matchai:
> I like sync count better than what I came up with.

I good with using `show_sync_count` then. 👍

wyze:
@matchai Should be good to go now.

src/modules/git_status.rs
```diff
@@ -66,17 +69,37 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+        let add_ahead = |m: &mut Module<'a>| {
+            m.new_segment("ahead", GIT_STATUS_AHEAD);
+
+            if show_sync_count {
+                m.new_segment("ahead_count", &ahead.to_string());
+            }
+        };
+
+        let add_behind = |m: &mut Module<'a>| {
+            m.new_segment("behind", GIT_STATUS_BEHIND);
```
matchai: Nice work separating the segments! 👍

tests/testsuite/git_status.rs
```diff
@@ -90,6 +150,42 @@ fn shows_diverged() -> io::Result<()> {
...
+    let expected = Color::Red.bold().paint(format!("[{}] ", "⇕")).to_string();
+
+    assert_eq!(expected, actual);
+
+    Ok(())
+}
+
+#[test]
+#[ignore]
+fn shows_diverged_with_count() -> io::Result<()> {
```
matchai: Tests look great, as usual! 😜

PR #282, opened 2019-09-03
refactor: Move `create_fixture_repo` into common in integration tests

wyze:
#### Description
I moved the fixture repo creation to common since it is now used by more than one module.

#### Motivation and Context
Remove duplicated code.

#### Types of changes

#### Checklist:

### DIFF
--- a/tests/testsuite/common.rs
+++ b/tests/testsuite/common.rs
@@ -1,7 +1,8 @@
 use lazy_static::lazy_static;
 use std::io::prelude::*;
 use std::path::{Path, PathBuf};
-use std::{io, process};
+use std::process::Command;
+use std::{env, io, process};
 
 lazy_static! {
     static ref MANIFEST_DIR: &'static Path = Path::new(env!("CARGO_MANIFEST_DIR"));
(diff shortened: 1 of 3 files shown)


wyze:
This might be setting the config in the wrong repo. If it fails, I think better solution would be to check for CI environment and only set the global config if running there.

chipbuster:
Oh wow. That's definitely not part of the spec ^^

Were you running the complete integration suite (with ignored tests) outside of the container environment? We might want to consider putting a warning on that, since we're (implicitly) using `#[ignore]` as an `in-docker-only` flag at the moment.

wyze:
Yeah, I usually just run `cargo test -- -Z unstable-options --include-ignored` locally.

matchai:
> Yeah, I usually just run cargo test -- -Z unstable-options --include-ignored locally.

I'd suggest running `./integration_test` to run the tests in a Docker container with all our dependencies locked at fixed versions. The `ignored` tests will produce side effects and aren't meant to be run outside of Docker ([detailed here](https://github.com/starship/starship/blob/master/CONTRIBUTING.md#acceptance-testing)).

You might have ended up with a few files generated in your home directory. 😅

wyze:
Haha, oops. Okay, I will do that in the future. I think it's still worth while to move the fixture repo creation logic to common. I can revert the checking of git config if you'd like.

matchai:
> I think it's still worth while to move the fixture repo creation logic to common. I can revert the checking of git config if you'd like.

Sounds perfect. 👍

matchai:
> Looks great! Out of curiosity, why is it called 'fixture'? I'm not familiar with the term.

[Test fixtures](https://en.wikipedia.org/wiki/Test_fixture#Software) are a pretty common term for pre-established static data for setting up the test context. 😄

PR #285, opened 2019-09-04
feat: Add the ability to configure per-module color styles

### LINKED ISSUE #128: Implement module style configuration

Now that we have module configuration implemented for many of the prompt's modules, it would be a good time to look into implementing style configuration to be able to alter the default style values.

Style configuration would be provided in string form, specifying the color and text style to be applied. For instance, you could set a module's style to `"red"`, but `"bold underlined red"` would also work.

Here's an example:
```toml
[character]
style_success = "bold green"
style_failure = "red"
```

An alternative option would be to use an array or object to specify the color and text style, but I think that using strings would be much easier to read.

chipbuster:
#### Description

Allows the user to specify styles on a per-module (and potentially even per-section) basis.

This has two main parts: the first is a new config function which can read a string specifier, e.g. `underlined bold green bg:75` and return the appropriate `ansi_term::Style` object to match that specifier.

The second will be integrating this config function in with existing modules.

#### Motivation and Context

Closes #128 and #208 

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -79,11 +79,15 @@ Style strings are a list of words, separated by whitespace. The words are not ca
 
   - `bold`
   - `underline`
+  - `dimmed`
   - `bg:<color>`
   - `fg:<color>`
   - `<color>`
+  - `none`
 
-where `<color>` is a color specifier (discussed below). The last two are currently considered equivalent, though this may change in the future. The order of words in the string does not matter. 
+where `<color>` is a color specifier (discussed below). `fg:<color>` and `<color>` currently do the same thing , though this may change in the future. The order of words in the string does not matter.
+
+The `none` token overrides all other tokens in a string, so that e.g. `fg:red none fg:blue` will still create a string with no styling. It may become an error to use `none` in conjunction with other tokens in the future.
 
 A color specifier can be one of the following:
 
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -72,3 +72,26 @@ precmd_functions+=(set_win_title)
 
 If you like the result, add these lines to your shell configuration file 
 (`~/.bashrc` or `~/.zsrhc`) to make it permanent.
+
+## Style Strings
+
+Style strings are a list of words, separated by whitespace. Each word can be one of the following:
+
+  - `bold`
+  - `underline`
+  - `bg:<color>`
+  - `fg:<color>`
+  - `<color>`
+
+where `<color>` is a color specifier (discussed below). The last two are currently considered equivalent, though this may change in the future. The order of words in the string does not matter.
+
+A color specifier can be one of the following:
+
+ - One of the standard terminal colors: `black`, `red`, `green`, `blue`,
+    `yellow`, `purple`, `cyan`, `white`. You can optionally prefix these
+    with `bright-` to get the bright version (e.g. `bright-white`).
+ - A `#` followed by a six-digit hexadecimal number. This specifies an
+   [RGB color hex code](https://www.w3schools.com/colors/colors_hexadecimal.asp).
+ - A number between 0-255. This specifies an [8-bit ANSI Color Code](https://i.stack.imgur.com/KTSQa.png).
+
+If multiple colors are specified for foreground/background, the last one in the string will take priority.
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -75,7 +75,7 @@ If you like the result, add these lines to your shell configuration file
 
 ## Style Strings
 
-Style strings are a list of words, separated by whitespace. Each word can be one of the following:
+Style strings are a list of words, separated by whitespace. The words are not case sensitive (i.e. `bold` and `BoLd` are considered the same string). Each word can be one of the following:
 
   - `bold`
   - `underline`
@@ -83,7 +83,7 @@ Style strings are a list of words, separated by whitespace. Each word can be one
   - `fg:<color>`
   - `<color>`
 
-where `<color>` is a color specifier (discussed below). The last two are currently considered equivalent, though this may change in the future. The order of words in the string does not matter.
+where `<color>` is a color specifier (discussed below). The last two are currently considered equivalent, though this may change in the future. The order of words in the string does not matter. 
 
 A color specifier can be one of the following:
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -44,10 +44,11 @@ are segments within it. Every module also has a prefix and suffix that are the d
 
 Most modules in starship allow you to configure their display styles. This is done with an entry (usually called `style`, but some modules have different names) which is a string. Here are some examples of style strings along with what they do. For details on the full syntax, consult the [advanced config guide](/advanced-config/).
 
-- `fg:green bg:blue` sets green text on a blue background
-- `bg:blue fg:bright-green` sets bright green text on a blue background
-- `bold fg:27` sets bold text with [ANSI color](https://i.stack.imgur.com/KTSQa.png) 27
-- `underline bg:#bf5700` sets underlined text on a burnt orange background
+- `"fg:green bg:blue"` sets green text on a blue background
+- `"bg:blue fg:bright-green"` sets bright green text on a blue background
+- `"bold fg:27"` sets bold text with [ANSI color](https://i.stack.imgur.com/KTSQa.png) 27
+- `"underline bg:#bf5700"` sets underlined text on a burnt orange background
+- `""` explicitly disables all styling on the module
 
 ## Prompt
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -42,13 +42,15 @@ are segments within it. Every module also has a prefix and suffix that are the d
 
 ### Style Strings
 
-Most modules in starship allow you to configure their display styles. This is done with an entry (usually called `style`, but some modules have different names) which is a string. Here are some examples of style strings along with what they do. For details on the full syntax, consult the [advanced config guide](/advanced-config/).
+Most modules in starship allow you to configure their display styles. This is done with an entry (usually called `style`) which is a string specifying the configuration. Here are some examples of style strings along with what they do. For details on the full syntax, consult the [advanced config guide](/advanced-config/).
 
 - `"fg:green bg:blue"` sets green text on a blue background
 - `"bg:blue fg:bright-green"` sets bright green text on a blue background
 - `"bold fg:27"` sets bold text with [ANSI color](https://i.stack.imgur.com/KTSQa.png) 27
 - `"underline bg:#bf5700"` sets underlined text on a burnt orange background
-- `""` explicitly disables all styling on the module
+- `""` explicitly disables all styling
+
+Note that what styling looks like will be controlled by your terminal emulator. For example, some terminal emulators will brighten the colors instead of bolding text, and some color themes use the same values for the normal and bright colors.
 
 ## Prompt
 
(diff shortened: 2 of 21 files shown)


chipbuster:
@starship/astronauts Looks like I have a working draft of this puppy (screenshot at end). The last thing I'd like to do before integrating it with our existing modules is to make sure the syntax is good.
@nickwb I'd also like your opinion, if you have one on the syntax here.

The current rules for the stylestring are as follows:

#### Valid tokens
   - The user can put `bold`, `underline`, `fg`, `bg`, or a color into the string, separated by spaces.
#### Valid colors
A color can take on one of the three following forms:
   - A number between 0-255. This will be interpreted as an [ansi color](https://jonasjacek.github.io/colors/)
   - A hash followed by 6 hexadecimal digits. This is interpreted as a hexadecimal RGB code.
   - A standard color string (e.g. `red`, `blue`, `cyan`, etc.)
#### Foreground/Background
   - Any color appearing after an `fg` will modify the text color, any color appearing after a `bg` modifies the background color.
   - If multiple colors appear to a specifier (e.g. `fg red green bg blue purple`), the last one on the line is applied (in this example, it's equivalent to `fg green bg purple`).
   - If neither `fg` nor `bg` appear in the string, all colors are assumed to apply to the foreground.
#### Bold and Italic
   - The tokens `bold` and `italic` may appear anywhere in the string, multiple times.

We can add additional styling codes that are supported by ANSI terminals (e.g. italic, dimmed, strikethrough), but given that a lot of terminal emulators don't support some of these codes, we might want to be judicious about what we support.

Does anyone have improvements to suggest for how this little DSL works before I bake it into the codebase? Do we want to add support for any additional styling?

image

nickwb:
Looks like a good start @chipbuster, I like the different options you've provided for specifying colors.

Regarding the syntax, I do find it a little odd that `bold` are `underline` are independent from other tokens, but `fg`, `bg` and the colors are not.

For example, if you consider the current parser implementation, these strings are valid.
- `bg underline green`
- `bg green underline red` (Set's the bg twice!) 

Some possible solutions:
- Merge the tokens together: `bg:green`, `fg:137`
- Have formal grammar rules which force a color to appear after `fg` or `bg`.

chipbuster:
I like the colon solution. That was actually what I originally had in mind, but I guess I just derped because I didn't consider having separators (and things like `fggreen` look really weird).

Do you have any opinions on behavior if the same color is set twice? We can ignore subsequent settings, set to the last occurrence, or cause the parse to fail (presumably while logging a message).

matchai:
> A standard color string (e.g. red, blue, cyan, etc.)

How are bright vs normal colors handled? Would it be `bright-red`?

> We can ignore subsequent settings, set to the last occurrence, or cause the parse to fail (presumably while logging a message).

This comes back to us not having an elegant solution for handling errors. Ideally, the error would only be logged once after a configuration change is made so that the prompt containing the error doesn't get in the way of the user trying to fix the issue. For the time being, we can simply respect the last occurrence, and eventually transition to error handling, like we have noted elsewhere in the codebase.

> Does anyone have improvements to suggest for how this little DSL works before I bake it into the codebase?

Does anyone think that the DSL might be too complicated for advanced style configuration? I think it certainly would be simple for the most common use cases: `green`, `bold red`, etc. 

Another option to allow for clearer control for advanced styling would be to use a nested table for style.

```toml
[[character]]
symbol = "👋"

  [character.style]
  foreground = "bold red"
  background = "blue"
```

Thoughts?

chipbuster:
> How are bright vs normal colors handled? Would it be bright-red?

Haven't thought that far ahead yet. There's actually not an explicit enum in `ansi_term` for bright colors, so I might just rework the match arms to return `Fixed(0..15)` instead of using the explicit color enums like I am right now.

> For the time being, we can simply respect the last occurrence, and eventually transition to error handling

Sounds good to me.

> Does anyone think that the DSL might be too complicated for advanced style configuration?

Do you mean too complicated in terms of for the user to understand, or for us to implement? I think with the change suggested by @nickwb the syntax becomes pretty simple to explain (whether it stays that way or not is another story)

chipbuster:
Alright, the current language is the following:

- `underline`, `bold`, `bg:<color>`, `fg:<color>`, or `<color>` (the last two are currently considered equivalent).
- `<color>` can be:
  + a name of one of the 8 basic colors, optionally prefixed with `bright-`
  + a decimal number between 0-255 (ANSI colors)
  + a `#` followed by a six-digit hex RGB code

We still take the last instance of any color if multiple colors are found in the format string.

docs/config/README.md
```diff
@@ -40,6 +40,15 @@ are segments within it. Every module also has a prefix and suffix that are the d
  "via "         "⬢"        "v10.4.1"       ""
 ```
 
+### Style Strings
+
+Most modules in starship allow you to configure their display styles. This is done with an entry (usually called `style`, but some modules have different names) which is a string. Here are some examples of style strings along with what they do. For details on the full syntax, consult the advanced config guide.
```
chipbuster: How do I link this to the actual advanced config guide?

matchai: It looks like the syntax would be `[advanced config guide](/advanced-config/)`

https://vuepress.vuejs.org/guide/markdown.html#links

chipbuster:
I accidentally introduced a bug which I think is useful as a feature: if the color config string is completely invalid, it acts the same way as a blank config (turns off all styling). I think this could be helpful if the user types in an invalid config string: they now know that starship has seen their update to the config, but that something is going wrong (as opposed to going to the default color, where you're not sure if its because starship didn't see the config change or not).

src/config.rs
```diff
@@ -157,11 +160,130 @@ impl Config for Table {
...
+ - 'bg:<color>'    (specifies that the color read should be a background color)
+ - 'underline'
+ - 'bold'
+ - '<color>'        (see the parse_color_string doc for valid color strings)
+*/
+fn parse_style_string(style_string: &str) -> Option<ansi_term::Style> {
+    let tokens = style_string.split_whitespace();
+    let mut style = ansi_term::Style::new();
+
+    // Should we color the foreground? If not, assume we color the background.
```
wyze: Is the sentence here correct? Seems like we default to foreground if no `fg:`/`bg:` token was found.

chipbuster: Ah, I think this is poorly worded.

The intent was to say that "if `col_fg` is `false`, then we're coloring the background", as opposed to "if `col_fg` is false, don't color anything".

chipbuster: The other option I had would have been to introduce an Enum for it, but that felt a little overkill for an either-or decision.

wyze: Thanks. It reads much better now.

matchai:
Could we have an example where the style of a module is explicitly set to nothing?
```toml
[character]
style = "none"
```

docs/advanced-config/README.md
```diff
@@ -72,3 +72,26 @@ precmd_functions+=(set_win_title)
...
+
+Style strings are a list of words, separated by whitespace. The words are not case sensitive (i.e. `bold` and `BoLd` are considered the same string). Each word can be one of the following:
+
+  - `bold`
+  - `underline`
+  - `bg:<color>`
+  - `fg:<color>`
+  - `<color>`
+
+where `<color>` is a color specifier (discussed below). The last two are currently considered equivalent, though this may change in the future. The order of words in the string does not matter.
```
matchai: To help avoid a mistake where we add an option to the end of the list, I think it's best if we reword this.

```suggestion
where `<color>` is a color specifier (discussed below). Both `fg:<color>` and `<color>` are currently considered equivalent, though this may change in the future. The order of words in the string does not matter. 
```

chipbuster:
@matchai Currently the way to do no styling is to set the style string to `""`. If you'd like I can add the `none` token to the list (I think I should add `dimmed` anyways, as it's used by the hostname module).

docs/config/README.md
```diff
@@ -40,6 +40,18 @@ are segments within it. Every module also has a prefix and suffix that are the d
...
 
+### Style Strings
+
+Most modules in starship allow you to configure their display styles. This is done with an entry (usually called `style`) which is a string specifying the configuration. Here are some examples of style strings along with what they do. For details on the full syntax, consult the [advanced config guide](/advanced-config/).
+
+- `"fg:green bg:blue"` sets green text on a blue background
+- `"bg:blue fg:bright-green"` sets bright green text on a blue background
+- `"bold fg:27"` sets bold text with [ANSI color](https://i.stack.imgur.com/KTSQa.png) 27
+- `"underline bg:#bf5700"` sets underlined text on a burnt orange background
+- `""` explicitly disables all styling
```
matchai: Thoughts on using `null` or `"none"` here?

```suggestion
- `null` explicitly disables all styling
```

docs/config/README.md
```diff
@@ -40,6 +40,18 @@ are segments within it. Every module also has a prefix and suffix that are the d
...
+
+Note that what styling looks like will be controlled by your terminal emulator. For example, some terminal emulators will brighten the colors instead of bolding text, and some color themes use the same values for the normal and bright colors.
+
 ## Prompt
```
matchai: Could be addressed in a separate PR, but we should probably have style configuration for the prompt elements that aren't part of modules (e.g. prefix, suffix).

src/config.rs
```diff
@@ -157,11 +160,130 @@ impl Config for Table {
...
+ There are three valid color formats:
+  - #RRGGBB      (a hash followed by an RGB hex)
+  - u8           (a number from 0-255, representing an ANSI color)
+  - colstring    (one of the 16 predefined color strings)
+*/
+fn parse_color_string(color_string: &str) -> Option<ansi_term::Color> {
+    // Parse RGB hex values
+    log::trace!("Parsing color_string: {}", color_string);
+    if color_string.starts_with('#') {
+        log::trace!(
```
matchai: Nice logging! 👍

src/config.rs
```diff
@@ -157,11 +160,130 @@ impl Config for Table {
...
+            token.trim_start_matches("fg:").to_owned()
+        } else if token.as_str().starts_with("bg:") {
+            col_fg = false;
+            token.trim_start_matches("bg:").to_owned()
+        } else {
+            col_fg = true; // Bare colors are assumed to color the foreground
+            token
+        };
+
+        match token.as_str() {
```
matchai: Looks good 👍

src/config.rs
```diff
@@ -210,4 +332,69 @@ mod tests {
...
+        table.insert(
+            String::from("plainstyle"),
+            toml::value::Value::String(String::from("")),
+        );
+        assert_eq!(
+            table.get_as_ansi_style("plainstyle").unwrap(),
+            ansi_term::Style::new()
+        );
+
+        // Test a string that's clearly broken
```
matchai: Good tests 👍

src/modules/username.rs
```diff
@@ -37,13 +38,13 @@ fn get_uid() -> Option<u32> {
...
 
-fn is_root(style: &mut Style) -> bool {
-    match get_uid() {
-        Some(uid) if uid == 0 => {
-            style.clone_from(&Color::Red.bold());
-
-            true
-        }
-        _ => false,
+fn get_mod_style(user_uid: Option<u32>, module: &Module) -> Style {
```
matchai: 👍

docs/advanced-config/README.md
chipbuster: Agreed.

docs/config/README.md
chipbuster: Let's do a separate PR. I'll open it once this gets merged.

docs/config/README.md
chipbuster: I've put in a new token `none` (as well as a new token `dimmed`).

PR #286, opened 2019-09-04
feat: add host module

### LINKED ISSUE #157: Add hostname module

When I was only using starship on one machine, it was pretty easy to keep track of where I was.

Now that I have it across 4 or 5, I find myself getting confused a little too often as to which host I'm actually on.

A module like `host` from spaceship, which prints out the hostname of the system.

Reference implementation: https://github.com/denysdovhan/spaceship-prompt/blob/master/sections/host.zsh

Nothing yet.

ahouts:
#### Description

Add a host module as requested by @chipbuster.
Displays the system hostname as provided by [gethostname](https://crates.io/crates/gethostname).

#### Motivation and Context

Closes #157

#### Types of changes

image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -277,7 +277,7 @@ The `hostname` module shows the system hostname.
 
 | Variable     | Default | Description                                             |
 | ------------ | ------- | ------------------------------------------------------- |
-| `ssh_only`   | `true`  | Only show hostname when connected to a ssh session.     |
+| `ssh_only`   | `true`  | Only show hostname when connected to an SSH session.     |
 | `prefix`     | `""`    | Prefix to display immediately before the hostname.      |
 | `suffix`     | `""`    | Suffix to display immediately after the hostname.       |
 | `disabled`   | `false` | Disables the `hostname` module.                         |
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -69,6 +69,7 @@ The `default_prompt_order` configuration option is used to define the order in w
 ```
 default_prompt_order = [
     "username",
+    "hostname",
     "directory",
     "git_branch",
     "git_status",
@@ -266,6 +267,31 @@ modified = "📝"
 staged = "➕"
 renamed = "👅"
 deleted = "🗑"
+```
+
+## Hostname
+
+The `hostname` module shows the system hostname.
+
+### Options
+
+| Variable     | Default | Description                                             |
+| ------------ | ------- | ------------------------------------------------------- |
+| `ssh_only`   | `true`  | Only show hostname when connected to a ssh session.     |
+| `prefix`     | `""`    | Prefix to display immediately before the hostname.      |
+| `suffix`     | `""`    | Suffix to display immediately after the hostname.       |
+| `disabled`   | `false` | Disables the `hostname` module.                         |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[hostname]
+ssh_only = false
+prefix = "⟪"
+suffix = "⟫"
+disabled = false
 ```
 
 ## Golang
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -277,7 +277,7 @@ The `hostname` module shows the system hostname.
 
 | Variable     | Default | Description                                             |
 | ------------ | ------- | ------------------------------------------------------- |
-| `ssh_only`   | `true`  | Only show hostname when connected to an SSH session.     |
+| `ssh_only`   | `true`  | Only show hostname when connected to an SSH session.    |
 | `prefix`     | `""`    | Prefix to display immediately before the hostname.      |
 | `suffix`     | `""`    | Suffix to display immediately after the hostname.       |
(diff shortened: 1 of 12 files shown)


tests/testsuite/host.rs
```diff
@@ -0,0 +1,107 @@
+use ansi_term::{Color, Style};
+use std::io;
+
+use crate::common;
+use crate::common::TestCommand;
+
+// TODO: test where hostname contains invalid UTF, triggering invalid hostname pattern
+
+#[test]
```
chipbuster: Just to make sure I'm understanding correctly: it looks like these could all be run as unit tests from within the module, if it weren't for the fact that you want to test things with the actual hostname/configs? (Excellent suite of tests, by the way)

Is there any way at all that these tests could fail if run in a production environment? Non-ignored tests failing will stop certain installation methods, so I'd like to be extra-sure. Looking over them once, my gut feeling is "no", but I might have missed something.

src/modules/host.rs
```diff
@@ -0,0 +1,40 @@
...
+    let ssh_connection = env::var("SSH_CONNECTION").ok();
+    if module.config_value_bool("ssh_only").unwrap_or(true) && ssh_connection.is_none() {
+        return None;
+    }
+
+    let os_hostname: OsString = gethostname::gethostname();
+
+    let (host, style) = match os_hostname.into_string() {
+        Ok(host) => (host, Color::Green.bold().dimmed()),
+        Err(_) => (String::from("<invalid UTF>"), Color::Red.bold()),
```
chipbuster: We generally don't display error conditions directly in the prompt. 

The preferred technique (that I've seen so far) is to make the module `None` in the event of an error, and then use `log::debug()` to emit a logging message that can be seen if the user runs with the right envars. See, for example, [battery.rs:23](https://github.com/starship/starship/blob/08aef016cdb1dbf53fc50b6e31ab0add2f40856b/src/modules/battery.rs#L23)

src/modules/host.rs
```diff
@@ -0,0 +1,40 @@
...
+
+/// Creates a module with the system hostname
+///
+/// Will display the hostname if all of the following criteria are met:
+///     - host.disabled is absent or false
+///     - host.ssh_only is false OR the user is currently connected as an SSH session (`$SSH_CONNECTION`)
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("host")?;
+
+    if module.config_value_bool("disabled").unwrap_or(false) {
```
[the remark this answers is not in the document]
chipbuster: ...we should probably document that better, because it seems everyone who writes a module (myself included) does that.

docs/config/README.md
```diff
@@ -69,6 +69,7 @@ The `default_prompt_order` configuration option is used to define the order in w
 ```
 default_prompt_order = [
     "username",
+    "host",
```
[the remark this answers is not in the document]
ahouts: Whoops, that's what I named it originally. Guess I'll revert that commit.

tests/testsuite/host.rs
ahouts: Yes they could be run there, I just assumed the testsuite directory was the correct place to put these.
I do not believe they could fail unless `gethostname` panics or returns invalid UTF. I'm changing the tests to not fail if `gethostname` returns invalid UTF.

src/modules/host.rs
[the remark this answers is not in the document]
ahouts: Fair enough, fixed.

ahouts:
@chipbuster changes from the review are pushed

docs/config/README.md
```diff
@@ -268,6 +269,31 @@ renamed = "👅"
...
 
+## Hostname
+
+The `hostname` module shows the system hostname.
+
+### Options
+
+| Variable     | Default | Description                                             |
+| ------------ | ------- | ------------------------------------------------------- |
+| `ssh_only`   | `true`  | Only show hostname when connected to a ssh session.     |
```
matchai: Just a little nitpick:
```suggestion
| `ssh_only`   | `true`  | Only show hostname when connected to an SSH session.     |
```

src/modules/hostname.rs
```diff
@@ -0,0 +1,39 @@
+use ansi_term::{Color, Style};
+use std::env;
+use std::process::Command;
+
+use super::{Context, Module};
+use std::ffi::OsString;
+
+/// Creates a module with the system hostname
+///
+/// Will display the hostname if all of the following criteria are met:
```
matchai: Nice documenting. 👍

ahouts: Thanks, tried to match the style from other modules.

tests/testsuite/host.rs
chipbuster: No, this is the correct place for them. I was just curious as to whether I was missing anything.

matchai:
Thank you for your contribution, Andrew! 😄 

@all-contributors Please add @ahouts for code, docs and test!

bbigras:
Thanks Andrew. I can't wait to use this.

PR #298, opened 2019-09-06
docs: install latest version using brew

hackerzgz:
#### Description

Attempt to update brew before install startship to avoid occur unexpected error in the older version.

#### Motivation and Context

Without running `brew update` I installed v0.7.0. After restart iTerm  I got an error message like:

**error: Found argument '0' which wasn't expected, or isn't valid in this context**

#### Types of changes

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ The prompt shows information you need while you're working, while staying sleek 
    #### Homebrew
 
    ```sh
-   $ brew install starship
+   $ brew update && brew install starship
    ```
 
    #### Rust (v1.33 or higher)


chipbuster:
Hmmm...

I do appreciate the sentiment here.

I'm not sure if we want to be telling users how to administer their own systems--let's say that (entirely stupid idea) someone is intentionally avoiding running a brew update to avoid some bug on their system. Or something like that--there's a lot of reasons someone might not want to run update their system, and starship *should* normally work fine under those conditions (the error you hit is a known bug from back in the 0.7.0 release).

@starship/astronauts What do y'all think?

wyze:
I don't want to say all, just in case I am not remembering correctly. But most installation instructions I see referencing Homebrew just have the install command and don't mention anything about updating first.

matchai:
I'm going to have to agree. Brew auto-updates unless updating is manually disabled. Developers that disable Brew's auto-update feature should be aware that the packages they install might be out-of-date.

Closing this for now. We can reopen this PR if we see repeated cases of bugs caused by out-of-date Brew installations. 🙂

PR #306, opened 2019-09-07
fix: Lazy load git repo and only run module if not disabled

### LINKED ISSUE #283: Let's not touch the dir if we none of the modules uses it
On one of the machines I majorly navigate a remote-mounted filesystem, that takes 1 second to "ls" or "stat". The machine contains no source files, so programming language modules are useless. I'd like to use "starship" for command time measurment and pleasant L&F, I've got used to it.  To avoid an annoying pause after each command, I setup

```toml
[directory]
disabled = true

[python]
disabled = true

[git_branch]
disabled = true

[git_status]
disabled = true

[golang]
disabled = true

[ruby]
disabled = true

[package]
disabled = true

[rust]
disabled = true

[nodejs]
disabled = true

[battery]
disabled = true

[nix_shell]
disabled = true
```

But each time I run a command the prompt still attempts to touch the filesystem. Tried look at the source code tells me that modules look individually to the directory. But I have disabled them all, so there should be no `stat`s.

My suspicion is this https://github.com/starship/starship/blob/dfe2ae643af77b882ae1cfbe81fa74d62250eb93/src/context.rs#L70

as `strace` shows "stat" calls on `/dir/.git`, `/dir/HEAD`, `/dir/commondir`. 

Let's avoid touching the dir if modules are off. I don't know if this is avoidable or not, please tell me if I miss something.

I tried the alternative to disable all modules, but it does not work.

wyze:
#### Description
A couple optimizations are done in this PR. One, we now will check config ahead of time to see if a module is disabled before running any module code. Also, we won't try to discover a git repository unless the module requests access to it.

#### Motivation and Context
Closes #283

#### Types of changes

#### Checklist:


wyze:
A few of the tests got deleted because they were no longer relevant with this change.

src/modules/git_state.rs
```diff
@@ -11,7 +11,8 @@ use super::{Context, Module};
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("git_state")?;
 
-    let repo_root = context.repo_root.as_ref()?;
+    let repo = context.get_repo().ok()?;
+    let repo_root = repo.root.as_ref()?;
```
nickwb: At the moment `git_state` re-opens the repository, but perhaps you could capture `Repository.state()` on to your Repo struct...

If necessary I can do this in a subsequent PR.

src/context.rs
```diff
@@ -24,13 +25,8 @@ pub struct Context<'a> {
     /// The map of arguments that were passed when starship was called.
     pub arguments: ArgMatches<'a>,
 
-    /// If `current_dir` is a git repository or is contained within one,
```
nickwb: Perhaps this is a separate PR, but should we use a similar pattern for `dir_files`?

src/context.rs
```diff
@@ -96,19 +83,22 @@ impl<'a> Context<'a> {
...
+    /// Can always return a new module as if a module was disabled it won't get
+    /// to this point.
     pub fn new_module(&self, name: &str) -> Option<Module> {
         let config = self.config.get_module_config(name);
 
+        Some(Module::new(name, config))
+    }
+
+    /// Check the `disabled` configuration of the module
+    pub fn is_module_enabled(&self, name: &str) -> bool {
```
nickwb: Yeah, makes sense to evaluate this at the same point in time for each module.

src/context.rs
```diff
@@ -96,19 +83,22 @@ impl<'a> Context<'a> {
 
     /// Create a new module
     ///
-    /// Will return `None` if the module is disabled by configuration, by setting
-    /// the `disabled` key to `true` in the configuration for that module.
+    /// Can always return a new module as if a module was disabled it won't get
+    /// to this point.
     pub fn new_module(&self, name: &str) -> Option<Module> {
```
nickwb: I suppose this could return `Module` rather than `Option<Module>` now?

src/modules/git_state.rs
wyze: I can definitely add that.

src/context.rs
wyze: I think I avoided this initially because I would have to touch every module, but maybe we should do that anyways.

matchai: Modules still do return `None` when there's nothing relevant to show. For example, the Node.js module when there are no relevant files.

wyze: This specific function can return just `Module`. Each `Module::module` function is still returning Option<Module>`, so we don't lose the functionality. We save a character in each module by removing the `?` from the end of `let module = context.new_module("");` call.

src/context.rs
wyze: I think it would be a pretty big change, so we should do it in a separate PR.

nickwb:
Had a peek your new changes to this PR. Looks great!

src/context.rs
nickwb: Unless you have plans for it, I might try to tackle this in the next few days.

src/context.rs
```diff
@@ -96,19 +83,22 @@ impl<'a> Context<'a> {
 
     /// Create a new module
     ///
-    /// Will return `None` if the module is disabled by configuration, by setting
-    /// the `disabled` key to `true` in the configuration for that module.
-    pub fn new_module(&self, name: &str) -> Option<Module> {
+    /// Can always return a new module as if a module was disabled it won't get
```
matchai: Nitpick: I would leave `///` doc comments above functions for usage information, rather than contextual information. I think we can assume that a new module can always be returned from the function signature.

src/modules/character.rs
```diff
@@ -20,7 +20,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     const ASSUMED_MODE: ShellEditMode = ShellEditMode::Insert;
     // TODO: extend config to more modes
 
-    let mut module = context.new_module("character")?;
+    let mut module = context.new_module("character");
```
matchai: Happy to see these go away. 👍

src/modules/git_state.rs
```diff
@@ -79,8 +80,8 @@ static AM_OR_REBASE_LABEL: StateLabel = StateLabel {
     message_default: "AM/REBASE",
 };
 
-fn get_state_description(repository: &mut Repository) -> StateDescription {
-    match repository.state() {
+fn get_state_description(state: RepositoryState, root: &PathBuf) -> StateDescription {
```
matchai: Could we please have a doc comment for this function?

src/print.rs
```diff
@@ -81,6 +81,7 @@ pub fn prompt(args: ArgMatches) {
 
     let modules = &prompt_order
         .par_iter()
+        .filter(|module| context.is_module_enabled(module))
```
matchai: 👍

matchai:
Thanks again for writing up this PR! 😄

wyze:
No problem! It's been fun contributing so far. :)

PR #307, opened 2019-09-07
fix: Fixture repo will now clone and set git config locally

### LINKED ISSUE #300: Tests overwrite global git user.name and user.email
#### Current Behavior
~/.gitconfig gets edited
See dfade6d6298b52e04dc8622c676ac8174a900587

#### Expected Behavior
Don't edit anything in the HOME

#### Environment
starship master

#### Suggested fix
Use a temporary home for tests, maybe

wyze:
#### Description
The common function in the test suite will now clone the repo into the fixture directory, and then set the git config locally to that repo.

#### Motivation and Context
Closes #300

#### Types of changes

#### Checklist:

### DIFF
--- a/tests/testsuite/common.rs
+++ b/tests/testsuite/common.rs
@@ -48,15 +48,8 @@ pub fn new_tempdir() -> io::Result<tempfile::TempDir> {
 /// Create a repo from the fixture to be used in git module tests
 pub fn create_fixture_repo() -> io::Result<std::path::PathBuf> {
     let fixture_repo_dir = new_tempdir()?.path().join("fixture");
+    let repo_dir = new_tempdir()?.path().join("rocket");
     let fixture = env::current_dir()?.join("tests/fixtures/rocket.bundle");
-
-    Command::new("git")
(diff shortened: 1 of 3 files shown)


tests/testsuite/common.rs
```diff
@@ -68,7 +61,19 @@ pub fn create_fixture_repo() -> io::Result<std::path::PathBuf> {
         ])
         .output()?;
 
-    Ok(fixture_repo_dir)
+    git2::Repository::clone(fixture_repo_dir.to_str().unwrap(), &repo_dir.as_path()).unwrap();
```
nickwb: Fairly minor, but can these `unwraps` potentially be `?` or `match` statements?

chipbuster:
Well...I fat-fingered my keyboard and accidentally punched in a git commit message before I was ready, but I think it still tells you the gist of what happened. Certainly seems like it would be less confusing to leave it in there than it would be to revert and apply the same patch with two more words in the commit message.

Anyways, @all-contributors let's add @nickwb for review!

PR #314, opened 2019-09-08
feat: Add Java Module

jakubclark:
Implemented a basic `Java` module, that shows the `Java` version, when inside a Java/Maven Project.

#### Description

* New module `java` that calls `javac --version`, to get the version of Java
* Simple tests for the formatting and call to `javac`
* Updated the `Dockerfile` to install `Java`

#### Motivation and Context
Just another language that can be supported.

#### Types of changes

MacOS:
java_mac

Linux (WSL):
java_wsl

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -493,6 +493,31 @@ impure_msg = "impure shell"
 pure_msg = "pure shell"
 ```
 
+## Java
+
+The `java` module shows the currently installed version of Java.
+The module will be shown if any of the following conditions are met:
+
+- The current directory contains a `pom.xml` file
+- The current directory contains a file with the `.java` extension
+
+### Options
+
+| Variable   | Default        | Description                                              |
+| ---------- | -------------- | -------------------------------------------------------- |
+| `symbol`   | `"☕ "`         | The symbol used before displaying the version of Java.  |
+| `style`    | `"bold RGB(166, 42, 42)"` | The style for the module.                     |
+| `disabled` | `false`        | Disables the `java` module.                              |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[java]
+symbol = "🌟 "
+```
+
 
 ## NodeJS
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -499,8 +499,8 @@ pure_msg = "pure shell"
 The `java` module shows the currently installed version of Java.
 The module will be shown if any of the following conditions are met:
 
-- The current directory contains a `pom.xml` file
-- The current directory contains a file with the `.java` extension
+- The current directory contains a `pom.xml` or `build.gradle` file
+- The current directory contains a file with the `.java`, `.class` or `.jar` extension
 
 ### Options
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -506,7 +506,7 @@ The module will be shown if any of the following conditions are met:
 | Variable   | Default        | Description                                              |
 | ---------- | -------------- | -------------------------------------------------------- |
 | `symbol`   | `"☕ "`         | The symbol used before displaying the version of Java.  |
-| `style`    | `"bold RGB(166, 42, 42)"` | The style for the module.                     |
+| `style`    | `"bold #a62a2a"`| The style for the module.                               |
 | `disabled` | `false`        | Disables the `java` module.                              |
 
 ### Example
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -93,6 +93,7 @@ prompt_order = [
     "rust",
     "python",
     "golang",
+    "java",
     "nix_shell",
     "cmd_duration",
     "line_break",
@@ -506,7 +507,7 @@ The module will be shown if any of the following conditions are met:
 | Variable   | Default        | Description                                              |
 | ---------- | -------------- | -------------------------------------------------------- |
 | `symbol`   | `"☕ "`         | The symbol used before displaying the version of Java.  |
-| `style`    | `"bold #a62a2a"`| The style for the module.                               |
+| `style`    | `"dimmed red"` | The style for the module.                               |
 | `disabled` | `false`        | Disables the `java` module.                              |
 
 ### Example
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -46,20 +46,15 @@ fn get_java_version() -> Option<String> {
         Err(_) => String::from("java"),
     };
 
-    // Temporary -> For CI
-    eprintln!("java_command={}", java_command);
-
     match Command::new(java_command).arg("-Xinternalversion").output() {
-        Ok(output) => {
-            // Temporary -> For CI
-            let res = String::from_utf8(output.stdout).unwrap();
-            eprintln!("res={}", res);
-            Some(res)
-        }
+        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
         Err(_) => None,
     }
 }
 
+/// Extract the java version from `java_stdout`.
+/// The expected format is similar to: "JRE (1.8.0_222-b10)".
+/// Some Java vendors don't follow this format: "JRE (Zulu 8.40.0.25-CA-linux64)").
 fn format_java_version(java_stdout: String) -> Option<String> {
     let start = java_stdout.find("JRE (")? + "JRE (".len();
     let end = start
@@ -67,7 +62,12 @@ fn format_java_version(java_stdout: String) -> Option<String> {
             '0'..='9' | '.' => false,
             _ => true,
         })?);
-    Some(format!("v{}", &java_stdout[start..end]))
+
+    if start == end {
+        None
+    } else {
+        Some(format!("v{}", &java_stdout[start..end]))
+    }
 }
 
 #[cfg(test)]
@@ -86,5 +86,20 @@ mod tests {
     fn test_format_java_version_oracle() {
         let java_8 = String::from("Java HotSpot(TM) Client VM (25.65-b01) for linux-arm-vfp-hflt JRE (1.8.0_65-b17), built on Oct  6 2015 16:19:04 by \"java_re\" with gcc 4.7.2 20120910 (prerelease)");
         assert_eq!(format_java_version(java_8), Some(String::from("v1.8.0")));
+    }
+
+    #[test]
+    fn test_format_java_version_redhat() {
+        let java_8 = String::from("OpenJDK 64-Bit Server VM (25.222-b10) for linux-amd64 JRE (1.8.0_222-b10), built on Jul 11 2019 20:48:53 by \"root\" with gcc 7.3.1 20180303 (Red Hat 7.3.1-5)");
+        let java_12 = String::from("OpenJDK 64-Bit Server VM (12.0.2+10) for linux-amd64 JRE (12.0.2+10), built on Jul 18 2019 14:41:47 by \"jenkins\" with gcc 7.3.1 20180303 (Red Hat 7.3.1-5)");
+        assert_eq!(format_java_version(java_8), Some(String::from("v1.8.0")));
+        assert_eq!(format_java_version(java_12), Some(String::from("v12.0.2")));
+    }
+
+    #[test]
+    fn test_format_java_version_zulu() {
+        // Not currently supported
+        let java_8 = String::from("OpenJDK 64-Bit Server VM (25.222-b10) for linux-amd64 JRE (Zulu 8.40.0.25-CA-linux64) (1.8.0_222-b10), built on Jul 11 2019 11:36:39 by \"zulu_re\" with gcc 4.4.7 20120313 (Red Hat 4.4.7-3)");
+        assert_eq!(format_java_version(java_8), None);
     }
 }
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -41,8 +41,21 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn get_java_version() -> Option<String> {
-    match Command::new("java").arg("-Xinternalversion").output() {
-        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
+    let java_command = match std::env::var("JAVA_HOME") {
+        Ok(java_home) => format!("{}/bin/java", java_home),
+        Err(_) => String::from("java"),
+    };
+
+    // Temporary -> For CI
+    eprintln!("java_command={}", java_command);
+
+    match Command::new(java_command).arg("-Xinternalversion").output() {
+        Ok(output) => {
+            // Temporary -> For CI
+            let res = String::from_utf8(output.stdout).unwrap();
+            eprintln!("res={}", res);
+            Some(res)
+        }
         Err(_) => None,
     }
 }
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -41,6 +41,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn get_java_version() -> Option<String> {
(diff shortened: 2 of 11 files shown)


src/print.rs
```diff
@@ -24,6 +24,7 @@ const DEFAULT_PROMPT_ORDER: &[&str] = &[
     "rust",
     "python",
     "golang",
+    "java",
```
chipbuster: Looks like you found all the places you need to add new modules to. Well done!

tests/testsuite/java.rs
```diff
@@ -0,0 +1,22 @@
+use std::fs::File;
+use std::io;
+
+use ansi_term::Color;
+
+use crate::common;
+
+#[test]
```
chipbuster: This test should probably be `#ignore`d, since it writes into the filesystem.

src/modules/java.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+        Err(_) => None,
+    }
+}
+
+fn format_java_version(java_stdout: String) -> String {
+    format!("v{}", java_stdout.replace("javac", "").trim())
+}
+
+#[cfg(test)]
```
chipbuster: Love me a unit test.

src/modules/java.rs
```diff
@@ -0,0 +1,71 @@
...
+            Some(module)
+        }
+        None => None,
+    }
+}
+
+fn get_java_version() -> Option<String> {
+    match Command::new("javac").arg("-version").output() {
+        Ok(output) => {
+            if output.stdout.is_empty() {
```
chipbuster: Whoa, do some versions of javac print the version info to stderr?

src/modules/java.rs
```diff
@@ -0,0 +1,71 @@
...
+        return None;
+    }
+
+    match get_java_version() {
+        Some(java_version) => {
+            const JAVA_CHAR: &str = "☕ ";
+            let module_color = Color::RGB(166, 42, 42).bold();
+
+            let mut module = context.new_module("java")?;
+            module.set_style(module_color);
```
chipbuster: Up until about 5 minutes ago, this was the correct way to set the color in modules. Unfortunately, you happened to open this PR right on the back of a pretty big change in how styling is handled in starship.

Fortunately, the change won't be too bad for this module. Take a look at the new e.g. [rust.rs](https://github.com/starship/starship/blob/master/src/modules/rust.rs#L27) for the new syntax: essentially, you ask the module to get its style, and if that fails, fall back to your selected color.

src/modules/java.rs
jakubclark: As far as I can tell, it only prints to stderr, but haven't found any definitive answer for this, which is why I included both.

chipbuster: Alright, good to know! Could you leave a comment here about that so that we know why we're reading from both input streams?

docs/config/README.md
```diff
@@ -463,6 +463,31 @@ impure_msg = "impure shell"
...
+
+- The current directory contains a `pom.xml` file
+- The current directory contains a file with the `.java` extension
+
+### Options
+
+| Variable   | Default        | Description                                              |
+| ---------- | -------------- | -------------------------------------------------------- |
+| `symbol`   | `"☕ "`         | The symbol used before displaying the version of Java.  |
+| `style`    | `"bold RGB(166, 42, 42)"` | The style for the module.                     |
```
chipbuster: The syntax for colors is `#RRGGBB`, so the style would be `bold #a62a2a`

docs/config/README.md
```diff
@@ -463,6 +463,31 @@ impure_msg = "impure shell"
 pure_msg = "pure shell"
 ```
 
+## Java
```
chipbuster: :+1:

jakubclark:
@chipbuster I've addressed your comments. Let me know if there is anything else to change!

docs/config/README.md
matchai: I think built-in modules should probably all use the 16 ANSI colors to remain compatible with as many emulators and configurations as possible.

jakubclark: Sure, I can change it to be red instead.

chipbuster: Didn't think about that. This is probably right. The color appears to be a darker version of the red color, perhaps `dimmed red` would be a good color?

chipbuster:
@jakubclark I'm really sorry about this, I should have done this a little sooner. I just pulled and tested the prompt from your fork, and the addition of the java module tanks performance.

To give you an idea, we can get directory, git branch, git status, package, rust, and python info in 28ms, but adding a single java file raises that to 137.3ms.

image

Do you have any ideas on how we could potentially mitigate this timing problem? I can tell you that it's definitely noticeable--the reason I reached for hyperfine was because I was surprised at how slow the prompt felt.

Maybe it would be possible to parse the information out of `java -version` instead?

With the current speeds I'm seeing, I don't think we can justify putting this module into the prompt :\

jakubclark:
@chipbuster I'll try getting the version from `java -version`.

I just chose `javac` because the output was simpler.

I'll do some performance testing and post the results here.

chipbuster:
@jakubclark Thanks. I think the number we'd love to hit for java module render (via `starship module java`) would be 35-40ms, but 50 might be tolerable. Anything higher than that will most likely be questionable.

jakubclark:
@chipbuster Unfortunately, using `java -version` doesn't help much:

Screenshot 2019-09-08 at 19 47 34

Looks like `java -version` and `javac -version` are simply not fast enough.

chipbuster:
@jakubclark Damn :disappointed: 

Unfortunately, while we're building towards support for several things that could help to mitigate this, like disabled-by-default modules, timeouts, caching data, and (perhaps someday) asynchronous rendering, none of those are quite ready at the moment. I'm afraid we might have to put this PR on hold until one of those comes in.

If you figure out some way to get the module time down without those, let us know and we'll be happy to merge this in!

nickwb:
I had a similar issue when I tried to implement #182 - I'm considering implementing the version detection myself 😅

hdevalke:
> @chipbuster Unfortunately, using `java -version` doesn't help much:
> Looks like `java -version` and `javac -version` are simply not fast enough.

At least on OSX the openjdk jvm installed a release file, I was able to get the version number under 5ms.

```
❯ hyperfine "cat $JAVA_HOME/release | grep JAVA_VERSION | grep -Eo '[0-9]+\\.[0-9]+\\.[0-9]+'"
Benchmark #1: cat /Library/Java/JavaVirtualMachines/jdk1.8.0_211.jdk/Contents/Home/release | grep JAVA_VERSION | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+'
  Time (mean ± σ):       3.2 ms ±   0.1 ms    [User: 2.6 ms, System: 4.2 ms]
  Range (min … max):     3.0 ms …   4.3 ms    566 runs
```

If JAVA_HOME is not set this could be used

```
❯ hyperfine "cat $(/usr/libexec/java_home)/release | grep JAVA_VERSION | grep -Eo '[0-9]+\\.[0-9]+\\.[0-9]+'"
Benchmark #1: cat /Library/Java/JavaVirtualMachines/jdk-12.0.1.jdk/Contents/Home/release | grep JAVA_VERSION | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+'
  Time (mean ± σ):       3.3 ms ±   0.7 ms    [User: 2.7 ms, System: 4.3 ms]
  Range (min … max):     3.0 ms …  16.1 ms    491 runs
```

Not sure for other OSes but caching `java -version` to some file could help to create a generic solution

hdevalke:
Just found out there is an option `-Xinternalversion` this is the output on my mac:

```
❯ hyperfine  'java -Xinternalversion'
Benchmark #1: java -Xinternalversion
  Time (mean ± σ):      24.2 ms ±   7.3 ms    [User: 12.5 ms, System: 8.5 ms]
  Range (min … max):    22.4 ms …  69.8 ms    41 runs
```

On my Linux it is even below 5ms.

I tried to add this functionallity https://github.com/hdevalke/starship/commit/51428ed9cb5dd08e580e5c8498b3eff2ebfdd48e

chipbuster:
@hdevalke I think the `-Xinternalversion` option will be workable, even though parsing the output will be one hell of a task (our main restriction is process launch time, not Rust runtime, so the cost will be code complexity, not runtime).

jakubclark:
@chipbuster @hdevalke I'll try using `java -Xinternalversion` to extract the Java version using regular expressions from inside Rust

jakubclark:
After making the necessary changes, here are the results from my mac:

Screenshot 2019-09-19 at 19 43 19

chipbuster:
@jakubclark 
image

I get no noticeable difference on my Linux machine either. Looks like we got that timing issue nailed! Great work, and thanks to @hdevalke for the hint!

Do you know if this works cross-platform with both OpenJDK/JRE and Oracle Java? I can try to test this out in a docker container later if you don't know.

src/modules/java.rs
```diff
@@ -0,0 +1,77 @@
...
+            module.new_segment("version", &formatted_version);
+
+            Some(module)
+        }
+        None => None,
+    }
+}
+
+fn get_java_version() -> Option<String> {
+    match Command::new("java").arg("-Xinternalversion").output() {
```
hdevalke: Should we take into account the `JAVA_HOME` variable?
If exists  the version could be `$JAVA_HOME/bin/java -Xinternalversion`. 
When using maven it will take priority over the java binary in the `PATH`.

chipbuster: I like this idea. I'm not a fan of relying on JAVA_HOME being set (simply because half the envars that really ought to be set usually aren't), but I think checking for the existence of a JAVA_HOME variable should be more than fast enough.

Do you know if that variable gets set on Windows as well, or is that a UNIX-like-only thing?

chipbuster: Answer to my own question: it seems like (as with many envars) [it's supposed to be set on windows, but a lot of installers fail to do so](https://stackoverflow.com/questions/2619584/how-to-set-java-home-on-windows-7).

jakubclark:
@chipbuster I have Oracle Java installed and it works for me. 

Also, the result of running `java -Xinternalversion` on my raspberry pi (which has Oracle Java version 1.8) is [here](https://github.com/starship/starship/pull/314/files#diff-d4a4c7fefd8315de006a05d6e2ff269bR74).

So I think it's safe to say that works for both OpenJDK and Oracle versions.

chipbuster:
Alright, excellent.

Looks like our CI is failing the standard test suite, though not the containerized tests...do we potentially have a test somewhere that should be `#[ignore]`d that isn't?

src/modules/java.rs
jakubclark: I don't think it's necessary to check `$JAVA_HOME`. Running `$JAVA_HOME/bin/java -Xinternalversion` should be identical to running `java -Xinternalversion`. If `java` isn't on the `PATH` environment variable, then Java simply isn't set up properly.

jakubclark:
For the test suites, I don't know what the issue is. Could be that the result of running `java -Xinternalversion` might be different than expected, but hard to debug the cause. [This is the specific failed test](https://github.com/starship/starship/pull/314/files#diff-0a85556eb8a9df742b134d1785f9be35R10).

src/modules/java.rs
jakubclark: https://www.baeldung.com/maven-java-home-jdk-jre

First checks `$JAVA_HOME`, otherwise defaults to `java` on the `$PATH`.

But I've never had to set up separate Java installations and setting up `$JAVA_HOME` for different projects...

So maybe we should follow the same strategy as Maven:

Check for `$JAVA_HOME`.
Otherwise just use the `java` that is on the `$PATH`.

jakubclark:
@chipbuster any suggestions on how to debug the Test Suites? I tried printing some data using `eprintln!()` but it was written to the logs.

chipbuster:
@jakubclark Unfortunately, the CI is not my specialty.

Looking at the logs, it looks like the Java version simply isn't being detected at all (the actual string for the Java version is just a `v` followed by a color reset code). That suggests to me that the Java version isn't being correctly found.

Perhaps you could try something like 

````rust
let ver = std::command::process("java") //...whatever you need to get the full version string
assert_eq(ver,"nope");
```

This will obviously fail, but it'll let you see what's going on when you try call the java binary to get the version.

hdevalke:
As this fails on Zulu maybe we could check the length of the version to be not zero:

```rust
if end == start {
  None
} else {
  Some(format!("v{}", &java_stdout[start..end]))
}
```

Support for openjdk could be the MVP, other JVM variants could be added later on.

hdevalke:
I tested some other jvm builds:

```sh
for image in {azul/zulu-openjdk,adoptopenjdk,amazoncorretto}; do docker run --rm $image java -Xinternalversion; done
OpenJDK 64-Bit Server VM (25.222-b10) for linux-amd64 JRE (Zulu 8.40.0.25-CA-linux64) (1.8.0_222-b10), built on Jul 11 2019 11:36:39 by "zulu_re" with gcc 4.4.7 20120313 (Red Hat 4.4.7-3)
OpenJDK 64-Bit Server VM (12.0.2+10) for linux-amd64 JRE (12.0.2+10), built on Jul 18 2019 14:41:47 by "jenkins" with gcc 7.3.1 20180303 (Red Hat 7.3.1-5)
OpenJDK 64-Bit Server VM (25.222-b10) for linux-amd64 JRE (1.8.0_222-b10), built on Jul 11 2019 20:48:53 by "root" with gcc 7.3.1 20180303 (Red Hat 7.3.1-5)
```

jakubclark:
I think you're right @hdevalke. It's at least a start.

jakubclark:
Okay, I think this PR is (finally) in a state that it can be merged.

It's able to extract the java version from the `java -Xinternalversion` command, assuming that the output of the command contains something similar to "JRE (12.0.2+10)". It seems that the most common distributions of Java follow this pattern. 

The Zulu distribution that comes with `actions/setup-java` does not follow this pattern, so an integration test wouldn't work in the CI.

[We could include a local copy of the jdkFile](https://github.com/actions/setup-java) from a specific vendor, but I don't think it makes sense to keep an entire JDK in this repo.

Thanks for all your comments and suggestions @chipbuster and @hdevalke !

src/modules/java.rs
```diff
@@ -0,0 +1,105 @@
+use std::process::Command;
+
+use ansi_term::Color;
+
+use super::{Context, Module};
+
+/// Creates a module with the current Java version
+///
+/// Will display the Java version if any of the following criteria are met:
+///     - Current directory contains a file with a `.java` extension
```
chipbuster: Nitpick: aren't these comments inaccurate now, since we're also scanning for jar/class files and `build.gradle`?

docs/config/README.md
```diff
@@ -493,6 +494,31 @@ impure_msg = "impure shell"
...
+
+The `java` module shows the currently installed version of Java.
+The module will be shown if any of the following conditions are met:
+
+- The current directory contains a `pom.xml` file
```
chipbuster: Same as other comment: do we need to update this?

src/modules/java.rs
jakubclark: Fixed

jakubclark:
At least now we know a number of ways to get the Java version 😄

chipbuster:
That we do! It's why I love working on these projects--you're never quite sure what you're going to learn next.

chipbuster:
Woo! What a marathon of an issue.

@all-contributors, please add @jakubclark for code, doc, and tests.

chipbuster:
@all-contributors please also add @hdevalke for their invaluable ideas on this PR.

hdevalke:
> @all-contributors please also add @hdevalke for their invaluable ideas on this PR.

Thank you!

PR #316, opened 2019-09-08
fix: Add display for unknown battery state

heyrict:
#### Description
This PR does the following minor changes to the battery module.

- display battery state when the state is `battery::State::Unknown`
- log the state when the battery state in not handled.

#### Motivation and Context
In my computer, the manufacturer has a setting on battery that prevents it from further charging when it reaches 60% capacity to make the lifespan of the battery longer. In this case, the battery will display `Unknown` state.
I made some modifications as the state is neither logged nor displayed in `starship` here.

#### Types of changes

#### Checklist:


nickwb:
Interesting! Thanks for your contribution. The code looks reasonable to me, but I would ask:

- Is there a better default symbol. Some type of question mark perhaps?
- Should we hide this feature behind a configuration option? I can see this causing some confusion if it's shown during default configuration.

heyrict:
@nickwb I have thought of hiding this feature if no `unknown_symbol` key exist in `[battery]` table, but I got stuck at rust borrowing error, as long as I use `module.new_segment` API.
It seems that I have to add a new API `new_segment_plain` to `Module` which does not read the config for the overwritten value.

heyrict:
> Is there a better default symbol. Some type of question mark perhaps?

Currently, unless a corresponding config of symbol is provided by the user, no segments will be added by the module.

> Should we hide this feature behind a configuration option? I can see this causing some confusion if it's shown during default configuration.

The extra options are hidden under a details-summary block now.

svartalf:
Hey, @heyrict, I'm the author of the [`battery`](https://crates.io/crates/battery), which is used here in `starship` to provide batteries information; I accidentally found your issue and this behavior is quite weird, it would be nice to fix it in `battery`, so everyone could benefit from the fix.

Could you post the output of the following commands when your battery is fully charged, either here if `starship` developers are okay with that or in a [new issue at the `battery` crate repo](https://github.com/svartalf/rust-battery/issues/new)?

This would help a lot:

`cat /sys/class/power_supply/*/uevent`
`cat /sys/class/power_supply/*/status`

heyrict:
@svartalf Thank you for your comments, but I don't think it is something related to `battery` as `Unknown` is what linux kernel shows.

Here is the output:

```bash
/tmp
⇡56% ❯ cat /sys/class/power_supply/*/status
Unknown

/tmp
?59% ❯ cat /sys/class/power_supply/*/uevent
POWER_SUPPLY_NAME=ACAD
POWER_SUPPLY_ONLINE=1
POWER_SUPPLY_NAME=BAT1
POWER_SUPPLY_STATUS=Unknown
POWER_SUPPLY_PRESENT=1
POWER_SUPPLY_TECHNOLOGY=Li-ion
POWER_SUPPLY_CYCLE_COUNT=0
POWER_SUPPLY_VOLTAGE_MIN_DESIGN=7600000
POWER_SUPPLY_VOLTAGE_NOW=7888000
POWER_SUPPLY_POWER_NOW=0
POWER_SUPPLY_ENERGY_FULL_DESIGN=35000000
POWER_SUPPLY_ENERGY_FULL=34070000
POWER_SUPPLY_ENERGY_NOW=20290000
POWER_SUPPLY_CAPACITY=59
POWER_SUPPLY_CAPACITY_LEVEL=Normal
POWER_SUPPLY_MODEL_NAME=003250324D34314C
POWER_SUPPLY_MANUFACTURER=53 4D 50 00 31 31 00 00 12
POWER_SUPPLY_SERIAL_NUMBER=2021
```

heyrict:
@svartalf Thank you for your comments, but I don't think it is something related to `battery` as `Unknown` is what linux kernel shows.

Here is the output:

```bash
/tmp
⇡56% ❯ cat /sys/class/power_supply/*/status
Unknown

/tmp
?59% ❯ cat /sys/class/power_supply/*/uevent
POWER_SUPPLY_NAME=ACAD
POWER_SUPPLY_ONLINE=1
POWER_SUPPLY_NAME=BAT1
POWER_SUPPLY_STATUS=Unknown
POWER_SUPPLY_PRESENT=1
POWER_SUPPLY_TECHNOLOGY=Li-ion
POWER_SUPPLY_CYCLE_COUNT=0
POWER_SUPPLY_VOLTAGE_MIN_DESIGN=7600000
POWER_SUPPLY_VOLTAGE_NOW=7888000
POWER_SUPPLY_POWER_NOW=0
POWER_SUPPLY_ENERGY_FULL_DESIGN=35000000
POWER_SUPPLY_ENERGY_FULL=34070000
POWER_SUPPLY_ENERGY_NOW=20290000
POWER_SUPPLY_CAPACITY=59
POWER_SUPPLY_CAPACITY_LEVEL=Normal
POWER_SUPPLY_MODEL_NAME=003250324D34314C
POWER_SUPPLY_MANUFACTURER=53 4D 50 00 31 31 00 00 12
POWER_SUPPLY_SERIAL_NUMBER=2021
```

svartalf:
@heyrict, yeah, you are right, there is nothing I can do with that. Thanks for a information though!

docs/config/README.md
```diff
@@ -118,6 +118,18 @@ The module is only visible when the device's battery is below 10%.
 | `display`            | [link](#battery-display) | Display threshold and style for the module.       |
 | `disabled`           | `false`                  | Disables the `battery` module.                    |
 
+<details>
```
chipbuster: Didn't know you could do that. Neat.

src/module.rs
```diff
@@ -74,6 +74,20 @@ impl<'a> Module<'a> {
         self.segments.last_mut().unwrap()
     }
 
+    /// Should config exists, get a reference to a newly created segment in the module
```
chipbuster: Sorry, I'm not 100% clear on why this function exists. Is this just creating a segment only if it's in the config file?

chipbuster: I guess clarifying a little more, I'm just confused about the naming and comments for this function. The actual functionality is pretty clear from reading the code.

heyrict: Hmm, sorry I can't find a better naming. Should we name it `new_segment_if_config_exists` or something?

chipbuster: @heyrict I think that would be a good name.

PR #317, opened 2019-09-08
docs: reorder misplaced Ruby option example

weihanglo:
Reorder misplaced Ruby option example

#### Description

Ruby example is misplaced at Linebreak section.

#### Types of changes

Docs updated

Ruby example shouldn't show at linebreak section.

image

Need not.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -427,15 +427,6 @@ disabled = true
 ```
 
 
-### Example
-
-```toml
-# ~/.config/starship.toml
-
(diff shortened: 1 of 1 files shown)


nickwb:
Whoops. Good catch. Thanks @weihanglo

weihanglo:
Maybe I should add some magic word like `[ci skip]` when there are only doc updates.
(or add a diff algorithm to reduce CI workloads)

PR #318, opened 2019-09-08
feat: Implement threshold based styling for battery module

heyrict:
#### Description
Add threshold based styling for battery.

Require changes to `starship.toml`:

- Remove key `style`
- Add key `display` which is an array of tables of
  ```
  threshold: i64 // required
  style: i64 // required
  ```

#### Motivation and Context
Fixes TODO described [here](https://github.com/starship/starship/blob/29eedeb698bb268bfc1659b5e3f0fe2ccef001a8/src/modules/battery.rs#L31)

#### Types of changes

#### Example
A config of
```toml
[battery]
  [[battery.display]]
  threshold = 10
  style = "bold red"

  [[battery.display]]
  threshold = 30
  style = "bold yellow"

  [[battery.display]]  # This is ignored
  threshold = 15
  style = "bold orange"
```
makes the style `bold red` if capacity is below 10, and `bold yellow` if capacity is below 30.
The first table matching `capacity <= threshold` is applied so the third table is ignored.
When capacity is over 30, the battery info will not be displayed.

#### Checklist:

Since I don't know how to write multiple lines in the table, the doc is written in JSON format and should better be fixed.
The doc is not well described though.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -108,13 +108,13 @@ The module is only visible when the device's battery is below 10%.
 
 ### Options
 
-| Variable             | Default      | Description                                       |
-| -------------------- | ------------ | ------------------------------------------------- |
-| `full_symbol`        | `"•"`        | The symbol shown when the battery is full.        |
-| `charging_symbol`    | `"⇡"`        | The symbol shown when the battery is charging.    |
-| `discharging_symbol` | `"⇣"`        | The symbol shown when the battery is discharging. |
-| `style`              | `"bold red"` | The style for the module.                         |
-| `disabled`           | `false`      | Disables the `battery` module.                    |
+| Variable             | Default                                  | Description                                       |
+| -------------------- | ---------------------------------------- | ------------------------------------------------- |
+| `full_symbol`        | `"•"`                                    | The symbol shown when the battery is full.        |
+| `charging_symbol`    | `"⇡"`                                    | The symbol shown when the battery is charging.    |
+| `discharging_symbol` | `"⇣"`                                    | The symbol shown when the battery is discharging. |
+| `display`            | `{"threshold": 10, "style": "bold red"}` | Display threshold and style for the module.       |
+| `disabled`           | `false`                                  | Disables the `battery` module.                    |
 
 ### Example
 
@@ -125,6 +125,14 @@ The module is only visible when the device's battery is below 10%.
 full_symbol = "🔋"
 charging_symbol = "⚡️"
 discharging_symbol = "💀"
+
+  [[battery.display]]
+  threshold = 10
+  style = "bold red"
+
+  [[battery.display]]
+  threshold = 30
+  style = "bold yellow"
 ```
 
 ## Character
--- a/src/modules/battery.rs
+++ b/src/modules/battery.rs
@@ -7,7 +7,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     const BATTERY_FULL: &str = "•";
     const BATTERY_CHARGING: &str = "⇡";
     const BATTERY_DISCHARGING: &str = "⇣";
-    const BATTERY_THRESHOLD: f32 = 10.0;
+    const BATTERY_THRESHOLD: i64 = 10;
     // TODO: Update when v1.0 printing refactor is implemented to only
     // print escapes in a prompt context.
     let shell = std::env::var("STARSHIP_SHELL").unwrap_or_default();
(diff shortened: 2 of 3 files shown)


heyrict:
@johnletey Well, I didn't get what you mean quite well. Could you give an example?

johnletey:
@heyrict So instead of having

```toml
[battery]
  [[battery.display]]
  threshold = 15
  style = "bold red"
```

You should have

```toml
[battery]
threshold = 15
style = "bold red"
```

Which matches the configuration style of the other prompt modules

heyrict:
Well, my first implementation ([The first commit](https://github.com/starship/starship/pull/318/commits/3bd7efb3af0c2ef27098fb3de0cffab3027aacf8)) uses that symtax, but I found it is not what the contributor who left the comment mean.

With an ungrouped syntax, the style will not *be set based on percentage* when threshold is modifiable.

Say, you may want to display battery capacity constantly and have a configuration file as below:

```toml
[battery]
threshold = 100
style = "bold green"
```

But the indicator will always be `bold green` in this case and cannot be changed to `bold red` for example if the capacity is low.

johnletey:
@heyrict I see, thank you for your explanation! Leave it as is then! Wonderful work!

docs/config/README.md
```diff
@@ -125,6 +125,36 @@ The module is only visible when the device's battery is below 10%.
...
+  [[battery.display]]
+  threshold = 10
+  style = "bold red"
+
+  [[battery.display]]
+  threshold = 30
+  style = "bold yellow"
+```
+
+### Battery Display
```
matchai: Thank you for adding this section to the docs. 👍 
Much clearer now that this is here! 😄

matchai: What are your thoughts around a sub-heading to "Battery Display" for "Options", with a similar table to option tables for modules?

It would be nice to specify that `threshold` acts as an upper bound for the `battery.display`.

src/modules/battery.rs
```diff
@@ -19,43 +19,79 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+            if let toml::Value::Table(config) = display_config {
+                if let Some(display_style) = BatteryDisplayStyle::from_config(config) {
+                    display_styles.push(display_style);
+                }
+            }
+        }
+
+        // Return display styles as long as display array exists, even if it is empty.
+        display_styles
+    } else {
```
matchai: 👍

src/modules/battery.rs
```diff
@@ -19,43 +19,79 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    let display_style = display_styles.iter().find(|display_style| {
+        let BatteryDisplayStyle { threshold, .. } = display_style;
+        percentage <= *threshold as f32
+    });
+
+    if let Some(display_style) = display_style {
+        let BatteryDisplayStyle { style, .. } = display_style;
+
+        // Set style based on percentage
+        module.set_style(*style);
```
chipbuster: I'm still a little green to Rust, so feel free to point me to the appropriate doc if you know where it is: why do we need to dereference this after destructuring?

src/modules/battery.rs
```diff
@@ -19,43 +19,79 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+
+    // Parse config under `display`
+    let display_styles = get_display_styles(&module);
+    let display_style = display_styles.iter().find(|display_style| {
+        let BatteryDisplayStyle { threshold, .. } = display_style;
+        percentage <= *threshold as f32
+    });
+
+    if let Some(display_style) = display_style {
+        let BatteryDisplayStyle { style, .. } = display_style;
```
chipbuster: Is `style` here the first element of the struct because `threshold` was already destructured out earlier?

src/modules/battery.rs
```diff
@@ -19,43 +19,79 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
 
-    let mut percent_string = Vec::<String>::with_capacity(2);
-    // Round the percentage to a whole number
-    percent_string.push(percentage.round().to_string());
-    percent_string.push(percentage_char.to_string());
-    module.new_segment("percentage", percent_string.join("").as_ref());
+fn get_display_styles(module: &Module) -> Vec<BatteryDisplayStyle> {
+    const DISPLAY_STYLES: &[BatteryDisplayStyle] = &[BatteryDisplayStyle {
+        threshold: 10,
+        style: Style {
```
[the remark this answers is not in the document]
chipbuster: This doesn't work because this is a `const`. Oh dear.

src/modules/battery.rs
nickwb: `iter` returns a reference, and when you destructure a reference you get fields which are themselves references.

src/modules/battery.rs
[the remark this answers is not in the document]
heyrict: Yes it do works, but instead of a constant, we have to use a `let` variable then.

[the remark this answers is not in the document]
chipbuster: I'm a little torn on this. On one hand, we'd like to trim whatever runtime costs we can, and on the other, it's *so* hard to read this thing and figure out what style you're getting from it...

src/modules/battery.rs
chipbuster: Thanks!

docs/config/README.md
heyrict: Done

src/modules/battery.rs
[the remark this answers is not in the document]
heyrict: Should be better now.

chipbuster:
Just as a note for any future code spelunkers: I think the `BatteryDisplayStyle` could be changed into some sort of generic` ThresholodedDisplayStyle`, but that falls squarely under YAGNI at the moment, so we'll probably keep it the way it is.

PR #321, opened 2019-09-08
docs: Include Docker related information in CONTRIBUTING

Hofer-Julian:
#### Description
Includes instruction to set up and use Docker.

#### Motivation and Context
Using Docker is hard.
Additional guidance might be helpful.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,7 +76,10 @@ The previous point should be emphasized: even seemingly innocuous ideas like "if
 
 Acceptance tests are located in the [`tests/`](tests) directory and are also written using the built-in Rust testing library.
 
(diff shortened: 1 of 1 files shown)


chipbuster:
@Hofer-Julian Did you end up deciding against including information on the individual build/run steps?

Hofer-Julian:
I wasn't sure if it still makes sense, when the SELinux problem is resolved in the script and both commands still have to be executed regularly. What do you think?

chipbuster:
> I wasn't sure if it still makes sense, when the SELinux problem is resolved in the script and both commands still have to be executed regularly.

That makes sense. It'll probably be a little more reasonable to add in those changes once we stabilize our modules a bit.

chipbuster:
@all-contributors please add @Hofer-Julian for doc!

PR #332, opened 2019-09-09
fix: Update Module::is_empty to check value of segments

### LINKED ISSUE #325: Remove status wrapper if option is empty [git_status]
## Feature Request
Sorry for the vague description, maybe I'm not using `git stash` correctly but I rarely clean it. So when I'm on what I call a "clean" git status I get the `[$]` which brings me to this feature request.

I'd like to have a way to be able to do something like:
```toml
[git_status]
stashed = ""
```

or

```toml
[git_status]
stashed = false
```

Not use git stash as a dumping ground.

wyze:
#### Description
Change the logic of `Module::is_empty` to use newly created `Segment::is_empty` to determine if the module is empty and should print.

#### Motivation and Context
Closes #325

#### Types of changes

#### Checklist:

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -271,7 +271,7 @@ mod tests {
         let name = "unit_test";
         let module = Module {
             config: None,
-            name: name.to_string(),
+            _name: name.to_string(),
             style: Style::default(),
             prefix: Affix::default_prefix(name),
             segments: Vec::new(),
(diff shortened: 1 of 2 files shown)


chipbuster:
Whoops, looks like we clashed with @nickwb's refactoring changes.

PR #335, opened 2019-09-10
perf: Lazy load files from directory

nickwb:
Currently starship is guaranteed to enumerate all files in the current directory, and it will not start evaluating modules until this operation completes. This PR will change it so that the list is lazily loaded when it is first used, in a similar fashion to the git details in #306 

#### Description
Updates `Context` to use `OnceCell<Vec<PathBuf>>`.

One side effect is `new_scan_dir` is now fallible. I've change the name to `try_begin_scan` and `scan` to `is_match`.

#### Motivation and Context
Allows modules which do not use `ScanDir` to be evaluated sooner, reducing total latency.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -20,7 +20,7 @@ pub struct Context<'a> {
     pub current_dir: PathBuf,
 
     /// A vector containing the full paths of all the files in `current_dir`.
-    pub dir_files: Vec<PathBuf>,
+    dir_files: OnceCell<Vec<PathBuf>>,
 
     /// The map of arguments that were passed when starship was called.
     pub arguments: ArgMatches<'a>,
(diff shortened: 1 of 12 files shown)


chipbuster:
One question I have about all this lazy loading: is there any possibility that it would cause issues with determining timeout if we have one? For example, module 1 tries to load something, takes too long, and times out. Can this (should this?) have any effect on any subsequent modules? What if modules 1 and 2 try to load concurrently?

I suppose I should have asked this back when lazy git loading was implemented, but I wasn't thinking about it at the time.

nickwb:
> is there any possibility that it would cause issues with determining timeout if we have one?

I think the depends a lot on how we implement timeouts. I daresay most of this code will need to become `async` before we can get to that point, at which stage this code will certainly be superseded.

> What if modules 1 and 2 try to load concurrently?

As it stands right now, because we are using `get_or_try_init`, we may attempt to enumerate the directory multiple times if we return an error. That does seem a bit odd for a prompt, so I think we could probably make the argument to store a `OnceCell<Result<Vec<PathBuf>, std::io::Error>`, and use `get_or_init`. We would then be guaranteed to initialise the cell on the first attempt, even if we end up storing an error rather than a list.

nickwb:
I think this PR is ready now. I've also included a handful of other cleanups to make various warnings go away when building/testing starship.

PR #338, opened 2019-09-10
refactor: Separate shell initialization into files

chipbuster:
#### Description

Shell inits are now in a separate directory in the source code, with each shell getting its own script. 

Also adds a little DRY + commenting in init.rs.

#### Motivation and Context

This gives us several advantages:
  - Makes it much easier to navigate init.rs
  - Increased ease of running shellcheck or other linters
  - Line numbers in error messages actually correspond to a line number in a file

These files are then included into init.rs via the [`include_str!`](https://doc.rust-lang.org/std/macro.include_str.html) mechanism, so they are still part of the Rust source code (and importantly, the starship binary)

This is mostly a quality of life change at the moment, but may become increasingly necessary as we move into tighter integration with some shells and more shells overall.

#### Types of changes

Refactor 

#### Checklist:


matchai:
For organization, would you mind maybe moving all the init scripts into `./init`, and renaming `init.rs` to `./init/main.rs` or something along those lines?

chipbuster:
> you haven't changed any of the init scripts, right?

No, I added the change from #336 to bash manually (because git wouldn't have detected a conflict otherwise) but aside from that, these were copy-paste jobs.

> For organization, would you mind maybe moving all the init scripts into src/init, and renaming init.rs to src/init/mod.rs?

Not at all. I'm still not 100% clear on how that works though. Is it just that `src/<name>/mod.rs` is treated equivalently to how `src/<name>.rs` is? Would I have to rename anything?

nickwb:
> is treated equivalently to how `src/<name>.rs` is

Correct. `mod.rs` is a bit like `index.js` in CommonJS or TypeScript, etc., if you are familiar with those.

nickwb:
Good change. It's nice to have proper syntax highlighting for those shell scripts. 👍

chipbuster:
@nickwb Thanks! I've actually never used either language before (which is probably related to the fact that I'm not familiar with that convention)

PR #341, opened 2019-09-11
perf: Don't build `git2` with `ssh` support

nickwb:
We don't need to push or pull repositories, so it should be safe to build `git2` with `ssh` disabled.

The benefits are:

- Faster compile times
- Smaller binary size
- Faster startup time

#### Description
I disable the `ssh` feature from the `git2` package.

#### Motivation and Context

- 🏎
-  🚀 
- Simpler = Better

#### Types of changes

- [] I have tested using **Windows**

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -306,8 +306,6 @@ dependencies = [
  "libc 0.2.58 (registry+https://github.com/rust-lang/crates.io-index)",
  "libgit2-sys 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "openssl-probe 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "openssl-sys 0.9.46 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


nickwb:
Hmm, I'll have to do some benchmarking on MacOS tomorrow - this branch actually seems to be 8% slower on windows 😅 

libgit2 may be heavily optimized towards the standard build configuration.

chipbuster:
Additional advantages: don't require the user to have libcrypto/libssl installed in order to run non-MUSL builds.

Granted these are *usually* installed on systems intended for use as desktop/laptops, but we've seen a few build failures that seem to stem from not having SSL libs in place.

nickwb:
Benchmarking on MacOS today, and it's mostly a wash:

```
# master
Benchmark #1: starship prompt
  Time (mean ± σ):     106.2 ms ±   4.2 ms    [User: 25.4 ms, System: 20.8 ms]
  Range (min … max):    95.8 ms … 110.7 ms    100 runs

# no-ssh
Benchmark #1: starship prompt
  Time (mean ± σ):     107.4 ms ±   4.1 ms    [User: 24.8 ms, System: 21.8 ms]
  Range (min … max):    95.9 ms … 114.0 ms    100 runs
```

PR #352, opened 2019-09-12
feat: Add support for logical paths in directory module

chipbuster:
Adds support for displaying logical (as opposed to physical) paths on in the directory module.

#### Description

Adds the ability for starship to display the logical path instead of the physical path. 

#### Motivation and Context

Physical paths are useful because there is usually only one canonical physical path. However, this physical path is sometimes not intuitive for users (e.g. `/storage/volumes/group/200/user/15ae2` as opposed to `~/storage`). The logical path is easier for the user to understand.

This PR adds an option to directory, `use_logical_path`. When it is set to `true` (which is its default value), the directory module attempts to use the `PWD` environment variable as the path instead of the path within context. If any of its lookups fail (e.g. because PWD is not valid unicode) it will fall back to using the physical path.

`PWD` appears to be a widely-accepted standard. For example, GNU Coreutils `pwd` uses this when running under `pwd -L` to try to get the logical path. Nonetheless, this may fail on systems like MacOS, BSD, Termux, and Windows.

Closes #

#### Types of changes

Breaking new feature?

If anyone is trying to parse the output of starship for some reason, this may break their setup because the current setting is to try to do logical paths by default. However, I don't anticipate any breakage in general.

#### Checklist:

- [x] I have updated the tests accordingly. (but they're failing)

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -32,8 +32,23 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .config_value_i64("fish_style_pwd_dir_length")
         .unwrap_or(FISH_STYLE_PWD_DIR_LENGTH);
 
+    // If use_logical_path is set, read from the envar PWD. This is done by many
+    // logical path tools, including coreutils pwd when run as `pwd -L`. If this
+    // fails, log an error and fall back to the path from context.
+    let use_logical_path = module.config_value_bool("use_logical_path").unwrap_or(true);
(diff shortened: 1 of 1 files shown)


chipbuster:
Also, this PR is still horribly broken and incomplete, so don't review it yet (was supposed to be opened as draft)

chipbuster:
Since GH Actions is broken on this PR, I'm going to see if closing this one and reopening a new PR helps.

PR #362, opened 2019-09-13
Make GitLocalize consistent

matchai:
[Check out a review request at GitLocalize](https://gitlocalize.com/repo/3018/ja/review/10430)


matchai:
Unfortunately not. VuePress does not have such a feature built-in.

PR #366, opened 2019-09-13
feat: add support for logical paths in directory

### LINKED ISSUE #270: Directory module shows physical path instead of logical path on mounted network drive.
#### Current Behavior
The directory module shows `diskxxxx/home/username`.

#### Expected Behavior
The directory module shows `~`.

The home directories on our machines are mounted on network drives. The `HOME` bash variable is set to `/home/username` while the fully resolved path (`realpath $HOME`) is `/diskxxxx/home/username`. When I enter home (`cd $HOME`) and execute `pwd` I get `/home/username`. However, the directory prompt shows `diskxxxx/home/username`.

#### Environment
- Starship version: starship 0.13.1
- Shell type: bash
- Shell version: 4.4.12(1)
- Shell plugin manager: None
- Terminal emulator: Terminal
- Operating system: Debian 9

chipbuster:
Adds support for displaying logical (as opposed to physical) paths on in the directory module.

This is a duplicate of #352, which was repeatedly failing due to (what seems like) a bug in GH Actions on clone. Hoping that a new PR works better.
#### Description

Adds the ability for starship to display the logical path instead of the physical path. 

#### Motivation and Context

Physical paths are useful because there is usually only one canonical physical path. However, this physical path is sometimes not intuitive for users (e.g. `/storage/volumes/group/200/user/15ae2` as opposed to `~/storage`). The logical path is easier for the user to understand.

This PR adds an option to directory, `use_logical_path`. When it is set to `true` (which is its default value), the directory module attempts to use the `PWD` environment variable as the path instead of the path within context. If any of its lookups fail (e.g. because PWD is not valid unicode) it will fall back to using the physical path.

`PWD` appears to be a widely-accepted standard. For example, GNU Coreutils `pwd` uses this when running under `pwd -L` to try to get the logical path. Nonetheless, this may fail on systems like MacOS, BSD, Termux, and Windows.

Closes #270 

#### Types of changes

Breaking new feature?

If anyone is trying to parse the output of starship for some reason, this may break their setup because the current setting is to try to do logical paths by default. However, I don't anticipate any breakage in general.

#### Checklist:

- [x] I have updated the tests accordingly. (but they're failing)


src/modules/directory.rs
```diff
@@ -32,8 +32,23 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .config_value_i64("fish_style_pwd_dir_length")
         .unwrap_or(FISH_STYLE_PWD_DIR_LENGTH);
 
+    // If use_logical_path is set, read from the envar PWD. This is done by many
+    // logical path tools, including coreutils pwd when run as `pwd -L`. If this
+    // fails, log an error and fall back to the path from context.
+    let use_logical_path = module.config_value_bool("use_logical_path").unwrap_or(true);
+    let current_dir = if use_logical_path {
```
chipbuster: I had to change from taking a borrow of the context variable to taking a clone. Here's why: when using the logical path, we read `$PWD` from the environment to construct a PathBuf. Since the PathBuf is created within the scope of the match, it's dropped at the closing brace, so there's no way to return a borrow of it without having the reference outlive the data.

There are two alternate solutions to this:
  - Set up `use_logical_path` as a prompt-wide config which controls how the `context` determines the `current_dir`
  - Declare a mutable outside the match block, set its value within the match, then return a reference to it.

nickwb: Here's another option which doesn't require a copy:

```rust
let pwd_dir = std::env::var("PWD").ok();
let current_dir = pwd_dir
    .as_ref()
    .map(|d| Path::new(d))
    .unwrap_or_else(|| context.current_dir.as_ref());
```

chipbuster: Thanks! I had the idea to use something like this while I was developing this thing, but I couldn't quite work out how to get the ownership to work.

docs/config/README.md
```diff
@@ -245,13 +245,14 @@ it would have been `nixpkgs/pkgs`.
...
-| `style`                     | `"bold cyan"` | The style for the module.                                                        |
-| `disabled`                  | `false`       | Disables the `directory` module.                                                 |
+| Variable                    | Default       | Description                                                                              |
+| --------------------------- | ------------- | ---------------------------------------------------------------------------------------- |
+| `truncation_length`         | `3`           | The number of parent folders that the current directory should be truncated to.          |
+| `truncate_to_repo`          | `true`        | Whether or not to truncate to the root of the git repo that you're currently in.         |
+| `fish_style_pwd_dir_length` | `0`           | The number of characters to use when applying fish shell pwd path logic.                 |
+| `use_logical_path`          | `true`        | Displays the logical path provided by the shell (`PWD`) instead of the path from the OS. |
```
matchai: I think the majority of users wouldn't be able to tell the difference, let alone knowing that they want to change this configuration option.

I really like what was being done here: https://github.com/starship/starship/pull/316/files/29f07ff0956e6a086baa3eb5ab067acb36bcb02e#diff-170a3f2c0a87c4f6c25b4d944865f1c8R121-R131
What are your thoughts around adding an "Advanced configuration" table that is hidden within a spoiler for configuration options that wouldn't be immediately visible to the average user?

src/modules/directory.rs
```diff
@@ -32,8 +32,25 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .config_value_i64("fish_style_pwd_dir_length")
         .unwrap_or(FISH_STYLE_PWD_DIR_LENGTH);
 
+    // Using environment PWD is the standard approach for determining logical path
+    let use_logical_path = module.config_value_bool("use_logical_path").unwrap_or(true);
+    let logical_current_dir = if use_logical_path {
```
matchai: Do all shells update `$PWD`?
I think that resorting to hiding the directory if `$PWD` is unavailable could make it difficult for users to fix the problem. What are your thoughts around falling back to the physical path rather than returning `None`?

tests/testsuite/directory.rs
```diff
@@ -418,3 +418,62 @@ fn directory_in_git_repo_truncate_to_repo_true() -> io::Result<()> {
...
+#[ignore]
+fn use_logical_and_physical_paths() -> io::Result<()> {
+    /* This test is a bit of a smoke + mirrors trick because all it shows is that
+    the application is reading the PWD envar correctly (if the shell doesn't
+    correctly set PWD, we're still in trouble). */
+    let tmp_dir = Path::new("/tmp/starship/porthole/viewport");
+    let dir = tmp_dir.join("directory");
+    let sym = tmp_dir.join("symlink_to_directory");
+    fs::create_dir_all(&dir)?;
+    std::os::unix::fs::symlink(&dir, &sym).unwrap();
```
matchai: Could we add conditional compilation to use `std::os::windows::fs::{symlink_file, symlink_dir}` so that we don't have to revisit this test when adding Windows support?

docs/config/README.md
chipbuster: I think that makes sense.

src/modules/directory.rs
chipbuster: I can't guarantee that all shells update PWD, but the GNU Coreutils `pwd` specifically says in its manpage that it works by trying to read the value of the envar `PWD`. I figure if it's good enough for coreutils, it's good enough for us.

tests/testsuite/directory.rs
chipbuster: Can do! I just have no idea how to test it...

src/modules/directory.rs
chipbuster: As far as falling back to physical path goes: that's actually the intention of this code, so I guess I need to add some comments to make that clear. If the logical path is unreadable, we log an error and return None (this is just so that the user will have a clue as to why the behavior has changed).

Further down, `None` values are converted into the physical path (or at least, the path returned by `std::env::current_dir`, which might not actually be the physical path).

The total effect of this is that if `use_logical_path` is false OR `PWD` is unreadable, the prompt will fall back to the physical path.

tests/testsuite/directory.rs
chipbuster: @matchai Done, but for now I'm just going to have to hope that this does what I want on Windows.

heyrict:
LGTM. May also fix #204.

PR #370, opened 2019-09-14
feat: Add module for showing exit codes of shell pipeline

### LINKED ISSUE #193: Option to show job exit status in prompt
When a job fails, starship displays the prompt character in red, but doesn't show me what the exit status was. This is a fairly subtle display, and the exit code can be useful information.

I'd like a `directory` option that adds the status code. My current prompt does this by putting e.g. `[3]` (in red) right before the `>`. I'd like something similar for starship.

My Bash prompt actually goes a step further and shows me the entire pipeline output if any command had a non-zero exit, even if the overall status was success (in that case it's colored grey instead of red). So for the Bash prompt `false | true` will display as `[1 0]` (in grey), `true | false` will display as `[0 1]` (in red) and longer jobs will add more numbers. And if the overall job status doesn't match the final command status (e.g. because I used `!`) it tacks the overall status on the end in parens, so e.g. `! false` becomes `[1 (0)]` (in grey).

The only reason my Fish prompt doesn't do what my Bash prompt does is because Fish doesn't (yet) have a `$pipestatus` variable.

I know this is kind of complicated, but it works really well for me, especially when putting together complex command-lines, and I'd love to see the ability to do this with starship.

---

<details><summary>Bash <code>__prompt_command</code> function that implements this</summary>

```bash
__prompt_command () 
{ 
    local status=$? pipestatus=("${PIPESTATUS[@]}");
    local prompt="";
    prompt+='\[\e[1;32m\]\u\[\e[m\]@\[\e[1;36m\]\h\[\e[m\]:\[\e[1;31m\]';
    prompt+='$(__prompt_path)';
    prompt+='\[\e[0;33m\]';
    prompt+='$(__git_ps1 " (%s)")';
    prompt+='\[\e[m\]';
    local promptstatus=;
    if (( status > 255 )); then
        promptstatus=$status;
    else
        local x;
        for x in "${pipestatus[@]}";
        do
            if (( x )); then
                promptstatus=${pipestatus[*]};
                break;
            fi;
        done;
        if (( status != pipestatus[${#pipestatus[@]}-1] )); then
            promptstatus+="${promptstatus:+ }($status)";
        fi;
    fi;
    if [[ -n $promptstatus ]]; then
        local color=31;
        if (( status == 0 )); then
            color=2;
        fi;
        prompt+=" \\[\\e[${color}m\\][$promptstatus]\\[\\e[m\\]";
    fi;
    prompt+='> ';
    PS1=$prompt
}
```

</details>

### LINKED ISSUE #882: Support for status codes
## Feature Request

Requesting better support for status codes:
- The ability to see the specific (nonzero) exit code
- Different behaviors for "known special" codes (signals, 127, etc)

Add a module that, at the very least, shows the code in (by default) red, when it's (by default) nonzero. 

None; this feature seems pretty straightforward.

nrabulinski:
#### Description
 I implemented `status` module which can show the exit code of the last command or all exit codes of the last pipeline

#### Motivation and Context
Closes #193
Closes #882

#### Types of changes

Default configuration:
image
Symbols for exit codes:
image
No pipeline:
image

#### Checklist:


nrabulinski:
Oh no, I forgot about the quirk @lilyball talked about - when using `! true` the exit code should be `1` even though `pipestatus` returns `0`.
Will fix in the next commit.

nickwb:
I must admit, I'm a little confused by all of the configuration options - hopefully some new documentation could help with that? 🙂 

This is just a thought - do you think there's value in modelling the display mode in an enum?

For example:
```
enum DisplayMode {
    ShowPipeline,              // Always show the pipeline
    ShowPipelineOnError,       // Show the pipeline on error, otherwise nothing
    ShowLastOrPipelineOnError, // Show the pipeline on error, otherwise the last status
    ShowLast,                  // Always show the last status
    ShowLastOnError            // Show the last status on error, otherwise nothing
}
```

nrabulinski:
> I must admit, I'm a little confused by all of the configuration options - hopefully some new documentation could help with that? slightly_smiling_face
> 
> This is just a thought - do you think there's value in modelling the display mode in an enum?
> 
> For example:
> 
> ```
> enum DisplayMode {
>     ShowPipeline,              // Always show the pipeline
>     ShowPipelineOnError,       // Show the pipeline on error, otherwise nothing
>     ShowLastOrPipelineOnError, // Show the pipeline on error, otherwise the last status
>     ShowLast,                  // Always show the last status
>     ShowLastOnError            // Show the last status on error, otherwise nothing
> }
> ```

I've already talked about it in the Discord server, but I don't personally think there's sense to add `ShowLastOrPipelineOnError`. Other than that I think it was a great idea and I knew I had to change all those bools anyway. I don't know however if the way I solved it is the best/most intuitive one out there, so if you have any other suggestions - I'm all ears!

lilyball:
Judging from the name, `ShowLastOrPipelineOnError` matches the behavior I use personally.

nrabulinski:
@lilyball I just don’t see how it’d be different from my current `OnError` other than it’d show `0` each time there’s no error which isn’t very helpful. Although if it is something that people want, I can implement it.

lilyball:
Skimming the code, it looks like what you've actually implemented doesn't do what I thought it did. In fact, it doesn't even implement the behavior of showing an error on `! false`, because that has a pipestatus length of `1` and `errored` is `false`.

nrabulinski:
@lilyball
And it’s only natural that a negated error is considered a success, isn’t it?
As I mentioned earlier - I don’t see the point of adding the mode you’re talking about since the only way it differs from `Always` is that it’d show `0` when there was no error instead of potentially whole pipeline. 

It’d probably be much more viable to iterate through all the status codes in pipeline and show them when at least one of them errored which I might add with the next commit but as it is and considering what (at least I think) you’re talking about - it doesn’t make sense to add.

nkoehring:
I'm new to the code of this project so I don't really know how other modules are doing it, but for the sake of readability I would suggest to move the default values (for example in [line 57](https://github.com/starship/starship/pull/370/files#diff-82610e3591a90c6ea080c9d05534c8cbR57), [line 59](https://github.com/starship/starship/pull/370/files#diff-82610e3591a90c6ea080c9d05534c8cbR59), [line 60](https://github.com/starship/starship/pull/370/files#diff-82610e3591a90c6ea080c9d05534c8cbR60) and others) into constants in (or close to) the beginning of status.rs.

lilyball:
@nrabulinski The whole point of my bash prompt status code setup is to convey the results of the previous job without extraneous info. The results include answering the question "is the status different from the pipestatus?"

So basically, my prompt shows pipestatus, unless it's all zeroes. If pipestatus disagrees with status, it shows that disagreement parenthesized. And it colors the output accordingly, so `! false` shows an output in grey of `[1 (0)]`; the `1` because it's nonzero, and the `(0)` because that's the status diasgreeing with pipestatus. But `! true` shows output in red of `[(1)]`, pipestatus is omitted since it's all zeroes, but status disagrees so it's shown. This same output works for `! true | true` as well, since again pipestatus is all zeroes. But e.g. `! false | false` is (in grey) `[1 1 (0)]`. And `! false | true` is (in red) `[1 0 (1)]`; the pipestatus is shown here because it has a nonzero component.

Ultimately, I'm left with the shortest prompt that doesn't discard any status info. I can tell at a glance what both `$?` and `$PIPESTATUS` are. And uninteresting output is omitted so I don't have to parse it just to find out that it's boring.

chipbuster:
@nrabulinski Do you plan to make more changes to this PR or are we good to review as-is?

nrabulinski:
@chipbuster I'll commit the changes @lilyball suggested in a day or two and then it should be ready to review

matchai:
@nrabulinski Sorry for neglecting this PR. Would you be able to spare some time to get this PR back into a mergable state? 😊

tests/testsuite/status.rs
```diff
@@ -0,0 +1,209 @@
+use ansi_term::Color;
```
matchai: Tests look great! 👍

chipbuster:
Looks like rustfmt failed on....whitespace? Not entirely sure what happened there but hopefully it's easy to fix.

lilyball:
It looks like the empty line has 2 stray spaces in it, which rustfmt wants to delete.

johnletey:
> Looks like rustfmt failed on....whitespace

I am still utterly confused by the fix 😅, but all tests are passing now!

matchai:
@nrabulinski A friendly reminder about those review comments. ☝️

src/configs/status.rs
```diff
@@ -0,0 +1,79 @@
...
+                 //     or there's an error in the pipeline
+}
+
+impl<'a> ModuleConfig<'a> for DisplayMode {
+    fn from_config(config: &toml::Value) -> Option<Self> {
+        Some(match config.as_str()? {
+            "always" => DisplayMode::Always,
+            "error" => DisplayMode::OnExitError,
+            "any error" => DisplayMode::OnError,
+            /*"mismatch" |*/ _ => DisplayMode::OnErrorOrMismatch,
```
Lucretiel: Recommend explicitly matching `"mismatch"` here, and panicking if it doesn't match one of those strings. Worst case, the panic is unreachable otherwise.

src/main.rs
```diff
@@ -25,6 +25,12 @@ fn main() {
         .help("The status code of the previously run command")
         .takes_value(true);
 
+    let pipestatus_arg = Arg::with_name("pipestatus")
+        .long("pipestatus")
+        .value_name("PIPESTATUS")
+        .help("Status codes from the previous pipeline")
```
Lucretiel: Recommend adding a `long_help` here that documents the specific argument we're looking for. Is it a single, space-separated argument? Several individual arguments? Something else? If possible, I'd consider using [Arg::multiple](https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.multiple), which means that the expected format would be: `--pipestatus 0 1 0 2` (as opposed to `--pipestatus "0 1 0 2"`.

nrabulinski: Ah, that’s cool, I’m not really familiar with clap so I wasn’t aware this was a thing, thank you very much!

src/init/starship.bash
```diff
@@ -24,7 +24,7 @@ starship_preexec() {
 # Will be run before the prompt is drawn
 starship_precmd() {
     # Save the status, because commands in this pipeline will change $?
-    STATUS=$?
+    local STATUS=$? PIPE="${PIPESTATUS[@]}";
```
Nemo157: It looks when using `bash-preexec` this will have to be `BP_PIPESTATUS`: https://github.com/rcaloras/bash-preexec/blob/master/bash-preexec.sh#L41-L45

sirMerr:
Closing this in favor of a newer one: https://github.com/starship/starship/pull/1165

PR #380, opened 2019-09-16
feat: Add ability for command duration in milliseconds

### LINKED ISSUE #320: Command duration should report milliseconds for short commands
Commands that take 0-2 seconds should have a finger-grained duration in milliseconds, rather than rounding to seconds. The fish bobthefish theme does this nicely with subtle gray text below a second (if I recall correctly) and then red above. It also switches the units from milliseconds to seconds and then minutes and hours, the same way starship currently goes from seconds to minutes to hours. This gives nice feedback for commands that were just a bit slow, and is precise enough to allow quick informal timing.

Command durations should be in integer milliseconds, or floating-point seconds. The `min_time` config parameter should be parsed as a floating-point number of seconds. Durations below around 2 seconds should be reported in milliseconds (ideally with a different style).

TomHotston:
#### TODOs
- [x] Implement for Bash
- [x] Implement for Fish
- [x] Implement for Zsh

#### Description

This PR aims to implement the millisecond command duration discussed in issue #320 

#### Motivation and Context

Closes #320 

#### Types of changes

```toml
[cmd_duration]
min_time = 1
```
starship changes screenshot

#### Checklist:


docs/config/README.md
```diff
@@ -199,7 +199,7 @@ use_symbol_for_status = true
 ## Command Duration
 
 The `cmd_duration` module shows how long the last command took to execute.
-The module will be shown only if the command took longer than two seconds, or
+The module will be shown only if the command took longer than 2000 milliseconds, or
```
chipbuster: At first I wasn't so sure about the utility of writing this in milliseconds, but after thinking about it, I'll agree that it's much better to retain consistency of units rather than have the intro be written with a common unit (seconds)

docs/config/README.md
```diff
@@ -216,7 +216,7 @@ running `eval $(starship init $0)`, and then proceed as normal.
 
 | Variable   | Default         | Description                         |
 | ---------- | --------------- | ----------------------------------- |
-| `min_time` | `2`             | Shortest duration to show time for. |
+| `min_time` | `2000`          | Shortest duration to show time for. |
```
chipbuster: Just in case someone skips the intro paragraph, can we clarify that this is in milliseconds?

src/modules/cmd_duration.rs
```diff
@@ -45,14 +45,16 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 // Render the time into a nice human-readable string
-fn render_time(raw_seconds: u64) -> String {
+fn render_time(raw_nanoseconds: u64) -> String {
     // Calculate a simple breakdown into days/hours/minutes/seconds
+    let raw_milliseconds = raw_nanoseconds / 1_000_000;
```
chipbuster: Good usage of numeric separators--I think I would have gone slightly insane without them. :+1:

docs/config/README.md
matchai: How about specifying:
> 2000 milliseconds (2 seconds)

src/modules/cmd_duration.rs
TomHotston: Thanks! I am rather new to rust. I didn't actually know these existed until I ran the linter on the code. It's a great feature though!

docs/config/README.md
TomHotston: I have added some clarification in the docs now.

docs/config/README.md
TomHotston: i have amended this as suggested:)

chipbuster:
Code looks good!

So I read over #320, and I just realized that they were asking for milliseconds only below the default threshold (2000 ms, at the moment). OP there wanted the text to be dimmed gray when it's below the threshold, and to just fall back to ms-free timing data when it's above that threshold (which makes sense, since milliseconds aren't terribly useful on a one hour command).

I can also imagine some users would just want to not bother if the command is below that threshold.

Would you mind adding these features? I can also add them by pushing to your fork if you're busy.

TomHotston:
Thanks for reviewing the 

> Code looks good!
> 
> So I read over #320, and I just realized that they were asking for milliseconds only below the default threshold (2000 ms, at the moment). OP there wanted the text to be dimmed gray when it's below the threshold, and to just fall back to ms-free timing data when it's above that threshold (which makes sense, since milliseconds aren't terribly useful on a one hour command).
> 
> I can also imagine some users would just want to not bother if the command is below that threshold.
> 
> Would you mind adding these features? I can also add them by pushing to your fork if you're busy.

That makes sense. I think I have Implemented that!

TomHotston:
> LGTM!

Thanks that's great!

chipbuster:
@all-contributors Please add @TomHotston for code, and doc.

nickwb:
This is broken on Mac OS 10.14.6, it seems to ship with an old version of GNU Date which does not support the `N` token.

TomHotston:
@nickwb that's interesting. Out of curiousity could you run `date --version` to allow me to recreate this on my machine if possible.

Would you like to open an issue to encapsulate elsewhere?

nickwb:
> Out of curiousity could you run date --version to allow me to recreate this on my machine if possible.

This is the first thing I tried to check actually. 😅 It seems whatever shipping with Mac OS does not support `-version`.

```
❯ /bin/date --version
/bin/date: illegal option -- -
usage: date [-jnRu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ...
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
```

It seems like this may not even be GNU date at all:
https://stackoverflow.com/questions/9804966/date-command-does-not-follow-linux-specifications-mac-os-x-lion

TomHotston:
@nickwb that's upsetting. Maybe the remedy is to `install coreutils` then alias over the Mac date command with `gdate` command

Not the friendliest way but it may work.

nickwb:
@TomHotston - Yeah, I've actually tried that. It works, but I don't think it's reasonable to expect starship users to do this.

nickwb:
Maybe we should consider the starship subcommand approach that was discussed in #320

matchai:
We're going to be reverting this merger so we can take the time to address this macOS compatibility issue. Thank you for continuing to work on a solution. 🙏

chipbuster:
@nickwb @TomHotston I'm sorry, I was a little hasty in trying to get this merged.

This has bitten us a few times in the past: the UNIX utilities that ship with MacOS are often the BSD flavor rather than the GNU flavor (see, e.g. the difference in behavior of `sed -i`).

I think our options here are pretty limited: implement a starship timer command, using the program as a wrapper around the system-specific timing calls, or to find some other utility which is available across all platforms.

nomaed:
Maybe the BDS version of the date result with the 'N' at the end can be caught and the N replaced by 000_000_000, so macOS will not have millisecond resolution support, but it will allow supporting it on GNU systems?

PR #383, opened 2019-09-17
refactor: Refactoring config

### LINKED ISSUE #132: Change how default config values are handed in the code
The most common way to handle default values for config options seems to be to attempt to look up the key in the toml, then assign a default value if that fails, something like `config.get_as_type(key).unwrap_or(default_value)`. There are several potential problems I can see with this approach:

- It allows the user to typo keys in their config and not be warned about it (e.g. `use_colour`)
- It allows the *programmer* to typo keys in their code and not be warned about it (granted this should usually be caught on testing/code review)
- If the documentation and code have accidentally diverged, you have to go to the source code of the module to look up the correct default arguments.

A possible alternative would be to create a config manager class that holds `enums` of all the config options, along with their default values. You would then be able to query it by asking it for the value of the enum, e.g. we would replace

```
    let signed_config_min = module.config_value_i64("min_time").unwrap_or(2);
```

with 

```
   let signed_config_min = module.get_config_value(config::min_time);
```

We could even start to encode things like implied arguments (if one value is specified in the config, then multiple other values are implicitly set to certain values), or incompatible arguments across modules (if doing A in one module, cannot do B in another) into this config manager, and give starship the ability to check the toml for correctness.

Advantages:
  - Code that accesses invalid configuration options will not compile
  - Since code now knows about all config options, it can check user's toml file for inconsistencies or invalid config values
  - Simplifies reading in of options in module code

Disadvantages:
  - Large initial pain to switch over
  - Cannot immediately see default value from within module code
  - Potential unseen complexities in implementation

Would this seem like a net positive for us, or does it seem closer to neutral for what we have at the moment?

heyrict:
#### TODOs
- [x] Make macro working
- [x] Add a config example
- [x] Implement `ModuleConfig` trait for scalar (e.g. i64, f64, &str, bool)
- [x] Implement `ModuleConfig` trait for Vec\<T\>
- [x] Implement `ModuleConfig` trait for Option\<T\>
- [x] Deprecate `config.rs`
- [x] Remove `config.rs` and rename `module_config.rs` -> `config.rs`
- [x] Use `toml::Value` instead of `toml::value::Table` in `context.rs` for config.
- [ ] Update configs in `modules/*.rs` (This is time-consuming and can break a lot of things including all PRs here. I will not get this done before the implementation is proven mature.)
- [ ] Update behavior of `fn new_segment` not to override config for segments.

#### Description
This PR refactors config and put configuration files for all modules in `configs/`.
It will break so-many things for developers, while it does not touch the structure of `starship.toml` (yet).

#### Examples
[See module_config::tests, including draft of future config usage](https://github.com/starship/starship/blob/dd98c7222358957e15ddc6c1d55ceec4fd32ebc7/starship/src/config.rs#L291)

#### Changes
master vs PR in `rust` module:

Screenshot from 2019-09-27 19-11-51

where,

```rust
pub struct RustConfig<'a> {
    pub symbol: &'a str,
    pub style: Style,
    pub disabled: bool,
}

impl<'a> RootModuleConfig<'a> for RustConfig<'a> {
    fn new() -> Self {
        RustConfig {
            symbol: "🦀 ",
            style: Color::Red.bold(),
            disabled: false,
        }
    }
}
```

#### Specifications
You need to use a struct for the config with `trait ModuleConfig`, either implement it your self or use the `#[derive(Clone, ModuleConfig)]` macro. `Clone` trait should be implemented too as the `ModuleConfig` macro use `Self.clone()` internally.

All data in the config should have `trait ModuleConfig` implemented. Default data types supported in toml, namely `&str`, `bool`, `i64`, `f64` have `trait ModuleConfig` implemented by default once you `use crate::module_config::ModuleConfig`.

To use an enum, simply implement `trait ModuleConfig` to it. e.g.
```rust
enum Switch {
    on,
    off,
}

impl<'a> ModuleConfig<'a> for Switch {
    fn from_config(config: &toml::Value) -> Option<Self> {
        match (config.as_str()?) {
            "on" => Some(Switch::on),
            "off" => Some(Switch::off),
            _ => None,
        }
    }
}
```

The `trait ModuleConfig` exposes two functions. 
One is `fn from_config(config: &toml::Value) -> Option<Self>`, which constructs `Self` from a toml value. The default behavior of this function is to return `None`.
Another is `fn load_config(&self, config: &'a toml::Value) -> Self`, which overwrites `Self` with `config`. The default behavior of this function is to return `Self::from_config(config).unwrap_or(self)`.

For variables, `from_config` should be re-implemented, and for structs, `load_config` and `from_config` should be re-implemented, by hand or by macros.

PS: If you are familiar with javascript, `load_config` may work like `Object.assign({}, default_config, config)` or `{ ...default_config, ...config }`.

##### About derive macro `ModuleConfig`
Prepending `#[derive(ModuleConfig)]` to automatically get `fn from_config()` and `fn load_config()` implemented, when:

- The struct has `Clone` trait implemented.
- All items in the struct have `ModuleConfig` trait implemented.

The default behavior of `fn from_config` is to return `Some(Self)` only when `item.from_config(config)` is not `None` for all items, otherwise it will return `None`.
`fn load_config` will automatically merge `self` with the config immutably no matter whether all configs exist or not.

#### Discussions
- Should we refactor the folder structure and put `configs/module-name.rs` and `modules/module-name.rs` in the same folder `modules/module-name/`?
- Maybe we can rename the derive macro `ModuleConfig` something more relevant to what it does (implementing `trait ModuleConfig` for structs only)

#### Motivation and Context
Closes #132

#### Checklist:


starship/src/configs/rust.rs
```diff
@@ -0,0 +1,73 @@
...
+            if let Some(config_str) = config.get("style") {
+                if let Some(style) = <Style>::from_config(config_str) {
+                    new_module_config.symbol = style;
+                }
+            }
+        }
+        new_module_config
+    }
+}
+*/
```
heyrict: This is what `#[derive(ModuleConfig)]` will generate for you.

TODO: delete the comment before merging.

heyrict:
The docker CI seems to be broken (probably after the change in folder structure) cc. @matchai 

Current issue is that I cannot `use crate::configs::rust` in other modules as stated in Discord. I'd appreciated if someone can fix it.

starship/src/module_config.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+impl<'a> ModuleConfig<'a> for f64 {
+    fn from_config(config: &toml::Value) -> Option<Self> {
+        config.as_float()
+    }
+}
+
+fn parse_style_string(style_string: &str) -> Option<ansi_term::Style> {
```
heyrict: Copied from `config.rs`.

starship_module_config_derive/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+fn impl_module_config(dinput: DeriveInput) -> TokenStream {
+    let struct_ident = &dinput.ident;
+    let (_impl_generics, ty_generics, where_clause) = dinput.generics.split_for_impl();
+
+    if let syn::Data::Struct(data) = dinput.data {
+        if let syn::Fields::Named(fields_named) = data.fields {
+            let mut tokens = quote! {};
+
+            for field in fields_named.named.iter() {
+                let ident = field.ident.as_ref().unwrap();
```
heyrict: Possible panic point. Can this happen when named fields does not have an identifier? Afaik, all fields in a struct has a name e.g. `symbol: &str`.

heyrict:
Ready for review & merge.

Two modules are updated.
- `rust`: for demo purpose, since no PRs will be affected by this change.
- `battery`: `config_value_array` won't work any more, which is used in this module.

This PR is getting so *huge*. Update of configs of other modules may be put in a separate PR.

chipbuster:
@heyrict Would it be an option to PR into a separate branch instead where we can update the modules one at a time until everything is ready, then PR that branch into master?

heyrict:
@chipbuster Yes, if you'd like to.

Currently this PR will not have any side effects except deprecation warnings and docker CI build fail, so merging it with master is also an option, if someone can get the docker CI build fixed :)

cappyzawa:
I added changes for fixing `Test in Docker` but the test doesn't start... :bow:

heyrict:
> I added changes for fixing `Test in Docker` but the test doesn't start... :bowing_man:

@cappyzawa Thanks for the fix! It seems that tests won't run if this PR has some conflicts with master branch.

cappyzawa:
@heyrict Sorry, the test didn't succeed after all.
But error message is changed. Do you know what is happing?

heyrict:
@cappyzawa Solved by copying the whole project to docker image (which may not be what the creator want). Thank you again for you help!

bijancn:
are these big moves into a `starship` subfolder really intended? Like `src/context.rs → starship/src/context.rs` ?

heyrict:
> are these big moves into a `starship` subfolder really intended? Like `src/context.rs → starship/src/context.rs` ?

Yes. Currently proc_macros have to be implemented in a separate package, and we should add a workspace to make the main package (currently in `starship` subfolder) co-exist with it.

starship/src/configs/rust.rs
bijancn: Do I understand it correctly that it will fail to load the config if any of the required properties are missing?

I really love the idea of automatic config loader derivation but I'm worried that it doesnt cover default values. So either users will have to set all properties even though they might only want to customize one or every property is marked optional with unwrap_or across the module implementations

heyrict: No, `load_config` will merge user's config with `Self`, which is probably the default config, while `from_config` will fail if any of the items is missing.

heyrict: You can refer to [this test](https://github.com/starship/starship/blob/dd98c7222358957e15ddc6c1d55ceec4fd32ebc7/starship/src/config.rs#L385)

bijancn: Uhh. I didn't see that. That's amazing!

heyrict:
@matchai Can you please review d55fc44 and 6400fc5? I just added a minimal segment config implementation for `ModuleConfig` while merging #378.

A new feature is that, we can style all undocumented segments with `segment_name.style`, which I will get documented in the follow-up PRs.

This PR is getting way too big and it is getting hard to maintain. If you think this feature is OK to merge, please get it merged as soon as possible. Should there been any regressions, I will provide some follow-ups to fix it.

Thanks for your review!

matchai:
Everything looks good to me! 👍
Thank you for taking the time to write up this big refactor.
Here goes nothing! 3... 2... 1... MERGE! :shipit:

matchai:
@all-contributors Please add @bijancn for review.

PR #398, opened 2019-09-20
fix: Use logical path instead of physical path when available

### LINKED ISSUE #204: Starship panics if $(pwd) doesn't exist
## Bug Report

Pointed out by user @lilyball on issue #199

#### Current Behavior

When starship is passed a non-existent directory as `--path`, it panics.

This can be triggered from a shell without needing to pass an argument explicitly: open a shell, go to a directory, then open a second shell and delete that directory. Starship will now panic in the first shell until you `cd` out of there.

#### Expected Behavior

Not a panic.

#### Environment
- Starship version: HEAD
- Shell type: all

heyrict:
#### Motivation and Context

Closes #204

#### Types of changes

Screenshot from 2019-09-20 19-17-43

#### Checklist:


heyrict:
Oh, sorry for the transparent background :pray:

matchai:
Is this PR obsolete now that #366 was merged in?

chipbuster:
@matchai It is not. #366 only added logic for how the path is displayed (within the directory module), while the panic occurs in `context.rs`, when the context tries to call `env::current_dir()` to populate the context.

Now there is the interesting problem of what happens here if `PWD` is somehow missing or corrupted, because other modules will start failing left and right, but I think we can deal with that when we get there--that would almost certainly be a shell bug.

heyrict:
This pr will also fix the wrong path when you are in a soft linked directory.

But no module will work if you overwrite PWD. Maybe using output of `pwd -L` instead should be more robust.

chipbuster:
Mm, this is actually going to interfere with #366 in a different way though.

Currently, we read PWD from the environment, and if that fails (or if the user disabled logical paths in their config), we fall back to the context-provided path.

If we have the context doing logical paths by default, then we're going to have to request the physical path inside `directory.rs` and handle any errors in that file.

chipbuster:
@heyrict At least in the GNU coreutils, `pwd -L` just works by reading `PWD`, so I doubt it'll be any more effective.

heyrict:
Ok, I think we should point the user to `.bashrc` if their $PWD got overridden.

heyrict:
> Mm, this is actually going to interfere with #366 in a different way though.

Just found that this PR will always use logical directory no matter whether `use_logical_path` exists in the settings after #366 is merged. (It seems a different manner to fix the logical path issue)

I think it is somewhat OK to drop the physical path option as most people will be confused to see this:

Screenshot from 2019-09-23 19-36-46

, and I can't come up with any use cases of physical path.

chipbuster:
@heyrict I don't personally, but I think we should provide the option. I think the way to do that would be to switch the logic in `directory.rs` so that we try to get the physical path if use_logical_path is false, then fall back to the logical path if that fails.

heyrict:
@chipbuster Got it.

nickwb:
Could we also address #415 in this PR?

I think we need to be a little careful in general, by lifting up `$PWD` to the shared `Context` there may be similar bugs in other modules.

The summary of the issue is, at some places we may be comparing some other absolute/canonical path to the `$PWD` path, and we may become confused if they do not share a common ancestor directory as expected.

heyrict:
> Could we also address #415 in this PR?

Sorry, but I can't reproduce that on my linux machine. Can you provide an example?

> The summary of the issue is, at some places we may be comparing some other absolute/canonical path to the $PWD path, and we may become confused if they do not share a common ancestor directory as expected.

Agreed, since this PR will make context.current_dir unpredictable. You cannot tell whether it is a logical or a physical path.

heyrict:
Currently no module other than directory uses `current_dir`

```bash
starship/src on  fix-unaccessible-directory [?] via 🦀 v1.39.0-nightly took 58s 
?56% ❯ rg -l current_dir
context.rs
modules/directory.rs
```

heyrict:
Hmm, the commit 430f4ad made things a little complex:

The `context` module tries to get logical path first, if not overwritten by `--path`, fallbacks to the physical path and sets `self.current_dir`

The `directory` module gets the config `directory.use_logical_path` and independently load logical or physical path judging from the config, ignoring the `--path` argument and making all tests fail.

chipbuster:
It seems that we have some conflicting goals here. I'm going to go ahead and enumerate them (scanning through the thread from top to bottom) just so we have an idea of what we want to accomplish.

We would like to: 
- Avoid panics on nonexistent or badly-permissioned directory
- Avoid unintentionally clashing with the changes made in #366 
- Avoid introducing bugs due to moving logical paths into `Context` (we're not 100% sure if these exist yet).

Using the logical path in `Context` presents the following challenges:
- If anyone uses `env::current_dir()` or an external tool to get paths relative to the working directory, they may not share any path components.
- Some tools might expect the physical path? (we don't know of any)

We don't know if we want to continue providing support for physical paths, since it seems like people rarely want to use them.

matchai:
>We don't know if we want to continue providing support for physical paths, since it seems like people rarely want to use them.

At this point, I'd certainly be in favor of dropping them completely.

nickwb:
> Sorry, but I can't reproduce that on my linux machine. Can you provide an example?

I was finally able to find a simple way to reproduce this.

If you have a repository which has a canonical path of `/path/to/repo_root`, then use your shell to navigate to that directory with "incorrect" case -
e.g., `cd /path/TO/repo_root`
or `cd /path/to/REPO_root`.

If you reproduce it successfully, you should see that navigation is successful, but `truncate_to_repo` is broken in starship.

matchai:
What's the consensus on this PR?
Would you all be in favor of dropping it to instead use logical paths everywhere?

chipbuster:
My vote is to use logical paths everywhere, but we'll have to keep in mind that we're using logical paths. If some other application decides to use physical paths, we may have to look into using something like `std::fs::canonicalize` to convert our logical path into a physical one.

heyrict:
Just updated `directory` module so that we can still get a physical path if we use logical one as default but have `directory.use_logical_path` set to `true` in the config.

tests/testsuite/directory.rs
```diff
@@ -449,66 +449,3 @@ fn git_repo_in_home_directory_truncate_to_repo_true() -> io::Result<()> {
     assert_eq!(expected, actual);
     Ok(())
 }
-
-#[test]
-#[ignore]
-fn use_logical_and_physical_paths() -> io::Result<()> {
```
heyrict: This test is removed because

- `--path` argument will overwrite the logical path, rather than the physical path. The physical path cannot be tested as it will always be the directory where we run `starship` command.
- Testing logical path only looks meaningless since it just tests if the function reads `$PWD` correctly, as stated in the comment.

PR #403, opened 2019-09-21
feat: add memory usage module

### LINKED ISSUE #392: Add a ram/swap monitor module
I often try to see how much memory/swap I am using and have to run `gotop` or `htop` to check (I don't like the format of `free`). Showing the currently used and maybe available ram/swap as a starship module would be nice.  

A ram usage module like the one in powerlevel9k: https://github.com/Powerlevel9k/powerlevel9k#ram

It has options to show either ram available, swap used, or both. I do think that disabling by default is a good idea for this module, this is also the approach powerlevel9k takes with it.

Of course there are lots of ways to see your ram usage, but I think that having a way to see it within the starship prompt would be neat.

ahouts:
#### Description

Adds a module to display system memory and swap usage.

#### Motivation and Context

Closes #392

#### Types of changes

Screenshot_20190920_223255

Screenshot_20190920_223208

#### Checklist:

Not sure how to meaningfully test this without the ability to mock `SystemExt`. Perhaps `Context` could expose "resources" to modules that could either be a real resource or a fake one defined by the test?


nickwb:
Looks like a great start 👍 

I think you are showing `available/total` rather than `in_use/total`, which would be more obvious based on the "using " prefix you have.

Thanks for including the `show_swap` and `show percentage` options. I think that offers good flexibility. Another option to consider is setting a threshold percentage, so the module would only show if you are above n% in use.

Also, I'd love to see the ram emojii in there: 🐏   😄

ahouts:
@nickwb Doh, you're totally right, fixed.
Also good idea about using the ram, I made that the default.

src/modules/memory_usage.rs
```diff
@@ -0,0 +1,88 @@
...
+    let (display_mem, display_swap) = if show_percentage {
+        (
+            format!("{:.0}%", percent_mem_used),
+            format!("{:.0}%", percent_swap_used),
+        )
+    } else {
+        fn format_kib(n_kib: u64) -> String {
+            let byte = Byte::from_unit(n_kib as f64, ByteUnit::KiB)
+                .unwrap_or_else(|_| Byte::from_bytes(0));
+            byte.get_appropriate_unit(true).format(0).replace(" ", "")
```
nickwb: It's minor, and it might not be worth changing, but you could avoid allocating the string twice here by using `retain` rather than `replace`.

src/modules/memory_usage.rs
```diff
@@ -0,0 +1,88 @@
...
+    } else {
+        module.new_segment("memory_usage", &display_mem);
+    }
+
+    let icon = module
+        .config_value_str("symbol")
+        .unwrap_or(DEFAULT_SYMBOL)
+        .to_string();
+
+    module.get_prefix().set_value(icon);
```
nickwb: I don't see any specific problem with this, but if we are being consistent with other modules, you would probably set the prefix to empty, and include a separate segment for the icon.

If you did make this change, and name the segment `symbol`, you would get the configuration lookup for free.

docs/config/README.md
```diff
@@ -488,6 +489,35 @@ The `line_break` module separates the prompt into two lines.
...
+The `memory_usage` module shows current system memory and swap usage.
+
+By default the swap usage is displayed if the total system swap is non-zero.
+
+### Options
+
+| Variable          | Default               | Description                                                   |
+| ----------------- | --------------------- | ------------------------------------------------------------- |
+| `show_percentage` | `false`               | Display memory usage as a percentage of the available memory. |
+| `show_swap`       | variable              | Display swap usage.                                           |
```
nickwb: Is `when above zero`, better than `variable` here?

src/modules/memory_usage.rs
```diff
@@ -0,0 +1,88 @@
...
+    let total_swap_kib = system.get_total_swap();
+
+    let percent_mem_used = (used_memory_kib as f64 / total_memory_kib as f64) * 100.;
+    let percent_swap_used = (used_swap_kib as f64 / total_swap_kib as f64) * 100.;
+
+    let threshold = module
+        .config_value_i64("threshold")
+        .unwrap_or(DEFAULT_THRESHOLD);
+
+    if percent_mem_used.round() < threshold as f64 {
```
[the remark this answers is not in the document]
ahouts: The threshold value the user can set and the threshold we display are in whole units. One thing to know about `format` is that it rounds to whatever place you specify. As a result, for the value to display only if it is >= the threshold, we need to round just like `format` does.

docs/config/README.md
ahouts: The logic is when total system swap is non-zero, so maybe `when total swap non-zero` works here.

src/modules/memory_usage.rs
ahouts: Didn't notice that icons were handled differently than prefixes usually, fixed.

src/modules/memory_usage.rs
ahouts: Not sure if the user will feel this one, but sure why not 😆

src/print.rs
```diff
@@ -26,6 +26,7 @@ const DEFAULT_PROMPT_ORDER: &[&str] = &[
     "golang",
     "java",
     "nix_shell",
+    "memory_usage",
```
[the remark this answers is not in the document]
ahouts: Good point, default changed to `75`.

ahouts:
@nickwb Thanks for the PR, I fixed or responded to all your comments.

src/modules/memory_usage.rs
```diff
@@ -0,0 +1,87 @@
...
+    const DEFAULT_SHOW_PERCENTAGE: bool = false;
+    const RAM_CHAR: &str = "🐏 ";
+
+    let mut module = context.new_module("memory_usage");
+
+    let module_style = module
+        .config_value_style("style")
+        .unwrap_or_else(|| Color::White.bold().dimmed());
+
+    let system = sysinfo::System::new();
```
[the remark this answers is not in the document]
ahouts: Good to know, updated with that fix.

chipbuster:
Gorgeous :3

PR #409, opened 2019-09-22
feat: Add Environment Variable module

yuri1969:
#### Description
Enables a basic display of a selected environment variable.

#### Motivation and Context
Allows displaying custom information (eg. used by in-house tooling). Inspired by Powerline's `powerline.segments.common.env.environment`.

#### Types of changes

#### Checklist:


src/modules/env_var.rs
```diff
@@ -0,0 +1,49 @@
...
+        .unwrap_or_else(|| Color::Black.bold().dimmed());
+
+    let env_name = module.config_value_str("variable");
+
+    // name is required
+    env_name?;
+
+    let default_value = module.config_value_str("default");
+
+    let env_value = match get_env_value(env_name.unwrap(), default_value) {
```
nickwb: This whole match is equivalent to: `get_env_value(env_name.unwrap(), default_value)?`

src/modules/env_var.rs
```diff
@@ -0,0 +1,49 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("env_var");
+    let module_style = module
+        .config_value_style("style")
+        .unwrap_or_else(|| Color::Black.bold().dimmed());
+
+    let env_name = module.config_value_str("variable");
+
+    // name is required
+    env_name?;
```
nickwb: Is it better to have this at the end of the `let env_name =` line?

i.e.,
`let env_name = module.config_value_str("variable")?;`

That way, `env_name` will be an `&str` rather than `Option<&str>`, and you can drop the `unwrap` from line 25.

src/modules/env_var.rs
yuri1969: Yea, this is way more elegant. I'm a total rookie in Rust, sorry.

src/modules/env_var.rs
yuri1969: Another rookie mistake, sorry.

docs/config/README.md
```diff
@@ -285,6 +286,35 @@ it would have been `nixpkgs/pkgs`.
...
+- The environment variable named `variable` is defined
+- The environment variable named `variable` is not defined and `default` config value exists
+
+### Options
+
+| Variable   | Default          | Description                                                                  |
+| ---------- | ---------------- | ---------------------------------------------------------------------------- |
+| `symbol`   | `""`             | The symbol used before displaying the variable value.                        |
+| `variable` | `""`             | The environment variable to be displayed.                                    |
+| `default`  | `""`             | The default value to be displayed when the selected variable is not defined. |
```
heyrict: `symbol`, `variable`, `default` don't have a default value. Just leave it blank.

yuri1969:
> I'm still considering about which syntax we can use to display multiple env variables, but currently this works great enough.

A full-fledged solution might allow multiple instances of the `env_var` module. This way it would allow defining a different style or symbol for each env variable.

docs/config/README.md
yuri1969: This is fixed now

docs/config/README.md
```diff
@@ -285,6 +286,35 @@ it would have been `nixpkgs/pkgs`.
 truncation_length = 8
 ```
 
+## Environment Variable
+
+The `env_var` module displays the current value of a selected environment variable.
+The module will be shown only if any of the following conditions are met:
+
+- The environment variable named `variable` is defined
```
matchai: I think this may be misleading.
My first impression is that it would be required to have `$VARIABLE` be set.
Maybe instead it should be something along these lines:

```suggestion
- The `variable` configuration option matches an existing environment variable
```

Does this sound clear to folks?

docs/config/README.md
```diff
@@ -285,6 +286,35 @@ it would have been `nixpkgs/pkgs`.
...
+| `style`    | `"dimmed black"` | The style for the module.                                                    |
+| `disabled` | `false`          | Disables the `env_var` module.                                               |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[env_var]
+variable = "SHELL"
```
matchai: Can we throw in default to the example?
```suggestion
variable = "SHELL"
default = "unknown shell"
```

docs/config/README.md
```diff
@@ -285,6 +286,36 @@ it would have been `nixpkgs/pkgs`.
 truncation_length = 8
 ```
 
+## Environment Variable
+
+The `env_var` module displays the current value of a selected environment variable.
+The module will be shown only if any of the following conditions are met:
+
+- The `variable` configuration option matches an existing environment variable
+- The environment variable named `variable` is not defined and `default` config value exists
```
matchai: I think this line should also receive the same treatment:
```suggestion
- The `variable` configuration option is not defined and the `default` configuration option is defined
```

matchai:
Thank you for your contribution! 😄 
@all-contributors Please add @yuri1969 for code, docs, tests. 👏

PR #419, opened 2019-09-24
feat: Add AWS module

### LINKED ISSUE #400: Add AWS profile module
No

In Spacefish, there was an [indicator](https://github.com/matchai/spacefish/blob/master/docs/Options.md#amazon-web-services-aws-aws) that would show me the value of `AWS_PROFILE`. There is no equivalent feature in starship, and I would really appreciate it if there was!

andytom:
#### Description

Have added a module for displaying the current AWS profile based on the
AWS_PROFILE env var.

#### Motivation and Context

Closes #400 

#### Types of changes

#### Checklist:


src/modules/aws.rs
```diff
@@ -0,0 +1,31 @@
+use std::env;
+
+use ansi_term::Color;
+
+use super::{Context, Module};
+
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    if let Ok(aws_profile) = env::var("AWS_PROFILE") {
```
nickwb: It might be neater to drop the if statement and use `let aws_profile = env::var("AWS_PROFILE").ok?;`

andytom:
> Great. Looks pretty simple. 👍
> 
> I have a few questions for some context:
> 
> * Under what conditions does a user usually have this env var set?
> * I assume it is set by the AWS CLI..?
> * Is this env var short-lived, or are AWS CLI users going to see this module at all times?

I can't speak for all users but I have a few [named profiles](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) setup and I will normally set the env var when I am doing some work on a particular account and the either unset the var after I am done or switch to a different account e.g

```bash
export AWS_PROFILE=foo
aws ... 
aws ...
export AWS_PROFILE=bar
# or 
unset AWS_PROFILE
```

chipbuster:
Question: would we be better served by extending functionality in #409? It seems to me right now that we have a bit of code duplication going on in these two PRs.

nickwb:
> Question: would we be better served by extending functionality in #409?

It's a good question. No doubt "show an environment variable" will be at the core of many modules we could add. I think there is value in have plug-and-play modules so users don't have to configure these things by themselves.

That said, we should probably have some common code which reduces a module like this down to just a couple of lines.

chipbuster:
@nickwb In that case, I think I'd be in favor of merging this as-is right now and then refactoring in the future (perhaps adding some common code in util?)

nickwb:
> @nickwb In that case, I think I'd be in favor of merging this as-is right now and then refactoring in the future (perhaps adding some common code in util?)

Yep, that works for me @chipbuster

jeefberkey:
Thank you for jumping on this! I use AWS profiles in the same way you do, and it's comforting to see `dev` in my shell instead of nothing or `production`.

khuongduybui:
Just wants to say thanks. I came here to request a feature only to see it already PR'ed and merged. Sweet!

matchai:
@all-contributors Please add @andytom for code, docs, and tests.

PR #427, opened 2019-09-25
refactor: Refactor config.rs to pure functional style

bijancn:
Hi maintainers 👋 ! This a bit of a test balloon if you are open to refactorings that move the code base towards a more pure functional style, e.g. avoiding mutable variables or using values instead of return statements. I'm an FP Scala developer by day so please excuse any Rust newbie mistakes made.

#### Description
- Replaced for loop that iterates over mutable state with a fold expression
- Unified the logging for the different accessors. The code is now screaming for further refactoring (`get_module_config`, `get_as_bool`, `get_as_str`, `get_as_i64` and `get_as_array` are basically the same up to higher order functions but I didn't manage to get the life times right)
- Increased test coverage (especially a test case for `none` not at the end)
- Removed code comments that literally repeated the code in the next line (see e.g. https://blog.usejournal.com/stop-writing-code-comments-28fef5272752)
- Added `TODO` for the problematic line that swallows the `None` and produces `Some(Style::new())` (this also happened before but in a less obvious way)

#### Motivation and Context
It increases test coverage, makes reasoning about the error cases easier and should pave the path for better error handling in the config in the future.

#### Types of changes
Behaviour should be identical

- [X] I have tested using **Docker**

#### Checklist:

- [X] I have updated the documentation accordingly. (nothing to update)

#### PS
I tried out `tarpaulin` and got a pretty nice coverage xml out of it that I could upload to codecov https://codecov.io/gh/bijancn/starship/src/8ef4843c234c4c026f4ffb26091aa3f1444e686e/src/config.rs . Would be pretty cool to add it to the build but I guess a maintainer has to do it. I've run it locally with
```
docker run --security-opt seccomp=unconfined -v "${PWD}:/volume" xd009642/tarpaulin sh -c "cargo tarpaulin --out Xml "
bash <(curl -s https://codecov.io/bash) -t MY-TOKEN-HERE
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -134,7 +134,8 @@ impl Config for Table {
     /// Get a text key and attempt to interpret it into an ANSI style.
     fn get_as_ansi_style(&self, key: &str) -> Option<ansi_term::Style> {
         // TODO: This should probably not unwrap to an empty new Style but inform the user about the problem
-        self.get_as_str(key).map(|x| parse_style_string(x).unwrap_or(ansi_term::Style::new()))
+        self.get_as_str(key)
+            .map(|x| parse_style_string(x).unwrap_or_default())
     }
 }
 
@@ -146,7 +147,11 @@ fn log_if_key_found(key: &str, something: Option<&Value>) {
     }
 }
 
-fn log_if_type_correct<T: std::fmt::Debug>(key: &str, something: &Value, casted_something: Option<T>) {
+fn log_if_type_correct<T: std::fmt::Debug>(
+    key: &str,
+    something: &Value,
+    casted_something: Option<T>,
+) {
     if let Some(casted) = casted_something {
         log::trace!(
             "Value under key \"{}\" has the expected type. Proceeding with {:?} which was build from {:?}.",
@@ -198,15 +203,13 @@ fn parse_style_string(style_string: &str) -> Option<ansi_term::Style> {
                     "none" => None,
 
                     // Try to see if this token parses as a valid color string
-                    color_string => parse_color_string(color_string).map(
-                        |ansi_color| {
-                            if col_fg {
-                                style.fg(ansi_color)
-                            } else {
-                                style.on(ansi_color)
-                            }
-                        },
-                    ),
+                    color_string => parse_color_string(color_string).map(|ansi_color| {
+                        if col_fg {
+                            style.fg(ansi_color)
+                        } else {
+                            style.on(ansi_color)
+                        }
+                    }),
                 }
             })
         })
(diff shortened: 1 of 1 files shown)


chipbuster:
As someone who took up Rust as a sort of methadone for his Haskell addiction, I'm certainly not opposed to the change.

Of course, I think what's more important than my opinion is the opinion of the astronauts team, so let's get their opinions on this as well.

src/config.rs
```diff
@@ -178,52 +179,40 @@ impl Config for Table {
...
-            // Try to see if this token parses as a valid color string
-            color_string => {
-                // Match found: set either fg or bg color
-                if let Some(ansi_color) = parse_color_string(color_string) {
-                    if col_fg {
-                        style = style.fg(ansi_color);
-                    } else {
-                        style = style.on(ansi_color);
-                    }
+    style_string
```
matchai: Love it! 😄

src/config.rs
```diff
@@ -64,107 +63,109 @@ impl Config for Table {
...
+
+fn log_if_key_found(key: &str, something: Option<&Value>) {
+    if something.is_some() {
+        log::trace!("Value found for \"{}\": {:?}", key, &something);
+    } else {
+        log::trace!("No value found for \"{}\"", key);
+    }
+}
+
+fn log_if_type_correct<T: std::fmt::Debug>(
```
matchai: 👍

src/config.rs
```diff
@@ -277,92 +266,108 @@ fn parse_color_string(color_string: &str) -> Option<ansi_term::Color> {
...
+        table.insert(String::from("integer"), Value::Integer(82));
         assert_eq!(table.get_as_i64("integer"), Some(82));
 
-        // Use with string value
+        table.insert(String::from("string"), Value::String(String::from("82")));
+        assert_eq!(table.get_as_bool("string"), None);
+    }
+
+    #[test]
+    fn table_get_as_array() {
```
matchai: 👍 👍

src/config.rs
```diff
@@ -178,52 +179,40 @@ impl Config for Table {
...
-                // Match found: set either fg or bg color
-                if let Some(ansi_color) = parse_color_string(color_string) {
-                    if col_fg {
-                        style = style.fg(ansi_color);
-                    } else {
-                        style = style.on(ansi_color);
-                    }
+    style_string
+        .split_whitespace()
+        .fold(Some(ansi_term::Style::new()), |maybe_style, token| {
```
chipbuster: Huh, didn't know that the standard library has a fold.

The more you know...

matchai:
@all-contributors Please add @bijancn for code and tests.

bijancn:
Cool! Thank you very much for the warm feedback ❤️

PR #436, opened 2019-09-27
chore: Add rustfmt config file

KokaKiwi:
#### Description
Add a Rustfmt configuration file and format the whole project with it.

#### Motivation and Context
Make the project respect a certain Rust style, this way the code would be consistent.

#### Types of changes
- [x] Add a rustfmt.toml file
- [x] Apply `cargo fmt` to format the project source code

- [x] It just compiles fine and all tests passes.

#### Checklist:
- [ ] Add test on CI for format check?
- [ ] Modify the format convention to another one (I actually used mine here)

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,16 @@
+condense_wildcard_suffixes = true
+edition = "2018"
+format_strings = true
+format_macro_matchers = true
+imports_layout = "HorizontalVertical"
+merge_imports = true
+max_width = 120
+normalize_comments = true
+reorder_impl_items = true
+use_field_init_shorthand = true
+use_try_shorthand = true
+format_code_in_doc_comments = true
+wrap_comments = true
+unstable_features = true
+version = "Two"
+normalize_doc_attributes = true
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,8 +2,7 @@ use crate::utils;
 use std::env;
 
 use dirs::home_dir;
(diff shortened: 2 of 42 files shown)


matchai:
It appeared to me that the codebase was already being formatted consistenly by `rustfmt` without the config file. 

It looks to me like this PR adds an optionated formatting style. Is that the case?

I'd personally prefer sticking to the `rustfmt` defaults so that we can avoid having future discussions and PRs weighing the benefits of possible formatting styling decision (in the same way Prettier does for JavaScript).

KokaKiwi:
> It appeared to me that the codebase was already being formatted consistently by `rustfmt` without the config file.
> 
> It looks to me like this PR adds an alternate optionated formatting style from the `rustfmt` defaults. Is that the case?
> 
> I'd personally prefer sticking to the `rustfmt` defaults so that we can avoid having future discussions and PRs weighing the benefits of possible formatting styling decision (in the same way Prettier does for JavaScript).

Indeed, I think I will close this PR if it really poses problems like that.

chipbuster:
I take quite a few issues with the way that rustfmt formats our code at the moment.

I also appreciate that, when it comes to code formatting, it's more important to be *consistent* than *correct*. 

I'm personally going to come down in favor of sticking with default formatting, especially given that we have a number of fairly large refactors coming through the pipeline right now. Perhaps someday when we've matured as a codebase, we can revisit this PR.

PR #428, opened 2019-09-26
feat: Add Elixir module

logicmason:
#### Description
- New module `elixir` that calls `elixir -v` to get the versions of Elixir and Erlang/OTP
- Simple tests for the formatting and call to `elixir`
- Adds Elixir and other missing languages to list in JA doc
- Alphabetizes languages
- Lints markdown (tables had minor issues)

#### Motivation and Context
This adds support for Elixir projects.

#### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?
- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [x] I have tested using **Windows**

#### Checklist:
- [ ] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #437, opened 2019-09-27
feat: Add Elixir module.

### LINKED ISSUE #149: Add Elixir on starship
## Feature Request
It could be really cool having Elixir (and if possible Erlang too) support on starship

KokaKiwi:
#### Description
Add an Elixir module, detecting Elixir and OTP versions when it detects a `mix.exs` file.

#### Motivation and Context
I have some Elixir projects and it missed me to not see versions while many others languages have them displayed.

Closes #149

#### Types of changes
- [x] Add `elixir` module

image

- [x] I added tests for parsing `elixir --version` output.

#### Checklist:
- [x] I have added acceptance tests.
- [ ] Maybe add some more detection code (like testing if there's `.ex` or `exs` files)

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -102,6 +102,7 @@ prompt_order = [
     "hg_branch",
     "package",
     "dotnet",
+    "elixir",
     "elm",
     "golang",
     "haskell",
@@ -468,6 +469,29 @@ The module will be shown only if any of the following conditions are met:
 [env_var]
 variable = "SHELL"
 default = "unknown shell"
+```
+
+## Elixir
+
+The `elixir` module shows the currently installed version of Elixir and Erlang/OTP.
+The module will be shown if any of the following conditions are met:
+
+- The current directory contains a `mix.exs` file.
+
+### Options
+
+| Variable   | Default      | Description                                            |
+| ---------- | ------------ | ------------------------------------------------------ |
+| `symbol`   | `"💧 "`      | The symbol used before displaying the version of Rust. |
+| `disabled` | `false`      | Disables the `elixir` module.                          |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[elixir]
+symbol = "🔮 "
 ```
 
 ## Git Branch
--- a/src/modules/elixir.rs
+++ b/src/modules/elixir.rs
@@ -1,0 +1,77 @@
+use ansi_term::Color;
+use regex::Regex;
+use std::process::Command;
+
+use super::{Context, Module};
+
+const ELIXIR_VERSION_PATTERN: &str = "\
+Erlang/OTP (?P<otp>\\d+)[^\\n]+
+
+Elixir (?P<elixir>\\d[.\\d]+).*";
+
+/// Create a module with the current Elixir version
+///
+/// Will display the Rust version if any of the following criteria are met:
+///     - Current directory contains a `mix.exs` file
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let is_elixir_project = context.try_begin_scan()?.set_files(&["mix.exs"]).is_match();
+
+    if !is_elixir_project {
+        return None;
+    }
+
+    let (otp_version, elixir_version) = get_elixir_version()?;
+
+    let mut module = context.new_module("elixir");
+    let module_style = module
+        .config_value_style("style")
+        .unwrap_or_else(|| Color::Purple.bold());
+    module.set_style(module_style);
+
+    module.new_segment("symbol", "🔮 ");
+    module.new_segment(
+        "version",
+        &format!("{} (OTP {})", elixir_version, otp_version),
+    );
+
+    Some(module)
+}
+
+fn get_elixir_version() -> Option<(String, String)> {
+    let output = Command::new("elixir")
+        .arg("--version")
+        .output()
+        .ok()
+        .and_then(|output| String::from_utf8(output.stdout).ok())?;
+
+    parse_elixir_version(&output)
+}
+
+fn parse_elixir_version(version: &str) -> Option<(String, String)> {
+    let version_regex = Regex::new(ELIXIR_VERSION_PATTERN).ok()?;
+    let captures = version_regex.captures(version)?;
+
+    let otp_version = captures["otp"].to_owned();
+    let elixir_version = captures["elixir"].to_owned();
+
+    Some((otp_version, elixir_version))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_elixir_version() {
+        const OUTPUT: &str = "\
+Erlang/OTP 22 [erts-10.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
+
+Elixir 1.9.0 (compiled with Erlang/OTP 22)
+";
+
+        assert_eq!(
+            parse_elixir_version(OUTPUT),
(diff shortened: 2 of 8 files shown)


src/modules/elixir.rs
```diff
@@ -0,0 +1,81 @@
...
+use std::process::Command;
+
+use super::{Context, Module};
+
+const ELIXIR_VERSION_PATTERN: &str = "\
+Erlang/OTP (?P<otp>\\d+)[^\\n]+
+
+Elixir (?P<elixir>\\d[.\\d]+).*";
+
+lazy_static! {
```
chipbuster: Does using `lazy_static` gain us any advantages over just initializing the regex at the start of the `module` function? Keep in mind that each prompt draw of starship corresponds to a new process, so we can only reuse lazy statics if we use this variable more than once per prompt draw.

src/print.rs
```diff
@@ -26,6 +26,7 @@ const DEFAULT_PROMPT_ORDER: &[&str] = &[
     "golang",
     "java",
     "nix_shell",
+    "elixir",
```
chipbuster: Looks like you caught all the places this needs to be added in the code. Well done!

src/modules/elixir.rs
```diff
@@ -0,0 +1,81 @@
...
+    let output = Command::new("elixir")
+        .arg("--version")
+        .output()
+        .ok()
+        .and_then(|output| String::from_utf8(output.stdout).ok())?;
+
+    parse_elixir_version(&output)
+}
+
+fn parse_elixir_version(version: &str) -> Option<(String, String)> {
```
chipbuster: As someone who has no idea how elixir works: what's the OTP version?

docs/config/README.md
```diff
@@ -340,6 +340,28 @@ variable = "SHELL"
 default = "unknown shell"
 ```
 
+## Elixir
```
chipbuster: I believe there's also a `Default Prompt Order` section in this doc somewhere--have you edited that to include this module?

src/modules/elixir.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_elixir_version() {
+        const OUTPUT: &str = "\
+Erlang/OTP 22 [erts-10.5] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
```
chipbuster: Phew, I can see why you want to use regexes for this!

chipbuster:
You know what, I just realized this is strongly overlapping with #428.

I'll review that one as well and see how we want to proceed.

docs/config/README.md
```diff
@@ -340,6 +340,28 @@ variable = "SHELL"
...
+The `elixir` module shows the currently installed version of Elixir and Erlang/OTP.
+The module will be shown if any of the following conditions are met:
+
+- The current directory contains a `mix.exs` file.
+
+### Options
+
+| Variable   | Default      | Description                                            |
+| ---------- | ------------ | ------------------------------------------------------ |
+| `symbol`   | `"🔮 "`      | The symbol used before displaying the version of Rust. |
```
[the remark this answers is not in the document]
KokaKiwi: Ah forgot that one indeed, will fix that soon, thanks :)

src/modules/elixir.rs
KokaKiwi: Forgot the process run every time, so yeah this line is totally useless actually, i will remove it and and add the regex where it's actually used.

src/modules/elixir.rs
KokaKiwi: OTP it the name of the Erlang runtime system and library (Erlang being the "core" of Elixir, which it compiles to), you can have more info on the OTP repository: https://github.com/erlang/otp

> OTP is a set of Erlang libraries, which consists of the Erlang runtime system, a number of ready-to-use components mainly written in Erlang, and a set of design principles for Erlang programs. [Learn more about Erlang and OTP.](http://erlang.org/doc/system_architecture_intro/sys_arch_intro.html)

docs/config/README.md
KokaKiwi: I think I missed that one, will fix that ^-^

src/modules/elixir.rs
KokaKiwi: Actually maybe I could just do some splits but it would make the code a bit ugly and the actual `regex` perfs make the parsing actually sufficiently fast i think and it's cleaner this way i think.

KokaKiwi:
> Overall, this looks pretty good. A few small doc changes (we're trying to improve the process, but there's just a whole bunch of places you have to include things at the moment, so not your fault that you missed a few) and two general questions for you:

>     * Do you think you'll add any acceptance tests to this module? 
I'll look into that, i'm not really good at making tests so I'll try to see what i can do :)

>     * Have you tested the performance of this PR on your machine?
Well I actually use it everyday as it comes from a fork i made to get some personal changes: https://gitlab.kokakiwi.net/contrib/starship ^^ (there's some changes i'll propose on upstream but it needs some more works for that, like documenting and stuff ^^')

And it actually run pretty ok, the only thing is that having a complex prompt makes it slower that the default one, but actually it's ok

> You know what, I just realized this is strongly overlapping with #428.
> 
> I'll review that one as well and see how we want to proceed.
Didn't that one :o

But i see a significantly difference from my PR in that it actually show the OTP version Elixir was compiled with and not the one actually instead, which seems a bit wrong to me (maybe i should put that in a comment on the other PR instead :thinking: )

src/modules/elixir.rs
logicmason: > the OTP version Elixir was compiled with and not the one actually instead

My reasoning for that was that the OTP version used to compile Elixir impacts an Elixir project. One example I've seen is that people with Erlang/OTP 21 installed [encountered problems](
https://github.com/elixir-lang/elixir/issues/5851#issuecomment-406853833) running Dialyzer on Elixir 1.6.6-otp-19.

In that case the output of `elixir -v` was:
```
Erlang/OTP 21 [erts-10.0.1] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1]
Elixir 1.6.6 (compiled with OTP 19)
```
and in order to get the new functionality, they needed to Elixir to be compiled on OTP 20+. This kind of issue is supposed to be rare, but it does happen.

KokaKiwi:
Hi, any news about merging this PR or the other (or not)? :x

matchai:
At the moment, this PR doesn't have any integration tests, which we should have for all future modules.
It is difficult right now to add new modules with integration tests until #289 is addressed, which I am hoping to take care of in the near future.

KokaKiwi:
> At the moment, this PR doesn't have any integration tests, which we should have for all future modules.
> It is difficult right now to add new modules with integration tests until #289 is addressed, which I am hoping to take care of in the near future.

Oh okay, i'll see about trying to add some of these tests, but yeah with #289 i think too it'd be easier, but i'll try with the difficult way first and maybe do changes if it's addressed by then

matchai:
#289 has been merged and is now being used when testing new modules.
You can take a look at #867 for reference when porting tests over.

jwaldrip:
would love to see this be merged.

KokaKiwi:
Just added tests using the new `utils::exec_cmd` stuff with dummy output for acceptance testing.

Kinda hope it should finally "clear" this PR from any missing stuff ^^

chipbuster:
@all-contributors Please add @KokaKiwi for code, doc, tests.

PR #451, opened 2019-09-30
fix: Disable memory_usage by default

matchai:
#### Description

This PR disables the `memory_usage` module by default.

#### Motivation and Context

The `memory_usage` module is very long  (many characters in length) by default, and can be active in any directory based on current memory usage. For many, this may be considered a little too noisy and unactionable.

The modules we usually have enabled by default should be affected by your working directory, keeping modules that aren't smaller or optional.

#### Types of changes

#### Checklist:


nickwb:
Yeah, if most people are seeing this in their prompt most of the time, then I think it's reasonable to disable it by default.

PR #454, opened 2019-09-30
refactor: Rewrite battery module to use module config

heyrict:
#### Motivation and Context
Related: #383

#### Types of changes

#### Checklist:

### DIFF
--- a/starship/src/configs/battery.rs
+++ b/starship/src/configs/battery.rs
@@ -1,25 +1,35 @@
-use crate::config::{ModuleConfig, RootModuleConfig};
+use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
 
 use ansi_term::{Color, Style};
 use starship_module_config_derive::ModuleConfig;
 
 #[derive(Clone, ModuleConfig)]
 pub struct BatteryConfig<'a> {
-    pub full_symbol: &'a str,
-    pub charging_symbol: &'a str,
-    pub discharging_symbol: &'a str,
-    pub unknown_symbol: Option<&'a str>,
-    pub empty_symbol: Option<&'a str>,
+    pub full_symbol: SegmentConfig<'a>,
+    pub charging_symbol: SegmentConfig<'a>,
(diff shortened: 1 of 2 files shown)


starship/src/configs/battery.rs
```diff
@@ -1,32 +1,46 @@
...
 
 impl<'a> RootModuleConfig<'a> for BatteryConfig<'a> {
     fn new() -> Self {
         BatteryConfig {
-            full_symbol: "•",
-            charging_symbol: "↑",
-            discharging_symbol: "↓",
+            full_symbol: SegmentConfig {
+                value: "•",
+                style: None,
```
bijancn: Mhh. You changed the meaning now, right? It used to be just the `symbol` and now it is `symbol` and `style`. How about simplifying the naming to `full`, `charging`, `discharging`, `unknown` and `empty`? That's a breaking change for users but this is breaking anyway, right?

heyrict: Renaming do make sense, but it will not break previous code.

This is equivalent to calling `config_value_segment_config` if you are familiar with #378.

bijancn: Mhh. I might be understanding it wrongly but I would assume that when I had earlier in my configuration
```
[battery]
full_symbol = "🔋"
```
This won't be recognized now anymore but has to be changed to
```
[battery.full_symbol]
    value = "🔋"
```
or is `value` a magic default `key`?

heyrict: No, both

```toml
[battery]
full_symbol = "B "
```

and

```toml
[battery]
full_symbol = { value = "B " }
```

will work as described [here](https://github.com/heyrict/starship/blob/26db0cef52ebc7fdcac0ce59a75448d854a58ce0/docs/config/README.md#styling) (currently undocumented due to discussion in #378)

If you want reference to code, [here it is](https://github.com/heyrict/starship/blob/master/starship/src/config.rs#L181-L220).

bijancn: Thanks for the explanation

PR #460, opened 2019-10-01
refactor: Rewrite cmd_duration, directory and env_var module to use module config

heyrict:
#### Description
This PR is a batched rewrite of the following modules:

- cmd_duration
- directory
- env_var

#### Motivation and Context
Related #383

#### Types of changes

#### Checklist:


starship/src/configs/directory.rs
```diff
@@ -0,0 +1,32 @@
...
+            truncation_length: 3,
+            truncate_to_repo: true,
+            fish_style_pwd_dir_length: 0,
+            use_logical_path: true,
+            style: Color::Cyan.bold(),
+            disabled: false,
+            path: SegmentConfig {
+                value: "",
+                style: None,
+            },
```
heyrict: `path` is an undocumented module and you can configure the style now with

```toml
[directory]
path.style = "normal"
```

docs/config/README.md
```diff
@@ -262,7 +262,7 @@ running `eval $(starship init $0)`, and then proceed as normal.
 | Variable   | Default         | Description                                                |
 | ---------- | --------------- | ---------------------------------------------------------- |
 | `min_time` | `2`             | Shortest duration to show time for.                        |
-| `prefix`   | `took`          | Prefix to display immediately before the command duration. |
+| `prefix`   | `took `         | Prefix to display immediately before the command duration. |
```
nickwb: Did you intend to make this change?

src/configs/cmd_duration.rs
```diff
@@ -0,0 +1,24 @@
+use crate::config::{ModuleConfig, RootModuleConfig};
+
+use ansi_term::{Color, Style};
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
+pub struct CmdDurationConfig<'a> {
+    pub min_time: i64,
+    // TODO: Rewrite prefix with SegmentConfig
```
nickwb: Are we going to address this `TODO` before landing the PR?

docs/config/README.md
heyrict: The default value seems to be `took ` when I inspect the code.

src/configs/cmd_duration.rs
heyrict: Oh sorry, I should probably open another PR instead of leaving a TODO comment here as it seems to be a new feature. Thanks for the question!

PR #463, opened 2019-10-01
feat: Implement timezone offset config option for the time module

### LINKED ISSUE #431: Specify time zone in time module
## Feature Request

The `time` module currently only shows the local system's time. It would be nice if you could choose a time zone, e.g. so you could always show the timestamp in UTC.

keawade:
#### Description

Allows users to configure the time module to display time with a timezone offset other than just their local timezone.

I've added an option, `utc_time_offset`, to the `time` config that accepts an integer or float for the timezone offset. This allows users to specify a timezone via the offset. The offset is then used to change the displayed time from the `time` module to show the current time with the given offset instead of local time.

For example, `utc_time_offset = "-8"` would configure the timezone offset to be the Pacific timezone.

The configuration supports floats to accommodate [30/45 minute offsets](https://www.timeanddate.com/time/time-zones-interesting.html).

#### Motivation and Context

Closes #431

#### Types of changes

#### Screenshots:

Screenshot from 2019-09-30 07-26-20

The time module doesn't have much testing due to existing constraints. I added tests to cover the new formatting function I needed to format the offset times matching the existing format function tests.

I have also manually reviewed the changes on my machine with `fish`.

#### Checklist:


.all-contributorsrc
```diff
@@ -383,6 +383,16 @@
...
+    },
+    {
+      "login": "keawade",
+      "name": "Keith Wade",
+      "avatar_url": "https://avatars0.githubusercontent.com/u/7308850",
+      "profile": "https://github.com/keawade",
+      "contributions": [
+        "code",
+        "test"
+      ]
```
bijancn: Note sure you are supposed to edit this file by hand 😄

starship/src/modules/time.rs
```diff
@@ -30,8 +35,33 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+            let utc_offset_in_seconds: i32 = (utc_time_offset_in_hours * 3600_f32) as i32;
+            let timezone_offset = FixedOffset::east(utc_offset_in_seconds);
+            log::trace!("Target timezone offset is {}", timezone_offset);
+
+            let target_time = utc_time.with_timezone(&timezone_offset);
+            log::trace!("Time in target timezone now is {}", target_time);
+
+            formatted_time_string = format_time_fixed_offset(&time_format, target_time);
+        } else {
+            // Fall back to local time if an invalid offset is provided.
```
bijancn: I would log this as an error.

starship/src/modules/time.rs
```diff
@@ -102,4 +136,81 @@ mod tests {
...
+    #[test]
+    fn test_format_with_paren_fixed_offset() {
+        let timezone_offset = FixedOffset::east(0);
+        let time = Utc
+            .ymd(2014, 7, 8)
+            .and_hms(15, 36, 47)
+            .with_timezone(&timezone_offset);
+        let formatted = format_time_fixed_offset("[%T]", time);
+        assert_eq!(formatted, "[15:36:47]");
+    }
```
bijancn: Cool that you added formatter tests. However I was assuming, you would also test your new offset logic here. Something like
- Happy path +3, -8, +5.5, -3.9, "local"
- Edge cases +24, -24
- Wrong input +5000, -200

starship/src/modules/time.rs
```diff
@@ -30,8 +35,33 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         time_format
     );
 
-    let local: DateTime<Local> = Local::now();
-    let formatted_time_string = format_time(&time_format, local);
+    let local_time: DateTime<Local> = Local::now();
+    log::trace!("Local time now is {}", local_time);
+
+    let formatted_time_string: String;
```
bijancn: you don't need to end the assignment here. It's a simple expression, so you can do
```rust
let formatted_time_string: String = if utc_time_offset_str == "local" {
  format_time(&time_format, local_time)
} else {
...
};
```

.all-contributorsrc
keawade: Good to know. I'll remove my changes. 😆

starship/src/modules/time.rs
keawade: Oh cool. Sorry, I'm pretty new to rust so I'm not aware of things like this. I'll update it. :)

bijancn: No worries. I'm new myself. I just have a head start from Scala 😉

starship/src/modules/time.rs
keawade: It is now logging.

There aren't any other instances of log level `error` though so I left it at `warn` so the logs won't yell at folks every time they render a prompt unless they purposefully change their log level.

starship/src/modules/time.rs
keawade: I extracted the offset logic into a function so I could test it more easily and added a bunch of tests.

starship/src/modules/time.rs
keawade: I've implemented this pattern. Thanks for pointing it out! Really cool. :smile:

keawade:
I'm not sure if I'm supposed to resolve conversations or if the person who opened them is supposed to resolve them. Sorry, not used to GitHub's PR tools. Please let me know if I got it wrong! :smile:

keawade:
@bijancn I was able to add tests for the cases that use the local time (`local` and error fallbacks) but I'm not sure how I can modify the timezone dependent tests so that they pass in the CI.

Any suggestions?

chipbuster:
@keawade Pretty much every PR I've seen on this repo, the person making the changes will mark the comment as resolved once the appropriate change has been made. If it doesn't satisfy the person doing the review, they can always unmark the resolved status or make a new comment.

matchai:
Thank you for taking the time to tackle this!

@all-contributors Please add @keawade for code and test. 😄

PR #469, opened 2019-10-02
feat: Add conda module

### LINKED ISSUE #263: Add conda module
The spaceship prompt shows the current Conda env -- this could be useful. A potential drawback is clutter with the python module, but I believe pyenv and conda env are not used together so this is irrelevant.

Scotsguy:
#### Description

This adds the `conda` module, mostly following [this implementation](https://github.com/denysdovhan/spaceship-prompt/blob/master/sections/conda.zsh)

#### Types of changes

#### Checklist:


src/modules/mod.rs
```diff
@@ -40,8 +41,16 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
         "directory" => directory::module(context),
         "character" => character::module(context),
         "cmd_duration" => cmd_duration::module(context),
+        "conda" => conda::module(context),
         "dotnet" => dotnet::module(context),
         "env_var" => env_var::module(context),
+        "nodejs" => nodejs::module(context),
```
chipbuster: Were these just all missing previously? :o

tests/testsuite/conda.rs
```diff
@@ -0,0 +1,29 @@
...
+    let actual = String::from_utf8(output.stdout).unwrap();
+
+    assert_eq!(expected, actual);
+    Ok(())
+}
+
+#[test]
+fn env_set() -> io::Result<()> {
+    let output = common::render_module("conda")
+        .env_clear()
```
chipbuster: Just FYI, [Render module already does this for us](https://github.com/starship/starship/blob/master/tests/testsuite/common.rs#L32).

src/modules/mod.rs
Scotsguy: I honestly don't know how those got in there, must've been something during a rebase

tests/testsuite/conda.rs
Scotsguy: I just copied these from the [aws testsuite](https://github.com/starship/starship/blob/master/tests/testsuite/aws.rs), basically

src/modules/mod.rs
chipbuster: I swear Git is just mocking me now...

Alright, if you don't mind, could you sort these into alphabetical order? Not entirely sure how they keep unsorting themselves, but we'd like to keep an order where we can (in accordance with the comment at the top of the `match`)

tests/testsuite/conda.rs
chipbuster: Alright, looks like we'll have some cleanup work to do in the future. Thanks for letting me know.

PR #473, opened 2019-10-03
fix: Don't print the first module prefix after a line-break

### LINKED ISSUE #461: Don't show the prefix of the first module after the `line_break`
#### Current Behavior

Currently, the first module after the `line_break` module will have its prefix visible. This makes the new line start with a prefix, which I think looks worse than it would without the prefix:

image

This is what the prompt looks like with just the `time` module enabled:
```toml
[time]
disabled = false
```

#### Expected Behavior

Similarly to how the first prefix in the prompt doesn't appear, neither should the first prefix after a `line_break` module.

#### Possible Solution

The implementation that hides the first prefix can be found here:
https://github.com/starship/starship/blob/master/starship/src/print.rs#L46-L50

Maybe similar logic should be implemented following after the `line_break` module is printed.

pt2121:
#### Description
Not show the prefix of the first module of the 2nd line.
I'd never used Rust before - not sure if there is any more elegant way to do this so feedback is welcome.

#### Motivation and Context
Closes #461

#### Types of changes

Screen Shot 2019-10-02 at 7 57 59 PM

I config'ed my prompt to have 2 lines and made sure the prefixes of the 1st modules not show up.

#### Checklist:

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -41,16 +41,20 @@ pub fn prompt(args: ArgMatches) {
         .flatten()
         .collect::<Vec<Module>>(); // Remove segments set to `None`
 
+    let mut print_without_prefix = true;
     let mut printable = modules.iter();
 
-    // Print the first module without its prefix
-    if let Some(first_module) = printable.next() {
-        let module_without_prefix = first_module.to_string_without_prefix();
-        write!(handle, "{}", module_without_prefix).unwrap()
-    }
+    for module in printable {
+        // Skip printing the prefix of a module after the line_break
+        if print_without_prefix {
+            let module_without_prefix = module.to_string_without_prefix();
+            write!(handle, "{}", module_without_prefix).unwrap()
+        } else {
+            write!(handle, "{}", module).unwrap();
+        }
 
-    // Print all remaining modules
-    printable.for_each(|module| write!(handle, "{}", module).unwrap());
+        print_without_prefix = module.get_name() == "line_break"
+    }
 }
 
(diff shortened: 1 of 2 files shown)


src/print.rs
```diff
@@ -49,7 +49,18 @@ pub fn prompt(args: ArgMatches) {
         write!(handle, "{}", module_without_prefix).unwrap()
     }
 
-    // Print all remaining modules
+    while let Some(module) = printable.next() {
+        write!(handle, "{}", module).unwrap();
+        if module.to_string().contains('\n') {
```
matchai: I think a slightly more robust and declarative way to handle this would be to check the module's name, rather than its string representation. 🙂

pt2121: Thanks for your review. updated!

I'll squash commits later if you prefer that.

src/print.rs
```diff
@@ -41,16 +41,19 @@ pub fn prompt(args: ArgMatches) {
         .flatten()
         .collect::<Vec<Module>>(); // Remove segments set to `None`
 
+    let mut line_break = true;
```
matchai: I think this may be a better variable name, since it's not only to do with the `line_break` module.

```suggestion
	// Skip printing the prefix of the first module
    let mut print_without_prefix = true;
```

pt2121: very good point. updated!

pt2121:
Awesome, thanks for this good looking prompt and patiently reviewing my PR!

matchai:
@all-contributors Please add @pt2121 for code. 😄

PR #477, opened 2019-10-04
feat: Add docker/docker-compose module

keawade:
#### Description
This PR adds a simple module for docker and docker-compose versions. The module will display the current docker version if a `Dockerfile` or `docker-compose.yml` file are present in the current directory.

docker-compose version can be displayed by setting the config `show_compose = true`. `show_compose` is defaulted to `false` because the docker-compose version command causes a noticeable pause between prompts.

#### Motivation and Context
Addresses some of #393 

#### Types of changes

Screenshot from 2019-10-03 21-08-48
Screenshot from 2019-10-03 21-09-47

I've run this locally and added tests to the test suite.

I'm not sure if those tests are being run though. I tried running the `./acceptance_tests` script and `cargo test` but I didn't see any output indicating they were run. Could someone double check that specifically?

#### Checklist:


matchai:
Is knowing the version of Docker being used actionable information?

This isn't meant to sound condescending but is a genuine question from someone who doesn't use Docker much. The little I've written and used, the installed version of Docker had no effect on the way I worked, but I'd be interested in hearing from people who use it frequently.

Just making sure the info we're adding to the prompt is useful and actionable. 🙂

keawade:
That's a good question. I was thinking about that while discussing the feature request over in #393 a little bit ago.

In my use, the specific version isn't particularly valuable to know but I appreciated having it in `spacefish` because it helped me know if I was in a directory that used docker or docker-compose.

nickwb:
Having a `v` in front of the docker version, but not in front of the docker-compose version, seems a little odd.

> In my use, the specific version isn't particularly valuable to know but I appreciated having it in spacefish because it helped me know if I was in a directory that used docker or docker-compose.

Maybe we should have a configuration option which causes the module to show without the version numbers. e.g., `🐳 +C`

keawade:
> Having a `v` in front of the docker version, but not in front of the docker-compose version, seems a little odd.

Oh sorry, that's just an oversight on my part. I'll update that tonight.

> Maybe we should have a configuration option which causes the module to show without the version numbers. e.g., `🐳 +C`

Yeah I can do that.

keawade:
I've fixed the `docker-compose` version text and added a `show_versions` config.

I've set the default value for `show_versions` to `true` so it matches the default behavior of other modules.

matchai:
Speaking honestly, I'm concerned about adding a module that adds so much data that would be unactionable to the majority of users seeing it. I don't think that showing the versions by default is the right approach.

Would just showing the docker symbol be useful to people?
Are there any other datapoints we can derive other than the number of expected containers running?
Maybe this would best be revisited when we have caching in place.

keawade:
> Speaking honestly, I'm concerned about adding a module that adds so much data that would be unactionable to the majority of users seeing it. I don't think that showing the versions by default is the right approach.

Does that mean you would prefer to show only the icon or that you would prefer to not add the module at all?

> Would just showing the docker symbol be useful to people?

Yes. The two use cases mentioned so far have been to:
* Note that you are in a directory with `docker`/`docker-compose` files (This is my primary use case)
* Differentiate between a `docker`/`docker-compose` directory and it's running environment (@re1 was requesting this in #393)

Displaying the `docker`/`docker-compose` versions are not necessary for that goal. Just the icon will suffice for that.

> Are there any other datapoints we can derive other than the number of expected containers running?

I think so, but the things I'm thinking of aren't really feasible without a daemon to do the checks in the background. I'll add a comment to #393 with the idea I'm thinking of so we can reference them later.

> Maybe this would best be revisited when we have caching in place.

I think this first piece of functionality is a useful starting point and is valuable even without the more advanced features we have discussed.

However, if we are planning to come back later then I think the `docker-compose` version functionality I've implemented here should be removed and added as part of that effort. I disabled showing the `docker-compose` version by default because it noticeably impacts prompt load speed.

PR #480, opened 2019-10-04
refactor: Go from Rust workspaces to a package with nested packages

matchai:
#### Description

When attempting to publish starship after the refactor, it looked like the workspace arrangement lead to the starship project no longer being able to use `README.md` at the root of the project.

I thought that simply setting the `readme` to `../README.md` in `Cargo.toml` would do the trick, but the manifest docs say the following:

> This (readme) points to a file under the package root (relative to this `Cargo.toml`).

https://doc.rust-lang.org/cargo/reference/manifest.html

That would mean that we would need a second README.md (or perhaps a symlink). Thinking it over, I don't think we've reached the complexity necessary for workspaces to be useful to us. 

For instance, [ripgrep](https://github.com/BurntSushi/ripgrep), a considerably larger project, has the project structured like this PR; a single root-level package, and several sub-packages within it.

Thoughts?


wyze:
This looks good. It also fixes the issue of the tests suite not running. Which it is now and has 2 failures.

heyrict:
LGTM. I shouldn't have moved src to a subfolder if it works without a workspace :+1:

PR #482, opened 2019-10-04
feat: Show AWS region in `aws` module

### LINKED ISSUE #443: Add AWS region module
## Feature Request

Continuing #400, it is nice to see the AWS profile I'm under. It would be even nicer to see the region.

I should be able to see the region I'm in, if set at all. The region could be retrieved by running `aws configure get region`.

Optionally I could also set a "default" region; the module should only tell me the region when I'm not in my default region.

pbzweihander:
#### Description

Make `aws` module also show AWS region

#### Motivation and Context

Closes #443 

#### Types of changes

screenshot_2019-10-04_23-01-21

#### Checklist:


pbzweihander:
I'm not sure how to test with `AWS_PROFILE` env var and `~/.aws/config` file. Should I create `~/.aws/config` file in the test code? Or in the Dockerfile?

andytom:
Would it be better to make this part of the existing aws module rather than a separate module?

pbzweihander:
@andytom I have updated this PR so `aws` module can show AWS region as well!

andytom:
> @andytom I have updated this PR so `aws` module can show AWS region as well!

Awesome, I'll try and take a proper look at this tomorrow morning.

src/modules/aws.rs
```diff
@@ -1,15 +1,72 @@
...
+
+    let region = region_line.split('=').nth(1)?;
+    let region = region.trim();
+
+    Some(region.to_string())
+}
+
+fn get_aws_region() -> Option<String> {
+    env::var("AWS_DEFAULT_REGION")
+        .or_else(|_| env::var("AWS_REGION"))
```
andytom: Is the precedent here correct? Should `AWS_REGION` not be prefered if both `AWS_REGION` and `AWS_DEFAULT_REGION` are both set?

src/modules/aws.rs
```diff
@@ -21,7 +78,13 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
 
     module.create_segment("symbol", &config.symbol);
-    module.create_segment("profile", &config.profile.with_value(&aws_profile));
+    module.create_segment(
+        "region",
+        &SegmentConfig {
+            value: &aws_region,
+            style: None,
+        },
+    );
```
andytom: Would it be better to have separate segments for the `profile` and the `region`?

src/modules/aws.rs
```diff
@@ -1,15 +1,72 @@
...
+fn get_aws_region() -> Option<String> {
+    env::var("AWS_DEFAULT_REGION")
+        .or_else(|_| env::var("AWS_REGION"))
+        .ok()
+        .or_else(|| {
+            let aws_profile = env::var("AWS_PROFILE").ok();
+            let aws_region = get_aws_region_from_config(&aws_profile);
+
+            match (aws_profile, aws_region) {
+                (Some(profile), Some(region)) => Some(format!("{}({})", region, profile)),
```
andytom: My personal preference would be to have the profile then the region rather than the region then the profile.

```suggestion
                (Some(profile), Some(region)) => Some(format!("{}({})", profile, region)),
```

src/modules/aws.rs
pbzweihander: My mistake! Actually, referring to https://github.com/aws/aws-cli/issues/735, `aws-cli` does not use `AWS_REGION` at all.

pbzweihander: But I will keep `AWS_REGION` at least precedent because I heard some of the AWS-related programs are still using `AWS_REGION` env var.

src/modules/aws.rs
```diff
@@ -1,16 +1,85 @@
...
+
+fn get_aws_region() -> Option<(String, String)> {
+    env::var("AWS_DEFAULT_REGION")
+        .ok()
+        .map(|region| (String::new(), region))
+        .or_else(|| {
+            let aws_profile = env::var("AWS_PROFILE").ok();
+            let aws_region = get_aws_region_from_config(&aws_profile);
+
+            if aws_profile.is_none() && aws_region.is_none() {
```
nickwb: This whole `if` can be replaced if you use the `?` operator on lines 52 & 53:

```
let aws_profile = env::var("AWS_PROFILE").ok()?;
let aws_region = get_aws_region_from_config(&aws_profile)?;
Some((aws_profile, aws_region))
```

(Note I didn't try to compile this, but it should be fairly close)

src/modules/aws.rs
```diff
@@ -1,16 +1,85 @@
...
+                Some((
+                    aws_profile.unwrap_or_default(),
+                    aws_region.unwrap_or_default(),
+                ))
+            }
+        })
+        .or_else(|| {
+            env::var("AWS_REGION")
+                .ok()
+                .map(|region| (String::new(), region))
```
nickwb: This is a bit of a hint that your return type is wrong. It seems like profile should be an `Option`.

I suspect you could neaten the whole module up with a struct:
```
struct AwsStatus {
    profile: Option<String>,
    region: Option<String>,
}
```

src/modules/aws.rs
pbzweihander: This is different semantic. I meant to return None when `aws_region` AND `aws_region` is both None.

src/modules/aws.rs
pbzweihander: In line 92 and 93, we need to make `SegmentConfig` which takes `&str`. If it is Option, we need to `unwrap_or_default` anyway.

src/modules/aws.rs
nickwb: Ok, yes - I see there is a slight difference. If you think the distinction is important, then you can leave it as-is.

PR #483, opened 2019-10-04
refactor: Rewrite `time` module to use module config

pbzweihander:
#### Description

Rewrite `time` module to use module config

Since the new module config uses struct fields, we cannot use `12hr` as a config key. So I updated `12hr` to `use_12hr`.

If `ModuleConfig` proc macro has `rename`-like attribute, we can keep backward compatibility.

#### Motivation and Context

Related #383

#### Types of changes

#### Checklist:

### DIFF
--- a/src/configs/time.rs
+++ b/src/configs/time.rs
@@ -1,0 +1,28 @@
+use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
+
+use ansi_term::{Color, Style};
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
+pub struct TimeConfig<'a> {
+    pub time: SegmentConfig<'a>,
+    pub use_12hr: bool,
+    pub format: Option<&'a str>,
+    pub style: Style,
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for TimeConfig<'a> {
+    fn new() -> Self {
+        TimeConfig {
+            time: SegmentConfig {
+                value: "",
+                style: None,
+            },
+            use_12hr: false,
+            format: None,
+            style: Color::Yellow.bold(),
+            disabled: true,
+        }
+    }
+}
--- a/src/configs/time.rs
+++ b/src/configs/time.rs
@@ -1,11 +1,10 @@
-use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
+use crate::config::{ModuleConfig, RootModuleConfig};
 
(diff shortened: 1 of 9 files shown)


src/configs/time.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for TimeConfig<'a> {
+    fn new() -> Self {
+        TimeConfig {
+            time: SegmentConfig {
+                value: "",
+                style: None,
+            },
```
heyrict: Sorry for the inconvenience but styling an undocumented segment is currently under discussion in #472. This syntax shouldn't be used until we get that issue resolved.

You can refer to b0bf17277e868026efe678d81915cc4941aa9e50 to drop it.

src/modules/time.rs
```diff
@@ -1,39 +1,33 @@
...
+    let mut module = context.new_module("time");
+    let config: TimeConfig = TimeConfig::try_load(module.config);
 
-    let default_format = if is_12hr { "%r" } else { "%T" };
-    let time_format = module
-        .config_value_str("format")
-        .unwrap_or(default_format)
-        .to_owned();
+    let default_format = if config.use_12hr { "%r" } else { "%T" };
+    let time_format = config.format.unwrap_or(default_format);
```
heyrict: Instead of using `format: Option<&str>` in the struct, you can use `format: &str` and provide a default format in `fn new()`, so that we can store all default value in the same place. What do you think?

heyrict: OK, my bad. It seems that default format depends on `use_12hr`.

pbzweihander:
I think failing tests will be fixed in #486

heyrict:
Tests should be fixed now.

src/modules/time.rs
```diff
@@ -1,39 +1,39 @@
...
+use crate::config::{RootModuleConfig, SegmentConfig};
+use crate::configs::time::TimeConfig;
+
 /// Outputs the current time
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
-    let mut module = context.new_module("time");
-
-    if module.config_value_bool("disabled").unwrap_or(true) {
-        return None;
-    }
```
andytom: I think this (or something similar) will be needed for now see #491.

matchai:
Thank you for taking the time to write this refactor, @pbzweihander! 😄 

@all-contributors Please add @pbzweihander for code.

matchai:
@all-contributors Also, please add @andytom for review. 😄

PR #485, opened 2019-10-04
feat: Add configuration for hostname truncation

zekesonxx:
#### Description
Added the simple ability to trim some text off the end of the hostname. Ex: `server.example.com`, if `trim_end=".example.com"`, can become just `server`

#### Motivation and Context
I would like to use starship for my work shell prompt, but having `servername.asdf.example.com` show up takes up a *lot* of space, and it needs to get trimmed down to `servername`.

I could trim it to the first dot (which is what bash and zfs like doing, and might not be a bad additional feature), except that some of my work machines are `desktop.servername.asdf.example.com` or `speedtest.servername.asdf.example.com`, and just having `desktop` or `speedtest` show up is annoying because I don't know *which* one I'm connected to

#### Types of changes

#### Checklist:


matchai:
Perhaps we could unify the two features you mentioned:
- Trimming the hostname
- Trimming to the first dot

If instead of `trim_end`, it could act as a `trim_from` or `trim_at`, trimming the remaining hostname once the value is matched.

```toml
[hostname]
# Input: speedtest.servername.asdf.example.com
trim_from = "."
# Output: speedtest

[hostname]
# Input: speedtest.servername.asdf.example.com
trim_from = ".asdf.example.com"
# Output: speedtest.servername
```

What are your thoughts, @zekesonxx?

zekesonxx:
@matchai that seems reasonable. Let me see if I can implement that...

zekesonxx:
Okay, so, `trim_at` implemented.

A selection of example results, as based on the hostname `speedtest.cluster.asdf.example.com`:
* `.`: `speedtest` (this is what bash and zsh do by default, and is now the default here too)
* `.asdf`: `speedtest.cluster`
* `.asdf.example.com`: `speedtest.cluster`
* `.example.com`: `speedtest.cluster.asdf`
* `t`: `speed`
* `e`: `sp`
* `ee`: `sp`

matchai:
Thank you for your contribution, Zach! 😄 

@all-contributors Please add @zekesonxx for code, docs, and test.

PR #488, opened 2019-10-05
fix: Disable Kubernetes module by default

### LINKED ISSUE #484: Kubernetes module is on by default and shown everywhere
The recently released kubernetes module is on by default and appears to be shown everywhere, rather than practicing heuristics to determine when it's appropriate to show kubernetes info in the prompt, like other modules do. This is a nuisance for people who happen to have kubernetes installed, but do not necessarily use it so often that they always want to see info about it.

#### Environment

- Starship version: v0.20.1

#### Possible Solution

Look for kubernetes configuration in the working directory to determine whether or not to display kubernetes info in the prompt.

andytom:
#### Description

Given the global nature of the Kubernetes module have disabled the
module by default. Have also taken this opertunity to refactor the
Kubernetes module to use the new config module.

#### Motivation and Context

Closes #484

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -574,13 +574,20 @@ can be done via `kubectl config set-context starship-cluster --namespace
 astronaut`. If the `$KUBECONFIG` env var is set the module will use that if
 not it will use the `~/.kube/config`.
 
+::: tip
+
+This module is disabled by default.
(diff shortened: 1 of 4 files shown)


matchai:
Thanks again for your contribution! 😄 

@all-contributors Please add @andytom for code.

PR #491, opened 2019-10-05
fix: Add disable option to kubernetes module

heyrict:
#### Description
This PR adds one line to skip the module if it is not enabled in the config.
It also renames `fn is_module_enabled` to better reflex its functionality.

#### Motivation and Context
I used to think `is_module_enabled` has a higher order functionality to judge if the module is disabled, but it turns out it only reads the user's config and ignores the default one. (Thanks to @andytom to point it out!)
Related #484 #486 #489

#### Types of changes

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -77,14 +77,14 @@ impl<'a> Context<'a> {
         Module::new(name, config)
     }
 
-    /// Check the `disabled` configuration of the module
-    pub fn is_module_enabled(&self, name: &str) -> bool {
+    /// Check if `disabled` option of the module is true in configuration file.
+    pub fn is_module_disabled_in_config(&self, name: &str) -> bool {
         let config = self.config.get_module_config(name);
 
         // If the segment has "disabled" set to "true", don't show it
         let disabled = config.and_then(|table| table.as_table()?.get("disabled")?.as_bool());
 
-        disabled != Some(true)
+        disabled == Some(true)
     }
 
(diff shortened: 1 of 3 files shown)


heyrict:
Also note that, this is just a hot-fix. The `disabled` option doesn't seem to be quite neat.

`print.rs` can filter out modules with `disabled = false` in user's config. Thus all modules enabled by default will continue to work, but modules disabled by default will have to add an extra line to get it disabled.

heyrict:
Another approach (as noted in discord) in brief is:

1. Add `fn is_disabled(&self) -> bool` to `trait RootModuleConfig`
1. Add `fn get_config(name: &str) -> Option<impl RootModuleConfig>` similar to `modules/mod.rs` to `configs/mod.rs`.
1. In `fn is_module_enabled`, use

    ```
    pub fn is_module_enabled(&self, name: &str) -> bool {
        let config = get_config(name);

        let disabled = config.and_then(|config| config.is_disabled());

        disabled == Some(true)
    }
    ```

matchai:
That sounds solid to me.

matchai:
Do you think this is something we can implement for this PR, or should I go ahead and merge this as it is?

heyrict:
@matchai Well, This PR is just a quick fix. I think it should be merged to fix the regression first.
I'll open a new PR related to https://github.com/starship/starship/pull/491#issuecomment-538651086 if I have enough time.

PR #492, opened 2019-10-05
chore: Add shared import types to `modules/mod.rs`

heyrict:
#### Description
As is suggested by @nrabulinski , it seems more idiomatic not to use `crate::config::*` and `crate::configs::*` together.

This is just a demo and there are many ways to fix this. e.g. Add `pub use crate::configs::*;` in `config.rs` and use `crate::config::{RootModuleConfig, module_name::ModuleNameConfig};` afterwards, following @nrabulinski 's suggestion.

Feel free to close this PR if it doesn't look good enough :)

#### Types of changes

#### Checklist:


Scotsguy:
Now also in [conda.rs](https://github.com/starship/starship/blob/master/src/modules/conda.rs)

PR #495, opened 2019-10-05
refactor: Default trait implementation for SegmentConfig

nrabulinski:
#### Description
I implemented the `Default` trait for `SegmentConfig` so there's no need for writing
```rust
SegmentConfig {
    value: "",
    style: None
}
```
for each segment in every module and instead one can just call `SegmentConfig::default()`.
I also added an associated function `new` which only takes `&'static str` as a parameter to write `SegmentConfig::new("🦀 ")` opposed to
```rust
SegmentConfig {
    value: "🦀 ",
    style: None
}
```

#### Motivation and Context

Reducing the amount of repeated code and making the codebase more idiomatic.

#### Types of changes

#### Checklist:


chipbuster:
Always love seeing cleaner code.

PR #497, opened 2019-10-06
refactor: Rewrite is_module_enabled

heyrict:
#### Description
This PR disables modules in `print.rs`.

It seems that we cannot multiple types with shared trait as return type. (#491)

Also, #486 can be safely merged after this PR got merged.

#### Motivation and Context
Related #486 #491

#### Types of changes

#### Checklist:

### DIFF
--- a/src/configs/mod.rs
+++ b/src/configs/mod.rs
@@ -59,3 +59,18 @@ impl<'a> RootModuleConfig<'a> for StarshipRootConfig<'a> {
         }
     }
 }
+
+/// Get the default disabled option in a module by name
+pub fn is_module_disabled_by_default(name: &str) -> bool {
+    match name {
+        "aws" => aws::AwsConfig::new().disabled,
+        "battery" => battery::BatteryConfig::new().disabled,
+        "character" => character::CharacterConfig::new().disabled,
+        "conda" => conda::CondaConfig::new().disabled,
+        "dotnet" => dotnet::DotnetConfig::new().disabled,
+        "kubernetes" => kubernetes::KubernetesConfig::new().disabled,
+        "rust" => rust::RustConfig::new().disabled,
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -77,14 +78,18 @@ impl<'a> Context<'a> {
...
         let config = self.config.get_module_config(name);
 
-        // If the segment has "disabled" set to "true", don't show it
-        let disabled = config.and_then(|table| table.as_table()?.get("disabled")?.as_bool());
+        // Check the disabled option in the config
+        let config_disabled = config.and_then(|table| table.as_table()?.get("disabled")?.as_bool());
 
-        disabled == Some(true)
+        if let Some(disabled) = config_disabled {
+            disabled
```
nickwb: This seems inverted to me. When `disabled = true`, we return `true` for `is_module_enabled`

src/configs/mod.rs
```diff
@@ -59,3 +59,18 @@ impl<'a> RootModuleConfig<'a> for StarshipRootConfig<'a> {
...
+    match name {
+        "aws" => aws::AwsConfig::new().disabled,
+        "battery" => battery::BatteryConfig::new().disabled,
+        "character" => character::CharacterConfig::new().disabled,
+        "conda" => conda::CondaConfig::new().disabled,
+        "dotnet" => dotnet::DotnetConfig::new().disabled,
+        "kubernetes" => kubernetes::KubernetesConfig::new().disabled,
+        "rust" => rust::RustConfig::new().disabled,
+        "time" => time::TimeConfig::new().disabled,
+        _ => true,
```
nickwb: Defaults to true? That seems incorrect to me.

heyrict:
I'm also thinking this approach not neat enough. I'll try Box and get another PR when it is ready.

PR #507, opened 2019-10-07
fix: Show Java version from OpenJ9 Java runtimes

### LINKED ISSUE #468: OpenJ9 Java runtimes don't show Java version in prompt
#### Current Behavior
When my `JAVA_HOME` environment variable points to an OpenJ9-based AdoptOpenJDK installation, the Java version is not displayed in the Starship prompt.

#### Expected Behavior
I would expect the prompt to include the detected Java version, something like `/tmp/test via ☕ v1.8.0 took 45s `.

Output of `java -Xinternalversion`:

```
Eclipse OpenJ9 OpenJDK 64-bit Server VM (1.8.0_222-b10) from macosx-x86_64 JRE with Extensions for OpenJDK for Eclipse OpenJ9 8.0.222.0, built on Jul 17 2019 23:00:22 by jenkins with i686-apple-darwin11-llvm-g++-4.2 (GCC) 4.2.1 (Based on Apple Inc. build 5658) (LLVM build 2336.11.00)
```

Screenshot 2019-10-02 at 11 45 22

#### Environment
- Starship version: `starship 0.19.0`
- Shell type: `bash`
- Shell version:  `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)`
- Shell plugin manager: none
- Terminal emulator: Terminal

#### Possible Solution
If we want to support this Java variant (I see that Zulu isn't supported, for example) I think the `format_java_version` method in java.rs needs to detect from the version string that it's an OpenJ9 variant, and then parse the version directly following `Eclipse OpenJ9 OpenJDK 64-bit Server VM (`.

yuri1969:
#### Description
This PR tries to improve the version detection across [multiple Java VM vendors](https://dzone.com/articles/an-overview-on-jdk-vendors). The module captures both `STDOUT` and `STDERR` outputs of the `java -Xinternalversion` call. 

Eclipse OpenJ9, Azul Zulu, SapMachine, Amazon Corretto and GraalVM outputs are unit tested now.

The module uses a very basic `nom` parser in order to obtain the version string. _Sorry, if I got this whole idea wrong..._

#### Motivation and Context
Closes #468

#### Types of changes

#### Checklist:


src/modules/java_version_parser.rs
```diff
@@ -0,0 +1,61 @@
+use nom::{
```
matchai: I think a better location for this might be `src/modules/utils/java_verison_parser.rs`
What do you think?

yuri1969: Moved the parser to `modules/utils`.

src/modules/utils/java_version_parser.rs
```diff
@@ -0,0 +1,61 @@
...
+///     "JRE (Zulu 8.40.0.25-CA-linux64) (1.8.0_222-b10)"
+///     "VM (1.8.0_222-b10)".
+///
+/// Some Java vendors might not follow this format.
+pub fn parse_jre_version(input: &str) -> IResult<&str, &str> {
+    let prefix = alt((jre_prefix, j9_prefix));
+    let version_or_zulu = alt((version, zulu));
+    let (input, (_, version, _)) = tuple((prefix, version_or_zulu, suffix))(input)?;
+
+    Ok((input, version))
```
matchai: What's the reason for returning a tuple including the `input`?
It looks to me like we don't use it anywhere, and makes the output hard to reason with without reading through the implementation of `parse_jre_version`.

src/modules/java.rs
```diff
@@ -98,8 +100,43 @@ mod tests {
...
+        let java_11 = String::from("OpenJDK 64-Bit Server VM (11.0.4+11-LTS-sapmachine) for linux-amd64 JRE (11.0.4+11-LTS-sapmachine), built on Jul 17 2019 08:58:43 by \"\" with gcc 7.3.0");
+        assert_eq!(format_java_version(java_11), Some(String::from("v11.0.4")));
+    }
+
+    #[test]
+    fn test_format_java_version_unknown() {
+        let unknown_jre = String::from("Unknown JRE");
+        assert_eq!(format_java_version(unknown_jre), None);
     }
 }
```
matchai: Excellent tests!! 👍

src/modules/utils/java_version_parser.rs
yuri1969: You are right, the parser API should hide its internals. Now it doesn't expose the leftovers from `IResult`.

PR #509, opened 2019-10-08
refactor: Allow starship to be better used programmatically

Southclaws:
#### Description

This makes it easier to embed Starship into other Rust programs such as shells written in Rust. It also decouples the arguments from the context for more programmatic initialisation of the context.

#### Motivation and Context

Partially this issue: https://github.com/nushell/nushell/issues/356 and also just making it easier to use Starship as a Rust library for those building shells in Rust.

#### Types of changes

Kind of N/A I suppose since the change is internal and tiny.

#### Checklist:

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -1,5 +1,6 @@
 use clap::ArgMatches;
 use rayon::prelude::*;
+use std::fmt::Write as FmtWrite;
 use std::io::{self, Write};
 
 use crate::context::Context;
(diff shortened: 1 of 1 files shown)


jonathandturner:
Speaking from the nushell side, we're definitely interested in making these work together so that nushell users can use starship as their prompt.

matchai:
@all-contributors Please add @Southclaws for code.

PR #515, opened 2019-10-10
refactor: Refactor memory_usage module to use module config.

matiaskotlik:
### Description
This PR refactors the memory_usage module to use the new module config from #383 and also fixes a wide variety of bugs with the old memory_usage module, like:
- the percent sign not displaying correctly on some terminal emulators, including `kitty`
- changing the symbol in the configuration file didn't do anything
- swap being shown even if the system didn't have any (I'm not sure if this was intentional -- but if it was, the docs were very unclear about it)

#### Motivation and Context
Related #383
Also fixes bugs with the module

#### Types of changes

#### Checklist:


docs/config/README.md
```diff
@@ -688,7 +688,7 @@ To enable it, set `disabled` to `false` in your configuration file.
 | Variable          | Default                  | Description                                                   |
 | ----------------- | ------------------------ | ------------------------------------------------------------- |
 | `show_percentage` | `false`                  | Display memory usage as a percentage of the available memory. |
-| `show_swap`       | when total swap non-zero | Display swap usage.                                           |
+| `show_swap`       | `true`         | Display swap usage if total swap is non-zero.                           |
```
heyrict: Can you get this aligned?

matiaskotlik: just fixed it, nice catch!

src/modules/memory_usage.rs
```diff
@@ -1,90 +1,72 @@
...
-    const DEFAULT_SHOW_PERCENTAGE: bool = false;
-    const RAM_CHAR: &str = "🐏 ";
+use super::{Context, Module, RootModuleConfig};
 
-    let mut module = context.new_module("memory_usage");
+use crate::configs::memory_usage::MemoryConfig;
 
-    if module.config_value_bool("disabled").unwrap_or(true) {
-        return None;
-    }
```
heyrict: Currently we do not have a high-order approach yet to decide if this module is disabled or not. This logic is required as the module is disabled by default.

matiaskotlik: ok, I put it back in. I thought I had read somewhere that starship didn't load the module if it was `disabled`'d... maybe that was with the old module config

heyrict:
It seems that you have an outdated version of `rustc`, which results in failure in the rustfmt CI.

You may fix it by running `git checkout master tests` or re-run `rustfmt` after upgrading `rustc` to latest.

matiaskotlik:
> It seems that you have an outdated version of `rustc`, which results in failure in the rustfmt CI.
> 
> You may fix it by running `git checkout master tests` or re-run `rustfmt` after upgrading `rustc` to latest.

heyrict:
> the percent sign not displaying correctly on some terminal emulators

I'm not familiar with all terminal emulators, we need comments from other reviewers.

heyrict:
Hmm.. Why are you hurry to close this? Changes in this PR looks reasonable to me.

> It seems that you have an outdated version of `rustc`...

This is just a tiny issue and can be fixed in a minute. Feel free to ping me if you have trouble getting it fixed. @matiaskotlik

matiaskotlik:
i clicked the "close and comment" button instead of comment... they are too close together! :(

matiaskotlik:
> It seems that you have an outdated version of `rustc`, which results in failure in the rustfmt CI.
> 
> You may fix it by running `git checkout master tests` or re-run `rustfmt` after upgrading `rustc` to latest.

the updated rustfmt also changed some files in the `tests` directory, hopefully that is normal!

heyrict:
> the updated rustfmt also changed some files in the tests directory, hopefully that is normal!

It just reverted your previous changes to `tests` :D

Thanks for your contribution! Changes to the code looks good to me.

matiaskotlik:
> > the percent sign not displaying correctly on some terminal emulators
> 
> I'm not familiar with all terminal emulators, we need comments from other reviewers.

Xfce4-terminal, kitty, and urxvt are all working.

matiaskotlik:
Huh, the percent signs are broken in bash but they work in zsh... Might be something to do with the prompt format for zsh/bash. This is what shows up in zsh (commit 7ae4eac):  ` 43% | 7%` and bash: ` 43%% | 7%%`. Anyone know what could be causing this?

matiaskotlik:
Found the issue, zsh uses % in its $PROMPT variable for expansions. Had nothing to do with terminal emulator. This commit gets it working in all the supported shells now (bash, fish, and zsh). 
DeepinScreenshot_kitty_20191015180145

heyrict:
You can check what `battery.rs` handles it in https://github.com/starship/starship/blob/master/src/modules/battery.rs#L9-L13

matiaskotlik:
> You can check how `battery.rs` handles it in https://github.com/starship/starship/blob/master/src/modules/battery.rs#L9-L13

ooh, this is perfect, thank you!

src/modules/memory_usage.rs
```diff
@@ -1,90 +1,86 @@
...
+    // print escapes in a prompt context.
+    let shell = std::env::var("STARSHIP_SHELL").unwrap_or_default();
+    let percent_sign = match shell.as_str() {
+        "zsh" => "%%", // % is an escape in zsh, see PROMPT in `man zshmisc`
+        "powershell" => "`%",
+        _ => "%",
+    };
 
-    if module.config_value_bool("disabled").unwrap_or(true) {
+    if config.disabled {
```
matchai: This is automatically provided for all modules. No need to add the logic. 👍

keawade: I think this might actually still be needed. https://github.com/starship/starship/pull/515#discussion_r333781415

When I was working on #561 I removed this line and started seeing the module appearing even though it was supposed to be disabled by default.

heyrict: No, disabling is currently not provided for modules unless a disable option is specified in the config. This is required as `memory` is disabled by default.

PR #526, opened 2019-10-12
fix: Show leading slash when truncating from root

### LINKED ISSUE #458: Root slash of the path shouldn’t be cut off
Move into `/sys/fs/cgroup` and the first slash of the path shown is cut off, so instead just shown as `sys/fs/gcroup`. I think this one character helps much while not needing much space, so I’d like to extend the shown path to keep that for one further step, and then directly switch to `fs/cgroup/<next folder>`. What do you think?

keawade:
#### Description
Fixed a bug on unix systems where the leading `/` of the path was being removed when exactly `truncation_length` directories deep into a path.

The issue was that the method used to find the depth leveraged a `<string>.split('/')` which resulted in a vector of `["", "foo", "bar", "baz"]` instead of `["foo", "bar", "baz"]`. This resulted in the leading `/` being removed because the logic considered the user to be 4 directories deep when they were actually only 3 directories deep.

#### Motivation and Context
Closes #458 

#### Types of changes

#### Screenshots:
It works!
Screenshot from 2019-10-11 21-12-58

#### Checklist:
- [N/A] I have updated the documentation accordingly.
- [N/A] I have updated the tests accordingly.

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -142,7 +142,13 @@ fn truncate(dir_string: String, length: usize) -> String {
         return dir_string;
     }
 
-    let components = dir_string.split('/').collect::<Vec<&str>>();
+    let mut components = dir_string.split('/').collect::<Vec<&str>>();
+
+    // If the first element is "" then there was a leading "/" and we should remove it so we can check the actual count of components
+    if (components[0] == "") {
+        components.remove(0);
+    }
+
     if components.len() <= length {
         return dir_string;
     }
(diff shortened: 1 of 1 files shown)


keawade:
@matchai, I've added some tests. :+1: 

Should I also try to apply this change to home directory paths like `~/foo/bar/baz`? Currently they display as `foo/bar/baz`.

matchai:
> Should I also apply this change to home directory paths like ~/foo/bar/baz? 

I don't think it should receive the same treatment. `~` represents a set of directories, though I'd be interested in hearing others' opinions. Thoughts  @starship/astronauts?

PR #546, opened 2019-10-16
feat: add Haskell Stack support

m0nhawk:
#### Description
* Add a Haskell Stack module, when it detects `stack.yaml` file.

#### Motivation and Context

~Closes #~
I have a bunch of Haskell projects and I'm really missing support for Haskell in `starship`.

#### Types of changes

Screen Shot 2019-10-16 at 12 27 15 PM

#### Checklist:

- [ ] Add support for other Haskell: cabal, `.hs` files etc.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -99,6 +99,7 @@ prompt_order = [
     "git_commit",
     "git_state",
     "git_status",
+    "haskell",
     "hg_branch",
     "package",
     "dotnet",
@@ -109,6 +110,7 @@ prompt_order = [
     "python",
     "ruby",
     "rust",
+    "terraform",
     "terraform",
     "nix_shell",
     "conda",
@@ -610,6 +612,23 @@ The module will be shown if any of the following conditions are met:
 [golang]
 symbol = "🏎💨 "
 ```
+## Haskell
+
+The `haskell` module shows the currently installed version of Haskell Stack version.
+The module will be shown if any of the following conditions are met:
+
+- The current directory contains a `stack.yaml` file
+
+### Options
+
+| Variable   | Default       | Description                                               |
+| ---------- | ------------- | --------------------------------------------------------- |
+| `symbol`   | `"λ "`        | The symbol used before displaying the version of Haskell. |
+| `style`    | `"bold red"`  | The style for the module.                                 |
+| `disabled` | `false`       | Disables the `haskell` module.                            |
+
+[haskell]
+symbol = "λx.x "
 
 ## Mercurial Branch
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -111,7 +111,6 @@ prompt_order = [
     "ruby",
     "rust",
     "terraform",
-    "terraform",
     "nix_shell",
     "conda",
     "memory_usage",
@@ -627,8 +626,15 @@ The module will be shown if any of the following conditions are met:
 | `style`    | `"bold red"`  | The style for the module.                                 |
 | `disabled` | `false`       | Disables the `haskell` module.                            |
 
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
 [haskell]
 symbol = "λx.x "
+```
 
 ## Mercurial Branch
 
--- a/src/configs/haskell.rs
+++ b/src/configs/haskell.rs
@@ -1,0 +1,23 @@
+use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
+
+use ansi_term::{Color, Style};
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
+pub struct HaskellConfig<'a> {
+    pub symbol: SegmentConfig<'a>,
+    pub version: SegmentConfig<'a>,
+    pub style: Style,
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for HaskellConfig<'a> {
+    fn new() -> Self {
+        HaskellConfig {
+            symbol: SegmentConfig::new("λ "),
+            version: SegmentConfig::default(),
+            style: Color::Red.bold(),
+            disabled: false,
+        }
+    }
+}
--- a/src/modules/haskell.rs
+++ b/src/modules/haskell.rs
@@ -1,0 +1,54 @@
+use ansi_term::Color;
+use std::process::Command;
+
+use super::{Context, Module};
+
+/// Creates a module with the current Haskell Stack version
+///
+/// Will display the Haskell version if any of the following criteria are met:
+///     - Current directory contains a `stack.yaml` file
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let is_haskell_project = context
+        .try_begin_scan()?
+        .set_files(&["stack.yaml"])
+        .is_match();
+
+    if !is_haskell_project {
+        return None;
+    }
+
+    match get_haskell_version() {
+        Some(haskell_version) => {
+            const HASKELL_CHAR: &str = "λ ";
+
+            let mut module = context.new_module("haskell");
+            let module_style = module
+                .config_value_style("style")
+                .unwrap_or_else(|| Color::Red.bold());
+            module.set_style(module_style);
+
+            let formatted_version = format_haskell_version(&haskell_version)?;
+            module.new_segment("symbol", HASKELL_CHAR);
+            module.new_segment("version", &formatted_version);
+
+            Some(module)
+        }
+        None => None,
+    }
+}
+
+fn get_haskell_version() -> Option<String> {
+    match Command::new("stack").arg("ghc").arg("--").arg("--numeric-version").arg("--no-install-ghc").output() {
+        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
+        Err(_) => None,
+    }
+}
+
+fn format_haskell_version(haskell_version: &str) -> Option<String> {
+    let version = haskell_version;
+
+    let mut formatted_version = String::with_capacity(version.len() + 1);
+    formatted_version.push('v');
+    formatted_version.push_str(version);
+    Some(formatted_version)
+}
--- a/src/modules/haskell.rs
+++ b/src/modules/haskell.rs
@@ -17,7 +17,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let haskell_version = utils::exec_cmd("stack", &["ghc", "--", "--numeric-version", "--no-install-ghc"])?.stdout;
+    let haskell_version = utils::exec_cmd(
+        "stack",
+        &["ghc", "--", "--numeric-version", "--no-install-ghc"]
+    )?
+    .stdout;
     let formatted_version = format_haskell_version(&haskell_version)?;
 
     let mut module = context.new_module("haskell");
--- a/src/modules/haskell.rs
+++ b/src/modules/haskell.rs
@@ -15,12 +15,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .is_match();
 
     if !is_haskell_project {
-        return None;
-    }
-
-    match get_haskell_version() {
-        Some(haskell_version) => {
-            let formatted_version = format_haskell_version(&haskell_version)?;
+        None
+    } else {
+        get_haskell_version().map(|haskell_version| {
+            let formatted_version = format_haskell_version(&haskell_version).unwrap();
 
             let mut module = context.new_module("haskell");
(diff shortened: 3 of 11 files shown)


src/modules/haskell.rs
```diff
@@ -0,0 +1,60 @@
...
+/// Will display the Haskell version if any of the following criteria are met:
+///     - Current directory contains a `stack.yaml` file
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let is_haskell_project = context
+        .try_begin_scan()?
+        .set_files(&["stack.yaml"])
+        .is_match();
+
+    if !is_haskell_project {
+        return None;
```
bijancn: there is no need to `return` here or `match` on the haskell version. You can simply say something like
```
if !is_haskell_project {
  None
} else {
  get_haskell_version().map(|haskell_version| {
  ...
  module
})
}
```

src/configs/haskell.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub version: SegmentConfig<'a>,
+    pub style: Style,
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for HaskellConfig<'a> {
+    fn new() -> Self {
+        HaskellConfig {
+            symbol: SegmentConfig::new("λ "),
+            version: SegmentConfig::default(),
```
bijancn: I still haven't worked with the new config style but why is the `version` part of the `Config` when it's not configurable?

src/modules/haskell.rs
```diff
@@ -0,0 +1,60 @@
...
+            const HASKELL_CHAR: &str = "λ ";
+
+            let mut module = context.new_module("haskell");
+            let module_style = module
+                .config_value_style("style")
+                .unwrap_or_else(|| Color::Red.bold());
+            module.set_style(module_style);
+
+            let formatted_version = format_haskell_version(&haskell_version)?;
+            module.new_segment("symbol", HASKELL_CHAR);
```
bijancn: Is it really necessary to repeat all the default values from `src/configs/haskell.rs`?

src/modules/haskell.rs
```diff
@@ -0,0 +1,60 @@
...
+            module.new_segment("version", &formatted_version);
+
+            Some(module)
+        }
+        None => None,
+    }
+}
+
+fn get_haskell_version() -> Option<String> {
+    match Command::new("stack")
```
bijancn: I'm not too aware of `stack` and `ghc` but should you maybe check the yaml file first for a set compiler? https://docs.haskellstack.org/en/v1.0.2/yaml_configuration/#compiler Reading from file is usually faster than starting some other command

src/modules/haskell.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+
+fn format_haskell_version(haskell_version: &str) -> Option<String> {
+    let version = haskell_version;
+
+    let mut formatted_version = String::with_capacity(version.len() + 1);
+    formatted_version.push('v');
+    formatted_version.push_str(version);
+    Some(formatted_version)
```
bijancn: I don't think there is a need for a mutable String here. All you really want is to prepend a `v` here? You could format or simply add them. See also https://users.rust-lang.org/t/what-is-right-ways-to-concat-strings/3780

src/configs/haskell.rs
heyrict: See #472

src/modules/haskell.rs
heyrict: It seems that you have mixed the old config system with the new one. Check https://github.com/starship/starship/blob/master/src/modules/ruby.rs#L26-L31 about how you can write the code.

src/modules/haskell.rs
m0nhawk: The `compiler` is optional and can't be relied on.

I'm also planning to add support for `package.yaml` and `.cabal` -styled projects, then it'll make sense to first check files for compiler version and if it isn't available - run `stack` or `ghc`.

src/modules/haskell.rs
m0nhawk: @bijancn That makes perfect sense. 👍

src/modules/haskell.rs
```diff
@@ -0,0 +1,50 @@
...
+        .arg("--no-install-ghc")
+        .output()
+    {
+        Ok(output) => Some(String::from_utf8(output.stdout).unwrap()),
+        Err(_) => None,
+    }
+}
+
+fn format_haskell_version(haskell_version: &str) -> Option<String> {
+    let formatted_version = format!("v {}", haskell_version);
```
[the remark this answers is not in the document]
bijancn: Btw is the `v` even necessary?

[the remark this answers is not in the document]
m0nhawk: I saw that other versions does have `v`. Though, there is no space after it, so I'm removing it.

bijancn:
Somehow Ubuntu and Windows Tests are still failing?

m0nhawk:
> Somehow Ubuntu and Windows Tests are still failing?

Same Python version issue addressed in this branch: [update-python](https://github.com/starship/starship/tree/update-python).

m0nhawk:
Rebased on top of `master`.

m0nhawk:
@bijancn I've rebased to latest `master` and all the tests are green now.

bijancn:
Sorry @m0nhawk I lost track of this. Looks great to me but I dont think I can merge it due to permissions. @matchai can you help?

docs/config/README.md
```diff
@@ -105,6 +106,7 @@ prompt_order = [
     "ruby",
     "rust",
     "terraform",
+    "terraform",
```
matchai: It looks like this line was duplicated. 🤔 
```suggestion
```

docs/config/README.md
```diff
@@ -605,6 +607,23 @@ The module will be shown if any of the following conditions are met:
...
+- The current directory contains a `stack.yaml` file
+
+### Options
+
+| Variable   | Default       | Description                                               |
+| ---------- | ------------- | --------------------------------------------------------- |
+| `symbol`   | `"λ "`        | The symbol used before displaying the version of Haskell. |
+| `style`    | `"bold red"`  | The style for the module.                                 |
+| `disabled` | `false`       | Disables the `haskell` module.                            |
+
```
matchai: It appears you're missing the "Example" heading that other modules have:

```
### Example

```toml
# ~/.config/starship.toml

[haskell]
symbol = "🍕 "
```

src/modules/haskell.rs
```diff
@@ -0,0 +1,50 @@
...
+
+            module.create_segment("symbol", &config.symbol);
+            module.create_segment("version", &SegmentConfig::new(&formatted_version));
+            module
+        })
+    }
+}
+
+fn get_haskell_version() -> Option<String> {
+    match Command::new("stack")
```
matchai: The use of `Command::new(...)` is replaced with our `exec_command` utility function where possible:
https://github.com/starship/starship/blob/edc62f45188b742a4d20b601170ba8b9bdb5c803/src/modules/ruby.rs#L22

m0nhawk: Done. 👍

docs/config/README.md
m0nhawk: Missed a bunch of stuff when rebasing and fixing conflicts. 🤦‍♂

docs/config/README.md
m0nhawk: Removed. 👍

m0nhawk:
@matchai I have added [mstksg/setup-stack](https://github.com/mstksg/setup-stack), because this requires `stack.yaml`, not only the compiler - GHC. Also, I've added few tests.

m0nhawk:
Also, I needed to export `$HOME` in `render_module` and `render_prompty`, because Haskell Stack requires `HOME` variable.

It would be nice to point how to better make it. 😄

chipbuster:
@heyrict Would you mind marking the changes you requested as resolved? (Or I can dismiss them for you if you'd like). It looks like this'll be ready to merge pretty soon and those requested changes will block the merge.

chipbuster:
@m0nhawk You can use `.env()` on the command from within the test--you don't have to modify the file in `common`, e.g.
```rust
  render_module("haskell")
  .use_config("blah")
  .env("HOME", var1)
```

See the test `git_repo_in_home_directory_truncate_to_repo_true` on line 443 of the directory tests for an example of how that's used.

We've also been a little skittish about using the user's actual home directory because of a slew of bugs that have come up in testing before (usually our assumptions about the state of user's `$HOME` is incorrect, which causes the tests to fail even though the module is working), so our preferred method is to make a temporary directory and then use that as the "fake home" for the tests. You can also see that in the function mentioned above.

m0nhawk:
@chipbuster Thank you. I've added `env()` to test with HOME directory and have a question.

Alternatively, I can make use of [`STACK_ROOT`](1).

I've also changed the workflow to include the steps to setup Haskell Stack for all OSs, but it's not running (because it's a change from the fork?).

[1]: https://docs.haskellstack.org/en/stable/GUIDE/#setting-stack-root-location

chipbuster:
@m0nhawk Since the test using `HOME` is `#[ignore]`d, I don't think that'll be an issue (I didn't notice that yesterday).

> and have a question.

What's that?

> ...but it's not running 

I'll try to figure out what's going on with that.

m0nhawk:
> What's that?
Sorry, was not very clear. It is about using `STACK_ROOT`, but probably it would complicate Github Actions setup.

> I'll try to figure out what's going on with that.
I remember that changing workflow definition in fork would result in them not running, but the only thing I can find is: [here][1]. Not sure if that's the case.

[1]: https://github.community/t5/GitHub-Actions/Run-a-GitHub-action-on-pull-request-for-PR-opened-from-a-forked/m-p/31147#M690

chipbuster:
Were the tests running before your latest set of commits? I seem to recall that they were, but I'm not certain.

(If they were, then that suggests that GH got confused by something that you added to the workflow YAML)

m0nhawk:
Yes, it was running right before the workflow change.

chipbuster:
Hmmm. The file looks like valid YAML, but is there any possibility that it's somehow not valid for GH Actions? Could you try reverting the YAML change and see if the tests run?

Sorry, I'm not particularly a specialist in GH Actions (I'm more familiar with Gitlab CI) and matchai doesn't see anything obviously wrong either, so we might be taking stabs in the dark here.

m0nhawk:
I reverted the changes and now it's running. Not sure what's wrong with it.

chipbuster:
Well, debugging this will be interesting, as [GitHub doesn't seem to have a YAML validator yet](https://github.community/t5/GitHub-Actions/YAML-validator-for-Github-Actions-possible-expansion-of/td-p/29557) (and the only 3rd party one I can find hasn't been updated since Actions was in early beta).

I'll try my best to take some time to play around with this later today, maybe copy the branch into my fork and try opening a (temporary) PR on that. If you figure out what happened before then, please do ping me so we can get this merged!

m0nhawk:
I'll also check this. I've already enabled my fork Github Actions, so I'll check if it runs there.

m0nhawk:
In my fork it fails with:
```
a step cannot have both the uses and run keys
```

I'll continue testing with my fork now on. Thanks for help! 😄

m0nhawk:
@chipbuster I managed to make it run on everything, except Windows. Will try that later locally.

m0nhawk:
@chipbuster Windows tests have some strange issues with environment variable, there is only `STACK_ROOT` and `LOCALAPPDATA` variables which are required for `stack` to run, but it doesn't work as expected. I'm ignoring the test for Windows for now.

chipbuster:
@m0nhawk Does it work on Windows at the moment?

m0nhawk:
Yes, `cargo test` is running fine on Windows, it's only Github Actions failing.

chipbuster:
@matchai Are we okay with merging this without functioning Windows CI on the module? It seems that the module and command *do* work on Windows, but we can't get GH Actions working.

matchai:
Yep, that's fine by me.
We'll get around to moving them over to the mocked tests soon enough 👍

chipbuster:
Alright, let's get this PR merged then!

(I should really get back into Haskell someday)

chipbuster:
@all-contributors Please add @m0nhawk for code, doc, and tests.

@m0nhawk Thanks so much for all your work on this PR, and for being patient with us through a whole bunch of changes in the project structure! We really appreciate it :smile:

m0nhawk:
@chipbuster Thank you (and other) for help and discussion. I found starship to be really awesome for cross-platform prompt (as a long time user of spaceship, zsh and fish). And learning some Rust was nice. :)

And, of course, started with the missing feature I need.

I will try to fix it for Windows in Github Actions, it looks like some missing environmental variables, except the one I mentioned is needed.

PR #556, opened 2019-10-18
feat: Add configuration to set how much AWS profile info is shown

### LINKED ISSUE #539: aws module doesn't display profile name since update 0.24
#### Current Behavior

Hello 🙂, since the update 0.24 the aws module display the region instead of the profile name and the region is always show whereas before the aws module was only displayed when we export an aws profile.

```bash
15:43:12 in ~ on ☁️  eu-west-1
$
```

#### Expected Behavior

It should be something like that (only when we export the aws profile):

```bash
15:43:12 in ~
$ aws-prod
15:43:13 in ~ on ☁️  aws-prod
$
```

#### Environment
- Starship version: 0.24.0
- Shell type: fish
- Shell version: 3.0.2
- Terminal emulator:  iTerm

```bash
function aws-prod
set -g -x AWS_PROFILE 'aws-prod'
set -g -x AWS_REGION 'eu-west-1'
set -g -x AWS_DEFAULT_REGION 'eu-west-1'
end
```

#### Starship Configuration

```bash
[aws]
style = "bold fg:208"
```

Kwelity:
#### Description
This PR aims to fix #539 where a regression happened that would cause the aws module to only display the region.

This fixes it and also add the ability to chose which item to be displayed.

The new config field `displayed_items` accepts those values:
- `All`: this will display both Profile and region e.g aws-profile (eu-west-1)
- `Profile`: this will only display the Profile e.g aws-profile
- `Region`: this will only display the Region e.g eu-west-1

#### Motivation and Context

The feature to have the ability to chose which item to display was motivated by my need to limit information on small screen (i.e I don't really need to know which region)

Closes #539

#### Types of changes

Tested on linux (arch) with zsh.
Basic tests with manually setted environment variables and different configurations (`displayed_items`)

#### Checklist:

#### Concerns:
I wanted to point out that I'm by no mean an expert in rust (in programming in general for that matter).
It's probably not idiomatic rust code.

I'm also not really satisfied with how I parse the enum variant from String.

This bit:

```rust
impl<'a> ModuleConfig<'a> for AwsItems {
    fn from_config(config: &Value) -> Option<AwsItems> {
        match config {
            Value::String(s) if s == "All" => Some(AwsItems::All),
            Value::String(s) if s == "Region" => Some(AwsItems::Region),
            Value::String(s) if s == "Profile" => Some(AwsItems::Profile),
            _ => None,
        }
    }
}
```

Finally I haven't really look into what get_aws_region_from_config does, I tried to implement my fix with the same logic from previous version. But I might have gotten something wrong.

Edit: A potential fix could simply be removing the feature of selecting which item to display.

Please let me know what I could improve.


Kwelity:
Ok, I've just figured out how tests works :/

I'll take a look at them and try to fix them.

docs/config/README.md
```diff
@@ -120,11 +120,12 @@ The `aws` module shows the current AWS region and profile. This is based on
...
-| ---------- | --------------- | ---------------------------------------------------------- |
-| `symbol`   | `"☁️  "`        | The symbol used before displaying the current AWS profile. |
-| `style`    | `"bold yellow"` | The style for the module.                                  |
-| `disabled` | `false`         | Disables the `AWS` module.                                 |
+| Variable          | Default         | Description                                                |
+| ----------        | --------------- | ---------------------------------------------------------- |
+| `symbol`          | `"☁️  "`         | The symbol used before displaying the current AWS profile. |
+| `style`           | `"bold yellow"` | The style for the module.                                  |
+| `disabled`        | `false`         | Disables the `AWS` module.                                 |
+| `displayed_items` | `All`           | Chose which item to display. Possible values [`All`, `Profile`, `Region`] |
```
matchai: For consistency with other string configuration options, like `style`, I think it would be best if the values were lowercase.
```suggestion
| `displayed_items` | `All`           | Chose which item to display. Possible values [`all`, `profile`, `region`] |
```

matchai: Typo fix.

```suggestion
| `displayed_items` | `All`           | Choose which item to display. Possible values [`All`, `Profile`, `Region`] |
```

src/modules/aws.rs
```diff
@@ -44,42 +44,34 @@ fn get_aws_region_from_config(aws_profile: &Option<String>) -> Option<String> {
...
-                    aws_region.unwrap_or_default(),
-                ))
-            }
-        })
-        .or_else(|| {
-            env::var("AWS_REGION")
-                .ok()
-                .map(|region| (String::new(), region))
-        })
+fn get_aws_region() -> (String, String) {
```
matchai: I think it can be a little clearer that this function returns a tuple of `(profile, region)`, either by using named values or with a comment.

src/modules/aws.rs
```diff
@@ -44,42 +44,34 @@ fn get_aws_region_from_config(aws_profile: &Option<String>) -> Option<String> {
...
+    match (
+        env::var("AWS_PROFILE").ok(),
```
matchai: Not a big deal, but right now it looks like we get environment variables that may or may not be used.
Would it be possible to structure it in a way that we check `config.displayed_items` before retrieving the necessary environment variables?

docs/config/README.md
Kwelity: Done, it might be possible to convert the value to lowercase at run time to make it case insensitive. For now, I'll just make it lowercase.

Kwelity: Fixed.

src/modules/aws.rs
Kwelity: I used alias typed (which I never used before).
Was it what you meant?

As for commenting the code, I need to take a look at the rest of the project to have an hint about the "style" used.

src/modules/aws.rs
Kwelity: Good idea. I tried to do just that.

I still do an unnecessarily call to AWS_PROFILE if AWS_DEFAULT_PROFILE is set. But it meant a bigger refacto.

Kwelity:
I fixed the old tests (by modifying the logic of my code to match the previous one).

I still haven't added test for the new feature as well as test that check the regression.

Bonus would be to add doc comment.

I'll try to tackle all that during the weekend.

Kwelity:
I've added some test for the new configuration field displayed_items.

One test that I'm not sure is `region_set_and_only_display_profile`

currently if no profile is set but with a region and a filter on profile. We'll show just the cloud icon.

To me, this make sens because we still in a aws 'folder' but we might want to show nothing...

src/configs/aws.rs
```diff
@@ -20,6 +35,18 @@ impl<'a> RootModuleConfig<'a> for AwsConfig<'a> {
...
+    }
+}
+
+impl<'a> ModuleConfig<'a> for AwsItems {
+    fn from_config(config: &Value) -> Option<AwsItems> {
+        match config {
+            Value::String(s) if s == "all" => Some(AwsItems::All),
+            Value::String(s) if s == "region" => Some(AwsItems::Region),
+            Value::String(s) if s == "profile" => Some(AwsItems::Profile),
+            _ => None,
```
heyrict: Check #383, I've put an example for parsing a enum value under the "Specification" header.

src/configs/aws.rs
```diff
@@ -2,6 +2,20 @@ use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
...
 use ansi_term::{Color, Style};
 use starship_module_config_derive::ModuleConfig;
+use toml::Value;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum AwsItems {
+    All,
+    Region,
+    Profile,
+}
```
heyrict: :+1: Good use case of enum values!

src/configs/aws.rs
Kwelity: Thanks for the heads-up it's actually a lot better.

andytom:
> I've added some test for the new configuration field displayed_items.
> 
> One test that I'm not sure is `region_set_and_only_display_profile`
> 
> currently if no profile is set but with a region and a filter on profile. We'll show just the cloud icon.
> 
> To me, this make sens because we still in a aws 'folder' but we might want to show nothing...

I'm not sure about this, there is no test if a folder is an aws folder or there wasn't when I initially added this module, the AWS module was only enabled if the `AWS_PROFILE` env var was set. So unless anything has changed this empty cloud will always be displayed. I'm not sure what the expected behaviour is here.

Kwelity:
> I'm not sure about this, there is no test if a folder is an aws folder or there wasn't when I initially added this module, the AWS module was only enabled if the `AWS_PROFILE` env var was set. So unless anything has changed this empty cloud will always be displayed. I'm not sure what the expected behaviour is here.

Ha, it makes sense. My workflow is a bit different, I dynamically load environment variable per directory. It allows me to have a "dedicated" aws folder.

andytom:
> > I'm not sure about this, there is no test if a folder is an aws folder or there wasn't when I initially added this module, the AWS module was only enabled if the `AWS_PROFILE` env var was set. So unless anything has changed this empty cloud will always be displayed. I'm not sure what the expected behaviour is here.
> 
> Ha, it makes sense. My workflow is a bit different, I dynamically load environment variable per directory. It allows me to have a "dedicated" aws folder.

Let's ask for some feedback on the issue from the users to see if we can workout what workflows different users use so we can work out what we need to support.

Kwelity:
@andytom after merging with master and testing a bit more, I found out some bugs. Maybe that was the issue you were referring to.

I've tweaked the code, returning `None` instead of `unwrap_or_default()` which is IMO better.

The tweak means that we won't show the cloud if the "correct" variable is not set.

So if we set `displayed_items` to region and AWS_DEFAULT_REGION or AWS_REGION is not set, nothing will be printed.

I've added a test for this example.

We should still await for feedback, but this tweaks make the code simpler.

tests/testsuite/aws.rs
```diff
@@ -4,12 +4,11 @@ use std::io::{self, Write};
 use ansi_term::Color;
```
matchai: I think it would be best if we added tests that explicitly tested the `displayed_items = "all"` configuration, in case we ever change the default behavior. 🙂

Kwelity: Fair point, I've added the tests.

PR #470, opened 2019-10-02
feat: Implement PowerShell support

davidkna:
#### Description

Now that #246 is closed I've taken another stab at implementing support for powershell.
Powershell is a cross-platform shell made by Microsoft.

The prompt loads and everything seems to work fine so far. If there's anything specific I should look out for please tell me.

#### Motivation and Context

Closes #126

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [x] I have tested using **Windows**
I haven't been able to test my battery status change yet, because I don't have a device handy for that.

#### Checklist:

- [x] I have updated the documentation accordingly.
I believe I changed the necessary parts. The Japanese part still needs updating.
- [ ] I have updated the tests accordingly.
I haven't found any test infrastructure for specific shells thus I haven't added any tests yet.

PR #562, opened 2019-10-19
feat: Add nodejs_npm and nodejs_yarn modules

### LINKED ISSUE #557: NPM Version next to Node Version

As a JavaScript developer I am never aware of a time when I would be using Node and not also using NPM. In addition, the version of NPM being used can have an impact on the way packages are resolved, lockfiles written, etc. It is also useful to know the NPM version when comparing my work environment to that of my coworkers/collaborators.

Include the NPM version next to the node version in the prompt when node version is displayed. This could be an optional mode activated via configuration.

The default approach for this would, presumably, be to leverage `npm -v` or `npm --version`. For broader support of the display of NPM libraries it might also be possible to leverage the fact that these libraries are all stored in the same global folder for each node installation. A user could configure an array of libraries they wish to see the version of (I could see arguments for wanting to see the global version of `grunt` or `create-react-app`, as examples) and starship could cache the versions in each `package.json` file for the globally installed libraries.

This approach would cover the NPM version use case since NPM is a globally installed package, and in that way it is a lot more flexible than just displaying the NPM version. However, it is harder to implement as a maintainer and harder to configure as a user. Among other potential concerns, one of the most obvious to me is that there would be difficulty when it came to specifying the colors/icons of multiple different packages such that they were visibly distinct in the prompt.

keawade:
#### Description
Adds modules for the two most used Node/JS package managers, [`npm`](https://www.npmjs.com/get-npm) and [`yarn`](https://yarnpkg.com).

Both of these modules are disabled by default.

#### Motivation and Context
From #557:
> As a JavaScript developer I am never aware of a time when I would be using Node and not also using NPM. In addition, the version of NPM being used can have an impact on the way packages are resolved, lockfiles written, etc. It is also useful to know the NPM version when comparing my work environment to that of my coworkers/collaborators.

Closes #557 

#### Types of changes

Screenshot from 2019-10-18 22-51-29

I've added acceptance tests to make sure both modules work as expected.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -95,6 +95,8 @@ prompt_order = [
     "golang",
     "java",
     "nodejs",
+    "nodejs_npm",
+    "nodejs_yarn",
     "python",
     "ruby",
     "rust",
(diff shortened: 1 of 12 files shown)


keawade:
@davidkna, I seem to be running into a similar issue testing the `npm` and `yarn` modules as you ran into with the `nodejs` module while implementing #470 but just adding the `SYSTEMROOT` wrapper didn't resolve the issue.

Do you have any suggestions as to what I could do to resolve this?

davidkna:
The only thing that's really different between a real environment and the test is that `render_module` clears env - which is why I suspect this is an issue here too.
I don't know which variables those could be. Maybe passing through all the usual default environment variables on windows (with defaults?) is an option. I only found `SYSTEMROOT` through trial and error.

keawade:
I think the testing changes introduced earlier this week will unblock these tests. I'll take a look when I've got time.

matchai:
Unfortunately the prompt is already very long in Node.js projects with `package`, `node`, and whichever other modules are already being used in the project. I don't think npm and yarn versions provide enough actionable value to justify adding to the prompt.

I'm closing this PR for the time being. If you would like to further discuss their value in the prompt, feel free to open an issue.

PR #563, opened 2019-10-19
fix: Fix the python module ignoring error codes

### LINKED ISSUE #542: Python module should use error code
#### Current Behavior
```
~/foo is 📦 v0.1.0 via 🐍 vasdf: No version set for command python
you might want to add one of the following in your .tool-versions file:

python 3.7.4 ❯
```
#### Expected Behavior
Starfish should treat the output from `python` (a shim provided by asdf) as an error instead of as a version string:

https://github.com/starship/starship/blob/master/src/modules/python.rs#L83-L97

Also, `.tool-versions` should be checked in addition to `.python-version`.

#### Environment
- Starship version: 0.24.0
- Shell type: fish 3.0.2
- Terminal emulator: iTerm
- Operating system: MacOS 10.15

andytom:
#### Description

This is a quick fix to stop the python module from displaying error
messages that have been printed to stderr as the version.

#### Motivation and Context

Closes #542

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -68,6 +68,13 @@ fn get_pyenv_version() -> Option<String> {
 fn get_python_version() -> Option<String> {
     match Command::new("python").arg("--version").output() {
         Ok(output) => {
+            if !output.status.success() {
+                log::warn!(
+                    "Non-Zero exit code '{}' when executing `python --version`",
+                    output.status
+                );
+                return None;
+            }
             // We have to check both stdout and stderr since for Python versions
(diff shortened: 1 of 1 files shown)


andytom:
Side note, long term it might be worth writing a little helper function for executing commands, we seem to have a similar use case in a lot of modules maybe it would be nice to have a common way of doing this similar to the [`read_file` util](https://github.com/starship/starship/blob/master/src/utils.rs#L6-L12)?

yuri1969:
IMHO, it might be useful to add a `log::warn!` there since encountering an error code is kind of unexpected.

PR #569, opened 2019-10-20
feat: Add the hg_branch module

rpl:
#### Description

Adds an hg_branch module (which displays the current active bookmark or branch while the current directory is detected as a mercurial repository).

The module is marked as disabled by default.

#### Motivation and Context

Mercurial is not as common as git, but it may be handy for users that work on mercurial repositories
(and personally I couldn't migrate to starship as my default prompt without it ;-))

#### Types of changes

- [X] ~~Bug fix (non-breaking change which fixes an issue)~~
- [X] ~~Breaking change (fix or feature that would cause existing functionality to not work as expected)~~

image

#### Checklist:


src/modules/hg_branch.rs
```diff
@@ -0,0 +1,90 @@
...
+            "\"truncation_length\" should be a positive value, found {}",
+            config.truncation_length
+        );
+        std::usize::MAX
+    } else {
+        config.truncation_length as usize
+    };
+
+    let mut branch_name = match get_hg_log_template(context, "{activebookmark}") {
+        None => get_hg_log_template(context, "{branch}").unwrap(),
```
matchai: Could we replace the `.unwrap()` with an early return of `None`?
It would be best to simply not show the module than to panic.

src/modules/hg_branch.rs
```diff
@@ -0,0 +1,90 @@
...
+        Some(bookmark) => bookmark,
+    };
+    if branch_name == "" {
+        branch_name = get_hg_log_template(context, "{branch}").unwrap();
```
matchai: Similarly here. ☝️

src/modules/hg_branch.rs
rpl: yeah, `.unwrap()` was quite rude ;-)

How about the updated version (which is using `.unwrap_or_default()`)?
I also added a couple of additional tests:
- if the current working dir isn't recognized as an hg repo, the module shouldn't produce any output at all
- if it is an hg repo, but the module fails to retrieve a bookmark or branch name (e.g. because it fails to execute hg, not unlikely if it is not installed), it defaults to "(no branch)" (which let the user know that, even if an hg repo has been detected, something went wrong and so the branch name is missing)

How that sounds/looks to you?

rpl:
> Sorry for the lengthy wait to have this PR reviewed. sweat_smile
> A couple small comments were added below.

No worries at all, I wasn't in rush (I was already able to use the same super fast prompt across all my machines, included my phone... I can only say "thank you so much for that" :-))

I really appreciate the review, though! 
(I find "contributing back" an important part of my "learning process", this is my first time contributing something written in Rust :-D)

rpl:
On the last version the workflow seems to be failing on one of the windows steps:
- https://github.com/starship/starship/runs/327161160#step:8:26 

but it is failing on the setup-dotnet action, which shouldn't actually be affected by the changes in this pull request.

src/modules/hg_branch.rs
matchai: This is certainly better, but I think the more semantic solution here is to use `None` instead of an empty string.

I think something along these lines would 
```rust
let get_branch_name = |tmpl| get_hg_log_template(tmpl, context, &config);

let mut branch_name = get_branch_name("{activebookmark}");

if branch_name.is_none() {
    branch_name = get_branch_name("{branch}");
}

if branch_name.is_none() {
    branch_name = "(no branch)".to_string();
}
```

Perhaps a simpler way of addressing this without the `if`s would be to use some `Option` methods:
```rust
// I haven't ran this code, so I don't know if it compiles
let get_branch_name = |tmpl| get_hg_log_template(tmpl, context, &config);

let mut branch_name = get_branch_name("{activebookmark}")
	.or_else(|| get_branch_name("{branch}"))
	.unwrap_or("(no branch)".to_string());
```

rpl: yeah, you are right, I just had to turn an empty string output into None on the `get_hg_log_template` side  (because the '{activebookmark}' output could be an empty string and it would be good to fallback to the branch name in that case)

Thanks! This is so much nicer.

src/modules/hg_branch.rs
```diff
@@ -0,0 +1,95 @@
...
+    module.create_segment(
+        "name",
+        &config.branch_name.with_value(&truncated_and_symbol),
+    );
+
+    Some(module)
+}
+
+fn get_hg_log_template(hgtmpl: &str, ctx: &Context, cfg: &HgBranchConfig) -> Option<String> {
+    let output = Command::new(cfg.hg_binary)
```
matchai: A more concise way of handling this might be to early return on `None` (also avoiding the `.clone()` for `ctx.current_dir`):
```rust
    let output = Command::new(cfg.hg_binary)
        .args(&["log", "-r", ".", "--template", hgtmpl])
        .current_dir(&ctx.current_dir)
        .output()
        .ok()
        .and_then(|output| String::from_utf8(output.stdout).ok())?;

    if output.is_empty() {
        None
    } else {
        Some(output)
    }
```

rpl: :+1: a small touch that makes it looks way better (I should have thought of returning earlier on None using `?`, thanks for pointing it out!)

src/configs/hg_branch.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#[derive(Clone, ModuleConfig)]
+pub struct HgBranchConfig<'a> {
+    pub symbol: SegmentConfig<'a>,
+    pub truncation_length: i64,
+    pub truncation_symbol: &'a str,
+    pub branch_name: SegmentConfig<'a>,
+    pub style: Style,
+    pub disabled: bool,
+    pub hg_binary: &'a str,
```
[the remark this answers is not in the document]
rpl: I'm pretty sure that none that use mercurial would have to use that configuration option.

I have to confess that I added it mainly to be able to easily test that the module does fallback gracefully when all the mercurial commands fails to run successfully.

I was already unhappy (also a bit "guilty" :-)) about such an hacky testing strategy and so I took the opportunity to think to something else that would be less invasive (as it doesn't need to customize the hg binary name).

In the updated version the single test case that was using this config option is now using a fake "broken hg repo" (and I removed any trace of this unnecessary config option).

matchai:
@all-contributors Please add @rpl for code, docs, and tests.

PR #584, opened 2019-10-23
feat: Enable the python module in virtual envs

andytom:
#### Description

This will enable the python module when a virtual environment has been
activated, this is detected via the `VIRTUAL_ENV` env var.

#### Motivation and Context

Since the python module displays the currently active virtual env it
makes sense to activate it when a virtual env is active.

#### Types of changes

#### Checklist:


andytom:
> From the test I'm concluding the virtual env was already working, it just had to be enabled?

Basically yes, the module was already displaying the virtualenv if the current dir had one of the python files in it but since we can detect if a virtualenv is active from an env var it makes sense to enable the python module when one is active.

PR #598, opened 2019-10-28
feat: Add stash count to git_status module

### LINKED ISSUE #595: Git Status: Stash Count
Ability to see the number of stashes in a git folder. First suggested here https://github.com/starship/starship/issues/423#issuecomment-534896545 by @etiennemabille.

New option `stash_count` with enabled and style attributes. When enabled shows a count after the stash symbol. e.g. `$3`

marblenix:
Added stash count to the git status module.

#### Description
The git_module has been updated to add the number of stashes for the current repo after the stash symbol ("$"). Documentation has been updated accordingly.

#### Motivation and Context
Closes #595

#### Types of changes

image

Tested on mac by running new unit tests and running the docker test. Also built a release build and manually tested that.

#### Checklist:

---

I'm new to rust so please correct me if I made any obvious mistakes! I'll update the PR as soon as I can.

### DIFF
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -4,6 +4,7 @@ use super::{Context, Module, RootModuleConfig};
 
 use crate::config::SegmentConfig;
 use crate::configs::git_status::{CountConfig, GitStatusConfig};
+use std::borrow::BorrowMut;
 
 /// Creates a module with the Git branch in the current directory
 ///
@@ -23,7 +24,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let repo = context.get_repo().ok()?;
     let branch_name = repo.branch.as_ref()?;
     let repo_root = repo.root.as_ref()?;
-    let repository = Repository::open(repo_root).ok()?;
+    let mut repository = Repository::open(repo_root).ok()?;
 
     let mut module = context.new_module("git_status");
     let config: GitStatusConfig = GitStatusConfig::try_load(module.config);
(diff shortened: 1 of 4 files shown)


src/modules/git_status.rs
```diff
@@ -52,9 +56,6 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         log::trace!("No stash object found");
     }
 
-    let repo_status = get_repo_status(&repository);
-    log::debug!("Repo status: {:?}", repo_status);
-
```
marblenix: I had to move this up to the first part in the file because of the immutable borrows above it. Please let me know if there's a better way to go about this.

src/modules/git_status.rs
```diff
@@ -114,7 +115,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     // Add the stashed segment
     if stash_object.is_ok() {
-        module.create_segment("stashed", &config.stashed);
```
brianlow: Do we still need the `stash_object` variable (used one line above)?

tests/testsuite/git_status.rs
```diff
@@ -259,33 +259,54 @@ fn doesnt_show_untracked_file_if_disabled() -> io::Result<()> {
 }
 
 #[test]
-#[ignore]
```
brianlow: Keep the `[ignore]` and add to new tests according to this comment: https://github.com/starship/starship/pull/434#discussion_r329673883

These tests can be run specifically with `cargo test git_status -- --ignored`

src/modules/git_status.rs
marblenix: It doesn't appear so. Without it all tests pass and the stash module seems unaffected overall.

tests/testsuite/git_status.rs
marblenix: Updated with `#[ignore]` back in place.

For the new tests, were there any edge cases you'd like me to explore further than what I did cover?

brianlow: Tests look good 👍🏼

src/modules/git_status.rs
```diff
@@ -244,6 +247,17 @@ fn is_untracked(status: Status) -> bool {
...
 }
 
+fn stashed_count(repository: &mut Repository) -> usize {
+    let mut count = 0;
+    repository
+        .stash_foreach(|_, _, _| {
+            count += 1;
+            true
+        })
+        .unwrap();
```
matchai: This `.unwrap()` could be a little dangerous here.
Ideally we would return an `Option` or `Result`, and early return the module as `None` if this is an `Error`.

matchai:
@all-contributors Please add @marblenix for code and test. 🎉

PR #602, opened 2019-10-28
feat: Add singularity module

ntadej:
#### Description

Singularity is one of the tools to run containers. A popular use case is to run it as a development environment, especially for older platforms.

#### Motivation and Context

As one might use several different images it is useful to display the image we are currently in.

Currently the default display is after hostname wrapped in `[` and `]` as it is the shortest. I'm open to suggestions if another defaults are more reasonable.

#### Types of changes

Screenshot 2019-10-28 at 23 26 39

#### Checklist:

### DIFF
--- a/src/configs/singularity.rs
+++ b/src/configs/singularity.rs
@@ -1,0 +1,27 @@
+use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
+
+use ansi_term::{Color, Style};
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
+pub struct SingularityConfig<'a> {
+    pub symbol: Option<SegmentConfig<'a>>,
+    pub label: &'a str,
+    pub prefix: &'a str,
+    pub suffix: &'a str,
+    pub style: Style,
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for SingularityConfig<'a> {
+    fn new() -> Self {
+        SingularityConfig {
+            symbol: None,
+            label: "",
+            prefix: "[",
+            suffix: "]",
+            style: Color::Blue.bold().dimmed(),
+            disabled: true,
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


src/configs/singularity.rs
```diff
@@ -0,0 +1,27 @@
+use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
+
+use ansi_term::{Color, Style};
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
+pub struct SingularityConfig<'a> {
+    pub symbol: Option<SegmentConfig<'a>>,
```
heyrict: I think using `symbol: SegmentConfig` with `""` as default value should be good enough, as long as you don't render the prompt conditionally.

src/configs/singularity.rs
```diff
@@ -0,0 +1,27 @@
...
+
+impl<'a> RootModuleConfig<'a> for SingularityConfig<'a> {
+    fn new() -> Self {
+        SingularityConfig {
+            symbol: None,
+            label: "",
+            prefix: "[",
+            suffix: "]",
+            style: Color::Blue.bold().dimmed(),
+            disabled: true,
```
heyrict: You need to explicitly add code to `modules/singularity.rs` to disable the module. Currently there is no high order approach for starship to judge the disabled option when the module is disabled as a default.

matchai:
Closing this PR as stale.
Feel free to reopen it when you get around to addressing the review comments.

ntadej:
Sorry for delaying this I rebased and fixed the comments. Let's see if I can figure out how to open this PR.

src/configs/singularity.rs
ntadej: Done.

src/configs/singularity.rs
ntadej: Done.

matchai:
Unfortunately, I no longer can since the branch was force-pushed or recreated. 😅 
Feel free to open a new PR.

PR #603, opened 2019-10-29
feat: Add separator config to the memory module

matiaskotlik:
#### Description
Improves memory module by adding a config option to change the seperator. The old seperator was `" | "`, now that is the default but it you can change it. I also added `SegmentConfig`'s for the ram and swap, mostly out of necessity, but now you can customize the style of those individually. I've left that undocumented because most of the other modules don't document stuff like that either.

#### Motivation and Context

I use the memory_usage module a lot and it was annoying me that the seperator took up 3 characters of the prompt. My config now has the line: `seperator = "/"` :).

#### Types of changes

DeepinScreenshot_select-area_20191028201602

#### Checklist:


docs/config/README.md
```diff
@@ -713,6 +713,7 @@ To enable it, set `disabled` to `false` in your configuration file.
 | `show_swap`       | `true`                | Display swap usage if total swap is non-zero.                 |
 | `threshold`       | `75`                  | Hide the memory usage unless it exceeds this percentage.      |
 | `symbol`          | `"🐏 "`               | The symbol used before displaying the memory usage.           |
+| `seperator`       | `" | "`               | The symbol or text that will seperate the ram and swap usage. |
```
r15ch13: Isn't it written `separator`? https://duckduckgo.com/?q=seperator 😃

matiaskotlik: wow, can't believe I spelled it wrong there and in the actual module code! Good catch :)

matchai:
@all-contributors Please add @matiaskotlik for code.

PR #611, opened 2019-10-31
Add Dart module

### LINKED ISSUE #609: Dart & Flutter support
Currently, there is no support for [Dart](https://dart.dev) and [Flutter](https://flutter.dev) development.

I would like to build support for [Dart](https://dart.dev) and [Flutter](https://flutter.dev). I'm completely new to Starship and Rust, so some general direction would be really welcome!

Alternatively, someone else could build support [Dart](https://dart.dev) and [Flutter](https://flutter.dev).

HofmannZ:
This is my first time writing Rust, so please be thorough with the code review.

#### Description

This PR adds a module that shows the current version of Dart.

#### Motivation and Context

Closes #609.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/dart.rs
+++ b/src/modules/dart.rs
@@ -20,21 +20,16 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    match get_dart_version() {
-        Some(dart_version) => {
-            let mut module = context.new_module("dartlang");
-            let config: DartConfig = DartConfig::try_load(module.config);
+    let formatted_version = format_dart_version(&get_dart_version()?)?;
 
-            module.set_style(config.style);
-            module.create_segment("symbol", &config.symbol);
+    let mut module = context.new_module("dart");
+    let config: DartConfig = DartConfig::try_load(module.config);
 
-            let formatted_version = format_dart_version(&dart_version)?;
-            module.create_segment("version", &config.version.with_value(&formatted_version));
+    module.set_style(config.style);
+    module.create_segment("symbol", &config.symbol);
+    module.create_segment("version", &config.version.with_value(&formatted_version));
 
-            Some(module)
-        }
-        None => None,
-    }
+    Some(module)
 }
 
 fn get_dart_version() -> Option<String> {
--- a/src/modules/dart.rs
+++ b/src/modules/dart.rs
@@ -52,12 +52,7 @@ fn format_dart_version(dart_stdout: &str) -> Option<String> {
 
     let version = dart_stdout.split(' ').nth(3)?;
 
-    let mut formatted_version = String::with_capacity(version.len() + 1);
-
-    formatted_version.push('v');
-    formatted_version.push_str(version);
-
-    Some(formatted_version)
+    Some(format!("v{}", version))
 }
 
 #[cfg(test)]
--- a/src/modules/dart.rs
+++ b/src/modules/dart.rs
@@ -33,20 +33,17 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn get_dart_version() -> Option<String> {
-    match Command::new("dart").arg("--version").output() {
-        Ok(output) => {
-            if !output.status.success() {
-                log::warn!(
-                    "Non-Zero exit code '{}' when executing `dart --version`",
-                    output.status
-                );
-                return None;
-            }
+    let output = Command::new("dart").arg("--version").output().ok()?;
 
-            Some(String::from_utf8(output.stderr).unwrap())
-        }
-        Err(_) => None,
+    if !output.status.success() {
+        log::warn!(
+            "Non-Zero exit code '{}' when executing `dart --version`",
+            output.status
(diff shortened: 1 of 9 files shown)


HofmannZ:
It seems like I'm still missing the `Run actions/setup-dart@v1` step.

src/modules/dart.rs
```diff
@@ -0,0 +1,80 @@
...
+                return None;
+            }
+
+            Some(String::from_utf8(output.stderr).unwrap())
+        }
+        Err(_) => None,
+    }
+}
+
+fn format_dart_version(dart_stdout: &str) -> Option<String> {
```
matiaskotlik: this function could look like this:
```rust
fn format_dart_version(dart_stdout: &str) -> Option<String> {
    // `dart --version` output will look somthing like this:
    // Dart VM version: 2.5.2 (Tue Apr 20 04:20:00 2019 +0200) on "macos_x64"

    let version = dart_stdout.split(' ').nth(3)?;

    Some(format!("v{}", version))
}
```
I think this way might be better than creating a new mutable String and appending to it.

src/modules/dart.rs
```diff
@@ -0,0 +1,80 @@
...
+            let formatted_version = format_dart_version(&dart_version)?;
+            module.create_segment("version", &config.version.with_value(&formatted_version));
+
+            Some(module)
+        }
+        None => None,
+    }
+}
+
+fn get_dart_version() -> Option<String> {
```
matiaskotlik: This function can also be simplified a bit, the `match` is not necessary here.
```rust
fn get_dart_version() -> Option<String> {
    let output = Command::new("dart").arg("--version").output().ok()?;
    if !output.status.success() {
        log::warn!(
            "Non-Zero exit code '{}' when executing `dart --version`",
            output.status
        );
        return None;
    }

    String::from_utf8(output.stderr).ok()
}
``` 
Another notable change is propagating None instead of panicing when the utf-8 decode fails, so that starship won't break if dart is really screwed up.

src/modules/dart.rs
```diff
@@ -0,0 +1,80 @@
...
+        .try_begin_scan()?
+        .set_files(&["pubspec.yaml"])
+        .set_extensions(&["dart"])
+        .is_match();
+
+    if !is_dart_project {
+        return None;
+    }
+
+    match get_dart_version() {
```
matiaskotlik: the match is not needed here, you can just use `?` to propagate the errors instead of using match
```rust
let formatted_version = format_dart_version(&get_dart_version()?)?;

    let mut module = context.new_module("dartlang");
    let config: DartConfig = DartConfig::try_load(module.config);

    module.create_segment("symbol", &config.symbol);

    module.create_segment("version", &config.version.with_value(&formatted_version));
```
    Some(module)
```

HofmannZ: Thanks for the tip! Will refactor that.

src/modules/dart.rs
HofmannZ: Thanks for the feedback! Just wondering how would `String::from_utf8(output.stderr).ok()` still propagate `None`? (I'm completely new, to Rust so probably it's newbie question.)

src/modules/dart.rs
HofmannZ: I totally stole that from the [Go module](https://github.com/starship/starship/blob/f2cb529d1bd44f996ad60c2d27ae14d1f9ad44ba/src/modules/golang.rs#L54-L69). Should we change that there too?

matiaskotlik: I think this way makes a lot more sense but ultimately they both work.

Even though it's not the biggest of issues, I think I'll still open a PR for it. This way no one will copy that code again :)

src/modules/dart.rs
matiaskotlik: whoops, I meant return. It's not propagating here, you are right :)

johnletey:
@HofmannZ I was looking through the PR, and it looks like you don't have Dart installed in both the GitHub Actions environment as well as the Docker environment. That's why your tests aren't passing.

src/modules/dart.rs
HofmannZ: Ah okay, but performance-wise it's the same right?

matiaskotlik: something like this is never gonna affect performance anyway, but the way I suggested should be more efficient unless Rust optimizes it because it only creates one `Option` (`.ok()` does this) while the way I described in my second comment creates two `Option`'s, one with the `.ok()` and another directly from the `Some` call.

HofmannZ: That's pretty cool thanks!

matchai:
Hi @HofmannZ 👋 
I'm sorry this PR slipped through the cracks. I really feel bad that your work didn't make it in.

I would love to see a Dart module added to the prompt, though the PR has now gotten stale with the refactoring that the project has seen in the last few months. If you're up for it, I encourage you to give it another go, referencing recently added modules.

For now, I'll be closing this PR as stale, but feel free to ping me if you need a hand if you give it another go.

PR #616, opened 2019-11-01
feat: Add option to select virtualenv visibility

### LINKED ISSUE #615: Option to hide Python path

Sometimes (often when using `nix`) paths look horrible and long.

For instance, I currently see this in my prompt:
```
backend on ? dev0 [?] via 🐍 v3.7.4 (pmsp58j8656j9xmppkj8gykkys065q87-python3-3.7.4-env) via impure took 10s 
```

Simply having an option `show_path` for removing the parentheses with the path inside  would work perfectly.

Perhaps it would be even better to have the option accept `yes`, `no` and `not_in_nixshell`, because one might have also have independent `venv`s, and in that case the info would be shorter and more useful.

None

Scotsguy:
#### Description
This adds the config option `show_venv`, which can be one of `always`, `never`, or `not_in_nix`, and selects whether to show the virtualenv name in brackets after the Python version.
always and never are self-explanatory, but not_in_nix only shows it when `$IN_NIX_SHELL` is *not* set.

I have put the enum in the config file, as I believe is only relevant in the context of configuration.

(Sorry for the large markdown diff, that's Prettier doing its thing)

#### Motivation and Context

Closes #615 

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -474,36 +474,36 @@ current directory.
 
 ### Options
 
-| Variable            | Default                    | Description                                                     |
-| ------------------- | -------------------------- | --------------------------------------------------------------- |
-| `conflicted`        | `"="`                      | This branch has merge conflicts.                                |
-| `conflicted_count`  | [link](#git-status-counts) | Show and style the number of conflicts.                         |
-| `ahead`             | `"⇡"`                      | This branch is ahead of the branch being tracked.               |
-| `behind`            | `"⇣"`                      | This branch is behind of the branch being tracked.              |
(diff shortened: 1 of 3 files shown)


heyrict:
Hmm, currently we are undergoing a discussion in #472 about styling (or even removing) the undocumented segments in the config.

I think using the following config will work well if we can get that issue addressed.

```toml
[python]
virtualenv.style = "none"
```

heyrict:
Also, if we can address #607 by @johnletey , we can have a default config similar to this:

```toml
virtualenv.style = [
  { env = "IN_NIX_SHELL", match = "none" },
  "default color",
]
```

andytom:

PR #614, opened 2019-11-01
feat: add k8s granular config

nomaed:
#### Description
This PR implements the suggestion from https://github.com/starship/starship/issues/570 to add an ability to customize the configuration of the `kubernetes` module style, based on the current context.

A new variable is added to the config section, called `environments`, which is a list of possible customizations. Each such customization is an object with a `name` variable, which includes a possible substring of an context name, and an optional `style` and `icon` that will override the global configuration, if the currently used context matched the `names` (substring match).

#### Motivation and Context
It is common to work with development, testing, staging and production environments when deploying to kubernetes clusters, especially with continuous deployment. However, it is too easy to be using a production context (for example, for inspecting logs) and forgetting to switch to the personal development context when deploying any changes. 

This change helps giving better visibility, by allowing to change the style of the kubernetes module in the prompt to draw attention, or changing the color to red (as an example) if using a context that should be used with care.

Closes #570 

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Screenshots (if appropriate):
Screen Shot 2019-10-28 at 11 36 26

#### How Has This Been Tested?

- [X] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [X] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #618, opened 2019-11-02
fix: update python to v3.7.5

spk:
#### Description

Update python to v3.7.5

#### Motivation and Context

 To update CI on https://github.com/starship/starship/pull/617

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -117,7 +117,7 @@ jobs:
       # Install Python at a fixed version
       - uses: actions/setup-python@v1
         with:
-          python-version: "3.7.4"
+          python-version: "3.7.5"
 
       # Install dotnet at a fixed version
       - uses: actions/setup-dotnet@v1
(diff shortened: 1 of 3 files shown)


johnletey:
@spk As I said in https://github.com/starship/starship/pull/614#issuecomment-549018461:

> This is problem that @matchai and I have noticed. Take a look at [actions/setup-python#34](https://github.com/actions/setup-python/issues/34). It's an issue that will keep happening as long as the GitHub Actions tool cache works the way it does. Matan has spoken to folks on the Actions team and let them know that it's painful for maintainers, and they will be fixing it, but not any time soon. So we have to wait for the tool caches for all OSes to be updated, before we can address this issue.

nomaed:
While it is best to use exactly the same version for all tests (including the patch version), would it not be reasonable to set the version to `3.7.x` for the time being, until the problem is solved?

I would _assume_ that this should not have negative consequences on the quality of the tests.

matchai:
Our tests generate the prompt based on the version of a runtime that is installed on the machine. We would need fixed versions on all platforms for our tests to pass the way they are right now.
This will not be a problem if we get around to addressing #289.

matchai:
Closing this PR until the tool cache is updated to add Python v3.7.5 support on all OSes.

PR #622, opened 2019-11-04
feat: Add global configuration paths

### LINKED ISSUE #565: Global/host-wide config file?
## Feature Request

Great project! Enjoying so far, but would like to add different configurations on different hosts in a manageable way.

In my business, I currently manage around 50 different Linux hosts. Some of these are production servers, some development, and these have different directory structures as well as hostname patterns. I'd like my prompt to show me exactly where I am. I can do this with Starship, but only by having different starship.toml configs on different hosts.

For example, many of our production web servers have hostnames where the first part is "web01" - "web03" ... but the second part of the hostname shows the actual client name of the server we manage. So the first part of the hostname doesn't tell me much but the second does. For dev servers, I don't care about the second part.

So for these production servers (and these servers only) I want to have this config:

[hostname]
trim_at = ".com"

Another scenario is on our dev servers, where we have multiple copies of a site -- dev, stage, etc. For these servers, this is very useful:

[directory]
truncate_to_repo = false

... as it shows enough directory info to show which site I'm in.

I also have other team members who might find these defaults quite useful.

I would like the configuration system to support a sitewide config file, in addition to a user config file -- e.g. /etc/starship.toml, which gets read first. Using normal config patterns, any config setting in this file can be overridden in a user's ~/.config/starship.toml file.

This solution would allow us to set different settings in /etc/starship.toml through our normal configuration management systems, and allow users to use their preferred dotfile management system for any personal preferences.

This is a common pattern for other software, don't know of any better solution...

marblenix:
#### Description
Added the ability to have global configuration settings per host.

#### Motivation and Context
Closes #565

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -55,6 +55,22 @@ Most modules in starship allow you to configure their display styles. This is do
 
 Note that what styling looks like will be controlled by your terminal emulator. For example, some terminal emulators will brighten the colors instead of bolding text, and some color themes use the same values for the normal and bright colors. Also, to get italic text, your terminal must support italics.
 
+### Global Config
+
+Default, system-wide configs can be added with the following config paths (later paths overrite former ones):
+
+#### Unix-like
+
+1. `/etc/starship.toml`
+2. `/usr/local/etc/starship.toml`
+3. `$HOME/starship.toml`
+
+#### Windows
+
+1. `%ProgramData%\starship.toml`
+2. `%AppData%\starship.toml`
+3. `%HomeDrive%%HomePath%\.config\starship.toml`
+
 ## Prompt
 
 This is the list of prompt-wide configuration options.
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,13 +1,15 @@
+use std::clone::Clone;
+use std::env;
+use std::marker::Sized;
+use std::path::PathBuf;
+
+use ansi_term::{Color, Style};
+use dirs::home_dir;
+use toml::map::Map;
+use toml::Value;
+
 use crate::configs::StarshipRootConfig;
 use crate::utils;
-use ansi_term::{Color, Style};
-
-use std::clone::Clone;
-use std::marker::Sized;
-
-use dirs::home_dir;
-use std::env;
-use toml::Value;
 
 /// Root config of a module.
 pub trait RootModuleConfig<'a>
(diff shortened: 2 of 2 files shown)


marblenix:
Attached is a trace output from a windows box.
[trace.txt](https://github.com/starship/starship/files/3802655/trace.txt)

marblenix:
rebased off of master

docs/config/README.md
```diff
@@ -55,6 +55,22 @@ Most modules in starship allow you to configure their display styles. This is do
...
 
+### Global Config
+
+Default, system-wide configs can be added with the following config paths (later paths overrite former ones):
+
+#### Unix-like
+
+1. `/etc/starship.toml`
+2. `/usr/local/etc/starship.toml`
+3. `$HOME/starship.toml`
```
matchai: Was it intentional that `~/.config/starship.toml` was replaced with `~/starship.toml`?

src/config.rs
```diff
@@ -128,20 +130,90 @@ pub struct StarshipConfig {
 impl StarshipConfig {
     /// Initialize the Config struct
     pub fn initialize() -> Self {
-        if let Some(file_data) = Self::config_from_file() {
-            StarshipConfig {
-                config: Some(file_data),
+        let mut config = toml::value::Table::new();
+
+        if let Some(global_config_path) = Self::global_config_path() {
```
matchai: Most of this logic appears to be repeated.
Could we perhaps create an iterator of directories to `.fold()` through with a merging function?

src/config.rs
```diff
@@ -153,21 +225,67 @@ impl StarshipConfig {
...
+            }
+            Err(err) => {
+                log::warn!("Failed to parse config: \n{}", err);
+                None
+            }
+        }
+    }
+
+    /// Returns a deeply merged map where the values of map2 will overwrite the values in map1 if there is a collision
+    fn deep_merge(map1: Map<String, Value>, map2: Map<String, Value>) -> Map<String, Value> {
```
matchai: Is a "deep merge" the expected behavior?
I can see it causing some unexpected behavior for users, rather than a regular merge.

docs/config/README.md
```diff
@@ -55,6 +55,22 @@ Most modules in starship allow you to configure their display styles. This is do
 
 Note that what styling looks like will be controlled by your terminal emulator. For example, some terminal emulators will brighten the colors instead of bolding text, and some color themes use the same values for the normal and bright colors. Also, to get italic text, your terminal must support italics.
 
+### Global Config
+
+Default, system-wide configs can be added with the following config paths (later paths overrite former ones):
```
olets: override or overwrite?

PR #626, opened 2019-11-04
feat: Add option python module to disable virtualenv name display.

DanCardin:
#### Motivation and Context
As someone who uses local .venv folders, all my virtualenv names end up being .venv, and the name is irrelevant. However seeing the version is nice!

To that end, I added 2 options:
* one to optionally only show anything if the virtualenvironment is active (because *really* i can't do anything with python until I'm in that virtualenv)
* one to disable the inclusion of the virtualenvironment name.

A couple of notes:
* I'm not at all married to the option names, feel free to suggest new ones if the changes are otherwise agreeable. 
* I wasn't sure what to do about pyenv. I dont use pyenv, and i wasn't sure what would be appropriate in that context.
* It wasn't obvious to me if/how i could add tests for the behavior, looking at the existing tests for python.
* It seems like, rather than adding a bunch of flag options, it'd be better to support a sort of interpolationy thing like:
```toml
[python]
pyenv = true/false
prompt = "🐍 ${version} ${virtualenv}"
```
because it'd replace: symbol, pyenv_prefix, display_virtualenv, and maybe pyenv_version_name. But I wasn't prepared to go down that road, myself.

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -835,13 +835,15 @@ The module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Variable             | Default         | Description                                                                 |
-| -------------------- | --------------- | --------------------------------------------------------------------------- |
-| `symbol`             | `"🐍 "`         | The symbol used before displaying the version of Python.                    |
-| `pyenv_version_name` | `false`         | Use pyenv to get Python version                                             |
-| `pyenv_prefix`       | `"pyenv "`      | Prefix before pyenv version display (default display is `pyenv MY_VERSION`) |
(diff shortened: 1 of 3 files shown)


Scotsguy:
#616, #624 ?

DanCardin:
Unfortunately #624 is probably a ways off.

#616 is certainly a subset of this; which i can remove because that impl is def more flexible. But I still want to only show anything when inside the venv, as opposed to whenever there's an indicator of a python project in the dir.

andytom:

PR #627, opened 2019-11-04
Suggest to use force flag with cargo install to allow installing updates

timvisee:
#### Description
The `cargo install` command requires use of the `-f` flag when recompiling/reinstalling the crate to update.

This PR adds this flag as default in the installation instructions.

---

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ The prompt shows information you need while you're working, while staying sleek 
    #### Rust (v1.38 or higher)
 
    ```sh
-   $ cargo install starship
+   $ cargo install starship -f
    ```
 
    #### Arch Linux (AUR)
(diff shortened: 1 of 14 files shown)


matchai:
I don't think it's necessary to add the `--force` flag by default. If a user tries installing starship without the flag, they'll be appropriately warned by Rust that starship is already installed, and they are given instructions explaining how to overwrite it.

```
$ cargo install starship
    Updating crates.io index
  Downloaded starship v0.26.2
  Downloaded 1 crate (2.8 MB) in 8.67s
error: binary `starship` already exists in destination as part of `starship v0.25.2 (/Users/matan/dev/starship)`
Add --force to overwrite
```

This can be a good way for users to find out that starship may be already installed on their machine, and doesn't add noisy flags to our installation process.

If there are added benefits that I hadn't considered, we can certainly reopen this issue. 😊

timvisee:
Alright!

It seems the situation has improved as well. Now it notifies users before compiling, not at the end of it. I didn't know about this.

Thanks for liking into it though.

PR #645, opened 2019-11-16
feat: Add util for parsing asdf files

### LINKED ISSUE #465: ASDF support for multiple programming language versions
## Feature Request

This is in contrast to, but not a replacement for, the issue described and discussed in #149 and being actively worked in #437 . I would love to contribute this support if possible, and have some scratch code locally that I may be able to clean up and publish this weekend. I am not a professional Rust developer, just an enthusiast, so mentoring and feedback is very welcome. 

Many Elixir developers use a toolchain management program called [asdf](https://github.com/asdf-vm/asdf/) together with its plugins [asdf-elixir](https://github.com/asdf-vm/asdf-elixir/) and [asdf-erlang](https://github.com/asdf-vm/asdf-erlang/) to install and select specific combinations of those languages' tooling at runtime. There is a [fairly broad ecosystem](https://github.com/asdf-vm/asdf-plugins) of other surrounding plugins as well, some of which overlap with pre-existing Starship modules, such as Node.

Notably, this ecosystem uses [a plaintext file with an easily parsable format](https://github.com/asdf-vm/asdf/blob/54ef3befaecb185968831a4b62e544782a6bc4b6/docs/core-configuration.md#tool-versions) to describe what versions to use within a given project, which can be read in with a ~single syscall versus several successive command invocations.

This means our projects can have a committed `.tool-versions` file that looks something like:

```
erlang 22.0.7
elixir 1.9.1-otp-22
nodejs 10.15.3
```

These are automatically respected by our shell commands as long as you have `asdf` as well as its plugins properly installed and configured.

I would like to introduce an optional Starship module that parses `.tool-versions` files and displays them with comparable formatting to the other per-language modules, and configurable to show one/some/all of those version-language pairings using similar UTF8 glyphs and color schemes. I believe this is referred to as `segments` in the existing code?

From some perspectives this can be considered to be either a superset of or parallel to the existing language modules, so for now I would recommend introducing this in a disabled-by-default and opt-in. I hold this view both in regards to the configuration file defaults and the compilation-feature defaults (i.e. the battery support is opt-in as well).

My individual needs would stop at Elixir/Erlang/Node.js support, but I think it's plausible to set convenient groundwork for the module and to maybe include the top N most popular plugins from the `asdf-plugins` link above.

Most (all?) of the existing language modules execute something analogous to `elixir --version`, which should continue to work under the presence of `asdf` in one's shell environment where those languages are already supported or would be in the future. If this feature request is viewed as feature-creep as a result, I can accept that stance.

andytom:
#### Description

Have added a simple helper function to fetch the version of a tool from
an asdf file if there is one present in the current directory. This
should be a fast and easy way to get the version for users who use asdf.

This just add the util function further PRs will be required to start using
this in the modules.

#### Motivation and Context

Closes #465

#### Types of changes

#### Checklist:

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::fs::File;
 use std::io::{Read, Result};
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


matchai:
Does using asdf make the usual `node --version` command not work correctly?
Just wondering if we need asdf-specific logic.

andytom:
Completely forgot about this, going to close this since I don't think there is much use having this at the moment.

PR #646, opened 2019-11-16
feat: Implement AWS region aliases

### LINKED ISSUE #637: Add friendly AWS region names to prompt

I work in multiple AWS regions on a daily basis and have my current prompt set to display a friendly region name. i.e. ap-southeast-2: au, us-east-1: va, etc.

Add to the AWS module some friendly names or make it a config option so people can use what ever name works for them. Whether to display the friendly names would be a config option too. Happy to do up a PR.

Could hack around this with #607 and set an alternative variable in my AWS scripts though this would be a bit of a hack.

mfs:
#### Description
Allows the user to configure AWS region aliases to be displayed along with the AWS region name.

#### Motivation and Context
For users that work with multiple AWS regions a friendly name helps to keep them straight.
Closes #637

#### Types of changes

Tests updated. May need to add more tests to cover all code changes.

#### Checklist:
May need more tests.


alexjurkiewicz:
This would be great. Rather than a user-customisable table, why not hardcode the city name for each region? So you could set `simple_region_name: true` to display `ap-southeast-2` as `Sydney`

mfs:
> This would be great. Rather than a user-customisable table, why not hardcode the city name for each region? So you could set `simple_region_name: true` to display `ap-southeast-2` as `Sydney`

Yeah, this was my initial idea and it has some pros and cons. One con is that I, and I guess others, like to keep their prompt fairly concise. These people probably wouldn't want 'Sydney' in full in their prompt. We could use abbreviations, e.g. 'syd' but that might not suit current internal naming conventions. Where I work we always refer to `ap-southeast-2` as 'au' not Sydney or any variation of.

If it was hard coded the table would need updating whenever a new region was created. Probably not a big hassle.

I guess I just went for what was the most configurable from the user point of view. i.e. add some config for just the regions you care about you are done.

andytom:
In the spirit of making this more configurable for the users would it be better to only display the alias rather than the region name and the alias? For example if there is an alias setup for `"ap-southeast-2"` to `"au"` display `☁️  au` rather than `☁️  ap-southeast-2 au`. This gives the user more flexibility about how to format the result, if the user wants to display both the alias and the original region they can do so by configuring the alias to be `"ap-southeast-2 au"`. @mfs What do you think?

mfs:
I like this idea a lot. I'll amend the PR tonight after work.

docs/config/README.md
```diff
@@ -129,6 +129,7 @@ The `aws` module shows the current AWS region and profile. This is based on
 | `style`           | `"bold yellow"` | The style for the module.                                                   |
 | `disabled`        | `false`         | Disables the `AWS` module.                                                  |
 | `displayed_items` | `all`           | Choose which item to display. Possible values: [`all`, `profile`, `region`] |
+| `region_aliases`  |                 | Table of region aliases to display in addition to the AWS name.             |
```
andytom: I think the convention for the docs is to put the `style`  and `disabled` vars at the bottom of the table.

matchai:
@all-contributors Please add @mfs for code, docs, and test. 🎉

PR #658, opened 2019-11-22
fix: Escape data that bash can execute

wendorf:
#### Description
This change sets the PS1 variable to call starship itself, so the
'$(foo)' construct returned by starship will not be evaluated by bash.

#### Motivation and Context
When PS1 includes something like '$(foo)' it will be evaluated. This
means that if the value of `starship prompt` includes such a construct,
that will be evaluated. It is possible for malicious parties to use this
behavior to execute code. See https://github.com/njhartwell/pw3nage for
a proof of concept.

#### Types of changes

bash 5.0.11: Issue present, this fixes it
zsh 5.7.1: Issue not present
fish: Untested

#### Checklist:

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -33,10 +33,10 @@ starship_precmd() {
     if [[ $STARSHIP_START_TIME ]]; then
         STARSHIP_END_TIME=$(date +%s)
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
-        PS1="$(::STARSHIP:: prompt --status=$STATUS --jobs="$(jobs -p | wc -l)" --cmd-duration=$STARSHIP_DURATION)"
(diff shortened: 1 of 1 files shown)


wendorf:
I'm not convinced that this is the best way to fix the issue (unexporting `STARSHIP_SHELL` is almost certainly wrong), but untangling escape characters, etc. proved to be pretty complicated. This PR as-is is intended to serve as a showcase of the issue and a possible fix. If anyone has guidance for how to do a better fix, please let me know and I'll try to improve this.

If someone has a better idea for how to mitigate this problem and wants to submit a different PR, please feel free to do so and I can close this.

lucarin91:
I propose a possible solution for this issue on the PR #884 .
Instead of changing the bash init script I propose to escape the character during the starship prompt building.

andytom:
Since PR #884 has been merged, going to close this PR.

PR #660, opened 2019-11-24
chore: Add a helper function for executing commands

### LINKED ISSUE #568: Add a helper function for running commands

I think it might be worth writing a little helper function for executing commands, we seem to have a similar use case in a lot of modules maybe it would be nice to have a common way of doing this similar to the [`read_file` util](https://github.com/starship/starship/blob/master/src/utils.rs#L6-L12).

_Originally posted by @andytom in https://github.com/starship/starship/pull/563#issuecomment-544119990_

A new function that wraps the Command function and fits the common use case we have across our different modules.

I've not really considered any other solutions but if someone has a better idea please add it here.

andytom:
#### Description

Have added a little helper function for executing commands. Note this just adds the function further PRs will be required to start using it.

#### Motivation and Context

Closes #568

This is my best guess at what we need most of the time from having a look at the existing modules. Please feel free to suggest alternatives if I've missed something.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,6 +1,7 @@
 use std::fs::File;
 use std::io::{Read, Result};
 use std::path::Path;
+use std::process::Command;
 
 /// Return the string contents of a file
 pub fn read_file<P: AsRef<Path>>(file_name: P) -> Result<String> {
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -10,3 +11,70 @@ pub fn read_file<P: AsRef<Path>>(file_name: P) -> Result<String> {
     file.read_to_string(&mut data)?;
     Ok(data)
 }
+
+/// Execute a command and return the output on stdout and stderr if sucessful
+pub fn exec_cmd(cmd: &str, args: &[&str]) -> Option<(String, String)> {
```
matchai: I think that working with a tuple here may make understanding how it's used a little more difficult than having a struct with `stdout` and `stderr`.

In most cases it would be used like this:
```rust
const version = exec_cmd("npm", &["--version"])?.0;
```

matchai:
For what it's worth, I've begun working on mocking capabilities a little bit ago, and this is what I opted to use for commands: https://github.com/starship/starship/compare/command-mocking#diff-97cb3a0ef783560461f4b6c7a1ae6678

I found that the single argument was the easiest to make sense of, but I do think we should make `stdout` and `stderr` accessible.

src/utils.rs
andytom: Have switched this over to return a simple struct with fields `stdout` and `stderr`, usage would now be something like
```rust
let ruby_version = utils::exec_cmd("ruby", &["-v"])?.stdout;
```

src/utils.rs
```diff
@@ -10,3 +11,97 @@ pub fn read_file<P: AsRef<Path>>(file_name: P) -> Result<String> {
...
+
+impl PartialEq for CommandOutput {
+    fn eq(&self, other: &Self) -> bool {
+        self.stdout == other.stdout && self.stderr == other.stderr
+    }
+}
+
+/// Execute a command and return the output on stdout and stderr if sucessful
+pub fn exec_cmd(cmd: &str, args: &[&str]) -> Option<CommandOutput> {
+    log::debug!("Executing command '{:?}' with args '{:?}'", cmd, args);
```
matchai: I think this would be better suited as a trace log as it explains how the code is executing.
```suggestion
    log::trace!("Executing command '{:?}' with args '{:?}'", cmd, args);
```

src/utils.rs
```diff
@@ -10,3 +11,97 @@ pub fn read_file<P: AsRef<Path>>(file_name: P) -> Result<String> {
...
+    match Command::new(cmd).args(args).output() {
+        Ok(output) => {
+            let stdout_string = String::from_utf8(output.stdout).unwrap();
+            let stderr_string = String::from_utf8(output.stderr).unwrap();
+
+            if !output.status.success() {
+                log::debug!("Non-zero exit code '{:?}'", output.status.code());
```
matchai: Likewise with these log statements.
I think they might add up to being a little too much for debug logs.

andytom:
> A couple small notes, but this looks ready to merge otherwise. 👍

All fixed.

PR #669, opened 2019-11-25
fix: multiple batteries support in the battery module

lucarin91:
#### Description

I update the battery module, in order to implement an aggregating view of all the batteries present in the notebook.
The percentage of the battery is calculated with a weighted average of the energy of all the batteries. Instead, the battery state is calculated with the following rule:
- Charging if at least one is Charging
- Discharging if at least one is Discharging
- Full if all are Full
- Empty if all are Empty
- Unknown otherwise 

#### Motivation and Context

As stated in issue #656 if there are multiple batteries in the notebook the battery module gives inconsistent values of the battery percentage and state.

#### Types of changes

I test it on my Lenovo T450s that has two batteries. For now, there is not any unit test on the battery module.

#### Checklist:

### DIFF
--- a/src/modules/battery.rs
+++ b/src/modules/battery.rs
@@ -75,27 +75,76 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
 fn get_battery_status() -> Option<BatteryStatus> {
     let battery_manager = battery::Manager::new().ok()?;
-    match battery_manager.batteries().ok()?.next() {
-        Some(Ok(battery)) => {
-            log::debug!("Battery found: {:?}", battery);
-            let battery_status = BatteryStatus {
-                percentage: battery.state_of_charge().value * 100.0,
-                state: battery.state(),
-            };
-
-            Some(battery_status)
-        }
-        Some(Err(e)) => {
-            log::debug!("Unable to access battery information:\n{}", &e);
-            None
-        }
-        None => {
-            log::debug!("No batteries found");
-            None
-        }
+    let batteries = battery_manager.batteries().ok()?;
+    let battery_contructor = batteries
+        .filter_map(|battery| match battery {
+            Ok(battery) => {
+                log::debug!("Battery found: {:?}", battery);
+                Some(BatteryInfo {
+                    energy: battery.energy().value,
+                    energy_full: battery.energy_full().value,
+                    state: battery.state(),
+                })
+            }
+            Err(e) => {
+                log::debug!("Unable to access battery information:\n{}", &e);
+                None
+            }
+        })
+        .fold(
+            BatteryInfo {
+                energy: 0.0,
+                energy_full: 0.0,
+                state: battery::State::Unknown,
+            },
+            |mut acc, x| {
+                acc.energy += x.energy;
+                acc.energy_full += x.energy_full;
+                acc.state = merge_battery_states(acc.state, x.state);
+                acc
+            },
+        );
+    if battery_contructor.energy_full != 0.0 {
+        let battery = BatteryStatus {
+            percentage: battery_contructor.energy / battery_contructor.energy_full * 100.0,
+            state: battery_contructor.state,
+        };
+        log::debug!("Battery status: {:?}", battery);
+        Some(battery)
+    } else {
+        None
     }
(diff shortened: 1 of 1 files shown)


src/modules/battery.rs
```diff
@@ -75,27 +75,76 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                state: battery::State::Unknown,
+            },
+            |mut acc, x| {
+                acc.energy += x.energy;
+                acc.energy_full += x.energy_full;
+                acc.state = battery_state_fold(acc.state, x.state);
+                acc
+            },
+        );
+    if battery_contructor.energy_full != 0.0 {
```
matchai: Do we need this condition?
`energy_full` could be `0.0` in the case of a depleted battery that is charging. I think we should still show the charging status in that case.

src/modules/battery.rs
```diff
@@ -75,27 +75,76 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    } else if state1 == Unknown {
+        state2
+    } else if state2 == Unknown {
+        state1
+    } else {
+        Unknown
     }
 }
 
+struct BatteryContructor {
```
matchai: I think a more fitting title might be `BatteryInfo`.

src/modules/battery.rs
```diff
@@ -75,27 +75,76 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+            |mut acc, x| {
+                acc.energy += x.energy;
+                acc.energy_full += x.energy_full;
+                acc.state = battery_state_fold(acc.state, x.state);
+                acc
+            },
+        );
+    if battery_contructor.energy_full != 0.0 {
+        let battery = BatteryStatus {
+            percentage: battery_contructor.energy / battery_contructor.energy_full * 100.0,
```
matchai: I can see this possibly returning a long decimal value. Perhaps we should limit the number of decimal places.

src/modules/battery.rs
```diff
@@ -75,27 +75,76 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-        None => {
-            log::debug!("No batteries found");
-            None
-        }
+/// the fold return Charging if at least one is charging
+///                 Discharging if at least one is Discharging
+///                 Full if both are Full or one is Full and the other Unknow
+///                 Empty if both are Empty or one is Empty and the other Unknow
+///                 Unknown otherwise
+fn battery_state_fold(state1: battery::State, state2: battery::State) -> battery::State {
```
matchai: I think a more descriptive function name would be `merge_battery_states`.

src/modules/battery.rs
lucarin91: We need to consider the case in which either the battery manager or the subsequent filter_map does not return any battery information. In those cases, the fold operation will return a `BatteryContructor` with both `energy` and `energy_full` at 0.0.
This will end up in a division by zero on line 109, but apart for that, we should return None in those situations.
I agree with you that this is not a very good condition, but I think that we need something like that.

src/modules/battery.rs
lucarin91: Ok, I will update it.

src/modules/battery.rs
lucarin91: Sorry, I do not understand, do you want to cast it to an integer value?

src/modules/battery.rs
lucarin91: Good name, I will update it.

src/modules/battery.rs
matchai: My mistake. I thought that `energy_full` was `energy`.
Yeah, this is a fine condition to have. 👍

src/modules/battery.rs
matchai: My bad. I had forgotten that we round the value later on in the file:
https://github.com/starship/starship/blob/master/src/modules/battery.rs#L61

matchai:
Thank you for your contribution, Luca! 😄 

@all-contributors Please add @lucarin91 for code. 🎉

PR #672, opened 2019-11-27
fix: Fix panic when using fish-style pwd with unicode symbols

weirongxu:
#### Description

image

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -1,4 +1,5 @@
 use path_slash::PathExt;
+use unicode_segmentation::UnicodeSegmentation;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


weirongxu:
> Thank you for submitting a fix @weirongxu!
> Since we already have [`unicode-segmentation`](https://github.com/unicode-rs/unicode-segmentation) as a dependency, this looks like a good time to separate by grapheme, rather than separate by character.
> What do you think?

Sure, I already use unicode-segmentation to instead of it

matchai:
@all-contributors Please add @weirongxu for code and tests.

PR #673, opened 2019-11-28
add git_commit module

houqp:
#### Description
This PR adds git commit module to show git commit hash in the prompt

#### Motivation and Context
Having git hash in the prompt is very useful for knowing which commit I am currently at after a git checkout operation.

Closes #

#### Types of changes

Screen Shot 2019-11-27 at 4 48 42 PM

#### Checklist:


houqp:
@johnletey do you know what needs to happen to get this merged into master?

matchai:
I'd be curious to hear what your use case is for this module. I don't personally have much use for the commit hash, so could you speak to its use in your workflow?

I'd be open to having it added to the prompt, but I don't think it would be enabled by default, since it's quite a long module that doesn't provide much actionable value.

Delapouite:
The hash commit is nice, but an information I find even more valuable is the **commit date**, expressed in relative terms like `5 days ago`.
This way, when I `cd` in a project I don't interact with regularly, the prompt tells me directly if the current branch seems _old_ and often reminds me that I have to `git fetch` / `git pull` the latest version.

houqp:
@matchai this information is very useful for those who heavily use `commit --amend` and rebase during day to day git workflow.

Sometimes, I would amend or rebase mistakenly and want to revert back to the old detached head. This has happened many times to me and I was able to find the right commit by going through zsh prompt in terminal scroll back.

I have changed the default to have this module disabled. Let me know if this looks good to you.

joshukraine:
FWIW, I just found and installed Starship yesterday (using fish), and this feature was one of the first things I looked for. I had this in my zsh prompt and really liked it. Would love to see this merged! 😁

houqp:
@matchai let me know what other change you want me to make to get this merged. Like @joshukraine mentioned, this is a feature that would benefit many zsh users.

docs/config/README.md
```diff
@@ -437,6 +438,29 @@ truncation_length = 4
 truncation_symbol = ""
 ```
 
+
+## Git Commit
+
+The `git_commit` module shows the active branch of the repo in your current directory.
```
matchai: I'd suggest adding a container clarifying that this module is disabled by default, as we do for the `time` module:
https://github.com/starship/starship/blob/ab310e2cf047e6ed57c7519d997d28e70e702fac/docs/config/README.md#L938-L943

src/configs/git_commit.rs
```diff
@@ -0,0 +1,25 @@
+use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
+
+use ansi_term::{Color, Style};
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
+pub struct GitCommitConfig<'a> {
+    pub style: Style,
```
matchai: Total nitpick.
Could we please reorder this to have unique configuration at the top, to maintain consistency with other module configuration. 😄
- hash
- prefix
- suffix
- style
- disabled

src/configs/git_commit.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub disabled: bool,
+    pub prefix: &'a str,
+    pub suffix: &'a str,
+    pub hash: SegmentConfig<'a>,
+}
+
+impl<'a> RootModuleConfig<'a> for GitCommitConfig<'a> {
+    fn new() -> Self {
+        GitCommitConfig {
```
matchai: Similarly here ☝️

src/modules/git_commit.rs
```diff
@@ -0,0 +1,52 @@
+use super::{Context, Module, RootModuleConfig};
+use git2::Repository;
+
+use crate::configs::git_commit::GitCommitConfig;
+
+pub fn id_to_hex_abbrev(bytes: &[u8], len: usize) -> String {
```
matchai: To maintain consistency with the other modules in the project, could we please move utility functions to the bottom of the file?

src/modules/git_commit.rs
```diff
@@ -0,0 +1,52 @@
...
+        .get_suffix()
+        .set_value(config.suffix)
+        .set_style(config.style);
+    module.set_style(config.style);
+
+    let repo = context.get_repo().ok()?;
+    let repo_root = repo.root.as_ref()?;
+    let git_repo = Repository::open(repo_root).ok()?;
+
+    let git_head = git_repo.head();
```
matchai: `if let Ok(git_head) = git_head {` below can be removed if we make the following update:
```suggestion
    let git_head = git_repo.head().ok()?;
```

src/modules/git_commit.rs
```diff
@@ -0,0 +1,52 @@
...
+        .collect::<Vec<String>>()
+        .join("")
+}
+
+/// Creates a module with the Git commit in the current directory
+///
+/// Will display the commit hash if the current directory is a git repo
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("git_commit");
+    let config = GitCommitConfig::try_load(module.config);
```
matchai: Because we don't yet have the logic to disable modules by default, the following condition has to be manually added for now:
https://github.com/starship/starship/blob/master/src/modules/time.rs#L14-L16

src/modules/git_commit.rs
```diff
@@ -0,0 +1,52 @@
...
+        let commit_oid = head_commit.id();
+        module.create_segment(
+            "hash",
+            &config
+                .hash
+                .with_value(&id_to_hex_abbrev(commit_oid.as_bytes(), 7)),
+        );
+    }
+
+    if module.is_empty() {
```
matchai: Can you think of a situation where this code path can be hit?
I'm having trouble thinking of any, if they've made it to here.

src/modules/git_commit.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    let git_head = git_repo.head();
+    if let Ok(git_head) = git_head {
+        let head_commit = git_head.peel_to_commit().ok()?;
+        let commit_oid = head_commit.id();
+        module.create_segment(
+            "hash",
+            &config
+                .hash
+                .with_value(&id_to_hex_abbrev(commit_oid.as_bytes(), 7)),
```
matchai: I find the hash to be rather long. I think convention for the abbreviated commit is for it to be 7 characters long, such as when you run `git log --abbrev-commit`.

Right now it looks like it's coming out to 14 characters.
What are your thoughts around shortening it further?

joshukraine: > I find the hash to be rather long. I think convention for the abbreviated commit is for it to be 7 characters long, such as when you run `git log --abbrev-commit`.
> 
> Right now it looks like it's coming out to 14 characters.
> What are your thoughts around shortening it further?

If I can throw in my two cents, I'd also vote for a 7-character hash. Both GitHub and BitBucket display that length for short commits, and it also takes up less screen real estate in the prompt. 🙂

houqp: So I did extensive research into this. Initially I was thinking of making it dynamic to match the behavior of `git rev-parse --short HEAD`.

After reading git's source code, it turned out to be a very deep rabbit hole. Git2 doesn't have an API for this. The actual implementation is at https://github.com/git/git/blob/master/sha1-name.c#L670. Which will result in hundreds lines of rust code counting all the helper functions it calls.

I ended up stick with the default length in Git to keep it simple as the first iteration, which is defined as 7: https://github.com/git/git/blob/master/builtin/rev-parse.c#L728. This should be long enough to work for majority of the code base. We can certainly port the dynamic length calculating algorithm from git if we ever get any request for increasing the length.

matchai:
As a friendly note, I'd suggest not force-pushing while having your code reviewed. 😊
Because there is no diff, the code has to be completely reviewed from scratch instead of just reviewing what has been changed.

src/modules/git_commit.rs
matchai: Since each byte is encoded as 2 hexadecimal characters, converting `7` bytes makes for a hash that is 14 characters long. A different approach may be needed to have 3.5 bytes displayed.

houqp: if we change a different encoding method, the displayed commit hash won't be useful for copy paste anymore. In order to use the displayed commit hash with the git cli, one would have to convert that 3.5 bytes into hex encoding first manually.

If you are still very concerned with the length, how about we make it configurable so users can pick whatever length they are most comfortable with?

matchai: > if we change to a different encoding method, the displayed commit hash won't be useful for copy paste anymore.

In my workflow, a 7-character hash is as useful as a 14-character hash, but I'm certainly open to the idea of adding a hash truncation configuration value. 👍 

TIL that `git` will allow you to use a hash as small as 4 characters:
image

houqp: yes, for smaller git repos, 7 bytes is more than enough, but will run into edge-cases for larger repos.

I have pushed an update to make the commit length configurable. It's named `min_commit_hash_length` to be consistent with git. This also means when we add dynamic length calculation in the future, the naming of this config will still make sense.

Let me know if you have any other thought on the implementation!

matchai: What would dynamic length calculation entail?
I was thinking that it would be a static value (`commit_hash_length`).

houqp: They way `git rev-parse --short HEAD` works is it starts with a user defined minimal length, and then traverse all git objects to to see if there is any other object shares the same hash prefix, if so, it increases the length until it becomes a unique prefix.

houqp: I have renamed it to `commit_hash_length`, we could also just remove this config when dynamic hash length is implemented.

docs/config/README.md
```diff
@@ -437,6 +438,38 @@ truncation_length = 4
...
+
+### Options
+
+| Variable                 | Default   | Description                                      |
+| ------------             | ----------| -------------------------------------------------|
+| `style`                  | `"green"` | The style for the module.                        |
+| `prefix`                 | `(`       | Prefix to display immediately before git commit. |
+| `suffix`                 | `)`       | Suffix to display immediately after git commit.  |
+| `disabled`               | `true`    | Disables the `git_commit` module.                |
+| `commit_hash_length` | `7`       | Minimal length for git commit hash to display    |
```
matchai: Nitpick: I find that this reads better.
```suggestion
| `commit_hash_length`     | `7`       | The length of the displayed git commit hash.        |
```

houqp:
@matchai test added!

src/configs/git_commit.rs
```diff
@@ -0,0 +1,28 @@
...
+            hash: SegmentConfig::default(),
+            prefix: "(",
+            suffix: ") ",
+            style: Color::Green.bold(),
+            disabled: true,
+            // be consistent with git by default, which has DEFAULT_ABBREV set to 7
+            commit_hash_length: 7,
```
matchai: It seems I've misunderstood what this value configures. Instead of being the character count of the truncated hash, it is in-fact the number of bytes of the truncated hash.
I don't think this comes across as very intuitive. Would you be open to changing the behavior?

tests/testsuite/git_commit.rs
```diff
@@ -0,0 +1,42 @@
...
+fn test_render_commit_hash() -> io::Result<()> {
+    let repo_dir = common::create_fixture_repo()?;
+
+    let mut git_output = Command::new("git")
+        .args(&["rev-parse", "HEAD"])
+        .current_dir(repo_dir.as_path())
+        .output()?
+        .stdout;
+    // 7 bytes hexed encoded equals 14 chars
+    git_output.truncate(14);
```
matchai: As mentioned earlier, I think a more reasonable default for the average user would be to have an output of 7 characters, rather than 7 bytes.

tests/testsuite/git_commit.rs
```diff
@@ -0,0 +1,42 @@
...
+                ",
+            )
+            .unwrap(),
+        )
+        .arg("--path")
+        .arg(repo_dir)
+        .output()?;
+
+    let actual = String::from_utf8(output.stdout).unwrap();
+    let expected = format!(
```
matchai: Slightly simpler than a second `format!()`:
```rust
    let expected = Color::Green.bold().paint(format!("({}) ", expected_hash)).to_string();
```

tests/testsuite/git_commit.rs
```diff
@@ -0,0 +1,42 @@
+use ansi_term::Color;
+use std::process::Command;
+use std::{io, str};
+
+use crate::common::{self, TestCommand};
+
+#[test]
+fn test_render_commit_hash() -> io::Result<()> {
```
matchai: Looks like a good test 👍 
I think the only remaining test would be another one testing `commit_hash_length` configuration.

src/configs/git_commit.rs
houqp: yeah, good point. I have changed it.

tests/testsuite/git_commit.rs
houqp: updated to 7 chars.

tests/testsuite/git_commit.rs
houqp: test added

houqp:
Looks like there is an issue with Clippy: ```
##[error]Unable to create clippy annotations! Reason: HttpError: Resource not accessible by integration
##[warning]It seems that this Action is executed from the forked repository.
##[warning]GitHub Actions are not allowed to create Check annotations, when executed for a forked repos. See https://github.com/actions-rs/clippy-check/issues/2 for details.
```

File: src/modules/git_commit.rs [hunk 14]
```diff
@@ -0,0 +1,55 @@
...
+
+/// len specifies length of hex encoded string
+pub fn id_to_hex_abbrev(bytes: &[u8], len: usize) -> String {
+    bytes
+        .iter()
+        .map(|b| format!("{:02x}", b))
+        .collect::<Vec<String>>()
+        .join("")
+        .chars()
+        .into_iter()
```
matchai: Despite clippy showing errors, it still displays the warnings from clippy beneath them:
image

It looks like `.into_iter()` is unnecessary here:
```suggestion
```

tests/testsuite/git_commit.rs
```diff
@@ -0,0 +1,68 @@
...
+        .arg(repo_dir)
+        .output()?;
+
+    let actual = String::from_utf8(output.stdout).unwrap();
+    let expected = Color::Green
+        .bold()
+        .paint(format!("({}) ", expected_hash))
+        .to_string();
+
+    assert_eq!(&*expected, actual);
```
matchai: I don't believe these should be necessary in the assertion:
```suggestion
    assert_eq!(expected, actual);
```

tests/testsuite/git_commit.rs
```diff
@@ -0,0 +1,68 @@
...
```
matchai:
```suggestion
    assert_eq!(expected, actual);
```

tests/testsuite/git_commit.rs
```diff
@@ -0,0 +1,68 @@
...
+    Ok(())
+}
+
+#[test]
+fn test_render_commit_hash_len_override() -> io::Result<()> {
```
matchai: The test here looks good. 👍

src/configs/git_commit.rs
```diff
@@ -0,0 +1,28 @@
...
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
+pub struct GitCommitConfig<'a> {
+    pub hash: SegmentConfig<'a>,
+    pub prefix: &'a str,
+    pub suffix: &'a str,
+    pub style: Style,
+    pub disabled: bool,
+    pub commit_hash_length: usize,
```
matchai: Nitpick: could we have configuration values unique to the module moved to the top of the list?

src/configs/git_commit.rs
matchai: Similarly here ☝️

docs/config/README.md
```diff
@@ -437,6 +438,38 @@ truncation_length = 4
...
+
+### Options
+
+| Variable              | Default   | Description                                      |
+| ------------          | ----------| -------------------------------------------------|
+| `style`               | `"green"` | The style for the module.                        |
+| `prefix`              | `(`       | Prefix to display immediately before git commit. |
+| `suffix`              | `)`       | Suffix to display immediately after git commit.  |
+| `disabled`            | `true`    | Disables the `git_commit` module.                |
+| `commit_hash_length`  | `7`       | The length of the displayed git commit hash.     |
```
matchai: And here ☝️

matchai:
@all-contributors Please add @houqp for code, docs, and test.

houqp:
@matchai thanks for the detailed review :)

joshukraine:
Thanks @houqp and @matchai! Awesome work! 😁🙌🏻

PR #682, opened 2019-12-02
refactor: Use format string

### LINKED ISSUE #624: Module configuration with format strings
With more and more configuration being introduced into Starship, we're starting to see a fair bit of configuration bloat that simply modifies the formatting of the output, rather than changing the way the prompt retrieves information or processes context. (e.g. #603, #623)

What I would like to propose is the following configuration:
```toml
[cmd_duration]
# ${duration} - The duration of the last running command

format = "took ${duration}!"
```

This would allow us to remove all the following configuration options in modules where they are static:
- `symbol`
- `prefix`
- `suffix`

In modules where the values are dynamic, they would remain. For example, the `character` module:

```toml
[character]
# ${prompt_symbol} - The prompt symbol

format = "${prompt_symbol} "
symbol = "❯"
error_symbol = "✖"
```

We could also remove the following configuration options from `git_status`, where enabling them would be implicit by them being in the format string:
- `conflicted_count`
- `untracked_count`
- `modified_count`
- etc...

This would, of course, require some changes in the way we educate users, but I think it would make for a less dense and complicated configuration system in the long run.

heyrict:
#### Description
Syntax like `format = "${prompt_symbol?style=fg:green bg:blue}"` should work.

This PR breaks many things:

- `prompt_order` now use a format string instead of an array of module names.

  ```rust
  prompt_order: "${username}${hostname}${kubernetes}${directory} \
               ${git_branch}${git_state}${git_status} \
               ${package}${dotnet}${golang}${java} \
               ${nodejs}${python}${ruby}${rust} \
               ${nix_shell}${conda}${memory_usage}${aws} \
               ${env_var}${cmd_duration}\n\
               ${jobs}${battery}${time}${character}",
  ```
- `prefix` and `suffix` configs in modules are deprecated in favor of `format`

  ```
  format: "took ${duration}"
  ```
- `format` key which is used to specify time display format, is renamed to `time_format`.
- All docs will break and should be rewritten.

User-defined styled text are provided via `styled` segment: `${styled?value=I'm green&style=bold green}`. (Maybe a shorter segment name e.g. `_` or `@`, looks better?)

Note: I have many code-style modifications in this PR as a new feature is introduced to starship. This PR is meant to make format string works, but won't provide a clean code style. Feel free point out if you think there is a better approach to get things done.

#### TODO list before merge
- [x] Update tests
- [x] Remove `prefix` and `suffix` from `src/module.rs`
- [x] Publish `spongy` or make starship's own syntax parser

#### Motivation and Context

Closes #624

#### Types of changes

#### Limitations
- `{`, `}` have to be written as `%7B`, `%7D` inside curly braces.
- When a module is not displayed (returns `None`) the format of the module is also not displayed.
  This behavior can be changed if we can pass a parameter (e.g. ?default=...) to a module.
- This PR won't add module parameters to `prompt_order`. An ideal form is that, we can specify different parameters for the same module, e.g. `env_var`.
- Conditional styling may not be customized. e.g. In character, we have a prompt character (say ">") that is by default `fg:green` if previous command succeeded and `fg:red` if previous command failed.  
  When it comes to custom format strings, esp. to powerline style,  we may want something like:

  ```toml
  [character]
  format="""\
  ${styled?value=" "&style=bg:(conditional-color)}
  ${styled?value=&style=fg:(conditional-color)}
  """
  [character.where]
  conditional-color = { success: "green", fail: "red" }
  ```

  But this feature won't be implemented in this PR.

- Also, a conditional segment would be good.

  ```toml
  prompt_order = "${git_branch}${when?config=after_git_branch}${when?config=between_git}${git_status}"
  
  [when.after_git_branch]
  # The string will be rendered if git_branch renders, but git_status does not
  git_branch = true
  git_status = false
  format = "${styled?value= &style=fg:purple}"
  
  [when.config.between_git]
  # The string will be rendered if git_branch and git_status renders
  git_branch = true
  git_status = true
  format = "${styled?value=&style=fg:purple bg:cyan}"
  
  [git_branch]
  format = """\
  ${styled?value= &style=fg:white bg:purple}\
  ${name?style=fg:white bg:purple}\
  """
  
  [git_status]
  format = """\
  ${symbols?style=fg:white bg:cyan}
  ${styled?value= &style=fg:cyan}\
  """
  ```
  

```toml
prompt_order = "${battery}${directory}"

[directory]
format = """\
    ${path?style=bg:#d33682 fg:bold white}\
    ${styled?value=&style=fg:#d33682}\
    """

[battery]
format = """\
    ${symbol?style=bg:#eee8d5 fg:bold #657b83}\
    ${percentage?style=bg:#eee8d5 fg:bold #657b83}\
    ${styled?value=% &style=bg:#eee8d5 fg:bold #657b83}\
    ${styled?value=&style=fg:#eee8d5 bg:#d33682}\
    """
```

Screenshot from 2019-12-03 11-15-22

Also we can have this with a complex config file:

Screenshot from 2019-12-08 09-11-24

#### Checklist:


Avi-D-coder:
How does this impact performance?

heyrict:
> How does this impact performance?

Screenshot from 2019-12-07 08-22-03

heyrict:
Almost done. It would be hard to review this PR, but modules can be reviewed one at a time as they are isolated with each other.

matchai:
Closing this for the time being as we finish discussing the implementation in #624 .

PR #691, opened 2019-12-05
fix: Long directory names trigger truncation in Fish

### LINKED ISSUE #585: long directory names trigger truncation
#### Current Behavior

When changing into a directory with a long name, the prompt reverts to ">" prompt for fish shell, as the prompt would have been longer than terminal width, behaviour discussed in fish-shell/fish-shell#904.  Otherwise it wraps.

#### Expected Behavior

Prompt should not revert to ">", and probably shouldn't wrap.  

#### Environment
- Starship version: 0.25.1
- Shell type: fish
- Shell version: fish=3.0.2, zsh==5.4.2, bash==4.4.19
- Shell plugin manager: fisher...
- Terminal emulator: wsltty
- Operating system: windows 10

I think none relevant

#### Starship Configuration

default

#### Possible Solution

Truncate directory names that are too long, for example if we would have this:

```bash
rich in ~/thisisalongdirectorynamesolongitmakesthepromptlongerthanterminalwidth/temp at coolserver
❯ 
```

Truncate like this

```bash
rich in ~/thisisalongdirec.../temp at coolserver
❯ 
```

or even just one letter (I think spaceship did something like this?)

```bash
rich in ~/t/temp at coolserver
❯ 
```

jonstodle:
After counting the module's number of characters, decide whether to print the module or skip it.

#### Description
This introduces a new CLI option, `--max-length` which can be used to specify the maximum length the prompt. If specified it's used to decide whether to print a module. When iterating through the modules their length is determined by way of the new `segments_len` function. If the length exceeds the remaining space the module is not printed.

This is mostly intended as a proof of concept and to see whether it's deemed a proper solution. I've not written any tests, only "anecdotally" tested it in my environment.

#### Motivation and Context
In Fish (shell), if the shell determines the length of the prompt exceeds the number of columns, it will use a default truncated prompt instead. This is to prevent undesired behavior when navigating the cursor in the shell. See the issue below for link to related issue in Fish repo.
Closes #585

#### Types of changes

I've tested this "anecdotally" to see that the proof of concept works.
OS: macOS Catalina (10.15.1 (19B88))
Terminal: Alacritty / Tmux

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1424,6 +1424,7 @@ dependencies = [
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.5.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "unicode-segmentation 1.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unicode-width 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "urlencoding 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "yaml-rust 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ os_info = "1.2.0"
 urlencoding = "1.0.0"
 open = "1.3.2"
 reqwest = "0.9.24"
+unicode-width = "0.1.7"
 
 [dev-dependencies]
 tempfile = "3.1.0"
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -9,7 +9,7 @@ function fish_prompt
     # Account for changes in variable name between v2.7 and v3.0
     set -l CMD_DURATION "$CMD_DURATION$cmd_duration"
     set -l starship_duration (math --scale=0 "$CMD_DURATION / 1000")
-    ::STARSHIP:: prompt --status=$exit_code --keymap=$keymap --cmd-duration=$starship_duration --jobs=(count (jobs -p))
+    ::STARSHIP:: prompt --status=$exit_code --keymap=$keymap --cmd-duration=$starship_duration --jobs=(count (jobs -p)) --max-length=$COLUMNS
 end
 
 # disable virtualenv prompt, it breaks starship
--- a/src/main.rs
+++ b/src/main.rs
@@ -95,7 +95,8 @@ fn main() {
                     .arg(&path_arg)
                     .arg(&cmd_duration_arg)
                     .arg(&keymap_arg)
-                    .arg(&jobs_arg),
+                    .arg(&jobs_arg)
+                    .arg(&max_length_arg),
             )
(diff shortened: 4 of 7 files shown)


chipbuster:
Hey @jonstodle, thanks for the PR!

It looks like your PR is failing our formatter autocheck right now--could you run `cargo fmt` from within the project to run rustfmt? You may need to run `rustup component add rustfmt` to install it first.

Looks like there's another CI failure, but I'm not sure if that's from the rust code or whether there's still some bugs in GH Actions--we'll find out once the format check is resolved.

jonstodle:
Forgot to configure it run automatically on my new dev box.

hedefalk:
I think this is what I need! I came here to the repo to ask about my issue:

I'm using fish in iterm2 on osx and I'm having problems that fish fallbacks to no prompt at all because I have a lot of modules rendering (git, node, java, aws with pretty long role names).

Instead of fallbacking to just `>` in fish I'd get a truncated starfish prompt with just the last module (probably aws) removed if surpassing available columns, is that correct?

jonstodle:
This fix would do that, yes.

If you keep hitting this very often, and don't mind a permanent extra line to your prompt, an extra `line_break` module could work for you. Put the `prompt_order` below in your config and it'll use two lines (always).

```toml
prompt_order = [
    "username",
    "hostname",
    "kubernetes",
    "directory",
    "git_branch",
    "git_commit",
    "git_state",
    "git_status",
    "hg_branch",
    "line_break", // <--- Notice this extra line_break
    "package",
    "dotnet",
    "golang",
    "java",
    "nodejs",
    "php",
    "python",
    "ruby",
    "rust",
    "terraform",
    "nix_shell",
    "conda",
    "memory_usage",
    "aws",
    "env_var",
    "cmd_duration",
    "line_break",
    "jobs",
    "battery",
    "time",
    "character",
]
```

jonstodle:
Sorry to nerd snipe you like this @matchai, but do you think maybe inserting line breaks would be better (instead of dropping the latter modules)?

matchai:
I would personally prefer it truncating modules for my own use, but I think the best solution, that might be worth later investigating, is to do something similar to p10k - dyanmically resizing the directory module: https://asciinema.org/a/258376

jonstodle:
If I understand you correctly, you prefer the version that skips printing modules when the exceed the limit? If so, I can revert the latest commits.

Or would you rather wait for a solution where the modules truncate internally? I'd be interested in taking a stab at that too.

matchai:
I would rather we don't change the truncation behavior between versions, so I guess it comes down to how soon you'd be willing to give it shot. 😛

jonstodle:
I agree that not changing behaviour would be preferable. I'll give give it a go over the holidays.

matchai:
Closing this PR for the time being. Feel free to reopen it if you don't get around to working on directory truncation. 👍

soraxas:
@jonstodle Are there any update on this? The solution seems to be decent and able to address the long prompt issues when there are a lot of information (e.g. virtual environments, git status, long directory name, etc.) being displayed!

PR #694, opened 2019-12-06
fix: Truncate long paths in conda environment names

### LINKED ISSUE #675: Abbreviation of conda environment paths
#### Current Behavior
Unnamed conda environments (e.g. those created via an explicit path, `conda create -p <path> [...]`) do not get abbreviated in the status line.

#### Expected Behavior
I would like `$CONDA_DEFAULT_ENV` to be shortened to everything after the last forward / backward slash, depending on your OS (i.e. to just the basename of the path). So in the screenshot below it would be "_via C my_conda_env_".

Screenshot 2019-11-28 at 14 48 00

#### Environment
- Starship version: 0.26.4
- Shell type: fish
- Shell version: 3.0.2
- Shell plugin manager: oh-my-fish
- Terminal emulator: iTerm2
- Operating system: macOS 10.14.6

#### Possible Solution

Extract the basename of `conda_env` here-ish: https://github.com/starship/starship/blob/f5bbaef82bb092d4e77859544af3f80e82641c9d/src/modules/conda.rs#L12-L13

Or use the directory module to abbreviate it just like the current working directory in the status line.

Scotsguy:
#### Description
Environment names created via `conda create -p [path]` tend to be too long for comfort, so this truncates them.
This uses the same function as the `directory` module, since I moved the function into `utils`.
#### Motivation and Context
Closes #675

#### Types of changes

image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -299,7 +299,7 @@ This does not suppress conda's own prompt modifier, you may want to run `conda c
 
 | Variable            | Default        | Description                                                                                                                                                                                   |
 | ------------------- | -------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `truncation_length` | `1`            | The number of directories the environment path should be truncated to, if the environment was created via `conda create -p [path]`. `0` means no truncation. Also see the `directory` module. |
+| `truncation_length` | `1`            | The number of directories the environment path should be truncated to, if the environment was created via `conda create -p [path]`. `0` means no truncation. Also see the [`directory`](#directory) module. |
 | `symbol`            | `"C "`         | The symbol used before the environment name.                                                                                                                                                  |
 | `style`             | `"bold green"` | The style for the module.                                                                                                                                                                     |
 | `disabled`          | `false`        | Disables the `conda` module.                                                                                                                                                                  |
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -288,15 +288,17 @@ prefix = "underwent "
 ## Conda
 
 The `conda` module shows the current conda environment, if `$CONDA_DEFAULT_ENV` is set.
+
 Note: This does not suppress conda's own prompt modifier, you may want to run `conda config --set changeps1 False`
 
 ### Options
 
-| Variable   | Default        | Description                                  |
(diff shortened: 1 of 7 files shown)


docs/config/README.md
```diff
@@ -288,15 +288,21 @@ prefix = "underwent "
 ## Conda
 
 The `conda` module shows the current conda environment, if `$CONDA_DEFAULT_ENV` is set.
-Note: This does not suppress conda's own prompt modifier, you may want to run `conda config --set changeps1 False`
+
+::: tip
```
matchai: Nice 👍

docs/config/README.md
```diff
@@ -288,15 +288,21 @@ prefix = "underwent "
...
 ### Options
 
-| Variable   | Default        | Description                                  |
-| ---------- | -------------- | -------------------------------------------- |
-| `symbol`   | `"C "`         | The symbol used before the environment name. |
-| `style`    | `"bold green"` | The style for the module.                    |
-| `disabled` | `false`        | Disables the `conda` module.                 |
+| Variable            | Default        | Description                                                                                                                                                                                   |
+| ------------------- | -------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `truncation_length` | `1`            | The number of directories the environment path should be truncated to, if the environment was created via `conda create -p [path]`. `0` means no truncation. Also see the `directory` module. |
```
matchai: _Non-blocking_
This seems like it would be convenient:
```suggestion
| `truncation_length` | `1`            | The number of directories the environment path should be truncated to, if the environment was created via `conda create -p [path]`. `0` means no truncation. Also see the [`directory`](#directory) module. |
```

src/modules/utils/directory.rs
```diff
@@ -0,0 +1,23 @@
+/// Truncate a path to only have a set number of path components
```
matchai: Good idea splitting this out into a utility 👍 
I think it would be best if we also moved the associated unit tests to this file:
https://github.com/Scotsguy/starship/blob/fix-conda-long-paths/src/modules/directory.rs#L238-L278

matchai:
@all-contributors Please add @Scotsguy for code and docs. 😊🎉

PR #696, opened 2019-12-06
feat: cmd_duration module optionally reports milliseconds

### LINKED ISSUE #320: Command duration should report milliseconds for short commands

lucarin91:
#### Description

I defined a new command, `starship time`, which returns time in milliseconds.
The command is then used inside the shell scripts in order to compute the command time with a millisecond resolution.

#### Motivation and Context

Closes #320 

#### Types of changes

#### Checklist:


src/init/starship.bash
```diff
@@ -61,9 +61,9 @@ else
     fi
 
     # Finally, prepare the precmd function and set up the start time.
-    PROMPT_COMMAND="starship_precmd;$PROMPT_COMMAND"
+    PROMPT_COMMAND="$PROMPT_COMMAND;starship_precmd"
```
matchai: I don't know if this change was intentional, but it appears to have broken bash initialization for me:

image

Reverting this change seems to solve the problem for me.

src/configs/cmd_duration.rs
```diff
@@ -14,7 +14,7 @@ pub struct CmdDurationConfig<'a> {
 impl<'a> RootModuleConfig<'a> for CmdDurationConfig<'a> {
     fn new() -> Self {
         CmdDurationConfig {
-            min_time: 2,
+            min_time: 2_000,
```
matchai: 👍

src/main.rs
```diff
@@ -42,7 +44,7 @@ fn main() {
         .short("d")
         .long("cmd-duration")
         .value_name("CMD_DURATION")
-        .help("The execution duration of the last command, in seconds")
+        .help("The execution duration of the last command, in milliseconds")
```
matchai: 👍

src/main.rs
```diff
@@ -108,6 +110,7 @@ fn main() {
                 .arg(&keymap_arg)
                 .arg(&jobs_arg),
         )
+        .subcommand(SubCommand::with_name("time").about("Prints time in milliseconds"))
```
matchai: Since `starship time` is sort-of an implementation detail, I think it would be best if we hid it from `starship --help` with the following Clap setting:

```suggestion
        .subcommand(SubCommand::with_name("time").about("Prints time in milliseconds").settings(&[AppSettings::Hidden]))
```

src/modules/cmd_duration.rs
```diff
@@ -44,14 +44,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 // Render the time into a nice human-readable string
-fn render_time(raw_seconds: u64) -> String {
-    // Calculate a simple breakdown into days/hours/minutes/seconds
+fn render_time(raw_millis: u128) -> String {
```
matchai: Nitpick: "milli" doesn't bring to mind "millisecond" to me intuitively.
I'd be happy using "millisecond" or "ms" for a little more clarity. 😊 

```suggestion
fn render_time(raw_milliseconds: u128) -> String {
```

src/modules/cmd_duration.rs
```diff
@@ -44,14 +44,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    // Calculate a simple breakdown into days/hours/minutes/seconds/milliseconds
+    let (millis, raw_seconds) = (raw_millis % 1000, raw_millis / 1000);
```
matchai: Similarly here ☝️

src/init/starship.bash
lucarin91: I did this change because the cmd_duration feature was broken on my bash configuration. I recently switch to Manjaro Linux and there the default bash configuration sets the  PROMPT_COMMAND variable and breaking the cmd_duration feature.
Changing the order of the concatenation seems to fix the problem for me.

Check my new commit. I hope to have fixed your issue.
In addition to that, I added a check to avoid the concatenation of multiple sharship_cmd function ones after the other in the case of multiple calls to `eval "$(starship init bash)"`.

src/main.rs
lucarin91: Done! thanks.

src/modules/cmd_duration.rs
lucarin91: I was using the same name that the Rust std library use. There is, for instance, the method `as_millis` in the `Duration` struct.
In any case, if you think that this is really misleading I will change it in the next commit.

lucarin91:
Check my last commit.
If the command duration is less than 1 second the milliseconds are hidden by default. I also add a `show_milliseconds` option in order to always show the milliseconds.

If you like this implementation I will also update the docs.

davidkna:
Please also switch the PowerShell init script to miliseconds by replacing `TotalSeconds` with `TotalMilliseconds` (it still needs rounding). The recently introduced ion shell support will also need updating.

matchai:
> If the command duration is less than 1 second the milliseconds are hidden by default. I also add a show_milliseconds option in order to always show the milliseconds.

@lucarin91 I was actually thinking that we would show milliseconds if the command is **<1s** since that means the user set their `min_time` to be <1s, so they were expecting to see an output.

matchai:
My bad. It appears that _is_ the implemented behavior.

PR #704, opened 2019-12-09
feat: Add support for the ion shell

### LINKED ISSUE #678: Add support for the ion shell

Currently starship doesn't support the [ion shell](https://gitlab.redox-os.org/redox-os/ion/). It would be awesome if we could add support for this.

It'd be awesome if we could use a similar setup to what other shells have. So for example being able to just add `eval "$(starship init ion)"` or something similar.

I think it'd be interested in trying to implement this if someone could point me in the right direction. I have only some basic experience with rust, but I think I can figure this out.

Currently it is possible to get it to kind of work with the ion shell. Using the following `PROMPT` function in the `initrc` configuration file:

```ion
fn PROMPT
	starship prompt --status=$? --jobs=$(jobs ^| wc -l)
end

export STARSHIP_SHELL="ion"
```

The `export STARSHIP_SHELL="ion"` is needed, because of issue #211.

BuggStream:
#### Description

This commit adds support for the jobs and status options of the starship prompt.
Support for the cmd-duration is also added if you are using the latest
ion shell. The cmd-duration is implemented using the ION_CMD_DURATION variable.
This variable keeps track of the time spent executing the last command. (This is
similar to the CMD_DURATION of the fish shell).

See this MR for more information: https://gitlab.redox-os.org/redox-os/ion/merge_requests/1176

#### Motivation and Context

Closes #678 

This change is useful for anyone who likes to use ion and starship together.

#### Types of changes

image

- [ ] ~~I have tested using **MacOS**~~
- [ ] ~~I have tested using **Windows**~~

The ion shell is only supported on linux and redox-os.

#### Checklist:

I have updated the documentation and README with references to ion. But I haven't updated the translations, since I don't speak those other languages.

I also haven't added any tests, since I couldn't find any tests that test all the different shells. If there are such tests, let me know and I will try to add them.

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -109,6 +109,10 @@ fi"#,
             );
             Some(script)
         }
+        Some("ion") => {
+            let script = format!("eval $({} init ion --print-full-init)", starship);
+            Some(script)
+        }
         None => {
             println!(
                 "Invalid shell name provided: {}\\n\
@@ -122,7 +126,7 @@ fi"#,
         Some(shell_basename) => {
             println!(
                 "printf \"\\n{0} is not yet supported by starship.\\n\
-                 For the time being, we support bash, zsh, and fish.\\n\
+                 For the time being, we support bash, zsh, fish and ion.\\n\
                  Please open an issue in the starship repo if you would like to \
                  see support for {0}:\\nhttps://github.com/starship/starship/issues/new\"\\n\\n",
                 shell_basename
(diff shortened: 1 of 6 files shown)


src/init/mod.rs
```diff
@@ -109,6 +109,10 @@ fi"#,
             );
             Some(script)
         }
+        Some("ion") => {
+            let script = format!("eval $({} init ion --print-full-init)", starship);
+            Some(script)
+        }
```
Scotsguy: Can't comment on the relevant part, but maybe change L.129: "For the time being, we support bash, zsh and fish."?

BuggStream: I added a reference to `ion` on the line you mentioned and also to the install.sh script. I think this should be all of it (ignoring the documentation in other languages).

charleschege:
Can we have this pull request merged?

BuggStream:
> Can we have this pull request merged?

As far as I am aware everything has been completed.

matchai:
@all-contributors Please add @BuggStream for code and docs. 🎉

PR #714, opened 2019-12-11
fix: Fix shellcheck issues in install.sh

Scotsguy:
#### Description
Running `shellcheck` on install.sh, there were quite a lot of warnings (and some errors).
This fixes them.

Many of these were just SC2059 and SC2155.
Also: SC2034, SC2145 (fixed by fixing SC2059), SC2162, SC2236.

#### Motivation and Context

Imo, even if small, these things would pile up over time if not fixed early.

#### Types of changes

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -131,7 +131,7 @@ detect_arch() {
 
 confirm() {
   if [ -z "${FORCE-}" ]; then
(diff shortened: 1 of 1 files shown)


matchai:
Thank you for your help, @Scotsguy 😊🙏

PR #727, opened 2019-12-13
feat: Add `emberjs` module

Turbo87:
#### Description

This PR introduces a new `emberjs` module that shows the https://emberjs.com/ version as e.g. `with 🐹 v3.14.0`

The implementation relies on the existance of the `ember-source` package in the `node_modules` folder, which should exist for all conventional Ember.js projects.

When working on multiple Ember.js projects it is very useful to see at a glance what Ember.js version is used by the project to figure out what features are available

#### Types of changes

Bildschirmfoto 2019-12-13 um 09 30 24

I ran this locally via `target/debug/starship prompt` in several projects and I've included a few unit tests that should test the basic functionality.

#### Checklist:

I will update the documentation accordingly once the implementation is approved 🙏

### DIFF
--- a/src/configs/emberjs.rs
+++ b/src/configs/emberjs.rs
@@ -1,0 +1,21 @@
+use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
+
+use ansi_term::{Color, Style};
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
+pub struct EmberjsConfig<'a> {
+    pub symbol: SegmentConfig<'a>,
+    pub style: Style,
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for EmberjsConfig<'a> {
+    fn new() -> Self {
+        EmberjsConfig {
+            symbol: SegmentConfig::new("🐹 "),
+            style: Color::Red.bold(),
+            disabled: false,
+        }
+    }
+}
--- a/src/configs/mod.rs
+++ b/src/configs/mod.rs
@@ -5,6 +5,7 @@ pub mod cmd_duration;
 pub mod conda;
 pub mod directory;
 pub mod dotnet;
+pub mod emberjs;
 pub mod env_var;
 pub mod git_branch;
 pub mod git_commit;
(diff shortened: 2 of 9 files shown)


johnletey:
> e.g. `with 🐹 v3.14.0`

@Turbo87 The "🐹" emoji is already used by the golang module. Is there any other emoji that you could use instead?

Turbo87:
@johnletey yeah, I noticed that in the README, but the official mascot of Ember.js is a hamster (see https://emberjs.com/mascots/) so it would be quite unfortunate to have to use something else :-/

I guess technically a gopher is more like a 🐀, but the Go devs probably wouldn't like that emoji that much 😅

Turbo87:
@matchai any thoughts on this PR? :)

matchai:
Truth be told, I'm a little apprehensive due to the prompt in a JS project already being quite long with the NodeJS and Package module. Adding another module is really going to make working with a smaller screen or terminal window difficult, and will also set the precedent for adding framework-specific modules.

If we do decide to proceed with Ember, I wonder if there's a way we can perhaps make this agnostic to JS frameworks:

```toml
[js_library]

	[[js_library.map]]
	dependency = "emberjs"
	style = "bold yellow"
	symbol = "🔥"
	
	[[js_library.map]]
	dependency = "react"
	style = "bold blue"
	symbol = "⚛"
```

Turbo87:
> due to the prompt in a JS project already being quite long with the NodeJS and Package modules

yeah, that makes sense. most Ember apps have their `version` set to `0.0.0` (+ `private: true`), so I'm wondering if we could/should hide the version in that case.

> and will also set the precedent for adding framework-specific modules.

absolutely true. I've been drawing inspiration for this from https://github.com/denysdovhan/spaceship-prompt, which also has an Ember.js module.


> If we do decide to proceed with Ember, I wonder if there's a way we can perhaps make this agnostic to JS frameworks

that is an interesting idea 🤔 

I'm also wondering if we should use something like https://crates.io/crates/node-resolve, or run `node -e "console.log(require('ember-source/package.json').version)"` to get the version more reliably in those cases

matchai:
Closing this in favor of #776. 😊

PR #738, opened 2019-12-15
feat(git_commit): Add `only_detached` option

### LINKED ISSUE #706: [Feature/git_commit] Only show git commit hash in detached HEAD state
## Feature Request

I recently switched to starship and really like it. Thanks!  
The new `git_commit` module is very useful, but I have a feature request, which maybe also others may find useful (as other prompts do it like that).

In version 0.28.0 the `git_commit` module was added. I find the commit hash very useful when in detached HEAD state (no branch checked out, but a specific commit, e.g. while interactively rebasing,...). When I'm on a branch I don't need this information.

IMHO an option to toggle the behavior would be great. E.g. `detached_only`:
* `false` (default): behaves like right now; always shows the current commit hash.
* `true`: only shows the commit hash when in detached HEAD state; not when on a branch.

Just live with it :)

nesmyslny:
#### Description

This adds a new option `only_detached` to the `git_commit` module. It doesn't change default behavior as its default value is `false`. Set to `true`, it will only show the commit hash if the repo is in detached HEAD state; if HEAD is on a branch, it will not output the hash. Also see #706.

#### Motivation and Context

Closes #706 

#### Types of changes

#### Checklist:


johnletey:
@matchai I think that this is a splendid idea! And very useful too! Ship it :rocket:

nesmyslny:
I'm happy to get it merged :) Just let me know if I should change the defaults once there's a consensus or feel free to change it yourself, of course :)

OT: Other prompts have this combined. If on branch show `master`, if detached only show hash `ba18e1a` (instead of `HEAD`). So basically merge the two modules `git_branch` and `git_commit`. But I understand this doesn't work for people, who prefer to always have the hash in the prompt. Or would it be an option to show the branch and the hash in one module?

matchai:
Unfortunately this PR got lost in the mix.
I would still love to see this merged in, but testing it locally, it seems to be showing the hash when I'm on master.

Can anyone else verify that it's working for them?

nesmyslny:
It works here, just checked. Have you set `only_detached = true` for the `git_commit` module? Default behavior is still to always show the hash, but as said before, just let me know if I should change that.  
(Also, I forced pushed (maybe a day) after submitting this PR to fix a logical bug, so make sure to be up-to-date :wink:)

matchai:
There we go! Yep, looked like a blunder on my end. 😅 
Everything looks like it works well. 👍 

Could you please set the default behavior to the following:
```toml
disabled = false
only_detached = true
```

You'll want to update the config, and remove these lines here:
https://github.com/starship/starship/blob/452ea5b2aaa4cdcd0288d43c75b55bba519da03c/src/modules/git_commit.rs#L12-L14

nesmyslny:
Sure, I changed it :)

Maybe someone can have a look at the commit. Especially the tests, if it's ok this way, or needs more work (I only apply the minimal changes to the tests, but maybe we want to change them, so they reflect the default better?).

About the line you suggested to remove: I'm not sure if understand. Isn't that still needed if someone wants to disable the module?

matchai:
> About the line you suggested to remove: I'm not sure if understand. Isn't that still needed if someone wants to disable the module?

That line was there due to us not yet implementing logic to handle disabled-by-default modules. We do, however, have logic to handle manually disabled modules via configuration here: https://github.com/starship/starship/blob/master/src/context.rs#L106-L113

nesmyslny:
Ok, I see. Thanks for explaining! :)

Removed that and rebased the branch on master.

PR #739, opened 2019-12-15
feat: Add config option to clear before printing prompt

### LINKED ISSUE #279: Fish: Using Alt-left/right doesn't update the prompt properly
#### Current Behavior
The Alt-left/right keys are bound to `prevd-or-backward-word` and `nextd-or-backward-word`, respectively, which change the working directory back and forward based on the directory stack. Using these keys leaves artifacts on the first line of the prompt from the previous directory.

#### Expected Behavior
The prompt should clear previous information and display updated information for new working directory.

First line of prompt when I'm in `~/.config`:
```
~/.config
```
After pressing Alt-left:
```
~ .config
```
The `/` was cleared but the rest of the line was not. After pressing enter the prompt is now:
```
~
```
Which is how it should have looked.

#### Environment
- Starship version: `starship 0.13.1`
- Shell type: fish
- Shell version: `fish, version 3.0.2`
- Terminal emulator: Konsole
- Operating system: Arch Linux

#### Starship Configuration
```bash
[directory]
fish_style_pwd_dir_length = 1
```

jonstodle:
Add a config option to use the escape code `\x1b[J` to "clear to the bottom" of the terminal before printing the prompt.

#### Description

This adds a new global config option to enable clearing to the bottom of the the terminal before printing the prompt. This ensures that inline "overwrites" of the prompt clears the previous prompt properly.

#### Motivation and Context

This works around a bug in fish (and possible other shells) where the shell might redraw the prompt without advancing to a new line, or "overwrite" the existing prompt. This is specifically encountered when using the "alt+left/right" shortcut in fish.
Closes #279 

#### Types of changes

As far as I can tell from manual testing it doesn't interfere with other parts of the code.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -61,20 +61,17 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Variable                | Default                       | Description                                                  |
-| ----------------------- | ----------------------------- | ------------------------------------------------------------ |
-| `clear_before_printing` | `false`                       | Clear the terminal to the bottom before printing the prompt. |
-| `add_newline`           | `true`                        | Add a new line before the start of the prompt.               |
-| `prompt_order`          | [link](#default-prompt-order) | Configure the order in which the prompt module occurs.       |
-| `scan_timeout`          | `30`                          | Timeout for starship to scan files (in milliseconds).        |
+| Variable       | Default                       | Description                                            |
+| -------------- | ----------------------------- | ------------------------------------------------------ |
+| `add_newline`  | `true`                        | Add a new line before the start of the prompt.         |
+| `prompt_order` | [link](#default-prompt-order) | Configure the order in which the prompt module occurs. |
+| `scan_timeout` | `30`                          | Timeout for starship to scan files (in milliseconds).  |
 
 ### Example
 
 ```toml
 # ~/.config/starship.toml
 
-# Clear the terminal from the prompt to the bottom of the terminal before printing.
-clear_before_printing = true
 # Disable the newline at the start of the prompt
 add_newline = false
 # Overwrite a default_prompt_order and  use custom prompt_order
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -61,17 +61,20 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Variable       | Default                       | Description                                            |
-| -------------- | ----------------------------- | ------------------------------------------------------ |
-| `add_newline`  | `true`                        | Add a new line before the start of the prompt.         |
-| `prompt_order` | [link](#default-prompt-order) | Configure the order in which the prompt module occurs. |
-| `scan_timeout` | `30`                          | Timeout for starship to scan files (in milliseconds).  |
+| Variable                | Default                       | Description                                                  |
+| ----------------------- | ----------------------------- | ------------------------------------------------------------ |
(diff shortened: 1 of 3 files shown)


johnletey:
@jonstodle Interesting feature! Just some quick feedback, but this doesn't seem to work in iTerm2:

Shot 2019-12-15 at 22 03 50

jonstodle:
It will only clear "to the end of the terminal" from where the cursor is, so any existing text above the will not be cleared.

Demo of prompt clearing

docs/config/README.md
```diff
@@ -61,17 +61,20 @@ This is the list of prompt-wide configuration options.
...
 ### Options
 
-| Variable       | Default                       | Description                                            |
-| -------------- | ----------------------------- | ------------------------------------------------------ |
-| `add_newline`  | `true`                        | Add a new line before the start of the prompt.         |
-| `prompt_order` | [link](#default-prompt-order) | Configure the order in which the prompt module occurs. |
-| `scan_timeout` | `30`                          | Timeout for starship to scan files (in milliseconds).  |
+| Variable                | Default                       | Description                                                  |
+| ----------------------- | ----------------------------- | ------------------------------------------------------------ |
+| `clear_before_printing` | `false`                       | Clear the terminal to the bottom before printing the prompt. |
```
matchai: Would there be any reason for users to set this to `true`?
I can't think of cases where this configuration may be useful.

jonstodle: I can't think of any. My thought was mainly not to potentially break someone's prompt by accident.

matchai: I'd be in favor of having `clear_before_printing` be the default behavior without configuration, at least until we find out it is needed. I have done some local testing in each of the supported shells and haven't run into any issues.

I think it will be otherwise difficult to explain the value of this configuration option to users, and it will just end up being a suggested fix to the rare user that reports a bug on GitHub.

Do you think that makes sense?

jonstodle: Yes.

matchai:
It appears this is causing #824. A better solution might be to have the clear only occur on Fish.

chipbuster:
If we're just printing the one escape character, would it be possible to just do it with an 'echo' command in `fish_prompt`? Not exactly the speediest but it seems easier than cluttering the Rust code with envar checks (which might bug out at a later date, if we restructure how shell detection works)

chipbuster:
Update: #824 and #834 can be fixed by wrapping this print statement in the appropriate shell-specific escape codes (`%{ %}` for zsh and `\[ \]` in bash). The bugs do also affect bash, but not as drastically (mostly just line wrapping being broken--you have to do some pretty aggressive tab-complete stuff in Bash to get it to fail).

This suggests that the right thing to do is to appropriately escape this code instead of restricting it to fish only: there may be uses for it in zsh as well (see e.g. #353, which I haven't tested, but smells similar to #279)

chipbuster:
As an aside, [this link](https://stackoverflow.com/questions/32226139/escaping-zero-length-characters-in-bash) suggests that bash is using the `\[ \]` as placeholders for emitting the ASCII control codes `\x01` and `\x02` ([START OF HEADER and START OF TEXT respectively](http://jkorpela.fi/chars/c0.html)).

If this is in fact the case, we may be able to get rid of our dependence on knowledge of the shell, which I think we only needed to decide the appropriate escape codes anyways.

PR #746, opened 2019-12-16
feat: add Fedora installation instructions

tim77:
#### Description

`starship` packaged and available in official [Fedora repos](https://src.fedoraproject.org/rpms/rust-starship). Now in Rawhide, soon will be for F31.

#### Motivation and Context

Users can easiely discover and install app via their native package managers. This solves https://github.com/starship/starship/issues/545

#### Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,12 @@ The prompt shows information you need while you're working, while staying sleek 
    scoop install starship
    ```
 
+   ##### Fedora
+
+   ```sh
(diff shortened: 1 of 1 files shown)


matchai:
For the time being, we are not adding installation instructions for more specialized or distro-specific package managers, in an effort to keep the docs minimal.

We may create an extended version of the setup docs in the future, which would include `dnf` installation steps. 😊

PR #751, opened 2019-12-17
feat: Implement `starship configure` command

dominikbraun:
#### Description
This is a implementation of the `starship configure` command that I proposed in https://github.com/starship/starship/issues/743. I don't even know if this enhancement is wanted or not, but if you want a preview, just checkout my branch and run `cargo run -- configure` or something.

#### Motivation and Context
Starship is highly configurable and customizing it is a very common thing, at least for me. It is a bit tedious to type `vi ~/.config/starship.toml` each time i want to configure Starship. This PR introduces the flexible and ergonomic `starship configure` command.

**Behavior:** `configure` opens the Starship configuration file in a desired editor. This may be a simple editor like nano or even an editor with a GUI. The command waits for the user to finish editing and then exists.

**Editor configuration and preferences:** Which editor is used depends on the user's configuration. The configured editor in the Starship root configuration has the highest priority:

```toml
add_newline = false
editor = "nano"
```

If the `editor` key does not exist, the `$EDITOR` environment variable will be checked. It has the second-highest priority and if it exists, the Starship configuration file will be opened using `$EDITOR`.

If this environment variable is not set, there's a default editor as fallback which can be found in `configure.rs` and is currently set to `vi`.

Tests and documentation will be updated accordingly in case you want to merge. Feel free to close this PR if you don't want to add the command and don't hesitate to propose any improvements!

#### Types of changes

image

image

#### Checklist:

### DIFF
--- a/src/configs/starship_root.rs
+++ b/src/configs/starship_root.rs
@@ -7,7 +7,6 @@ pub struct StarshipRootConfig<'a> {
     pub add_newline: bool,
     pub prompt_order: Vec<&'a str>,
     pub scan_timeout: u64,
-    pub editor: &'a str,
 }
 
 impl<'a> RootModuleConfig<'a> for StarshipRootConfig<'a> {
@@ -54,7 +53,6 @@ impl<'a> RootModuleConfig<'a> for StarshipRootConfig<'a> {
                 "character",
             ],
             scan_timeout: 30,
-            editor: "",
         }
     }
 }
--- a/src/configs/starship_root.rs
+++ b/src/configs/starship_root.rs
@@ -7,6 +7,7 @@ pub struct StarshipRootConfig<'a> {
     pub add_newline: bool,
     pub prompt_order: Vec<&'a str>,
     pub scan_timeout: u64,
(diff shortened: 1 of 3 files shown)


matchai:
I don't know that having `editor` configuration will be all too useful. I think just using `$EDITOR` is sensible behavior.

dominikbraun:
I removed the `editor` key from the root configuration, we're simply opening the file with `$EDITOR` now.

Scotsguy:
I know I'm late, but consider checking for `$VISUAL` first, then for `$EDITOR`, and then falling back to the hardcoded default: https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference/

Scotsguy:
Also, the command should probably fail if `$HOME` can't be found, instead of editing a literal file `<unknown config>`. I'll make a PR for these myself when I get around to it tomorrow.

dominikbraun:
I'm fine with that. In case you shouldn't get around to it, I can adjust that as well.

PR #756, opened 2019-12-19
fix: Improvements to `starship configure`

Scotsguy:
#### Description
- look for $VISUAL first, then $EDITOR, then the default
- panic if we can't find the home dir

#### Motivation and Context

- [This seems to be the standard](https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference/)
- If we can't find the home dir, there's not much reason to continue
  (originally, this would open a file called `<unknown config>`

#### Types of changes

#### Checklist:

### DIFF
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -15,13 +15,8 @@ pub fn edit_configuration() {
 }
 
 fn get_editor() -> String {
-    match env::var("VISUAL") {
-        Ok(val) => val,
-        Err(_) => match env::var("EDITOR") {
-            Ok(val) => val,
-            Err(_) => STD_EDITOR.to_string(),
-        },
-    }
+    let editor = env::var("VISUAL").or_else(|_| env::var("EDITOR"));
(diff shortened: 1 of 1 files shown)


src/configure.rs
```diff
@@ -15,23 +15,19 @@ pub fn edit_configuration() {
 }
 
 fn get_editor() -> String {
```
matchai: Totally nitpicky, but I would like to strip away the nested indentation here if possible.

I think something along these lines is more easily scannable:
```rust
fn get_editor() -> String {
    let editor = env::var("VISUAL").or_else(|_| env::var("EDITOR"));
    editor.unwrap_or_else(|_| STD_EDITOR.to_string())
}
```

Scotsguy: I'll get on it tomorrow (yay timezones)

matchai:
Perfect! Thank you, @Scotsguy 😊

PR #759, opened 2019-12-19
docs: Add Fedora instructions

ignatenkobrain:


matchai:
Dupe of #746.

For the time being, we are not adding installation instructions for more specialized or distro-specific package managers, in an effort to keep the docs minimal.

We may create an extended version of the setup docs in the future, which would include dnf installation steps. 😊

PR #763, opened 2019-12-20
ci: Use fixed version of setup-php in test workflow

shivammathur:
#### Description
- Use `v1` tag instead of `master`
- Do not specify patch in `php-version` as the latest one is installed

#### Motivation and Context

The action now advocates using `v1` tag.
Versions with patch are not supported, specifying just major and minor like `php-version 7.3` setups the latest patch of PHP for `7.3`

#### Types of changes

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -127,7 +127,7 @@ jobs:
       # Install PHP at a fixed version
       - uses: shivammathur/setup-php@v1
         with:
-          php-version: "7.3.8"
+          php-version: "7.3"
 
       # Install Mercurial (pre-installed on linux, installed from pip on macos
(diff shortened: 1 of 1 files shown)


matchai:
Thank you @shivammathur for setting the `v1` tag. 😊 
Since our tests depend on a specific PHP version to pass, I'll be re-adding the patch number:
https://github.com/starship/starship/blob/master/tests/testsuite/php.rs#L36

shivammathur:
@matchai Using `php-version: "7.3.8"` will not install `PHP 7.3.8`. It will install the latest patch version available.

matchai:
Sorry, I misread your message above.
Thank you for clarifying. 😊

PR #768, opened 2019-12-21
feat: Add ability to mock commands during tests

### LINKED ISSUE #289: Mocking external binaries in Starship tests

Right now, all the tests we're running outside of unit tests act as full-on acceptance tests,  which depend on external binaries to be installed at fixed versions. As a result, our tests are long, brittle, and difficult to write.

By mocking those external applications, we could reduce the complexity of our tests, making the feedback loop considerably faster, and reducing the barrier to entry for contributors. We could also remove the requirement for Docker to be installed when running our test suite, lowering the barrier for contributing.

To ensure we are still compatible with installations across platforms, we can write contract tests that will install binaries on each OS, ensuring that version strings are as we expect them to be.

I think the best solution might be to use dependency injection to mock `std::process::Command`, outputting a fixed version string given the name of a binary.

I'm no Rust expert, so I'd be interested in hearing alternative solutions.

andytom:
#### Description

This adds the ability to mock the output of `utils::exec_cmd` when running tests. Further work will be required to update the existing tests.

#### Motivation and Context

Closes #289

#### Types of changes

#### Checklist:

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -25,7 +25,29 @@ impl PartialEq for CommandOutput {
 }
 
 /// Execute a command and return the output on stdout and stderr if sucessful
+#[cfg(not(test))]
 pub fn exec_cmd(cmd: &str, args: &[&str]) -> Option<CommandOutput> {
+    internal_exec_cmd(&cmd, &args)
+}
+
+#[cfg(test)]
+pub fn exec_cmd(cmd: &str, args: &[&str]) -> Option<CommandOutput> {
+    let command = match args.len() {
+        0 => String::from(cmd),
+        _ => format!("{} {}", cmd, args.join(" ")),
+    };
+
+    match command.as_str() {
+        "dummy_command" => Some(CommandOutput {
+            stdout: String::from("stdout ok!"),
+            stderr: String::from("stderr ok!"),
+        }),
+        // If we don't have a mocked command fall back to executing the command
+        _ => internal_exec_cmd(&cmd, &args),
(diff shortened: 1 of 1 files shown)


matchai:
Looks great so far! 👍

andytom:
I've finally marked this as ready for review sorry about the delay. As mentioned in the description this just add the ability to mock commands when running the tests but I'm going to need some help switching the test to use this as I'm not very familiar with how the modules are printed and how would be best to set up the tests. If anyone has any ideas about how best to do that please let me know and I'll try to pick this up when I get sometime, alternatively please feel free to implement the tests yourself if that is easier.

chipbuster:
I think this will also close #755.

matchai:
I don't think we should be closing any issues until the actual tests are mocked.
This PR just makes it so we can get started on that.

chipbuster:
@matchai Ah, that's my bad, I misunderstood then. For some reason, I thought we were going to implement mocking in this PR once we verified that the mock code was good.

andytom:
@chipbuster and @matchai, yep as it currently stands this just adds the ability to mock the binaries. We can work on updating the tests as part of this branch or do it in separate PRs once this has been merged. I'm ok with either but as mentioned above I need some help with setting up the tests. In addition there are a few modules that still need to be switched over to use the `exec_cmd` (`dotnet`, `php` and `rust`) and switching them and mocking the tests in a single branch/PR might make the review process a bit of a nightmare.

PR #776, opened 2019-12-22
feat: Add `js_library` module 

Turbo87:
#### Description

This PR introduces a new `js_library` module, as suggested in https://github.com/starship/starship/pull/727#issuecomment-567644936

Under the hood it is checking all the specified libraries in parallel by using rayon and the [node-resolve](https://docs.rs/node-resolve/) crate to look for the corresponding `package.json` file.

By default the module is looking for `ember-source` and `react`.

#### Motivation and Context

> Why is this change required? What problem does it solve?

When working on multiple JS projects it is very useful to see at a glance what framework version is used by the project to figure out what features are available

#### Types of changes

Bildschirmfoto 2019-12-13 um 09 30 24

I ran this locally via `target/debug/starship prompt` in several projects and I've included a few unit tests that should test the basic functionality.

#### Checklist:

I will update the documentation accordingly once the implementation is approved 🙏


Turbo87:
@matchai as requested 😉

Turbo87:
@matchai any thoughts on this implementation?

matchai:
It, unfortunately, does look cluttered at this point.
What would be the ideal way to present all this information?

I was thinking that the Node version probably doesn't matter as much if there is a frontend library like Ember or React, but that wouldn't be the case if `js_library` was used for a backend library, like Express. 😕

Turbo87:
@matchai Ember **apps** usually use `version: "0.0.0"` because they are usually not published to npm, which makes the package version display in starship somewhat unnecessary in that case. 

do you think hiding the package version if the `package.json` contains `private: true` or if the version is `0.0.0` is viable for starship?

tests/testsuite/js_library.rs
```diff
@@ -0,0 +1,138 @@
+use ansi_term::Color;
```
matchai: New module tests are now kept alongside the module they test (e.g. https://github.com/starship/starship/blob/master/src/modules/nodejs.rs#L38)

Turbo87: no worries, I'll get it rebased and then move it around :)

Turbo87:
unfortunately I currently don't have time to finish this. if anyone wants to take over feel free to rebase this :)

PR #782, opened 2019-12-23
feat: Update colors of package and php

matchai:
#### Description

This PR changes the colors of two modules:
- `package` red -> green
  Too many of the runtime modules are red, which makes the `package` module hard to read and distinguish from the runtime module. I propose we make it green so it will be a unique color compared to all the runtimes

  image

- `php` red -> purple
  PHP has a very identifiable purple logo. Why not make this module purple instead of _yet another red_ module?

  

#### Checklist:


matchai:
I didn't realize that the `nodejs` module is also green, which would make both `package` and `nodejs` green in a Node.js project. In that case, our color options for `package` can't be red (Rust), green (Node.js), yellow (Python), or purple (PHP).

Maybe a light brown or a blue? There needs to be more colors 😛

chipbuster:
Would it be possible to check the status of `$COLORTERM` or similar and then set different palettes based on that? There's a pretty great overview of terminal color support on [this gist](https://gist.github.com/XVilka/8346728), and it looks like the only super common terminals (that I know of) that don't support true color are Hyper.js, MacOS Terminal, urxvt, and xterm.

I think we'd also want a way to fall back to just the 16-color scheme, since I think those are the only colors overridden by terminal themes (so anything outside of those colors won't be respecting the user's color theme choices)

johnletey:
I personally don't use the PHP module, but love that purple 😍

I wouldn't be opposed to a light brown, but I agree, we definitely need more colours!

matchai:
Maybe orange?

johnletey:
> Maybe orange?

@matchai I'm a fan 😄

PR #784, opened 2019-12-25
fix: Remove trailing semicolon from `$PROMPT_COMMAND`

### LINKED ISSUE #771: PROMPT_COMMAND syntax error near unexpected token `;;'
#### Current Behavior
After each command bash shows following error:

```
-bash: PROMPT_COMMAND: line 8: syntax error near unexpected token `;;'
-bash: PROMPT_COMMAND: line 8: `(_z --add "$(command pwd -P 2>/dev/null)" 2>/dev/null &);;starship_precmd'
```

Also shell prompt does not update (it stays in default state).

#### Expected Behavior

- no error :)
- update PS1 to use starship prompt

#### Environment
- Starship version: 0.32.1
- bash version: GNU bash, version 5.0.11(1)-release (x86_64-pc-linux-gnu)

- Operating system: Manjaro Linux
- Terminal emulator: Termite 

```bash
# for examples

case $- in
    *i*) ;;
      *) return;;
esac

# Attach to existing tmux session
if command -v tmux >/dev/null; then
  if [[ -z "$TMUX" ]] ;then
    ID="$( tmux ls | grep -vm1 attached | cut -d: -f1 )" # get the id of a deattached session
    if [[ -z "$ID" ]] ;then # if not available create a new one
      tmux new-session
    else
      tmux attach-session -t "$ID" # if available attach to it
    fi
  fi
fi

HISTCONTROL=ignoreboth

shopt -s histappend

HISTSIZE=1000
HISTFILESIZE=2000

shopt -s checkwinsize

#shopt -s globstar

#[ -x /usr/bin/lesspipe ] && eval "$(SHELL=/bin/sh lesspipe)"

fi

case "$TERM" in
    xterm-color) color_prompt=yes;;
esac

#force_color_prompt=yes

	color_prompt=yes
    else
	color_prompt=
    fi
fi

else
fi

case "$TERM" in
xterm*|rxvt*)
    ;;
*)
    ;;
esac

    alias ls='ls --color=auto'
fi

# Alias definitions.

[ -f ~/.bash_aliases ] && source ~/.bash_aliases
[ -f ~/.config/bash/colors ] && source ~/.config/bash/colors
[ -f ~/.config/bash/prompt ] && source ~/.config/bash/prompt
[ -f ~/.config/bash/functions ] && source ~/.config/bash/functions
[ -f ~/.config/bash/docker ] && source ~/.config/bash/docker

# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
    . /etc/bash_completion
  fi
fi

[ -f ~/.fzf.bash ] && source ~/.fzf.bash

if test -d ~/.composer/vendor/bin; then
  PATH="$PATH:~/.composer/vendor/bin"
  export PATH
fi

export NVM_DIR="$HOME/.nvm"
[ -f /usr/share/nvm/nvm.sh ] && source "/usr/share/nvm/nvm.sh"

# default editor
export EDITOR=nvim

GPG_TTY=$(/usr/bin/tty)
SSH_AUTH_SOCK=$(gpgconf --list-dir agent-ssh-socket)
export GPG_TTY SSH_AUTH_SOCK

gpgconf --launch gpg-agent

[[ -r "/usr/share/z/z.sh" ]] && source /usr/share/z/z.sh
[[ -f /usr/share/git/git-prompt.sh ]] && . /usr/share/git/git-prompt.sh

eval "$(direnv hook bash)"
eval "$(starship init bash)"
```

#### Starship Configuration

```toml
[hostname]
ssh_only = false
disabled = false

[time]
disabled = false

```

InNoobWeTrust:
Remove any trailing semicolon in $PROMPT_COMMAND before appending starship_precmd, to prevent syntax error.

#### Description

Remove trailing semicolon if exist in PROMPT_COMMAND before appending starship_precmd
#### Motivation and Context

Closes #771 
Latest starship release breaks on my system with direnv and pyenv hooks enabled at login. Here is the error message
```shell
-bash: PROMPT_COMMAND: line 5: syntax error near unexpected token `;;'
-bash: PROMPT_COMMAND: line 5: `_direnv_hook;_pyenv_virtualenv_hook;;starship_precmd'
```
Semicolon is duplicated on the line and make the command become syntactically wrong. The fix is to remove any trailing semicolon from previous `$PROMPT_COMMAND` before appending `;starship_precmd`
#### Types of changes

Just tested the variable assigning, not yet build starship
#### Checklist:

Don't know much, I'm just started using starship recently and haven't read the doc (lol)

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -70,7 +70,8 @@ else
     if [[ -z "$PROMPT_COMMAND" ]]; then
         PROMPT_COMMAND="starship_precmd"
     elif [[ "$PROMPT_COMMAND" != *"starship_precmd" ]]; then
-        PROMPT_COMMAND="$PROMPT_COMMAND;starship_precmd"
+        # Remove any trailing semicolon before appending (PR #784)
+        PROMPT_COMMAND="${PROMPT_COMMAND%;};starship_precmd"
     fi
 fi
 


src/init/starship.bash
```diff
@@ -70,10 +70,10 @@ else
     if [[ -z "$PROMPT_COMMAND" ]]; then
         PROMPT_COMMAND="starship_precmd"
     elif [[ "$PROMPT_COMMAND" != *"starship_precmd" ]]; then
-        PROMPT_COMMAND="$PROMPT_COMMAND;starship_precmd"
+        PROMPT_COMMAND="${PROMPT_COMMAND%;};starship_precmd"
```
InNoobWeTrust: Ref [Posix parameter expansion](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02)

> ${parameter%word}
    Remove Smallest Suffix Pattern. The word shall be expanded to produce a pattern. The parameter expansion shall then result in parameter, with the smallest portion of the suffix matched by the pattern deleted.

InNoobWeTrust:
> 
> 
> Thanks for the PR!
> 
> Two short requests: could you add a comment above the line explaining what the code does and why it's needed? (If you want, you can just reference this PR for additional details).
> 
> Also, could you let us know if this is the bug reported in #771? It seems to me that they're the same thing but it'd be nice to know if the symptoms are exactly the same or slightly different.

I don't quite get what you mean by "just reference this PR for additional details" but updated my first comment.

For second question, it's the same issue, same syntax error which happened on my system. Some hooks just append the last semicolon after them and then we append `;starship_precmd` which make wrong syntax, then the shell refuse to execute the command...

Don't know if we should but seems like we need to consider adding trailing semicolon like others (e.g `;starship_precmd ;`, to have it play nice if activated before others hooks?

chipbuster:
>I don't quite get what you mean by "just reference this PR for additional details" but updated my first comment.

Ah, I'm sorry, I didn't write clearly. I meant to add a comment in the code itself so that anyone reading the source in the future would be able to look up this issue. Thanks for the parameter expansion quote though---that'll be great for anyone referencing this in the future.

> For second question, it's the same issue...

Excellent! I'll modify the issue + PR so that this will automatically close that once merged.

> Don't know if we should but seems like we need to consider adding trailing semicolon like others to have it play nice if activated before others hooks?

Yes, this sounds like a good idea! (though arguably, this means that whoever's writing the other hook would need to check for semicolons on the end...not sure if there's an established convention for this).

I'd personally be in favor of doing this in another PR. Would you be up for that? If not, I can open an issue against it and whoever has time can work on it.

InNoobWeTrust:
> 
> 
> Yes, this sounds like a good idea! (though arguably, this means that whoever's writing the other hook would need to check for semicolons on the end...not sure if there's an established convention for this).
> 
> I'd personally be in favor of doing this in another PR. Would you be up for that? If not, I can open an issue against it and whoever has time can work on it.

I would like to help, but since I mainly use Gerrit and quite inactive on Github so not really used to the way of doing pull request based on another pull request (this pull request). Maybe after this is merged then. 😉

src/init/starship.bash
chipbuster: Just put a comment into the source code (instead of the PR) and we'll be ready to rumble!

PR #788, opened 2019-12-28
refactor(nodejs): Add more matches for files

pizzafox:
#### Description
This adds several other filenames to the matcher for the Node.js module, Like the npm, Yarn, and pnpm package manager lockfiles.

#### Motivation and Context

This will make the matcher for Node.js slightly more effective since it is checking for other Node.js project files that may be present.

#### Types of changes

- [x] I am committing this straight from the GitHub web editor

#### Checklist:
- [ ] I have updated the documentation accordingly ([relevant section](https://starship.rs/config/#nodejs)).
- [ ] I have updated the tests accordingly ([related tests](https://github.com/starship/starship/blob/master/tests/testsuite/nodejs.rs#L45)).

### DIFF
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -17,7 +17,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             "package-lock.json",
             "yarn.lock",
             "pnpm-workspace.yaml",
-            "pnpm-lock.yaml"
+            "pnpm-lock.yaml",
         ])
         .set_extensions(&["js"])
         .set_folders(&["node_modules"])
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -12,7 +12,13 @@ use crate::utils;
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let is_js_project = context
         .try_begin_scan()?
-        .set_files(&["package.json"])
+        .set_files(&[
+            "package.json",
+            "package-lock.json",
+            "yarn.lock",
+            "pnpm-workspace.yaml",
+            "pnpm-lock.yaml"
+        ])
         .set_extensions(&["js"])
         .set_folders(&["node_modules"])
         .is_match();


src/modules/nodejs.rs
```diff
@@ -12,7 +12,13 @@ use crate::utils;
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let is_js_project = context
         .try_begin_scan()?
-        .set_files(&["package.json"])
+        .set_files(&[
+            "package.json",
+            "package-lock.json",
```
matchai: Each of these additions increases the time needed to gather context before printing the prompt.
I'm thinking that the lockfiles might not be necessary (how often do you have `package.json` without a lockfile?)

pizzafox: How much time do you think it would add? I feel like a filesystem operation like that should be super fast, but you're the expert here.

johnletey:
I'm not quite sure if I benchmarked this correctly (@matchai please correct me if I'm wrong), but these new checks adds significant time needed to gather context before printing the prompt:

Shot 2020-01-12 at 08 37 25

matchai:
Those numbers seem unusual to be honest. I wouldn't expect that considerable of a difference from these changes.

johnletey:
> Those numbers seem unusual to be honest.

Now that I think of it, you're right. Even though these numbers are wrong, I still agree with you that there is an increase in the time required by the module, just not quite sure how to show it visually.

davidkna:
I believe the `cargo run` part has some overhead. Also the `cargo run` is running without `--release`. Is the other binary built with or without `--release`? Can you try again with `starship` vs `target/release/starship`?

johnletey:
> Can you try again with `starship` vs `target/release/starship`?

Thanks @davidkna, this seems to give more reasonable benchmarks:

Shot 2020-01-12 at 19 51 42

matchai:
I think the most noticeable difference would be in a directory with many files that _isn't_ a Node project.

chipbuster:
Benchmarks of this branch against its merge-base with master. Hyperfine with default settings, all release builds with rustc 1.40.0.

The CPU is a quad-core E3 Xeon clocked at 3.6 GHz, but the disk is a remote-mounted NFS partition, so this is probably one of the practical worst cases in terms of file access (and in fact, I did see performance issues in the first two runs as the NFS cache adjusted, but it was stable after that).

#### In starship repo, main level, no node files
master: 36.3 ms ±   3.1 ms
branch: 35.2 ms ±   2.7 ms

#### In starship `docs` directory, file `package.json`
master: 42.3 ms ±   3.7 ms
branch: 42.0 ms ±   3.1 ms

#### In directory w/ single file `pnpm-lock.yaml`
master: 4.1 ms ±   0.6 ms
branch: 6.5 ms ±   0.8 ms

#### In directory with 26447 files, no node files
master: 241.4 ms ± 261.7 ms
branch: 363.6 ms ± 669.5 ms

You can see that the NFS hits *hard* on that last benchmark. There are a few hyperfine runs that just freeze in-place altogether (and the max time for both branches is at least 1200ms).

Is this an undue burden any more so than our default module set? I'm not certain...

matchai:
Putting the performance impact aside, what would be the situation where a lockfile is present without a `package.json`?

chipbuster:
Someone else will have to answer that, unfortunately. @pizzafox perhaps? I don't work much with JS projects.

pizzafox:
It's very rare that you would just encounter a lockfile. I figured adding them would help for a few edge cases you might find yourself in, but I'm surprised by the performance cost. Surely a filesystem API call that checks if a file exists should be *super* fast, right?

davidkna:
> It's very rare that you would just encounter a lockfile. I figured adding them would help for a few edge cases you might find yourself in, but I'm surprised by the performance cost. Surely a filesystem API call that checks if a file exists should be _super_ fast, right?
I looked at the code it should only do an extra comparison and not an extra file op. I'm also a bit surprised by the cost. Maybe that code path should switch to a `HashSet` (one for filenames and one for extensions each)?

matchai:
> Maybe that code path should switch to a HashSet (one for filenames and one for extensions each)?

That's what I've done on a rewrite I'm working on. It's definitely not perfect as it is today.

matchai:
Closing this PR for the time being.
Since JS projects shouldn't be missing their `package.json`, this addition may be a little overly defensive.

PR #795, opened 2019-12-30
fix: Consider `$STARSHIP_CONFIG` in `configure`

Scotsguy:
#### Description
Makes `starship configure` consider the `$STARSHIP_CONFIG` variable before falling back to the default of `~/.config/starship.toml`.
#### Motivation and Context

It seems that @dominikbraun and I both forgot this environment variable.
#### Types of changes

#### Checklist:

### DIFF
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -40,11 +40,16 @@ fn get_editor_internal(visual: Option<OsString>, editor: Option<OsString>) -> Os
 }
 
 fn get_config_path() -> OsString {
-    dirs::home_dir()
-        .expect("Couldn't find home directory")
-        .join(".config/starship.toml")
-        .as_os_str()
-        .to_owned()
+    let config_path = env::var_os("STARSHIP_CONFIG").unwrap_or_else(|| "".into());
+    if config_path.is_empty() {
+        dirs::home_dir()
+            .expect("couldn't find home directory")
+            .join(".config/starship.toml")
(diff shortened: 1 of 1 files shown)


src/configure.rs
```diff
@@ -40,11 +40,16 @@ fn get_editor_internal(visual: Option<OsString>, editor: Option<OsString>) -> Os
 }
 
 fn get_config_path() -> OsString {
-    dirs::home_dir()
-        .expect("Couldn't find home directory")
-        .join(".config/starship.toml")
-        .as_os_str()
-        .to_owned()
+    let config_path = env::var_os("STARSHIP_CONFIG").unwrap_or_else(|| "".into());
+    if config_path.is_empty() {
```
dominikbraun: Personally, I'd probably write

```rust
if !config_path.is_empty() {
	config_path
}

dirs::home_dir()
	.expect("couldn't find home directory")
	.join(".config/starship.toml")
	.as_os_str()
	.to_owned()
```

... skipping the `else` block, but that's up to you of course.

Scotsguy: I think the if/else is clearer here (as opposed to get_editor_internal, because there it's two after another)

dominikbraun: I might be a bit Go-influenced here. 😃

Scotsguy: Now that you mention it, that *really* looks like Go, yeah :P

PR #803, opened 2020-01-03
perf: hg_branch, fix: show active branch not previous branch

### LINKED ISSUE #721: hg_branch module thinks i'm in a repo but i'm not
#### Current Behavior
`(no branch)` is appearing on my terminal in every directory under `~` because it thinks everything is an hg repo

#### Expected Behavior
not to appear in folders that aren't hg repos

i have a user level `.hgignore` that it seems to be thinking makes something an hg repo

#### Environment
- Starship version: starship 0.29.0
- Shell type: bash
- Shell version: 4.4.20(1)-release (x86_64-pc-linux-gnu)
- Shell plugin manager:
- Terminal emulator: Terminal
- Operating system: WSL Ubuntu 18.04

#### Possible Solution
hg_branch should not use `.hgignore` to decide something is an hg repo. `.hg` is the only requirement for an hg repo.

dten:
#### Description

perf: don't use `hg` process to get branch / bookmark information, each call was taking 250ms and it was making on average 2 calls.
the branch and bookmark information is available in files in the root of the repo, use that instead.
now takes 300 micro seconds

fix: do not use `.hgignore` to decide if it is an hg repo. was providing false positives. #721 

fix: after changing branch the new branch was not show (the old branch was shown), the new branch is now correctly shown. #722 

fix?: if no branch override was set it would say `(no branch)` however if you commited in that state you would commit to the branch `default` so say that instead

#### Motivation and Context

Closes #721 and #722 
Perf: 402ms -> 373.4µs (1000x faster)

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/hg_branch.rs
+++ b/src/modules/hg_branch.rs
@@ -60,7 +60,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 fn get_hg_branch_name(ctx: &Context) -> String {
     std::fs::read_to_string(ctx.current_dir.join(".hg").join("branch"))
         .map(|s| s.trim().into())
-        .unwrap_or_else(|_| "(no branch)".to_string())
+        .unwrap_or_else(|_| "default".to_string())
 }
 
 fn get_hg_current_bookmark(ctx: &Context) -> Option<String> {
--- a/src/modules/hg_branch.rs
+++ b/src/modules/hg_branch.rs
@@ -11,7 +11,6 @@ use crate::configs::hg_branch::HgBranchConfig;
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let is_hg_repo = context
         .try_begin_scan()?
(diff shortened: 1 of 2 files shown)


matchai:
Not being an `hg` user myself, I would think that the current implementation is able to give repo information wherever you are in the repo, whereas this would make it so that we only get repo information at the root of the repo. Is that correct?

dten:
it's true that `hg` is _capable_  however the module is only enabled if it found `.hg` directory in the current directory. so in reality it only told you before if you were in the hg root

matchai:
Ah, fair. Ideally it would behave similarly to the git modules, but given its current behavior, this is certainly an upgrade.

dten:
Yea sadly hg itself is a little slow to execute and doesn't lend itself to this use. Personally I was having to disable the module as it was causing the prompt to be unusable

matchai:
@all-contributors Please add @dten for code 🎉

dten:
Thanks 🥰

PR #813, opened 2020-01-07
localization: add Spanish as WIP localization

ivanhercaz:
#### Description
Add Spanish as working in progress localization. However, I just commit the link because I have not the icon nor know the font to make the icon of the Spain flag. I say Spain flag because in translate.starship.rs the translation appears as es-ES, although I usually prefer to translate directly to Spanish and not to use flags (this is because I consider the Spanish language uniform and not necessary to split it, but that's my opinion).

Please, don't merge it yet. @matchai, @cappyzawa, as managers of translate.starship.rs, could you help with the flag image? Thanks in advance!

#### Motivation and Context

The motivation is to show to another possible translators or Spanish-speakers that there is a translation in progress for this language.

This is part of #801.

#### Types of changes

None of the above. This is an addition to the README.

#### Checklist:

It isn't necessary.


matchai:
I've gone ahead and added `flag-es` to the pull request 😄

ivanhercaz:
Oh! Really nice. Thank you a lot @matchai! Now it seems to be completed.

PR #819, opened 2020-01-09
docs: Added dark theme to documentation

Brettm12345:
I added a dark theme to the starship documentation. Currently it only supports the `prefers-colorscheme` media query with no manual toggle

### DIFF
--- a/docs/.vuepress/styles/palette.styl
+++ b/docs/.vuepress/styles/palette.styl
@@ -1,4 +1,5 @@
 $accentColor = #DD0B78
+$accentDarkColor = #f757aa
 $textColor = #2c3e50
 $borderColor = #eaecef
 $codeBgColor = #282c34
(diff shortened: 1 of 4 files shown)


matchai:
Woah, great idea!
I'm amazed how easy it is to add dark theme to a vuepress site!

The only issue I see is the accent color for the dark theme being too bright, making the call-to-action hard to read with too low a contrast-ratio:

image
image

Unfortunately the default color is too dark on black:

image
image

If we can force the CTA button to keep the default accent color, it looks like this should be good to merge. 😊

Brettm12345:
I brought it up to 4.57

Brettm12345:
Also here is how I configured Starship for nixos
https://github.com/Brettm12345/nixos-config/blob/master/modules/applications/prompt.nix

docs/.vuepress/styles/palette.styl
```diff
@@ -1,4 +1,5 @@
 $accentColor = #DD0B78
+$accentDarkColor = ##e50a6c
```
matchai: It looks like the site isn't building due to the extra hash here:
```suggestion
$accentDarkColor = #e50a6c
```

Brettm12345:
Fixed. My bad

PR #643, opened 2019-11-15
docs: Simplify installation process and docs

matchai:
#### Description

Add a one-line installer for starship. For the time being, it will install the latest suitable version for the user's platform and architecture.

I also took the opportunity to clean up the installation docs, which have recently gotten cluttered, giving the impression that we have a lengthy installation process at a glance. 

[**📝 Click here to view the new README.md**](https://github.com/starship/starship/blob/update-install/README.md)

#### Motivation and Context

Closes #

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #820, opened 2020-01-09
docs: Explicitly mention 'cargo install' in README

BatmanAoD:
For users who already have Rust/Cargo installed, installation with Cargo may be the preferred approach.

#### Description
The "Install Latest Version" section now contains two subsections, "From binary" (for the existing curlbash command) and "From source (with Rust)" (for the new command).

#### Motivation and Context
Since attempting to go "off the beaten path" when installing software often leads to unintended problems, I installed Starship using a recommended approach (`brew`) before attempting to install with Cargo, even though I generally prefer Cargo. Assuming there is no special reason _not_ to recommend Cargo, I think it's worth mentioning it in the README so that other users don't avoid it unnecessarily.

#### Types of changes

- [x] Documentation of existing feature

I just `brew unlink`ed my first install, then ran `cargo install starship` and confirmed that Starship worked (after re-sourcing my Bash config to update the `PROMPT_COMMAND` and `PS1` to use the new binary path).

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,13 +138,13 @@ The prompt shows information you need while you're working, while staying sleek 
 
    #### Install Latest Version
 
-   ##### From binary:
+   ##### From prebuilt binary, with Shell:
 
    ```sh
    curl -fsSL https://starship.rs/install.sh | bash
    ```
 
-   ##### From source (with Rust):
+   ##### From source on [crates.io](https://crates.io/):
 
    ```sh
    cargo install starship
--- a/README.md
+++ b/README.md
@@ -138,10 +138,16 @@ The prompt shows information you need while you're working, while staying sleek 
 
    #### Install Latest Version
 
-   ##### With Shell:
+   ##### From binary:
 
    ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,12 +138,18 @@ The prompt shows information you need while you're working, while staying sleek
 
    #### Install Latest Version
 
-   ##### With Shell:
```
matchai: I think this should remain "With Shell", so we can add a "With PowerShell" heading when #690 is completed.

BatmanAoD:
@chipbuster Looks like it was deleted when `install.sh` was added: https://github.com/starship/starship/pull/643

README.md
BatmanAoD: How about `Prebuilt Binary, With Shell`?

matchai: `Prebuilt binary with Shell` works for me. 👍

BatmanAoD:
@matchai Better?

I couldn't see whether there's currently a minimum supported Rust version; if there is, that may be worth mentioning, but it could also be a hassle to keep it up-to-date.

PR #676, opened 2019-11-28
refactor: Refactor modules to use the exec_cmd util

andytom:
#### Description

Have refactored the golang, java, nodejs, python, ruby and
username modules to use the new `exec_cmd` util.

I am not very familiar with how the command function is being using
in the `rust` and `dotnet` modules so leaving them for a later, or
someone else if they are more familiar than me. 

#### Motivation and Context

This is just a refactor to reduce some common code.

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #826, opened 2020-01-11
refactor(dotnet): Use `exec_cmd` util

andytom:
#### Description

Have refactored the dotnet module to use the util::exec_cmd rather than
the Command module directly.

#### Motivation and Context

Related to #676, #768 and #289

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/dotnet.rs
+++ b/src/modules/dotnet.rs
@@ -2,11 +2,11 @@ use std::ffi::OsStr;
 use std::iter::Iterator;
 use std::ops::Deref;
 use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
 use std::str;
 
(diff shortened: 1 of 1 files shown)


andytom:
I don't use Dotnet so have been unable to test this locally, the PR passes the tests but I'm not sure if there is any performance impact. @nickwb you initially implemented this hopefully you can can let me know if I've done anything that would inadvertently break the module.

nickwb:
@andytom - I've checked out your branch, built it, and done a little bit of real-world testing. It seems to be working fine, plus all of the tests are passing.

I did a quick benchmark, and it looks Ok, close enough to what it was before.

There's an extra string copy in the new version, so theoretically I'd expect it to be a bit slower - but I think the benefit of mocking and consistency in the code base probably outweighs that.

PR #844, opened 2020-01-17
refactor: replace reqwest with ureq

davidkna:
#### Motivation and Context

reqwest is quite a heavy dependency in terms of file size and pulls in a lot of async related dependencies (tokio) despite starship only using the blocking API.
By replacing reqwest with ureq binary size on windows release builds is reduced by ~30%.

#### Types of changes

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,11 +147,6 @@ version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "bytes"
-version = "0.5.3"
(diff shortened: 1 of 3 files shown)


matchai:
Thank you for your PR, David! 😊

q66:
This breaks starship on any architecture other than x86 and ARM by pulling in `ring` (via `rustls`, which is needed for the tls feature) which is assembly-only without portable fallbacks, and doesn't seem to have any plans to change this, or accept additional platform support. It'd be great if you could reconsider this in favor of some other, better solution.

PR #849, opened 2020-01-17
fix: Wrap clear screen character in unprintable escapes

### LINKED ISSUE #824: With Starship, multiline editing erases previous input
#### Current Behavior
With Starship, multiline editing erases previous input. Additionally, typing tab after `cd` duplicates `cd`.

#### Expected Behavior
Multiline editing should just continue in the next line.

Here's a screencast showing before/after Starship behavior.

issue

Locale and PS1:
```sh
~ ❯ echo $PS1
%{%}~%{%} %{%}❯%{%}

~ ❯ echo $PS1 | od -t x1a  
0000000    1b  5b  4a  25  7b  1b  5b  31  3b  33  36  6d  25  7d  7e  25
         esc   [   J   %   { esc   [   1   ;   3   6   m   %   }   ~   %
0000020    7b  1b  5b  30  6d  25  7d  20  25  7b  1b  5b  31  3b  33  31
           { esc   [   0   m   %   }  sp   %   { esc   [   1   ;   3   1
0000040    6d  25  7d  e2  9d  af  25  7b  1b  5b  30  6d  25  7d  20  0a
           m   %   }   ?  9d   ?   %   { esc   [   0   m   %   }  sp  nl
0000060

~ ❯ locale
LANG="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL=
```

#### Possible Solution
Here's what I tried, to no avail:

 - using Terminal and iTerm2.
 - toggling "Use Unicode version 9+ widths" in iTerm2.
 - used different fonts, with and without ligatures.
 - see the output of `locale`, looks good ad UTFy to me.

#### Environment
- Starship version: 0.33.0
- zsh version: zsh 5.7.1 (x86_64-apple-darwin19.0)
- Operating system: Mac OS 10.15.2

```bash
fpath=(/usr/local/share/zsh-completions $fpath)

eval "$(starship init zsh)"

# The following lines were added by compinstall

zstyle ':completion:*' completer _complete _ignored
zstyle ':completion:*' menu select
zstyle :compinstall filename '/Users/nsabovic/.zshrc'

autoload -Uz compinit
compinit
# End of lines added by compinstall
```

#### Starship Configuration

```toml
add_newline = false
aws.disabled = true
conda.disabled = true
dotnet.disabled = true
golang.disabled = true
hg_branch.disabled = true
java.disabled = true
line_break.disabled = true
nix_shell.disabled = true
nodejs.disabled = true
php.disabled = true
python.disabled = true
ruby.disabled = true
rust.disabled = true
rust.symbol = "⚙️ "
terraform.disabled = true

```

### LINKED ISSUE #834: Cursor location issues only when line_break module is disabled (same as #290 or #110)
#### Current Behavior
Only with `line_break` disabled cursor location shifts when tab completion. For example `ls<TAB>` will leave the prompt as follow:
starship01
The `ls` pointed by the arrow was not manually inserted and the `<BACKSPACE>` is able to delete only the two character inserted leaving the PROMPT as follow:
starship02-after_backspace_to_the_beginning

#### Expected Behavior
Prompt should not interfere with TAB-completion also when `line_break` module is disabled.

#### Environment
- Starship version: 0.33.0
- zsh version: zsh 5.7.1 (x86_64-debian-linux-gnu)
- Operating system: Debian testing

I'm able to reproduce the issue using this minimal `zshrc`:
```bash
export PATH=~/.cargo/bin:$PATH  # for starship itself
eval "$(starship init zsh)"
```

#### Starship Configuration

```toml
add_newline = false

[directory]
truncation_length = 8
style = "cyan"

[git_status]
staged.style = "green"
staged_count.enabled = true
staged_count.style = "green"

# Issue https://github.com/starship/starship/issues/290 seems to be
# reproducible disabling line_break.
[line_break]
disabled = true

```

chipbuster:
#### Description

Minor refactor + bugfix.

One solution to the bugs linked below is to have the clear screen control code wrapped in shell-specific sequences that indicate a zero-length string. Unfortunately, our current escape wrapping code is hardcoded within `module.rs` and cannot be reused.

This PR does two things:
- Moves the wrapping code to its own function and allows it to wrap any specified pair of delimiters (as opposed to being hardcoded to wrap color escape codes previously)
- Uses this new function to escape the clear screen character introduced in #739 

#### Motivation and Context

Closes #824 
Closes #834

#### Types of changes

Tested before + after on Linux, using Konsole, Alacritty, and xterm, in bash, fish, and zsh.

Point of concern: there is still a remnant of this issue present in VSCode terminal when the autocomplete menu pops up in ZSH. I'm not clear whether that's an xterm.js issue or whether this is another bug, but this patch fixes the issue in all the standard terminal emulators I've tried.

EDIT: This bug persists even when rolling back to starship v0.30.0 (before #739 was merged) so I'm inclined to believe it's an xterm.js bug.

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -92,6 +92,23 @@ jobs:
       # Checkout the branch being tested
       - uses: actions/checkout@v2
 
+      # Cache files between builds
+      - name: Cache cargo registry
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache cargo index
+        uses: actions/cache@v1
+        with:
+          path: ~/.cargo/git
+          key: ${{ runner.os }}-cargo-index-${{ hashFiles('**/Cargo.lock') }}
+      - name: Cache cargo build
+        uses: actions/cache@v1
+        with:
+          path: target
+          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
+
       # Install all the required dependencies for testing
       - uses: actions-rs/toolchain@v1
         with:
@@ -110,14 +127,14 @@ jobs:
           go-version: "1.12.1"
 
       # Install Ruby at a fixed version
-      - uses: actions/setup-ruby@v1
+      - uses: eregon/use-ruby-action@v1
         with:
           ruby-version: "2.6.3"
 
       # Install Python at a fixed version
       - uses: actions/setup-python@v1
         with:
-          python-version: "3.7.5"
+          python-version: "3.7.6"
 
       # Install dotnet at a fixed version
       - uses: actions/setup-dotnet@v1
@@ -129,16 +146,12 @@ jobs:
         with:
           php-version: "7.3"
 
-      # Install Mercurial (pre-installed on linux, installed from pip on macos
-      # and from choco on windows),
+      # Install Mercurial (pre-installed on Linux and windows)
       - name: Install Mercurial (macos)
         if: matrix.os == 'macOS-latest'
         env:
           HGPYTHON3: 1
         run: pip install mercurial
-      - name: Install Mercurial (windows)
-        if: matrix.os == 'windows-latest'
-        run: choco install hg
 
       # Install Terraform at a fixed version
       - uses: volcano-coffee-company/setup-terraform@v1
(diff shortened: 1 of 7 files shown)


chipbuster:
It appears #808 is also caused by the same issue, and we can't fix it by wrapping this in escapes.

I'm going to keep most of the changes here and re-introduce this PR as a refactor since I think the escaping logic is a good (if minor) refactoring of code, and then introduce an independent fix for the issues.

nsabovic:
I can confirm that this fixes #824.

PR #851, opened 2020-01-18
feat(kubernetes): Configure module output

Mekk:
Adding two new configuration settings: `show_namespace` and `show_context` which let user manage Kubernetes module display. 

Both settings matter only if namespace is known (if not, we show context name and that's it) and simply enable/disable appropriate segments. So by default (with both set to true) we get `context (namespace)` but we setting one of those to false we can get `context` or `namespace`.

This in particular resolves https://github.com/starship/starship/issues/839 

Warning: this is my first contact with Rust, after briefing some book yesterday.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -692,10 +692,17 @@ threshold = 4
 
 ## Kubernetes
 
-Displays the current Kubernetes context name and, if set, the namespace from
-the kubeconfig file. The namespace needs to be set in the kubeconfig file, this
-can be done via `kubectl config set-context starship-cluster --namespace astronaut`. If the `$KUBECONFIG` env var is set the module will use that if
-not it will use the `~/.kube/config`.
+Displays the current Kubernetes context name and/or the namespace -
+from the kubeconfig file (file pointed by `$KUBECONFIG` env var, or
+`~/.kube/config` by default).
+
+Module output depends whether Kubernetes namespace is configured
+for current context (this can be done via `kubectl config set-context starship-cluster
+--namespace astronaut`):
+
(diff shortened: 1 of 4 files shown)


andytom:
@Mekk, thank you for the PR I've added a few suggestions. We are also having a discussion about shortening name Kubernetes names in #800 if you have any suggestions or thoughts please feel free to get involved.

Mekk:
As I'd like to use this issue also to clarify the approach to further patches: what's best approach to improve the patch (with those styling fixes, and maybe with some extra tweaks): should I just make next commit(s) on the branch, or should I prepare alternative branch from scratch with cleaner fix and offer it as replacement fix, or sth else?

andytom:
> As I'd like to use this issue also to clarify the approach to further patches: what's best approach to improve the patch (with those styling fixes, and maybe with some extra tweaks): should I just make next commit(s) on the branch, or should I prepare alternative branch from scratch with cleaner fix and offer it as replacement fix, or sth else?

I think if you fix the style changes here this should be fine to merge, I think being able to enable/disabled the namespace and context is a nice feature on it's own. If you would like to do some work on #800 I think that should be done on it's own branch.

Mekk:
I tried accepting commit above but it turns out it doesn't compile (str vs String mismatch). Never trusted editing code via browser ;-) Will try to improve it.

samtgarson:
Hey folks, I'd love to be able to disable the namespace portion—any updates on it? 

If any help would be accepted to get it over the line, I'd be happy to volunteer, although it sounds like its just getting that styling fixed and it's good to go? Thanks!

andytom:

PR #861, opened 2020-01-20
feat: Trim "Anaconda, Inc." from Python version

pjbull:
#### Description
The `python --version` command will return "Python 3.6.10 :: Anaconda, Inc." if it is an Anaconda build of Python. This makes the prompt very long. This PR trims that extra text so that the version segment is just the version.

#### Motivation and Context
Having the version of python include this text makes it very long.

#### Types of changes

Here's an example of what I see before this fix:
image

Added a test case with the version string that appears for Anaconda versions of python.

#### Checklist:
Docs say "The `python` module shows the currently installed version of Python." which is still accurate.

Added test including the Anaconda text in the version string

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -73,7 +73,13 @@ fn get_python_version() -> Option<String> {
 }
 
 fn format_python_version(python_stdout: &str) -> String {
-    format!("v{}", python_stdout.trim_start_matches("Python ").trim())
+    format!(
+        "v{}",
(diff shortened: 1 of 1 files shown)


matchai:
@all-contributors Please add @pjbull for code and test. 🙏

matchai:
Whoops! Fixed. 😅 
@all-contributors Please add @pjbull for code and test. 🙏

PR #865, opened 2020-01-22
fix: Restrict clear screen control code printing to fish

chipbuster:
#### Description

#739 caused issues in shells that are not fish. After considering several solutions, I have decided to simply wrap the offending print statement in a fish-specific print statement.

Rejected solutions:
- **Move the clear code into the shell adapter**. The clear code needs to occur after the newline being printed at the start of `get_prompt`, or it doesn't clear the things it's supposed to. This would effectively reopen #279.
- **Escape the clear code with zero-length sequences**. This would not fix #808.

#### Motivation and Context

Closes #834 
Closes #824 
Closes #808

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

This has been tested on Linux using fish, bash, and zsh, with `line_break` both disabled and enabled. The only artifact I notice is that the fish message "hit end of history" when using ALT+left/right is erased when `line_break` is enabled, but I consider this a very minor bug (and it was already present, regardless)

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #866, opened 2020-01-22
refactor: Move ANSI escaping into its own functions

chipbuster:
#### Description

Moves ANSI colorseq escape functionality into its own functions and adds tests.

#### Motivation and Context

One of the proposed changes in #865 involved escaping an ANSI code from within `print.rs`. Currently, escaping these sequences is done inside of a closure inside `module`, which prevents it from being used elsewhere.

This change allows us to escape unprintable characters from anywhere inside starship, as well as allowing a greater variety of escape sequences and tests specific to the escape parser (as opposed to just the module string generator).

#### Types of changes

Refactoring

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -270,7 +270,7 @@ fn get_current_branch(repository: &Repository) -> Option<String> {
     shorthand.map(std::string::ToString::to_string)
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub enum Shell {
     Bash,
     Fish,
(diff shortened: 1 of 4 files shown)


matchai:
LGTM if we can get the conflicts sorted out 👍

chipbuster:
Oh....yeahhhhh :sweat_smile: 

So I tried to rebase this after the shell enum change, failed miserably, and somehow wound up with a pile of code that had merge conflict markers committed into it. I had planned to try again at some point, and I guess I just forgot.

chipbuster:
**headdesk**

What the heck did I just do??

PR #867, opened 2020-01-22
test(nodejs): Port nodejs module tests from E2E to integraton

matchai:
#### Description

Replaces the existing `nodejs` module end-to-end tests with integration tests that don't require preinstalled environmental dependencies.

- Moved the tests to the same file as the module they test
- Created a `render_module` utility function for rendering modules within tests
- Removed Node.js installation during CI setup
- Add `Shell` to `Context` to allow for tests to not run shell-specific code

#### Motivation and Context

This is the first in a long series of PRs making our project easily testable! 💪 

#### Types of changes

#### Checklist:


chipbuster:
This causes a failure on my Linux boxen:

```
---- modules::nodejs::tests::folder_with_node_modules stdout ----
thread 'modules::nodejs::tests::folder_with_node_modules' panicked at 'assertion failed: `(left == right)`
  left: `Some("via \u{1b}[1;32m⬢ v12.0.0\u{1b}[0m ")`,
 right: `Some("via %{\u{1b}[1;32m%}⬢ %{\u{1b}[0m%}%{\u{1b}[1;32m%}v12.0.0%{\u{1b}[0m%} ")`', src/modules/nodejs.rs:85:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- modules::nodejs::tests::folder_with_js_file stdout ----
thread 'modules::nodejs::tests::folder_with_js_file' panicked at 'assertion failed: `(left == right)`
  left: `Some("via \u{1b}[1;32m⬢ v12.0.0\u{1b}[0m ")`,
 right: `Some("via %{\u{1b}[1;32m%}⬢ %{\u{1b}[0m%}%{\u{1b}[1;32m%}v12.0.0%{\u{1b}[0m%} ")`', src/modules/nodejs.rs:73:9

---- modules::nodejs::tests::folder_with_package_json stdout ----
thread 'modules::nodejs::tests::folder_with_package_json' panicked at 'assertion failed: `(left == right)`
  left: `Some("via \u{1b}[1;32m⬢ v12.0.0\u{1b}[0m ")`,
 right: `Some("via %{\u{1b}[1;32m%}⬢ %{\u{1b}[0m%}%{\u{1b}[1;32m%}v12.0.0%{\u{1b}[0m%} ")`', src/modules/nodejs.rs:62:9


failures:
    modules::nodejs::tests::folder_with_js_file
    modules::nodejs::tests::folder_with_node_modules
    modules::nodejs::tests::folder_with_package_json
```

It looks like it's incorrectly picking up on `STARSHIP_SHELL` from my prompt and then adding the ZSH escape sequences into the render.

matchai:
Unfortunately we no longer have the `.env_clear()` part of the module render function in tests, so this is bound to come up. I can't think of a quick solution that doesn't involve adding test-specific logic to the `STARSHIP_SHELL` checks. 😥

matchai:
I think the simplest solution would be to add the value of `STARSHIP_SHELL` to `Context`, so it can be customized within `render_module()`.

chipbuster:
While environment variable lookups are pretty fast, I've never particularly liked the fact that we're reading from an arbitrary source of data at arbitrary points in the program. I'm in favor of moving it to `Context`.

matchai:
The tests should now run the same on all environments.
Should be ready for review. 👀

src/modules/utils/test.rs
```diff
@@ -0,0 +1,12 @@
+use crate::config::StarshipConfig;
+use crate::context::{Context, Shell};
+use std::path::Path;
+
+/// Render a specific starship module by name
+pub fn render_module(module_name: &str, path: &Path) -> Option<String> {
+    let mut context = Context::new_with_dir(clap::ArgMatches::default(), path);
+    context.config = StarshipConfig { config: None };
+    context.shell = Shell::Unknown;
```
matchai: To ensure the prompt renders consistently, we explicitly set the shell here ☝️

src/print.rs
```diff
@@ -73,7 +74,7 @@ pub fn explain(args: ArgMatches) {
         .into_iter()
         .filter(|module| !dont_print.contains(&module.get_name().as_str()))
         .map(|module| {
-            let ansi_strings = module.ansi_strings_for_prompt(false);
+            let ansi_strings = module.ansi_strings();
```
matchai: A drive-by improvement since this is functionally the same.

PR #880, opened 2020-01-26
perf: Optimize files, folder and extension lookups in Context

### LINKED ISSUE #852: Replace the use of Vector<String> with HashSet<String> for Context dir_files
## Feature Request

As suggested by @davidkna in #788, we should probably be using a `HashSet` to store the files, folders, and extensions in the current directory within `Context` for constant lookup times, instead of linear lookup times, increasing as we search for more files.

Lookup times are increasing as more files, folders, and extensions are looked for by various modules.

Replace the use of `Vector<String>` with a `HashSet<String>` for `Context.dir_files`

samwho:
#### Description

Changes the `Vec<PathBuf>` containing all files and folders in context.rs to be a new struct called `DirContents` that optimises for lookup time on various relevant entities, e.g. files, folders and file names.

#### Motivation and Context

Closes #852.

#### Types of changes 
 
#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -169,9 +169,8 @@ impl DirContents {
                     folders.insert(path);
                 } else {
                     if !path.to_string_lossy().starts_with('.') {
-                        path.extension().map(|ext| {
-                            extensions.insert(ext.to_string_lossy().to_string())
-                        });
+                        path.extension()
+                            .map(|ext| extensions.insert(ext.to_string_lossy().to_string()));
                     }
                     files.insert(path);
                 }
@@ -182,7 +181,11 @@ impl DirContents {
             SystemTime::now().duration_since(start).unwrap()
         );
 
-        Ok(DirContents { folders, files, extensions })
+        Ok(DirContents {
+            folders,
+            files,
+            extensions,
+        })
     }
 
     pub fn files(&self) -> impl Iterator<Item = &PathBuf> {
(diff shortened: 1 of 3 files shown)


samwho:
Will work on getting all of the tests to pass over the next week, time allowing. I thought I had run all of the tests 🤔

samwho:
I _think_ this is ready now. The last MacOS test seems to be taking a very long time, but the rest passed.

chipbuster:
I'll want to run more benchmarks on various systems in various configurations, but the single data point I have so far makes this look amazing.

On my home PC (quad-core 2013 CPU, SATA SSD), this change drops the render time for a 100k-file directory (with no special files) from 185ms to 34.5ms.

samwho:
I was going to ask if we have some accepted set of benchmark we use for changes like this. Intuitively it makes sense for this to have a positive effect on large projects, but you're absolutely right to get some numbers. 😄

chipbuster:
We've been trying to get a standard set of benchmarks written up and included with `cargo bench`, but as of right now that's more of a "we should do that someday" item than a reality.

src/context.rs
```diff
@@ -139,26 +138,101 @@ impl<'a> Context<'a> {
...
+    // HashSet of all file names, e.g. the last section without any folders, as strings.
+    file_names: HashSet<String>,
+    // HashSet of all folders, relative to the base directory given at construction.
+    folders: HashSet<PathBuf>,
+    // HashSet of all extensions found, without dots, e.g. "js" instead of ".js".
+    extensions: HashSet<String>,
+}
+
+impl DirContents {
+    fn from_path(base: &PathBuf) -> Result<Self, std::io::Error> {
```
matchai: It looks like this function is unused. Perhaps we should remove it?

samwho: It's used in the tests to avoid having to pass in a bogus Duration. Is that okay?

Akoneto9:
Check mind too

PR #881, opened 2020-01-26
feat: Add clap shell completions

### LINKED ISSUE #1145: Shell autocompletions for starship subcommands/arguments
`starship` should provide/install autocomplete information, so that the shell can complete its subcommands and arguments.

Frequent use cases are:
- typing `starship c<TAB>` should be completed to `starship configure` by the shell
- typing `starship e<TAB>` should be completed to `starship explain` by the shell

It may be possible to emit autocompletion information from `starship init <shell> --print-full-init`, at least for bash.

Typing subcommands by hand. It's not as fast, and requires calls to `--help` when I forget something.

Lucretiel:
#### Description

This PR adds a new subcommand, `completions`, which generates `starship` shell completions. This is a [built-in feature](https://docs.rs/clap/2.33.0/clap/struct.App.html#method.gen_completions_to) for clap, so it's a fairly small change.

#### Motivation and Context

Shell completions are self explanatory: it's helpful for `starship` users to be able to explore and understand the command-line subcommands and their arguments.

Fixes #1145 

#### Types of changes

Screen Shot 2020-01-26 at 1 39 19 PM

Just manual testing for now. Because the output will depend on both `starship`'s CLI design and `clap` completion generator, it was unclear to me how to write a robust integration test for this feature, though I'm open to suggestions or ideas.

#### Checklist:

   - Added help entries to the CLI subcommand. The website doesn't appear to summarize the subcommands, and there doesn't seem to be a man page, so I'm not sure where else (if anywhere) the docs should be updated. This feature is discoverable through the `--help` summary.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,3 +1,5 @@
+use std::io;
+use std::str::FromStr;
 use std::time::SystemTime;
 
 #[macro_use]
@@ -16,7 +18,7 @@ mod segment;
 mod utils;
 
 use crate::module::ALL_MODULES;
-use clap::{App, AppSettings, Arg, SubCommand};
+use clap::{App, AppSettings, Arg, Shell, SubCommand};
 
 fn main() {
     pretty_env_logger::init();
@@ -68,7 +70,7 @@ fn main() {
         .long("print-full-init")
         .help("Print the main initialization script (as opposed to the init stub)");
 
-    let matches =
+    let mut app =
         App::new("starship")
             .about("The cross-shell prompt for astronauts. ☄🌌️")
             // pull the version number from Cargo.toml
@@ -125,7 +127,20 @@ fn main() {
             .subcommand(
                 SubCommand::with_name("explain").about("Explains the currently showing modules"),
             )
-            .get_matches();
+            .subcommand(
+                SubCommand::with_name("completions")
+                    .about("Generate starship shell completions for your shell to stdout")
+                    .arg(
+                        Arg::with_name("shell")
+                            .takes_value(true)
+                            .possible_values(&Shell::variants())
+                            .help("the shell to generate completions for")
+                            .value_name("SHELL")
+                            .required(true),
+                    ),
+            );
+
+    let matches = app.clone().get_matches();
 
     match matches.subcommand() {
         ("init", Some(sub_m)) => {
@@ -161,6 +176,12 @@ fn main() {
             }
         }
         ("explain", Some(sub_m)) => print::explain(sub_m.clone()),
-        _ => {}
+        ("completions", Some(sub_m)) => {
+            let shell_name = sub_m.value_of("shell").expect("Shell name missing.");
+            let shell = Shell::from_str(shell_name).expect("Invalid shell");
+
+            app.gen_completions_to("starship", shell, &mut io::stdout().lock());
+        }
+        (command, _) => panic!("Invalid subcommand: {}", command),
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,4 @@
 use std::io;
-use std::str::FromStr;
 use std::time::SystemTime;
 
 #[macro_use]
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -125,7 +127,20 @@ fn main() {
...
+            .subcommand(
+                SubCommand::with_name("completions")
+                    .about("Generate starship shell completions for your shell to stdout")
+                    .arg(
+                        Arg::with_name("shell")
+                            .takes_value(true)
+                            .possible_values(&Shell::variants())
+                            .help("the shell to generate completions for")
+                            .value_name("SHELL")
+                            .required(true),
```
jonstodle: We could get around this requirement by reading the `STARSHIP_SHELL` env variable by default, and let the user override by explicitly specifying.

src/main.rs
```diff
@@ -161,6 +176,12 @@ fn main() {
...
         }
         ("explain", Some(sub_m)) => print::explain(sub_m.clone()),
-        _ => {}
+        ("completions", Some(sub_m)) => {
+            let shell_name = sub_m.value_of("shell").expect("Shell name missing.");
+            let shell = Shell::from_str(shell_name).expect("Invalid shell");
+
+            app.gen_completions_to("starship", shell, &mut io::stdout().lock());
+        }
+        (command, _) => panic!("Invalid subcommand: {}", command),
```
jonstodle: Is there a reason we'd want to add this? It seems it's not used as _clap_ will catch any invalid sub commands.

src/main.rs
```diff
@@ -161,6 +176,12 @@ fn main() {
             }
         }
         ("explain", Some(sub_m)) => print::explain(sub_m.clone()),
-        _ => {}
+        ("completions", Some(sub_m)) => {
+            let shell_name = sub_m.value_of("shell").expect("Shell name missing.");
+            let shell = Shell::from_str(shell_name).expect("Invalid shell");
```
chipbuster: Is it safe to let clap generate completions for an unknown shell? (I tried looking this up, but I can't seem to find a quick answer)

src/main.rs
Lucretiel: Better correct than sorry, right? It's better to have a panic that'll never happen than have incorrect logic that relies on a check at-a-distance.

Lucretiel: The obvious protection here is that if someone adds a subcommand but doesn't update this `match`, it'll break loudly instead of silently doing the wrong thing.

src/main.rs
Lucretiel: Clap won't generate completions for an unknown shell. The valid shells are in the [`clap::Shell`](https://docs.rs/clap/2.33.0/clap/enum.Shell.html) enum; anything else will trigger an error.

src/main.rs
jonstodle: Makes sense.

Just want to mention that `unreachable!()` might be more expressive, in that we don't actually expect it to be reached.

Lucretiel: Sure, that makes sense; I'll make the.change

andytom:
Hi @Lucretiel, just a friendly reminder about this PR. If you need some help just shout.

Lucretiel:
Sorry, completely lost track of it! Will follow up today

src/main.rs
Lucretiel: Done

src/main.rs
Lucretiel: Is that actually preferable? It doesn't seem like the requirement is bad, and it does seem like the only other place it's used is in `custom command`, where it's used for something unrelated (specifying the shell to run the custom command in). Typically that would be `sh`, even if I'm using something like `fish` as my interactive shell.

matchai: It's used in a few places, including our bug-report mechanism. I'd be in favor of having the default behavior be to generate shell completions for the current shell.

Lucretiel: I agree, but its use in "custom command" implies that `STARSHIP_SHELL` isn't necessarily your current shell? Since it's used to specify the shell that those commands are run in, which is nearly always going to be something like `sh` instead of `fish` or whatever

matchai: Custom commands accept a `shell` value, which will fallback to `STARSHIP_SHELL` if unset. That is because we set `STARSHIP_SHELL` in the Starship initialization script; we don't expect it to be set by the user. It is assumed to be the currently running shell.

Lucretiel: Ahh, I see that. I wasn't looking at the init scripts. Very well, will update.

Lucretiel: Probably this means that `starship init` should be updated too, right? For consistency?

Lucretiel: Fixed

Lucretiel: I confirmed that the `env()` and `required()` interaction is correct; the argument is accepted at the command line or via env var. It's an error if both are absent, or if the input doesn't match one of the possible choices.

PR #884, opened 2020-01-27
fix: escape "$" character to avoid bash interpreting

lucarin91:
#### Description
An alternative approach to fix the issue of PR #658.
I modify the method `ansi_strings_modified` in the `module` in order to replace the character `$` with `\$`. This avoid the possibility to inject code with the git branch name or other sources.

#### Motivation and Context
Alternative solution for PR #658

#### Types of changes

I used the repo [pw3nage](https://github.com/njhartwell/pw3nage) to test the solution.

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -270,7 +270,7 @@ fn get_current_branch(repository: &Repository) -> Option<String> {
     shorthand.map(std::string::ToString::to_string)
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum Shell {
     Bash,
     Fish,
--- a/src/module.rs
+++ b/src/module.rs
@@ -191,18 +191,17 @@ fn ansi_strings_modified(ansi_strings: Vec<ANSIString>, shell: Shell) -> Vec<ANS
                             _ => x.to_string(),
                         }
                     }
-                    MAYBE_ESCAPE_END => {
-                        if escaped {
-                            escaped = false;
-                            match shell {
-                                Shell::Bash => String::from("m\u{5c}\u{5d}"), // => m\]
-                                Shell::Zsh => String::from("m\u{25}\u{7d}"),  // => m%}
-                                _ => x.to_string(),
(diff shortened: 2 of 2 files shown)


wendorf:
Thanks for the PR! I've tested it locally, and it does fix the issue. And much more cleanly than my PR :)

src/context.rs
```diff
@@ -270,7 +270,7 @@ fn get_current_branch(repository: &Repository) -> Option<String> {
     shorthand.map(std::string::ToString::to_string)
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
chipbuster: Thank'ee kindly--I meant to do this in my last PR and just forgot.

src/module.rs
```diff
@@ -191,18 +191,15 @@ fn ansi_strings_modified(ansi_strings: Vec<ANSIString>, shell: Shell) -> Vec<ANS
...
-                            x.to_string()
+                    MAYBE_ESCAPE_END if escaped => {
+                        escaped = false;
+                        match shell {
+                            Shell::Bash => String::from("m\u{5c}\u{5d}"), // => m\]
+                            Shell::Zsh => String::from("m\u{25}\u{7d}"),  // => m%}
+                            _ => x.to_string(),
                         }
                     }
+                    '$' if Shell::Bash == shell => String::from("\\$"),
```
chipbuster: Minor complaint: I'd like to have an explanation for why this is needed. It can either go here or in the block comment at the top of the function--you can even reference the issue number if you want to keep it short. Just something so if someone wants to refer back in the future, they can find some information, e.g. `// Escape $ to avoid `$()` injection attack, see #658 for more`.

lucarin91: Yes sorry, you are right. I will add a documentation comment right now.

chipbuster:
Looks like we're getting smacked by intermittent CI failures again. I've opened an issue to track it--in the meantime, let's just rerun the CI until it works...

PR #890, opened 2020-01-30
fix: Use newline as separator for PROMPT_COMMAND

### LINKED ISSUE #842: Possible regresion of issue 771
#### Current Behavior

Same as described in issue #771 

#### Expected Behavior

Same as described in issue #771

No error with version 0.32.2, with other newests versions, the issue occurs.

#### Environment
- Starship version: 0.33.1
- Shell type: bash
- Shell version: `GNU bash, version 5.0.11(1)-release (x86_64-pc-linux-gnu)`
- Shell plugin manager: bash-it
- Terminal emulator: Tilix
- Operating system: Debian Testing

https://github.com/edubxb/dotfiles/blob/master/.bashrc

#### Starship Configuration
https://github.com/edubxb/dotfiles/blob/master/.config/starship.toml

InNoobWeTrust:
fix: Prevent any further syntax error by separating `starship_precmd` from the current content of PROMPT_COMMAND and any hook that may be appended later by newline character.

#### Description

Previously, `starship_precmd` is terminated by semicolon, which can cause syntax error if other hooks appended later prepend theirs with a semicolon. Instead of semicolon, we can use newline to separate hooks to avoid breaking anything before and after.
#### Motivation and Context

Closes #842 

#### Types of changes

#### Checklist:

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -69,9 +69,9 @@ else
     # add multiple instances of the starship function and keep other user functions if any.
     if [[ -z "$PROMPT_COMMAND" ]]; then
         PROMPT_COMMAND="starship_precmd"
-    elif [[ "$PROMPT_COMMAND" != *"starship_precmd" ]]; then
-        # Remove any trailing semicolon before appending (PR #784)
-        PROMPT_COMMAND="${PROMPT_COMMAND%;};starship_precmd;"
+    elif [[ "$PROMPT_COMMAND" != *"starship_precmd"* ]]; then
+        # Prevent syntax error by separating with newline character (PR #890)
+        PROMPT_COMMAND=${PROMPT_COMMAND:+$PROMPT_COMMAND$'\n'}'starship_precmd'$'\n'
     fi
 fi
 


chipbuster:
@edubxb Would you mind fetching this PR and testing it in a build? Or you can use the tarball linked below: it's a binary release build of this PR.

[starship.tar.gz](https://github.com/starship/starship/files/4142143/starship.tar.gz)
The SHA256sum of the tarball is `922e81587b98ceab03252319464c9a07b05063a77c689ef4b5b41030fe81587f`

Our test suite does not cover stuff that happens in the init file very well, so I'd prefer to make sure that this exact patchset fixes your problem and doesn't cause any additional ones before we merge this. Unfortunately, I've been unable to get your failure to occur on my dev machine, even using your bashrc and starship.toml*, so it would be a great help if you could confirm that this change does work for you.

*I'm almost certain I'm doing something wrong since I don't work with bash much, but I can't work out what

edubxb:
Sure, I will test it!

edubxb:
:sob:

```
bash: PROMPT_COMMAND: line 8: syntax error near unexpected token `;'
bash: PROMPT_COMMAND: line 8: `;preexec_invoke_cmd'
```

But is a Bash-it problem, that [adds](https://github.com/Bash-it/bash-it/blob/7c09252ac86f430b9a6eaeebfc19b20a9f6fe3ce/lib/preexec.bash#L126) a `;` before adding the `preexec_invoke_cmd` function.

InNoobWeTrust:
I'm starting to wonder if we can avoid using `PROMPT_COMMAND` in bash? 🤔
Or just continue thinking about better way to play nice with other hooks...

chipbuster:
I don't think we have much of a choice for bash, unfortunately. For zsh, you have `precmd_functions` which is an actual array (as opposed to character-delimited commands like in bash), but bash only gives us `PROMPT_COMMAND`.

Now the immediate error in #771 can be fixed by using [rcaloras's bash-preexec](https://github.com/rcaloras/bash-preexec), since both starship and bash-it are preexec aware, but this doesn't fix the more general problem of misbehaving frameworks stepping on each others' toes in `PROMPT_COMMAND`.

chipbuster:
@InNoobWeTrust Do you think it would be a possibility to apply @edubxb's fix permanently in starship? 
The fix was to put `PROMPT_COMMAND=${PROMPT_COMMAND/;;/;}` into the initscript.

Or should we potentially just file a bug against bash-it?

romkatv:
> The fix was to put `PROMPT_COMMAND=${PROMPT_COMMAND/;;/;}` into the initscript.

This would break valid configs. For example:

```bash
PROMPT_COMMAND='case "$PWD" in /home/*) do_something;; esac'
```

Or another one:

```zsh
PROMPT_COMMAND='do_something ";;"'
```

chipbuster:
Bah, I had a feeling it might break something but I couldn't come up with an example off the top of my head.

Thanks for the catch.

InNoobWeTrust:
I think we should inform bash-it guys and let's see what they will do, maybe we don't even need this commit.

Or just revert the semicolon terminator commit and hope that most other repos will just append theirs?

For long term, I'm not sure about possibility but it's better if we can avoid using `PROMPT_COMMAND`. Maybe starship will just act as pseudo shell and pipe everything to underlying shell, like tmux? Though there are concerns about compatibilities... 🤔

edubxb:
As a Bash-it (kind of) "regular contributor", I will try to manage the issue personally, opening the issue and probably fixing it (if I can).

edubxb:
Finally I found some time to make the [PR](https://github.com/Bash-it/bash-it/pull/1519) to the Bash-it protect :sweat:

edubxb:
Problem fixed in Bash-it, the PR was merged!

chipbuster:
Awesome! Thanks so much for your help!

I'm going to go ahead and close this PR in the meantime.

PR #892, opened 2020-01-31
fix: make echo test use /bin/sh

kylesferrazza:
#### Description

Changed a test that failed on my machine (NixOS - `echo` is not in `/bin`):
https://github.com/starship/starship/blob/2968aac3ca173954cfc9eaef22c062a6ec0946cc/src/utils.rs#L105
 to use `/bin/sh` like the tests that follow it:
https://github.com/starship/starship/blob/2968aac3ca173954cfc9eaef22c062a6ec0946cc/src/utils.rs#L116
https://github.com/starship/starship/blob/2968aac3ca173954cfc9eaef22c062a6ec0946cc/src/utils.rs#L127

#### Motivation and Context

Test no longer fails when `echo` is not in `/bin`.

#### Types of changes

#### Checklist:

#### Other
I have been using starship for a few months now and am loving it. Looking to start contributing back, and just found this issue when running the tests for the first time.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -102,7 +102,7 @@ mod tests {
 
     #[test]
     fn exec_with_output_stdout() {
-        let result = internal_exec_cmd("/bin/echo", &["-n", "hello"]);
+        let result = internal_exec_cmd("/bin/sh", &["-c", "echo -n hello"]);
         let expected = Some(CommandOutput {
             stdout: String::from("hello"),
             stderr: String::from(""),
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -102,9 +102,9 @@ mod tests {
 
     #[test]
     fn exec_with_output_stdout() {
-        let result = internal_exec_cmd("/bin/sh", &["-c", "echo -n hello"]);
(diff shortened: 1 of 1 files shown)


src/utils.rs
```diff
@@ -102,7 +102,7 @@ mod tests {
 
     #[test]
     fn exec_with_output_stdout() {
-        let result = internal_exec_cmd("/bin/echo", &["-n", "hello"]);
+        let result = internal_exec_cmd("/bin/sh", &["-c", "echo -n hello"]);
```
andytom: This is failing on OSX looks the `-n` argument is getting ignored. If you change this

```suggestion
        let result = internal_exec_cmd("/bin/sh", &["-c", "echo hello"]);
```

src/utils.rs
```diff
@@ -102,7 +102,7 @@ mod tests {
...
         let expected = Some(CommandOutput {
             stdout: String::from("hello"),
```
andytom: and this it should fix it.
```suggestion
             stdout: String::from("hello\n"),
```

PR #895, opened 2020-02-01
revert: "refactor: replace reqwest with ureq (#844)"

matchai:
#### Description

This reverts commit d869b3a09870d432a71d63623bb235b5345e680f, as was merged in #844.

#### Motivation and Context

[As mentioned by @q66](https://github.com/starship/starship/pull/844#issuecomment-581069176):

> This breaks starship on any architecture other than x86 and ARM by pulling in ring (via rustls, which is needed for the tls feature) which is assembly-only without portable fallbacks, and doesn't seem to have any plans to change this, or accept additional platform support.

#### Types of changes

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,7 +2,7 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.7"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "memchr 2.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


davidkna:
Keep in mind the way this is configured it will still use rustls by default but changing that based on demand should be lot easier.
I looked for other alternative lighter http clients that support both openssl or native-ts and rustls backends and the best bet looks like `attohttpc` which currently has an incoming pr for rustls support. Once that has landed I will submit a pr if that library yields similiar space savings.
In the meantime this should be merged.

waldyrious:
> `attohttpc` which currently has an incoming pr for rustls support

Link for convenience: https://github.com/sbstp/attohttpc/pull/51

adamreichold:
@davidkna @waldyrious If you are looking for something with a similar approach to IO but a more minimal API compared to attohttpc, the zeptohttpc crate already contains the above code for rust-tls and native-tls.

davidkna:
I actually liked some of the less minimal API in attohttpc (form support) for our use-case, but with #922 on the horizon it might make sense not to switch to either of these options at all and stay with reqwest if tokio becomes a direct dependency. I know it's a chicken and egg problem but I'd also prefer a library with higher adoption or at leasts something that's been around a little longer.

PR #911, opened 2020-02-05
fix: Change command duration variable to CMD_DURATION

jonstodle:
ion shell has apparently changed the name of the command duration variable at some point from `ION_CMD_DURATION` to `CMD_DURATION`. This breaks the prompt script and presents the following message instead of the prompt:

```
ion: prompt expansion failed: expansion error: Variable does not exist
>>> 
```

#### Description

This changes the variable in the prompt script from `ION_CMD_DURATION` to `CMD_DURATION`.

This could introduce a potential issue where starship would break in older versions of ion. As ion so far has to be compiled from source to use it on MacOS and most LInux distros, there is no _current_ version, so ion users will have starship break whenever they re-compile.

#### Motivation and Context

This makes starship work on version of ion compiled the last month (and a half?) or so.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/init/starship.ion
+++ b/src/init/starship.ion
@@ -4,6 +4,7 @@ fn PROMPT
     let STARSHIP_ION_STATUS = $?
     # Save the ION_CMD_DURATION before running any other commands, but after the status since the
     # duration is not updated during variable assignment
+    # Account for variable name change sometime during December 2019 - January 2020
     let STARSHIP_ION_CMD_DURATION = $or($CMD_DURATION $ION_CMD_DURATION)
     let STARSHIP_ION_CMD_DURATION = $((STARSHIP_ION_CMD_DURATION * 1000))
 
--- a/src/init/starship.ion
+++ b/src/init/starship.ion
@@ -4,7 +4,7 @@ fn PROMPT
     let STARSHIP_ION_STATUS = $?
     # Save the ION_CMD_DURATION before running any other commands, but after the status since the
     # duration is not updated during variable assignment
-    let STARSHIP_ION_CMD_DURATION = $CMD_DURATION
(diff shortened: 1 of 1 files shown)


src/init/starship.ion
```diff
@@ -4,7 +4,7 @@ fn PROMPT
     let STARSHIP_ION_STATUS = $?
     # Save the ION_CMD_DURATION before running any other commands, but after the status since the
     # duration is not updated during variable assignment
-    let STARSHIP_ION_CMD_DURATION = $ION_CMD_DURATION
```
matchai: Do you think it would be worth supporting both, for the time being, to avoid breaking it for folks using an older build of Ion?

We do that for fish right now:
https://github.com/starship/starship/blob/master/src/init/starship.fish#L9-L10

davidkna:
I think `cmd_duration` on ion is still broken by #696 on ion shell so maybe also multiply this new variable by 1000 to convert the duration to ms, it looks like this new `$CMD_DURATION` env var still reports in seconds.

jonstodle:
Didn't find the correct command first time around. Now it falls back to the old variable.

src/init/starship.ion
jonstodle: Didn't find the correct command first time around. Now it falls back to the old variable.

src/init/starship.ion
```diff
@@ -4,7 +4,8 @@ fn PROMPT
...
+    let STARSHIP_ION_CMD_DURATION = $or($CMD_DURATION $ION_CMD_DURATION)
```
matchai: I think we should probably add a comment here explaining why we have both `$CMD_DURATION` and  `$ION_CMD_DURATION`.

PR #914, opened 2020-02-06
feat: Adds support to ask sudo password if needed

### LINKED ISSUE #793: Install script failed due to write permission to /usr/local
The installation script cannot complete successfully by a non-root shell due to the requirement of write permission to /usr/local

### LINKED ISSUE #810: Cannot install starship via curl
I cannot install starship via curl in MacOS version 10.14.3

```
➜  ~ curl -fsSL https://starship.rs/install.sh | sudo bash
Password:

  Configuration
> Bin directory: /usr/local/bin
> Platform:      apple-darwin
> Arch:          x86_64

> Tarball URL: https://github.com/starship/starship/releases/latest/download/starship-x86_64-apple-darwin.tar.gz
? Install Starship latest to /usr/local/bin? [y/N] y
> Installing Starship, please wait…
tar: could not chdir to '/usr/local/bin'


x Command failed (exit code 23): curl --silent --fail --location https://github.com/starship/starship/releases/latest/download/starship-x86_64-apple-darwin.tar.gz

> This is likely due to Starship not yet supporting your configuration.
> please create an issue requesting a build for x86_64-apple-darwin:
> https://github.com/starship/starship/issues/new/
```

### LINKED ISSUE #835: Modify curl install script to escalate privileges if needed

#793 and #810 show that sometimes, the `curl` install fails due to permissions issues.

Modify the install script so that it will escalate privileges if they are needed.

Modify install instructions to use `sudo` in the shell command. While this is arguably simpler for both us and the user, I prefer to explicitly explain what we need escalated privileges for.

djornada:
#### Description

Ask sudo password if needed

#### Motivation and Context

Closes #835 
Closes #793
Closes #810

#### Types of changes

#### Checklist:

- [ ] I have updated the documentation accordingly. (No needed)

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -82,6 +82,19 @@ fetch() {
   fi
 }
 
+install() {
+ local sudo 
+ if [ -w $BIN_DIR ]; then
+	 sudo=''
+ else
+	 info "Root privileges are required to proceed the installation"
+	 sudo='sudo'
+ fi
+ fetch "${URL}" \
+	| ${sudo} tar xzf${VERBOSE} - \
+	-C "${BIN_DIR}"
+}
+
 # Currently supporting:
 #   - win (Git Bash)
 #   - darwin
@@ -240,9 +253,7 @@ confirm "Install Starship ${GREEN}latest${NO_COLOR} to ${BOLD}${GREEN}${BIN_DIR}
 
 info "Installing Starship, please wait…"
 
-fetch "${URL}" \
-  | tar xzf${VERBOSE} - \
-    -C "${BIN_DIR}"
+install
 
 complete "Starship installed"
 echo
--- a/install/install.sh
+++ b/install/install.sh
@@ -83,13 +83,18 @@ fetch() {
 }
 
 install() {
- local sudo 
+ local sudo
+ local msg
  if [ -w $BIN_DIR ]; then
-	 sudo=''
+	 sudo=""
+	 msg="Installing Starship, please wait…"
  else
-	 info "Root privileges are required to proceed the installation"
-	 sudo='sudo'
+	 warn "Escalated permission are required to install to ${BIN_DIR}"
+	 sudo -v || (error "Aborting installation(Pleave provide root password)";exit 1)
+	 sudo="sudo -v"
+	 msg="Installing Starship, please wait…"
  fi
+ info "$msg"
  fetch "${URL}" \
 	| ${sudo} tar xzf${VERBOSE} - \
 	-C "${BIN_DIR}"
@@ -251,11 +256,9 @@ info "Tarball URL: ${UNDERLINE}${BLUE}${URL}${NO_COLOR}"
 check_bin_dir "${BIN_DIR}"
 confirm "Install Starship ${GREEN}latest${NO_COLOR} to ${BOLD}${GREEN}${BIN_DIR}${NO_COLOR}?"
 
-info "Installing Starship, please wait…"
-
 install
-
 complete "Starship installed"
+
 echo
 info "Please follow the steps for your shell to complete the installation:
 
--- a/install/install.sh
+++ b/install/install.sh
@@ -90,9 +90,9 @@ install() {
 	 msg="Installing Starship, please wait…"
  else
 	 warn "Escalated permission are required to install to ${BIN_DIR}"
-	 sudo -v || (error "Aborting installation(Pleave provide root password)";exit 1)
-	 sudo="sudo -v"
-	 msg="Installing Starship, please wait…"
+	 sudo -v || (error "Aborting installation (Please provide root password)";exit 1)
+	 sudo="sudo"
+	 msg="Installing Starship as root, please wait…"
  fi
  info "$msg"
  fetch "${URL}" \
--- a/install/install.sh
+++ b/install/install.sh
@@ -85,7 +85,7 @@ fetch() {
 install() {
  local sudo
  local msg
- if [ -w $BIN_DIR ]; then
+ if [ -w "$BIN_DIR" ]; then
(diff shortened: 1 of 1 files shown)


install/install.sh
```diff
@@ -240,9 +253,7 @@ confirm "Install Starship ${GREEN}latest${NO_COLOR} to ${BOLD}${GREEN}${BIN_DIR}
 
 info "Installing Starship, please wait…"
```
matchai: It may make more sense if we could escalate privileges before saying "Installing, please wait..."

Maybe the flow should be:
```sh
confirm "Install Starship ${GREEN}latest${NO_COLOR} to ${BOLD}${GREEN}${BIN_DIR}${NO_COLOR}?"
check_permissions
info "Installing Starship, please wait…"
install
complete "Starship installed"
```

What are your thoughts?

install/install.sh
```diff
@@ -82,6 +82,19 @@ fetch() {
   fi
 }
 
+install() {
+ local sudo 
+ if [ -w $BIN_DIR ]; then
+	 sudo=''
+ else
+	 info "Root privileges are required to proceed the installation"
```
matchai: I worry that the important visual queue that user interaction is needed isn't very clear with `info`:

image

What are your thoughts on using a blue exclamation mark for "input required"?
image

matchai: Perhaps we should use the opportunity to explain why root privileges are needed, specifying that it's because they don't have write access to that directory:

image

install/install.sh
djornada: Yeah I agree, will move the *info* to install function =)

install/install.sh
djornada: Good catcha! I'll chagne to warn instead info

install/install.sh
```diff
@@ -82,6 +82,19 @@ fetch() {
...
+	 sudo='sudo'
+ fi
+ fetch "${URL}" \
+	| ${sudo} tar xzf${VERBOSE} - \
```
davidkna: I remember some users having trouble with piping stuff directly into sudo and the script being treated as password - maybe do `sudo -v` first here? Correct me if I'm wrong about the semantics here.
Might also want to handle dismissed sudo in a loop or with `exit 1`.
Not sure if this is a priority but maybe missing `sudo` should be handled too either with early exit or fallback to `su`.

djornada: > I remember some users having trouble with piping stuff directly into sudo

Recenttly?

>  maybe do `sudo -v` first here?

Yes! That will be the update

> dismissed sudo

Do you mean command not found?

> fallback to `su`

I'm not sure if I like it cause *su* requires *root* password,  as fallback I preffer to give a message to user provide the bin location for this specific case

davidkna: > Recenttly?

I think it was the issue in #810.

> Do you mean command not found?

Command not found is an issue, but with "dismissed" I meant stuff like failing to provide the proper password, `ctrl+c` and not being in the sudoers file.

> I'm not sure if I like it cause su requires root password, as fallback we can give a message to guide user the needed to provide a bin location for this specific case
What are your thoughts?

I think a message is sufficient since every out-of-the-box setup should have sudo available.

install/install.sh
```diff
@@ -83,13 +83,18 @@ fetch() {
...
  if [ -w $BIN_DIR ]; then
-	 sudo=''
+	 sudo=""
+	 msg="Installing Starship, please wait…"
  else
-	 info "Root privileges are required to proceed the installation"
-	 sudo='sudo'
+	 warn "Escalated permission are required to install to ${BIN_DIR}"
+	 sudo -v || (error "Aborting installation(Pleave provide root password)";exit 1)
+	 sudo="sudo -v"
```
davidkna: I'm pretty sure this won't work and this line needs to be kept as `sudo='sudo'` because with `-v` it will "update user's timestamp **without running a command**" according to `--help`

djornada: > I'm pretty sure this won't work

Do you mean the whole change?

davidkna: No only that line. Btw the error message appears to miss a space.

install/install.sh
```diff
@@ -82,6 +82,24 @@ fetch() {
...
+ if [ -w $BIN_DIR ]; then
+	 sudo=""
+	 msg="Installing Starship, please wait…"
+ else
+	 warn "Escalated permission are required to install to ${BIN_DIR}"
+	 sudo -v || (error "Aborting installation(Pleave provide root password)";exit 1)
+	 sudo="sudo -v"
+	 msg="Installing Starship, please wait…"
+ fi
+ info "$msg"
```
matchai: It looks to me like `$msg` is the same in either case.

install/install.sh
djornada: I didn't get your point? is it only related to the `sudo -v` or the line that check if sudo password was dismissed?

davidkna: Your change sets the `sudo` variable that gets used for `${sudo} tar` to `sudo -v` in L94, but when `sudo` is used with `-v` flag, according to the docs the command will not get executed (here `tar`) and only the `sudo` validation check is performed (my local `sudo` actually drops me into usage help).

install/install.sh
```diff
@@ -82,6 +82,24 @@ fetch() {
   fi
 }
 
+install() {
+ local sudo
+ local msg
+ if [ -w $BIN_DIR ]; then
```
davidkna: Please quote `$BIN_DIR` and change this to `if [ -w "$BIN_DIR" ]; then`

install/install.sh
djornada: Ok now I got it, and I noticed that about the `-v`

PR #915, opened 2020-02-06
feat: Added cabal and hpack support to the haskell module

Brettm12345:
Added support for [cabal](https://www.haskell.org/cabal/) and [hpack](sol/hpack)


matchai:
Oh whoops! Looks like we have some failing tests.
Should be ready to merge as soon as those are fixed. 😊

Brettm12345:
Fixed

PR #918, opened 2020-02-07
fix: Modified Install Script to work in WSL

severinkehding:
Added sudo for the tar command to be able to extract the archive with in WSL propperly

If you want to install starship.rs within the WSL the install.sh script fails because of the missing permissions.
image

#### Description
added sudo prior to the extract command so WSL is able to extract the archive

#### Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

- [ x ] I have tested using **MacOS**
- [ x ] I have tested using **Linux**


matchai:
Thank you for opening a PR though this seems like a duplicate of #914.

PR #920, opened 2020-02-08
style: Added symbol and changed nix-shell color to blue

Brettm12345:
# Before

Before

# After

After


Brettm12345:
I am very new to rust. Please let me know if there is a better way of doing this

tests/testsuite/nix_shell.rs
```diff
@@ -28,7 +28,7 @@ fn pure_shell() -> io::Result<()> {
         .output()?;
     let actual = String::from_utf8(output.stdout).unwrap();
 
-    let expected = format!("via {} ", Color::Red.bold().paint("pure"));
+    let expected = format!("via {} ", Color::Blue.bold().paint(" pure"));
```
romkatv: Is this U+F313? It falls into the *private use area*, meaning that every font is free to display any glyph at all for this code point. It's usually not a good idea to use private use area Unicode code points unless your application controls the font.

matchai: Good point. 👍 
It may be best to use `❄️` here, and add the NerdFont symbol to the NerdFont preset: 
https://github.com/starship/starship/blame/master/docs/presets/README.md#L17

src/modules/nix_shell.rs
```diff
@@ -28,7 +28,14 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     module.set_style(config.style);
 
-    let shell_type = env::var("IN_NIX_SHELL").ok()?;
+    let in_nix_shell = env::var("IN_NIX_SHELL");
```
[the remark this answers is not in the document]
Brettm12345: Thank you!

andytom:
Thanks for the contribution @Brettm12345, unfortunately I didn't notice this PR before merging #1058. I'm going to close this since it looks like this has resolved the issue.

PR #922, opened 2020-02-10
WIP: Asnyc

Brettm12345:
This is my attempt at making this prompt handle data asynchronously. All the shell commands are asynchronous now but the prompt still waits for everything to be finished loading to render.

Ideally I would like to use something like streams to handle how the prompt renders


Brettm12345:
I'm looking for some help on this pull request. I was able to make all the shell functions asynchronous but I'm having a lot of trouble figuring out how to asynchronously render the prompt. I'm having a lot of trouble with how Rust handles ownership and I'd really appreciate some help

matchai:
Hey @Brettm12345 👋 
Thank you for giving async rendering a shot. Unfortunately, few of the prompts Starship supports are capable of async rendering. You should be able to process the modules asynchronously, but I'm not sure what the plan would be after that. Do you have a plan in mind for the rendering step?

tracker1:
Given the relatively small output from the prompt, I'm not sure you gain much value in streaming the output, given you probably want a definitive output order, that side is probably more trouble than it's worth... just making all the lookup calls async is probably a significant boost by itself, and likely would help with things like #831 with just the async lookups.

matchai:
Closing this PR as stale. Feel free to reopen it when you get around to finishing it.

PR #932, opened 2020-02-12
feat: Add singularity module

ntadej:
(Reopened #602)

#### Description
Singularity is one of the tools to run containers. A popular use case is to run it as a development environment, especially for older platforms.

#### Motivation and Context
As one might use several different images it is useful to display the image we are currently in.

Currently the default display is after hostname wrapped in `[` and `]` as it is the shortest. I'm open to suggestions if another defaults are more reasonable.

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1091,6 +1091,33 @@ The module will be shown if any of the following conditions are met:
 symbol = "⚙️ "
 ```
 
+## Singularity
+
+The `singularity` module shows the current singularity image, if inside a container
+and `$SINGULARITY_NAME` is set.
+
+:::
+
+### Options
+
+| Variable   | Default              | Description                                      |
+| ---------- | -------------------- | ------------------------------------------------ |
+| `label`    | `""`                 | Prefix before the image name display.            |
+| `prefix`   | `"["`                | Prefix to display immediately before image name. |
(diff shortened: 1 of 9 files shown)


docs/config/README.md
```diff
@@ -1091,6 +1091,39 @@ The module will be shown if any of the following conditions are met:
...
+### Options
+
+| Variable   | Default              | Description                                      |
+| ---------- | -------------------- | ------------------------------------------------ |
+| `label`    | `""`                 | Prefix before the image name display.            |
+| `prefix`   | `"["`                | Prefix to display immediately before image name. |
+| `suffix`   | `"]"`                | Suffix to display immediately after image name.  |
+| `symbol`   | `""`                 | The symbol used before the image name.           |
+| `style`    | `"bold dimmed blue"` | The style for the module.                        |
+| `disabled` | `true`               | Disables the `singularity` module.               |
```
matchai: I don't think it's necessary for this module to be disabled by default if it is only present when an environment variable exists. The cost of checking for an environment variable is low enough.

Feel free to have it enabled by default. 😄

ntadej: Done.

matchai:
Looks good to me! 👍 
Thank you for your contribution, @ntadej! 🙏

PR #936, opened 2020-02-13
test: Refactor modules to use mocked commands 

andytom:
#### Description

Have refactored the ruby and php modules tests to use a mocked
command and no longer depend on having a particular version of
Ruby or PHP installed. While I was here I also added a quick test for
checking the formatting of the ruby output.

#### Motivation and Context

This will make testing easier and less fragile.

#### Types of changes

#### Checklist:


andytom:
I have been looking at migrating some of the other modules (python, terraform, etc), but it looks like they depend on having some env vars set. I've done a bit off Googling and I'm not sure what the best way to handle them in the test render is, does anyone have any suggestions?

matchai:
I'm afraid it's not going to be easy. We may have create an abstraction around `std::env::var`, and a setter for tests.

For example:
```rust
#[test]
fn check_env_var() -> io::Result<()> {
  // starting now, `getEnvVar("USER")` will return `Some("root")`
  mockEnvVar("USER", "root");

  let actual = render_module("username", dir.path());
}
```

How does that sound to you, @andytom?

src/utils.rs
```diff
@@ -38,6 +38,21 @@ pub fn exec_cmd(cmd: &str, args: &[&str]) -> Option<CommandOutput> {
         0 => String::from(cmd),
         _ => format!("{} {}", cmd, args.join(" ")),
     };
+
```
matchai: Did you mean to include these changes in this PR?

andytom: Doh, no I was going to open a different PR for the Java changes. I'll pop this commit out.

andytom:
> In fact, this may also be the preferable pattern for mocking command execution, as it lets us keep the expected mock context near the test code.
> 
> What are your thoughts, @andytom?

I was thinking something similar, while it made the initial implementation easy the current way of doing things does have it's limitations. In addition to keeping the tests and the mocks closer together it would allow us to set different responses for on a per test basis, this would be useful for Java and Python where some versions print the version strings on `stdout` and others use `stderr`. Also the case statement is getting a bit unwieldy and we haven't even migrated all the other commands in.

PR #961, opened 2020-02-26
fix: check if editor is not found for `configure`

chrisgzf:
#### Description
STD_EDITOR falls back on `vi` if `$EDITOR` and `$VISUAL` is not set.
However, some machines might not have `vi` symlinked or installed and a bad error message is displayed.
image
This commit adds a better error message.
image

#### Motivation and Context
Was trying to configure starship on a fresh Arch Linux install where I:
- have yet to properly set `$EDITOR` and `$VISUAL`
- have `vim` installed but not `vi`
- have yet to symlink `vi` to `vim`

I was pretty confused about the first error message and had to do some digging to resolve the issue, so I decided to improve the error message.

#### Types of changes

See above

#### Checklist:

Not really sure how to write tests for this specific scenario.

Also, these are literally the first few lines of rust I have ever written, so do let me know if I'm doing anything incorrectly or if anything can be improved. Thanks!

### DIFF
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -1,4 +1,5 @@
 use std::env;
+use std::io::ErrorKind;
 use std::ffi::OsString;
 use std::process::Command;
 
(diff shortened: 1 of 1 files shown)


chipbuster:
Thanks for the fix!

I wonder if we could take this a step further: does this have to be a panic if we've identified the error and handled it? It seems to me we could just `eprintln!()` and then call `std::process::exit()` with a non-zero exit code.

chrisgzf:
Hey @chipbuster thanks for the comments! I've made the changes. Let me know if they are okay and if there's anything wrong with the style or anything.

chrisgzf:
Hello @chipbuster, just wanna check if you have any comments on this?

PR #975, opened 2020-03-04
feat: package: Ignore private JS packages

Turbo87:
#### Description

as discussed in https://github.com/starship/starship/pull/776#pullrequestreview-368933093, this PR adjusts the `package` module to ignore JS packages that have specified `private: true` in their `package.json` file.

#### Motivation and Context

see https://github.com/starship/starship/pull/776#issuecomment-594555121

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -41,6 +41,11 @@ fn extract_cargo_version(file_contents: &str) -> Option<String> {
 
 fn extract_package_version(file_contents: &str) -> Option<String> {
     let package_json: json::Value = json::from_str(file_contents).ok()?;
+
+    if package_json.get("private").and_then(json::Value::as_bool) == Some(true) {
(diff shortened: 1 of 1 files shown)


src/modules/package.rs
```diff
@@ -174,6 +179,40 @@ mod tests {
             extract_package_version(&package_without_version),
             expected_version
         );
+
```
chipbuster: Would it be possible to split this off into its own test? It's a lot easier to see in CI that `test_extract_private_package_version` failed than it is to have to dig through line number/backtrace info to figure out exactly which test in here failed.

Turbo87: sure, was just following what was already there, but I'm happy to split them apart :)

Turbo87: done 👍

PR #993, opened 2020-03-10
feat: Make git_branch and package prefix configurable

jnskw:
#### Description
I added `prefix` to the config for the `git_branch` and the `package` module. The default config still uses the previous values `on ` and `is ` respectively.

#### Motivation and Context
Using prefixes for every module can be very verbose. While some people are asking for more sophisticated methods to format the prompt (#346), this change aims provide a simple fix in the meantime.

#### Types of changes

`starship master v0.37.0` (a simple prompt with prefixes set to "")

#### Checklist:


jnskw:
@johnletey thanks for the review.
That sounds reasonable. Doing it manually was not fun. I will remove that part.

matchai:
Closing this PR as we won't be adding configuration for every module's prefix and suffix individually.

We're working on a more robust solution over at #624. Feel free to join the conversation. 😊

PR #996, opened 2020-03-11
feat: add docker context module

### LINKED ISSUE #995: Display Current Docker Context
## Feature Request

I hop between Docker contexts a lot when deploying and managing services running in different environments.

Sometimes I forget to check or change my Docker context and end up deploying stuff to the wrong server!

So, it would be really neat to have the currently active Docker context available in the prompt!

### How

Luckily it's super simple: read the `currentContext` key from the `config.json` file, which is (usually, I think) located in `$HOME` (`C:\Users\Username\` or `/home/Username/` or `/Users/Username/`)

Southclaws:
#### Description

Adds a simple module that checks for a Docker config file and if present, reads the `currentContext` value out and displays on the prompt with a whale.

#### Motivation and Context
Closes #995 

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -387,12 +387,12 @@ The `docker_context` module shows the currently active
 
 ### Options
 
-| Variable                | Default       | Description                                                                  |
-| ----------------------- | ------------- | ---------------------------------------------------------------------------- |
-| `symbol`                | `"🐳 "`       | The symbol used before displaying the Docker context .                       |
-| `style`                 | `"bold blue"` | The style for the module.                                                    |
-| `only_with_compose_yml` | `false`       | Only show when there's a `docker-compose.yml` file in the current directory. |
-| `disabled`              | `true`        | Disables the `docker_context` module.                                        |
+| Variable          | Default       | Description                                                                  |
+| ----------------- | ------------- | ---------------------------------------------------------------------------- |
+| `symbol`          | `"🐳 "`       | The symbol used before displaying the Docker context .                       |
+| `style`           | `"bold blue"` | The style for the module.                                                    |
+| `only_with_files` | `false`       | Only show when there's a `docker-compose.yml` file in the current directory. |
+| `disabled`        | `true`        | Disables the `docker_context` module.                                        |
 
 ### Example
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -100,6 +100,7 @@ prompt_order = [
     "git_state",
     "git_status",
     "hg_branch",
+    "docker_context",
     "package",
     "dotnet",
     "elixir",
@@ -376,6 +377,29 @@ a single character. For `fish_style_pwd_dir_length = 2`, it would be `/bu/th/ci/
 
 [directory]
 truncation_length = 8
+```
+
+## Docker Context
+
+The `docker_context` module shows the currently active
+[Docker context](https://docs.docker.com/engine/context/working-with-contexts/) if it's not set to
+`default`.
+
+### Options
+
+| Variable   | Default       | Description                                            |
+| ---------- | ------------- | ------------------------------------------------------ |
+| `symbol`   | `"🐳 "`       | The symbol used before displaying the Docker context . |
+| `style`    | `"bold blue"` | The style for the module.                              |
+| `disabled` | `true`        | Disables the `docker_context` module.                  |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[docker_context]
+symbol = "🐋 "
 ```
 
 ## Dotnet
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -387,11 +387,12 @@ The `docker_context` module shows the currently active
 
 ### Options
 
-| Variable   | Default       | Description                                            |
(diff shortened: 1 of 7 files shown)


andytom:
Thanks for the contribution @Southclaws.
 Would it make sense to check the current working directory and enable the module if it contains a docker or docker compose file?

Southclaws:
That's probably worth adding an option for as it's not something I'd use - I often run docker commands outside a directory with a compose file in (non-compose things like `docker ps` etc)

Southclaws:
Done, however I can't compile or test because Cargo/crates.io has just completely stopped working.

src/modules/docker_context.rs
```diff
@@ -0,0 +1,48 @@
...
+const DOCKER_CONFIG_FILE: &str = ".docker/config.json";
+
+/// Creates a module with the currently active Docker context
+///
+/// Will display the Docker context if the following criteria are met:
+///     - There is a file named `$HOME/.docker/config.json`
+///     - The file is JSON and contains a field named `currentContext`
+///     - The value of `currentContext` is not `default`
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module(DOCKER_CONFIG_FILE);
```
andytom: I think this should be
```suggestion
    let mut module = context.new_module("docker_context");
```

src/modules/docker_context.rs
```diff
@@ -0,0 +1,48 @@
...
+    let config: DockerContextConfig = DockerContextConfig::try_load(module.config);
+
+    if config.only_with_compose_yml
+        && !context
+            .try_begin_scan()?
+            .set_files(&["docker-compose.yml"])
```
andytom: Would it be useful to also enable the module if the dir has a `Dockerfile` too?
```suggestion
            .set_files(&["docker-compose.yml", "Dockerfile"])
```

If you make this change then maybe it would be better to change the `only_with_compose_yml` config var to something more generic maybe `only_with_docker_files` or `only_with_files`?

src/modules/docker_context.rs
```diff
@@ -0,0 +1,48 @@
...
+            .is_match()
+    {
+        return None;
+    }
+
+    let config_path = home_dir()?.join(".docker/config.json");
```
andytom: Should this be

```suggestion
    let config_path = home_dir()?.join(DOCKER_CONFIG_FILE);
```

docs/config/README.md
```diff
@@ -378,6 +379,30 @@ a single character. For `fish_style_pwd_dir_length = 2`, it would be `/bu/th/ci/
...
+`default`.
+
+### Options
+
+| Variable                | Default       | Description                                                                  |
+| ----------------------- | ------------- | ---------------------------------------------------------------------------- |
+| `symbol`                | `"🐳 "`       | The symbol used before displaying the Docker context .                       |
+| `style`                 | `"bold blue"` | The style for the module.                                                    |
+| `only_with_compose_yml` | `false`       | Only show when there's a `docker-compose.yml` file in the current directory. |
+| `disabled`              | `true`        | Disables the `docker_context` module.                                        |
```
andytom: We generally put the `style` and `disabled` variables at the bottom of the table.
```suggestion
| `only_with_compose_yml` | `false`       | Only show when there's a `docker-compose.yml` file in the current directory. |
| `style`                 | `"bold blue"` | The style for the module.                                                    |
| `disabled`              | `true`        | Disables the `docker_context` module.                                        |
```

Southclaws:
Ah, yes, sorry I completely forgot about this PR! Will clean up today 👍

docs/config/README.md
```diff
@@ -378,6 +379,30 @@ a single character. For `fish_style_pwd_dir_length = 2`, it would be `/bu/th/ci/
...
+The `docker_context` module shows the currently active
+[Docker context](https://docs.docker.com/engine/context/working-with-contexts/) if it's not set to
+`default`.
+
+### Options
+
+| Variable          | Default       | Description                                                                  |
+| ----------------- | ------------- | ---------------------------------------------------------------------------- |
+| `symbol`          | `"🐳 "`       | The symbol used before displaying the Docker context .                       |
+| `only_with_files` | `false`       | Only show when there's a `docker-compose.yml` file in the current directory. |
```
andytom, line 393:
```suggestion
| `only_with_files` | `false`       | Only show when there's a `docker-compose.yml` or `Dockerfile` in the current directory. |
```

src/configs/docker_context.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+impl<'a> RootModuleConfig<'a> for DockerContextConfig<'a> {
+    fn new() -> Self {
+        DockerContextConfig {
+            symbol: SegmentConfig::new("🐳 "),
+            context: SegmentConfig::default(),
+            style: Color::Blue.bold(),
+            only_with_files: false,
+            disabled: true,
```
andytom, line 22: What do you think about switching these so the module is enabled by default but only when there is are some docker related files?

```suggestion
            only_with_files: true,
            disabled: false,
```

Southclaws:
Thanks! Those suggestions made sense. LTGM! 👍

PR #997, opened 2020-03-11
docs: Add details about musl build to readme

### LINKED ISSUE #1052: Cannot get running under Centos
#### Current Behavior
Cannot get it to work under Centos 7. Whether installed via binary, Cargo, Snap, or otherwise, upon launch I get the error:

```
starship: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

#### Expected Behavior
To not get that error

It's looking for shared libraries in `/usr/local/lib64` that exist only in `/usr/lib64`

#### Possible Solution
Change the way the code looks for libraries under Centos

#### Environment
- Starship version: 0.38.1
- Shell type: bash
- Shell version: 4.2.46(2)-release
- Shell plugin manager: none
- Terminal emulator: iTerm
- Operating system: Centos 7.7.1908

```bash
# .bashrc

# User specific aliases and functions

alias rm='rm -i'
alias cp='cp -i'
alias mv='mv -i'

# Source global definitions
if [ -f /etc/bashrc ]; then
	. /etc/bashrc
fi
. "/root/.acme.sh/acme.sh.env"
eval "$(starship init bash)"
```

#### Starship Configuration
Does not exist

Daniel15:
## Description
Added an explanation of the musl build to the readme, in case anyone else tries to use starship with CentOS 7

#### Motivation and Context
Just a documentation update.
References #342
Fixes #1052

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,14 +147,6 @@
    ```sh
    curl -fsSL https://starship.rs/install.sh | bash
    ```
-   
-   ##### From prebuilt binary, Linux `musl` build
-   
-   If you get an error like "*version 'GLIBC_2.18' not found (required by starship)*" when using the prebuilt binary (for example, on CentOS 6 or 7), you can use a binary compiled with `musl` instead of `glibc`:
-   
-   ```sh
-   curl -fsSL https://starship.rs/install.sh | bash -s -- --platform unknown-linux-musl
-   ```
 
    ##### From source on [crates.io](https://crates.io/):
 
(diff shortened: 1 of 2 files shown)


chipbuster:
Thanks for the PR!

I'm in favor of adding this info to the docs, but I think putting it in the main README is going to cause some headaches for poor matchai when he needs to refactor the README *again* to keep its length down (we've already had several rounds of that).

Do you think this would be appropriate as a section in the FAQ instead?

Daniel15:
I think it'd be okay to just be on the website, however the installation instructions on the site seem to pull from the readme.

chipbuster:
I believe `docs/faq/README.md` maps to https://starship.rs/faq/ on the website.

PR #999, opened 2020-03-11
feat: replace reqwest with attohttpc

### LINKED ISSUE #933: Dependency on `ring` breaks build on non-x86/ARM systems, even after going back to reqwest
It seems even after reverting the `reqwest` replacement (https://github.com/starship/starship/pull/844) there is still a dependency on `ring` right now. Running `cargo tree` indicates that it comes in via `reqwest`->`hyper-rusttls`->`ct-logs`->`sct`->`ring`.

It seems the reason for this is using `rustls-tls` feature for `reqwest`; removing it from `features` (leaving only `blocking`) fixes the problem.

There shouldn't be an issue building `openssl` on musl unless the `vendored` feature is used, which shouldn't generally be used, and I don't think it's used by default, as it uses `openssl-sys` with it...

Tried on 6 different PowerPC variants (64le/64/32, `glibc` and `musl` for each, ~~ppc32/musl segfaults rustc but that's a compiler bug and seen elsewhere too~~ fixed with rust 1.41) and it compiled fine on all of them. Also built on more conventional architectures like x86_64, with `glibc` and `musl`, in the Void Linux buildsystem.

davidkna:
#### Description

`reqwest` is a heavy dependency. Replace it with the lighter alternative `attohttpc` (-35% binary size on windows release build).

Also (fixes #933):
Opt to use `native-tls` instead of `rustls` on cpu architectures not supported by `rustls` (architectures other than ARM and x86) and on MacOS/iOS and Windows because it's shrinks binary size (at least on Windows) and doesn't appear to cause any trouble there.

#### Motivation and Context

Context #844 #895

#### Types of changes

#### Checklist:


hmble:
reqwest is already an optional dependency. See [PR 979](https://github.com/starship/starship/pull/979)

davidkna:
> 
> 
> reqwest is already an optional dependency. See [PR 979](https://github.com/starship/starship/pull/979)

True and I made that pull request myself but lightening the load when https is enabled in is still beneficial.

Cargo.toml
```diff
@@ -51,13 +51,19 @@ regex = "1.3.1"
...
 urlencoding = "1.0.0"
 open = "1.4.0"
-# OpenSSL causes problems when building a MUSL release. Opting to use native SSL implementation
-# see: https://github.com/richfelker/musl-cross-make/issues/65#issuecomment-509790889
-reqwest = { version = "0.10.4", default-features = false, optional = true, features = ["blocking", "rustls-tls"] }
 unicode-width = "0.1.7"
 textwrap = "0.11.0"
 term_size = "0.3.1"
 
+# OpenSSL causes problems when building a MUSL release.
```
matchai: Oof that's hairy. I wish there was a more concise way to address this. 😰

hmble:
Oopsie my bad. Sorrry I did not saw your name in that PR @davidkna

Cargo.toml
davidkna: In #933 @q66 mentions everything builds fine for them with MUSL. Is the trouble related with MUSL and openssl related to the CI setup or is it something else?

Should I start a test run to see if MUSL builds fine with openssl on CI now? Beyond that do you know how well the rust openssl crate handles different openssl library versions and  locations across different systems these days? I remember having trouble with that back in the day.

q66: yes, no issues building on musl (x86_64 or any other architecture) on my system when using `openssl-sys`, it just dynamically links against that and works fine

i think the issue for you might be that upstream rust musl builds prefer static linkage by default, while we override this in Void to use dynamic linkage by default; that may result in the openssl crate building its own openssl rather than linking system's...

so, maybe make the condition something like `cfg(target_feature = "crt-static")`? if using `crt-static`, use `rustls`, otherwise use `native-tls`... then see if it builds on CI, and I can test if it builds here.

davidkna: > i think the issue for you might be that upstream rust musl builds prefer static linkage by default, while we override this in Void to use dynamic linkage by default; that may result in the openssl crate building its own openssl rather than linking system's...

Well I just tried cross-compiling to FreeBSD and I just the same error as when I tried to compile to MUSL (`Could not find directory of OpenSSL installation` which apparently means it needs a cross-compiled openssl). This means that the error is likely not related to `crt-static` (and isn't that anly about the c runtime and not other dependencies?). I suppose vendoring openssl could be a fix as opposed it to being the issue.

> Ideally, it'd be good to make this overridable too, to force either of these options even if the default is different.

Not sure how this could be solved. Is it possible to do feature-aware dependency specification?

q66: Yes, if you cross compile you obviously need OpenSSL/libressl for target in the cross root, that works fine though (we build all ARM packages as crosscompiled in Void and there are no issues)

davidkna: I think the MUSL builds here are mainly intended for maximum platform compablity so I guess vendoring works fine here.

q66: It makes sense as an option, i don't think musl is otherwise special though

davidkna: Yes. I added a disabled-by-default `tls-vendored` feature in the recent commits that gets enabled only in the CI MUSL release builds.

matchai:
Thank you both, @davidkna and @q66! 😄 
LGTM! 👍

PR #1014, opened 2020-03-17
docs: Add disabled = true to [Battery] section in docs

### LINKED ISSUE #1013: Add `disable = false` to [battery] section in config documentation
#### Current Behavior

[battery] in config docs is `disabled = true` by default, but in the sample example, since` disabled = false` is not described, there is a problem that copying and pasting does not work.

#### Expected Behavior

In the sample example, set it to `disabled = true` so that it can be used as it is.

minakawa-daiki:
#### Description

Added a fix to add `disabled = false` to [Battery] in the config documentation

#### Motivation and Context

Closes #1013

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/de-DE/config/README.md
+++ b/docs/de-DE/config/README.md
@@ -185,6 +185,7 @@ Note: Battery indicator will be hidden if the status is `unknown` or `empty` unl
 # ~/.config/starship.toml
 
 [battery]
+disabled = false
 full_symbol = "🔋"
 charging_symbol = "⚡️"
 discharging_symbol = "💀"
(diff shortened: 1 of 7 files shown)


andytom:
Maybe I am missing something but the battery module is enabled by default, see https://starship.rs/config/#battery. Is this not the case?

matchai:
The battery is in-fact enabled by default.
The default functionality is to only show the battery value when it is below 10%, as detailed in the "battery display" configuration: https://starship.rs/config/#battery-display

PR #1019, opened 2020-03-20
feat: Add Undistract Me Feature

### LINKED ISSUE #1397: feat: optionally use the bell to notify the user after a long-running command
When I run long-running commands, I change windows and forget about them.

`echo -e "\07"` if the command's duration is higher that some value maybe

Maybe there's shell (zsh, fish) plugins for that with desktop notifications and things like that.

schrieveslaach:
#### Description
Often it is handy to get notified when the execution of a command finished. This feature is inspired by https://github.com/jml/undistract-me/ and includes `rust-notify` to generate a desktop notification. 

This PR is probably not yet done. For example, it should be turned off by default (I think). I want to elaborate if this is an extension worthy to be included into starship.

#### Motivation and Context
When long commands are done, it is often handy to get a desktop notification in order to return to the shell and continue with your work.

Fixes #1397

#### Types of changes

grafik

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -27,6 +27,9 @@ jobs:
           profile: minimal
           components: rustfmt
 
+      - name: Setup | libdbus (ubuntu)
+        run: sudo apt-get install libdbus-1-dev
+
       - name: Build | Format
         run: cargo fmt --all -- --check
 
@@ -50,6 +53,10 @@ jobs:
             ~/.cargo/git
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Setup | libdbus (ubuntu)
+        if: matrix.os == 'ubuntu-latest'
+        run: sudo apt-get install libdbus-1-dev
 
       - name: Setup | Rust
         uses: actions-rs/toolchain@v1
@@ -108,6 +115,11 @@ jobs:
       - name: Setup | Checkout
         uses: actions/checkout@v2
 
+      - name: Setup | libdbus (ubuntu)
+        if: matrix.os == 'ubuntu-latest'
+        run: sudo apt-get install libdbus-1-dev
+
+      # Cache files between builds
       - name: Setup | Cache Cargo
         uses: actions/cache@v2
         with:
@@ -132,4 +144,4 @@ jobs:
 
       # Run the ignored tests that expect the above setup
       - name: Build | Test
-        run: cargo test -- -Z unstable-options --include-ignored
+        run: cargo test --all-features -- -Z unstable-options --include-ignored
--- a/src/context.rs
+++ b/src/context.rs
@@ -198,6 +198,10 @@ impl<'a> Context<'a> {
             _ => Shell::Unknown,
         }
     }
+
+    pub fn get_cmd_duration(&self) -> Option<u128> {
+        self.properties.get("cmd_duration")?.parse::<u128>().ok()
+    }
 }
 
 #[derive(Debug)]
--- a/src/modules/cmd_duration.rs
+++ b/src/modules/cmd_duration.rs
@@ -11,13 +11,6 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("cmd_duration");
     let config: CmdDurationConfig = CmdDurationConfig::try_load(module.config);
 
-    let props = &context.properties;
-    let elapsed = props
-        .get("cmd_duration")
-        .unwrap_or(&"invalid_time".into())
-        .parse::<u128>()
-        .ok()?;
-
     /* TODO: Once error handling is implemented, warn the user if their config
     min time is nonsensical */
     if config.min_time < 0 {
@@ -28,7 +21,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    if elapsed < config.min_time as u128 {
+    let elapsed = context.get_cmd_duration()?;
+    let config_min = config.min_time as u128;
+
+    if elapsed < config_min {
         return None;
     }
 
@@ -53,7 +49,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         }
     });
 
-    Some(module)
+    Some(undistract_me(module, &config, elapsed))
 }
 
 // Render the time into a nice human-readable string
@@ -84,6 +80,49 @@ fn render_time_component((component, suffix): (&u128, &&str)) -> String {
         0 => String::new(),
         n => format!("{}{}", n, suffix),
     }
+}
+
+#[cfg(not(feature = "notify-rust"))]
+fn undistract_me<'a, 'b>(
+    module: Module<'a>,
+    config: &'b CmdDurationConfig,
+    _elapsed: u128,
+) -> Module<'a> {
+    if config.show_notifications {
+        log::debug!("This version of starship was built without notification support.");
+    }
+
+    module
+}
+
+#[cfg(feature = "notify-rust")]
+fn undistract_me<'a, 'b>(
+    module: Module<'a>,
+    config: &'b CmdDurationConfig,
+    elapsed: u128,
+) -> Module<'a> {
+    use ansi_term::{unstyle, ANSIStrings};
+    use notify_rust::{Notification, Timeout};
+
+    if config.show_notifications && config.min_time_to_notify as u128 <= elapsed {
+        let body = format!(
+            "Command execution {}",
+            unstyle(&ANSIStrings(&module.ansi_strings()))
+        );
+
+        let mut notification = Notification::new();
+        notification
+            .summary("Command finished")
+            .body(&body)
+            .icon("utilities-terminal")
+            .timeout(Timeout::Milliseconds(750));
+
+        if let Err(err) = notification.show() {
+            log::trace!("Cannot show notification: {}", err);
+        }
+    }
+
+    module
 }
 
 #[cfg(test)]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -96,6 +96,12 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4efd02e230a02e18f92fc2735f44597385ed02ad8f831e7c1c1156ee5e1ab3a5"
+
+[[package]]
+name = "bitflags"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
@@ -110,6 +116,12 @@ dependencies = [
  "arrayvec",
  "constant_time_eq",
 ]
+
+[[package]]
+name = "block"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
 
 [[package]]
 name = "block-buffer"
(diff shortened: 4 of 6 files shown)


chipbuster:
Thanks for the suggestion!

Personally, I would definitely use this feature sometimes, especially since I've **never** gotten the notification feature working in fedora-patched gnome-terminal on Arch.

In the larger scheme of things, I wonder if this is an appropriate thing for starship. This is a shell prompt, not a full configuration framework. The expectation is that the user will configure other things (like terminal-title changing) instead of having us do it.

Do we get any advantages (in speed, configuration complexity, etc.) by putting this in the prompt? If so, it might be a candidate to keep.

schrieveslaach:
@chipbuster, thanks for your feedback. I'm also unsure if this features belongs to starship. However, this would lower the bar to use such notifications. As you mentioned, you couldn't get undistract-me running and I had trouble too; and undistract-me works only for bash. 

Including this feature into starship would make the usage of such notification very easy and works for more shells out of the box.

schrieveslaach:
Thanks for your feedback. 

I'm wondering too if this feature is appropriate for starship.

schrieveslaach:
I tested it on MacOs as well and it works as well.

matchai:
I'd be open to adding this feature to the prompt, but I think it may come as a surprise to folks that a sudden notification is coming from their prompt, of all things.

I think it would be sensible to have it be disabled by default and configurable within the `cmd_duration` module.

schrieveslaach:
@matchai, I've created a version that adds this configuration. What do you think?

chipbuster:
After thinking about this more, I agree with matchai: I'd be happy to merge this as a modification to the existing cmd_duration module. We could make arguments for having the user set up their own hooks, using terminal-based notifiers (libvte in particular), or use a custom module, but each of those sets up some complexity having to deal with ordering or environment.

That being said, given that this essentially reads a cmd_duration and conditionally calls a crate (and doesn't render anything to the prompt!), I don't think it needs its own standalone module. We could probably simplify a lot of the changes by merging it as a change to cmd_duration.

schrieveslaach:
That's good news.

I can adjust the code based on your suggestions so this PR can be merged anytime soon.

andytom:
I don't know if it is worth waiting but notify-rust just merged support for Windows (https://github.com/hoodie/notify-rust/pull/69) which would allow us to enable this feature on all platforms which would be nice.

schrieveslaach:
@andytom, I've upgaded `notify-rust` to `4.0.0-beta.4` and it compiles on Windows as well (verified by the GitHub action for Windows).

schrieveslaach:
@andytom, `notify-rust` v4 has been released and I updated the PR.

src/modules/mod.rs
```diff
@@ -150,3 +151,31 @@ pub fn description(module: &str) -> &'static str {
...
+            let body = format!(
+                "Command execution {}",
+                unstyle(&ANSIStrings(&module.ansi_strings()))
+            );
+
+            let mut notification = Notification::new();
+            notification
+                .summary("Command finished")
+                .body(&body)
+                .icon("bash")
```
chipbuster, line 171: According to the notify_rust docs `.icon()` needs to be freedesktop.org compliant. Would ["utilities-terminal"](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#names) be a more standard choice here?

More importantly, does "utilities-terminal" even work, or does it throw up a generic "no icon" icon?

src/modules/mod.rs
```diff
@@ -150,3 +151,31 @@ pub fn description(module: &str) -> &'static str {
         _ => "<no description>",
     }
 }
+
+fn undistract_me<'a>(context: &'a Context, module: Option<Module<'a>>) -> Option<Module<'a>> {
```
chipbuster, line 155: It's not clear to me why this has to be a separate module-creating function instead of just a modification to `cmd_duration`, but if we're going this route, this function should be placed in cmd_duration, not in `mod.rs`.

If at all possible, I would prefer that this PR just modify the behaviour of the cmd_duration module so that we don't have to deal with the cognitive overhead of this wrapping when reading code in the future.

src/context.rs
```diff
@@ -171,6 +171,10 @@ impl<'a> Context<'a> {
             _ => Shell::Unknown,
         }
     }
+
+    pub fn elapsed(&self) -> Option<u128> {
```
chipbuster, line 175: Can we call this something like `get_cmd_duration`?

As a local variable, it was fine for it to have a more ambiguous name since that would only be used within the function. As a method on a struct that's touched by almost every single other module, I'd like something a little more specific.

src/modules/mod.rs
schrieveslaach: Yes, utilities-terminal is a better choice.

src/modules/mod.rs
schrieveslaach: At the beginning I wasn't sure where to put it. I moved it to `cmd_duration`.

src/context.rs
schrieveslaach: done.

schrieveslaach:
> Pretty happy with the logic in here. Have a few questions about the organization/layout of the code.

I've updated the code. Please, have a look again.

chipbuster:
Hit the wrong button :(

The code here is good enough that I'd be happy to add docs and hit "merge."

Unfortunately, something popped up in testing which I should have noticed earlier, namely the requirement on libdbus, as it's currently set up, will completely prevent the prompt from running on any system that doesn't have dbus installed. It's not obvious to me how big of a deal this is, or whether this will work on Termux or other systems.

@andytom @matchai Do either of you have a feeling for how often we'd encounter a platform where DBus isn't installed? Will this interfere with our ability to make our MUSL builds?

andytom:
> @andytom @matchai Do either of you have a feeling for how often we'd encounter a platform where DBus isn't installed? Will this interfere with our ability to make our MUSL builds?

I'm have no idea, I assumed DBus would be installed by default on most Linux distros but this PR had to update the CI config to install it on Ubuntu so that might not be the best assumption. There is an open issue to remove the libdbus requirement and move to a pure rust version (see https://github.com/diwic/dbus-rs/issues/100) but it looks like that work is still in it's infancy.

chipbuster:
@schrieveslaach I think I have an idea for how we can get this into the codebase without potentially breaking binary distribution: can use [Cargo features](https://doc.rust-lang.org/cargo/reference/features.html) to make this a compile-time option? That way, we have an easy way to disable the linking to libraries when building for platforms that don't have the available libraries.

If we do that, we'd probably want to add a version of this code for when building without it enabled that logs a message for DEBUG mode, something like "This version of starship was built without notification support."

schrieveslaach:
@chipbuster, `notify-rust` is now an optional dependency.

chipbuster:
@matchai We've got an interesting problem here.

We have a feature that should generally be disabled at compile-time (because if the recipient doesn't have the right libraries installed on Linux, a binary compiled with the feature will fail to run).

At the same time, we want to run CI tests with this feature enabled (otherwise we can't know if it's broken until someone files a bug report).

Is there any easy way to resolve this?

bbigras:
Can't you add the deps and use `cargo build --features` while building with the github actions?

.github/workflows/workflow.yml
```diff
@@ -79,6 +82,9 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
+      - name: Install libdbus (ubuntu)
+        run: sudo apt-get install dbus libdbus-1-3 libdbus-glib-1-dev
+
```
ATiltedTree, line 87: `cargo check` only compiles and does not link to the needed native libraries.

```suggestion
```

.github/workflows/workflow.yml
```diff
@@ -47,6 +47,9 @@ jobs:
...
```
ATiltedTree, line 52: `cargo clippy` only compiles and does not link to the needed native libraries.

```suggestion
```

.github/workflows/workflow.yml
```diff
@@ -104,8 +110,13 @@ jobs:
...
 
-      - name: Setup | Cache Cargo
-        uses: actions/cache@v2
+      - name: Install libdbus (ubuntu)
+        if: matrix.os == 'ubuntu-latest'
+        run: sudo apt-get install dbus libdbus-1-3 libdbus-glib-1-dev
+
+      # Cache files between builds
+      - name: Cache cargo registry
+        uses: actions/cache@v1
```
ATiltedTree, line 119: We only need `libdbus-1-dev` because `dbus` and `libdbus-1-3` is preinstalled and `libdbus-glib-1-dev` is not needed.

```suggestion
      - name: Setup | libdbus (ubuntu)
        if: matrix.os == 'ubuntu-latest'
        run: sudo apt-get install libdbus-1-dev

      # Cache files between builds
      - name: Setup | Cache Cargo
        uses: actions/cache@v2
```

src/modules/cmd_duration.rs
```diff
@@ -86,6 +82,43 @@ fn render_time_component((component, suffix): (&u128, &&str)) -> String {
...
+
+    module
+}
+
+#[cfg(feature = "notify-rust")]
+fn undistract_me(module: Module, elapsed: u128) -> Module {
+    use ansi_term::{unstyle, ANSIStrings};
+    use notify_rust::{Notification, Timeout};
+
+    let config: CmdDurationConfig = CmdDurationConfig::try_load(module.config);
```
ATiltedTree, line 100: I think passing the config into this function would be better since we wouldn't have to load it twice.

ATiltedTree:
My bad.
We still have to install `libdbus-1-dev` for `check` and `clippy` because `pkg-config` searches for them.
If you could add them back that would be wonderful.

schrieveslaach:
@ATiltedTree, all checks are green now. :green_heart:

schrieveslaach:
@ATiltedTree, done. This PR is now up-to-date with master.

DMRobertson:
> > @andytom @matchai Do either of you have a feeling for how often we'd encounter a platform where DBus isn't installed? Will this interfere with our ability to make our MUSL builds?
> 
> I'm have no idea, I assumed DBus would be installed by default on most Linux distros but this PR had to update the CI config to install it on Ubuntu so that might not be the best assumption. There is an open issue to remove the libdbus requirement and move to a pure rust version (see [diwic/dbus-rs#100](https://github.com/diwic/dbus-rs/issues/100)) but it looks like that work is still in it's infancy.

Ubuntu user here. I think I might have hit this. Details for completeness:

I found that notifications weren't working on Ubuntu 20.04 after a plain `cargo install spaceship`. I added `show_notifications = true` to the config but still wasn't seeing notifications. After doing some digging I tried `cargo install starship --features notify-rust` which was unhappy:

```
--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Failure { command: "\"pkg-config\" \"--libs\" \"--cflags\" \"dbus-1\" \"dbus-1 >= 1.6\"", output: Output { status: ExitStatus(ExitStatus(256)), stdout: "", stderr: "Package dbus-1 was not found in the pkg-config search path.\nPerhaps you should add the directory containing `dbus-1.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'dbus-1\' found\nPackage dbus-1 was not found in the pkg-config search path.\nPerhaps you should add the directory containing `dbus-1.pc\'\nto the PKG_CONFIG_PATH environment variable\nNo package \'dbus-1\' found\n" } }', /home/dmr/.cargo/registry/src/github.com-1ecc6299db9ec823/libdbus-sys-0.2.1/build.rs:6:5
```
After installing `apt install libdbus-glib-1-dev` I was able to build starship as above. Notifications work now :)

chipbuster:
@DMRobertson Thanks for the report! Just to be clear, was this an Ubuntu desktop (like one where you sit down and work on the machine with a monitor) or was it a remote machine you access over SSH?

DMRobertson:
@chipbuster Just a regular desktop, installed to disk from a DVD a few months back. Happy to provide any other information if it's helpful!

PR #1021, opened 2020-03-21
feat: Add parser for format strings

### LINKED ISSUE #624: Module configuration with format strings

heyrict:
#### Description
This PR implements the parser of format strings described in https://github.com/starship/starship/issues/624#issuecomment-564831405

Usage can be found [here](https://github.com/heyrict/starship/compare/98a854d...716e5d3).

#### Motivation and Context

Works toward closing #624

#### Design
```
text1 [text2 $var](style)
──┬── ──────────────────┐
Text   ──┬──  ─┬─  ──┬──│
       Text    │   Text TextGroup
            variable
       ────┬────   ──┬──
     FormatElement StyleElement
───────────┴─────────────
```

This format string will be parsed into the following vector, `Vec<FormatElement>`:

```rust
[
    FormatElement::Text("text1 "),
    FormatElement::TextGroup(
        TextGroup {
            format: [
                FormatElement::Text("text2 "),
                FormatElement::Variable("var"),
            ],
            style: [
                StyleElement::Text("style"),
            ],
        },
    ),
]
```

##### Variable
Variable is a string of `[a-zA-Z0-9_]+` which is preceded by a `$`.

When `StringFormatter.new(format)` is called, the format string `format` will be converted to a `Vec<FormatElement>` stated above, and then it will extract variables into `self.variables`, which is a BTreeMap. What each variable represents will be filled when `self.map*` is called.

##### TextGroup
TextGroup is a pair of `format` and `style`: `[format](style)`.

Lacking either of `[`, `]`, `(`, `)` will make `StringFormatter::new()` returns an `Error`.

Each `format` is handled as a sub-format-string, and thus nesting TextGroups is allowed.

##### Style
Style is a style string that will be fed into `crate::config::parse_style_string`.

There is a `Variable` enum in `StyleElement`, but variables in `()` are not handled in this PR.

##### Escape
One of the following chars can be escaped in the parser: `[]()$\`, followed by a literal `\`, which means, you need to write `format = \\[` or `format = '\['` inside toml to correctly escape a `[`, and write `format = \\\\` or `format = '\\'` in order to make `format` a literal `\`.

#### Types of changes

#### Checklist:


heyrict:
Here is the comparison of the performance between v0.37.0 and this PR, no significant performance downgrade monitored. Tested under Ubuntu 18.04 w/ Intel i5-5200U @ 2.2GHz

Screenshot from 2020-03-21 16-40-43

heyrict:
These following works are not done in this PR, either having break changes, or better be split out from this PR.

- Refactoring `prompt_order` in `StarshipRootConfig` to use format string.

  As long as the parser converts a format string (`&str`) and variables (`Vec<Segment>`) into a sequence of `Segment` (`Vec<Segment>`), it would break the convention of using `Module` for each module when filling a variable with a module. I'm afraid it will break things such as `starship explain`, logging, and other functions that requires a module to be rendered as `Module`.

- Using variables in style.

  When I mean style, I mean text `[](here)`. In this PR, text in `()` are directly parsed as a style string and do not allow variables. This is easy to implement but won't be implemented in this PR as further discussions are required since it is not mentioned in @matchai 's comment.

- Style combining

  ```rust
  "[text1 [text2](green)](red bold)
  ```

  In this example, `text2` will be rendered as normal `green` rather than `green bold` in this PR. Styles are overwritten rather than combined for the limitation of `parse_style_string`.

- Refactoring modules to use format string

  This is left for further work to avoid conflicts with other PRs. To make sure that modules can work properly with the new formatting system, I've made [a branch](https://github.com/heyrict/starship/tree/feat/format-string-simplified-rust) that has module `Rust` refactored.

- Depracate `SegmentConfig`, `Affix`, `Suffix`

  ,once all modules use the new format system.

PR #1030, opened 2020-03-25
feat: add a julia module

cappyzawa:
#### Description

* I added module for [The Julia Language](https://julialang.org/)

#### Motivation and Context

Closes #

#### Types of changes

Screen Shot 2020-03-25 at 11 32 57

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -107,6 +107,7 @@ prompt_order = [
     "golang",
     "haskell",
     "java",
+    "julia",
     "nodejs",
     "php",
     "python",
@@ -794,6 +795,31 @@ symbol = "+ "
 threshold = 4
 ```
 
+## Julia
+
+The `julia` module shows the currently installed version of Julia.
+The module will be shown if any of the following conditions are met:
+
+- The current directory contains a `Project.toml` file
+- The current directory contains a `Manifest.toml` file
+- The current directory contains a file with the `.jl` extension
+
+### Options
+
+| Variable   | Default       | Description                                              |
+| ---------- | ------------- | -------------------------------------------------------- |
+| `symbol`   | `"👸 "`       | The symbol used before displaying the version of Julia. |
+| `style`    | `"bold purple"` | The style for the module.                                |
+| `disabled` | `false`       | Disables the `julia` module.                            |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[golang]
+symbol = "🤴 "
+```
 ## Kubernetes
 
 Displays the current Kubernetes context name and, if set, the namespace from
--- a/src/configs/julia.rs
+++ b/src/configs/julia.rs
@@ -1,0 +1,23 @@
+use crate::config::{ModuleConfig, RootModuleConfig, SegmentConfig};
+
+use ansi_term::{Color, Style};
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig)]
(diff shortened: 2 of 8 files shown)


docs/config/README.md
```diff
@@ -794,6 +795,31 @@ symbol = "+ "
...
+| `symbol`   | `"👸 "`       | The symbol used before displaying the version of Julia. |
+| `style`    | `"bold purple"` | The style for the module.                                |
+| `disabled` | `false`       | Disables the `julia` module.                            |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[golang]
```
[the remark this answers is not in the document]
cappyzawa: Oh..
Thanks!!

cappyzawa:
@chipbuster 
> Holy cow the number of things we need to change to add a new module is getting high. We should probably make a list somewhere...

I think so.

> As an aside, do you think the "therefore" emoji (∴) might be an appropriate default symbol?

That's nice!
I was having trouble finding a good symbol.

heyrict:
It seems that julia uses `Project.toml` to store version info. What do you think about adding some code to `package` module to display the version?

cappyzawa:
> It seems that julia uses Project.toml to store version info. What do you think about adding some code to package module to display the version?

It's sound good.
However, I want to implement it in another PR because it is not included in the julia module.

docs/config/README.md
```diff
@@ -794,6 +795,31 @@ symbol = "+ "
...
+| `style`    | `"bold purple"` | The style for the module.                                |
+| `disabled` | `false`       | Disables the `julia` module.                            |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[julia]
+symbol = "👸 "
```
waldyrious, line 821: I believe this example may not be compliant with [Julia's community standards](https://julialang.org/community/standards/):

> do not sexualize the term "Julia" or any other aspects of the project. While "Julia" is a female name in many parts of the world, the programming language is not a person and does not have a gender.

I'd suggest using something different, e.g. "Mathematical Bold Capital J" (U+1D409):

    symbol = "𝐉 "

or "Mathematical Sans-serif Bold Small J" (U+1D5F7), which looks similar to the j in [Julia's logo](https://julialang.org):

    symbol = "𝗷 "

WDYT?

matchai: Agreed, it would probably be best if we changed it.
I like both your suggestions, @waldyrious. Another option if the "Therefore" symbol, present in the Julia logotype and favicon: `∴` or `ஃ`

waldyrious: The therefore symbol is actually already the default :) if I understood the code correctly, this was an example of customizing the symbol. However, the second symbol you listed is an even better match to the Julia logo! I'd suggest replacing the current default with it and using one of the J symbols I proposed as an example of customization.

matchai: > The therefore symbol is actually already the default :)

Ah right. My bad. 🤦‍♂
This is what happens when I swoop into issues without reading the context. 😅 

> However, the second symbol you listed is an even better match to the Julia logo! I'd suggest replacing the current default with it and using one of the J symbols I proposed as an example of customization.

Sounds good to me. 👍 
Would you be open to creating a PR for the change, Waldir?

cappyzawa: Oh.. I didn't know https://julialang.org/community/standards/ :bow:

I will fix this as soon as possible.

PR #1033, opened 2020-03-27
docs: Fix broken installation link anchor on README

jmfederico:
README does not include the "🚀" anchor for the installation sub-section.
This adds it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@
 <p align="center">
   <a href="https://starship.rs">Website</a>
   ·
-  <a href="#-installation">Installation</a>
+  <a href="#🚀-installation">Installation</a>
   ·
   <a href="https://starship.rs/config/">Configuration</a>
 </p>
--- a/README.md
+++ b/README.md
@@ -130,6 +130,8 @@
 </p>
 
 
+<a name="🚀-installation"></a>
+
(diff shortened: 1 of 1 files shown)


matchai:
It seems to me like the link works correctly as-is, and doesn't work when updated to include the emoji. I'm running Brave browser, so I imagine other Chromium browsers behave the same.

What browser are you using, which ran into this issue?

jmfederico:
I am using Chrome 80.0.3987.149 and Firefox 74.0 on macOS.

The sidebar link with the rocket works, the body link fails on both browsers.

## Firefox
20200328101337-f675b4c18c gif-2-mp4 com

## Chrome
20200328101409-f675b4c18c gif-2-mp4 com

andytom:
I think the problem is that when Github renders the `README.md` it doesn't include the 🚀 in the the anchors (`https://github.com/starship/starship/#-installation`) but when VuePress renders does (`https://starship.rs/guide/#%F0%9F%9A%80-installation`) unfortunately I'm not sure what the fix would be.

jmfederico:
I hackish solution is to include a additional/fake anchor.

In this case I created a fake anchor for GitHub, but can be done the other way around. This should work in both cases.

PR #1034, opened 2020-03-28
perf(dotnet): fail faster in the dotnet module

davidkna:
#### Description

For finding the necessary files to determine the version and if the module should be enabled the `dotnet` module is currently iterating through all files in the current directory (`O(n)`) to find relevant files.

This PR makes the module exit early in case the relevant files aren't present and only does the work of finding these files if they are present.

Possible breakage:
`estimate_dotnet_version` is case insentive but `context.is_match()` is not.

#### Motivation and Context

Tested with hyperfine in `starship/src/modules`:

before:
`starship module dotnet 63.9 ms ±   4.6 ms`

after:
`starship module dotnet 23.9 ms ±   2.5 ms`

#### Types of changes

#### Checklist:


davidkna:
It looks like #886 wasn't fixed after all.

matchai:
Thank you for the performance improvement, @davidkna! 😄

PR #1056, opened 2020-04-07
refactor: Replace prompt_order with format string

heyrict:
#### Description
This PR adds `format` key and deprecates `add_newline` and `prompt_order` keys in the config

Prefix and suffix will be ignored since this PR gets merged.

**Merging This PR will cause all modules not using format string printing no `prefix` and `suffix`**

#### Types of changes

#### Checklist:


andytom:
> I think we can now remove `[pre|suf]fix` and all calls to `get_[pre|suf]fix().set_value("")` because their the same across all modules.

In #1057 that is listed as a separate step, that I think is part of https://github.com/starship/starship/milestone/4

PR #1061, opened 2020-04-07
feat: Prefix & Suffix support for hostname & username

### LINKED ISSUE #705: Provide the option of formatting user, host, and path as `user@host:path`
While the current `<user> on <host> in <directory>` formatting works perfectly well and is easy to read, it would be nice to have the option of displaying those values as `<user>@<host>:<directory>`.

In particular, this would facilitate easy (read: lazy) formulation of `scp`, `rsync`, and `git` commands. If I have two terminal sessions open, one local to my machine and the other an SSH session to remote host, and I want to `rsync` something from my local machine to my current directory on the remote host, all I have to type is the beginning of the command: `rsync <local_file> `.

Then I copy the `<user>@<host>:<directory>` string from the session on the remote host, 
paste it into the local session, hit return, and I'm off to the races.

This would of course require the display of full host name and path to be useful, but I already have those enabled.

(This request might, of course, be rendered moot by the ongoing effort to generalize the formatting of module output.)

diminishedprime:
#### Description
Instead of having `via` and `on` hardcoded, use the prefix & suffix values similar to other modules. 

#### Motivation and Context
I'd like to be able to format my prompt as `username@hostname:path`. Since `on` and `via` were hard-coded in the source, I couldn't configure them without a fork. 

Fixes #705 

#### Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

This is technically a breaking change because if someone had configured the hostname prefix, they will no longer also get `on`. 

`cargo test`

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -755,8 +755,8 @@ The `hostname` module shows the system hostname.
 | Variable   | Default               | Description                                                                                                                          |
 | ---------- | --------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
 | `ssh_only` | `true`                | Only show hostname when connected to an SSH session.                                                                                 |
-| `prefix`   | `""`                  | Prefix to display immediately before the hostname.                                                                                   |
-| `suffix`   | `""`                  | Suffix to display immediately after the hostname.                                                                                    |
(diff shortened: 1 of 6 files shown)


matchai:
We're currently working toward a more robust solution for all modules. Please keep an eye on #1021 for updates. 😊

PR #1064, opened 2020-04-08
Added module support for z.lua.

jimleroyer:
#### Description

Added module support for [z.lua](https://github.com/skywind3000/z.lua). This also covers configuration, tests and mock command. The way the module is detected is either via:

1- The presence of the environment variables `ZLUA_LUAEXE` and `ZLUA_SCRIPT` which are usually configured by the *z.lua* installation script.
2- The explicit configuration of the Lua executable location and z.lua script location.

Number 2 is necessary for customization as the *z.lua* installation script might not render the variables in the environment scope but in some others unfortunately, such as  is the case for PowerShell where it resides in `$Script` scope rather than `$env`.

Regarding the icon for z.lua, it would have been nice to use the project's lightning emoji but I had a hard time to get some of these to correctly display on the prompt. I left it out to the user to customize this to his liking. 

#### Motivation and Context

The [z.lua](https://github.com/skywind3000/z.lua) project offers a way to quickly change directory based on keywords and frecency algorithms without having to type the full paths. It's a convenient alternative for the classic shell [z command](https://github.com/rupa/z). It supports various platforms (including Windows) and shells. It's generally more performant than its counterparts. 

This module display the rank of the current directory. It's useful to me so I can track which directories are better ranked and manually intervene on a folder that should not be. 

#### Warning

I'm a rookie when it comes to Rust. Feel free to direct me on better coding practices. 

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -95,6 +95,7 @@ prompt_order = [
     "hostname",
     "kubernetes",
     "directory",
+    "z.lua",
     "git_branch",
     "git_commit",
     "git_state",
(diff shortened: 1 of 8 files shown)


matchai:
Hey Jimmy! Thank you for your PR. 😊 

As a user of one of the many z/jump tools, I'm afraid this seems like it may be a little too specific to be added as a core module in Starship. I don't think that most users using such tools would consider the rank to be actionable data to be added to the prompt.

I'd suggest keeping an eye on #916, which would allow for such a module to be used without being a core module.

I'm open to being convinced otherwise, but I'm afraid this may be a little too niche a use-case for us to have added to the codebase for us to continually maintain.

jimleroyer:
Hey Matan. Yeah that totally makes sense. Admittedly, I wanted to try out my hands on Rust again and that was a pet project for me. Good job BTW on being able to say no, not all features should get in a project.

Regarding the custom commands, that seems like a nice feature. I'll probably re-implement this functionality using it once it's out and micro-benchmark it to see the impact, that would be an interesting data point to have a feature natively implemented and the other with the extension mechanism. 

I'd like to contribute otherwise to the project. If you see something for me to contribute with, feel free to ping me on Discord. I saw there is some upcoming grunt work on migrating modules using the new format mechanism, it's something I could help with.

matchai:
Thank you for your understanding, Jimmy. 😊 

> I'll probably re-implement this functionality using it once it's out and micro-benchmark it to see the impact.

That would be interesting, for sure.
In the long run, I'd love to eventually work out a convenient way for users to share their custom modules, inspired by how [Espanso] managed to make their package system.

> I saw there is some upcoming grunt work on migrating modules using the new format mechanism, it's something I could help with.

There's certainly a whole lot to do there! We have an umbrella issue outlining the work needed on that side of things if you want to give it a shot: https://github.com/starship/starship/issues/1057 👍

PR #1063, opened 2020-04-08
refactor(rust): Use format strings

heyrict:
#### Description
This PR deprecates `symbol` `version` `style` keys in `RustConfig`, and adds `format` key instead.

Also can be used as an example of refactoring existing modules.

#### Motivation and Context

Related #1057

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #1066, opened 2020-04-08
refactor(golang): Use format strings

jletey:
#### Description

This PR deprecates `symbol` `version` `style` keys in `GoConfig`, and adds `format` key instead.

Based off of #1063.

#### Motivation and Context

Related #1057

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Screenshots (if appropriate):

Screenshot 2020-04-08 at 17 28 45

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #1071, opened 2020-04-08
refactor(hostname): Use format strings

### LINKED ISSUE #1102: Hostname module doesn't respect prefix/suffix
#### Current Behavior
The `hostname` module doesn't treat the values of `hostname.prefix` and `hostname.suffix` in `starship.toml` as an actual prefix and suffix. Instead, it wraps them into the main hostname value and then forces a prefix of `"on "` before that value, as below:

```rust
let hostname_stacked = format!("{}{}{}", config.prefix, host, config.suffix);
module.create_segment("hostname", &SegmentConfig::new(&hostname_stacked));
module.get_prefix().set_value("on ");
```

#### Expected Behavior
The values of `hostname.prefix` and `hostname.suffix` as set by the user replace the defaults, and appear with the same formatting (i.e. unformatted) as the default `"on "` prefix. In other words, the following `starship.toml`...

```toml
[hostname]
prefix = "@ "
```

...and the hostname `Foo` would result in the `hostname` module printing...

```
@ Foo
```

...rather than the default:

```
on Foo
```

> Environment and shell configuration trimmed, as the cause seems to be a directly apparent design choice in the code - I can add them again if needed, though

jletey:
#### Description

This PR deprecates `prefix` `suffix` `style` keys in `HostnameConfig`, and adds `format` key instead.

Based off of #1063.

#### Motivation and Context

Related #1057
Fixes #1102

#### Types of changes

Screenshot 2020-04-08 at 17 23 02

#### Checklist:


src/configs/hostname.rs
```diff
@@ -1,26 +1,21 @@
...
 use starship_module_config_derive::ModuleConfig;
 
 #[derive(Clone, ModuleConfig)]
 pub struct HostnameConfig<'a> {
     pub ssh_only: bool,
-    pub prefix: &'a str,
-    pub suffix: &'a str,
     pub trim_at: &'a str,
-    pub style: Style,
+    pub format: &'a str,
```
heyrict, line 9: Same as #1066

src/modules/hostname.rs
```diff
@@ -42,10 +43,20 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-    module.get_prefix().set_value("on ");
+    let formatter = if let Ok(formatter) = StringFormatter::new(config.format) {
+        formatter.map(|variable| match variable {
+            "hostname" => Some(host.to_string()),
+            _ => None,
+        })
+    } else {
+        log::warn!("Error parsing format string in `hostname.format`");
+        return None;
+    };
```
heyrict, line 54: Ditto

heyrict:
Thanks for your initial work on this! Going to merge it now.

PR #1078, opened 2020-04-10
feat: Add migration script for format strings

heyrict:
#### Description
This adds a binary `starship_migrate_0_40` that updates the config file.

Changes proposed in #1056 are applied.

#### Motivation and Context

See https://github.com/starship/starship/issues/1057#issuecomment-611582686

#### Types of changes

#### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,3 +78,7 @@ lto = true
 [[bin]]
 name = "starship"
 path = "src/main.rs"
+
+[[bin]]
+name = "starship_migrate_0_40"
+path = "scripts/starship_migrate_0_40.rs"
--- a/scripts/starship_migrate_0_40.rs
+++ b/scripts/starship_migrate_0_40.rs
@@ -1,0 +1,224 @@
+/*
+ * starship_migrate_0_40
+ *
+ * A script for migrating starship configuration file from before v0.40 to v0.40.
+ *
+ */
+use dirs::home_dir;
+use std::env;
+use std::fs::{copy, File};
+use std::io::Write;
+use std::path::PathBuf;
+use toml::Value;
+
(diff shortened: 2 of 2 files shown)


heyrict:
I'm wondering if this PR should be included in the repo `starship/starship` or in separate repo. Actually it is a onetime-off script and won't need to be compiled every time with `starship` after v0.40. It seems cleaner if we put it in a separate repo, but it will force developers to open two PRs to refactor a module, one to `starship/starship`, and another to the migration script.

I'm thinking about adding the migration binary here and move the code to another repo when all modules get refactored. Any thoughts @matchai ?

nickwb:
Are we going to force users to run this script manually? I think that would be a unfortunate outcome..

An alternative solution might look like this:

1. `config_schema = n` becomes a part of the root config. (where `n` is a number). If the property is missing, then we assume `n = 1` (i.e., all current config files are schema version 1).
2. When loading the config during a normal `starship prompt`, "hot migrate" any config from the old version to the latest version, using as many migration steps as is necessary. 
3. Write a warning out to stderr whenever we needed to migrate prior to drawing the prompt: `"Your starship configuration is using an outdated format. Run 'starship config migrate' to update it."`

heyrict:
@nickwb Thanks for your comments. I agree with you in point 1 and 3. I don't think 2 works good as it will change the format (probably deleting all comments) of the config file without users' confirmation.

This PR is not included in the umbrella issue. I think it is safe to be closed as per discussion in https://github.com/starship/starship/discussions/1116#discussioncomment-6579-permalink . Most modules will work the same as before with the config examples in `docs/config/README.md` except a few (`character`, `memory` and `time` afaik).

andytom:
I have created two issues for the points that have been raised here.

PR #1082, opened 2020-04-10
fix(ruby): Do not assume length when retrieving version

jonknapp:
#### Description

This changes the version parsing logic to not assume a max length of the Ruby version number.

#### Motivation and Context

 It is currently incorrect for a version like `2.1.10`.

#### Types of changes

I've added a test example which is not dependent on the OS. It does assume that `p` is part of the `ruby --version` output.

#### Checklist:


src/modules/ruby.rs
```diff
@@ -38,7 +38,10 @@ fn format_ruby_version(ruby_version: &str) -> Option<String> {
         .split_whitespace()
         // return "2.6.0p0"
         .nth(1)?
-        .get(0..5)?;
+        // split into ["2.6.0", "0"]
+        .split('p')
+        // return "2.6.0"
+        .nth(0)?;
```
andytom, line 44: It's not obvious from the CI output but I think this line should be
```suggestion
        .next()?;
```

See https://rust-lang.github.io/rust-clippy/master/index.html#iter_nth_zero

jonknapp: Sounds good; force pushed an update

PR #1084, opened 2020-04-11
refactor(java): Added formatter support for Java module.

jimleroyer:
#### Description

This PR deprecates `symbol` and `style` keys in `JavaConfig` to replace it with the `format` key instead.

#### Motivation and Context

Closes #

Related to issue #1057

#### Types of changes

image

- [x] I have tested using **Windows Linux Subsystem**

#### Checklist:


docs/config/README.md
```diff
@@ -877,11 +877,20 @@ The module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Variable   | Default        | Description                                            |
-| ---------- | -------------- | ------------------------------------------------------ |
-| `symbol`   | `"☕ "`         | The symbol used before displaying the version of Java. |
-| `style`    | `"dimmed red"` | The style for the module.                              |
-| `disabled` | `false`        | Disables the `java` module.                            |
+| Variable   | Default                             | Description                                     |
```
heyrict, line 880:
```suggestion
| Option     | Default                             | Description                                     |
```
To avoid naming conflicts with variable in format string

src/modules/java.rs
```diff
@@ -25,12 +26,35 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                        "symbol" => Some(config.symbol),
+                        _ => None,
+                    })
+                    .map_style(|variable| match variable {
+                        "style" => Some(Ok(config.style)),
+                        _ => None,
+                    })
+                    .map(|variable| match variable {
+                        // This may result in multiple calls to `get_module_version` when a user have
+                        // multiple `$version` variables defined in `format`.
```
heyrict, line 43:
```suggestion
```

docs/config/README.md
```diff
@@ -877,11 +877,20 @@ The module will be shown if any of the following conditions are met:
...
+| ---------- | ----------------------------------- | ----------------------------------------------- |
+| `format`   | `"via [$symbol $version]($style) "` | The format for the module.                      |
+| `symbol`   | `"☕"`                               | A format string representing the symbol of Java |
```
heyrict, line 883:
```suggestion
| `format`   | `"via [$symbol$version]($style) "` | The format for the module.                      |
| `symbol`   | `"☕ "`                             | A format string representing the symbol of Java |
```
Well, this change looks reasonable to me, but it will break the config before the format string release. I think it should use exactly the same config as before.

src/configs/java.rs
```diff
@@ -1,21 +1,22 @@
...
 
 impl<'a> RootModuleConfig<'a> for JavaConfig<'a> {
     fn new() -> Self {
         JavaConfig {
-            symbol: SegmentConfig::new("☕ "),
-            style: Color::Red.dimmed(),
+            format: "via [$symbol $version]($style) ",
             disabled: false,
+            style: "red dimmed",
+            symbol: "☕",
```
heyrict, line 19:
```suggestion
            format: "via [$symbol$version]($style) ",
            disabled: false,
            style: "red dimmed",
            symbol: "☕ ",
```

src/modules/java.rs
```diff
@@ -25,12 +26,35 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                        "version" => Some(Ok(formatted_version.clone())),
```
heyrict, line 44: Cloning isn't required here

docs/config/README.md
jimleroyer: Arf... not ideal but I get it for backward compatibility. Let me change that.

src/modules/java.rs
jimleroyer: Without it, I get this error:
```
cannot move out of `formatted_version`, a captured variable in an `Fn` closure
  --> src\modules\java.rs:42:46
42 |                         "version" => Some(Ok(formatted_version)),
   |                                              ^^^^^^^^^^^^^^^^^ move occurs because `formatted_version` has type `std::string::String`, which does not implement the `Copy` trait
```

jimleroyer: Okay I changed it to a reference using `&` and it works. 👍

PR #1085, opened 2020-04-11
refactor(git_status): Use format strings with lazy initialization

heyrict:
#### Description
This PR rewrites  `git_status` to use format strings instead of a combination of `SegmentConfig` and `CountConfig`.

The following variables are available in `format`:

- stashed, ahead, behind, conflicted, deleted, renamed, modified, staged, untracked:
  
  Content of these variables are defined under key `*_format`, where `*` is the name of the variable.

  `*_format`: A format string with a variable `count` that prints the count of corresponding commits.

- diverged:

  Content of this variable in defined in `diverged_format`, which is a format string with the following two variables available:

  - ahead_count: The count of commits ahead of the tracking branch
  - behind_count: The count of commits behind of the tracking branch

- ahead_behind:

  Will display `$ahead` `$behind` or `$diverged` based on the current status.

- all_status:

  Will display all the variables except $ahead, $behind and $diverged.

All these variables will be computed in parallel on demand.

#### Motivation and Context

Related #1057

#### Types of changes

Screenshot from 2020-04-11 20-42-22

#### Checklist:


heyrict:
This PR uses `Mutex` to avoid data races, where `OnceCell` may suit more to the use case.

heyrict:
I'm going to merge this for now as it is totally functional and should not block the process of format string release. You can use the following techniques for further improvements in another issue.

1. Use `OnceCell` instead of `Mutex` locks
2. Parse the output of `git status` instead of using `libgit2`.

PR #1086, opened 2020-04-11
feat(Azure): Azure module

andymac4182:
#### Description
This adds support for showing the Azure subscription you are currently executing as for commands with the Azure CLI.

#### Motivation and Context
Myself and a few people I know are managing multiple Azure Subscriptions and need to ensure we are currently selected into the correct one.

#### Types of changes

I have tested against the current version of the Azure CLI with and without the environment variable specifying where to find the configuration. I am not set up currently to test MacOS or Linux sorry.

#### Checklist:

I am have updated the English documentation but I don't know other languages to do those.

I just noticed the big move to format strings. Is there an easy bit of doco on their format? I will move this module over to using that.

### DIFF
--- a/src/modules/azure.rs
+++ b/src/modules/azure.rs
@@ -1,0 +1,89 @@
+use dirs::home_dir;
+use encoding_rs::*;
+use std::env;
+use std::fs::File;
+use std::io::{BufRead, BufReader, Read};
+use std::path::PathBuf;
+use std::str::FromStr;
+
+use super::{Context, Module, RootModuleConfig};
+
+type JValue = serde_json::Value;
+
+use crate::configs::azure::AzureConfig;
+use crate::formatter::StringFormatter;
+
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("azure");
+    let config = AzureConfig::try_load(module.config);
+
+    let subscription_name = get_azure_subscription_name()?;
+
+    let formatter = if let Ok(formatter) = StringFormatter::new(config.format) {
+        formatter.map(|variable| match variable {
+            "subscription" => Some(subscription_name.to_string()),
+            _ => None,
+        })
+    } else {
+        log::warn!("Error parsing format string in `azure.format`");
+        return None;
+    };
+
+    module.set_segments(formatter.parse(None));
+
+    module.get_prefix().set_value("");
+    module.get_suffix().set_value("");
+
+    Some(module)
+}
+
+fn get_azure_config_file_location() -> Option<PathBuf> {
+    env::var("AZURE_CONFIG_DIR")
+        .ok()
+        .and_then(|path| PathBuf::from_str(&path).ok())
+        .or_else(|| {
+            let mut home = home_dir()?;
+            home.push(".azure");
+            Some(home)
+        })
+}
+
+fn get_azure_subscription_name() -> Option<String> {
+    let mut azure_profile_json = get_azure_config_file_location()?;
+    azure_profile_json.push("azureProfile.json");
+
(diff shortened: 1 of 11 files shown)


andymac4182:
Updated to support format strings

src/modules/azure.rs
```diff
@@ -0,0 +1,89 @@
...
+    module.get_prefix().set_value("");
+    module.get_suffix().set_value("");
+
+    Some(module)
+}
+
+fn get_azure_config_file_location() -> Option<PathBuf> {
+    env::var("AZURE_CONFIG_DIR")
+        .ok()
+        .and_then(|path| PathBuf::from_str(&path).ok())
```
nickwb, line 43: I think there is an implicit allocate + copy here which you an avoid by using `PathBuf::from(path)` instead.

src/modules/azure.rs
```diff
@@ -0,0 +1,89 @@
...
+    let bytes = buffer.as_mut_slice();
+    let decodedbuffer = UTF_8.decode_with_bom_removal(bytes).0;
+
+    let parsed_json: JValue = serde_json::from_str(&decodedbuffer).ok()?;
+
+    let subscriptions = parsed_json["subscriptions"].as_array()?;
+
+    for sub in subscriptions {
+        let subscription_id = sub["id"].as_str()?;
+        if subscription_id == region {
```
nickwb, line 83: `std::iter::find` is probably more idiomatic here, rather than the for loop. Or even: `subscriptions.iter().filter_map(...).next()`, so you can keep the `?` syntax.

src/modules/azure.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    let region = region_line.split('=').nth(1)?;
+    let region = region.trim();
+
+    let file = File::open(&azure_profile_json).ok()?;
+    let mut buffer: Vec<u8> = Vec::new();
+    let mut reader = BufReader::new(file);
+    reader.read_to_end(&mut buffer).ok()?;
+    let bytes = buffer.as_mut_slice();
+    let decodedbuffer = UTF_8.decode_with_bom_removal(bytes).0;
```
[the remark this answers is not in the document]
andymac4182: I was originally using this util and had an issue with UTF-8 BOM. I moved to using encoding since that appeared to be the best way to handle BOM.

src/modules/azure.rs
andymac4182: Thanks for this. Got some great reading about iterators in Rust. Ended up going with find_map 👍

andymac4182:
Updated based on feedback and is good to go I believe.

nickwb:
Built and tested on MacOS, looking good :)

image

gertjvr:
I just build and run tests on ubuntu 
image

andytom:
Thanks for the contribution @andymac4182, since this contains this is a new module with format strings, I'm going to put this PR on hold until that work has been completed. Sorry for the delay.

andymac4182:
Curious why can't it be included in the format strings Milestone?

andytom:
> Curious why can't it be included in the format strings Milestone?

For now we are trying to keep the `feat/format-string` just for the adding format strings and the refactor of the existing modules, to simplify debugging issues, tracking work, etc.

andymac4182:
Any update on when this can be taken off `on hold`?

matchai:
We're putting `on hold` on PRs ready to merge once the format strings is released.

andymac4182:
@matchai Sorry why was this closed?

matchai:
Whoops! I must have accidentally hit the "Close and comment" button 😅

matchai:
It looks like I can't reopen it since the upstream branch no longer exists. 😓 
Would you mind recreating the PR?

PR #1089, opened 2020-04-11
chore: use an Action for updating Homebrew formula

dawidd6:
This PR switches the Homebrew formula bumping logic from custom script to an Action.

Example of a project already using it: https://github.com/jesseduffield/lazygit

Link to Action repo: https://github.com/dawidd6/action-homebrew-bump-formula

Disclaimer: I'm the author.

#### Description

Use an Action instead of custom script.

#### Motivation and Context

Less maintenance burden, I guess.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -164,24 +164,15 @@ jobs:
 
   update_brew_formula:
     if: startsWith(github.ref, 'refs/tags/v')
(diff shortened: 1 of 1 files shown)


andytom:
Sorry @dawidd6, it looks like this PR got lost and we switched to use an action in #1391. I'm going to close PR, sorry that your changes didn't make it in.

PR #1093, opened 2020-04-11
fix: branch_name in fresh repo (unborn master)

### LINKED ISSUE #686: Untracked files and staged files are not shown in empty git repository
#### Current Behavior

In an empty git repository (`git init` in an empty dir), when files are created (e.g. `touch file`) then Starship doesn't show there are untracked files. Similarly, when something is staged (e.g. `git add file`), Starship doesn't show these untracked or staged files.

Note that Starship does detect the user is in a Git repository: note the (default config) prompt changes after `git init`.

#### Expected Behavior

As a user I expect Starship to show me my Git status in a just initialised Git repo, i.e. untracked files (and count) and staged files (and count). Maybe there are more possibilities in an empty repo concerning status, but these two are common.

N/A

#### Environment
- Starship version: `starship 0.27.0`
- Shell type: bash
- Shell version: `GNU bash, version 5.0.11(1)-release (x86_64-apple-darwin18.6.0)`
- Shell plugin manager: N/A
- Terminal emulator: Terminal
- Operating system: macOS 10.14.6 (18G1012)

N/A

#### Starship Configuration

default

#### Possible Solution

It appears that Starship detects that the user is in a Git repo: the repo's name appears instead of the working directory. Somehow Starship should/could detect that there are no Git commits or branches, or maybe some Git log tells the latest action was `git init`? Somehow Starship should/could keep track of the current `git status`, which is a valid command at this stage.

hrombach:
#### Description

Since #1035 didn't quite do what it was supposed to, I had another look and figured out the problem is actually that libgit2 doesn't return a `HEAD` if it's pointing to an unborn branch (like in the case of a freshly initialised repo). 

AFAIK (please correct me if I'm wrong) the only valid case where this error would occur is a fresh repo, so I think it's fine if the module assumes the branch name to be "master" in that case, since the git cli acts the same.

#### Motivation and Context

Closes #686 

#### Types of changes

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -3,7 +3,7 @@ use crate::module::Module;
 
 use crate::modules;
 use clap::ArgMatches;
-use git2::{Repository, RepositoryState};
+use git2::{ErrorCode::UnbornBranch, Repository, RepositoryState};
 use once_cell::sync::OnceCell;
 use std::collections::{HashMap, HashSet};
 use std::env;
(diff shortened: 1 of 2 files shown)


hrombach:
@andytom ofc, I'll get on that as soon as I'm done with work for today

hrombach:
@andytom sorry it took me a while to get around to it, but I added a test now, I hope it's adequate.

andytom:
> @andytom sorry it took me a while to get around to it, but I added a test now, I hope it's adequate.

Not a problem thanks for adding the test.

PR #1095, opened 2020-04-12
feat: Modify config keys from shell

### LINKED ISSUE #1080: Make it possible to modify the config from shell

Currently the configuration is static and cannot change unless you edit the file

Commands like `starship configure set character.symbol "$"`, like [bspwm](https://github.com/baskerville/bspwm) is doing with their config

jletey:
#### Description

- Alias `starship configure` to `starship config`
- You can now pass in a key (e.g. `character.symbol`) and value (e.g. `"$"`) to `starship config` and it will update Starship's configuration file for you.

#### Motivation and Context

Closes #1080

#### Types of changes

#### Checklist:

### DIFF
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -18,7 +18,7 @@ pub fn update_configuration(name: &str, value: &str) {
     let keys: Vec<&str> = name.split('.').collect();
     if keys.len() != 2 {
         log::error!("Please pass in a config key with a '.'");
-        process::abort();
+        process::exit(1);
     }
 
     let starship_config = StarshipConfig::initialize();
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -1,6 +1,7 @@
 use std::env;
 use std::ffi::OsString;
 use std::io::ErrorKind;
+use std::process;
 use std::process::Command;
 
 use dirs::home_dir;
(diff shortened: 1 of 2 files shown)


sandorex:
You should probably keep `starship configure` as an alias to `starship config`

jletey:
@sandorex Good idea :+1:

src/configure.rs
```diff
@@ -3,8 +3,55 @@ use std::ffi::OsString;
...
+
+fn get_file_path() -> Option<PathBuf> {
+    if let Ok(path) = env::var("STARSHIP_CONFIG") {
+        // Use $STARSHIP_CONFIG as the config path if available
+        Some(PathBuf::from(path))
+    } else {
+        // Default to using ~/.config/starship.toml
+        home_dir().map(|path| path.join(".config/starship.toml"))
+    }
+}
```
andytom, line 53: Would it be better to use the `get_config_path` function rather than duplicate the code here?

jletey:
@andytom Thanks for taking a look at the PR 😊 You can now pass in integers and booleans and they will be parsed correctly. Also when you pass in an invalid configuration key, the program halts and "soft" errors.

jletey:
@andytom I have made your suggested change. I think using `process::exit` is better than `process::abort`. It seems like the dotnet module tests are failing, but I guess this has nothing to do with my changes.

PR #1098, opened 2020-04-13
fix(conda): Do not depend on user's config in conda tests

sbarzowski:
#### Description
The user's config is automatically loaded in context.new_module. This change explicitly sets the empty config to avoid depending on the environment.

#### Motivation and Context
Without this fix, the tests do not pass for users who have a custom symbol defined for conda. This prevents installing/upgrading from Arch Linux AUR.

#### Types of changes

N/A

Tests (`cargo test`) no longer fail on my machine.

#### Checklist:

### DIFF
--- a/tests/testsuite/conda.rs
+++ b/tests/testsuite/conda.rs
@@ -2,12 +2,14 @@ use ansi_term::Color;
 use std::io;
 
 use crate::common;
+use crate::common::TestCommand;
 
 #[test]
 fn not_in_env() -> io::Result<()> {
     let output = common::render_module("conda")
         .env_clear()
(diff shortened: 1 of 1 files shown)


tests/testsuite/conda.rs
```diff
@@ -2,12 +2,15 @@ use ansi_term::Color;
 use std::io;
 
 use crate::common;
+use crate::common::TestCommand;
+use toml::Value;
```
andytom, line 6: This isn't used so can be removed to make Clippy happy.

```suggestion
```

sbarzowski: Thanks for catching this, done.

PR #1109, opened 2020-04-15
feat(package): configurable displaying package version for packages marked as private

### LINKED ISSUE #1053: Revert hiding npm version for private packages

After the change in #975, the version of private npm packages is no longer displayed. I disagree with this change. If I have a package.json with a non-zero version, then I'm likely using that version for something and am interested in seeing it displayed. Why should starship care about the publish status of of the package?

In my use case, I have Node microservices that use npm for dependency management. The services are published as Docker images, using the version from package.json as an image tag and also as a runtime service version. So even though I have `"private": "true"` in my package.json to prevent publishing an npm module, the package version is still very important to my workflow.

I would like the change in #975 reverted. I'm fine with replacing it with a check for `"version": "0.0.0"` as suggested in the conversation linked from the PR; however, other members of the community might feel differently.

None

kudlatyamroth:
#### Description
This PR adds config `display_private` in package scope.
If true, it will display version for `package.json` marked as private

#### Motivation and Context
In my job, we mark as private company packages that are not published to npm registry,
without this i can't see current package version

Closes #1053

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -13,11 +13,11 @@ use crate::configs::package::PackageConfig;
 ///
 /// Will display if a version is defined for your Node.js or Rust project (if one exists)
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
-    match get_package_version(&context.current_dir) {
-        Some(package_version) => {
-            let mut module = context.new_module("package");
-            let config: PackageConfig = PackageConfig::try_load(module.config);
+    let mut module = context.new_module("package");
+    let config: PackageConfig = PackageConfig::try_load(module.config);
 
+    match get_package_version(&context.current_dir, &config) {
+        Some(package_version) => {
             module.set_style(config.style);
             module.get_prefix().set_value("is ");
 
@@ -38,10 +38,11 @@ fn extract_cargo_version(file_contents: &str) -> Option<String> {
     Some(formatted_version)
 }
 
-fn extract_package_version(file_contents: &str) -> Option<String> {
+fn extract_package_version(file_contents: &str, display_private: bool) -> Option<String> {
     let package_json: json::Value = json::from_str(file_contents).ok()?;
 
-    if package_json.get("private").and_then(json::Value::as_bool) == Some(true) {
+    if !display_private && package_json.get("private").and_then(json::Value::as_bool) == Some(true)
+    {
         return None;
     }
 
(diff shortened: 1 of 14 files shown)


kudlatyamroth:
@jletey tests refactored :)
if something is wrong let me know :) thats my first contact with Rust :P

jletey:
> @jletey tests refactored :)
> if something is wrong let me know :) thats my first contact with Rust :P

@kudlatyamroth Looks amazing! Thanks so much for your contribution and quickly adding my suggestion :)

kudlatyamroth:
anyone can merge this? its annoying to not see version of package

kudlatyamroth:
@andytom ok, PR updated to upstream and tests moved to module
i also added third parameter to render_module, with config for starship as i need it in one test

andytom:
> edit: is there a way to restart tests?

You should be able to restart the tests from the Checks tab at the top of the PR, I've tried rerunning the jobs a couple of time but still failing with a network issue 😞.

kudlatyamroth:
i can't see anything that i can click to rerun jobs...
but it seems that it finished successfully :)

src/modules/package.rs
```diff
@@ -367,91 +401,149 @@ java {
...
+        let text = String::from(contains.unwrap_or(""));
+        let expected = Some(format!(
+            "is {} ",
+            Color::Fixed(208).bold().paint(format!("📦 {}", text))
+        ));
+
+        match contains {
+            Some(_contains) => assert_eq!(actual, expected),
+            None => assert_eq!(actual, None),
+        }
```
heyrict, line 545: I think it would be better written as the following as you don't need the value `contains`.
```suggestion
        match contains.is_some() {
            true => assert_eq!(actual, expected),
            false => assert_eq!(actual, None),
        }
```

kudlatyamroth: changed and updated with upstream

kudlatyamroth: after update with upstream, i need to update erlang tests with third parameter in render_module

kudlatyamroth:
anyone can rerun test checks?

src/modules/package.rs
```diff
@@ -366,91 +400,149 @@ java {
...
+        let text = String::from(contains.unwrap_or(""));
+        let expected = Some(format!(
+            "is {} ",
+            Color::Fixed(208).bold().paint(format!("📦 {}", text))
+        ));
+
+        match contains.is_some() {
+            true => assert_eq!(actual, expected),
+            false => assert_eq!(actual, None),
+        }
```
andytom, line 544: Clippy is suggesting this should be an if/else statement (see https://github.com/starship/starship/pull/1109/checks?check_run_id=625039316#step:4:400) rather than a match so it is failing.

kudlatyamroth: ah, ok, i saw that it cannot download some lib and i think that it was same error that last time

PR #1114, opened 2020-04-16
fix(directory): avoid confusing modules with PowerShell paths

jeanga:
#### Description

Powershell supports PSDrives (https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-psdrive?view=powershell-7) that allows to create "logical" drives mapped to actual Windows drives.

#### Motivation and Context

When using PSDrives, you can use paths that are related to a logical drive.
For example, test: can be "mapped" to c:\test folder.
You can then use "cd test:" to change the current location.

Many other drives exist be default (env, function, variable etc...). 

These "Powershell" paths confuse starship modules that fail to return the correct result.

Changing the --path argument to a Windows path (instead of the Powershell one) fixes the issue.

In our example, with the fix, when "cd test:\" is used, starship is called with --path C:\test (instead of test:\).

Closes:

#### Types of changes
This change introduces a call to Convert-Path in startship.ps1 to convert the variable $PWD (a "PowerShell" path to a Windows path. This change avoids modules (like Rust, Git, ...) being confused by a PowerShell path.

Created a PSDrive:
     New-PSDrive -Scope 1 -Name test -PSProvider FileSystem -root c:\test -Description "PSDrive test with starship"
     Checked basic modules (Rust&Git) are now working properly.

- [X] I have tested using **Windows Subsystem for Linux**

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -7,12 +7,14 @@ function global:prompt {
     # @ makes sure the result is an array even if single or no values are returned
     $jobs = @(Get-Job | Where-Object { $_.State -eq 'Running' }).Count
 
+    $current_directory = (Convert-Path $PWD)
+
     if ($lastCmd = Get-History -Count 1) {
         $duration = [math]::Round(($lastCmd.EndExecutionTime - $lastCmd.StartExecutionTime).TotalMilliseconds)
         # & ensures the path is interpreted as something to execute
-        $out = @(&::STARSHIP:: prompt "--path=$PWD" --status=$lastexitcode --jobs=$jobs --cmd-duration=$duration)
+        $out = @(&::STARSHIP:: prompt "--path=$current_directory" --status=$lastexitcode --jobs=$jobs --cmd-duration=$duration)
     } else {
-        $out = @(&::STARSHIP:: prompt "--path=$PWD" --status=$lastexitcode --jobs=$jobs)
+        $out = @(&::STARSHIP:: prompt "--path=$current_directory" --status=$lastexitcode --jobs=$jobs)
     }
 
     # Convert stdout (array of lines) to expected return type string
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -7,6 +7,7 @@ function global:prompt {
     # @ makes sure the result is an array even if single or no values are returned
     $jobs = @(Get-Job | Where-Object { $_.State -eq 'Running' }).Count
 
+    $env:PWD = $PWD
     $current_directory = (Convert-Path $PWD)
 
     if ($lastCmd = Get-History -Count 1) {
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -1,5 +1,5 @@
 use path_slash::PathExt;
-use std::path::Path;
+use std::path::{Path,PathBuf};
 use unicode_segmentation::UnicodeSegmentation;
 
 use super::{Context, Module};
@@ -29,7 +29,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     // Using environment PWD is the standard approach for determining logical path
     // If this is None for any reason, we fall back to reading the os-provided path
     let physical_current_dir = if config.use_logical_path {
-        None
+        match std::env::var("PWD") {
+            Ok(x) => {
+                Some(PathBuf::from(x))
+            },
+            Err(e) => {
+                log::debug!("Error getting PWD environment variable: {}", e);
+                None
+            }
+        }
     } else {
         match std::env::current_dir() {
             Ok(x) => Some(x),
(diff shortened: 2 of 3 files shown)


jimleroyer:
Being a PowerShell user myself, I gave this a spin locally. It does improve the behavior with PSDrive. Some examples (would have been nice to have these SS submitted with the PR -- before / after images of behavior):

The root is properly displayed within a PSDrive:
jeanga-proper-root

This is what's the best about this fix to me though. We can see that the master build breaks on the information display when we are in git repository with modules that should be detected (Rust+git in this case):
jeanga-full-info

Looks good to me! 👏

andytom:
Also I'm not sure but it looks like this will also fix #1106, is this correct?

jeanga:
Sorry for the formatting errors & test failures ("cargo test" did not fail on my machine for a reason I still ignore). 
My only excuse is to be new to Rust&Starship. 
Thank you for your understanding!

andytom:
> Sorry for the formatting errors & test failures ("cargo test" did not fail on my machine for a reason I still ignore).
> My only excuse is to be new to Rust&Starship.
> Thank you for your understanding!

Not a problem, we have a few tests that are ignored locally because they depend on the machine running the tests being setup in a particular way (e.g specific versions of an application being installed or some particular files existing).

@jimleroyer and @ylor can you have a quick look at this to confirm it works as expected and I'll get this merged in for the next release.

jimleroyer:
I don't think it works as expected yet. The root folder works fine by displaying `C:/` but getting into a folder shows double forward slashes, which some of the old code prevented to do (and now does again).

image

I would suggest a) to remove the double forward slashes as it seems like a regression and b) to add a test that triggers this behavior by getting into a folder after a root, so we are sure it does not repeat again.

davidkna:
> 
> 
> I don't think it works as expected yet. The root folder works fine by displaying `C:/` but getting into a folder shows double forward slashes, which some of the old code prevented to do (and now does again).
> 
> image
> 
> I would suggest a) to remove the double forward slashes as it seems like a regression and b) to add a test that triggers this behavior by getting into a folder after a root, so we are sure it does not repeat again.
> 
> I can help with that if @jeanga wants me to submit a PR fixing that on his repo.

The `C:\` to `/c` is broken anyway because it only works with `C:\`. e.g. `D://foo` is broken. This brokenness is at least consistent.

Getting rid of the `//` should be possible by replacing `path_slash` 
e.g. with

```rust
use std::path::Path;

fn to_slash(path: &Path) -> Option<String> {
    assert!(path.is_absolute());

    let size = path.as_os_str().len();
    let mut out = String::with_capacity(size);

    let mut iterator = path.iter();

    // Don't append / before root directory
    out.push_str(&iterator.next()?.to_string_lossy());
    
    // skip prevents double //
    for component in iterator.skip(1) {
        out.push('/');
        out.push_str(&component.to_string_lossy());
    }

    Some(out)
}
```
Obviously windows-only.

jeanga:
Hi davidkna,

I noticed that behavior but I do not think it is a regression of this PR.
It was already behaving this way before.

Are you suggesting I add you to the members of my forked repo?
If so, I can certainly do that :-)

Jean

davidkna:
> 
> 
> Hi davidkna,
> 
> I noticed that behavior but I do not think it is a regression of this PR.
> It was already behaving this way before.
> 
> Are you suggesting I add you to the members of my forked repo?
> If so, I can certainly do that :-)
> 
> Jean

The removal of `replace_c_dir()` makes this something of a regression, but I'm not sure that if that change would better of in a seperate PR.

jeanga:
Yes, but the behavior showed with anything _not_ c:\ like d:\ ... z:\ :-)

Jean

ylor:
> 
> 
> I don't think it works as expected yet. The root folder works fine by displaying `C:/` but getting into a folder shows double forward slashes, which some of the old code prevented to do (and now does again).
> 
> image
> 
> I would suggest a) to remove the double forward slashes as it seems like a regression and b) to add a test that triggers this behavior by getting into a folder after a root, so we are sure it does not repeat again.
> 
> I can help with that if @jeanga wants me to submit a PR fixing that on his repo.

And the slashes are still going the wrong way. Not sure if this PR is for solving that though. Please see the behavior I expect when working in Powershell (that is to say not WSL).

Annotation 2020-04-27 204940

jeanga:
This PR tries to be as close as the previous behavior as possible while fixing annoying bugs for Windows Powershell users.
As previously stated, the double slash (D://Windows) and replacing the backslash with a slash was the previous way it behaved. 
If there is a consensus on preserving \ and removing these double //, I'll be happy to submit a fix in another PR.

ylor:
@jeanga Thanks for your work.

andytom:
@jeanga thank you for your contribution and thank you to everyone else for your help testing this. I've left #1106 open and maybe that would be a good place to start discussing how to display paths in Powershell on Windows.

PR #1126, opened 2020-04-20
refactor(python): Added formatter support for the Python module.

jimleroyer:
#### Description

This PR deprecates `symbol` and `style` keys in the Python module to replace it with the `format` key instead.

Because the module got some complexity by supporting version provided by either pyenv or python itself, also supporting virtualenv, that would make the format string configuration more complicated than necessary. We'd need to support conditional display of text based on the presence of certain variable. This could be feasible but not in short immediate plan as this require proper design planning. 

After some discussion with @matchai , we determined we could split the python module in two for virtualenv to live in its own module. This should for the time being reduce complexity and ease up configuration.

#### Motivation and Context

Related to issue #1057

#### Types of changes

image

#### Checklist:


jimleroyer:
> Code generally lgtm, while I think there could be unexpected behavior when `is_py_project == false` and `is_venv == true`.
> 
> Previously this will print `:python: v3.7.6 (venv) ` for example, but with a separate `virtualenv` module, this will print `(venv) ` only.

Yes, that will happen with a separate module. Not only that, but it simply requires to get out of the Python project to keep the virtualenv environment displayed and have the Python version being removed.

That said, I think it's okay, as it behaves the same as the virtualenv activation script. The latter will continue to display the virtualenv name on the prompt even when getting out of the project, until the activation is disabled. Because the virtualenv overrides binaries in the standard PATH, even when getting out of the Python directory: it indicates to the user that normal $PATH behavior is not to be expected. 

Hmm honestly, to my personal preference, it's not a module that I would enable as this seems not so useful, as the activation script already displays that information. Here in that screenshot, you can see two `venv` text fragment in the prompt: one displayed by the activation script and the other by starship. Both displays that information even when outside of a Python project's folder.

image

heyrict:
> as the activation script already displays that information

Maybe this is true for windows, but in Linux with bash or zsh, the `(venv)` pretended by activation script will be overwritten when using starship.

Imo, we should pertain max compatibility with previous behavior in a module in the refactoring process. If you think the changes are reasonable, I will leave this PR open and wait for another one that is backward compatible.

jimleroyer:
That's ok, I can add a filter on the `virtualenv` to  only display within Python's folder, just like what the Python module does.

heyrict:
> That's ok, I can add a filter on the `virtualenv` to only display within Python's folder, just like what the Python module does.

@jimleroyer You didn't get my point. I mean, adding a `virtualenv` module is fine, but I don't think it should be part of the format string refactor. One python module instead of two with the same filter is much preferred.

jimleroyer:
@heyrict Hmm, that was not clear. After the [discussion thread about the string formatter issue](https://github.com/starship/starship/discussions/1116), I thought that separating this module was the way to go for simplicity. Raising the explicit decision to not favor configuration simplicity and not split the module would have saved me time here. 

I see the point of trying to align best with existing mainline so we minimize code friction. But given that this separation would actually provide simplicity and fix some other issues ([this was just opened for example](https://github.com/starship/starship/issues/1208)), I feel this could have some consideration.

But it's up to you, it's your project. I will start another branch with just the Python module and you pick up the one you prefer. 

In that new PR, the necessary format string will be (not tested): 
```via [${symbol}${version}( \\($virtualenv\\))]($style) ```

I hope we can move the whole of the string format feature in mainline soon. 🤞

jimleroyer:
https://github.com/starship/starship/pull/1210

heyrict:
Thank you for your understanding. I'll try my best to review the PRs ;)

andytom:
Closing this since #1210 has been merged.

PR #1129, opened 2020-04-21
feat(erlang): Add Erlang module

filmor:
#### Description
Add Erlang module and more precise OTP version reading (could be used in the Elixir module as well).

#### Types of changes

#### Checklist:

### DIFF
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -11,6 +11,7 @@ mod dotnet;
 mod elixir;
 mod elm;
 mod env_var;
+mod erlang;
 mod git_branch;
 mod git_commit;
 mod git_state;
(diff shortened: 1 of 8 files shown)


src/modules/mod.rs
```diff
@@ -60,6 +61,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
         "dotnet" => dotnet::module(context),
         "elixir" => elixir::module(context),
         "elm" => elm::module(context),
+        "erlang" => erlang::module(context),
         "env_var" => env_var::module(context),
         "git_branch" => git_branch::module(context),
         "git_commit" => git_commit::module(context),
```
andytom, line 67: Can you also add erlang to the module description at the bottom of this file? This will allow it to work with `starship explain`.

filmor:
Getting proper tests to run is a bit involved, maybe I'll change the Erlang code being called to also read the file, that way it's a single command like for most other tools.

filmor:
I have no idea why this other test is failing, but since I didn't touch anything related to it, I'm not going to waste time on it :)

andytom:
> I have no idea why this other test is failing, but since I didn't touch anything related to it, I'm not going to waste time on it :)

We have some tests in Github Actions that are a bit flaky (see #886) but will usually pass if we rerun them.

filmor:
Then this is ready. As soon as it's merged, I'll do a second PR, improving the Elixir version evaluation (using the language itself instead of parsing the output with a regex) and converting to the new format-string style.

PR #1143, opened 2020-04-26
feat(format_string):  Add TFM support to dotnet module and update to format string

andymac4182:
#### Description
This detects the Target Framework Moniker being used by the current project. 
https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-framework-versions

This also updates to use the new format string branch. https://github.com/starship/starship/issues/1057

#### Motivation and Context
There is multiple TFMs that are targeted under the same SDK. Currently this module just shows the SDK not the TFM.

#### Types of changes

#### Checklist:


src/configs/dotnet.rs
```diff
@@ -1,25 +1,24 @@
...
+    pub symbol: &'a str,
+    pub style: &'a str,
+    pub heuristic: bool,
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for DotnetConfig<'a> {
+    fn new() -> Self {
+        DotnetConfig {
+            format: "[$symbol$version]($style)( 🎯 $tfm)($style) ",
```
heyrict, line 17: Do you mean this?

```suggestion
            format: "[$symbol$version]($style)( [🎯 $tfm]($style)) ",
```

andymac4182: Thanks. I have updated this. :)

andymac4182:
Test fail seems to be related to another module and something in new ubuntu.

src/modules/dotnet.rs
```diff
@@ -46,13 +49,89 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    module.get_prefix().set_value("");
+    module.get_suffix().set_value("");
 
     Some(module)
 }
 
+fn find_current_tfm<'a>(files: &[DotNetFile<'a>]) -> Option<String> {
+    let get_file_of_type = |t: FileType| files.iter().find(|f| f.file_type == t);
+
+    let relevant_file = get_file_of_type(FileType::ProjectFile).or_else(|| files.iter().next())?;
```
nickwb, line 88: The `or_else` doesn't seem to have a purpose here, given the match below will ignore anything which isn't a `ProjectFile`.

src/modules/dotnet.rs
```diff
@@ -46,13 +49,89 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+
+    let mut in_tfm = false;
+    let mut tfm = Option::None;
+    let mut buf = Vec::new();
+
+    loop {
+        match reader.read_event(&mut buf) {
+            // for triggering namespaced events, use this instead:
+            // match reader.read_namespaced_event(&mut buf) {
+            Ok(Event::Start(ref e)) => {
```
nickwb, line 109: Is there an `Event::End`? We could stop parsing the file early if `in_tfm` is true...?

src/modules/dotnet.rs
```diff
@@ -46,13 +49,89 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+
+    let formatter = if let Ok(formatter) = StringFormatter::new(config.format) {
+        formatter
+            .map_style(|variable| match variable {
+                "style" => Some(config.style.to_string()),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "symbol" => Some(config.symbol.to_string()),
+                "version" => Some(version.0.to_string()),
```
nickwb, line 62: I'm not super familiar with the `StringFormatter`, but is it possible to just move `version` and `current_tfm` rather than copying them?

src/modules/dotnet.rs
```diff
@@ -46,13 +49,89 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                match e.name() {
+                    b"TargetFrameworks" => in_tfm = true,
+                    b"TargetFramework" => in_tfm = true,
+                    _ => in_tfm = false,
+                }
+            }
+            // unescape and decode the text event using the reader encoding
+            Ok(Event::Text(e)) => {
+                if in_tfm {
+                    tfm = Some(e.unescape_and_decode(&reader).unwrap())
```
nickwb, line 121: Another option would be simply to return from here..?

Also, can we drop the `unwrap`. `ok()?` should be equivalent, and won't panic.

src/modules/dotnet.rs
heyrict: Yes, it's currently possible to move them in `feat/format-string`.

```
                "version" => Some(version.0),
```

I'm going to review all these changes once the design of `StringFormatter` is mature.

src/modules/dotnet.rs
andymac4182: Fixed up

src/modules/dotnet.rs
andymac4182: Thanks :) I think I have this updated correctly.

src/modules/dotnet.rs
andymac4182: There is a buffer clear just outside the loop. I have updated that when it finds the next close tag it breaks the loop at that point and clears the buffer and returns. I wasn't sure if I just returned if the buffer would clear properly?

nickwb: The `buf.clear()` is for a `Vec` which will go out of scope as soon as you return. Rust guarantees through its memory safety that it will be dropped. Clearing it, moments before it is deallocated, seems to be of little benefit 😃

andymac4182: Ahh thanks. I have updated to return once we find the text we care about :)

src/modules/dotnet.rs
nickwb: These are still copying, actually. Your closure returns `Option<&str>` and the `Into<String>` on map will happily clone them.

I cloned your branch and tried to make this work, but unfortunately `map` accepts a `Fn` rather than a `FnOnce`, so you can't move values into that closure anyway. Plus, having an `FnOnce` doesn't make too much sense with the multiple invocations required with different variables. Solving this would probably require changing the `StringFormatter` to take a `HashMap`, or similar, rather than using this multiply-invoked function.

Anyway, I don't think this needs to be solved in your branch.

You can't solve the copying, but I think you can still neaten up a bit:

```
    "symbol" => Some(config.symbol),
    "version" => Some(&version.0),
    "tfm" => current_tfm.as_deref(),
    _ => None,
})
```

heyrict: I do have solved the copying with `Into<Cow<'a, str>>` in another PR (#1148). This PR has some API changes and I'm going to leave it a while for review. Even if the design changes are rejected, I'll open a PR that reduces unnecessary copying wherever  possible, so do not worry.

nickwb: @heyrict - Yes, the `Cow` with the lifetime of the `StringFormatter` looks like a good solution 👍

andymac4182: Sorry I am new to Rust. How is the best way to determine if `dotnet_files` is thread safe?

heyrict: If you don't know, just make the changes and have the compiler determine it. A variable(struct, enum, etc.) can be safely shared between threads only when all its members can be safely shared between threads, which is indicated by trait Sync.

nickwb: Yeah, just try it. It's a `Vec` of a simple data type, and you're already consuming it by shared reference, so I'd expect it to be fine.

andymac4182: Sorry I think this one is outside my knowledge. I tried to make the update but I had to flow the change from `String` to `&str` all the way through the various functions.

heyrict: It seems that you have to use two `map` as one cannot return two different types in the same function.

Should we change the API to return a `Cow<str>` or `Box<dyn Into<Cow<str>>>`? Though I'm personally happy with the current implementation.

andymac4182: That would be a question for someone else sorry I am starting to understand what you are talking about but not quite there yet

heyrict: Oh, sorry. I should have given a better explanation on what happened.

Here is the type of `StringFormatter::map`:

```rust
pub fn map<T: Into<String>>(mut self, mapper: impl Fn(&str) -> Option<T> + Sync) -> Self;
```

This method accepts a callback function `Fn(&str) -> Option<T> + Sync` that converts the variable name (of type `&str`) to a type `Option<impl Into<String>> + Sync`.

The type `Option<impl Into<String>> + Sync` is a `Option<impl Into<String>>` that can be shared safely between threads. and `Into<String>` means the type can be converted to string, e.g. `&str`, or `String` itself.

While this method is flexible and can accept any type of `T` that implements `Into<String>`, there is only one type `T` and cannot accept both `&str` and `String` as a return type, though they are both valid `impl Into<String>`.

An alternative in rust is `dyn Into<String>` that is a dynamic type that accepts both `String` and `&str`, but since it is dynamic, rust compiler does not know its size (it does not implements `Sized` trait). So it should be allocated in the heap, as the size of pointer is fixed for them. That's why `Box<dyn Into<String>>` works.

Another alternative is to convert `&str` to `String` manually.

docs/config/README.md
```diff
@@ -449,15 +449,31 @@ as running `dotnet --version`, but it may show an incorrect version if your .NET
...
 
+The module will also show the Target Framework Moniker 
+(https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-framework-versions)
+when there is a csproj file in the current directory. 
+
 ### Options
 
 | Variable    | Default       | Description                                              |
 | ----------- | ------------- | -------------------------------------------------------- |
+| `format`   | `"v[$symbol$version]($style)( 🎯 $tfm)($style) "` | The format for the module.                             |
```
heyrict, line 460:
```suggestion
| `format`   | `"[$symbol$version( 🎯 $tfm)]($style) "` | The format for the module.                             |
```

src/modules/dotnet.rs
```diff
@@ -39,20 +42,89 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+            })
+            .map(|variable| match variable {
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "version" => {
+                    let version = if enable_heuristic {
+                        let repo_root = context.get_repo().ok().and_then(|r| r.root.as_deref());
+                        estimate_dotnet_version(&dotnet_files, &context.current_dir, repo_root)?
```
heyrict, line 60: You may want to return an error (`Some(Err(StringFormatterError::new(..)))`) here.

Originally failure in finding dotnet version will make **the module** empty. Returning `None` will make **the variable** `$version` empty, while throwing error makes the format string fail to render.

andymac4182:
@andytom I imagine this PR will also need to be put on hold?

andytom:
> @andytom I imagine this PR will also need to be put on hold?

I don't think so from #1057 this is the PR to add format strings to the Dotnet module or am I missing something?

heyrict:
@andytom This PR adds format string support for dotnet, but it also adds an extra feature of displaying the TFM variable. The first is required to the format string release, but the second is not. It should not be put on hold imo.

andytom:
@andymac4182 thank you for your contribution.

andymac4182:
Thanks for reviewing and merging in :)

PR #1147, opened 2020-04-27
fix(haskell): avoid creating .lock files or downloading GHC

### LINKED ISSUE #1146: Haskell module creates stack.yaml.lock files that irritate git
#### Current Behavior
I had checked out a [Haskell repo](https://github.com/ucsd-progsys/liquidhaskell) and wanted to update to the latest HEAD with `git pull --rebase`.

However, it fails due to the presence of an untracked file:
```
liquidhaskell on  develop [⇣] via λ v8.6.5                                                                                                                              
✦ ❯ gup                                                                             
Updating 6f29ece8f..e38a5a17e                                                                                                                                            
error: The following untracked working tree files would be overwritten by merge:                                                                                         
        stack.yaml.lock                                                                                                                                                  
Please move or remove them before you merge.                                                                                                                             
Aborting                                                                                                                                                                 
First, rewinding head to replay your work on top of it...                                                                                                                
error: The following untracked working tree files would be overwritten by checkout:                                                                                      
        stack.yaml.lock                                                                                                                                                  
Please move or remove them before you switch branches.                                                                                                                   
Aborting                                                                                                                                                                 
fatal: Could not detach HEAD
```
Removing the file didn't help, since it was recreated every time I hit enter. I opened a bash prompt temporarily to delete the lock file and do a plain `git pull` without Starship running `stack`, then dropped back to `zsh`.

This raised another issue: the latest changes to that repo happened to bump the required Haskell compiler (GHC) version to 8.10, which was not installed on my machine. This caused the Starship prompt to hang completely, so I disabled the Haskell module for now.

A quick perusal of the Haskell module shows this:
```rust
    let haskell_version = utils::exec_cmd(
        "stack",
        &["ghc", "--", "--numeric-version", "--no-install-ghc"],
    )?
    .stdout;
```
I'm guessing this is equivalent to the shell command: `stack ghc -- --numeric-version --no-install-ghc`? If so, it doesn't prevent `stack` from trying to build the latest GHC -- I think the `--no-install-ghc` argument needs to go before `ghc`. I might do a PR to try to fix this.

#### Expected Behavior
1. Git pulls the latest changes from the remote repo.
1. The prompt re-appears really quickly.

#### Possible Solution
There seems to be a `--lock-file` [argument to Stack](https://github.com/commercialhaskell/stack/blob/0a9711c71adff8d6cdcda113f05dbec48876018e/src/Stack/Options/GlobalParser.hs#L64) that might avoid this behaviour if set to `read`.

#### Environment
- Starship version: 0.40.1
- zsh version: zsh 5.7.1 (x86_64-ubuntu-linux-gnu)
- Operating system: Ubuntu 19.10

```bash
# WTF
. $HOME/bin/antigen.zsh
antigen use oh-my-zsh
antigen bundle git
antigen bundle zsh-users/zsh-syntax-highlighting
antigen theme robbyrussell
antigen apply
unsetopt share_history

#THIS MUST BE AT THE END OF THE FILE FOR SDKMAN TO WORK!!!
export SDKMAN_DIR="/home/omf/.sdkman"
[[ -s "/home/omf/.sdkman/bin/sdkman-init.sh" ]] && source "/home/omf/.sdkman/bin/sdkman-init.sh"
export DEFAULT_USER=omf
source $HOME/.asdf/asdf.sh
source $HOME/.asdf/completions/asdf.bash

# OPAM configuration
. /home/omf/.opam/opam-init/init.zsh > /dev/null 2> /dev/null || true

export PATH="$HOME/.yarn/bin:$PATH"
export PATH=/home/omf/.local/bin:$PATH

# completions
if [ -f ~/.config/exercism/exercism_completion.zsh ]; then
  source ~/.config/exercism/exercism_completion.zsh
fi

[[ -s "/home/omf/.gvm/scripts/gvm" ]] && source "/home/omf/.gvm/scripts/gvm"
eval "$(starship init zsh)"
```

#### Starship Configuration

```toml
[haskell]
disabled = true
```

DestyNova:
#### Description
This PR changes the arguments to `stack` in the Haskell module to avoid creating a `stack.yaml.lock` file which interferes with Git. It also reorders the arguments to avoid downloading GHC if the required version isn't already installed.

#### Motivation and Context
Starship was creating `stack.yaml.lock` files, even right after deleting them. This made it awkward for me to pull new changes to a Git repo containing a `stack.yaml` -- in fact I had to open a bash shell without Starship.
After I had updated the repo, Starship hung for quite a while before I killed it. I'm not 100% sure, but suspect that it was downloading and building a new version of GHC.

In terms of both the `stack.yaml.lock` and downloading GHC behaviour, I think Starship should generally be side-effect free.

Closes #1146.

#### Types of changes
- [?] Breaking change (fix or feature that would cause existing functionality to not work as expected)

1. Executed `stack` manually and messed with the command line args until it had the desired behaviour
1. Updated the Haskell module to use the same args.
1. Built with `cargo build`, then tested in a new shell by pasting the results of `./target/debug/starship prompt` and checking that lock files are not created.

#### Checklist:

#### Problems

~~`cargo test modules::haskell::tests` passes all 4 relevant tests, but `cargo test` fails for those exact same tests. I have absolutely no clue as to why this would be the case, and spent some time debugging to see if `cargo test modules::haskell::tests` was somehow skipping the actual assertions (by making it so they must fail), but it's definitely running the tests in both cases.~~

~Any insight here would be appreciated!~

Ok, figured it out and the test suite passes normally now!

### DIFF
--- a/src/modules/haskell.rs
+++ b/src/modules/haskell.rs
@@ -22,7 +22,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let haskell_version = utils::exec_cmd(
         "stack",
-        &["ghc", "--", "--numeric-version", "--no-install-ghc"],
+        &["--no-install-ghc", "--lock-file", "read-only", "ghc", "--", "--numeric-version"],
+
     )?
     .stdout;
     let formatted_version = Some(format!("v{}", haskell_version.trim()))?;
--- a/src/modules/haskell.rs
+++ b/src/modules/haskell.rs
@@ -22,8 +22,14 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let haskell_version = utils::exec_cmd(
         "stack",
-        &["--no-install-ghc", "--lock-file", "read-only", "ghc", "--", "--numeric-version"],
-
+        &[
+            "--no-install-ghc",
+            "--lock-file",
+            "read-only",
+            "ghc",
+            "--",
+            "--numeric-version",
+        ],
     )?
     .stdout;
     let formatted_version = Some(format!("v{}", haskell_version.trim()))?;
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -73,10 +73,12 @@ pub fn exec_cmd(cmd: &str, args: &[&str]) -> Option<CommandOutput> {
             stdout: String::from("ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux-gnu]"),
             stderr: String::default(),
         }),
-        "stack ghc -- --numeric-version --no-install-ghc" => Some(CommandOutput {
-            stdout: String::from("8.6.5"),
-            stderr: String::default(),
-        }),
+        "stack --no-install-ghc --lock-file read-only ghc -- --numeric-version" => {
+            Some(CommandOutput {
+                stdout: String::from("8.6.5"),
+                stderr: String::default(),
+            })
+        }
         "elixir --version" => Some(CommandOutput {
             stdout: String::from(
                 "\


DestyNova:
Thanks @andytom -- I did suspect there must have been a mocked command but couldn't figure it out last night. Will take another look after work then.

PR #1156, opened 2020-04-30
[DO NOT MERGE] feat: Per-session messaging

### LINKED ISSUE #1075: Implement error messaging for the prompt

To date, we have not had a system for surfacing error messaging outside of debug logs. Error messages should serve as a way to alert the user when Starship is misconfigured, or warn them of upcoming deprecations.

Since starship is stateless and runs anew every time the prompt is rendered, we need a way to make error messages non-repetitive.

I was thinking that one way of doing that would be to only show a particular error message once per session. To do that, we would need to know when the last session was, which would require some form of statefulness (e.g. logging when the last prompt was rendered as well as which errors were printed that time).

Additionally, we will need a way to collect error messages from modules, which will require some rearranging of how modules work.

Modules should be able to return multiple errors each.

heyrict:
#### Description
This PR allows modules to add prepend messages to the prompt on the first time they start a console.

* Note 1: This PR is based on #1056 and should be blocked before it is merged.

* Note 2: This is just a PoC. Detailed design should be put under further discussion.

* Note 3: This workaround requires changing on init scripts. This PR currently only includes changes on `bash`, `zsh` and `fish`. It may not work if you use other shells.

#### Motivation and Context

Closes #1075 if merged

#### Types of changes

#### Checklist:


matchai:
Converting PR to draft to block merge 😜

PR #1158, opened 2020-04-30
feat(format-string): add format string support to battery module

lucarin91:
#### Description
Add format-string support to battery module.

#### Motivation and Context

part of #1057

#### Types of changes

#### Checklist:


docs/config/README.md
```diff
@@ -171,13 +171,14 @@ The module is only visible when the device's battery is below 10%.
...
 ### Options
 
-| Variable             | Default                  | Description                                       |
-| -------------------- | ------------------------ | ------------------------------------------------- |
-| `full_symbol`        | `"•"`                    | The symbol shown when the battery is full.        |
-| `charging_symbol`    | `"⇡"`                    | The symbol shown when the battery is charging.    |
-| `discharging_symbol` | `"⇣"`                    | The symbol shown when the battery is discharging. |
-| `display`            | [link](#battery-display) | Display threshold and style for the module.       |
-| `disabled`           | `false`                  | Disables the `battery` module.                    |
+| Variable             | Default                            | Description                                       |
```
heyrict, line 174:
```suggestion
| Option               | Default                            | Description                                       |
```
To prevent naming conflict with variable in a format string.

src/configs/battery.rs
```diff
@@ -1,40 +1,39 @@
...
-    pub charging_symbol: SegmentConfig<'a>,
-    pub discharging_symbol: SegmentConfig<'a>,
-    pub unknown_symbol: Option<SegmentConfig<'a>>,
-    pub empty_symbol: Option<SegmentConfig<'a>>,
-    pub display: Vec<BatteryDisplayConfig>,
+    pub full_symbol: &'a str,
+    pub charging_symbol: &'a str,
+    pub discharging_symbol: &'a str,
+    pub unknown_symbol: &'a str,
+    pub empty_symbol: &'a str,
```
heyrict, line 11: I think it would be better written as `Option<&'a str>` as `battery` module will not be displayed when battery status is `UNKNOWN` or `EMPTY`.

src/modules/battery.rs
```diff
@@ -14,60 +16,60 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                        battery::State::Full => Some(Ok(config.full_symbol)),
+                        battery::State::Charging => Some(Ok(config.charging_symbol)),
+                        battery::State::Discharging => Some(Ok(config.discharging_symbol)),
+                        battery::State::Unknown => Some(Ok(config.unknown_symbol)),
+                        battery::State::Empty => Some(Ok(config.empty_symbol)),
+                        _ => {
+                            log::debug!("Unhandled battery state `{}`", state);
+                            None
+                        }
+                    },
```
heyrict, line 52: I think `symbol` should be configured as a format string instead of a plain string.

It should be fine most of the case but there may be some users trying to configure a rainbow colored `FULL` or so.

Note that `map_meta` accepts a format string as a return value.

src/configs/battery.rs
```diff
@@ -4,11 +4,11 @@ use starship_module_config_derive::ModuleConfig;
 
 #[derive(Clone, ModuleConfig)]
 pub struct BatteryConfig<'a> {
-    pub full_symbol: &'a str,
-    pub charging_symbol: &'a str,
-    pub discharging_symbol: &'a str,
-    pub unknown_symbol: &'a str,
-    pub empty_symbol: &'a str,
+    pub full_symbol: Option<&'a str>,
```
heyrict, line 7: No, this will make users overwrite the default config with None, even if one does not specify the key `full_symbol`. Same updates should be done with charging and discharging symbol.

Use this.

```suggestion
    pub full_symbol: &'a str
```

lucarin91: Ok, sorry you are right. I will update it.

lucarin91:
Thanks for the suggestion @heyrict. I hope that now I implemented it correctly.
Moreover, someone knows why mac tests continue to fail? I do not think it is something related to my PR.

heyrict:
It's a bug in ci. Fix for the bug has landed in master in #1181

PR #1164, opened 2020-05-03
feat: add more configurable prefixes

### LINKED ISSUE #418: Allow configuration of prefixes and improve documentation on them

Currently, many modules have hard-coded prefixes, such as "in " on the `directory` module or "via " on all version modules. The prefix handling is pretty unclear to me; for example, in the default configuration, the `directory` module does not display a prefix, while the version modules do ("via ").

```
~ via ☕ v12.0.2
>
```

When moving the `directory` module to the next line (i.e. after `line_break`), it suddenly displays its prefix with no way to disable it, while the version module does not. My goal in this particular case is to remove the "in " prefix.

```
☕ v12.0.2
in ~ >
```

After some testing, I came to the conclusion that the prefix is only displayed if the module is preceded by another module (which the `line_break` module does not "reset", hence why `directory` displays its prefix even when alone on a new line). However, I'm not totally sure that's the case.

If the behavior I described is correct, it may be a good idea for modules coming directly after a `line_break` to not display their prefix (or to make this configurable). An even better solution could be to make the prefixes changeable or removable on all modules, or a combination of both.

**More specifically**:

* Add an explanation to the docs on what determines whether the prefix is shown or not.
* Make the prefix changeable or removable on all modules that include one.
* ~~(Possibly) Do not show the prefix on the  first module after a `line_break`.~~ (Fixed by #473)

### LINKED ISSUE #1102: Hostname module doesn't respect prefix/suffix

scarvalhojr:
#### Description

Add configurable prefixes for username, hostname, git_branch and
python modules.

#### Motivation and Context

Unlike with other modules, it's not possible to customise prefixes in the username, hostname, git_branch and python modules.

Closes #418 
Closes #1102 

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -583,6 +583,7 @@ The `git_branch` module shows the active branch of the repo in your current dire
 
 | Variable            | Default         | Description                                                                           |
 | ------------------- | --------------- | ------------------------------------------------------------------------------------- |
+| `prefix`            | `"on "`         | Prefix to display immediately before the Git branch.                                  |
 | `symbol`            | `" "`          | The symbol used before the branch name of the repo in your current directory.         |
 | `truncation_length` | `2^63 - 1`      | Truncates a git branch to X graphemes                                                 |
 | `truncation_symbol` | `"…"`           | The symbol used to indicate a branch name was truncated. You can use "" for no symbol |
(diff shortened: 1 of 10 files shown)


scarvalhojr:
I understand there is work going on to introduce format strings, which would make this change irrelevant, but it will allow for better configuration while format strings are not released.

Let me know if there is any interest in merging this and I can update the docs and tests accordingly.

heyrict:
Hey, thank you for your PR! Currently format-string related PRs are in review process. It is expected to see it landed in master *soon*. If you are interested, there is a umbrella issue in #1057.

Going to close this PR.

PR #1178, opened 2020-05-07
fix(directory): fix logic for logical (shell dependent) path management  

jeanga:
#### Description
Some supported shells, like Powershell, support "logical path" that are not "file system" paths.
(typically cert: drive for certificates, variable: for vars, env: for env variables, etc... you can even create your owns).

This PR introduces a new command line argument --logical-path to clearly indicate starship the distinction between the "logical" passed with --logicapl-path and the "physical" or "file system" path.

This PR introduces a new configuration setting for the directory module: separator.
This "string" setting allows to configure the path separator (usually/typically "/") with a string of user's choosing.
Typically: 
image
Will give:
image

#### Motivation and Context
The discussion on issue #1106 was the main driver behind this PR to reliably and durably fix logical/physical distinction starship. The previous reliance on $PWD/--path wasn't distinctive enough.

Closes [#1106](https://github.com/starship/starship/issues/1106)

#### Types of changes

New separator setting:
image

#### Checklist:


jeanga:
defaulting to std::path::MAIN_SEPARATOR instead of "/" has had a nasty side effect on directory tests on Windows.
I am fixing them with a number of:
```
        .use_config(toml::toml! {
            [directory]
            separator = "/"
        })
```
I chose this solution as duplicating all tests (for Windows/non-Windows) would be (IMHO) counter productive.

src/modules/directory.rs
```diff
@@ -161,6 +152,57 @@ fn to_fish_style(pwd_dir_length: usize, dir_string: String, truncated_dir_string
         .join("/")
 }
 
+#[cfg(target_os = "windows")]
+fn to_separator(path: &Path, separator: &str) -> Option<String> {
```
davidkna, line 156: 1. Have you investigated unifying the windows and -nix implementations? Gating platform behaviour in `#[cfg(windows)]` and/or `if cfg!(windows)`?
2. I think this will never return `None`, maybe switch to plain `String`?

src/modules/directory.rs
```diff
@@ -161,6 +152,57 @@ fn to_fish_style(pwd_dir_length: usize, dir_string: String, truncated_dir_string
...
+    let first = iter.next();
+    if let Some(first_component) = first {
+        out.push_str(&first_component.to_string_lossy());
+        for component in iter {
+            out.push_str(&separator);
+            out.push_str(&component.to_string_lossy());
+        }
+    }
+
+    if path_str.ends_with('\\') || path_str.ends_with('/') {
```
davidkna, line 173: When does this trigger?

src/modules/directory.rs
```diff
@@ -161,6 +152,57 @@ fn to_fish_style(pwd_dir_length: usize, dir_string: String, truncated_dir_string
...
+    }
+
+    Some(out)
+}
+
+#[cfg(not(target_os = "windows"))]
+fn to_separator(path: &Path, separator: &str) -> Option<String> {
+    let path_str = path.as_os_str().to_string_lossy();
+
+    if path_str == "/" {
```
davidkna, line 184: Maybe use `std::path::MAIN_SEPARATOR` here instead of `/`?

davidkna:
> 
> 
> defaulting to std::path::MAIN_SEPARATOR instead of "/" has had a nasty side effect on directory tests on Windows.
> I am fixing them with a number of:
> 
> ```
>         .use_config(toml::toml! {
>             [directory]
>             separator = "/"
>         })
> ```
> 
> I chose this solution as duplicating all tests (for Windows/non-Windows) would be (IMHO) counter productive.

Considering that and all the trouble with using `std::path::MAIN_SEPARATOR` as a default it might make sense to default to `/` instead.

src/modules/directory.rs
jeanga: I did this today for the Linux tests to pass but I probably should remove this now that most of them are separated with #[cfg(windows)].

src/modules/directory.rs
jeanga: Yes, will do. I will have to exclude empty string first and test on len ( std::path::MAIN_SEPARATOR is a char, not a str, remember ;-))

jeanga:
Well, I reckon it created some trouble in the code :-)
However, I think it is for the better on the long run as Windows path will 'always' be \ separated, it makes sense to be consistent and use the native path separator. 
The directory module has to do a great job on directories ;-) (per #1106 )

src/modules/directory.rs
jeanga: 1: Not sure, windows&unix paths have very unique characteristics, like absolute paths always starting with / on Linux and on Windows they are relative to the current drive...
2: Sure, done.

jeanga:
Okay, after taking some time to think about it, here is what I intend to implement to sort out the current situation with tests and std::path::MAIN_SEPARATOR issue:
If std::path::MAIN_SEPARATOR is _not_ "/", then replace all occurrences of std::path::MAIN_SEPARATOR with "/" in all inputs (home, repo, path, logical path.
When everything is "normalized", we go through truncations, fish style, repo&home replacement etc...
Once the "final" string is computed, we replace all occurrences of "/" with the configured separator in configuration (if not "/") (and, the home string we're also adding (if not "~")).
This, IMHO, is the best way forward... 
Expect some new commit on this PR or a brand new one in the next very few (hopefully) days.

jeanga:
OK. After some longer than I thought was needed, I took a completely different approach to this problem and rewrote the entire module. 
I will close this PR now and will soon submit a new one (also closing #1174 .
Thank you for your patience

PR #1183, opened 2020-05-09
feat(directory): Add directory substitutions

### LINKED ISSUE #1065: Feature request: current directory substitutions
## Feature Request
I'd like an option for the directory module for a list of arbitrary substitutions to be made to the PWD to shorten it in a context-aware way. It could happen instead of (or before?) the truncation to `truncation_length`. I currently use this in my bash prompt to remove a long common prefix (network drive) from my paths so I can put them on $PS1. Specifically, instead of `/.../.../.../A/.../B/C`, I get `A/./B/C`, where `...` is multiple letters and `.` is just one.

It could be configured as a list of (literal, not regex) substitutions to make (in order) like:
```
[directory]
substitute = [
  [ "/foo/bar", "/f" ],
  [ "/f", "/baz" ]
]
```
What do you think, would anyone find this useful?

rbutoi:
Hello, this is my first real piece of rust code, so any tips are greatly appreciated! I left some review comments myself.

#### Description
This adds an option to provide a list of pairs of strings to substitute in the directory string. If this option is used, fish-style PWD is disabled.

The reason for fish-style PWD being disabled is an implementation detail -- I perform the `replace` on the `dir_string`, since it is a string type rather than a path type. To make this work with fish-style PWD would need more code, and IMO I don't see the need to use both features at once. Would definitely appreciate more input here.

#### Motivation and Context
I'm currently using this at work to shorten paths that have a long common prefix, interesting middle, and some long development directories. It's a more tailored way of shortening paths while retaining more information than, say the fish-style.
Closes #1065.

#### Types of changes

Added a unit test and two "testsuite" tests. They test the interaction with `truncation_length` and (non-)interaction with `fish_style_pwd_dir_length`.

#### Checklist:

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -1,4 +1,5 @@
 use path_slash::PathExt;
+use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use unicode_segmentation::UnicodeSegmentation;
 
@@ -10,11 +11,14 @@ use crate::configs::directory::DirectoryConfig;
 
 /// Creates a module with the current directory
 ///
-/// Will perform path contraction and truncation.
+/// Will perform path contraction, substitution, and truncation.
 /// **Contraction**
 ///     - Paths beginning with the home directory or with a git repo right
 /// inside the home directory will be contracted to `~`
 ///     - Paths containing a git repo will contract to begin at the repo root
+///
+/// **Substitution**
+/// Paths will undergo user-provided substitutions of substrings
 ///
 /// **Truncation**
 /// Paths will be limited in length to `3` path components by default.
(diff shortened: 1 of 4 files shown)


src/modules/directory.rs
```diff
@@ -223,6 +246,15 @@ mod tests {
         assert_eq!(output, "C:/");
     }
 
+    #[test]
+    fn substitute_prefix_and_middle() {
+        let full_path = "/absolute/path/foo/bar/baz";
+        let substitutions = vec![["/absolute/path", ""].to_vec(), ["/bar/", "/"].to_vec()];
```
rbutoi, line 252: Is `vec![[].to_vec(), ...]` the best way to write this vector-of-vectors in code? Related, does taking a `&[Vec<&str>]` parameter to pass a `&Vec<Vec<&str>>` make sense?  I was mostly following compiler suggestions here.

nickwb:
The list-of-list for substitutions seems a little odd. Minimally, this can be upgraded to `Vec<(String, String)>`, though I wonder if a map/set would be nicer from a configuration perspective..?

The feature seems like a good idea.

rbutoi:
> The list-of-list for substitutions seems a little odd. Minimally, this can be upgraded to `Vec<(String, String)>`, though I wonder if a map/set would be nicer from a configuration perspective..?

Agreed that the list-of-lists isn't ideal here; the inner lists must be of length two. However, TOML doesn't have tuples (they ended up becoming inline tables: https://github.com/toml-lang/toml/issues/131, https://github.com/toml-lang/toml/pull/154, https://github.com/toml-lang/toml/pull/235). Going forward, I see two possibilities: a list of inline tables with "from" and "to" keys, or a simple map, as you mention. The difference between the two is that map iteration, and therefore substitution order, is undefined. This shouldn't be a problem, and I can't think of any reasons for a multi-stage substitution that would need an explicit order, but worth noting.

I've pushed a commit that changes the config to a map, which is the simpler option of the two.

> The feature seems like a good idea.

Glad to hear that!

rbutoi:
^ Above check failures seem unrelated, they all passed before and I only changed the config type from vec to hashmap.

matchai:
The current failures were caused by a change in the GitHub Actions toolcache. Should be fixed in #1243.

andytom:
@rbutoi thank you for the contribution and @nickwb thank you for reviewing this.

PR #1185, opened 2020-05-10
fix: character module last command status for bash

MunifTanjim:
#### Description

For **bash**, appending `starship_precmd` to `PROMPT_COMMAND` breaks previous command's status (`$?`) checking when other commands already exist in `PROMPT_COMMAND`. Because those commands can modify the actual `$?` variable.

#### Motivation and Context

Related to #954, which was closed without solving the underlying issue that caused it.

#### Types of changes

Screenshot from 2020-05-11 05-24-02

#### Checklist:

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -22,7 +22,7 @@ starship_preexec() {
         PREEXEC_READY=false
         STARSHIP_START_TIME=$(::STARSHIP:: time)
     fi
-    
+
     : "$PREV_LAST_ARG"
 }
 
@@ -64,14 +64,14 @@ else
         }
         trap 'starship_preexec_all "$_"' DEBUG
     fi
- 
-    # Finally, prepare the precmd function and set up the start time. We will avoid to 
+
+    # Finally, prepare the precmd function and set up the start time. We will avoid to
     # add multiple instances of the starship function and keep other user functions if any.
     if [[ -z "$PROMPT_COMMAND" ]]; then
         PROMPT_COMMAND="starship_precmd"
     elif [[ "$PROMPT_COMMAND" != *"starship_precmd" ]]; then
-        # Remove any trailing semicolon before appending (PR #784)
-        PROMPT_COMMAND="${PROMPT_COMMAND%;};starship_precmd;"
+        # prepending starship_precmd (appending it breaks $? status checking)
+        PROMPT_COMMAND="starship_precmd;${PROMPT_COMMAND}"
     fi
 fi
 


MunifTanjim:
Currently, I'm doing this as a workaround:

```sh
eval "$(starship init bash)"
export PROMPT_COMMAND="starship_precmd;${PROMPT_COMMAND/starship_precmd;/}"
```

It adds the `starship_precmd` command to the start of `PROMPT_COMMAND` and removes it from the end.

X-dark:
This breaks command duration for me as idle time become counted as well.

MunifTanjim:
> This breaks command duration for me as idle time become counted as well.

Strange. The workaround works fine. But the change in this PR breaks **command duration**. :thinking: 

@X-dark can you confirm if the workaround (without the change in this PR) works for you?

X-dark:
> @X-dark can you confirm if the workaround (without the change in this PR) works for you?
Sorry I have tested the workaround, not the PR.

Without any fix or workaround, my PROMPT_COMMAND looks like this:

```
__vte_prompt_command; printf "\033]0;%s@%s:%s\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\~}";starship_precmd;
```

MunifTanjim:
Thanks @X-dark . I've updated the PR to handle those cases.

src/init/starship.bash
```diff
@@ -64,14 +66,18 @@ else
...
-    # Finally, prepare the precmd function and set up the start time. We will avoid to 
+
+    # Finally, prepare the precmd function and set up the start time. We will avoid to
     # add multiple instances of the starship function and keep other user functions if any.
     if [[ -z "$PROMPT_COMMAND" ]]; then
         PROMPT_COMMAND="starship_precmd"
     elif [[ "$PROMPT_COMMAND" != *"starship_precmd" ]]; then
-        # Remove any trailing semicolon before appending (PR #784)
-        PROMPT_COMMAND="${PROMPT_COMMAND%;};starship_precmd;"
+        # Appending to PROMPT_COMMAND breaks exit status ($?) checking.
```
chipbuster, line 75: Thanks for the comment here--will really help in the future.

chipbuster:
I have to speak frankly here for a second: the number of hoops we've jumped through on this project to try to get things working perfectly in Bash is starting to get ludicrous. First we set, then we appended, then we prepended. We've changed separators between semicolons and whitespace, we've had to deal with other shell helpers breaking starship with their own initscripts. Every single time, I've been optimistic that **this** change will finally fix everything, and every time, I've been proven wrong. And judging by the number of conflicts we've gotten with other shell frameworks, I'm not sure they're doing much better than us.

I certainly hope that this will fix our problems once and for all (and looking at it, I do think it will!), but in the event that I'm proven wrong again, I'm going to have to suggest that we push that responsibility on the user in one of two ways:

- Request that they install bash-preexec, which abstracts away this problem and we already have support for.
- Force the user to decide their own ordering: make an init flag for bash only (e.g. `starship init bash --funcs-only`) which will only define `starship_precmd` and `starship_preexec`, and decide that it's the user's responsibility to figure out how they want their hooks ordered.

PR #1209, opened 2020-05-14
refactor(directory): Rewrite directory module with support for separator&home configuration settings and --path&--logical-path disctinction

### LINKED ISSUE #1174: Allow us to modify home directory symbol (~) in [directory] options
Please give us a possibility to replace the default symbol for the **home directory** _(tilde character)_ **'~'** to any other symbol, such as home emoticon -🏠.

Add an option under the **[directory]** settings in configuration file ```starship.toml``` _(default)_:
```
[directory]
home_directory_symbol = "🏠"
```

Which should give us an excepted result like this:
```
via USER at HOSTNAME in 📂: 🏠/Projects/ExampleDirectory
```

**Why?**
I strongly believe the main benefit of a Starship is to be highly customizable. Some people, like me, love to add more icons. 😉

jeanga:
#### Description
Some supported shells, like Powershell, support "logical path" that are not "file system" paths.
(typically cert: drive for certificates, variable: for vars, env: for env variables, etc... you can even create your owns).

This PR introduces a new command line argument --logical-path to clearly indicate starship the distinction between the "logical" passed with --logicapl-path and the "physical" or "file system" path.

This PR introduces two configuration settings for the directory module: separator&home.
This "string" setting allows to configure the path separator (usually/typically "/") with a string of user's choosing.
Typically: 
image
Will give:
image

And for home:
image

#### Motivation and Context
The discussion on issue #1106 was the main driver behind this PR to reliably and durably fix logical/physical distinction starship. The previous reliance on $PWD/--path wasn't distinctive enough.

Closes [#1106](https://github.com/starship/starship/issues/1106)
Closes #1174 

#### Types of changes

New separator setting:
image

#### Checklist:


docs/de-DE/config/README.md
```diff
@@ -360,6 +360,8 @@ For example, given `~/Dev/Nix/nixpkgs/pkgs` where `nixpkgs` is the repo root, an
 | `truncation_length` | `3`           | Die Anzahl der übergeordneten Ordner, die angezeigt werden.                      |
```
matchai, line 360: Please remove the updates to the translated docs. Only the English docs need to be updated.
These files will be updated automatically by a Crodwin PR.

jeanga: Arggh....  I did not know that... sorry.
Will fix that.

jeanga: Sorry for asking a stupid question but should I cancel this PR and submit a new one without those files modified or there is another option?

jeanga:
I am not sure I understand this git_repo_in_home_directory_truncate_to_repo_true test.
Should'nt truncate_to_repo=true imply that the directory is truncated to fuel-gauge ?
(i.e. not preserve the ~/src/meters/?)
and truncation_length = 5 will remove the root of the repo if you are deep inside the repo tree?

Thank you for your help!
Jean

docs/config/README.md
```diff
@@ -383,6 +383,8 @@ it would have been `nixpkgs/pkgs`.
 | `truncation_length` | `3`           | The number of parent folders that the current directory should be truncated to.  |
 | `truncate_to_repo`  | `true`        | Whether or not to truncate to the root of the git repo that you're currently in. |
 | `prefix`            | `"in "`       | Prefix to display immediately before the directory.                              |
+| `separator`         | `"/"`         | Separator used between each element of the directory path                        |
```
davidkna, line 386: `"/"` is not the actual default separator in your code on Windows (and other platform that don't use `/` as the main seperator)

docs/de-DE/config/README.md
```diff
@@ -1,1302 +0,0 @@
-# Konfiguration
```
davidkna, line 1: I don't think you should delete all the readme files.

src/modules/directory.rs
```diff
@@ -101,169 +121,862 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-            .to_slash()
-            .unwrap()
-    )
+#[derive(PartialEq, Clone, Debug)]
+pub enum DirectoryComponent {
+    Prefix(String),
+    Root,
+    Current,
+    Parent,
+    Home,
```
davidkna, line 130: Would `Normal(options.home)` work instead of `Home`?

src/modules/directory.rs
```diff
@@ -101,169 +121,862 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+
+    fn to_separator(&self, separator_option: Option<&str>, home_options: Option<&str>) -> String {
+        let default_separator = std::path::MAIN_SEPARATOR.to_string();
+        let separator = separator_option.unwrap_or(&default_separator);
+        let home = home_options.unwrap_or(HOME_SYMBOL);
+
+        let mut size = self
+            .components
+            .iter()
+            .fold(0u32 as usize, |count, comp| match comp {
```
davidkna, line 302: I think `.map(...).sum()` is cleaner

docs/config/README.md
jeanga: ok

docs/de-DE/config/README.md
jeanga: Sorry, messed up... :-( misunderstood what what the delete in the PR is doing... 
I will find a way to fix that.

src/modules/directory.rs
jeanga: Probably but keeping it as a string would only confuse when replacing with the configured symbol is the only use for it.
This way, we don't have to check if this "Normal" is a home.
Also, we could evolve it into Home(String) if we want to support ~user/folder which Normal(options.Home) won't support.

docs/de-DE/config/README.md
davidkna: Download a clean repo, copy the readme files and commit them.

src/modules/directory.rs
jeanga: yes, agreed.

docs/de-DE/config/README.md
jeanga: k

jeanga:
I don't have any clue why this test is failing:
```
---- directory::git_repo_in_home_directory_truncate_to_repo_true stdout ----
thread 'directory::git_repo_in_home_directory_truncate_to_repo_true' panicked at 'assertion failed: `(left == right)`
  left: `"in \u{1b}[1;36mfuel-gauge\u{1b}[0m "`,
 right: `""`', tests/testsuite/directory.rs:672:5
```
It passed just fine on my WSL (Ubuntu 20.04) install.

src/configs/directory.rs
```diff
@@ -22,6 +24,15 @@ impl<'a> RootModuleConfig<'a> for DirectoryConfig<'a> {
             fish_style_pwd_dir_length: 0,
             use_logical_path: true,
             prefix: "in ",
+            separator: match std::path::MAIN_SEPARATOR {
```
matchai, line 27: **question:**
What does this `match` do that a `char.to_string()` wouldn't do?

src/configs/directory.rs
```diff
@@ -22,6 +24,15 @@ impl<'a> RootModuleConfig<'a> for DirectoryConfig<'a> {
...
+                '/' => "/",
+                '\\' => "\\",
+                '.' => ".",
+                ':' => ":",
+                '>' => ">",
+                _ => panic!("Unknown Seperator"),
```
matchai, line 33: **issue**:
I don't think it would be appropriate for the prompt to panic if the `MAIN_SEPARATOR` isn't on this list. Perhaps we should fallback on `/`?

src/context.rs
```diff
@@ -50,11 +51,13 @@ impl<'a> Context<'a> {
                 })
             });
 
-        Context::new_with_dir(arguments, path)
+        let logical_path: Option<PathBuf> = arguments.value_of("logical-path").map(From::from);
+
+        Context::new_with_dir(arguments, path, logical_path)
     }
 
     /// Create a new instance of Context for the provided directory
```
matchai, line 59: **nitpick:**
Could the comment please be updated to reflect the new argument. Perhaps explaining when `logical_path` is to be used.

src/context.rs
```diff
@@ -50,11 +51,13 @@ impl<'a> Context<'a> {
...
-    pub fn new_with_dir<T>(arguments: ArgMatches, dir: T) -> Context
+    pub fn new_with_dir<T>(arguments: ArgMatches, path: T, logical_path: Option<PathBuf>) -> Context
```
matchai, line 60: **suggestion:**
Since we already have the `T: Into<PathBuf>` here, it seems sensible to use it for `logical_path` too:
```suggestion
    pub fn new_with_dir<T>(arguments: ArgMatches, path: T, logical_path: Option<T>) -> Context
```

src/context.rs
```diff
@@ -70,15 +73,15 @@ impl<'a> Context<'a> {
             .map(|(a, b)| (*a, b.vals.first().cloned().unwrap().into_string().unwrap()))
             .collect();
 
-        // TODO: Currently gets the physical directory. Get the logical directory.
-        let current_dir = Context::expand_tilde(dir.into());
-
+        let current_dir = Context::expand_tilde(path.into());
+        let logical_dir = logical_path;
```
matchai, line 77: ☝️ 
```suggestion
        let logical_dir = logical_path.map(Into::into);
```

src/context.rs
```diff
@@ -70,15 +73,15 @@ impl<'a> Context<'a> {
...
         let shell = Context::get_shell();
 
         Context {
             config,
             properties,
             current_dir,
+            logical_dir,
             dir_contents: OnceCell::new(),
```
matchai, line 85: **thought:**
It looks like we're starting to have a good number of directory-specific properties in `Context`. It may be sensible to create a `Directory` struct within `Context` to house them.

src/main.rs
```diff
@@ -35,7 +35,14 @@ fn main() {
...
         .value_name("PATH")
-        .help("The path that the prompt should render for")
+        .help("The filesystem path that the prompt should render for")
+        .takes_value(true);
+
+    let logical_path_arg = Arg::with_name("logical-path")
+        .short("P")
+        .long("logical-path")
+        .value_name("LOGICALPATH")
+        .help("The shell path that the prompt should render for")
```
matchai, line 45: **suggestion:**
Please update the help string to distinguish between `path` and `logical-path`.

src/configs/directory.rs
davidkna: see my earlier comments

src/configs/directory.rs
davidkna: I didn't suggest that so that any such emerging problems can be handled quickly. Fallback sounds good too.

src/context.rs
jeanga: Just pushed a new commit for this

src/context.rs
jeanga: As a matter of fact, current_dir & dir_contents are used by a variety of modules and are "general" purpose stuff (like repo is to the directory module). As I read it, only logical_dir is directory module specific.
Well, you call really...

src/configs/directory.rs
jeanga: Your call guys. Either paths are Ok to me.

jeanga:
@matchai this PR is 95% rewrite of the directory module, mostly diverging from the previous "split then join then split again then join again" pattern. Removing the logical/file system path distinction will put us back in the troubles that #1114 & #1106 tried to solve.

Only to_separator implements the home & separator substitution.
I can put that function into a separate PR if you'd like.

src/configs/directory.rs
matchai: How about `str::utf8` as an alternative solution:
```suggestion
            separator: std::str::from_utf8(&[std::path::MAIN_SEPARATOR as u8]).unwrap(),
```

davidkna: Given that `std::path::MAIN_SEPARATOR` is `u32` should `try_into` be used here? Also I don't think all single byte utf8 chars map one-to-one to `u32` Unicode scalar values.

jonathandturner:
Is there anything we can do from the Nushell side to help with this PR? We'd like to be able to have native path separators in Windows for Nu, and would be happy to help whee we can to make that possible.

src/modules/directory.rs
```diff
@@ -101,169 +121,863 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    fn truncate_larger_path_than_provided_length_from_root_windows() {
+        let mut dir = Directory::from(r"C:\starship\engines\booster\rocket");
+        dir.truncate(3);
+        assert_eq!(
+            dir.to_separator(Some(r"\"), None),
+            r"engines\booster\rocket"
+        );
+    }
+
+    #[test]
```
Tadaboody, line 589: Shouldn't this be windows-only as well?

jeanga: I am afraid there has been too much changes to this module's code since this PR was submitted.
Not sure it can still be merged now after so much time in the backlog.

DiegoMagdaleno:
Is this still happening? I would love to be able to use \ instead of / on Windows

korrat:
As far as I can tell, there has been no activity on this PR for since June 2020. Also, the logical path handling seems to be implemented in #2104 already. Maybe this PR should be closed in favor of a new PR that focuses on allowing configuration of the path separator?

PR #1216, opened 2020-05-15
docs: Updated config with a tip for PowerShell user

xeho91:
Added a tip with information about how a PowerShell user can
override default starship.toml file location path in his PowerShell
profile.

#### Motivation and Context

It took me a while to figure out a way how I can change the default file location for starship configuration in PowerShell. The main motivation for me was to make one config file, which could be shared across shells on my computer (zsh and PowerShell).

Closes #

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -33,6 +33,11 @@ You can change default `starship.toml` file location with `STARSHIP_CONFIG` envi
 export STARSHIP_CONFIG=~/.starship
 ```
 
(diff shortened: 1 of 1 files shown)


andytom:
Thanks for the contribution @xeho91 

> That makes me realize that the documentation is heavier on the POSIX compliant systems. Maybe we should have different sections for different environments for the documentation on more sensitive configuration.

@jimleroyer That is a great idea would it be possible for you to create an issue for this so it doesn't get lost?

PR #1221, opened 2020-05-16
refactor(git_state): Use format strings

ATiltedTree:
#### Description

This PR deprecates `progress_divider` and `style` keys in the `git_state` module to replace it with the `format` key instead.
It also enables the test that were ignored before.

#### Motivation and Context

Related to issue #1057

#### Types of changes

image

#### Checklist:

### DIFF
--- a/src/modules/git_state.rs
+++ b/src/modules/git_state.rs
@@ -1,8 +1,9 @@
 use git2::RepositoryState;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
-use super::{Context, Module, RootModuleConfig, SegmentConfig};
+use super::{Context, Module, RootModuleConfig};
 use crate::configs::git_state::GitStateConfig;
+use crate::formatter::StringFormatter;
 
 /// Creates a module with the state of the git repository at the current directory
 ///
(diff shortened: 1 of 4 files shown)


src/modules/git_state.rs
```diff
@@ -12,37 +13,43 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    let state_description = get_state_description(repo_state, repo_root, &config)?;
+
+    let parsed = StringFormatter::new(config.format).and_then(|formatter| {
+        formatter
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "state" => Some(Ok(state_description.label)),
```
heyrict, line 29: What do you think about making `label` format strings, so that users can write e.g. `[REBASING](bold green)` to change the styles?

```suggestion
            .map_meta(|variable, _| match variable {
                "state" => Some(state_description.label),
```

src/modules/git_state.rs
```diff
@@ -12,37 +13,43 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "progress_current" => state_description.current.as_ref().map(Ok),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "progress_total" => state_description.total.as_ref().map(Ok),
+                _ => None,
+            })
```
heyrict, line 39: Can these two be combined?

```suggestion
            .map(|variable| match variable {
                "progress_current" => state_description.current.as_ref().map(Ok),
                "progress_total" => state_description.total.as_ref().map(Ok),
                _ => None,
            })
```

src/modules/git_state.rs
ATiltedTree: I think that's a good change! Should I update the docs to reflect this somehow?

heyrict: Yes, go ahead :+1:

PR #1223, opened 2020-05-16
refactor(conda): Use format strings

ATiltedTree:
#### Description

This PR deprecates `symbol` and `style` keys in the `conda` module to replace it with the `format` key instead.
It also changes the default symbol to the reference implementation.

#### Motivation and Context

Related to issue #1057

#### Types of changes

image

#### Checklist:

### DIFF
--- a/tests/testsuite/conda.rs
+++ b/tests/testsuite/conda.rs
@@ -2,15 +2,10 @@ use ansi_term::Color;
 use std::io;
 
 use crate::common;
-use crate::common::TestCommand;
 
 #[test]
 fn not_in_env() -> io::Result<()> {
-    let output = common::render_module("conda")
-        .env_clear()
-        .env("PATH", env!("PATH"))
-        .use_config("".parse().unwrap())
-        .output()?;
+    let output = common::render_module("conda").output()?;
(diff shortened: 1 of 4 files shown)


tests/testsuite/conda.rs
```diff
@@ -2,15 +2,10 @@ use ansi_term::Color;
...
-use crate::common::TestCommand;
 
 #[test]
 fn not_in_env() -> io::Result<()> {
-    let output = common::render_module("conda")
-        .env_clear()
-        .env("PATH", env!("PATH"))
-        .use_config("".parse().unwrap())
-        .output()?;
+    let output = common::render_module("conda").output()?;
```
heyrict, line 8: Hmm, is this change required? I'm wondering if removing `.env_clear()` will make the development environment messed up with the tests.

ATiltedTree: `.env_clear()` is actually already called by `common::render_module()`.
https://github.com/starship/starship/blob/77c42dca734b9ee537c5c6351d1422bb025b2d30/tests/testsuite/common.rs#L32-L44

PR #1229, opened 2020-05-17
refactor(nix_shell): Use format strings 

ATiltedTree:
#### Description

This PR deprecates `symbol`, `style` and `use_name` keys in the `nix_shell` module to replace it with the `format` key instead.
It also removes the exception for `lorri` as it was fixed [here](https://github.com/target/lorri/pull/164).

#### Motivation and Context

Related to issue #1057

#### Types of changes

image

#### Checklist:

### DIFF
--- a/src/modules/nix_shell.rs
+++ b/src/modules/nix_shell.rs
@@ -1,11 +1,9 @@
 use std::env;
 
-use super::{Context, Module, RootModuleConfig, SegmentConfig};
+use super::{Context, Module, RootModuleConfig};
 
 use crate::configs::nix_shell::NixShellConfig;
-
-// IN_NIX_SHELL should be "pure" or "impure" but lorri uses "1" for "impure"
-// https://github.com/target/lorri/issues/140
+use crate::formatter::StringFormatter;
 
 /// Creates a module showing if inside a nix-shell
 ///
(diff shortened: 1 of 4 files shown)


src/modules/nix_shell.rs
```diff
@@ -1,55 +1,66 @@
...
 
-    module.set_style(config.style);
-    module.create_segment("symbol", &config.symbol);
-
+    let shell_name = env::var("name").ok();
     let shell_type = env::var("IN_NIX_SHELL").ok()?;
-    let shell_type_segment: SegmentConfig = match shell_type.as_ref() {
-        "1" | "impure" => config.impure_msg,
+    let shell_type_format = match shell_type.as_ref() {
+        "impure" => config.impure_msg,
```
heyrict, line 29: Sorry, I'm not familiar with nix_shell. Does that mean `$IN_NIX_SHELL` will never be `1`?

ATiltedTree: I'm not familiar with it either but it says in the docs that `$IN_NIX_SHELL` should only be `pure` or `impure`:
> IN_NIX_SHELL
>
>   Indicator that tells if the current environment was set up by nix-shell. Since Nix 2.0 the values are "pure" and "impure"

Reference: https://nixos.org/nix/manual/#sec-common-env

PR #1230, opened 2020-05-17
refactor(terraform): Use format strings

ATiltedTree:
#### Description

This PR deprecates `symbol`, `style` and `show_version` keys in the `terraform` module to replace it with the `format` key instead.

#### Motivation and Context

Related to issue #1057

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1435,7 +1435,7 @@ format = "[📦 \\[$env\\]]($style) "
 
 The `terraform` module shows the currently selected terraform workspace and version.
 By default the terraform version is not shown, since this is slow on current versions of terraform when a lot of plugins are in use.
-I you still want to enable it, [follow the example shown below](#with-version).
+If you still want to enable it, [follow the example shown below](#with-version).
 The module will be shown if any of the following conditions are met:
 
 - The current directory contains a `.terraform` folder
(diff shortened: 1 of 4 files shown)


src/modules/terraform.rs
```diff
@@ -25,20 +27,42 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("terraform");
     let config: TerraformConfig = TerraformConfig::try_load(module.config);
 
-    module.set_style(config.style);
-    module.create_segment("symbol", &config.symbol);
+    let terraform_version =
+        format_terraform_version(&utils::exec_cmd("terraform", &["version"])?.stdout.as_str());
+    let terraform_workspace = get_terraform_workspace(&context.current_dir);
```
heyrict, line 32: Well, since calculating terraform version is very slow on large workspaces, you'd better move this to the `.map` function so that it will be calculated only when the formatter saw a `$version` variable.

Also users should be informed about this in the doc.

docs/config/README.md
```diff
@@ -1435,27 +1435,50 @@ format = "[📦 \\[$env\\]]($style) "
 
 The `terraform` module shows the currently selected terraform workspace and version.
 By default the terraform version is not shown, since this is slow on current versions of terraform when a lot of plugins are in use.
+I you still want to enable it, [follow the example shown below](#with-version).
```
heyrict, line 1438:
```suggestion
If you still want to enable it, [follow the example shown below](#with-version).
```

PR #1237, opened 2020-05-19
fix(custom): improve handling of Powershell prompts

### LINKED ISSUE #1220: Custom Module On Windows (Powershell) Crashing
#### Current Behavior
When configure a custom module and not specify any `shell`, the terminal crashing when any condition is met.

#### Expected Behavior
Output of custom module command in terminal.

When any condition of custom module are met, the terminal crashing and it's possible to see many extras powershell instances spawning.

image

When add some extra parameters in `shell` segment of the custom module, the problem is solved. Ex:

```toml
[custom.foo]
command = "time /T"
files = ["foo"]
prefix = " transcending "
shell = "powershell -NoProfile -Command"
```

#### Possible Solution
Document additional params necessary to `shell` segment to powershell or develop some mechanism to detect shell recursion.

#### Environment
- Starship version: 0.41.1
- powershell version: 5.1
- Terminal emulator: Windows Terminal with Powershell

71:
#### Description
In custom modules, add the arguments `-Command -` to the shell arguments when using PowerShell.

#### Motivation and Context
Closes #1220.

#### Types of changes

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1358,7 +1358,7 @@ will simply show all custom modules in the order they were defined.
 | ------------- | ------------------- | ---------------------------------------------------------------------------- |
 | `command`     |                     | The command whose output should be printed.                                  |
 | `when`        |                     | A shell command used as a condition to show the module. The module will be shown if the command returns a `0` status code. |
-| `shell`       |                     | The path to the shell to use to execute the command. If unset, it will fallback to STARSHIP_SHELL and then to "sh". |
+| `shell`       |                     | [See below](#custom-command-shell)                                           |
 | `description` | `"<custom module>"` | The description of the module that is shown when running `starship explain`. |
 | `files`       | `[]`                | The files that will be searched in the working directory for a match.        |
 | `directories` | `[]`                | The directories that will be searched in the working directory for a match.  |
@@ -1368,6 +1368,22 @@ will simply show all custom modules in the order they were defined.
 | `prefix`      | `""`                | Prefix to display immediately before the command output.                     |
 | `suffix`      | `""`                | Suffix to display immediately after the command output.                      |
 | `disabled`    | `false`             | Disables this `custom` module.                                               |
+
+#### Custom command shell
+
+`shell` accepts a non-empty list of strings, where:
+- The first string is the path to the shell to use to execute the command.
+- Other following arguments are passed to the shell.
+
+If unset, it will fallback to STARSHIP_SHELL and then to "sh" on Linux, and "cmd /C" on Windows.
+
+If `shell` is not given or only contains one element and Starship detects PowerShell will be used,
+the following arguments will automatically be added: `-NoProfile -Command -`.  
+This behavior can be avoided by explicitly passing arguments to the shell, e.g.
+
+```toml
+shell = ["pwsh", "-Command", "-"]
+```
 
 ### Example
 
--- a/src/modules/custom.rs
+++ b/src/modules/custom.rs
@@ -35,7 +35,7 @@ pub fn module<'a>(name: &'a str, context: &'a Context) -> Option<Module<'a>> {
 
     if !is_match {
         if let Some(when) = config.when {
-            is_match = exec_when(when, config.shell);
+            is_match = exec_when(when, &config.shell.0);
         }
 
         if !is_match {
@@ -57,7 +57,7 @@ pub fn module<'a>(name: &'a str, context: &'a Context) -> Option<Module<'a>> {
         module.create_segment("symbol", &symbol);
     }
 
-    if let Some(output) = exec_command(config.command, config.shell) {
+    if let Some(output) = exec_command(config.command, &config.shell.0) {
         let trimmed = output.trim();
 
         if trimmed.is_empty() {
@@ -77,26 +77,31 @@ pub fn module<'a>(name: &'a str, context: &'a Context) -> Option<Module<'a>> {
 
 /// Return the invoking shell, using `shell` and fallbacking in order to STARSHIP_SHELL and "sh"
 #[cfg(not(windows))]
-fn get_shell(shell: Option<&str>) -> std::borrow::Cow<str> {
-    if let Some(forced_shell) = shell {
-        forced_shell.into()
+fn get_shell<'a, 'b>(shell_args: &'b [&'a str]) -> (std::borrow::Cow<'a, str>, &'b [&'a str]) {
+    if !shell_args.is_empty() {
+        (shell_args[0].into(), &shell_args[1..])
     } else if let Ok(env_shell) = std::env::var("STARSHIP_SHELL") {
-        env_shell.into()
+        (env_shell.into(), &[] as &[&str])
     } else {
-        "sh".into()
+        ("sh".into(), &[] as &[&str])
     }
 }
 
 /// Attempt to run the given command in a shell by passing it as `stdin` to `get_shell()`
 #[cfg(not(windows))]
-fn shell_command(cmd: &str, shell: Option<&str>) -> Option<Output> {
-    let command = Command::new(get_shell(shell).as_ref())
+fn shell_command(cmd: &str, shell_args: &[&str]) -> Option<Output> {
+    let (shell, shell_args) = get_shell(shell_args);
+    let mut command = Command::new(shell.as_ref());
+
+    command
+        .args(shell_args)
         .stdin(Stdio::piped())
         .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .spawn();
+        .stderr(Stdio::piped());
 
-    let mut child = match command {
+    handle_powershell(&mut command, &shell, shell_args);
+
+    let mut child = match command.spawn() {
         Ok(command) => command,
         Err(_) => {
             log::debug!(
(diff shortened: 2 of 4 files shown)


src/modules/custom.rs
```diff
@@ -207,6 +214,19 @@ fn exec_command(cmd: &str, shell: Option<&str>) -> Option<String> {
...
+/// If the specified shell refers to PowerShell, adds the arguments "-Command -" to the
+/// given command.
+fn handle_powershell(command: &mut Command, shell: &str) {
+    let is_powershell = shell.ends_with("pwsh.exe")
+        || shell.ends_with("powershell.exe")
+        || shell.ends_with("pwsh")
+        || shell.ends_with("powershell");
+
+    if is_powershell {
+        command.arg("-Command").arg("-");
```
jimleroyer, line 226: What is the `-` argument for? I can't see this [through the documentation](https://docs.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Core/About/about_PowerShell_exe?view=powershell-5.1). Testing it locally also outputs an error and help message.

We might want the `-NoProfile` argument on the other hand. That would speed up execution of the custom command as the profile loading is skipped. Testing locally a `hello world` echo command with and without that argument shows substantial delay (when profile is loaded). 

```suggestion
        command.arg("-NoProfile").arg("-Command");
```

71: I am not sure about -NoProfile, since the user may have functions defined in their profile that they use in the prompt.

Omitting it _does_ slow down significantly the prompt. Not sure what the best course of action is.

As for `-`, it is actually part of the documentation, it's just subtle (`-Command - | ...`, see that first alternative `-`?). It means "read from stdin."

rcjeferson: Guys, and if we add a new param to add custom args? This attend all shells and allow the user customize which args he wants.

What do you think?

71: @rcjeferson That's kinda possible now, and we can extend it further by splitting `shell` into shell + args (which seems to be automatic on Windows?), but we still need a sane default for the time being.

rcjeferson: Yes, in this case we can assume without args as default and put `-NoProfile` as suggested value to this new param (args) in documentation. That way, who need the profile config get by default and who not need, can add `-NoProfile` arg.

I agree that in some cases I would like to use my alias configured in my profile, but in mostly cases isn't necessary, so it will be possible disable with this new param.

What do you think about this?

71: If we're going down that way, maybe `-NoProfile -Command -` should be the default, since I think users will more often run built-in commands. Then for "advanced" usages, the user may remove `-NoProfile` at the cost of speed of execution.

jimleroyer: Thanks @71 for the explanation around usage of `-`.

+1 for the `-NoProfile -Command -` as default, and making this a priority in the time being.

+1 for the possibility for overriding it. Could be done later if we can't squeeze into it right now?

71: I think the best thing to do here would be to change `shell` from accepting a string to accepting a list of strings, but that means breaking backward compatibility.

71: Commit updated for this, please take a look!

docs/config/README.md
```diff
@@ -1369,6 +1369,22 @@ will simply show all custom modules in the order they were defined.
...
+If unset, it will fallback to STARSHIP_SHELL and then to "sh" on Linux, and "cmd /C" on Windows.
+
+If `shell` is not given or only contains one element and Starship detects PowerShell will be used,
+the following arguments will automatically be added: `-NoProfile -Command -`.  
+This behavior can be avoided by explicitly passing arguments to the shell, e.g.
+
+```toml
+shell = ["pwsh", "-Command", "-"]
+```
+
```
jimleroyer, line 1387: I've opened a PR to update the documentation for the current problem, with the code in `master`:
https://github.com/starship/starship/pull/1242

If that PR can be closed quicker, I would be in favor of using your documentation section instead and close my PR.

That said, I'd suggest we add a :::warning section about the potential terminal freeze if the shell is misconfigured, inspired by what I have in my PR. Would you be in favor to add this in your PR?

src/modules/custom.rs
```diff
@@ -77,26 +77,31 @@ pub fn module<'a>(name: &'a str, context: &'a Context) -> Option<Module<'a>> {
...
+
+    command
+        .args(shell_args)
         .stdin(Stdio::piped())
         .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .spawn();
+        .stderr(Stdio::piped());
+
+    handle_powershell(&mut command, &shell, shell_args);
```
jimleroyer, line 102: Well, I wasn't aware [we could run PowerShell on Linux](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-7). That's pretty awesome, cool to see you got it cover here.

src/modules/custom.rs
```diff
@@ -120,23 +125,30 @@ fn shell_command(cmd: &str, shell: Option<&str>) -> Option<Output> {
...
-        Some(std::borrow::Cow::Borrowed(shell))
+fn shell_command(cmd: &str, shell_args: &[&str]) -> Option<Output> {
+    let (shell, shell_args) = if !shell_args.is_empty() {
+        (
+            Some(std::borrow::Cow::Borrowed(shell_args[0])),
+            &shell_args[1..],
+        )
     } else if let Ok(env_shell) = std::env::var("STARSHIP_SHELL") {
-        Some(std::borrow::Cow::Owned(env_shell))
+        (Some(std::borrow::Cow::Owned(env_shell)), shell_args)
```
jimleroyer, line 135: The code here makes me believe that it's possible to specify shell arguments without specifying a shell and relying on the `STARSHIP_SHELL` environment variable -- I think that's great as we can define custom parameters such as loading profile if desired. But from the first and previous condition, it seems that as soon as the `shell` configuration variable is set, the first parameters is assumed to be the shell. Hence we'll never have valid `shell_args` here. Is this on purpose?

I might be missing something.. can you show me an example of defining shell parameters without defining the shell?

src/modules/custom.rs
```diff
@@ -207,14 +219,27 @@ fn exec_command(cmd: &str, shell: Option<&str>) -> Option<String> {
...
 
+/// If the specified shell refers to PowerShell, adds the arguments "-Command -" to the
+/// given command.
+fn handle_powershell(command: &mut Command, shell: &str, shell_args: &[&str]) {
+    let is_powershell = shell.ends_with("pwsh.exe")
+        || shell.ends_with("powershell.exe")
+        || shell.ends_with("pwsh")
+        || shell.ends_with("powershell");
+
+    if is_powershell && shell_args.is_empty() {
```
jimleroyer, line 230: 🎉

src/modules/custom.rs
jimleroyer: Do we have some guidelines on breaking backward compatibility? cc @matchai 

If only @rcjeferson used that feature, I guess we're good to go but I'm wondering if more people uses this. 😅

matchai: We don't quite yet.
https://github.com/starship/starship/issues/1075 would ideally give us the tools necessary to surface broken configuration, and link users to docs with necessary steps, but that's not yet in place. 😰

That said, I would assume that Powershell + custom modules is a rare combination and is worth pushing a fix for, noting the breaking change in our changelog.

src/modules/custom.rs
71: I did it to trick the type checker: in these branches `shell_args` are empty, but they have the right lifetime to be returned. After some digging, `&[]` has a static lifetime so that works too. Pushing this change for clarity.

src/modules/custom.rs
71: Yep, and that's pretty new!

src/modules/custom.rs
71: Another possibility I had in mind was to allow both a string (that would be internally converted to a vector) and a list, but that requires a new implementation of `ModuleConfig` (e.g. `ModuleConfig` for an enum `VectorOrString`, or `ModuleConfig` for a new type `VecOrString<'a>(Vec<&'a str>)` that can handle both strings and lists in TOML).

docs/config/README.md
71: I think your PR is a nice addition to mine, so we should keep both?

71: Although there will be some conflicts.

src/modules/custom.rs
matchai: I think that sounds like a worthwhile addition 👍

docs/config/README.md
jimleroyer: Yes that sounds good! I can handle the conflict, that's okay. I'm just wondering of the ordering of our PRs. We go with yours first and then we accommodate mine?

src/modules/custom.rs
jimleroyer: Okay, so it's more of a pragmatic reason.

I assume then that we'll have to override the shell by defining it as a first parameter to also provide the arguments? That make sense, thinking about it again. The user couldn't be 100% sure which shell that Starship will choose and overriding the shell will lead to the path of least surprises. 

Thank you!

docs/config/README.md
71: I think so, yes. The proposed fix in the warning (using `shell = [...]`) is currently not possible because right now `shell` only accepts a string, so it would have to be slightly changed before and after merging my PR.

PR #1224, opened 2020-05-16
refactor(crystal): Use format strings

ATiltedTree:
#### Description

This PR deprecates `symbol` and `style` keys in the `crystal` module to replace it with the `format` key instead.

#### Motivation and Context

Related to issue #1057

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Screenshots (if appropriate):
image

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #1251, opened 2020-05-24
refactor(zig): Use format strings

filalex77:
#### Description

This PR deprecates `symbol` and `style` keys in the `zig` module to replace it with the `format` key instead. Heavily based on #1224.

#### Motivation and Context

Related to issue #1057.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/configs/zig.rs
+++ b/src/configs/zig.rs
@@ -13,7 +13,7 @@ pub struct ZigConfig<'a> {
 impl<'a> RootModuleConfig<'a> for ZigConfig<'a> {
     fn new() -> Self {
         ZigConfig {
-            format: "via [$symbol$version]($style)",
+            format: "via [$symbol$version]($style) ",
             symbol: "↯ ",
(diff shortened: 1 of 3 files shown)


src/configs/zig.rs
```diff
@@ -1,22 +1,21 @@
...
     pub disabled: bool,
 }
 
 impl<'a> RootModuleConfig<'a> for ZigConfig<'a> {
     fn new() -> Self {
         ZigConfig {
-            symbol: SegmentConfig::new("↯ "),
-            version: SegmentConfig::default(),
-            style: Color::Yellow.bold(),
+            format: "via [$symbol$version]($style)",
```
andytom, line 16: I think you need to add a trailing space here to match the output of the old module and the updated config docs.
This should also fix the tests.

```suggestion
            format: "via [$symbol$version]($style) ",
```

andytom:
@filalex77 Thank you for the contribution.

PR #1254, opened 2020-05-24
chore: create stale.yml

sirMerr:
#### Description

Looking into adding [Stale](https://probot.github.io/apps/stale/) to Starship. This PR adds the configuration necessary to turn it on. Keep in mind that this will send a lot of notifications once turned on due to the number of issues we have.

The configuration is still uncertain so please comment with your suggestions :).

#### Motivation and Context

With the refactoring work going on, it would be great to freshen up our issues. It would also be good to have reminders for the maintainers. 

#### Checklist:


sirMerr:
Closing this in favor of the stale action

PR #1278, opened 2020-05-31
build(deps): disabled logs in release

anouarkappitou:
#### Description
Setting RUST_LOG prints debug messages to the console, in startship release binary mode.

#### Motivation and Context
Related to issue https://github.com/starship/starship/issues/1277


andytom:
Hi @anouarkappitou, while I understand that the current behaviour might be frustrating for rust developers we can not disable logging in the release builds as it would make it impossible for us to debug a lot of the issues that are reported. I'm going to close this PR and will add some additional suggestions on the related issue.

PR #1297, opened 2020-06-06
feat(python): Add option to change the python binary

andytom:
#### Description

We are going to start to have problems with the python binary as python2
is removed and replaced with python3. To make the transition easier I
have added an option to the python module to allow the user to pick a
particular binary, e.g `python3`, for the module to use when selecting
the version of python. I have also refactored the python tests moving
almost all of them into the module and removing the dependency on the
version of python that is installed on the system.

#### Motivation and Context

Related to #1287, also provides a workaround for #632.

#### Checklist:


docs/config/README.md
```diff
@@ -1188,14 +1188,15 @@ The module will be shown if any of the following conditions are met:
...
-| `scan_for_pyfiles`   | `true`          | If false, Python files in the current directory will not show this module.  |
-| `style`              | `"bold yellow"` | The style for the module.                                                   |
-| `disabled`           | `false`         | Disables the `python` module.                                               |
+| Variable             | Default         | Description                                                                   |
+| -------------------- | --------------- | ----------------------------------------------------------------------------- |
+| `symbol`             | `"🐍 "`         | The symbol used before displaying the version of Python.                      |
+| `pyenv_version_name` | `false`         | Use pyenv to get Python version                                               |
+| `pyenv_prefix`       | `"pyenv "`      | Prefix before pyenv version display (default display is `pyenv MY_VERSION`)   |
+| `python_binary`      | `python`        | Use this to change the python binary if, for example, you have an alias setup |
```
andytom, line 1196: I have just added this as an option here but should this go in an advanced configuration section?

matchai: I would be in favor of having it in Advanced Configuration.

andytom: Have added an Advanced Configuration section, (sorry for the force pushing but it's only the new commit that got forced)

PR #1298, opened 2020-06-06
feat(directory): Show lock symbol if current directory is read only

### LINKED ISSUE #1041: add lock symbol if directory is write-protected
## Feature Request
**Is your feature request related to a problem? Please describe.**

The `directory` module could have an option to show a "padlock" if the current directory is write-protected. Like [this](https://github.com/denysdovhan/spaceship-prompt/blob/master/docs/Options.md#directory-dir).

Aelnor:
The directory module shows little 'lock' symbol when the current directory can't be written by the current user.

#### Description
It was requested to Starship have the similar feature as zsh does -- showing a special symbol whenever the current directory can't be written by the current user.
The *nix implementation is straightforward: it uses a built-in sh feature to determine the matter. Windows implementation is complex involving Windows Security API to perform an `AccessCheck`.

#### Motivation and Context

Closes #1041

Annotation 2020-06-06 194759
Screenshot from 2020-06-06 20-12-35

#### Checklist:

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -95,6 +95,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         String::from("")
     };
     let final_dir_string = format!("{}{}", fish_prefix, truncated_dir_string);
+    let lock_symbol = String::from(config.read_only_symbol);
 
     let parsed = StringFormatter::new(config.format).and_then(|formatter| {
         formatter
@@ -107,7 +108,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "path" => Some(Ok(&final_dir_string)),
                 "read_only" => {
                     if is_readonly_dir(current_dir.to_str()?) {
-                        Some(Ok(config.read_only_symbol))
+                        Some(Ok(&lock_symbol))
                     } else {
                         None
                     }
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -1,3 +1,7 @@
+#[cfg(not(target_os = "windows"))]
+use super::utils::directory_nix as directory_utils;
+#[cfg(target_os = "windows")]
+use super::utils::directory_win as directory_utils;
 use path_slash::PathExt;
 use std::collections::HashMap;
 use std::iter::FromIterator;
@@ -96,10 +100,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         formatter
             .map_style(|variable| match variable {
                 "style" => Some(Ok(config.style)),
+                "read_only_style" => Some(Ok(config.read_only_symbol_style)),
                 _ => None,
             })
             .map(|variable| match variable {
                 "path" => Some(Ok(&final_dir_string)),
+                "read_only" => {
+                    if is_readonly_dir(current_dir.to_str()?) {
+                        Some(Ok(config.read_only_symbol))
+                    } else {
+                        None
+                    }
+                }
                 _ => None,
             })
             .parse(None)
@@ -114,6 +126,20 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     });
 
     Some(module)
+}
+
+fn is_readonly_dir(path: &str) -> bool {
+    match directory_utils::is_write_allowed(path) {
+        Ok(res) => !res,
+        Err(e) => {
+            log::debug!(
+                "Failed to detemine read only status of directory '{}': {}",
+                path,
+                e
+            );
+            false
+        }
+    }
 }
 
 /// Contract the root component of a path
--- a/src/modules/utils/directory_nix.rs
+++ b/src/modules/utils/directory_nix.rs
@@ -2,7 +2,7 @@ use nix::sys::stat::Mode;
 use nix::unistd;
 use nix::unistd::{Gid, Uid};
 use std::fs;
-use std::os::linux::fs::MetadataExt;
+use std::os::unix::fs::MetadataExt;
 use std::os::unix::fs::PermissionsExt;
 
 const USER_WRITE_BIT: u32 = Mode::S_IWUSR.bits();
@@ -17,10 +17,10 @@ pub fn is_write_allowed(folder_path: &str) -> Result<bool, &'static str> {
     if euid.is_root() {
         return Ok(true);
     }
-    if meta.st_uid() == euid.as_raw() {
+    if meta.uid() == euid.as_raw() {
         Ok(perms & USER_WRITE_BIT != 0)
-    } else if (meta.st_gid() == Gid::effective().as_raw())
-        || (get_supplementary_groups().contains(&meta.st_gid()))
+    } else if (meta.gid() == Gid::effective().as_raw())
+        || (get_supplementary_groups().contains(&meta.gid()))
     {
         Ok(perms & GROUP_WRITE_BIT != 0)
     } else {
--- a/src/modules/utils/directory_nix.rs
+++ b/src/modules/utils/directory_nix.rs
@@ -1,0 +1,60 @@
+extern crate libc;
+#[cfg(all(unix, not(target_os = "macos")))]
+use libc::gid_t;
+use std::ffi::CString;
+
+pub fn is_write_allowed(folder_path: &str) -> std::result::Result<bool, &'static str> {
+    let c_string = CString::new(folder_path).unwrap();
+    unsafe {
+        let mut buf: Vec<u8> = Vec::with_capacity(std::mem::size_of::<libc::stat>());
+        let res = libc::stat(c_string.as_ptr(), buf.as_mut_ptr() as *mut libc::stat);
+
+        if res != 0 {
+            return Err("Unable to stat() directory");
+        }
+
+        let stat_struct: *const libc::stat = buf.as_ptr() as *const libc::stat;
+        let mode = (*stat_struct).st_mode;
+        if (*stat_struct).st_uid == libc::geteuid() {
+            return Ok(mode & libc::S_IWUSR != 0);
+        }
+        if (*stat_struct).st_gid == libc::getgid() {
+            return Ok(mode & libc::S_IWGRP != 0);
+        }
+
+        let num_groups = libc::getgroups(0, ::std::ptr::null_mut());
+        if num_groups == -1 {
+            return Err("Unable to get suplementary groups for the current process");
+        }
+        #[cfg(all(unix, target_os = "macos"))]
+        let mut groups: Vec<u32> = vec![0; 1024];
+        #[cfg(all(unix, not(target_os = "macos")))]
+        let mut groups: Vec<gid_t> = vec![0; 1024];
+        let res = libc::getgroups(num_groups, groups.as_mut_ptr());
+        if res == -1 {
+            return Err("Unable to get suplementary groups for the current process");
+        }
+
+        for i in 0..num_groups {
+            if groups[i as usize] == (*stat_struct).st_gid {
+                return Ok(mode & libc::S_IWGRP != 0);
+            }
+        }
+
+        Ok(mode & libc::S_IWOTH != 0)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn read_only_test() {
+        assert_eq!(is_write_allowed("/etc"), Ok(false));
+        assert_eq!(
+            is_write_allowed("/i_dont_exist"),
+            Err("Unable to stat() directory")
+        );
+    }
+}
--- a/src/modules/utils/directory_nix.rs
+++ b/src/modules/utils/directory_nix.rs
@@ -1,47 +1,45 @@
-extern crate libc;
-#[cfg(all(unix, not(target_os = "macos")))]
-use libc::gid_t;
-use std::ffi::CString;
+use nix::sys::stat::Mode;
+use nix::unistd;
+use nix::unistd::{Gid, Uid};
+use std::fs;
+use std::os::linux::fs::MetadataExt;
+use std::os::unix::fs::PermissionsExt;
 
-pub fn is_write_allowed(folder_path: &str) -> std::result::Result<bool, &'static str> {
-    let c_string = CString::new(folder_path).unwrap();
-    unsafe {
-        let mut stat: libc::stat = std::mem::zeroed();
-        let res = libc::stat(c_string.as_ptr(), &mut stat);
+const USER_WRITE_BIT: u32 = Mode::S_IWUSR.bits();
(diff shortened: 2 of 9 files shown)


Aelnor:
I'll take a look at the broken tests.

src/modules/directory.rs
```diff
@@ -104,11 +108,48 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
     module.get_prefix().set_value(config.prefix);
 
     Some(module)
 }
 
+#[cfg(not(target_os = "windows"))]
+fn is_readonly_dir(path: &str) -> bool {
+    let child = Command::new("sh")
+        .arg("-c")
+        .arg(format!("if [ ! -w {} ]; then exit 1; fi", path))
```
Scotsguy, line 129: Calling out to `sh` once per prompt redraw is super expensive, performance-wise.
Why do this manually on windows but shell out on *nix?

Aelnor: > Calling out to `sh` once per prompt redraw is super expensive, performance-wise.
> Why do this manually on windows but shell out on *nix?

In my opinion Starship does a lot of really expensive stuff on every redraw (for example, calling go binary or examining git status). An extra `sh` call is not expensive at all in comparison.

But I can rework it for sure if that's the consensus.

Scotsguy: It is not as expensive as, say, examining git status for a big repo, but it gets called indiscriminately on each directory, that's why I'm more worried about it than otherwise

Aelnor: Well, I do agree. I'll fix this, thanks

Aelnor:
At the moment this implementation has several Linux issues, I'm digging into

Aelnor:
Okay. Linux is fixed, will fix macOS shortly (tomorrow probably)

chipbuster:
@Aelnor Sorry about my lateness to this party.

I just checked the source diffs, and I think it's highly unlikely we'll be merge the changeset as it currently stands.

Starship currently uses a grand total of zero lines of unsafe code. As far as I know, adding `unsafe` to the codebase has only been proposed a few times, and each time we were able to get around it by using safe abstractions or other crates. This PR introduces a whole bunch of `unsafe` code into the codebase that we're going to have to continually audit.

Is there not some crate that could simplify this procedure? Doesn't `std::fs` provide something for this? I'm pretty sure `PermissionsExt::mode()` already does at least what some of the code here does.

ATiltedTree:
I have looked into this also a little:
There is [`std::fs::metadata`](https://doc.rust-lang.org/stable/std/fs/fn.metadata.html).
You give it any path and it return the metadata of that path.
You then can call [`Metadata.permissions()`](https://doc.rust-lang.org/stable/std/fs/struct.Metadata.html#method.permissions) which returns you the [permissions of that path](https://doc.rust-lang.org/stable/std/fs/struct.Permissions.html).
And finally you can call `.readonly()`.
And the docs state that it should work on all platforms:
> This module only currently provides one bit of information, readonly, which is exposed on all currently supported platforms.

Hopefully that helps.

Aelnor:
> And the docs state that it should work on all platforms:



> 
> 
> I have looked into this also a little:
> There is [`std::fs::metadata`](https://doc.rust-lang.org/stable/std/fs/fn.metadata.html).
> You give it any path and it return the metadata of that path.
> You then can call [`Metadata.permissions()`](https://doc.rust-lang.org/stable/std/fs/struct.Metadata.html#method.permissions) which returns you the [permissions of that path](https://doc.rust-lang.org/stable/std/fs/struct.Permissions.html).
> And finally you can call `.readonly()`.
> And the docs state that it should work on all platforms:
> 
> > This module only currently provides one bit of information, readonly, which is exposed on all currently supported platforms.
> 
> Hopefully that helps.

On Windows, it uses `GetFileAttributes`, which is not a reliable way to draw conclusions. I haven't played with it on *nix yet though, but I'm very sceptic about `std::file::permissions`. The linked ticket has additional information, feel free to take a look.

Aelnor:
> 
> 
> @Aelnor Sorry about my lateness to this party.
> 
> I just checked the source diffs, and I think it's highly unlikely we'll be merge the changeset as it currently stands.
> 
> Starship currently uses a grand total of zero lines of unsafe code. As far as I know, adding `unsafe` to the codebase has only been proposed a few times, and each time we were able to get around it by using safe abstractions or other crates. This PR introduces a whole bunch of `unsafe` code into the codebase that we're going to have to continually audit.
> 
> Is there not some crate that could simplify this procedure? Doesn't `std::fs` provide something for this? I'm pretty sure `PermissionsExt::mode()` already does at least what some of the code here does.

`mode` is one piece of the puzzle and to understand if dir is read-only we need to have a lot more: effective uid, gid, directory owner, directory group, other groups current user belongs to, etc.
I believe, unsafe is a tool provided by the language for a reason. One doesn't really have to 'continuously audit' the code I'm proposing since it is very simple, straightforward and one dimensional. The only reason one would like to change it is when something is wrong with it.
To be honest, I dislike the idea of searching for other crates because that leads to having a lot of dependencies which are maintained by other people in order to have very simple and straightforward 150 lines of code, 15 of which are unsafe. And still, these other crates will use unsafe code in order to do the same.
But if you insist, I'm going to dig a little bit to try to find something suitable.

Scotsguy:
https://github.com/Scotsguy/starship/commit/1adc31b855256f61438a87985e95bdd4c1be2f84
No unsafe on the nix version, not sure what to do about windows

Aelnor:
> 
> 
> [Scotsguy@f209894](https://github.com/Scotsguy/starship/commit/f20989440ffe5727f4f7def5a672bb401786fc70)
> No unsafe on the nix version, not sure what to do about windows

Thanks
I believe this code fails when suplementary groups provide access to the directory. Have you checked it?

Scotsguy:
Annoyingly, the function to get supplementary groups isn't available on macOS
https://docs.rs/nix/0.17.0/nix/unistd/fn.getgroups.html

Scotsguy:
Please don't squash your commits, it's impossible to tell what changed between the individual commits

Aelnor:
> Please don't squash your commits, it's impossible to tell what changed between the individual commits

At the moment it does not matter much. I've fixed the macOS version and a part of clippy warnings. 
But until we are ok with using unsafes the whole PR is not worth analyzing.

chipbuster:
> I believe, unsafe is a tool provided by the language for a reason. 

Agreed.

> One doesn't really have to 'continuously audit' the code I'm proposing since it is very simple, straightforward and one dimensional. The only reason one would like to change it is when something is wrong with it.

Perhaps to someone familiar with the necessary contracts and APIs, this is true. I can mostly make out what the UNIX one is doing, but I personally have no clue why the calls in directory_win are necessary, or how they could potentially fail.

> To be honest, I dislike the idea of searching for other crates because that leads to having a lot of dependencies which are maintained by other people in order to have very simple and straightforward 150 lines of code, 15 of which are unsafe. And still, these other crates will use unsafe code in order to do the same.

Perhaps I'm misreading this code, but it looks like essentially all of the unix function and most of the Windows one is unsafe at the moment, for a grand total of about 140 lines of unsafe. Are you saying that this code essentially boils down to just a few lines of unsafe?

Aelnor:
> I'm not hard set against this, I'm just a little wary of dropping 140 lines of unsafe code into our codebase when I don't fully understand why it's needed. If it turns out that there is no other better option, I'll take it, but as @Scotsguy points out, there are at least a few crates that provide some safe abstractions over libc (whether any of them completely satisfy our needs is a different story).

As I can see at the moment (thanks, @Scotsguy), we can have decent and safe code for Linux which in my opinion completely satisfies our needs. We can have decent safe code for macOS which works correctly in the majority of the cases (ignoring ACL and other mac specific stuff). This is good news.

At the moment I have no idea what to do with the Windows version. I keep digging at the moment, but no luck so far.
We can either leave it as is (unsafe) or completely turn the feature off on Windows if we'd like to keep everything safe in Starship unless we're able to find something covering WinAPI.

chipbuster:
Alright, if we can't find anything else, let's restrict the scope of `unsafe` as much as possible (so that future code readers can tell exactly which blocks are unsafe) and throw it in then.

Aelnor:
*nix is safe now, minor Windows refactoring incoming. 
By the way, what do you think about how it looks like? I'm not sure I'm liking the *nix version, but have no idea how to improve it.

src/modules/utils/directory_nix.rs
```diff
@@ -2,7 +2,7 @@ use nix::sys::stat::Mode;
 use nix::unistd;
 use nix::unistd::{Gid, Uid};
 use std::fs;
-use std::os::linux::fs::MetadataExt;
+use std::os::unix::fs::MetadataExt;
 use std::os::unix::fs::PermissionsExt;
 
 const USER_WRITE_BIT: u32 = Mode::S_IWUSR.bits();
```
Scotsguy, line 8: actual type: nix::sys::stat::mode_t

Scotsguy: nevermind, too late :P

Aelnor:
Would you like me at this point to squash all the commits?

chipbuster:
@Aelnor We usually do the squash from the GitHub end when merging in the PR. If you'd like to do it for cleanliness, feel free to, but I likely won't perceive much of a difference since I do most reviews in the web interface.

Aelnor:
I've fixed the conflict that came up recently.
What do you think about current implementation?

src/modules/utils/directory_nix.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+#[cfg(all(unix, target_os = "macos"))]
+fn get_supplementary_groups() -> Vec<u32> {
+    // at the moment nix crate does not provide it for macOS
+    Vec::new()
+}
+
+#[cfg(test)]
```
chipbuster, line 40: I think this test needs to be `#[ignore]`d, as it assumes the existence of `/etc`. [This is not a valid assumption on Termux](https://wiki.termux.com/wiki/Differences_from_Linux) which is a platform we support.

src/modules/utils/directory_nix.rs
```diff
@@ -0,0 +1,52 @@
+use nix::sys::stat::Mode;
+use nix::unistd::{Gid, Uid};
+use std::fs;
+use std::os::unix::fs::MetadataExt;
+use std::os::unix::fs::PermissionsExt;
+
+pub fn is_write_allowed(folder_path: &str) -> Result<bool, &'static str> {
```
chipbuster, line 7: This looks great! I wouldn't mind a doc comment just generally explaining what this function is looking for (e.g. checks permission bits against current user's effective uid + list of groups) but otherwise I think it's ready to merge.

src/modules/utils/directory_win.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub fn is_write_allowed(folder_path: &str) -> std::result::Result<bool, &'static str> {
+    let folder_name: Vec<u16> = OsStr::new(folder_path)
+        .encode_wide()
+        .chain(iter::once(0))
+        .collect();
+    let mut length: DWORD = 0;
+
+    let rc = unsafe {
+        securitybaseapi::GetFileSecurityW(
```
chipbuster, line 27: Alright, we're about to enter the area of the code I know nothing about, so forgive me if I sound like a total rube here. I just want to understand things well enough to where I could (at least try to) modify this if I need to in the future.

First, what is the difference between `GetFileSecurityA` (which I can find documentation for) and `GetFileSecurityW`?

Second, am I correct in saying that this call is only getting information about `length` so that we can allocate a buffer of the correct size?

Finally (and perhaps the gaps in my knowledge of `unsafe` are showing here), is there an implicit coercion that occurs from `&mut` to `*mut`? The crate docs say that you need an `LPDWORD` (i.e. `*mut DWORD`) for the last argument.

src/modules/utils/directory_win.rs
```diff
@@ -0,0 +1,113 @@
...
+            length,
+            &mut length,
+        )
+    };
+
+    if rc != 1 {
+        return Err("GetFileSecurityW failed to retrieve the security descriptor");
+    }
+
+    let mut token: HANDLE = 0 as HANDLE;
```
chipbuster, line 57: Clearly, we need to create an impersonation token here, but I'm having trouble understanding why. Is this because `AccessCheck` is a much more powerful API than just checking for file permissions, and so needs an impersonation token for other uses of it? Is this something that you just have to do when you want to call `AccessCheck`?

src/modules/utils/directory_nix.rs
chipbuster: Of course, looking further down I see that our integration tests directly assume the existence of `/etc`...which I also think is in error, but now I'm not as sure.

src/modules/utils/directory_win.rs
Scotsguy: A suffix = ANSI string, W suffix = Unicode String
https://docs.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings

src/modules/utils/directory_win.rs
jeanga: AccessCheck requires an impersonation token (per https://docs.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-accesscheck).
You can find some sample code on how to do call DuplicateToken here: 
https://github.com/jeanga/wil-rs/blob/abc43b2d3ec45896ebf2f3d19b380e7bfed085e6/src/token.rs#L139

HTH

src/modules/utils/directory_nix.rs
Aelnor: Done

src/modules/utils/directory_nix.rs
Aelnor: Done

src/modules/utils/directory_win.rs
Aelnor: The second question: yes, one can use type coercion to get a raw pointer from a reference in Rust. More info here: https://doc.rust-lang.org/std/primitive.pointer.html

Aelnor:
Rebased for the latest changes.

chipbuster:
Okay, I think we're good to go here!

The only last thing I'd be worried about is that checking this info might be slow on a network-mounted drive, but I'm going to file that under YAGNI for now. If someone has an issue where this causes things to be slower-than-usual, they can file an issue and we'll add a timeout.

Aelnor:
> Okay, I think we're good to go here!
> 
> The only last thing I'd be worried about is that checking this info might be slow on a network-mounted drive, but I'm going to file that under YAGNI for now. If someone has an issue where this causes things to be slower-than-usual, they can file an issue and we'll add a timeout.

The current design is: if the format string in directory modue has no {read_only}, it isn't calculated. So at the very least we have a workaroud

chipbuster:
@Aelnor Thanks so much for your work on this PR!

PR #1299, opened 2020-06-07
fix(directory): Fix path contractions for symlinked git repos

foriequal0:
#### Description
This PR fixes path contractions
1. when `full_path` contains symlinks (a logical path), and `top_level_path` is given as a physical path.
2. and when git repository's subdirectory is symlinked to outside of the repo tree.

##### Fix 1
`&repo.root` returns physical path, not the logical path.
`git rev-parse --show-toplevel` returns worktree [path](https://github.com/git/git/blob/v2.27.0/builtin/rev-parse.c#L803), which is [initialized](https://github.com/git/git/blob/v2.27.0/environment.c#L261) with [`strbuf_realpath`](https://github.com/git/git/blob/v2.27.0/abspath.c#L81).
So, it returns a physical path of a repository (there isn't a symlink in its ancestors). I think libgit2 does too so `&repo.root` does.
However, `PWD` and `std::env::current_dir` gives you a logical path (it might contain symlinks in its ancestors).

If I have a following structure of repositories, `starship` mispredict that I'm not inside of the repository since `current_dir` doesn't contain `&repo.root`.
```sh
~/ws> ls -l
... repo
... repo2 -> repo

~/ws> cd ~/ws/repo
repo> echo $(pwd)  $(realpath pwd) $(git rev-parse --show-toplevel)
/home/user/ws/repo /home/user/ws/repo

repo> cd ~/ws/repo2
/home/user/ws/repo2> # note that the prompt is not repo2
/home/user/ws/repo2> echo $(pwd)  $(realpath pwd) $(git rev-parse --show-toplevel)
/home/user/ws/repo2 /home/user/ws/repo /home/user/ws/repo
```

In `repo2`, `$(pwd)` doesn't start with $(git rev-parse --show-toplevel), so the prompt is shown as `/home/user/ws/repo2`, not `repo2`.

So I made a function `realpath` after git's `strbuf_realpath` (they should do the same thing), and `contract_repo_path` walks through ancestors of `full_path` and whether their physical paths are matches with `top_level_path`. Then it trims the matching part.
As I've commented, this will properly handle following case.
```
weird_repo> ls -l src
... src -> ../src
weird_repo> cd src/src/src/src
weird_repo/src/src/src/src>
```
If I'd just applied `strip_prefix`, then it would've been:
```
weird_repo> ls -l src
... src -> ../src
weird_repo> cd src/src/src/src
weird_repo/src>
```

##### Fix 2
When you have a repository at `~/ws/repo`, and you made a symlink `ln -s ~/ws/repo/some/module ~/tmp/module`.
In `~/tmp/module`, You can see that `git rev-parse --show-toplevel` returns `/home/user/ws/repos`.
However, the current version also fails to contract path. It prompts `/home/user/tmp/module> ` again.
It is weird to contract it as `repo/module>`. So when the path is out of the repo tree, it just contracted to `HOME`.

I've made a repository, and its symlinks like the above scenarios (`ln -s` on Linux, `mklink` on Windows).
- [x] I have tested using **MacOS** (Tested by @nickwb)
- [x] I have tested using **Linux**: I've tested it on WSL1, and Ubuntu 20.04.
- [x] I have tested using **Windows**: I just checked that `git rev-parse --show-toplevel` (and `&repo.root`) returns a logical path, instead of a physical path (Weird 1). Also, It is already failed to recognize a repo root in the scenario of Fix 2 (However, `git rev-parse --show-toplevel` does it well. It might be a libgit2's bug. Weird 2). So there shouldn't be a difference after this PR.

#### Checklist:
  It is a bug fix. I think updating the documentation isn't needed.
  ~It involves `read_link` which requires a real filesystem. Existing tests are pure functions and don't need a real filesystem. I'm not sure how do I add tests.~ Added tests

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -62,6 +62,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let dir_string = match &repo.root {
         Some(repo_root) if config.truncate_to_repo && (repo_root != &home_dir) => {
+            log::debug!("Repo root: {:?}", repo_root);
             let repo_folder_name = repo_root.file_name().unwrap().to_str().unwrap();
 
             // Contract the path to the git repo root
@@ -70,6 +71,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         // Contract the path to the home directory
         _ => contract_path(current_dir, &home_dir, HOME_SYMBOL),
     };
+    log::debug!("Dir string: {}", dir_string);
 
     let substituted_dir = substitute_path(dir_string, &config.substitutions);
 
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -152,10 +152,7 @@ fn contract_repo_path(full_path: &Path, top_level_path: &Path) -> Option<String>
         }
 
         let components: Vec<_> = full_path.components().collect();
-        let repo_name = components[components.len() - i - 1]
-            .as_os_str()
-            .to_str()
-            .unwrap();
+        let repo_name = components[components.len() - i - 1].as_os_str().to_str()?;
 
         if i == 0 {
             return Some(repo_name.to_string());
@@ -166,7 +163,7 @@ fn contract_repo_path(full_path: &Path, top_level_path: &Path) -> Option<String>
             "{repo_name}{separator}{path}",
             repo_name = repo_name,
             separator = "/",
-            path = path.to_slash().unwrap()
+            path = path.to_slash()?
         ));
     }
     None
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -1,5 +1,6 @@
 use path_slash::PathExt;
 use std::collections::HashMap;
+use std::iter::FromIterator;
 use std::path::{Path, PathBuf};
 use unicode_segmentation::UnicodeSegmentation;
 
@@ -63,10 +64,9 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let dir_string = match &repo.root {
         Some(repo_root) if config.truncate_to_repo && (repo_root != &home_dir) => {
             log::debug!("Repo root: {:?}", repo_root);
-            let repo_folder_name = repo_root.file_name().unwrap().to_str().unwrap();
-
             // Contract the path to the git repo root
-            contract_path(current_dir, repo_root, repo_folder_name)
+            contract_repo_path(current_dir, repo_root)
+                .unwrap_or_else(|| current_dir.to_slash().unwrap())
         }
         // Contract the path to the home directory
         _ => contract_path(current_dir, &home_dir, HOME_SYMBOL),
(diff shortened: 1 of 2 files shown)


andytom:
Hi @foriequal0, thank you for the contribution, you can add tests that require a real fs to [`tests/testsuite/directory.rs`](https://github.com/starship/starship/blob/master/tests/testsuite/directory.rs).

nickwb:
I built this, and started testing on MacOS. I got one scenario to work as expected, but I found another scenario which did not.

Would you expect this to work?

Repo: `~/my_repo/`
Symlink: `~/my_tests` -> `~/my_repo/tests`
Expected Result: Prompt shows `myrepo/tests`
Actual Result: Prompt shows `~/my_tests`

Relevant logs:

```
 DEBUG starship::modules::directory > Current directory: "/Users/nickwb/my_tests"
 DEBUG starship::modules::directory > Repo root: "/Users/nickwb/my_repo/"
 DEBUG starship::modules::directory > Dir string: ~/my_test
```

foriequal0:
@nickwb I deliberately choose that behavior. Whenever I command `cd ..` when I see `some/dir` on prompt, I expect that I should be in `some/`. It is less confusing and consistent with the prompt without truncation at repo.

foriequal0:
How about prompt both? Instead of choose one, show both results would not be a completely irrational choice, like
```
~/my_tests (which is myrepo/tests) on master
> 
```

nickwb:
> @nickwb I deliberately choose that behavior. Whenever I command `cd ..` when I see `some/dir` on prompt, I expect that I should be in `some/`. It is less confusing and consistent with the prompt without truncation at repo.

Ah - ok then! Sorry, I mis-read your original post.

> How about prompt both? Instead of choose one, show both results would not be a completely irrational choice...

This would definitely be clear, but it might make the prompt longer than desired. I'm Ok with your original choice. Maybe there is another whole module we could consider writing (e.g., `reveal_symlink`)

Happy to ✔️  for MacOS. The contraction though a symlink cycle, or direct to the repo root, behaves as expected.

foriequal0:
I'll bring some test cases as @andytom noted.

src/modules/directory.rs
```diff
@@ -134,6 +136,60 @@ fn contract_path(full_path: &Path, top_level_path: &Path, top_level_replacement:
...
+        let repo_name = components[components.len() - i - 1]
+            .as_os_str()
+            .to_str()
+            .unwrap();
+
+        if i == 0 {
+            return Some(repo_name.to_string());
+        }
+
+        let path = PathBuf::from_iter(&components[components.len() - i..]);
```
nickwb, line 164: `PathBuf::from_str` should work here? One less memory allocation.

src/modules/directory.rs
```diff
@@ -134,6 +136,60 @@ fn contract_path(full_path: &Path, top_level_path: &Path, top_level_replacement:
...
+        let ancestor_real_path = real_path(ancestor);
+        if ancestor_real_path != top_level_real_path {
+            continue;
+        }
+
+        let components: Vec<_> = full_path.components().collect();
+        let repo_name = components[components.len() - i - 1]
+            .as_os_str()
+            .to_str()
+            .unwrap();
```
nickwb, line 158: Should we panic here? Maybe `.to_str()?` is better?

src/modules/directory.rs
```diff
@@ -134,6 +136,60 @@ fn contract_path(full_path: &Path, top_level_path: &Path, top_level_replacement:
...
+        if i == 0 {
+            return Some(repo_name.to_string());
+        }
+
+        let path = PathBuf::from_iter(&components[components.len() - i..]);
+        return Some(format!(
+            "{repo_name}{separator}{path}",
+            repo_name = repo_name,
+            separator = "/",
+            path = path.to_slash().unwrap()
```
nickwb, line 169: Can we also avoid this panic?

src/modules/directory.rs
foriequal0: I think there is no way to get sub-`&Path` or `&str` or `&OsStr` from a `full_path.components().skip(N)...`. Similar situation to that `"some string".chars().skip(5)` cannot get &"string"

chipbuster:
@foriequal0 Any reason you're force-pushing these commits?

foriequal0:
@chipbuster Sorry. That's my bad habit. I initially just appended commits but after that, I force-pushed fixes for typos and `cargo fmt` warnings of the commits. Reviewed commits are not touched though.

chipbuster:
No worries! Was just curious. Force-pushing sometimes causes weird things to happen to the GitHub review interface, but it's not *too* big a headache to work around it.

src/modules/directory.rs
nickwb: Ah, sorry @foriequal0 - I mistakenly thought that `components` was a string slice.

Your existing implementation is fine 👍

nickwb:
Just looking at your new tests and noticed the `#[cfg(not(target_os = "windows"))]` - I'm wondering if this has been tested on NTFS with directory junctions?

It might even be possible to run the tests in windows by invoking `cmd.exe /c mklink/d <link> <target>`.

foriequal0:
Oh, I didn't know that `std::os::windows::fs::symlink_dir` doesn't require administrator privilege. I've removed `#[cfg(not(target_os = "windows"))]`

nickwb:
> Oh, I didn't know that `std::os::windows::fs::symlink_dir` doesn't require administrator privilege. I've removed `#[cfg(not(target_os = "windows"))]`

Oh, nice - I didn't realise that there was such a function. 💯

andytom:
> **edit** I don't know why `modules::crystal::tests::folder_with_cr_file` fails on nightly. I didn't touched that module. I removed `#[cfg(not(target_os = "windows"))]` except a test that sets `$HOME`.

😞 We have some tests that are a bit flaky when run in Github Actions (see #886). Normally  they pass after rerunning the tests.

nickwb:
Hey team (@andytom @matchai @chipbuster) - This PR is all green on the build now, and has been reviewed.

I notice there are a few other PRs outstanding for `directory`, so it might be nice to get this merged in the interest of a smoother rebase cadence 😄

chipbuster:
Sorry y'all, ever since I peeked into the Fish repo history I've been wanting to make really pretty squash commit messages for PRs. Their logs are sooooo nice .____.'

@foriequal0 Thanks so much for the PR! Sorry the process dragged out a bit.

@nickwb Thanks for reviewing! :)

foriequal0:
Thanks! I'm happy to contribute to a nice project 😄.

PR #1303, opened 2020-06-10
feat(dotnet): adds additional files to detect .net projects

### LINKED ISSUE #1302: Enhance dotnet support by supporting Directory.Build.{props,targets} and Packages.props
Not really, just willing to help contribute.

There are a couple other files that may hint at a .NET directory:

* `Directory.Build.{props,targets}` is a way of configuring .NET projects one directory up from the project file. These are just standard MSBuild files. For instance `MyLibrary` will inherit the configuration from the props file:
   ```
   ./MyLibraries.sln
   ./src/Directory.Build.props
   ./src/MyLibrary/MyLibrary.csproj
   ```
   In the above example, the `src` directory is not detected.

* `Packages.props` is a central way to configure NuGet packages. These are just standard MSBuild files.

None really.

daniel-white:
#### Description

Adds additional .net specific files to detect.

#### Motivation and Context

These are additional files that can be used to detect a .NET Core project.

Closes #1302

Tested with `cargo run` on mac and unit tests

#### Checklist:

### DIFF
--- a/src/modules/dotnet.rs
+++ b/src/modules/dotnet.rs
@@ -12,6 +12,9 @@ type JValue = serde_json::Value;
 
 const GLOBAL_JSON_FILE: &str = "global.json";
 const PROJECT_JSON_FILE: &str = "project.json";
+const DIRECTORY_BUILD_PROPS_FILE: &str = "Directory.Build.props";
+const DIRECTORY_BUILD_TARGETS_FILE: &str = "Directory.Build.targets";
+const PACKAGES_PROPS_FILE: &str = "Packages.props";
 
 /// A module which shows the latest (or pinned) version of the dotnet SDK
 ///
@@ -23,7 +26,13 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     // check for the version using the JSON files
     let is_dotnet_project = context
         .try_begin_scan()?
-        .set_files(&[GLOBAL_JSON_FILE, PROJECT_JSON_FILE])
+        .set_files(&[
+            GLOBAL_JSON_FILE,
+            PROJECT_JSON_FILE,
+            DIRECTORY_BUILD_PROPS_FILE,
+            DIRECTORY_BUILD_TARGETS_FILE,
+            PACKAGES_PROPS_FILE,
+        ])
         .set_extensions(&["sln", "csproj", "fsproj", "xproj"])
         .is_match();
 
@@ -196,6 +205,7 @@ fn get_dotnet_file_type(path: &Path) -> Option<FileType> {
     match extension_lower.as_ref().map(|f| f.as_ref()) {
         Some("sln") => return Some(FileType::SolutionFile),
         Some("csproj") | Some("fsproj") | Some("xproj") => return Some(FileType::ProjectFile),
+        Some("props") | Some("targets") => return Some(FileType::MsBuildFile),
         _ => (),
     };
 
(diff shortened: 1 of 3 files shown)


chipbuster:
@daniel-white Thanks for the PR!

It looks like the tests you added are default-ignored (as they should be, since they seem to touch the filesystem). To run them, you'll need to run `cargo test dotnet -- --ignored.`

daniel-white:
finally i think i figured out what this module was doing

src/modules/dotnet.rs
```diff
@@ -196,6 +207,7 @@ fn get_dotnet_file_type(path: &Path) -> Option<FileType> {
     match extension_lower.as_ref().map(|f| f.as_ref()) {
         Some("sln") => return Some(FileType::SolutionFile),
         Some("csproj") | Some("fsproj") | Some("xproj") => return Some(FileType::ProjectFile),
+        Some("props") | Some("targets") => return Some(FileType::MsBuildFile),
```
daniel-white, line 210: the general classification of these files are msbuild specific files, i didn't see a reason to break it out right now

src/modules/dotnet.rs
```diff
@@ -23,14 +26,22 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+            PACKAGES_PROPS_FILE
+        ])
         .set_extensions(&["sln", "csproj", "fsproj", "xproj"])
         .is_match();
 
     if !is_dotnet_project {
         return None;
     }
 
+    eprintln!("File found!!!!!!DW");
```
daniel-white, line 43: let me fix that up

daniel-white:
@chipbuster do you or someone have time to review this and merge? TIA!

tests/testsuite/dotnet.rs
```diff
@@ -12,12 +12,39 @@ fn shows_nothing_in_directory_with_zero_relevant_files() -> io::Result<()> {
...
+    workspace.close()
+}
+
 #[test]
 #[ignore]
 fn shows_latest_in_directory_with_solution() -> io::Result<()> {
     let workspace = create_workspace(false)?;
     touch_path(&workspace, "solution.sln", None)?;
-    expect_output(&workspace, ".", Some("•NET v2.2.402"))?;
+    expect_output(&workspace, ".", Some("•NET v"))?;
```
chipbuster, line 47: While I now understand why these changes were necessary/correct, I spent about 30 minutes being very confused over what was going on here.

Could we leave a comment (either at the top of the dotnet test module or at the first time it's used) explaining why this form is the correct output to expect, and isn't just us breaking the tests so that the version doesn't need to display?

Alternatively, is there a way we could make this match better, e.g. by employing a regex in testing so that we at least know we're getting valid version numbers on the end and not random strings?

daniel-white: i could add a regex to this to make it more robust. i'm dabbling for the first time with rust for this so it'll take me a bit.

daniel-white: @chipbuster great thanks! i just refactored to account for a pattern string to being passed to the `expect_output` function. ill get the formatting fixed in a jiffy

daniel-white:
@chipbuster thanks! i just used the inherited pattern. i think im going to use this chance to learn some rust. i like what i see so far.

andytom:
> i think im going to use this chance to learn some rust. i like what i see so far.

@daniel-white, welcome aboard this is also how I got started with Rust :D.

PR #1305, opened 2020-06-10
Fix install script

vladimyr:
#### Description

Address `shellcheck` reports. (Mainly fixing problematic `printf` invocations.)

#### Motivation and Context

Making sure install script passes `shellcheck`.

#### Checklist:


vladimyr:
@chipbuster Please take a look 😉

install/install.sh
```diff
@@ -34,27 +34,27 @@ MAGENTA="$(tput setaf 5 2>/dev/null || echo '')"
...
+  printf "%s✓%s %s\n" "${GREEN}" "${NO_COLOR}" "$@"
 }
 
 # Gets path to a temporary file, even if 
 get_tmpfile(){
   local suffix
   suffix="$1"
   if hash mktemp; then
-    printf "$(mktemp).${suffix}"
+    printf "%s.%s" "$(mktemp)" "${suffix}"
```
chipbuster, line 57: This one I'll definitely accept without too much quibble.

install/install.sh
```diff
@@ -210,7 +210,7 @@ detect_arch() {
 
 confirm() {
   if [ -z "${FORCE-}" ]; then
-    printf "${MAGENTA}?${NO_COLOR} $@ ${BOLD}[y/N]${NO_COLOR} "
+    printf "%s?%s %s %s[y/N]%s " "${MAGENTA}" "${NO_COLOR}" "$@" "${BOLD}" "${NO_COLOR}"
```
chipbuster, line 213: I'm not certain whether fixing the shellcheck lint in this way results in more readable code or not. Consider the diff:

```
printf "${MAGENTA}?${NO_COLOR} $@ ${BOLD}[y/N]${NO_COLOR} "
```

This is a little tricky to read, but I can go through left-to-right in a single pass and rattle things off:
  - First symbol will be magenta
  - First symbol is a `?`
  - Reset color afterwards
  - Print all arguments
etc.etc.

Now look at the shellcheck-approved version:

```
printf "%s?%s %s %s[y/N]%s " "${MAGENTA}" "${NO_COLOR}" "$@" "${BOLD}" "${NO_COLOR}"
```
What's the first symbol printed? Evaluate the first argument to printf. Oh, that's a magenta color code. Now skip your eyes back to where you left off (good luck finding it in the soup of `%s`!). The symbol is a `?`, so that's a magenta `?`.. What gets printed next?  Another interpolated string. Was this argument 1 or 2? (Count the format string). Oh it's argument 2. Which argument is argument 2? (Count the arguments). It's `NO_COLOR`. etc.etc.

We shouldn't necessarily leave the shellcheck warnings in, but I think we need to think a little more closely about how we fix these. Do we need to factor color printing out into functions or special variables? Do we need to refactor the shellcode? Or should we just say that the issues in [SC2059](https://github.com/koalaman/shellcheck/wiki/SC2059) aren't relevant here and silence those warnings?

PR #1314, opened 2020-06-11
fix: remove errant whitespace which broke tar command

### LINKED ISSUE #1313: Install fails on Ubuntu 20.04 (x64)
#### Current Behavior

When running the following command `curl -fsSL https://starship.rs/install.sh | bash` the install fails due to "Starship not yet supporting your configuration."
#### Expected Behavior

Starship is expected to install as normal.

```sh
> Installing Starship, please wait…
tar (child): : Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now

x Command failed (exit code 23): curl --silent --fail --location https://github.com/starship/starship/releases/latest/download/starship-x86_64-unknown-linux-gnu.tar.gz

> This is likely due to Starship not yet supporting your configuration.
> please create an issue requesting a build for x86_64-unknown-linux-gnu:
> https://github.com/starship/starship/issues/new/\n
```

#### Possible Solution

The problem seems to be here: https://github.com/starship/starship/blob/master/install/install.sh#L82

Running the following command:

```sh
curl --fail --location "https://github.com/starship/starship/releases/latest/download/starship-x86_64-unknown-linux-gnu.tar.gz"
```

Returns the the following:

```sh
❯ curl --fail --location "https://github.com/starship/starship/releases/latest/download/starship-x86_64-unknown-linux-gnu.tar.gz"
Warning: Binary output can mess up your terminal. Use "--output -" to tell 
Warning: curl to output it to your terminal anyway, or consider "--output 
Warning: <FILE>" to save to a file.

```

Adding the `-O` flag, resolves the issue.

```sh
curl --silent --fail --location "https://github.com/starship/starship/releases/latest/download/starship-x86_64-unknown-linux-gnu.tar.gz" -O
``

#### Environment
- Starship version: 0.42.0
- bash version: GNU bash, version 5.0.16(1)-release (x86_64-pc-linux-gnu)

- Operating system: Ubuntu 20.04

```bash
# for examples

[ -z "$PS1" ] && return

# don't put duplicate lines in the history. See bash(1) for more options
# ... or force ignoredups and ignorespace
HISTCONTROL=ignoredups:ignorespace

shopt -s histappend

HISTSIZE=1000
HISTFILESIZE=2000

shopt -s checkwinsize

if [ -z "$debian_chroot" ] && [ -r /etc/debian_chroot ]; then
fi

case "$TERM" in
    xterm-color) color_prompt=yes;;
esac

#force_color_prompt=yes

	color_prompt=yes
    else
	color_prompt=
    fi
fi

else
fi

case "$TERM" in
xterm*|rxvt*)
    ;;
*)
    ;;
esac

    alias ls='ls --color=auto'
    #alias dir='dir --color=auto'

fi

# some more ls aliases
alias ll='ls -alF'
alias la='ls -A'
alias l='ls -CF'

# Alias definitions.

    . ~/.bash_aliases
fi

# sources /etc/bash.bashrc).
#if [ -f /etc/bash_completion ] && ! shopt -oq posix; then
#    . /etc/bash_completion
#fi

eval "$(starship init bash)"
```

#### Starship Configuration

```toml
<unknown config>
```

chipbuster:
#### Description

HOTFIX: curl installer will be broken until this is merged.

There was a space added to a tar command in #1305. While it seems innocuous, it causes tar to either try to extract the file `"v"` or the file `""` instead of reading from stdin like we want. This reverts that change by putting the `VERBOSE` invocation in the middle of the command string, hopefully making it more obvious if it breaks in the future.

Curiously, this did *not* break the script when executing it directly (`./install.sh`) but *did* break it when running it in a new shell (`bash install.sh`). I've determined that that has something to do with the environment, but haven't worked out anything deeper than that.

#### Motivation and Context

Closes #1313

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -114,7 +114,7 @@ fetch_and_unpack() {
   # I'd like to separate this into a fetch() and unpack() function, but I can't
   # figure out how to get bash functions to read STDIN/STDOUT from pipes
   if [ "${EXT}" = "tar.gz" ]; then
-    fetch "${URL}" | ${sudo} tar xzf "${VERBOSE}" - -C "${BIN_DIR}"
+    fetch "${URL}" | ${sudo} tar xz"${VERBOSE}"f - -C "${BIN_DIR}"
   elif [ "${EXT}" = "zip" ]; then
     # According to https://unix.stackexchange.com/q/2690, zip files cannot be read
     # through a pipe. We'll have to do our own file-based setup.


install/install.sh
```diff
@@ -114,7 +114,7 @@ fetch_and_unpack() {
   # I'd like to separate this into a fetch() and unpack() function, but I can't
   # figure out how to get bash functions to read STDIN/STDOUT from pipes
   if [ "${EXT}" = "tar.gz" ]; then
-    fetch "${URL}" | ${sudo} tar xzf "${VERBOSE}" - -C "${BIN_DIR}"
+    fetch "${URL}" | ${sudo} tar xz"${VERBOSE}"f - -C "${BIN_DIR}"
```
vladimyr, line 117: Damn, I really messed it up, sorry 🤦 
Btw, maybe we can make it even more obvious:
```suggestion
    fetch "${URL}" | ${sudo} tar -xz"${VERBOSE}"f - -C "${BIN_DIR}"
```

chipbuster: Nah, I was the reviewer there, the onus is on me. (Also, I'm a maintainer, making that doubly true).

So I don't know if this is still true, but back when I was first learning command line shenanigans a decade ago, I found that adding a `-` to tar options made them more order-sensitive, sometimes in ways that didn't agree with each other on different platforms.

Do we know that this problem no longer exists? Or perhaps it's an issue that I entirely imagined?

chipbuster: As a simple example, `tar fxvz file.tgz` works on my system, while `tar -fxvz file.tgz` generates 
```
tar: You must specify one of the '-Acdtrux', '--delete' or '--test-label' options
Try 'tar --help' or 'tar --usage' for more information.
```

vladimyr: I think `-f` needs to come last: `tar -xvzf` because that is the only one that receives argument. Does that work?

chipbuster: Yes, I'm just not sure how that works on other platforms.

I guess another way to think of the question is this: can we think of a value of VERBOSE that will break the dashless form in a bad way (e.g. extracting the wrong file, passing inadvertent flags to `tar`) but won't break the dash form?

vladimyr: > Yes, I'm just not sure how that works on other platforms.

Which platform you have? If this works on a Mac (I tested it, it does) it will work on other BSDs. The only other case would be GNU `tar` which I believe is fully backward compatible with BSD implementation.

> I guess another way to think of the question is this: can we think of a value of VERBOSE that will break the dashless form in a bad way (e.g. extracting the wrong file, passing inadvertent flags to tar) but won't break the dash form?

Well, that is a really hard question to answer... I guess we are a bit safer with dashed form because anything before `-f` would be treated as a flag, assuming that bad flag is better than operating on a bad file...

grimm26:
it would also be explicit to put a `--output -` with the curl command that is equivalent to the `-O-` on the wget command.

PR #1315, opened 2020-06-11
refactor(install): refactor install script

vladimyr:
#### Description

Complete rewrite from `bash` to POSIX `sh` containing the following changes:
- rename `complete` to `success`
- implement `has` helper for verifying program existence
- implement cross-platform `mktmpdir` with graceful degradation
- support tarball download using BSD's `fetch`
- change `curl` invocation to use short flags
- swap `curl --fail` with explicit HTTP status code check
- fix silent/verbose unzipping output
- split `fetch` and `unpack`
- avoid pipes by downloading and unpacking to tmpdir
- inline `install` steps
- split platform detection (`uname_platform`) from platform check (`platform_check`)
- split arch detection (`uname_arch`) from arch check (`arch_check`)
- split `check_bin_dir` into `bin_dir_check` and `path_check`
	_Former checks the existence of a directory later checks its presence in `$PATH`._
- support `PREFIX` environment variable

Added support for `PREFIX` environment variable which enables the following _standardish_ invocation style:
```sh
curl -fsSL https://starship.rs/install.sh | PREFIX=$HOME/.local sh
# starship ends up inside $HOME/.local/bin
```

**PS** It includes fixed `tar` invocation from #1314 

#### Motivation and Context

Keeping install procedure readable & maintainable 😄 

#### Checklist:

### DIFF
--- a/docs/de-DE/faq/README.md
+++ b/docs/de-DE/faq/README.md
@@ -49,5 +49,5 @@ Die Eingabeaufforderung verwendet so viel Kontext wie möglich, aber keine Flagg
 If you get an error like "*version 'GLIBC_2.18' not found (required by starship)*" when using the prebuilt binary (for example, on CentOS 6 or 7), you can use a binary compiled with `musl` instead of `glibc`:
 
 ```sh
-curl -fsSL https://starship.rs/install.sh | sh -s -- --platform unknown-linux-musl
+curl -fsSL https://starship.rs/install.sh | bash -s -- --platform unknown-linux-musl
 ```
--- a/docs/de-DE/faq/README.md
+++ b/docs/de-DE/faq/README.md
@@ -49,5 +49,5 @@ Die Eingabeaufforderung verwendet so viel Kontext wie möglich, aber keine Flagg
 If you get an error like "*version 'GLIBC_2.18' not found (required by starship)*" when using the prebuilt binary (for example, on CentOS 6 or 7), you can use a binary compiled with `musl` instead of `glibc`:
 
 ```sh
-curl -fsSL https://starship.rs/install.sh | bash -s -- --platform unknown-linux-musl
+curl -fsSL https://starship.rs/install.sh | sh -s -- --platform unknown-linux-musl
 ```
--- a/install/install.sh
+++ b/install/install.sh
@@ -67,8 +67,8 @@ fetch() {
 
   if has curl; then
     cmd="curl -w '%{http_code}' -sL -o $file $url"
-    code=$($cmd)
-    if [ "$code" == "200" ]; then
+    code=$(curl -w '%{http_code}' -sL -o "$file" "$url")
+    if [ "$code" = "200" ]; then
       rc=0
     else
       rc=$code
--- a/install/install.sh
+++ b/install/install.sh
@@ -54,11 +54,13 @@ has() {
 }
 
 mktmpdir() {
-  tmpdir=$(mktemp -dqt "$1") || {
-    error "Can't create temp file, exiting…"
-    return 1
-  }
-  echo "$tmpdir"
+  if tmpdir=$(mktemp -dqt "$1"); then
+    echo "$tmpdir"
+    return 0
+  fi
+
+  error "Can't create temp file, exiting…"
+  return 1
 }
 
 fetch() {
--- a/install/install.sh
+++ b/install/install.sh
@@ -63,7 +63,7 @@ mktmpdir() {
   return 1
 }
 
-fetch() {
+download() {
   file="$1"
   url="$2"
 
@@ -387,7 +387,7 @@ else
   elevated=1
 fi
 
-fetch "$archive" "$url"
+download "$archive" "$url"
 unpack "$archive"
 if [ -n "$elevated" ]; then
   sudo install "$bin" "$BIN_DIR"
--- a/install/install.sh
+++ b/install/install.sh
@@ -234,14 +234,14 @@ arch_check() {
 }
 
 bin_dir_check() {
-  if [ ! -d "$BIN_DIR" ]; then
-    error "Installation location $BIN_DIR does not appear to be a directory"
-    printf "\n" >&2
-    info "Make sure the location exists and is a directory, then try again."
-    return 1
+  if [ -d "$BIN_DIR" ]; then
+    return 0
   fi
 
-  return 0
+  error "Installation location $BIN_DIR does not appear to be a directory"
+  printf "\n" >&2
+  info "Make sure the location exists and is a directory, then try again."
+  return 1
 }
 
 path_check() {
--- a/install/install.sh
+++ b/install/install.sh
@@ -258,15 +258,12 @@ path_check() {
 
 confirm() {
   printf "%s " "${MAGENTA}?${NO_COLOR} $* ${BOLD}[y/N]${NO_COLOR}"
-  {
-    read -r yn </dev/tty
-    rc=$?
-  } || true
-
-  if [ $rc != 0 ]; then
+  if ! read -r yn </dev/tty; then
     error "Error reading from prompt (please re-run with the \`--yes\` option)"
     return 1
-  elif [ "$yn" != "y" ] && [ "$yn" != "yes" ]; then
+  fi
+
+  if [ "$yn" != "y" ] && [ "$yn" != "yes" ]; then
     error 'Aborting (please answer "yes" to continue)'
     return 1
   fi
--- a/install/install.sh
+++ b/install/install.sh
@@ -206,7 +206,7 @@ uname_arch() {
     arch=i386
   fi
 
-  case $arch in
+  case "$arch" in
     x86) arch="i386" ;;
     i686) arch="i386" ;;
     i386) arch="i386" ;;
@@ -221,7 +221,7 @@ uname_arch() {
 
 arch_check() {
   arch=${arch:-"$(uname_arch)"}
-  case $arch in
+  case "$arch" in
     x86_64) return 0 ;;
   esac
 
@@ -357,7 +357,7 @@ if [ -n "$verbose" ]; then
 fi
 
 tmp=$(mktmpdir "starship")
-case $platform in
+case "$platform" in
   pc-windows-msvc)
     bin="$tmp/starship.exe"
     ext="zip"
--- a/install/install.sh
+++ b/install/install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # Options
 #
@@ -20,7 +20,7 @@
 #   -B, --base-url
 #     Override the base URL used for downloading releases
 
-set -euo pipefail
+set -eu
 printf "\n"
 
 BOLD="$(tput bold 2>/dev/null || echo '')"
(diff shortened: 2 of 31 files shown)


install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
-      exit 1
+      rc=$code
     fi
+  elif has wget; then
+    cmd="wget -q -O $file $url"
+    {
+      wget -q -O "$file" "$url"
+      rc=$?
+    } || true
+  elif has fetch; then
```
chipbuster, line 82: Doesn't this cause an infinite recursion on systems that don't have curl/wget installed? `fetch` should be defined at this point (since we're calling it), which will cause this block to infinitely succeed.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
+    pc-windows-msvc) return 0 ;;
+    unknown-linux-musl) return 0 ;;
+    unknown-linux-gnu) return 0 ;;
+  esac
+
+  arch=$(uname_arch)
 
-  echo "${platform}"
+  error "Builds for $platform are not yet available for Starship"
+  printf "\n" >&2
```
chipbuster, line 192: Why not place this `\n` in the error string?

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
+path_check() {
+  # https://stackoverflow.com/a/9663359
+  case :$PATH: in
+    *:$BIN_DIR:*) return 0 ;;
+  esac
+
+  warn "Bin directory $BIN_DIR is not in your \$PATH"
+  printf "\n"
+
+  return 0
```
chipbuster, line 256: Might be good to have a comment explaining why we return 0 even if it's not in `PATH`. I did a double-take until I remembered how this function is used.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
-  )
+  return 0
+}
 
-  if [ "${good}" != "1" ]; then
-    warn "Bin directory ${bin_dir} is not in your \$PATH"
-  fi
+path_check() {
+  # https://stackoverflow.com/a/9663359
+  case :$PATH: in
```
chipbuster, line 249: This has a failure case where the user sets `BIN_DIR` to be something with a `:` in it. While I'm not super worried about it, I don't know if there are whitespace-based conditions I'm not thinking of.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
-  ARCH="$(detect_arch)"
-fi
+  return 0
+}
 
-if [ -z "${BASE_URL-}" ]; then
-  BASE_URL="https://github.com/starship/starship/releases"
-fi
+# defaults
+PREFIX=${PREFIX:-"/usr/local"}
```
chipbuster, line 278: Why separate out `PREFIX` and `BIN_DIR`? This has the potential to be a bit confusing when the thing you're asked by the prompt (install location) and the environment variable you set (`PREFIX`) are different.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
-  -f=* | -y=* | --force=* | --yes=*)
-    FORCE="${1#*=}"
-    shift 1
-    ;;
-
-  *)
-    error "Unknown option: $1"
-    exit 1
-    ;;
+    -p | --platform)
```
chipbuster, line 288: Not opposed to capitalization changes, but curious if there's a rationale for them.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
+if [ -n "$elevated" ]; then
+  sudo install "$bin" "$BIN_DIR"
+else
+  install "$bin" "$BIN_DIR"
+fi
 
-install
-complete "Starship installed"
+success "Starship installed"
+printf "\n"
```
chipbuster, line 402: Similar question to before: move this into the success message?

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
-      wget -O- -q "$1"
-      rc=$?
-      set -e
+  file="$1"
+  url="$2"
+
+  if has curl; then
+    cmd="curl -w '%{http_code}' -sL -o $file $url"
+    code=$(curl -w '%{http_code}' -sL -o "$file" "$url")
+    if [ "$code" = "200" ]; then
```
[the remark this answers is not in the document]
chipbuster: Unlike my previous comment on mktempdir, this can easily be simulated in production by just `curl`ing a large file and then shutting off your network. The HTTP code will be 200, but the file will be incomplete.

install/install.sh
```diff
@@ -20,7 +20,7 @@
 #   -B, --base-url
 #     Override the base URL used for downloading releases
 
-set -euo pipefail
+set -eu
```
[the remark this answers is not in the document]
vladimyr: POSIX does not define it so `sh` lacks it. Which is why I swapped piped untarring with download to a temp file and extract tarball from disk.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
   printf "%s\n" "${RED}x $*${NO_COLOR}" >&2
 }
 
-complete() {
+success() {
```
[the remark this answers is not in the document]
vladimyr: Actually, I had to change it because complete is clashing with _something_ inside `sh` causing the script to error.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
-# Test if a location is writeable by trying to write to it. Windows does not let
-# you test writeability other than by writing: https://stackoverflow.com/q/1999988
-test_writeable() {
-  local path
-  path="${1:-}/test.txt"
-  if touch "${path}" 2>/dev/null; then
-    rm "${path}"
-    return 0
-  else
+mktmpdir() {
```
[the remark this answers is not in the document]
vladimyr: Shouldn't or operator prevent that? 🤔
FWIW, I agree with your line of thinking regarding `mktemp` availability but I'm not sure what is the situation in git bash on windows.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
-      set +e
-      command="wget -O- -q $1"
-      wget -O- -q "$1"
-      rc=$?
-      set -e
+  file="$1"
+  url="$2"
+
+  if has curl; then
+    cmd="curl -w '%{http_code}' -sL -o $file $url"
```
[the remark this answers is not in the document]
vladimyr: I've c/p it 🤣

install/install.sh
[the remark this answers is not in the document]
vladimyr: Wow, that's a really good catch! I went with HTTP codes because the curl won't fail on 404s.

install/install.sh
vladimyr: Probably 🤦
I'll rename function to avoid that.

install/install.sh
vladimyr: Unfortunately, print modifiers work only inside the first `printf` argument.

install/install.sh
vladimyr: A path with a semicolon is not a valid path? Even if you extend this to support bin dir with semicolon that would inevitably fail later on? 🤔

install/install.sh
vladimyr: This is actually a convention in *nix world. https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

install/install.sh
vladimyr: I did it to emphasize the distinction between standard variables overridable from the environment (prefix, bin dir) and color modifiers vs other variables. Former are left uppercased while later got lowercased.

install/install.sh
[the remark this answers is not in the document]
chipbuster: If you install it with the default set of minimal UNIX utilities, `mktemp` exists. I'm not sure what happens if you go for a true bare minimum install, but given that you're unlikely to be able to `curl` things, or `unzip` them, it's probably safe to ignore that case.
 
(If someone opens an issue, we can just tell them to use the scoop/chocolatey installs)

install/install.sh
vladimyr: The same answer, I would love to but sadly I can't. 😥

install/install.sh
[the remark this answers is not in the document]
chipbuster: > Shouldn't or operator prevent that? 

Everything I read suggests that it should, but when I tested it by deleting mktemp, it didn't 😖

install/install.sh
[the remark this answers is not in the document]
chipbuster: Makes sense.

install/install.sh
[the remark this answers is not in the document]
chipbuster: Does the `--fail` flag make it fail on 404s?

According to the link, it doesn't work for some 4xx errors involving auth, but I don't think we need to worry about that.

https://superuser.com/questions/590099/can-i-make-curl-fail-with-an-exitcode-different-than-0-if-the-http-status-code-i

install/install.sh
chipbuster: Oh heck, I just realized....this creates an additional blank line, doesn't it?

I thought it was needed to make the first newline appear. Whoops.

install/install.sh
```diff
@@ -45,328 +45,362 @@ error() {
...
+if [ -z "$force" ]; then
+  confirm "Install Starship ${GREEN}latest${NO_COLOR} to ${BOLD}${GREEN}${BIN_DIR}${NO_COLOR}?"
 fi
+path_check
 
-URL="${BASE_URL}/latest/download/starship-${ARCH}-${PLATFORM}.${EXT}"
-info "Tarball URL: ${UNDERLINE}${BLUE}${URL}${NO_COLOR}"
-confirm "Install Starship ${GREEN}latest${NO_COLOR} to ${BOLD}${GREEN}${BIN_DIR}${NO_COLOR}?"
-check_bin_dir "${BIN_DIR}"
+if [ -w "$BIN_DIR" ]; then
```
chipbuster, line 384: I'll need to test this again, but I thought that `-w` on Windows doesn't actually tell you if the directory is writeable or not.

install/install.sh
chipbuster: Okay, so at the moment, if I wanted to install starship to `DIR`, I would set `BIN_DIR=$DIR` because we don't install stuff to share/include/lib/libexec/whatever. I was thinking of `PREFIX` as the final install location.

install/install.sh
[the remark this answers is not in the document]
vladimyr: Actually if used with `--fail` it will result in 22. Ok, time to revert this. 😉

install/install.sh
vladimyr: PREFIX is really just a prefix/parent dir. 🙃

install/install.sh
vladimyr: Yup, generates additional blank line.

install/install.sh
vladimyr: I wanted to ask you about that. It really should, previous workaround with `type` was meant to be used inside batch scripts. This is something that bash should provide to us.

install/install.sh
chipbuster: The only invalid characters in UNIX paths are `NULL` and `/`, everything else is fair game (which includes Unicode on modern systems).

This is a pretty troll edge case, but with the following setup:
  - I have a directory at `/bin:/sbin`, as in the directory in the root is called `bin:`, and it has a child directory called `sbin`.
  - I ask for `BIN_DIR="/bin:/sbin"`
  - My `PATH` is `/usr/bin:/bin:/sbin:.local/bin`

This will successfully install the binary without warning me that it's not on my PATH, because the `BIN_DIR` is a valid directory, albeit one with a weird name.

Now, if anyone actually pulls this stunt and then files an issue, I think I'm well within my rights to ask what they think they're doing. My point is that glob-based matches can often have unexpected behavior, especially around word splitting. I'm not terribly familiar with the rules around globs in `case` statements: can we argue that whitespace splitting isn't an issue? If the PATH has whitespace, or BIN_DIR has whitespace, can we confidently say that there won't be breaks or incidental matches?

install/install.sh
chipbuster: Yeah, that seems like a good idea.

install/install.sh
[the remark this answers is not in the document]
chipbuster: Let's at least convert this to long-flag form then. It won't be perfect, but I think `curl --write-out '${http_code}' --silent --location --output $file $url` at least gives a few more hints as to what each flag is doing, and is slightly easier to look up in the manpage if needed.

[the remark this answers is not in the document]
vladimyr: Printing out HTTP code is actually redundant since we are going back to the `--fail` approach. Regarding other flags, well, I've converted them to short ones in order to make diagnostic output (error trace) shorter. TBH I don't care much about their form, so I'll just change it as you suggested.

install/install.sh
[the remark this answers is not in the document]
vladimyr: > Everything I read suggests that it should, but when I tested it by deleting mktemp, it didn't 😖

Please try again with the current version. This should be handled now 🤞

install/install.sh
vladimyr: > The only invalid characters in UNIX paths are `NULL` and `/`, everything else is fair game (which includes Unicode on modern systems).

What can I say? First I was a bit ashamed to forget such an important fact. And then it inspired me to handle even those troll edge cased 😄

install/install.sh
```diff
@@ -45,328 +45,343 @@ error() {
...
+  error "Installation location $BIN_DIR does not appear to be a directory"
+  printf "\n" >&2
+  info "Make sure the location exists and is a directory, then try again."
+  return 1
 }
 
-check_bin_dir() {
-  local bin_dir="$1"
+path_check() {
+  echo "$PATH" | tr ':' '\0' | xargs -0 -I"{}" [ "{}" != "$BIN_DIR" ] || return 0
```
vladimyr, line 234: Now everyone can enjoy installing starship inside `/bin:/sbin` :trollface:

docs/de-DE/faq/README.md
```diff
@@ -49,5 +49,5 @@ Die Eingabeaufforderung verwendet so viel Kontext wie möglich, aber keine Flagg
 If you get an error like "*version 'GLIBC_2.18' not found (required by starship)*" when using the prebuilt binary (for example, on CentOS 6 or 7), you can use a binary compiled with `musl` instead of `glibc`:
 
 ```sh
-curl -fsSL https://starship.rs/install.sh | bash -s -- --platform unknown-linux-musl
+curl -fsSL https://starship.rs/install.sh | sh -s -- --platform unknown-linux-musl
```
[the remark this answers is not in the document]
vladimyr: To avoid any potential problems I've removed all changes from translated docs.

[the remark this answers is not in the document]
matchai: Looks like I missed this! Yep, it's best if we don't touch any localized docs.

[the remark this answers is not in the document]
vladimyr: > Looks like I missed this! Yep, it's best if we don't touch any localized docs.

Thanks for the info, I've already removed them from this PR 👍

chipbuster:
Finally got a chance to boot into Windows + Git Bash and test this:

```
chipb@onnyx MINGW64 ~
$ [ -w "C:\\Program Files" ] && echo "LAWL"
LAWL

chipb@onnyx MINGW64 ~
$ touch "C:\\Program Files\\lol"
touch: cannot touch 'C:\Program Files\lol': Permission denied
```

This tells me that Git Bash thinks that I can write directly into "C:\Program Files", even though I empirically cannot. I'm not sure exactly why this happens, but I'm pretty sure relying on the behavior of `[ -w ]` on Windows is going to lead to false positives. Maybe there's a different flag you have to use?

install/install.sh
chipbuster: Just to confirm: for some reason or other, `-w` on Git Bash in windows incorrectly believes that you can write into system directories. Maybe we should bring back the old test-writeability function that actually tries to write a file?

vladimyr: > Just to confirm: for some reason or other, `-w` on Git Bash in windows incorrectly believes that you can write into system directories. Maybe we should bring back the old test-writeability function that actually tries to write a file?

Windows, again 😫  Ok, it seems we need to bring it back. I'll update this later on 👍

vladimyr:
> EDIT: Just to be safe, I tested on a bunch of other stuff--pretty much the only time the `-w` test doesn't return true is when the path doesn't exist.

That is probably a consequence of the flaky implementation of Unix permissions in Windows made on top of NT because spec clearly says it checks permissions: https://tldp.org/LDP/abs/html/fto.html#:~:text=%2Dw,(for%20the%20user%20running%20the%20test) 😩 

I'm starting to think that using _curl-pipe-sh_ installer should be _verboten_ on Windows and separate posh install script should be offered 🤔  Something like [Scoop's installer](https://scoop.sh/#installs-in-seconds:~:text=Invoke%2DExpression%20(New%2DObject%20System.Net.WebClient).DownloadString('https%3A%2F%2Fget.scoop.sh'),iwr%20%2Duseb%20get.scoop.sh%20%7C%20iex) (source https://raw.githubusercontent.com/lukesampson/scoop/master/bin/install.ps1)

andytom:
I'm going to close this, some significant refactoring has already been done in #2228 so I'm not sure if this PR is needed anymore. If it is please feel free to reopen the PR.

PR #1322, opened 2020-06-13
feat(config): Adding prefix/suffix config options to most modules

ZeePal:
#### Description
I have added `suffix` and `prefix` config options to most modules. The defaults they had hard coded have been moved to default config values.

I have also added an optional symbol for `cmd_duration`:
https://github.com/starship/starship/pull/1322/files#diff-e77ee7c3a1c69e40334459a90d4f0d8fR9

I also reordered the statements in the `jobs` module as config loading isn't needed if their are no detected jobs :D:
https://github.com/starship/starship/pull/1322/files#diff-00b059f845b3814d13363448ac0b4fa0R18

#### Motivation and Context

I believe we want Starship to be as configurable as possible which this helps with.
I also would like a smaller command prompt (Removing all the "via " words!)

I have tested with my starship config here: https://github.com/ZeePal/workstation-ubuntu/blob/master/roles/starship/tasks/main.yml
I have also run `cargo test`.

#### Checklist:


ZeePal:
Oh didn't see the format strings milestone: https://github.com/starship/starship/issues/1057

At least this is a stop gap for me.

matchai:
Thank you for the PR ZeePal! 😄 
I'll be closing this PR as we will not be adding any new module affixes until the release of the format strings milestone you linked to.

PR #1336, opened 2020-06-16
feat: git branch: read from HEAD on newly initialized repo

### LINKED ISSUE #1327: Git Branch: get initial branch name (immediately after git init, no commits)
On initializing a git repo, the git branch module will always show `master` as the name, because no branch has been created until a commit is made and the default is `master`.

I hope that there is a way to retrieve the actual name of initial branch. `git status --porcelain --branch` can return `## No commits yet on main`, so there's some way somewhere; perhaps it's in `git2` already (I looked, but may have missed it). I filed an issue in `git2`: https://github.com/rust-lang/git2-rs/issues/570

I guess that always showing `master` as default is the alternative. Or perhaps showing something like `no branch` as default, but that would probably be confusing.

Some other statusline parses the output of the cli, but it doesn't make sense to switch away from `git2`.

hwchen:
#### Description, motivation, context
On a newly initialized git repo, there are no branches created until a
commit is made. Previously, starship handled this by hardcoding a default
value "master" for when branch `head` could not be read.

However, if a user wants to set a different default branch name, that
name won't appear in starship until a commit is made to the branch.

If git2 provides a way to read the default branch name, we can use that,
but at the moment it's not obvious how.

For the moment, we can directly read `.git/HEAD`, which contains the
name of the default branch head reference. This commit implements this
strategy.

Closes #1327

I've updated the `git_branch::...unborn_master` to `git_branch::...unborn_default_branch`, where the default branch name is set in `.git/HEAD`.

I've also done manual testing on Linux, where I created a new git repo with default branch `main` in `.git/HEAD`, then run the locally built version of starship in that repo as `../starship/target/debug/starship module git_branch`.

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -322,7 +322,8 @@ fn get_current_branch(repository: &Repository) -> Option<String> {
                 head_path.push("HEAD");
 
                 // get first line, then last path segment
-                fs::read_to_string(&head_path).ok()?
+                fs::read_to_string(&head_path)
+                    .ok()?
                     .lines()
                     .next()?
                     .trim()
--- a/src/context.rs
+++ b/src/context.rs
@@ -317,8 +317,18 @@ fn get_current_branch(repository: &Repository) -> Option<String> {
         Err(e) => {
             return if e.code() == UnbornBranch {
                 // HEAD should only be an unborn branch if the repository is fresh,
-                // in that case assume "master"
-                Some(String::from("master"))
+                // in that case read directly from `.git/HEAD`
+                let mut head_path = repository.path().to_path_buf();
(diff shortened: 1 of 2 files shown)


Zegnat:
I am very surprised we’re having to manually read the file inside `.git`, which would generally be something to shy away from 😕 I would have wanted to run something like the following instead:
```bash
git symbolic-ref --short HEAD
```
But I can’t find any way to access the reference for HEAD through the libgit2 bindings at all? Grabbing HEAD is what gives the UnbornBranch error in the first place. Odd!

mijailr:
Any update? I'm waiting for this ❤️

PR #1338, opened 2020-06-16
refactor: Switch from `dirs` to `dirs-next` dependency

### LINKED ISSUE #1319: dirs -> dirs-next
According to [several posts](https://www.reddit.com/r/rust/comments/gli4pc/directoriesnext_dirsnext_the_new_home_for/), the `dirs` crate is going on a multi-year hiatus (it sounds like a mini-replay of actix-web, but I don't like to think about the actix-web incident).

Someone has forked `dirs` to make `dirs-next`, should we consider switching?

juliangehring:
#### Description
Replaces the `dirs` dependency with the maintained fork `dirs-next`.

#### Motivation and Context
As the `dirs` dependency will be unmaintained for the foreseeable future, switch to the maintained fork `dirs-next`. This change has been suggested in #1319.

Closes #1319.

#### Checklist:
- [ ] I have updated the documentation accordingly: Should not be needed since the change of the dependency is not visible to the user.

### DIFF
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -109,7 +109,7 @@ fn get_editor_internal(visual: Option<OsString>, editor: Option<OsString>) -> Os
 fn get_config_path() -> OsString {
     let config_path = env::var_os("STARSHIP_CONFIG").unwrap_or_else(|| "".into());
     if config_path.is_empty() {
-        dirs::home_dir()
+        dirs_next::home_dir()
             .expect("couldn't find home directory")
             .join(".config/starship.toml")
             .as_os_str()
--- a/src/context.rs
+++ b/src/context.rs
@@ -89,7 +89,7 @@ impl<'a> Context<'a> {
     fn expand_tilde(dir: PathBuf) -> PathBuf {
         if dir.starts_with("~") {
             let without_home = dir.strip_prefix("~").unwrap();
-            return dirs::home_dir().unwrap().join(without_home);
+            return dirs_next::home_dir().unwrap().join(without_home);
         }
         dir
     }
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -55,7 +55,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             .unwrap_or_else(|| &context.current_dir),
     );
 
-    let home_dir = dirs::home_dir().unwrap();
+    let home_dir = dirs_next::home_dir().unwrap();
     log::debug!("Current directory: {:?}", current_dir);
 
     let repo = &context.get_repo().ok()?;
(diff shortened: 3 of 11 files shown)


src/configure.rs
```diff
@@ -109,7 +110,7 @@ fn get_editor_internal(visual: Option<OsString>, editor: Option<OsString>) -> Os
 fn get_config_path() -> OsString {
     let config_path = env::var_os("STARSHIP_CONFIG").unwrap_or_else(|| "".into());
     if config_path.is_empty() {
-        dirs::home_dir()
+        home_dir()
```
chipbuster, line 113: I don't think we need to `use dirs_next::home_dir;` for a single call. I don't usually memorize the imports of a module before reading the code--this makes it harder to figure out where this function comes from, while making the overall length of the code longer.

src/context.rs
```diff
@@ -89,7 +90,7 @@ impl<'a> Context<'a> {
     fn expand_tilde(dir: PathBuf) -> PathBuf {
         if dir.starts_with("~") {
             let without_home = dir.strip_prefix("~").unwrap();
-            return dirs::home_dir().unwrap().join(without_home);
+            return home_dir().unwrap().join(without_home);
```
chipbuster, line 93: Same here.

src/modules/directory.rs
```diff
@@ -55,7 +56,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             .unwrap_or_else(|| &context.current_dir),
     );
 
-    let home_dir = dirs::home_dir().unwrap();
+    let home_dir = home_dir().unwrap();
```
chipbuster, line 59: And here.

src/modules/kubernetes.rs
```diff
@@ -48,7 +49,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             .filter_map(|filename| parse_kubectl_file(&filename))
             .next(),
         Err(_) => {
-            let filename = dirs::home_dir()?.join(".kube").join("config");
+            let filename = home_dir()?.join(".kube").join("config");
```
chipbuster, line 52: And here.

tests/testsuite/directory.rs
```diff
@@ -1,5 +1,5 @@
 use ansi_term::Color;
-use dirs::home_dir;
+use dirs_next::home_dir;
```
chipbuster, line 2: This one is fine and arguably should have been in here in the first place, since we use it so often :+1:

juliangehring:
Closing this in favour of #1361. The CI test doesn't seem to pick up code changes in my latest commits.

PR #1344, opened 2020-06-18
Favor finding Starship on the PATH over its absolute path on disk (#910)

### LINKED ISSUE #910: Shell does not work after upgrading with Homebrew
#### Current Behavior
```
brew upgrade
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (homebrew/core).
==> New Formulae
rubyfmt
==> Updated Formulae
starship ✔  clp         composer    erlang@21   hugo        protobuf    ungit
cfn-lint    coinutils   erlang      gitmoji     paket       syncthing

==> Upgrading 1 outdated package:
starship 0.33.1 -> 0.34.1
==> Upgrading starship
==> Downloading https://linuxbrew.bintray.com/bottles/starship-0.34.1.x86_64_linux.bottl
==> Downloading from https://akamai.bintray.com/bb/bba0c68907b0ef7d3cbc08c307d9ff45a659a
######################################################################## 100.0%
==> Pouring starship-0.34.1.x86_64_linux.bottle.tar.gz
🍺  /home/linuxbrew/.linuxbrew/Cellar/starship/0.34.1: 7 files, 5.7MB
Removing: /home/linuxbrew/.linuxbrew/Cellar/starship/0.33.1... (6 files, 7.5MB)
Removing: /home/thedrow/.cache/Homebrew/starship--0.33.1.tar.gz... (4.9MB)
==> Checking for dependents of upgraded formulae...
==> No dependents found!
bash: /home/linuxbrew/.linuxbrew/Cellar/starship/0.33.1/bin/starship: No such file or directory
bash: /home/linuxbrew/.linuxbrew/Cellar/starship/0.33.1/bin/starship: No such file or directory

bash: /home/linuxbrew/.linuxbrew/Cellar/starship/0.33.1/bin/starship: No such file or directory
bash: /home/linuxbrew/.linuxbrew/Cellar/starship/0.33.1/bin/starship: No such file or directory
```
No further commands can be inputted since the previous executable was removed.
#### Expected Behavior

The shell should continue to function until it is closed.

None
#### Possible Solution

None

#### Environment
- Starship version: 0.34.1

- Terminal emulator: Alacritty 0.4.1

```bash
# for examples

case $- in
    *i*) ;;
      *) return;;
esac

HISTCONTROL=ignoreboth

shopt -s histappend

HISTSIZE=1000
HISTFILESIZE=2000

shopt -s checkwinsize

#shopt -s globstar

fi

case "$TERM" in
esac

#force_color_prompt=yes

	color_prompt=yes
    else
	color_prompt=
    fi
fi

else
fi

case "$TERM" in
xterm*|rxvt*)
    ;;
*)
    ;;
esac

    alias ls='ls --color=auto'
    #alias dir='dir --color=auto'

fi

# some more ls aliases
alias ll='ls -alF'
alias la='ls -A'
alias l='ls -CF'

#   sleep 10; alert

# Alias definitions.

    . ~/.bash_aliases
fi

# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
    . /etc/bash_completion
  fi
fi
export PATH="/home/linuxbrew/.linuxbrew/bin:$PATH"
export MANPATH="/home/linuxbrew/.linuxbrew/share/man:$MANPATH"
export INFOPATH="/home/linuxbrew/.linuxbrew/share/info:$INFOPATH"

eval "$(pyenv init -)"
export EDITOR=nano

# added by travis gem
[ -f /home/thedrow/.travis/travis.sh ] && source /home/thedrow/.travis/travis.sh

pyenv virtualenvwrapper_lazy

export PATH="/home/linuxbrew/.linuxbrew/sbin:$PATH"
eval "$(hub alias -s)"
export NVM_DIR="$HOME/.nvm"
  [ -s "/home/linuxbrew/.linuxbrew/opt/nvm/nvm.sh" ] && . "/home/linuxbrew/.linuxbrew/opt/nvm/nvm.sh"  # This loads nvm
  [ -s "/home/linuxbrew/.linuxbrew/opt/nvm/etc/bash_completion.d/nvm" ] && . "/home/linuxbrew/.linuxbrew/opt/nvm/etc/bash_completion.d/nvm"

eval "$(starship init bash)"
eval $(thefuck --alias)
```

#### Starship Configuration

```toml
<unknown config>
```

nirvdrum:
The idea here is to search for Starship on the PATH when printing out its command path during shell command generation. This is particularly helpful for people using Homebrew, which deletes the old version of any software it installs upon upgrade. Prior to this change, _starship_ would print out the fully resolved path to the binary on disk, and not the symlink that Homebrew creates and is generally available on the path. Consequently, `starship` would print out a path that no longer exists after Homebrew's cleanup, and this in turn would break any already open terminal.

Apologies, but I'm not familiar with conventional commits and don't know whether this is a feature or a fix.

#### Description

This change looks for `starship` on the `PATH` by way of the [which crate](https://crates.io/crates/which). _which_ is cross-platform and is lightweight both in terms of its dependency list and its overall project size.

#### Motivation and Context

Closes #910 

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,34 +14,34 @@ version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
- "getrandom 0.2.7",
+ "getrandom 0.2.8",
  "once_cell",
  "version_check",
 ]
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
 name = "android_system_properties"
-version = "0.1.4"
+version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7ed72e1635e121ca3e79420540282af22da58be50de153d36f81ddc6b83aa9e"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
 dependencies = [
  "libc",
 ]
 
 [[package]]
 name = "anyhow"
-version = "1.0.62"
+version = "1.0.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1485d4d2cc45e7b201ee3767015c96faa5904387c9d87c6efdd0fb511f12d305"
+checksum = "216261ddc8289130e551ddcd5ce8a064710c0d064a4d2895c67151c92b5443f6"
 
 [[package]]
 name = "arc-swap"
@@ -93,16 +93,16 @@ dependencies = [
 
 [[package]]
 name = "async-io"
-version = "1.8.0"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ab006897723d9352f63e2b13047177c3982d8d79709d713ce7747a8f19fd1b0"
+checksum = "e8121296a9f05be7f34aa4196b1747243b3b62e048bb7906f644f3fbfc490cf7"
 dependencies = [
+ "async-lock",
  "autocfg",
  "concurrent-queue",
  "futures-lite",
  "libc",
  "log",
- "once_cell",
  "parking",
  "polling",
  "slab",
@@ -113,11 +113,12 @@ dependencies = [
 
 [[package]]
 name = "async-lock"
-version = "2.5.0"
+version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e97a171d191782fba31bb902b14ad94e24a68145032b7eedf871ab0bc0d077b6"
+checksum = "c8101efe8695a6c17e02911402145357e718ac92d3ff88ae8419e84b1707b685"
 dependencies = [
  "event-listener",
+ "futures-lite",
 ]
 
 [[package]]
@@ -139,9 +140,9 @@ checksum = "7a40729d2133846d9ed0ea60a8b9541bccddab49cd30f0715a1da672fe9a2524"
 
 [[package]]
 name = "async-trait"
-version = "0.1.57"
+version = "0.1.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76464446b8bc32758d7e88ee1a804d9914cd9b1cb264c029899680b0be29826f"
+checksum = "1e805d94e6b5001b651426cf4cd446b1ab5f319d27bab5c644f61de0a804360c"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -176,9 +177,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "base64"
-version = "0.13.0"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
 name = "bitflags"
@@ -203,9 +204,9 @@ dependencies = [
 
 [[package]]
 name = "block-buffer"
-version = "0.10.2"
+version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0bf7fe51849ea569fd452f37822f606a5cabb684dc918707a0193fd4664ff324"
+checksum = "69cce20737498f97b993470a6e536b8523f0af7892a4f928cceb1ac5e52ebe7e"
 dependencies = [
  "generic-array",
 ]
@@ -242,16 +243,17 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.11.0"
+version = "3.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1ad822118d20d2c234f427000d5acc36eabe1e29a348c89b63dd60b13f28e5d"
+checksum = "572f695136211188308f16ad2ca5c851a712c464060ae6974944458eb83880ba"
 
 [[package]]
 name = "byte-unit"
-version = "4.0.14"
+version = "4.0.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95ebf10dda65f19ff0f42ea15572a359ed60d7fc74fdc984d90310937be0014b"
+checksum = "581ad4b3d627b0c09a0ccb2912148f839acaca0b93cf54cbe42b6c674e86079c"
 dependencies = [
+ "serde",
  "utf8-width",
 ]
 
@@ -290,9 +292,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.73"
+version = "1.0.74"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2fff2a6927b3bb87f9595d67196a70493f627687a71d87a0d692242c33f58c11"
+checksum = "581f5dba903aac52ea3feb5ec4810848460ee833876f1f9b0fdeab1f19091574"
 
 [[package]]
 name = "cfg-if"
@@ -377,11 +379,21 @@ checksum = "218d6bd3dde8e442a975fa1cd233c0e5fded7596bccfe39f58eca98d22421e0a"
 
 [[package]]
 name = "cmake"
-version = "0.1.48"
+version = "0.1.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e8ad8cef104ac57b68b89df3208164d228503abbdce70f6880ffa3d970e7443a"
+checksum = "db34956e100b30725f2eb215f90d4871051239535632f84fea3bc92722c66b7c"
 dependencies = [
  "cc",
+]
+
+[[package]]
+name = "codespan-reporting"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3538270d33cc669650c4b093848450d380def10c331d38c768e34cac80576e6e"
+dependencies = [
+ "termcolor",
+ "unicode-width",
 ]
 
 [[package]]
@@ -416,18 +428,18 @@ dependencies = [
 
 [[package]]
 name = "const_format"
-version = "0.2.26"
+version = "0.2.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "939dc9e2eb9077e0679d2ce32de1ded8531779360b003b4a972a7a39ec263495"
+checksum = "7309d9b4d3d2c0641e018d449232f2e28f1b22933c137f157d3dbc14228b8c0e"
 dependencies = [
  "const_format_proc_macros",
 ]
 
 [[package]]
 name = "const_format_proc_macros"
-version = "0.2.22"
+version = "0.2.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef196d5d972878a48da7decb7686eded338b4858fbabeed513d63a7c98b2b82d"
+checksum = "d897f47bf7270cf70d370f8f98c1abb6d2d4cf60a6845d30e05bfb90c6568650"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -458,9 +470,9 @@ checksum = "5827cebf4670468b8772dd191856768aedcb1b0278a04f989f7766351917b9dc"
 
 [[package]]
 name = "cpufeatures"
-version = "0.2.4"
+version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc948ebb96241bb40ab73effeb80d9f93afaad49359d159a5e61be51619fe813"
+checksum = "28d997bd5e24a5928dd43e46dc529867e207907fe0b239c3477d924f7f2ca320"
 dependencies = [
  "libc",
 ]
@@ -497,26 +509,24 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.10"
+version = "0.9.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "045ebe27666471bb549370b4b0b3e51b07f56325befa4284db65fc89c02511b1"
+checksum = "f916dfc5d356b0ed9dae65f1db9fc9770aa2851d2662b988ccf4fe3516e86348"
 dependencies = [
  "autocfg",
  "cfg-if 1.0.0",
  "crossbeam-utils",
  "memoffset",
- "once_cell",
  "scopeguard",
 ]
 
 [[package]]
 name = "crossbeam-utils"
-version = "0.8.11"
+version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51887d4adc7b564537b15adcfb307936f8075dfcd5f00dde9a9f1d29383682bc"
+checksum = "edbafec5fa1f196ca66527c1b12c2ec4745ca14b50f1ad8f9f6f720b55d11fac"
 dependencies = [
  "cfg-if 1.0.0",
- "once_cell",
 ]
 
 [[package]]
@@ -530,14 +540,59 @@ dependencies = [
 ]
 
 [[package]]
-name = "dashmap"
-version = "5.3.4"
+name = "cxx"
+version = "1.0.80"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3495912c9c1ccf2e18976439f4443f3fee0fd61f424ff99fde6a66b15ecb448f"
+checksum = "6b7d4e43b25d3c994662706a1d4fcfc32aaa6afd287502c111b237093bb23f3a"
+dependencies = [
+ "cc",
+ "cxxbridge-flags",
+ "cxxbridge-macro",
+ "link-cplusplus",
+]
+
+[[package]]
+name = "cxx-build"
+version = "1.0.80"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "84f8829ddc213e2c1368e51a2564c552b65a8cb6a28f31e576270ac81d5e5827"
+dependencies = [
+ "cc",
+ "codespan-reporting",
+ "once_cell",
+ "proc-macro2",
+ "quote",
+ "scratch",
+ "syn",
+]
+
+[[package]]
+name = "cxxbridge-flags"
+version = "1.0.80"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e72537424b474af1460806647c41d4b6d35d09ef7fe031c5c2fa5766047cc56a"
+
+[[package]]
+name = "cxxbridge-macro"
+version = "1.0.80"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "309e4fb93eed90e1e14bea0da16b209f81813ba9fc7830c20ed151dd7bc0a4d7"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "dashmap"
+version = "5.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "907076dfda823b0b36d2a1bb5f90c96660a5bbcd7729e10727f07858f22c4edc"
 dependencies = [
  "cfg-if 1.0.0",
  "hashbrown",
  "lock_api",
+ "once_cell",
  "parking_lot_core",
 ]
 
@@ -583,11 +638,11 @@ dependencies = [
 
 [[package]]
 name = "digest"
-version = "0.10.3"
+version = "0.10.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f2fb860ca6fafa5552fb6d0e816a69c8e49f0908bf524e30a90d97c85892d506"
+checksum = "adfbc57365a37acbd2ebf2b64d7e69bb766e2fea813521ed536f5d0520dcf86c"
 dependencies = [
- "block-buffer 0.10.2",
+ "block-buffer 0.10.3",
  "crypto-common",
 ]
 
@@ -742,14 +797,14 @@ dependencies = [
 
 [[package]]
 name = "filetime"
-version = "0.2.17"
+version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e94a7bbaa59354bc20dd75b67f23e2797b4490e9d6928203fb105c79e448c86c"
+checksum = "4b9663d381d07ae25dc88dbdf27df458faa83a9b25336bcac83d5e452b5fc9d3"
 dependencies = [
  "cfg-if 1.0.0",
  "libc",
  "redox_syscall",
- "windows-sys",
+ "windows-sys 0.42.0",
 ]
 
 [[package]]
@@ -781,31 +836,30 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "form_urlencoded"
-version = "1.0.1"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5fc25a87fa4fd2094bffb06925852034d90a17f0d1e05197d4956d3555752191"
+checksum = "a9c384f161156f5260c24a097c56119f9be8c798586aecc13afbcbe7b7e26bf8"
 dependencies = [
- "matches",
  "percent-encoding",
 ]
 
 [[package]]
 name = "fragile"
-version = "1.2.1"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "85dcb89d2b10c5f6133de2efd8c11959ce9dbb46a2f7a4cab208c4eeda6ce1ab"
+checksum = "6c2141d6d6c8512188a7891b4b01590a45f6dac67afb4f255c4124dbb86d4eaa"
 
 [[package]]
 name = "futures-core"
-version = "0.3.23"
+version = "0.3.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2acedae88d38235936c3922476b10fced7b2b68136f5e3c03c2d5be348a1115"
+checksum = "04909a7a7e4633ae6c4a9ab280aeb86da1236243a77b694a49eacd659a4bd3ac"
 
 [[package]]
 name = "futures-io"
-version = "0.3.23"
+version = "0.3.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "93a66fc6d035a26a3ae255a6d2bca35eda63ae4c5512bef54449113f7a1228e5"
+checksum = "00f5fb52a06bdcadeb54e8d3671f8888a39697dcb0b81b23b55174030427f4eb"
 
 [[package]]
 name = "futures-lite"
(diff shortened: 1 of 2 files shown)


vladimyr:
> The idea here is to search for Starship on the PATH when printing out its command path during shell command generation. This is particularly helpful for people using Homebrew, which deletes the old version of any software it installs upon upgrade. Prior to this change, _starship_ would print out the fully resolved path to the binary on disk, and not the symlink that Homebrew creates and is generally available on the path.

While I understand the problem you are trying to solve I'm not sure this is the right way. I'm not sure that expecting starship to be on the `PATH` is a safe assumption? 🤔 In this case, it is enforced by Homebrew and in general, install script warns you if target dir is not on the `PATH` but nothing really prevents you to install it wherever you like and invoke init from there?
Writing this while using starship directly from my dev workspace located at `~/workspace/github.com/starship/starship` 🤷

nirvdrum:
@vladimyr This doesn't force _starship_ to be on the PATH. But, it will favor a binary on the PATH over an arbitrary binary. I've tested it out both with with and without an entry in the PATH. The only problem I could see is if it you custom built Starship and had it installed on the PATH somewhere, but updated your shell init scripts to use the custom built version. Now the `init` command will use use the path to _starship_ found in the PATH.

If you've tried out the branch and see a problem, please let me know.

vladimyr:
TBH, yes I failed to realize this is just favoring executable found on `PATH` and providing current binary path as a fallback. Still, the exact same thing you just described happened to me because I do have a custom build I use and one available on `PATH` installed by Homebrew. Yes that is my weird development setup and doesn't represent real-world use-case so we can ignore it but while playing with your branch I figured out something else:
```
starship on  master [$?] is 📦 v0.42.0 via 🦀 v1.42.0 
❯ $(realpath $(which starship)) init zsh
source <("/usr/local/Cellar/starship/0.42.0/bin/starship" init zsh --print-full-init)%                                  

starship on  master [$?] is 📦 v0.42.0 via 🦀 v1.42.0 
❯ $(brew --prefix)/bin/starship init zsh
source <("/usr/local/bin/starship" init zsh --print-full-init)%  
```
If you leverage `$(brew --prefix)` it will solve your problem without any changes required on starship end because `/usr/local/bin/starship` will always point to currently installed/brewed version, right?
Maybe it is better to update the docs for Homebrew users instead of introducing a special case to binary resolving algorithm, what do you think?

vladimyr:
Actually, even this seems to work:
```
starship on  master [$?] is 📦 v0.42.0 via 🦀 v1.42.0 
➜ which starship
/usr/local/bin/starship

starship on  master [$?] is 📦 v0.42.0 via 🦀 v1.42.0 
➜ starship init zsh
source <("/usr/local/bin/starship" init zsh --print-full-init)% 
```

Regardless of how I invoke it, starship will always print the path to symlink instead of the actual binary located inside Homebrew's cellar. How did you manage to get a resolved path? 😕

nirvdrum:
I'm using Linuxbrew, yes, but that doesn't really matter here.

The problem here is the output of the `init` command:

```
❯ which starship
/home/linuxbrew/.linuxbrew/bin/starship

dev/bin/graalvm-ee-java11-20.1.0 on ☁️  cloudtruth(us-east-1) 
❯ starship init fish
source ("/home/linuxbrew/.linuxbrew/Cellar/starship/0.42.0/bin/starship" init fish --print-full-init | psub)
```

Since `env::current_exe` always resolves to the binary's path on disk, it always returns the versioned binary from the Cellar. If you have a shell open with that version of Starship and then you delete the binary, the shell breaks. By default, Homebrew deletes the old version of software after installing a new version. Normally, I think this works out okay because neither macOS nor Linux really care if you delete the binary while its running. The problem though is the replacement version is in a completely different directory. If the already running _starship_ binary needs to use its `path_to_starship` for any reason, it fails, because the binary no longer exists on disk.

I can't say for certain, but I suspect most other software doesn't have this problem because of one of two reasons:

1. They check _argv_ to get the path to the command and just re-use that. With C, at least, that doesn't include a `readlink` call.

2. They assume the command is already on the PATH and invoke it that way.

Maybe this is a problem with Homebrew, but Starship is the only piece of software I've ever run into this problem with.

vladimyr:
Makes sense, thanks for the elaborate explanation and sorry for jumping out of the clear sky I should've really read the background story first. Tho I'm still wondering can use of `$(brew --prefix)` help mitigate it? What is the output of:
```
$(brew --prefix)/bin/starship init fish
```

chipbuster:
> Since env::current_exe always resolves to the binary's path on disk, it always returns the versioned binary from the Cellar.

Yes, this is our primary problem. The [behavior of current_exe is platform-specific](https://doc.rust-lang.org/std/env/fn.current_exe.html). Unfortunately, I doubt there's much we can do to change this--the PR that weakened its behavior from its previous version specifically noted that, because of platform-specific issues, it is impossible to provide a stronger guarantee than "it'll give you *some* path you can use to resolve the binary."

nirvdrum:
I had to change the command a bit because I use the fish shell and command substitutions are a bit different:

```
~ via 💎 v2.6.6
❯ set PREFIX (brew --prefix)

~ via 💎 v2.6.6
❯ echo $PREFIX
/home/linuxbrew/.linuxbrew

~ via 💎 v2.6.6
❯ $PREFIX/bin/starship init fish
source ("/home/linuxbrew/.linuxbrew/Cellar/starship/0.42.0/bin/starship" init fish --print-full-init | psub)           
```

vladimyr:
And it again followed symlink to the actual location 🙃

nirvdrum:
The docs for [env::current_exe](https://doc.rust-lang.org/std/env/fn.current_exe.html) indicate it's platform-dependent how symlinks are resolved:

> If the executable was invoked through a symbolic link, some platforms will return the path of the symbolic link and other platforms will return the path of the symbolic link’s target.

I don't think it's something we can rely on.

vladimyr:
> I can't say for certain, but I suspect most other software doesn't have this problem because of one of two reasons:
> 
> 1. They check _argv_ to get the path to the command and just re-use that. With C, at least, that doesn't include a `readlink` call.

And what happens if you use `argv0`? I swapped `env::current_exe` with `args[0]` and got logical path but I'm on a mac so that doesn't really count.

chipbuster:
@vladimyr Won't work for people on nix, see #220 for original issue and #224 for details.

I think we simply *cannot* attempt to rely on the behavior of `current_exe`: even if it correctly resolves the symlink now, on a particular platform, it may change to resolving the physical path at a later date or on a different platform, esp. once we start releasing builds for other architectures. If the `$(brew --prefix)` trick works for some people, that's great, but that is in no way a patch that's guaranteed to continue working.

That being said, @vladimyr's scenario of having one starship in the path but attempting to invoke a different one is a valid concern. We could even take it a step further and ask what happens if we have two different programs named `starship`, and the one in `PATH` is meant to do something else. In this case, `/path/to/my/prompt/tool/starship init` would appear to work, but then the wrong program altogether gets called in the shell.

@nirvdrum Do you think we could get around it by doing the following? Examine the 0-th argument of `std::env::args`:
- If it is just `starship` (or something similar), we use the current PR's PATH lookups to determine what the correct path to binary is. 
- If it is given as an absolute path, we assume that the user knows exactly what they're doing and just use that path instead.
- If it is given as a relative path.....(???)

I don't know what the correct behavior is for a relative path, unfortunately. I don't even know what this would signify, but there's at least one example of it in #220.

nirvdrum:
I'll have to dig through Starship and see how the path string is being used. argv[0] might be a relative path. E.g., if I read argv and run with `cargo run` I get:

```
starship on  find-starship-on-path is 📦 v0.42.0 via 🦀 v1.44.0
❯ cargo run init fish
source ("target/debug/starship" init fish --print-full-init | psub)
```

I can't put that into `~/.config/fish/conf.d/` because the path only makes sense relative to where the command was invoked. We could maybe check if the path is relative and if so, append it to `env::current_dir`.

I'd wager the case where someone has one version on their PATH and wants to both invoke and load a different version in their shell is probably quite small. I'd think most people would just adjust their PATH at that point. But, I can appreciate it might affect some people.

nirvdrum:
Ugh, relying on `env::current_dir` is probably just going to get us back into the same mess where the binary is now resolved in its versioned directory.

chipbuster:
@nirvdrum Yeah, only the shell would know what the logical path was...maybe we could try reading `$PWD` from the environment and then fall back to `current_dir()`?

vladimyr:
> Ugh, relying on `env::current_dir` is probably just going to get us back into the same mess where the binary is now resolved in its versioned directory.

Umm, not really because only the binary is symlinked, not its parent dir. FWIW according to POSIX spec `getcwd` should always return fully resolved path but in this case that is irrelevant because you are going to invoke starship from `$PREFIX/bin` and that is not symlink.

> @nirvdrum Yeah, only the shell would know what the logical path was...maybe we could try reading `$PWD` from the environment and then fall back to `current_dir()`?
> 
> EDIT: [Reading PWD is the approach that the GNU Coreutil `pwd` uses when it is asked for a logical path](https://github.com/coreutils/coreutils/blob/master/src/pwd.c?rgh-link-date=2020-06-19T00%3A53%3A26Z#L297) so it at least doesn't seem insane.

That is indeed good workaround but it is shell specific https://stackoverflow.com/a/1545838

nirvdrum:
Setting aside duplication, which I'd clean up, this is basically what I was suggesting in my latest comment:

```
fn path_to_starship() -> io::Result<String> {
    let exe_name = option_env!("CARGO_PKG_NAME").unwrap_or("starship");
    let argv0 = env::args().nth(0).unwrap_or(String::from(""));

    if (argv0.eq(exe_name)) {
        match (which(exe_name)) {
            Ok(path) => {
                let current_exe = path
                    .to_str()
                    .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
                    .to_string();
                Ok(current_exe)
            }
            Err(error) => {
                let current_exe = env::current_exe()?
                    .to_str()
                    .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
                    .to_string();
                Ok(current_exe)
            }
        }
    } else if (Path::new(&argv0).is_absolute()) {
        Ok(argv0)
    } else {
        let current_exe = env::current_exe()?
            .to_str()
            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
            .to_string();
        Ok(current_exe)
    }
}
```

The basic idea is:
* If the command is invoked bare, look it up on the PATH
* If the command is fully qualified, use it as is
* In all other cases, use `env::current_exe`

The PATH case is the one this PR already addresses. The fully qualified path case allows `/home/linuxbrew/.linuxbrew/bin/starship init fish` to return the Homebrew symlink without a `readlink` on it. The catchall cases handles relative calls to _starship_ (e.g., `target/debug/starship`) to resolve to the canonicalized path of the binary.

I could use `fs::canonicalize` on `argv0` instead of `env::current_exe`.

If that doesn't sound good, I'll look into `PWD`.

chipbuster:
> The catchall cases handles relative calls to starship (e.g., target/debug/starship) to resolve to the canonicalized path of the binary.

**D'oh.** Thank you, I wasn't thinking correctly. So the only error cases now are if the PWD or the `argv[0]` contain symlinks and the referent of those symlinks are moved somewhere else, is that correct?

If so, I'm inclined to ignore that error case until we know for sure that that's a use case, since it seems like an incredibly obscure use case that could probably be handled by the user.

chipbuster:
Meant to address this comment earlier but got caught up in the discussion:

> I'm not familiar with conventional commits and don't know whether this is a feature or a fix.

That makes two of us :grin:. Given that the correct behavior is (evidently) non-trivial, let's call it a feature.

nirvdrum:
> **D'oh.** Thank you, I wasn't thinking correctly. So the only error cases now are if the PWD or the `argv[0]` contain symlinks and the referent of those symlinks are moved somewhere else, is that correct?

Yeah, I overlooked that case. I could add a call to check if the path still exists and fall back to `env::current_exe` if it doesn't. You'll still have a corner case where the path gets deleted after the existence check, since none of this is atomic. But, I really hope that's a virtually non-existent problem. That would look like the following (again, subject to clean up):

```rust
fn path_to_starship() -> io::Result<String> {
    let exe_name = option_env!("CARGO_PKG_NAME").unwrap_or("starship");
    let argv0 = env::args().nth(0).unwrap_or(String::from(""));

    if (argv0.eq(exe_name)) {
        match (which(exe_name)) {
            Ok(path) => {
                let current_exe = path
                    .to_str()
                    .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
                    .to_string();
                Ok(current_exe)
            }
            Err(error) => {
                let current_exe = env::current_exe()?
                    .to_str()
                    .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
                    .to_string();
                Ok(current_exe)
            }
        }
    } else {
        let argv0_path = Path::new(&argv0);

        if (argv0_path.exists() && argv0_path.is_absolute()) {
            Ok(argv0)
        } else {
            let current_exe = env::current_exe()?
                .to_str()
                .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
                .to_string();
            Ok(current_exe)
        }
    }
}
```

If we wanted to relax the requirement of printing out a full path, we could drop _which_ and just let the user's PATH deal with it. That'd amount to changing the output to something like:

```
source (starship init fish --print-full-init | psub)
```

If we need to invoke it ourselves, `process::Command` will search the `PATH` for non-fully qualified command names.

vladimyr:
What do you think about this:
```rust
fn path_to_starship() -> io::Result<String> {
    let argv0 = env::args().nth(0).unwrap_or(String::from(crate_name!()));

    let path = Path::new(&argv0);
    let components = path.components().collect::<Vec<_>>();

    let resolved_path;
    if components.len() == 1 {
        resolved_path = which(argv0).or_else(|_| env::current_exe())?
    } else if components[0] == Component::CurDir {
        resolved_path = env::current_dir()?.as_path().join(argv0).to_path_buf();
    } else {
        resolved_path = path.to_path_buf();
    }

    let current_exe = fs::canonicalize(resolved_path)?
        .to_str()
        .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
        .to_string();

    Ok(current_exe)
}
```

It is essentially cleaned up version of @nirvdrum did. 

> The basic idea is:
> 
> * If the command is invoked bare, look it up on the PATH
> * If the command is fully qualified, use it as is
> * In all other cases, use `env::current_exe`

I implemented in the following way:

1. bare command (_path_ consists of a single component) - look it up on the `PATH` and fallback to `env::current_exe`
2. relative path - resolve it using `env::current_dir` (resolves symlinks, but that doesn't matter because the binary is symlinked and parent dir probably isn't)
3. absolute path - use it as is

with path normalization being applied on the resolved path.

vladimyr:
> > **D'oh.** Thank you, I wasn't thinking correctly. So the only error cases now are if the PWD or the `argv[0]` contain symlinks and the referent of those symlinks are moved somewhere else, is that correct?
> 
> Yeah, I overlooked that case. I could add a call to check if the path still exists and fall back to `env::current_exe` if it doesn't. You'll still have a corner case where the path gets deleted after the existence check, since none of this is atomic. But, I really hope that's a virtually non-existent problem.

I believe we should ignore that because isn't that case of pure misconfiguration from the user end? I mean the user should be responsible to make sure that symlinks actually point to the starship binary?

chipbuster:
> I could add a call to check if the path still exists and fall back to `env::current_exe` if it doesn't. You'll still have a corner case where the path gets deleted after the existence check, since none of this is atomic. But, I really hope that's a virtually non-existent problem.

You know, I think we might be overthinking this. We want to support the homebrew behaviour without additional config, since that's a really common use case. We want to retain support for NixOS (#220) because we've had support for that for a long time and don't want to break it.

But for the case where symlinks exist in the current `PWD` and the user inputs a relative path, or where they're doing some really weird shenanigans with links, why don't we just have them tell us where the executable is? We could have the init code read an envar like `STARSHIP_EXE_PATH` (or something) and then just have that override the autodetected binary location. That way, we don't have to worry about symlink components in the path or anything like that. I agree that a lot of the stuff we've talked about here seems unlikely, and it's going to complicate the code a fair bit, so foisting the responsibility off on the user seems like the easiest decision.

nirvdrum:
@chipbuster I've read that a couple times and I'm not entirely sure what you're proposing, so please forgive me if my reply here is not in conflict with what you're suggesting.

While I'm certainly biased on this one, I'd strongly recommend "just working" for the Homebrew case. I wouldn't want to require `STARSHIP_EXE_PATH` for that case because people won't learn there's a problem until they try to upgrade via Homebrew. At that point, the frustration has already set in. For my part, I stopped updating Starship before eventually finding Homebrew's `HOMEBREW_NO_INSTALL_CLEANUP` to avoid cleaning upon upgrade. I was tempted to dropping Starship because both solutions were impractical.

I can see favoring a `STARSHIP_EXE_PATH` if one is set. I think most people expect setting the `PATH` to be sufficient, but having a special env var has some precedent with things like `JAVA_HOME` and `GOROOT`. But, I also don't know anyone happy with those solutions :-/

I'm ambivalent on the corner case support here. I think the changes are fairly localized and the code is straightforward to read, so I don't have any problem with supporting them. But, I'm also not a core maintainer of this project, so I can't really speak to the maintenance burden.

nirvdrum:
@vladimyr I like the cleanups you made, but calling `fs::canonicalize` resolves symlinks, which then breaks the Homebrew case again. I don't think we want to resolve symlinks.

vladimyr:
> @vladimyr I like the cleanups you made, but calling `fs::canonicalize` resolves symlinks, which then breaks the Homebrew case again. I don't think we want to resolve symlinks.

True that... Quick googling reveals this gem https://github.com/rust-lang/rust/issues/59117 😥

Well, I guess we are good if you just strip canonicalize out?

chipbuster:
> While I'm certainly biased on this one, I'd strongly recommend "just working" for the Homebrew case.

Yes, I definitely agree with this.

Let me try to clarify what I was thinking. I want to achieve three main goals:
   1. Support the Homebrew use case without additional configuration on the user's part.
   2. Keep compatibility with the nixOS use case presented in #220.
   3. Allow the user to specify a specific path to use, to allow for overriding the default behavior if needed.

The idea I had was that we could achieve (3) by using an environment variable. On further reflection, this is only needed if the user wants to specify their starship binary as a `PWD`-relative path without resolving symlinks--an incredibly unlikely case.

I think the logic you posted in [this comment](https://github.com/starship/starship/pull/1344#issuecomment-646382896) should be sufficient for what we need. Let's not worry too much about relative paths that aren't unqualified `starship`.

nirvdrum:
I need to run through some cases, but I suspect `else if components[0] == Component::CurDir` isn't sufficient if using a path relative to the parent (e.g., `../some/path`). You may to do the `is_absolute` check like I was originally. Unless I'm mistaken, relative paths should always be relative to the current directory.

vladimyr:
> I need to run through some cases, but I suspect `else if components[0] == Component::CurDir` isn't sufficient if using a path relative to the parent (e.g., `../some/path`). You may to do the `is_absolute` check like I was originally. Unless I'm mistaken, relative paths should always be relative to the current directory.

Oh, that is a very good point 👍 My matching wasn't exhaustive at all https://doc.rust-lang.org/std/path/enum.Component.html

vladimyr:
@chipbuster @nirvdrum This might be the stupidest question in the whole thread but maybe it isn't... If this is 2 phase init why don't we just take verbatim argv0 of the first phase and just pass it to the second phase? 🤔 

It essentially reduces `path_to_starship` to:
```rust
fn path_to_starship() -> String {
    env::args().next().unwrap_or_else(|| String::from(crate_name!()))
}
```
That should work in the Homebrew case because no path modifications would happen. You would pass `$(brew --prefix)/bin/starship` and you would get it back - with expanded Homebrew prefix, but without resolving symlinks.
I believe the same applies to `nixOS` and essentially eliminates the need for `STARSHIP_EXE_PATH` because the user just needs to make sure to use it as an invocation path instead:
```bash
# hypothetical STARSHIP_EXE_PATH
eval $($STARSHIP_EXE_PATH init bash)
# starship on PATH
eval $(starship init bash)
# Homebrew
eval $($(brew-prefix)/bin/starship init bash)
# nixOS
eval $(${pkgs.starship}/bin/starship init bash)
```
Do note that I never really used nix so I might be wrong about last one 🤷

vladimyr:
@chipbuster On a different note what was the point of https://github.com/starship/starship/blob/ee3c0ec/src/init/mod.rs#L36 Escaping quotes inside the path? 😕

vladimyr:
Also, the background behind https://github.com/starship/starship/pull/1344#issuecomment-646904650 is that I believe we are heading in the wrong direction. I can't find any example out there doing anything similar. This reminded me of [`process.execPath`](https://nodejs.org/api/process.html#process_process_execpath) that `node` implements but quick checking revealed the following:
```
~
➜ node -e 'console.log(process.argv0 + "\n" + process.execPath)'
node
/Users/vladimyr/.fnm/node-versions/v10.17.0/installation/bin/node

~
➜ `which node` -e 'console.log(process.argv0 + "\n" + process.execPath)'
/var/folders/1r/tl9g2ytj4dq1w036t_0p0vr40000gn/T/fnm-shell-110968/bin/node
/Users/vladimyr/.fnm/node-versions/v10.17.0/installation/bin/node

```
They use [`uv_exepath`](http://docs.libuv.org/en/v1.x/misc.html#c.uv_exepath) and they are going pretty much opposite direction. `process.execPath` always returns the physical path of the binary.  That is a step ahead of `env::current_exe` because it is backed by a ton of platform-specific code:
- [*nix](https://github.com/libuv/libuv/blob/e4087dedf837f415056a45a838f639a3d9dc3ced/src/unix/procfs-exepath.c#L28-L45) implementation
- implementation for [BSDs](https://github.com/libuv/libuv/blob/3d7136639a39db46bc4a9074922559a564e49514/src/unix/freebsd.c#L60-L113)
- [Darwin](https://github.com/libuv/libuv/blob/2ab3dc1567d67924b634be6f478d25f4528ce574/src/unix/darwin.c#L63-L92) implementation 
- [Windows](https://github.com/libuv/libuv/blob/ff3ab317a2082fb450818b892156840af5e0cc35/src/win/util.c#L94-L149) implementation

And that raised a flag inside my head because:
1. not only that is demonstrably hard to go the opposite way but reading through their implementation it is easy to figure out that changing direction likely won't reduce complexity
2. even if you decide to go through all the required steps I'm not even sure it is always possible
3. seems nobody does that

And then I went to SO to do quick sanity check just to reveal another one:
https://stackoverflow.com/questions/21116519/get-current-logical-path-of-executable#comment31773535_21116519

Seems it is pretty much a doomed effort so I started asking myself why do we even need it...

nirvdrum:
@vladimyr I'd need to take a deeper look through more of Starship. My guess is if `starship` just forked, it wouldn't be a problem. But, it's trying to invoke the binary by path and doesn't function any longer once deleted or moved. I suspect most programs don't have a similar problem because they either fork or expect to be on the `PATH` and a replacement binary also available on the `PATH`.

vladimyr:
> But, it's trying to invoke the binary by path and doesn't function any longer once deleted or moved.

True that, but if you pass symlink that won't get deleted but updated like in Homebrew's case you just solved that?

chipbuster:
> why don't we take verbatim argv0 of the first phase and just pass it to the second phase?

Because this does not work for the case presented in #220.

> On a different note what was the point of https://github.com/starship/starship/blob/ee3c0ec/src/init/mod.rs#L36 

Don't remember. You can pull the commit where the initscripts refactor occured, get its parent, then run a git blame to find out.

vladimyr:
> > why don't we take verbatim argv0 of the first phase and just pass it to the second phase?
> 
> Because this does not work for the case presented in #220.

From what I read there I don't think that is actually correct. Back then pure invocations were hardcoded inside second stage (aka `starship ...`) which assumed availability of starship on `$PATH`. I don't see that my proposal was ever really implemented nor tested on nix?

vladimyr:
> > On a different note what was the point of [/src/init/mod.rs@`ee3c0ec`#L36](https://github.com/starship/starship/blob/ee3c0ec/src/init/mod.rs?rgh-link-date=2020-06-20T01%3A44%3A36Z#L36)
> 
> Don't remember. You can pull the commit where the initscripts refactor occured, get its parent, then run a git blame to find out.

I did that: https://github.com/starship/starship/commit/923855320baf4f931dd83957441f0af120c218ee but didn't find anything useful. https://github.com/starship/starship/pull/224#issuecomment-524051715 gives some insight. It most certainly looks like shell escaping but I'm still wondering why would you want to do that anyway 🤔

nirvdrum:
> True that but if you pass symlink that won't get deleted but updated like in Homebrew's case you just solved that?
> 
> I mean if I invoke it using later there will be troubles but if I pass the former and it gets reused in second stage it should work?

Yes. That was part of the point of my [follow-up solution](https://github.com/starship/starship/pull/1344#issuecomment-646391204). If argv0 is an absolute path, just use it without canonicalizing. You run into a problem where argv0 might not actually be the path to the command invoked, and that may end up being a security issue (no one is running this as _root_, I hope)? The plan then was to just fall over to using `env::current_exe`.

We can adjust the precedence rules and/or drop the `PATH` search. To the best of my knowledge, the three use cases we care about are:

* Starship is not available on the `PATH` (nix case)
* Starship is installed in a versioned directory and invoked via symlink (Homebrew case)
* Starship is installed into some location on the `PATH`, but a completely different version is invoked directly and we want that latter version to emit shell code to load that will use that latter binary (the development team use case)

The Homebrew case dovetails with invoking _starship_ from the `PATH`. All the shell setup commands in the README invoke _starship_ from the path (e.g., `
eval "$(starship init bash)"`), so we can't rely on argv0 being a path to the _starship_ binary. That's why I pulled in the `which` crate, but I structured it such that the `PATH` search is only performed if argv0 is exactly the string "starship".

vladimyr:
>You run into a problem where argv0 might not actually be the path to the command invoked, and that may end up being a security issue (no one is running this as _root_, I hope)? The plan then was to just fall over to using `env::current_exe`.

In which case argv0 isn't a path or does not resolve to a path to actual binary? I don't understand potential security concern around it 🤔

> All the shell setup commands in the README invoke _starship_ from the path (e.g., `
> eval "$(starship init bash)"`), so we can't rely on argv0 being a path to the _starship_ binary. That's why I pulled in the `which` crate, but I structured it such that the `PATH` search is only performed if argv0 is exactly the string "starship".

And that is exactly my point. I would rather go and update the documentation to tell users ok if you installed it using Homebrew or nix use following symlink instead of doing smart guessing inside the code. It seems logical to give users full control instead of trying figuring out the correct second stage invocation for them?

vladimyr:
And actually, let's imagine for a second that docs stay the same but code changes to just pass original argv0. In Homebrew's case that would mean users would use starship as a path which ultimately triggers search for it on the PATH. In my specific case, actual resolved value then becomes /usr/local/bin/starship that points to versioned binary inside the cellar. What gets passed to the second stage is just starship triggering the same process again and that is fine because you'll never have versioned binary directly inside your path and symlink is guaranteed to always point to it regardless of underlying changes/updates.

nirvdrum:
@vladimyr I'm sorry, but I've lost track of what you're arguing for or against. It sounds like the only part you don't like is searching for _starship_ on the `PATH`, which both seems like one of the simplest cases and which I said we could eliminate if we want. I think you'll still end up having a lot of people with problems that way. There will be people that have already set up their shell command using the docs that won't know that invoking from the `PATH` is not supported. It also seems like a confusing issue to explain to people that _starship_ may not work the way people expect when invoking from the `PATH`.

I think all the tricky cases here have to do with invoking not from the `PATH`. Predominately, invoking with a relative path to the binary. For me, handling the invocation from the `PATH` is sufficient. If you just rely on argv0 having the path, you run into potential problems with relative paths and the current working directory. But, I suspect in the common case of the user just slapping `starship init <shell>` into a shell script, this won't be much of a problem because the current working directory will not have change.

The other potential problem is that argv0 isn't actually the command used to invoke _starship_, per the StackOverflow post you linked. I don't think of it as much of a problem because the use case here is the shell is the one invoking the command. But, if we need to worry about some arbitrary program invoking _starship_ and feeding in garbage data into argv0 and then the user blindly executing that within their shell init, then we can't rely on argv0. I can see how that would be problematic, but if someone can already execute arbitrary programs on the user's system, they can already do whatever damage they'd like. I think we could mitigate that problem if we wanted to, though.

I'll update my PR with what we've discussed so far and that should give us a way to re-center the discussion. It's gotten difficult to keep track of with code appearing in different comments and such.

vladimyr:
> I'm sorry, but I've lost track of what you're arguing for or against.

It was bound to happen 🙃 That is why I opened another draft so you can comment and test my proposal: https://github.com/starship/starship/pull/1365 without trying to guess what actually lies behind my messy explanations 🙃 

> It sounds like the only part you don't like is searching for _starship_ on the `PATH`, which both seems like one of the simplest cases and which I said we could eliminate if we want. I think you'll still end up having a lot of people with problems that way. There will be people that have already set up their shell command using the docs that won't know that invoking from the `PATH` is not supported. It also seems like a confusing issue to explain to people that _starship_ may not work the way people expect when invoking from the `PATH`.

It is not that I don't like it but I believe we don't need it. Users execute `starship` binary twice. The first time it is by direct request and second is pseudo-forking. In the first case, os would find binary on the path and if we just pass that to the second phase, literal string: `starship` os would find it again, no need for us to fully resolve it?
So for those who already have `$(starship init bash)` in their rc file (and `starship` binary available on `PATH`), it will simply continue to work.

> But, I suspect in the common case of the user just slapping `starship init <shell>` into a shell script, this won't be much of a problem because the current working directory will not have change.

Exactly, that makes it a non-issue.

> The other potential problem is that argv0 isn't actually the command used to invoke _starship_, per the StackOverflow post you linked. I don't think of it as much of a problem because the use case here is the shell is the one invoking the command. But, if we need to worry about some arbitrary program invoking _starship_ and feeding in garbage data into argv0 and then the user blindly executing that within their shell init, then we can't rely on argv0. I can see how that would be problematic, but if someone can already execute arbitrary programs on the user's system, they can already do whatever damage they'd like. I think we could mitigate that problem if we wanted to, though.

Agree with everything you just said and it again seems like a non-issue because we shouldn't really worry about someone doing _crazy_ things on their own.

> I'll update my PR with what we've discussed so far and that should give us a way to re-center the discussion. It's gotten difficult to keep track of with code appearing in different comments and such.

At this point, it is very much needed 😄 Looking forward to discussing it 👍

davidkna (approved):

PR #1360, opened 2020-06-20
ci: Auto publish for the AUR

ATiltedTree:
#### Description

This adds a action step that automatically bumps the AUR package of starship (and starship-bin) to the latest version.
But it needs some special attention before that. Which means one of these two options:
1. The current maintainer sets the secret `AUR_SSH_PRIVATE_KEY` to his AUR ssh private key. And the TBD messages should be replaced to his.
2. The current maintainer transfers the packages to a stub account (ex. starship-bot) and an extra ssh private key is generated for it. Also replacing TBD's with something sensible.

#### Motivation and Context

I like to have new versions as fast as possible and the current maintainer does not have to worry about version bumps.

#### Checklist:


chipbuster:
@ATiltedTree We already have different maintainers for `starship/starship-bin` and `starship-git` so I'd rather not confuse that further. Would it be an option to add a co-maintainer stub account for `starship-bin` and then use a private key for that, or does it have to be the primary maintainer's key?

ATiltedTree:
No, adding co-maintainer would be enough to push new versions.

chipbuster:
Sorry for the slow progress here--I've been waylaid by other work and have had very little time to look at anything starship recently.

I've recently created the starship-bot account on AUR and added it into our secrets store. Next I'll need to add the keypair and register it with AUR, then update this PR with the appropriate values.

Question: is there a way to do tests in an Arch chroot with the current state GH Actions? The starship package is still frozen at 0.42.0 because of failures in `cargo test` in a chroot (occurs with both v0.43.0 and v0.44.0), and I don't particularly like the idea of pushing broken builds into the AUR, because they tend to break AUR helpers (most helpers, when they fail to build a package, will abort an entire upgrade operation, meaning that you have to baby the helper or come back a few hours later to find that nothing happened)

ATiltedTree:
The action I choose tries to built the package before it pushes it and when it failes it aborts.
Also which test are failing? I might be able to help there.

chipbuster:
Got it.

I'm seeing failures in python::with_active_venv and python::with_virtual_env. You can see the PKGBUILD and how I do chroot setup in this tarball (it's possible I'm doing something wrong with the chroot, I'm not sure):

https://www.dropbox.com/s/6nk89i3tvb5e7te/starship_aur.tar.gz?dl=1

ATiltedTree:
I was able to fix the problem. The solution was to add `python` to the `checkdepends`.
And the underlying problem is that we have `utils::exec_cmd` which is different if we run test or not but when integration test are run it actually seems to use the non test `utils::exec_cmd` as we hardcode the path our binary to `./target/debug/starship` but that is not the binary cargo builds for unit tests (its actually this binary `./target/debug/deps/starship-XXXXXXXXXXXXXXX` the `X` being random characters) which is why it uses the non test variant of `utils::exec_cmd`.

chipbuster:
Nice catch! Can confirm that this fixed things for me.

Shouldn't that be considered a filesystem-dependent test and `#[ignore]`d?

It also looks like the action relies on `pkgver` and `sha256sums` being the only variables that need to be changed, so I'll take a look at the current PKGBUILD and make sure that's the case.

(Somewhat related: at some point, we need to go through our testsuite and make sure that appropriate tests are actually being ignored--I tried to test the package outside of chroot at some point and got a bunch of stuff dropped in my home directory...)

ATiltedTree:
> Shouldn't that be considered a filesystem-dependent test and #[ignore]d?

It definitely should be ignored but because we have a mock command for it the person who added it probably thought it would get mocked as well (which it unfortunately does not).

> (Somewhat related: at some point, we need to go through our testsuite and make sure that appropriate tests are actually being ignored--I tried to test the package outside of chroot at some point and got a bunch of stuff dropped in my home directory...)

In the end it should be our goal that no test is ignored meaning that all tempdirs and files are actually closed and that we finally should have proper env mocking.
Sadly some modules (e.g. dotnet) actually require the cli to do things and not only to print it's version which we can't mock.

Also I discovered that `tempfile` now uses `remove_dir_all` which means we only have to properly close them and not manually remove them.

chipbuster:
@ATiltedTree Got the substituted values added. I've also:

- Added user starship-bot as a co-maintainer on both the `starship` and `starship-bin` AUR packages
- Added a public key to the AUR account
- Added the corresponding private key under `AUR_SSH_PRIVATE_KEY` on this project's secrets

From what I understand, that should be everything we need to allow this action to fire. Is there anything I'm missing? Is there a way to test it before the next major release, or will we just have to wait and see?

.github/workflows/deploy.yml
```diff
@@ -143,3 +143,22 @@ jobs:
...
+          package_name: starship
+          commit_username: "Starship Bot"
+          commit_email: starship.bot.noreply@gmail.com
+          ssh_private_key: ${{ secrets.AUR_SSH_PRIVATE_KEY }}
+      - name: Build | Publish AUR package (-bin version)
+        uses: guumaster/aur-publish-docker-action@v1
+        with:
+          package_name: starship-bin
+          commit_username: "Starship Bot"
+          commit_email: starship.bot.noreply@gmail.com
```
chipbuster, line 163: Is it a problem if the email isn't monitored? I'm going out on a limb and assuming that very few people use the committer email in the AUR git repo as information on how to contact the maintainer.

ATiltedTree: They shouldn't. If they want to contact us they should use the comments in the AUR web ui or create an issue here on GitHub.

ATiltedTree:
We could just basically build the Docker container from the action and run the exact same commands (but obviously not running the git push).
If anything fails I could fork the action and fix the problems that we encounter.

chipbuster:
Seems the CI has randomly failed in MacOS this time. Strange.

It was randomly failing tests in Ubuntu before this--not sure what's going on today.

ATiltedTree:
AFAIK macos testing on GitHub actions is still a bit janky. Looking at https://github.com/actions/virtual-environments/issues there are a some that report weird behavior with macos. I think we the macos failure was just janky GitHub actions behavior and the Linux tests were related to our race conditions.

chipbuster:
Alright. Is this ready to come out of draft?

ATiltedTree:
After a bit of testing with the other action it turns out that they forgot the `-s` `makepkg` flag meaning download all required dependencies.
I forked it, fixed the issue and added a fix from some other fork that allows for multiple source files.
It *should* be good to go now. We will see if everything works properly on the next release!

PR #1364, opened 2020-06-20
fix: remove extra space for nix-shell segment

payasrelekar:
Now matches documentation

#### Motivation and Context
Removes unnecessary and unsymmetrical extra space

#### Checklist:
- [x] I have updated the documentation accordingly. --> Match present document


vladimyr:
> #### Motivation and Context
> Removes unnecessary and unsymmetrical extra space

Can you paste a screenshot that shows actual unsymmetrical output?

Those double spaces (here and in AWS module) are there for a reason. Without them, text that follows becomes smashed together with proceeding emoji symbol.

andytom:
If I remember correctly these two training spaces are to work around some issues with double width emojis, so this is intentional.

PR #1365, opened 2020-06-20
fix(init): pass verbatim invocation between phases

### LINKED ISSUE #910: Shell does not work after upgrading with Homebrew

vladimyr:
#### Description

Alter 2-phase init so the first phase takes original invocation and passes it to the second phase without any modifications being made.

#### Motivation and Context

Closes #910 

#### Checklist:

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -23,9 +23,13 @@ struct StarshipPath {
 }
 impl StarshipPath {
     fn init() -> io::Result<Self> {
-        Ok(Self {
-            native_path: env::current_exe()?,
-        })
+        let argv0_path = env::args().next().map(PathBuf::from);
+        let native_path = match argv0_path {
+            Some(path) => path,
+            None => env::current_exe()?,
+        };
+
+        Ok(Self { native_path })
     }
     fn str_path(&self) -> io::Result<&str> {
         let current_exe = self
@@ -91,6 +95,9 @@ impl StarshipPath {
 init code. The stub produces the main init script, then evaluates it with
 `source` and process substitution */
 pub fn init_stub(shell_name: &str) -> io::Result<()> {
+    let starship = path_to_starship();
+
+    log::debug!("Starship: {}", starship);
     log::debug!("Shell name: {}", shell_name);
 
     let shell_basename = Path::new(shell_name)
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -95,17 +95,15 @@ impl StarshipPath {
 init code. The stub produces the main init script, then evaluates it with
 `source` and process substitution */
 pub fn init_stub(shell_name: &str) -> io::Result<()> {
-    let starship = path_to_starship();
-
-    log::debug!("Starship: {}", starship);
-    log::debug!("Shell name: {}", shell_name);
+    let starship = StarshipPath::init()?;
 
     let shell_basename = Path::new(shell_name)
         .file_stem()
         .and_then(OsStr::to_str)
         .unwrap_or(shell_name);
 
-    let starship = StarshipPath::init()?;
+    log::debug!("Starship path: {:?}", starship.str_path());
+    log::debug!("Shell name: {}", shell_name);
 
     match shell_basename {
         "bash" => print!(


src/init/mod.rs
```diff
@@ -17,24 +17,24 @@ using whatever mechanism is available in the host shell--this two-phase solution
...
-fn path_to_starship() -> io::Result<String> {
-    let current_exe = env::current_exe()?
-        .to_str()
-        .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
-        .to_string();
-    Ok(current_exe)
+fn path_to_starship() -> String {
+    env::args()
+        .next()
+        .unwrap_or_else(|| String::from(crate_name!()))
```
nickwb, line 23: I'm curious, for the fallback why use `crate_name` rather than the current `env::current_exe` approach?

vladimyr: TBH it shouldn't ever happen unless someone explicitly messes with `argv0` like using `exec -a something starship...`. I put it there just for the sake of having avoiding forceful unwrapping.
Now when you ask me it might be better to use `env::current_exe` but it has both pros and cons compared to `crate_name`. `crate_name` is resolved during compile time which is obviously plus and it is deterministic regardless of platform. Yes, it might be the wrong thing to fall back onto because starship might not be available on the path. OTOH, `env::current_exe` is runtime call and completely unreliable (https://doc.rust-lang.org/std/env/fn.current_exe.html#platform-specific-behavior). So the real question is should we fall back to:
1. something that might be wrong but at least is always the same
2. something that might be wrong but is pretty much a moving target
3. should we ditch this altogether and proclaim it will never happen 🤔

chipbuster:
Unfortunately, in its current state, this does not work for a relative-path invocation of starship in the init command.

image

chipbuster:
Since #1344 has been merged, I'm going to close this.

PR #1383, opened 2020-06-24
docs: instruct users to install a nerd-font

matchai:
#### Description

I just went through the fresh setup process on a new computer and realized that the suggestion to install Fira Code left users without any of the Nerd Font icons we occasionally use.

This PR updates the docs to suggest folks install a Nerd Font.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@
 
 ### Prerequisites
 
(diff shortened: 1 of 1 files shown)


matchai:
@vladimyr - Agreed. I'd like to get around to making a per-platform step-by-step installation instructions on the website, and that would certainly be the best place to have platform-specific instructions like that.

PR #1385, opened 2020-06-24
feat(shlvl): Add shlvl module

### LINKED ISSUE #1286: SHLVL support
`starship` could show the current value of `$SHLVL` which shows the "nesting level" of current shell. Having this in prompt is very useful for anyone who lanches custom shells as direct child of the one you're working with (eg. sudo, pipenv). This module should allow for inhibiting the display based on the value - for example normal shell will have `SHLVL=1`, but if you're using `screen` or `tmux`, most of your shells witll have `SHLVL=2`. In zsh, this can be achieved via `%(3L.%L. .)` in the prompt substitution - display `SHLVL` when it's `>=3`.

`env_var` module can kind of do it, except:
* you can't make it conditional based on the value
* you can only have one `env_var`

daniel-white:
#### Description

Adds support for the `SHLVL` environment variable as new module.

#### Motivation and Context

Closes #1286 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1993,14 +1993,14 @@ set and greater than the specified threshold.
 
 ### Options
 
-| Variable    | Default                | Description                                      |
-| ----------- | ---------------------- | ------------------------------------------------ |
-| `symbol`    | `"↕️ "`                 | The symbol used before displaying the SHLVL.     |
-| `threshold` | `3`                    | Display threshold.                               |
-| `prefix`    | `""`                   | Prefix to display immediately the value.         |
-| `suffix`    | `""`                   | Suffix to display immediately after the value.   |
-| `style`     | `"bold dimmed yellow"` | The style for the module.                        |
-| `disabled`  | `true`                 | Disables the `shlvl` module.                     |
+| Variable    | Default         | Description                                      |
+| ----------- | --------------- | ------------------------------------------------ |
+| `symbol`    | `"↕️ "`          | The symbol used before displaying the SHLVL.     |
+| `threshold` | `2`             | Display threshold.                               |
+| `prefix`    | `""`            | Prefix to display immediately the value.         |
+| `suffix`    | `""`            | Suffix to display immediately after the value.   |
+| `style`     | `"bold yellow"` | The style for the module.                        |
+| `disabled`  | `true`          | Disables the `shlvl` module.                     |
 
 ### Example
 
@@ -2010,6 +2010,7 @@ set and greater than the specified threshold.
 [shlvl]
 disabled = false
 suffix = " levels down "
+threshold = 3
 ```
 
 ## Singularity
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -178,6 +178,7 @@ format = """
 
 $username\
 $hostname\
+$shlvl\
 $kubernetes\
 $directory\
 $git_branch\
@@ -1983,6 +1984,32 @@ The module will be shown if any of the following conditions are met:
 
 [rust]
 format = "via [⚙️ $version](red bold)"
+```
+
+## SHLVL
+
+The `shlvl` module shows the current SHLVL image, if it is set and greater than the 
+specified threshold.
+
+### Options
+
+| Variable    | Default                | Description                                      |
+| ----------- | ---------------------- | ------------------------------------------------ |
+| `symbol`    | `"↕️ "`                 | The symbol used before displaying the SHLVL.     |
+| `threshold` | `3`                    | Display threshold.                               |
+| `prefix`    | `""`                   | Prefix to display immediately the value.         |
+| `suffix`    | `""`                   | Suffix to display immediately after the value.   |
+| `style`     | `"bold dimmed yellow"` | The style for the module.                        |
+| `disabled`  | `true`                 | Disables the `shlvl` module.                     |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[shlvl]
+disabled = false
+suffix = " levels down "
 ```
 
 ## Singularity
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1988,8 +1988,8 @@ format = "via [⚙️ $version](red bold)"
 
 ## SHLVL
 
-The `shlvl` module shows the current SHLVL image, if it is set and greater than the 
-specified threshold.
+The `shlvl` module shows the current SHLVL ("shell level") environment variable, if it is
+set and greater than the specified threshold.
 
 ### Options
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1993,14 +1993,13 @@ set and greater than the specified threshold.
 
 ### Options
 
-| Variable    | Default         | Description                                      |
-| ----------- | --------------- | ------------------------------------------------ |
-| `symbol`    | `"↕️ "`          | The symbol used before displaying the SHLVL.     |
-| `threshold` | `2`             | Display threshold.                               |
-| `prefix`    | `""`            | Prefix to display immediately the value.         |
-| `suffix`    | `" "`           | Suffix to display immediately after the value.   |
-| `style`     | `"bold yellow"` | The style for the module.                        |
-| `disabled`  | `true`          | Disables the `shlvl` module.                     |
+| Variable    | Default                      | Description                                      |
+| ----------- | ---------------------------- | ------------------------------------------------ |
+| `threshold` | `2`                          | Display threshold.                               |
+| `format`    | `"[$symbol$shlvl]($style) "` | The format for the module.                       |
+| `symbol`    | `"↕️ "`                       | The symbol used to represent the SHLVL.          |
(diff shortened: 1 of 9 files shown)


docs/config/README.md
```diff
@@ -1317,6 +1318,32 @@ The module will be shown if any of the following conditions are met:
 symbol = "⚙️ "
 ```
 
+## SHLVL
+
+The `shlvl` module shows the current SHLVL ("shell level") environment variable, if it is
+set and greater than the specified threshold.
```
yacoob, line 1324: s/set/available? And maybe mention that this is supported in bash and zsh only:
https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-SHLVL
http://zsh.sourceforge.net/Doc/Release/Parameters.html#Parameters-Set-By-The-Shell-1

daniel-white: i didn't want to be too prescriptive incase that changes.

src/modules/shlvl.rs
```diff
@@ -0,0 +1,54 @@
...
+                Some(module)
+            } else {
+                None
+            }
+        }
+        None => None,
+    }
+}
+
+fn get_shlvl_value() -> Option<i64> {
```
nickwb, line 43: Ok, a few notes. If you are going to convert from OsString to String, you might as well use `env::var` rather than `env::var_os`, which does that conversion for you.

I believe you should be able to write the whole function as:
`env::var(SHLVL_ENV_VAR).ok()?.parse::<i64>().ok()`

(But I didn't try to compile this, I might have missed something subtle.)

src/modules/shlvl.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use super::{Context, Module, SegmentConfig};
+
+use crate::config::RootModuleConfig;
+use crate::configs::shlvl::ShLvlConfig;
+
+const SHLVL_ENV_VAR: &str = "SHLVL";
+
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let shlvl = get_shlvl_value();
```
nickwb, line 11: You can remove the match below if you use the `?` operator, i.e.:
`let shlvl = get_shlvl_value()?;`

src/configs/shlvl.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub suffix: &'a str,
+    pub style: Style,
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for ShLvlConfig<'a> {
+    fn new() -> Self {
+        ShLvlConfig {
+            symbol: Some(SegmentConfig::new("↕️ ")),
+            threshold: 3,
```
nickwb, line 20: I'm curious why `3` is your default? Would `2` not make more sense? Is there some kind of common scenario you are accounting for here?

daniel-white: the original issue said 3, but i can change that since it is a config option :)

nickwb: Ah, I should have read the original issue 😅 

The `tmux`/`screen` scenario makes sense.

src/modules/shlvl.rs
daniel-white: thanks. for my own edification, does that essentially destructure the `Option` and return `None` if the inner one is?

nickwb: Yes.

Think of `x?` like:
`match x { Some value => value, None => return None }`

It's a little subtle, because its effect encompasses the local function, not just the current expression.

This is why you can only use the syntax with a function which returns `Option<T>`, or a function which returns `Result<T, E>` (The de-sugaring is slightly different for result functions, as you might imagine).

nickwb:
The code and the tests looks good 👍  - I'd like to build and test for myself locally before I give you the green tick, hopefully I can find some time to do that in a few hours time.

src/modules/shlvl.rs
```diff
@@ -0,0 +1,40 @@
...
+
+const SHLVL_ENV_VAR: &str = "SHLVL";
+
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let shlvl = get_shlvl_value()?;
+
+    let mut module = context.new_module("shlvl");
+    let config: ShLvlConfig = ShLvlConfig::try_load(module.config);
+
+    if config.disabled {
```
vladimyr, line 16:
```suggestion
    if config.disabled || shlvl < config.threshold {
```

Use early return to avoid indenting the main portion of the function.

nickwb:
Finally got around to testing this @daniel-white. Sorry for the delay.

Was it your intention to keep the module disabled by default? I don't have a strong opinion either way, but because of the default threshold set, I don't think it would be a problem to enable this by default.

One minor note, is you're overwriting the normal default suffix of `" "` to instead have a default of `""`. This means it encroaches awkwardly on the next module:

image

If we instead use the normal space as the default, we get:

image

daniel-white:
@nickwb thanks! i'm pushing a change that will set the default to `" "` for the suffix. as for the default, i thought it wasn't a feature that everyone would want on - coupled with the fact theres a lot of edge cases

docs/config/README.md
```diff
@@ -1317,6 +1318,33 @@ The module will be shown if any of the following conditions are met:
...
+
+### Options
+
+| Variable    | Default         | Description                                      |
+| ----------- | --------------- | ------------------------------------------------ |
+| `symbol`    | `"↕️ "`          | The symbol used before displaying the SHLVL.     |
+| `threshold` | `2`             | Display threshold.                               |
+| `prefix`    | `""`            | Prefix to display immediately the value.         |
+| `suffix`    | `""`            | Suffix to display immediately after the value.   |
```
daniel-white, line 1333:
```suggestion
| `suffix`    | `" "`           | Suffix to display immediately after the value.   |
```

src/configs/shlvl.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub disabled: bool,
+}
+
+impl<'a> RootModuleConfig<'a> for ShLvlConfig<'a> {
+    fn new() -> Self {
+        ShLvlConfig {
+            symbol: Some(SegmentConfig::new("↕️  ")), // extra space for emoji
+            threshold: 2,
+            prefix: "",
+            suffix: "",
```
daniel-white, line 22:
```suggestion
            suffix: " ",
```

daniel-white:
i can't seem to retrigger the build...

daniel-white:
got to update this to support the new format stuff

src/modules/shlvl.rs
```diff
@@ -0,0 +1,53 @@
...
+
+const SHLVL_ENV_VAR: &str = "SHLVL";
+
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let shlvl = get_shlvl_value()?;
+
+    let mut module = context.new_module("shlvl");
+    let config: ShLvlConfig = ShLvlConfig::try_load(module.config);
+
+    if config.disabled || shlvl < config.threshold {
```
andytom, line 17: This is a bit of a nitpick but the description says that this is only displayed if the level is greater than the threshold but this will display if the level is greater than or equal to the threshold. Should this be

```suggestion
    if config.disabled || shlvl <= config.threshold {
```
Or maybe the description should be changed.

tests/testsuite/shlvl.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+#[test]
+fn formatting() -> io::Result<()> {
+    let output = common::render_module("shlvl")
+        .env_clear()
+        .use_config(toml::toml! {
+            [shlvl]
+            format = "$symbol going down [$shlvl]($style) GOING UP "
+            suffix = " level(s) "
```
andytom, line 151: I'm not sure this line does anything should this be removed?

```suggestion
```

src/modules/shlvl.rs
daniel-white: @andytom thanks! i updated - made it a little more clear to reduce ambiguity

tests/testsuite/shlvl.rs
daniel-white: removed

andytom:
@daniel-white thank you for your contribution and thank you @vladimyr, @nickwb and @yacoob for reviewing.

src/configs/shlvl.rs
```diff
@@ -16,9 +16,9 @@ pub struct ShLvlConfig<'a> {
 impl<'a> RootModuleConfig<'a> for ShLvlConfig<'a> {
     fn new() -> Self {
         ShLvlConfig {
-            threshold: 3,
+            threshold: 2,
             format: "[$symbol$shlvl]($style)",
-            symbol: "↕️ ",
+            symbol: "↕️  ", // extra space for emoji
```
fadeevab, line 19: Hey @daniel-white, it's time to dive into the past :) **The question: why 2 spaces?** It just looks like too much (asymmetrical):
image 
image

Now, the 1 space (👍GOOD!):
image
image

Are you okay if I fix it?

daniel-white: The macOS emoji

PR #1391, opened 2020-06-25
chore: bump Homebrew formula on new release

vladimyr:
#### Description

Bump Homebrew formula as part of the release workflow.

#### Motivation and Context

Automate all the things 🤖 
Special thanks go to @ATiltedTree who's PR #1360 inspired me to do this.

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -175,21 +175,11 @@ jobs:
     name: Update Brew Formula
     runs-on: macos-latest
     steps:
-      - uses: actions/checkout@v2
-      - run: |
-          echo "https://matchai:$HOMEBREW_GITHUB_API_TOKEN@github.com" > ~/.git-credentials
-          git config --global credential.helper store
-          git config --global user.name "Matan Kushner"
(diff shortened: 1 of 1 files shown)


matchai:
It looks like you might've not noticed this part of our workflow 😅 
https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L173-L192

vladimyr:
> It looks like you might've not noticed this part of our workflow 😅

Yeah, it is official, I'm an idiot 🤦 

Let me know what do you think about swapping that snippet with Mislav's action.

.github/workflows/workflow.yml
```diff
@@ -175,21 +175,11 @@ jobs:
     name: Update Brew Formula
     runs-on: macos-latest
```
matchai, line 176: It looks like we would no longer need to run this on `macos-latest` with this new action:
```suggestion
    runs-on: ubuntu-latest
```

vladimyr: Indeed 👍

PR #1394, opened 2020-06-26
chore: Deprecate affix and module style

heyrict:
#### Description
Deprecation of prefix and suffix for all modules.
the `style` member will also be deprecated in preference of `config.style`.

#### Motivation and Context

Related #1057 

#### Checklist:

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -1,7 +1,6 @@
 use crate::context::Shell;
 use crate::segment::Segment;
 use crate::utils::wrap_colorseq_for_shell;
-use ansi_term::Style;
 use ansi_term::{ANSIString, ANSIStrings};
 use std::fmt;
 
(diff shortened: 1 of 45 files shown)


src/module.rs
```diff
@@ -152,9 +118,6 @@ impl<'a> Module<'a> {
             .map(Segment::ansi_string)
             .collect::<Vec<ANSIString>>();
 
-        ansi_strings.insert(0, self.prefix.ansi_string());
-        ansi_strings.push(self.suffix.ansi_string());
-
```
ATiltedTree, line 157: Now that we are not modifying `ansi_strings` anymore it does not have to mut anymore. But the match statement has to be changed to return it's value. So something like this:
```rust
match shell {
    Shell::Bash => ansi_strings_modified(ansi_strings, shell),
    Shell::Zsh => ansi_strings_modified(ansi_strings, shell),
    _ => ansi_strings,
}
```

heyrict: If you'd like, you can directly push that to my branch as this is a minor change. Or I'll do that this Saturday.

ATiltedTree: I directly pushed it so that we can get things moving a bit quicker.

PR #1374, opened 2020-06-22
refactor: Refactor modules with format strings

heyrict:
#### Description
The **final piece** of [format strings release](https://github.com/starship/starship/milestone/3)

#### Motivation and Context
Blocked by #1056 
Closes #1057 

<details>
<summary>, and other associated issues</summary>

Closes #1208 
Closes #346 
Closes #337 
Closes #687 

</details>

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #1396, opened 2020-06-26
feat: Make git_branch prefix configurable

brendanmaguire:
#### Description
Allows the `prefix` symbol for the `git_branch` module to be configured

#### Motivation and Context
Provides functionality that is commonly available in other modules

I ran `cargo build` and ran `target/debug/starship prompt`. It defaults to the previous `prefix` value, `"on "`, but now also supports a configured prefix in the `~/.config/starship` file.

#### Checklist:


andytom:
Hi @brendanmaguire, thank you for the contribution but unfortunately given how close we are to releasing the format string feature which changes how the prefixes work for all modules, see #1374, I don't think this would actually be released. I'm going to close this PR, sorry about that.

brendanmaguire:
@andytom No worries. I should have looked around and saw other closed commits for the same change :sweat_smile:

PR #1411, opened 2020-06-29
fix(git_status): Remove renames_index_to_workdir() option from git status

### LINKED ISSUE #1371: Git status deleted and renamed show up for both changes in the working directory and in staging.
#### Current Behavior
Git status `deleted` and `renamed ` symbols show up for both changes in the working directory and for when a file is staged.

#### Expected Behavior
Git status `deleted` and `renamed ` symbols show up for when a file is staged.

In the [documentation](https://starship.rs/config/#git-status), `deleted ` and `renamed` are both described as being related to when a file is in the staging area. But it looks like `deleted` and `renamed` symbols show up regardless of whether they are in staging or the working directory.

Here's an example, given this setup:
```sh
example-repo on  release 
❯ tree
.
├── one.txt
└── two.txt

0 directories, 2 files
```

If I delete **one.txt**, and rename **two.txt** to **three.txt**, I see the `deleted` and `renamed` symbols even though the files are part of the working directory.

```sh
example-repo on  release [✘»] 
❯ git status --long
On branch release
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    one.txt
	deleted:    two.txt

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	three.txt

no changes added to commit (use "git add" and/or "git commit -a")
```

And adding them to staging, the symbols remain (matching what's stated in the documentation).

```sh
example-repo on  release [✘»] 
❯ git status --long
On branch release
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	renamed:    one.txt -> three.txt
	deleted:    two.txt
```

#### Environment
- Starship version: starship 0.42.0
- Shell type: zsh
- Shell version: zsh 5.8
- Operating system: macOS 10.14.6

Aelnor:
This option causes advanced files rename detection which causes inconsistency between `git status` and Starship reports.
Closes #1371

#### Checklist:

### DIFF
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -260,7 +260,6 @@ fn get_repo_status(repository: &mut Repository) -> Result<RepoStatus, git2::Erro
     status_options
         .renames_from_rewrites(true)
         .renames_head_to_index(true)
-        .renames_index_to_workdir(true)
         .include_unmodified(true);
 
     let statuses = repository.statuses(Some(&mut status_options))?;
--- a/tests/testsuite/git_status.rs
+++ b/tests/testsuite/git_status.rs
@@ -503,48 +503,6 @@ fn shows_deleted_file_with_count() -> io::Result<()> {
     remove_dir_all(repo_dir)
 }
 
-#[test]
-#[ignore]
-fn prefix() -> io::Result<()> {
-    let repo_dir = common::create_fixture_repo()?;
-    File::create(repo_dir.join("prefix"))?.sync_all()?;
-    let output = common::render_module("git_status")
-        .arg("--path")
-        .arg(&repo_dir)
-        .env_clear()
-        .use_config(toml::toml! {
-            [git_status]
-            prefix = "("
-            style = ""
-        })
-        .output()?;
-    let actual = String::from_utf8(output.stdout).unwrap();
-    let expected = "(";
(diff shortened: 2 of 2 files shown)


Aelnor:
The faulty behaviour is caused by [`renames_index_to_workdir()` ](https://docs.rs/git2/0.7.5/git2/struct.StatusOptions.html#method.renames_index_to_workdir) option (src/modules/git_status.rs:202)

I'm not sure if Starship needs it, but I've found out it was introduced here (which doesn't help either): https://github.com/starship/starship/pull/247/files

andytom:
Thanks for the investigation @Aelnor, I think this should be fine but do you think it would be possible to add a test for this?

Aelnor:
> Thanks for the investigation @Aelnor, I think this should be fine but do you think it would be possible to add a test for this?

Sure. Done, please take a look

Aelnor:
That's weird. These tests don't fail in my environment and they were okay before the latest rebase :)
Will investigate.

Aelnor:
Tests are ok now. I'd like someone to take a look at this PR please.

andytom:
@Aelnor, Thank you for the contribution.

PR #1433, opened 2020-07-05
fix: manage sandboxed version of OCaml

### LINKED ISSUE #1395: Use sandboxed version of OCaml if any
#### Current Behavior

OCaml module display OCaml version from global.

1. Since opam 2 and esy.sh may sandbox OCaml in a directory it may be the wrong version.
1. OCaml project managed by esy.sh may be detect as nodejs project because it also use a package.json manifest 

#### Expected Behavior

- [ ] Should not display nodejs info for esy.sh OCaml project

- [ ] Should display sandboxed version of OCaml instead of global version if any (should be 4.08 not 4.10)

Capture d’écran 2020-06-26 à 11 43 28

- [ ] display sandboxed version of OCaml when managed with esy.sh without global OCaml installation (should be 4.08 not empty)

Capture d’écran 2020-06-26 à 11 42 45

#### Possible Solution

- When `esy.lock` file exists `package.json` is not related to nodejs but OCaml
- When`esy.lock` exists read sandboxed OCaml version 
`$  esy ocaml -vnum`

#### Environment
- Starship version: 0.42.0
- Shell type:  zsh
- Shell version: zsh 5.7.1 (x86_64-apple-darwin19.0)
- Shell plugin manager:oh-my-zsh
- Terminal emulator:  iTerm
- Operating system: macOS 10.15.5

oteku:
#### Description

- Should not display nodejs info for esy.sh OCaml project
- Should display sandboxed version of OCaml instead of global version if any (should be 4.08 not 4.10)
- display sandboxed version of OCaml when managed with esy.sh without global OCaml installation (should be 4.08 not empty)

#### Motivation and Context

Closes #1395

#### Checklist:

### DIFF
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -18,7 +18,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .set_folders(&["node_modules"])
         .is_match();
 
-    if !is_js_project {
+    let is_esy_project = context
+        .try_begin_scan()?
+        .set_folders(&["esy.lock"])
+        .is_match();
+
+    if !is_js_project || is_esy_project {
         return None;
     }
 
@@ -59,6 +64,19 @@ mod tests {
 
         let actual = render_module("nodejs", dir.path(), None);
         let expected = Some(format!("via {} ", Color::Green.bold().paint("⬢ v12.0.0")));
+        assert_eq!(expected, actual);
+        dir.close()
+    }
+
+    #[test]
+    fn folder_with_package_json_and_esy_lock() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        File::create(dir.path().join("package.json"))?.sync_all()?;
+        let esy_lock = dir.path().join("esy.lock");
+        fs::create_dir_all(&esy_lock)?;
+
+        let actual = render_module("nodejs", dir.path(), None);
+        let expected = None;
         assert_eq!(expected, actual);
         dir.close()
     }
--- a/src/modules/ocaml.rs
+++ b/src/modules/ocaml.rs
@@ -29,12 +29,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .set_folders(&["esy.lock"])
         .is_match();
 
-    let ocaml_version =  if is_esy_project {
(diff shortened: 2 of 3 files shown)


vladimyr:
Unfortunately, you just created a bigger mess:

image

As you can see looking at my screenshot I have:
1. globally installed OCaml 4.09.0
2. esy managed local OCaml 4.08.1 installation

All is fine and dandy while I'm inside the directory containing `esy.lock`. As soon as I navigate away to any of the subfolders OCaml module fails to locate lockfile and reports global version instead. Simply put you can't just stick following snippet and be done with it:
```rust
let is_esy_project = context
    .try_begin_scan()?
    .set_folders(&["esy.lock"])
    .is_match();
```
You need to do a full topological search climbing up on the filesystem tree...

**PS** Definitely not your fault but also note those blank lines coming from untrimmed `(esy) ocaml -vnum` output...

vladimyr:
Btw, you could theoretically cut some corners using `esy status` which would give you:
```
fnm/library on  master via 🐫 v4.09.0
 took 10s
❯ esy status
{
  "isProject": true,
  "isProjectSolved": true,
  "isProjectFetched": true,
  "isProjectReadyForDev": true,
  "rootBuildPath": "/Users/vladimyr/workspace/github.com/Schniz/fnm/_esy/default/store/b/fnm-9eeaf197",
  "rootInstallPath": "/Users/vladimyr/workspace/github.com/Schniz/fnm/_esy/default/store/i/fnm-9eeaf197",
  "rootPackageConfigPath": "/Users/vladimyr/workspace/github.com/Schniz/fnm/package.json"
}
```
but that is no go in practice! It is implemented with filesystem side effects, more specifically it generates `_esy` dir polluting projects that aren't necessarily meant to be managed using it.

tcoopman:
Running `esy ocaml -vnum` in directories where you haven't run `esy` yet will result in command timed out and keep `esy` running in the background until everything has been build. So when you actually start `esy` manually you get conflicts because an other `esy` command is already running...

This doesn't seem like a great thing, is it possible to disable this, or to make sure to kill `esy` when the command times out?

chipbuster:
@tcoopman That sounds like a pretty significant problem. Could you open a new issue on it?

PR #1444, opened 2020-07-06
feat(cmd_duration): custom duration format

vladimyr:
#### Description

Enable configuring command duration display using a format string with time unit tokens.

#### Motivation and Context

Improve command duration output readability. :bowtie: 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -498,6 +498,7 @@ running `eval $(starship init $0)`, and then proceed as normal.
 | Option               | Default                                                  | Description                                                |
 | -------------------- | -------------------------------------------------------- | ---------------------------------------------------------- |
 | `min_time`           | `2_000`                                                  | Shortest duration to show time for (in milliseconds).      |
+| `show_zero_units`    | `false`                                                  | Display zero time units. (Turns `2h30s` into `2h0m30s`.)   |
 | `show_milliseconds`  | `false`                                                  | Show milliseconds in addition to seconds for the duration. |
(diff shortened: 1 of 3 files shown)


matchai:
This mostly comes down to personal preference.
I personally prefer to keep it how it is, but if the majority of maintainers would rather have the added whitespace, I'd be happy to see it changed.

vladimyr:
> This mostly comes down to personal preference.
> I personally prefer to keep it how it is...

@matchai Honestly, I don't care much about actual format. It caught my eye only because I was longtime spaceship user before switching over and spaceship formats it with spaces 🤷

chipbuster:
My vote would be to leave it as it is. No good reason aside from some vague sense of personal preference.

If this is an issue, we could introduce a similar option to format strings for time display.

vladimyr:
> If this is an issue, we could introduce a similar option to format strings for time display.

So If I want to both have and eat my ~cake~ spaces it is time for me to learn how to display compound duration in Rust using externally provided format string 😁 
So, I went and looked it up on Rosetta Code: https://rosettacode.org/wiki/Convert_seconds_to_compound_duration#Rust
Note the spaces 😝

ealap:
> If this is an issue, we could introduce a similar option to format strings for time display.
For me, it's a readability issue. So I would really appreciate it there is an option to format it according to my preference.

vladimyr:
> For me, it's a readability issue. So I would really appreciate it if there is an option to format it according to my preference.

I've just updated this PR to enable the duration format to be configured! Check out updated configuration docs. :tada:

OlaoluwaM:
Hi! Any chance of this getting merged?

PR #1449, opened 2020-07-07
ci(python): Use the default version of python in CI

andytom:
#### Description

This removes the hard requirement for a particular version of python in
the CI, and just uses the default version. Since #1297 was merged we
have not depended on the version of python that is installed. 

#### Motivation and Context

This should stop us from having to update the version of python when the
github action updates the available versions.

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -124,11 +124,6 @@ jobs:
           toolchain: stable
           override: true
 
-      # Install Python at a fixed version
-      - uses: actions/setup-python@v1
-        with:
-          python-version: "3.7.7"
-
       # Install dotnet at a fixed version
       - uses: actions/setup-dotnet@v1
         with:


.github/workflows/workflow.yml
```diff
@@ -124,16 +124,17 @@ jobs:
...
       - uses: actions/setup-dotnet@v1
         with:
           dotnet-version: "2.2.402"
 
+      # Ensure we have a recent version of python installed for Mercurial
+      - uses: actions/setup-python@v1
+        if: matrix.os == 'macOS-latest'
+        with:
+          python-version: "3.8"
+
```
ATiltedTree, line 137: I think we don't even need that since [mercurial requires python 3.5](https://www.mercurial-scm.org/wiki/SupportedPythonVersions) and [python 3.7 is preinstalled](https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md#language-and-runtime).

andytom:
> LGTM.
> Also shouldn't we mock terraform as well and remove it from here since we only want it's version?

We should but if I remember correctly the terraform tests depend a lot on the env vars that are set and we don't have a way to mock env vars yet.

ATiltedTree:
> > LGTM.
> > Also shouldn't we mock terraform as well and remove it from here since we only want it's version?
> 
> We should but if I remember correctly the terraform tests depend a lot on the env vars that are set and we don't have a way to mock env vars yet.

Looking through the tests I can't see any direct calls to terraform and all the directories and envvars are being set by us in the integration tests. And the unit test only test the correctness of the format function.

andytom:
> > > LGTM.
> > > Also shouldn't we mock terraform as well and remove it from here since we only want it's version?
> > 
> > 
> > We should but if I remember correctly the terraform tests depend a lot on the env vars that are set and we don't have a way to mock env vars yet.
> 
> Looking through the tests I can't see any direct calls to terraform and all the directories and envvars are being set by us in the integration tests. And the unit test only test the correctness of the format function.

Ok, I had a quick look the two tests that depend on terraform are the ones that get the version [`folder_with_dotterraform_with_version_no_environment`](https://github.com/starship/starship/blob/master/tests/testsuite/terraform.rs#L123) and [`folder_with_dotterraform_with_version_with_environment`](https://github.com/starship/starship/blob/master/tests/testsuite/terraform.rs#L148), I think they could be moved into [`src/modules/terraform.rs`](https://github.com/starship/starship/blob/master/src/modules/terraform.rs) which would allow us to mock out the terraform version call, but the rest of the tests would need to stay there until we fix do some more work on #941.

PR #1450, opened 2020-07-07
ci: General overhaul

ATiltedTree:
#### Description

A general overhaul for our CI.
Some of the changes include:

- Switching from [actions-rs/toolchain](https://github.com/actions-rs/toolchain) to [ATiltedTree/setup-rust](https://github.com/ATiltedTree/setup-rust) as it's pretty stale
- Adding a consistent naming scheme to our steps
- Separating our deploy steps into a different file
- Condense our cache hits

Also I have been running *a lot* of CI and have yet to see this arbitrarily fail like in #886. But that might change at any moment.

All of this was thoroughly tested and a test release was also created [on my fork](https://github.com/ATiltedTree/starship/releases/tag/v0.45.0). It includes more commit than it should because my fork is missing some tags.

#### Motivation and Context

Clean code is better code.
Might solve #886

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -21,17 +21,17 @@ jobs:
           rust-version: stable
 
       - name: Build | Publish
-        run: cargo publish --dry-run #--token ${{ secrets.CRATES_IO_TOKEN }}
+        run: cargo publish --token ${{ secrets.CRATES_IO_TOKEN }}
 
-  # update_brew_formula:
-  #   name: Update Brew Formula
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #     - uses: mislav/bump-homebrew-formula-action@v1.6
-  #       with:
-  #         formula-name: starship
-  #       env:
-  #         COMMITTER_TOKEN: ${{ secrets.HOMEBREW_GITHUB_API_TOKEN }}
+  update_brew_formula:
+    name: Update Brew Formula
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v1.6
+        with:
+          formula-name: starship
+        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,147 @@
+name: Deploy
+on:
+  push:
+    branches:
+      - master
```
matchai, line 5: I think we will only want releases on pushes to `v*` tags. No need to release for unversioned pushed to master.
```suggestion
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,147 @@
...
+  # Publish starship to Crates.io
+  cargo_publish:
+    name: Publish Cargo Package
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup | Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup | Rust
+        uses: ATiltedTree/setup-rust@v1
```
matchai, line 19: Is there anything specific lacking from `actions-rs/toolchain`?
I have no doubt that your action works perfectly, but I would prefer to stick with something widely known and used if it does the job as expected.

ATiltedTree: I created my own version and changed to it because while `actions-rs/toolchain` works perfectly you always have to use `actions-rs/cargo` if you want annotations in your code.
Also `actions-rs/cargo` might actually be the cause of the random check failure as I have never seen it fail with my current changes.

But I can always change it back if you want to.

matchai: We could keep using `ATiltedTree/setup-rust` to see if it solves the random failures, but I'm fairly confident that the flakiness of our tests is due to race conditions and brittleness in how our tests work.
If we continue to see flakiness, I think it'd be best if we brought it back to `actions-rs/toolchain`.

PR #1461, opened 2020-07-09
feat(hg_branch): display topic of current repo

### LINKED ISSUE #1384: [mercurial] support for topic (lightweight branches)

arthurzenika:
#### Description

Attempt to extract information from `.hg/topic` in hg_branch module

I am not a rust developer and so this is untested code. 

#### Motivation and Context

(attempts to start) fixes  #1384

#### Checklist:

started, but does it need to be in all languages ? 

... tried

### DIFF
--- a/src/modules/hg_branch.rs
+++ b/src/modules/hg_branch.rs
@@ -79,8 +79,7 @@ fn get_hg_branch_name(ctx: &Context) -> String {
 fn get_hg_topic_name(ctx: &Context) -> String {
     std::fs::read_to_string(ctx.current_dir.join(".hg").join("topic"))
         .map(|s| s.trim().into())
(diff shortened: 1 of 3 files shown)


src/modules/hg_branch.rs
```diff
@@ -76,6 +76,11 @@ fn get_hg_branch_name(ctx: &Context) -> String {
         .unwrap_or_else(|_| "default".to_string())
 }
 
+fn get_hg_topic_name(ctx: &Context) -> String {
+    std::fs::read_to_string(ctx.current_dir.join(".hg").join("topic"))
+        .map(|s| s.trim().into())
+        .ok()
+
```
andytom, line 83: I think most of these issues are caused by a missing `}` here. I think this should be fix it
```suggestion
}

```

PR #1463, opened 2020-07-09
feat(helm): Add Helm module

### LINKED ISSUE #1443: Helm Support
Helm as simply as possible is a package manager for Kubernetes. We simply deploy applications to the Kubernetes using Helm's package format called Chart.

Sometimes we might need to know about Helm's version that I am using on Kubernetes cluster. We can simply get the version information of Helm with version command like below:

```
$ helm version
version.BuildInfo{Version:"v3.2.1", GitCommit:"fe51cd1e31e6a202cba7dead9552a6d418ded79a", GitTreeState:"clean", GoVersion:"go1.13.10"}
```

I thought it might be good If I could see the Helm version on my terminal that I am using with the Helm's logo.

developer-guy:
Signed-off-by: Furkan <furkan.turkal@hotmail.com>

#### Motivation and Context

Closes #1443 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -193,6 +193,7 @@ $elixir\
 $elm\
 $erlang\
 $golang\
+$helm\
 $java\
 $julia\
 $nim\
@@ -1112,6 +1113,42 @@ The module will be shown if any of the following conditions are met:
 
 [golang]
 format = "via [🏎💨 $version](bold cyan) "
+```
+
+## Helm
+
+The `helm` module shows the currently installed version of Helm.
+The module will be shown if any of the following conditions are met:
+
+- The current directory contains a `helmfile.yaml` file
+- The current directory contains a `Chart.yaml` file
+
+### Options
+
+| Option     | Default                            | Description                                    |
+| ---------- | ---------------------------------- | ---------------------------------------------- |
+| `format`   | `"via [$symbol$version]($style) "` | The format for the module.                     |
+| `symbol`   | `"⎈ "`                            | A format string representing the symbol of Helm. |
+| `style`    | `"bold white"`                      | The style for the module.                      |
+| `disabled` | `false`                            | Disables the `helm` module.                  |
+
+### Variables
+
+| Variable | Example   | Description                          |
+| -------- | --------- | ------------------------------------ |
+| version  | `v3.1.1` | The version of `helm`                  |
+| symbol   |           | Mirrors the value of option `symbol` |
+| style\*  |           | Mirrors the value of option `style`  |
+
+\*: This variable can only be used as a part of a style string
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[helm]
+format = "via [⎈ $version](bold white) "
 ```
 
 ## Hostname
--- a/src/modules/helm.rs
+++ b/src/modules/helm.rs
@@ -1,0 +1,127 @@
+use super::{Context, Module, RootModuleConfig};
+
+use crate::configs::helm::HelmConfig;
+use crate::formatter::StringFormatter;
+use crate::utils;
+
+/// Creates a module with the current Helm version
+///
+/// Will display the Helm version if any of the following criteria are met:
+///     - Current directory contains a `helmfile.yaml` file
+///     - Current directory contains a `Chart.yaml` file
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let is_helm_project = context
+        .try_begin_scan()?
+        .set_files(&["helmfile.yaml", "Chart.yaml"])
+        .is_match();
+
+    if !is_helm_project {
+        return None;
+    }
+
+    let mut module = context.new_module("helm");
+    let config = HelmConfig::try_load(module.config);
+    let parsed = StringFormatter::new(config.format).and_then(|formatter| {
+        formatter
+            .map_meta(|var, _| match var {
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "version" => format_helm_version(
+                    &utils::exec_cmd("helm", &["version", "--short", "--client"])?
+                        .stdout
+                        .as_str(),
+                )
+                .map(Ok),
+                _ => None,
+            })
+            .parse(None)
+    });
+
+    module.set_segments(match parsed {
+        Ok(segments) => segments,
+        Err(error) => {
+            log::warn!("Error in module `helm`:\n{}", error);
+            return None;
+        }
+    });
+
+    Some(module)
+}
+
+fn format_helm_version(helm_stdout: &str) -> Option<String> {
+    // `helm version --short --client` output looks like this:
+    // v3.1.1+gafe7058
+    // `helm version --short --client` output looks like this for Helm 2:
+    // Client: v2.16.9+g8ad7037
+
+    Some(
+        helm_stdout
+            // split into ["v3.1.1","gafe7058"] or ["Client: v3.1.1","gafe7058"]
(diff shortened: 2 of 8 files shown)


Dentrax:
Hey @andytom, we have tried to add Helm support as we talked about at #1443. But somehow, these tests does just fail by Ubuntu. Even so, tests are passing by `nightly` [unlike](https://github.com/starship/starship/runs/856881272?check_suite_focus=true) `stable`, that's where we couldn't figure it out why. Thanks! :)

andytom:
> Hey @andytom, we have tried to add Helm support as we talked about at #1443. But somehow, these tests does just fail by Ubuntu. Even so, tests are passing by `nightly` [unlike](https://github.com/starship/starship/runs/856881272?check_suite_focus=true) `stable`, that's where we couldn't figure it out why. Thanks! :)

Unfortunately some of our tests are a bit unstable, see #886, we are trying to figure out the root cause but not quite there yet. I'll rerun the tests as that normally fixes the issue.

src/modules/helm.rs
```diff
@@ -0,0 +1,119 @@
...
+	// v3.1.1+gafe7058
+
+	let version = helm_stdout
+		// split into ["v3.1.1","gafe7058"]
+		.splitn(2, '+')
+		// return "v3.1.1"
+		.next()?;
+
+	Some(version.to_string())
+}
```
andytom, line 69: Can you update this function to also support helm 2, this should just be a case of removing the prefixed `Client: ` from the version string.
```
helm version --short --client
Client: v2.16.9+g8ad7037
```

src/modules/helm.rs
```diff
@@ -0,0 +1,119 @@
...
+		let expected = Some(format!("via {} ", Color::Blue.bold().paint("⎈ v3.1.1")));
+		assert_eq!(expected, actual);
+		dir.close()
+	}
+
+	#[test]
+	fn test_format_helm_version() {
+		let input = "v3.1.1+ggit afe7058";
+		assert_eq!(format_helm_version(input), Some("v3.1.1".to_string()));
+	}
```
andytom, line 118: Maybe change this to also test for helm2 versions.

 ```suggestion
	fn test_format_helm_version() {
		assert_eq!(format_helm_version("v3.1.1+gafe7058"), Some("v3.1.1".to_string()));
		assert_eq!(format_helm_version("Client: v2.16.9+g8ad7037"), Some("v2.16.9".to_string()));
	}
```

File: src/modules/helm.rs [hunk 3]
```diff
@@ -0,0 +1,119 @@
...
+/// Creates a module with the current Helm version
+///
+/// Will display the Helm version if any of the following criteria are met:
+///     - Current directory contains a `helmfile.yaml` file
+///     - Current directory contains a `Chart.yaml` file
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+	let is_helm_project = context
+		.try_begin_scan()?
+		.set_files(&["helmfile.yaml", "Chart.yaml"])
+		.set_folders(&["templates"])
```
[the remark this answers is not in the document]
Dentrax: In this case, i don't have to check any folders anymore, right?

src/modules/helm.rs
Dentrax: I just changed this parser logic, now should be support both Helm 1 and 2 version commands, i guess. :)

src/modules/helm.rs
Dentrax: Done

docs/config/README.md
```diff
@@ -1150,6 +1151,79 @@ trim_at = ".companyname.com"
...
+
+- The current directory contains a `helmfile.yaml` file
+- The current directory contains a `Chart.yaml` file
+
+### Options
+
+| Option     | Default                            | Description                                    |
+| ---------- | ---------------------------------- | ---------------------------------------------- |
+| `format`   | `"via [$symbol$version]($style) "` | The format for the module.                     |
+| `symbol`   | `"🐹 "`                            | A format string representing the symbol of Helm. |
```
cappyzawa, line 1167: based on https://github.com/starship/starship/pull/1463/files#diff-3009cc0091e546b48a6d01d63548e923R17.

```suggestion
| `symbol`   | `"⎈ "`                            | A format string representing the symbol of Helm. |
```

Dentrax: Updated

src/modules/helm.rs
```diff
@@ -0,0 +1,127 @@
...
+        let expected = Some(format!("via {} ", Color::Blue.bold().paint("⎈ v3.1.1")));
+        assert_eq!(expected, actual);
+        dir.close()
+    }
+
+    #[test]
+    fn test_format_helm_version() {
+        let helm_1 = "v3.1.1+ggit afe7058";
+        let helm_2 = "Client: v2.16.9+g8ad7037";
+        assert_eq!(format_helm_version(helm_1), Some("v3.1.1".to_string()));
```
andytom, line 124: Minor consistency correction here.

```suggestion
        let helm_3 = "v3.1.1+ggit afe7058";
        let helm_2 = "Client: v2.16.9+g8ad7037";
        assert_eq!(format_helm_version(helm_3), Some("v3.1.1".to_string()));
```

docs/config/README.md
```diff
@@ -1150,6 +1151,79 @@ trim_at = ".companyname.com"
...
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[helm]
+format = "via [⎈ $version](bold blue) "
+```
+
+## Hostname
```
andytom, line 1190: It looks like you have duplicated the hostname entry here.

src/modules/helm.rs
Dentrax: Done

docs/config/README.md
Dentrax: 👍 some copy-paste problems. :) removed.

Dentrax:
> Have added a few more suggestions, something else to think about is the default symbol and style are very similar to the defaults for the kubernetes module which I think a lot of helm users will also be using, can you think of any other style and symbols we could use?

I think you are right. Both of looks similar. But I couldn't find any other nice symbol unicodes. :(

andytom:
> > Have added a few more suggestions, something else to think about is the default symbol and style are very similar to the defaults for the kubernetes module which I think a lot of helm users will also be using, can you think of any other style and symbols we could use?
> 
> I think you are right. Both of looks similar. But I couldn't find any other nice symbol unicodes. :(

How about changing the default color to white maybe that will be enough?

Dentrax:
> I've been testing this locally and just realised there are a few other little changes that need to be made. You will need to add the module to the `PROMPT_ORDER` in `src/configs/starship_root.rs` and `ALL_MODULES` in `src/module.rs`.

Fixed. :)



> > > Have added a few more suggestions, something else to think about is the default symbol and style are very similar to the defaults for the kubernetes module which I think a lot of helm users will also be using, can you think of any other style and symbols we could use?
> > 
> > 
> > I think you are right. Both of looks similar. But I couldn't find any other nice symbol unicodes. :(
> 
> How about changing the default color to white maybe that will be enough?

Yep, White seems to be more appropriate for this case.

docs/config/README.md
```diff
@@ -1114,6 +1115,42 @@ The module will be shown if any of the following conditions are met:
...
+
+\*: This variable can only be used as a part of a style string
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[helm]
+format = "via [⎈ $version](bold blue) "
```
cappyzawa, line 1151: Based on https://github.com/starship/starship/pull/1463/files#diff-3009cc0091e546b48a6d01d63548e923R18

```suggestion
format = "via [⎈ $version](bold white) "
```

andytom:
Thanks for your contribution @Dentrax and @developer-guy and thank you for reviewing @cappyzawa.

PR #1474, opened 2020-07-13
docs: add uninstallation instructions to FAQ

matchai:
#### Description

According to our docs website analytics, the single most searched term is "uninstall", which signals that we need more clarity on that front.

This PR aims to address that gap in our docs.

#### Checklist:

### DIFF
--- a/docs/faq/README.md
+++ b/docs/faq/README.md
@@ -46,7 +46,7 @@ The prompt will use as much context as is provided, but no flags are "required".
 
 ## How do I run Starship on Linux distributions with older versions of glibc?
 
-If you get an error like "*version 'GLIBC_2.18' not found (required by starship)*" when using the prebuilt binary (for example, on CentOS 6 or 7), you can use a binary compiled with `musl` instead of `glibc`:
+If you get an error like "_version 'GLIBC_2.18' not found (required by starship)_" when using the prebuilt binary (for example, on CentOS 6 or 7), you can use a binary compiled with `musl` instead of `glibc`:
 
 ```sh
 curl -fsSL https://starship.rs/install.sh | bash -s -- --platform unknown-linux-musl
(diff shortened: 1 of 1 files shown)


docs/faq/README.md
```diff
@@ -78,3 +78,19 @@ If either symbol fails to display correctly, your system is still misconfigured.
 Unfortunately, getting font configuration correct is sometimes difficult. Users
 on the Discord may be able to help. If both symbols display correctly, but
 you still don't see them in starship, [file a bug report!](https://github.com/starship/starship/issues/new/choose)
+
+## How do I uninstall Starship?
+
+Starship is just as easy to uninstall as it is to install in the first place.
+
+1. Remove any lines in your shell config (e.g. `~/.bashrc`) used to initialize Starship.
+1. Delete the Starship binary.
```
chipbuster, line 87: This renders correctly in the preview. Is it idiomatic to use the same number in markdown lists and just let the renderer sort out the actual numbering?

docs/faq/README.md
```diff
@@ -78,3 +78,19 @@ If either symbol fails to display correctly, your system is still misconfigured.
...
+
+If Starship was installed using a package manager, please refer to their docs for uninstallation instructions.
+
+If Starship was installed using the `curl | bash` script, the following command will delete the binary:
+
+```sh
+# Locate and delete the starship binary
+rm -rf $(which starship)
```
chipbuster, line 95: This is potentially dangerous if the user makes poor file naming choices (e.g. if starship is installed at `/Users/<user>/Documents 2/programs/starship`, this will cause a deletion of their `~/Documents` folder).

```suggestion
rm "$(which starship)"
```

docs/faq/README.md
matchai: It comes down to personal preference. I've run into the problem of removing a step and being left with incorrect numbering one too many times. I wouldn't mind changing it though.

docs/faq/README.md
matchai: Good call! 👍

docs/faq/README.md
chipbuster: Nah, I just didn't realize markdown could do this for you. Might start doing that myself in the future--I've had similar issues with numbering not updating.

docs/faq/README.md
```diff
@@ -78,3 +78,19 @@ If either symbol fails to display correctly, your system is still misconfigured.
...
+1. Remove any lines in your shell config (e.g. `~/.bashrc`) used to initialize Starship.
+1. Delete the Starship binary.
+
+If Starship was installed using a package manager, please refer to their docs for uninstallation instructions.
+
+If Starship was installed using the `curl | bash` script, the following command will delete the binary:
+
+```sh
+# Locate and delete the starship binary
+rm "$(which starship)"
```
davidkna, line 95: I think it could be useful to explain that `sudo` might be needed to remove the starship binary given that was a problem with the former install script.

PR #1479, opened 2020-07-14
fix(ocaml): Stop new line being included in ocaml version

andytom:
#### Description

This fixes a regression in the ocaml module that caused the excess whitespace
to not be trimmed from the version.

#### Motivation and Context

Related to #1436 and 

#### Checklist:

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -84,11 +84,11 @@ active boot switches: -d:release\n",
             stderr: String::default(),
         }),
         "ocaml -vnum" => Some(CommandOutput {
-            stdout: String::from("4.10.0"),
+            stdout: String::from("4.10.0\n"),
             stderr: String::default(),
         }),
         "esy ocaml -vnum" => Some(CommandOutput {
-            stdout: String::from("4.08.1"),
+            stdout: String::from("4.08.1\n"),
             stderr: String::default(),
         }),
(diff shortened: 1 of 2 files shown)


src/utils.rs
```diff
@@ -84,11 +84,11 @@ active boot switches: -d:release\n",
             stderr: String::default(),
         }),
         "ocaml -vnum" => Some(CommandOutput {
-            stdout: String::from("4.10.0"),
+            stdout: String::from("4.10.0\n"),
             stderr: String::default(),
         }),
         "esy ocaml -vnum" => Some(CommandOutput {
-            stdout: String::from("4.08.1"),
+            stdout: String::from("4.08.1\n"),
```
vladimyr, line 91: Not sure what to think about this 🤔  Thing is, most of the other stdout/stderr outputs also lack trailing newline so we should probably fix all of them inside separate PR?

andytom: > Not sure what to think about this 🤔  Thing is, most of the other stdout/stderr outputs also lack trailing newline so we should probably fix all of them inside separate PR?

Yep, this is something we need to look at. The mocks should match the output of the commands exactly, including trailing whitespace, this is something that we haven't been doing very well so far and this is the second regression that this has caused the other being #1465. I'll create an issue to fix this for all the existing modules.

PR #1165, opened 2020-05-03
feat: add the status module

glehmann:
this allows to display more precisely the last command exit code
and to configure starship to not change the last charcter of the
prompt even in case of failure. It is disabled by default, because
it seems a bit redundant with the character module in its default
configuration.

#### Motivation and Context

Closes #193
Closes #882

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #1481, opened 2020-07-14
proof of concept for showing the exit status

oconnor663:
I'd like a way to automatically print the exit status of failed
commands, in addition to the fact that they failed. Here's an initial
stab at enabling something like this. (I haven't updated tests yet, so
for now this change causes failures.) With this change, I add this to my
config:

    [character]
    error_format = "[$code](red) [❯](bold red) "

Then my prompt looks like this:

image

I'm not sure if this is the best approach, either design-wise or
implementation-wise, but I wanted to put something up to get
suggestions. Another option would be to print an extra line above the
prompt containing the status; that might make is clearer that it's
associated with the previous command too.

### DIFF
--- a/src/configs/character.rs
+++ b/src/configs/character.rs
@@ -4,20 +4,18 @@ use starship_module_config_derive::ModuleConfig;
 
 #[derive(Clone, ModuleConfig)]
 pub struct CharacterConfig<'a> {
-    pub format: &'a str,
-    pub success_symbol: &'a str,
-    pub error_symbol: &'a str,
-    pub vicmd_symbol: &'a str,
+    pub success_format: &'a str,
+    pub error_format: &'a str,
+    pub vicmd_format: &'a str,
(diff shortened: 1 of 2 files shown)


jeffbyrnes:
@oconnor663 my current, pure bash implementation, which I used until discovering Starship, looks like this:

image

The second element, after the time, before my username, is the last exit code in parentheses.

As such, I think this is a great addition! I’d love to see it as its own module, so it can be positioned & configured separately from the `character` module. I can’t write a lick of Rust, or I’d lend a hand, but wanted to give some feedback nonetheless & an example of how someone else might want this to work.

jankatins:
https://github.com/starship/starship/pull/1165 has a status module which can be configured independently from the character.

jankatins:
And an updaed version of a status module is now in https://github.com/starship/starship/pull/1651

PR #1482, opened 2020-07-15
fix(explain): align table correctly

### LINKED ISSUE #1480: "explain" output alignment is foiled by multibyte unicode
#### Current Behavior
Running `starship explain` results in a misaligned table when the prompt segments are multibytes

```
 Pytest             -  The current working directory
  pytest-6.0.0rc1  -  The active branch of the repo in your current directory
 🐍 v3.8.2          -  The currently installed version of Python
 ☁️  pdx            -  The current AWS region and profile
```

#### Expected Behavior
I expect an aligned table:

```
 Pytest             -  The current working directory
  pytest-6.0.0rc1  -  The active branch of the repo in your current directory
 🐍 v3.8.2           -  The currently installed version of Python
 ☁️  pdx.             -  The current AWS region and profile
```

Just in case markdown rendering mucks with it:

image

#### Environment
- Starship version: 0.44.0
- Operating system: Mac OS 10.15.5
- Terminal emulator: iTerm.app 3.3.10beta4

```bash
<unknown config>
```

#### Starship Configuration

```toml
[prompt]
add_newline = false

[line_break]
disabled = true

[aws.region_aliases]
us-east-1 = "iad"
us-east-2 = "cmh"
us-west-2 = "pdx"
```

davidkna:
#### Description

The table in `explain` wasn't aligned correctly. I think this happens because since v0.12.0 textwrap doesn't count ANSI escape sequences anymore. This also means `max_ansi_width` is no longer needed.
I couldn't get `{:width$}` to work the way I wanted so I replaced it with `" ".repeat(width)`.
~Also by using `UnicodeWidthStr::width` the width calculation itself should be a bit better because it can work on the level of graphemes instead of characters.~ Sadly both `UnicodeWidthStr::width` and `UnicodeWidthStr::width` appear to fail at this. I replaced them with a custom textwrapper.

#### Motivation and Context

Closes #1480 

Before:
```
starship on  master [!?] is 📦 v0.44.0 via 🦀 v1.44.1
❯ starship explain

 starship          -  The current working directory
  master          -  The active branch of the repo in your current directory
 !?                -  Symbol representing the state of the repo
 📦 v0.44.0  -  The package version of the current directory's project
 🦀 v1.44.1        -  The currently installed version of Rust
```

After (this does not look aligned here but it does in my terminal):
```
 starship         -  The current working directory
 on  master      -  The active branch of the repo in your current directory
 [!?]             -  Symbol representing the state of the repo
 is 📦 v0.44.0    -  The package version of the current directory's project
 via 🦀 v1.44.1   -  The currently installed version of Rust
 ❯                -  A character (usually an arrow) beside where the text is entered in your terminal

```

#### Checklist:

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -4,7 +4,7 @@ use rayon::prelude::*;
 use std::collections::BTreeSet;
 use std::fmt::{self, Debug, Write as FmtWrite};
 use std::io::{self, Write};
-use unicode_width::UnicodeWidthChar;
+use unicode_width::UnicodeWidthStr;
 
 use crate::configs::PROMPT_ORDER;
 use crate::context::{Context, Shell};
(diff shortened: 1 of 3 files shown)


src/print.rs
```diff
@@ -104,50 +104,38 @@ pub fn explain(args: ArgMatches) {
...
-            info.value.chars().count() + count_wide_chars(&info.value),
-        );
-        max_module_width = std::cmp::max(max_module_width, info.value_len);
-    }
+    let max_module_width = modules.iter().map(|i| i.value_len).max().unwrap_or(0);
 
     let desc_width = term_size::dimensions()
         .map(|(w, _)| w)
-        .map(|width| width - std::cmp::min(width, max_ansi_module_width));
+        .map(|width| width - std::cmp::min(width, max_module_width + 6));
```
davidkna, line 117: I think it makes sense to fall back to `None` instead of allowing only one char per line. Something like 4 could be a good  cutoff. Any opinions on this?

davidkna: Also should I replace the 6 with a 7 so there's padding on both the left and right side?

davidkna:
My newest commit doesn't show up here. I'll make a new PR tomorrow if this persists.

davidkna:
It looks like most terminals have trouble properly displaying this kind of graphme clusters anyway and they'll actually end up as the wrong 8 characters wide.
Should I revert the custom textwrapping changes from this PR?

src/print.rs
andytom: I've tested this locally and 6 LGTM, only suggestion here would be to maybe add a comment to explain this so if we want to change it in the future it's nice an easy to find what we need to change.

src/print.rs
```diff
@@ -104,50 +105,71 @@ pub fn explain(args: ArgMatches) {
...
-        );
-        max_module_width = std::cmp::max(max_module_width, info.value_len);
-    }
+    let max_module_width = modules.iter().map(|i| i.value_len).max().unwrap_or(0);
 
     let desc_width = term_size::dimensions()
         .map(|(w, _)| w)
-        .map(|width| width - std::cmp::min(width, max_ansi_module_width));
+        // In addition to the module width itself there are also 6 padding characters in each line.
+        // This needs to be updated here and later in the function on line formatting changes.
```
andytom, line 119: Sorry didn't realise this earlier but if the 6 is used here are later on should we use a constant for this so they don't end up getting out of sync?

davidkna: I've replaced it with a constant.

andytom:
Thanks for the contribution @davidkna and for the quick responses.

PR #1490, opened 2020-07-18
test: Env mocking and a big refactor

### LINKED ISSUE #941: Create an abstraction around `std::env::var`, and a setter for tests.

We are currently working on writing better unit test for modules and moving away from having tests that depend on having the development machine setup in a particular way. However a lot of our modules depend on the values of env vars. There is no way to easily set or unset env vars when running unit tests since Rust will run the tests in parallel which could cause some race conditions in our tests.

> I'm afraid it's not going to be easy. We may have create an abstraction around `std::env::var`, and a setter for tests.
>
> For example:
```rust
#[test]
fn check_env_var() -> io::Result<()> {
  // starting now, `getEnvVar("USER")` will return `Some("root")`
  mockEnvVar("USER", "root");

  let actual = render_module("username", dir.path());
}
```

_Originally posted by @matchai in https://github.com/starship/starship/pull/936#issuecomment-586539715_

I've not really considered anything else but interested in any suggestions.

ATiltedTree:
#### Description
This PR adds a couple of things:
- A `ModuleRenderer` that allows for easy module testing and env mocking.
- Moves all integration test to the unit test with that new `ModuleRenderer`.
- Adds mocking for dotnet
- Refactors our contributing guidelines

There are still some problems but these can only be solved by switching to `git status --porcelain`.

#### Motivation and Context
It merges the two test together which were previously held apart due to the lack of env mocking.
It was mainly done to remove some of the jankyness that the integration test added. E.g. tests could only be run in debug mode as cargo only builds `target/release/starship` in release mode but we checked for `target/debug/starship`.
It also makes a slight move into the "drop git2" direction as none of the tests use git2 anymore.

Fixes #941

#### Checklist:


matchai:
This PR is looking great! 👍
I haven't had a chance to finish reviewing it, but thank you for taking up the challenge of implementing env mocking. This should greatly improve confidence and reliability in our tests! 🙌

CONTRIBUTING.md
```diff
@@ -56,37 +102,81 @@ rustup component add rustfmt
...
+      assert_eq!(actual, expected);
+
+      // Close the tempdir
+      tempdir.close()
+   }
+}
+```
 
-### Unit Testing
+If a module depends on output of another program then that output should be added to the match statement in [`utils.rs`](src/utils.rs). The match has to be exactly the same as the call to `utils::exec_cmd()` that means that the program will we joined with the argument by a `" "` meaning that the call `utils::exec_cmd("program", &["arg", "more_args"])` would turn into `program arg more_args`.
```
matchai, line 167: A few adjustments to make it easier to read: 
```suggestion
If a module depends on output of another program, then that output should be added to the match statement in [`utils.rs`](src/utils.rs). The match has to be exactly the same as the call to `utils::exec_cmd()`, including positional arguments and flags. The array of arguments are joined by a `" "`, so `utils::exec_cmd("program", &["arg", "more_args"])` would match with the `program arg more_args` match statement.
```

CONTRIBUTING.md
```diff
@@ -56,37 +102,81 @@ rustup component add rustfmt
...
 
-Unit tests are written using the built-in Rust testing library in the same file as the implementation, as is traditionally done in Rust codebases. These tests can be run with `cargo test`.
+If the program cannot be mocked (e.g. It performs some filesystem actions either writing or reading files) then it has to added to the project's GitHub Actions workflow file([`.github/workflows/workflow.yml`](.github/workflows/workflow.yml)) and the test has to be marked with a `#[ignored]` so that anyone can test their build without having to have some special programs.
```
matchai, line 169:
```suggestion
If the program cannot be mocked (e.g. It performs some filesystem operations, either writing or reading files) then it has to added to the project's GitHub Actions workflow file([`.github/workflows/workflow.yml`](.github/workflows/workflow.yml)) and the test has to be marked with an `#[ignored]`. This ensures that anyone can run the test suite locally without needing to pre-configure their environment. The `#[ignored]` attribute is bypassed during CI runs in GitHub Actions.
```

CONTRIBUTING.md
```diff
@@ -56,37 +102,81 @@ rustup component add rustfmt
...
-For tests that depend on having preexisting state, whatever needed state will have to be added to the project's GitHub Actions workflow file([`.github/workflows/workflow.yml`](.github/workflows/workflow.yml)).
-
 ### Test Programming Guidelines
 
 Any tests that depend on File I/O should use [`sync_all()`](https://doc.rust-lang.org/std/fs/struct.File.html#method.sync_all) when creating files or after writing to files.
 
-Any tests that use `tempfile::tempdir` should take care to call `dir.close()` after usage to ensure the lifecycle of the directory can be reasoned about.
-
-Any tests that use `create_fixture_repo()` should remove the returned directory after usage with  `remove_dir_all::remove_dir_all()`.
+Any tests that use `tempfile::tempdir` should take care to call `dir.close()` after usage to ensure the lifecycle of the directory can be reasoned about this includes `fixture_repo()` as it returns a TempDir that should be closed.
```
matchai, line 179:
```suggestion
Any tests that use `tempfile::tempdir` should take care to call `dir.close()` after usage to ensure the lifecycle of the directory can be reasoned about. This includes `fixture_repo()` as it returns a TempDir that should be closed.
```

Cargo.toml
```diff
@@ -31,7 +31,7 @@ tls-vendored = ["native-tls/vendored"]
 clap = "2.33.1"
 ansi_term = "0.12.1"
 dirs-next = "1.0.1"
-git2 = { version = "0.13.8", default-features = false, features = [] }
+git2 = { version = "0.13.8", default-features = false }
```
matchai, line 34: 👍

src/main.rs
```diff
@@ -1,11 +1,7 @@
+use clap::{crate_authors, crate_version};
 use std::io;
 use std::time::SystemTime;
 
-#[macro_use]
-extern crate clap;
-#[macro_use]
-extern crate pest_derive;
-
```
matchai, line 8: 👍

src/modules/aws.rs
```diff
@@ -47,19 +46,22 @@ fn get_aws_region_from_config(aws_profile: Option<&str>) -> Option<Region> {
     Some(region.to_string())
 }
 
-fn get_aws_profile_and_region() -> (Option<Profile>, Option<Region>) {
+fn get_aws_profile_and_region(contex: &Context) -> (Option<Profile>, Option<Region>) {
```
matchai, line 49: Looks like there might be a typo here and in future uses of the parameter:
```suggestion
fn get_aws_profile_and_region(context: &Context) -> (Option<Profile>, Option<Region>) {
```

src/modules/directory.rs
```diff
@@ -355,15 +368,15 @@ mod tests {
 
     #[test]
     fn fish_style_with_no_contracted_path() {
-        // `truncatation_length = 2`
+        // `truncation_length = 2`
```
matchai, line 371: 👍

src/modules/dotnet.rs
```diff
@@ -135,8 +135,8 @@ fn get_tfm_from_project_file(path: &Path) -> Option<String> {
     None
 }
 
-fn estimate_dotnet_version<'a>(
```
matchai, line 138: 👍

src/modules/aws.rs
ATiltedTree: I mistyped the parameter and auto-completion hid the typo from me. Fixed throughout the whole function.

matchai:
Should be ready to merge as soon as the conflicts are addressed.

ATiltedTree:
> Should be ready to merge as soon as the conflicts are addressed.

Merged master and fixed the new `gcloud` and `shlvl` modules.

ATiltedTree:
Also something I notices while writing this PR:
We run tests twice. Once for the "binary" starship and once for the "library" starship.
This is because we have a `main.rs` and a `lib.rs`. Maybe we should remove `lib.rs` to at least reduce the flakiness of our tests because it's not used for it's actual purpose?

davidkna:
> 
> 
> Also something I notices while writing this PR:
> We run tests twice. Once for the "binary" starship and once for the "library" starship.
> This is because we have a `main.rs` and a `lib.rs`. Maybe we should remove `lib.rs` to at least reduce the flakiness of our tests because it's not used for it's actual purpose?

`nushell` is probably using that `lib.rs`. `cargo test --bin starship` should suffice to solve the issue of running tests twice.

PR #1496, opened 2020-07-21
refactor(java): parse version using regex

vladimyr:
#### Description
Swap Java version parsing logic with simple regex. Mock version retrieval & extend module test suite with rendering tests.

#### Motivation and Context

1. Simplifying codebase by swapping handmade parser with regex
2. Increasing coverage by testing actual rendering

#### Checklist:


andytom:
Hi @vladimyr, If I remember correctly from when this was originally implemented a custom parser was used rather than a regex because each different "flavour" of java display their versions. I'm not sure how easy it is going to be to write a regex that supports all that. I don't know if you want to keep persevering or not but I think it is worth adding the java tests if you don't want to continue down the regexp path.

vladimyr:
Hey, @andytom before I did this I actually went through original PR #507 (and issue it addressed #468) to make sure I'm familiar with the context and that I got everything right.

> this was originally implemented a custom parser was used rather than a regex because each different "flavour" of java display their versions.

As @chipbuster stated back then https://github.com/starship/starship/issues/468#issuecomment-537616740 this can be solved either by using regex or utilizing custom parser. Original author (@yuri1969) took parser route but that doesn't disqualify regex as a valid option.

> I'm not sure how easy it is going to be to write a regex that supports all that.

It is that easy: `(?P<version>[\d\.]+)[^\s]+\s(?:built|from)` 🙃 
I reused the existing test suite and all checks are still green.

> but I think it is worth adding the java tests if you don't want to continue down the regexp path.

I'm not sure I understand this bit, what do you mean by _adding the java tests_?
As I said I practically haven't touched tests at all. The only change on that front is adding the output of `java -Xinternalversion` to mocked commands.

andytom:
> Hey, @andytom before I did this I actually went through original PR #507 (and issue it addressed #468) to make sure I'm familiar with the context and that I got everything right.
> 
> > this was originally implemented a custom parser was used rather than a regex because each different "flavour" of java display their versions.
> 
> As @chipbuster stated back then [#468 (comment)](https://github.com/starship/starship/issues/468#issuecomment-537616740) this can be solved either by using regex or utilizing custom parser. Original author (@yuri1969) took parser route but that doesn't disqualify regex as a valid option.
> 
> > I'm not sure how easy it is going to be to write a regex that supports all that.
> 
> It is that easy: `(?P<version>[\d\.]+)[^\s]+\s(?:built|from)` upside_down_face
> I reused the existing test suite and all checks are still green.

Sorry, this was caused by me being a bit lazy when I looked at the test failures. I was assumed that there was something wrong with the regexp since the tests were failing but it actually looks like a problem with the command not being mocked and falling back to the system Java. It looks like the regex is much simpler. :+1: 

> > but I think it is worth adding the java tests if you don't want to continue down the regexp path.
> 
> I'm not sure I understand this bit, what do you mean by _adding the java tests_?
> As I said I practically haven't touched tests at all. The only change on that front is adding the output of `java -Xinternalversion` to mocked commands.

Sorry this is me not explaining it very well. I was trying to suggest that if you didn't want to fiddle with the regexp to get it work then it would still be worth keeping the mocked tests that you had added.

src/utils.rs
```diff
@@ -74,6 +74,10 @@ Elixir 1.10 (compiled with Erlang/OTP 22)\n",
             stdout: String::from("v3.1.1+gafe7058\n"),
             stderr: String::default(),
         }),
+        "java -version" => Some(CommandOutput {
```
andytom, line 77: I think this might need to be changed since the actual command that is run depends on if the  `JAVA_HOME` env var is set or not. So this command could either be `${JAVA_HOME}/bin/java -Xinternalversion` or `java -Xinternalversion`.

Maybe something like 
```suggestion
        s if s.ends_with("java -Xinternalversion" => Some(CommandOutput {
```

vladimyr: Oh, this is really silly mistake from my end 🤦 
Thanks for catching it 👍

andytom:
Ok, now I have no idea why the tests are failing now. They work locally for me and also they work in GH Actions when I pushed your branch to my fork with not changes https://github.com/andytom/starship/runs/916499966.

vladimyr:
@andytom That makes two of us. It also works for me locally. I shuffled between installed JDKs to make sure mock is being used and verified that. Why this fails is still a mystery beyond me 🕵️

andytom:
> @andytom That makes two of us. It also works for me locally. I shuffled between installed JDKs to make sure mock is being used and verified that. Why this fails is still a mystery beyond me detective

And now the tests have started passing :confused:

andytom:
As always @vladimyr thanks for the contribution and dealing with the broken CI and my comments :D.

vladimyr:
It is a pleasure, as always 😉

PR #1497, opened 2020-07-21
refactor: unify version parsing & formatting

vladimyr:
#### Description

Unify version parsing & formatting across modules.

#### Motivation and Context

Make codebase more uniform and easier to maintain.

#### Checklist:

### DIFF
--- a/src/modules/cmake.rs
+++ b/src/modules/cmake.rs
@@ -18,6 +18,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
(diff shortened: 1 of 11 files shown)


andytom (changes requested):
I think we should actually go the opposite way with this and execute the version commands lazily where possible. If we make the changes in the PR then even if the version is not used in the format string we still execute the command. This is particularly a problem for the `terraform` module as the `terraform version` command takes a loooong time depending on the number of plugins (see #2084).

andytom:
I'm going to close this since, we are going to do a lazy evaluation. If you would like to help with this the umbrella issue is #2111.

PR #1499, opened 2020-07-21
perf: Only check timeout every 256 files

aquarhead:
#### Description

When listing directory content, only check timeout every 256 files

#### Motivation and Context

I noticed when listing directory content, the iterator will check system time against timeout for every single item. I thought changing it to check once in a while instead would be faster for large directories. I tried to compare timing locally for a large directory and it decreased the time by about 40-45% so I decided to open this PR to see if this is actually a good idea.

Note I actually use a stripped down fork from a quite old version (0.25.2) though it looks like this particular code hasn't changed that much between now and then. (but the log messages are different it seems)

I tested this change in a directory with 1685 files, without the change the logs read:

```
 TRACE starship::context             > Building a vector of directory files took 4.948ms
```

and with the change:

```
 TRACE starship::context             > Building a vector of directory files took 2.841ms

 TRACE starship::context             > Building a vector of directory files took 2.828ms

 TRACE starship::context             > Building a vector of directory files took 2.882ms

 TRACE starship::context             > Building a vector of directory files took 2.963ms
```

An alternative method is to use `.trailing_zeros()` instead of bitwise operations, which has the benefit of skipping the first entry as well, though the first interval is not consistent - say using `trailing_zeros() != 8` then the first check happens for index 256 (from 0) but after that it checks every 512 entries. Both methods provide similiar time though so I'm using the more "consistent" one in this PR.

I've also tried a bunch of other numbers than `8` and most others seem to provide (tiny bit) worse or rather identical result.

I'm not sure if there're existing benchmark for this particular scenario for `starship`? Or should I setup a more comprehensive test to show improvements?

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -204,8 +204,12 @@ impl DirContents {
         let mut extensions: HashSet<String> = HashSet::new();
 
         fs::read_dir(base)?
-            .take_while(|_| SystemTime::now().duration_since(start).unwrap() < timeout)
-            .filter_map(Result::ok)
+            .enumerate()
+            .take_while(|(n, _)| {
+                n & 0xFF != 0 // only check SystemTime once every 2^8 entries
+                || SystemTime::now().duration_since(start).unwrap() < timeout
+            })
+            .filter_map(|(_, entry)| entry.ok())
             .for_each(|entry| {
                 let path = PathBuf::from(entry.path().strip_prefix(base).unwrap());
                 if entry.path().is_dir() {


davidkna:
The time-out was added for slow filesystems. I feel like 256 files is quite high for that. Maybe a smaller threshold is more appropriate?
You might also want to check if `% n` has the same speed as this bitwise operation, because it could be optimized for this special case. Also I'm not sure if this works on big endian (we had someone use starship on PowerPC).

aquarhead:
> You might also want to check if % n has the same speed as this bitwise operation, because it could be optimized for this special case.

Wouldn't `%` be optimized into bitwise operations?

> Also I'm not sure if this works on big endian (we had someone use starship on PowerPC).

Switching to `.trailing_zeros()` might be better then? It seems [it uses CPU intrinsics](https://doc.rust-lang.org/src/core/num/mod.rs.html#2685-2687).

> I feel like 256 files is quite high for that.

I think doing `n.trailing_zeros() != 3` also works just fine with similiar timing, but I only have my local testing result, not a comprehensive benchmark to compare performance for all scenarios.

davidkna:
> Wouldn't `%` be optimized into bitwise operations?

Maybe. It's more obvious what  `%` does and it should work on big endian while being just as efficient if it's optimized into bitwise operations.

> Switching to `.trailing_zeros()` might be better then? It seems [it uses CPU intrinsics](https://doc.rust-lang.org/src/core/num/mod.rs.html#2685-2687).

Not sure if it works. The docs talk about the number of trailing zeros in the "binary representation" which could refer to the in-memory representation.

davidkna:
On second thought `& 0xFF` should work on both endians.

whatisaphone:
That code should be using `Instant::now()`, not `SystemTime`. `SystemTime` is not monotonic, meaning the time could go backwards. The `unwrap()` in that code could panic if the system syncs with an NTP server, or daylight savings time happens, or the user is on a train traveling west, etc. [See docs](https://doc.rust-lang.org/std/time/struct.Instant.html)

And since this issue is about perf, I would guess that `Instant::now` is faster than `SystemTime::now`, since it doesn't have to sync with the system clock.

aquarhead:
I agree but that's outside the scope of this PR.

andytom:
Thanks for the suggestion @whatisaphone I've created an issue so that this doesn't get forgotten about.

andytom:
@aquarhead thank you for your contribution and @davidkna thank you for your help reviewing.

PR #1506, opened 2020-07-23
Do not try to analyze write permissions on network locations

### LINKED ISSUE #1476: directory(windows) lock symbol is always shown on SMB network shares
#### Current Behavior

```
\\NAS\MY-SHARE🔒
❯
```

Even though it is writable and `New-Item -Type File -Path \\NAS\MY-SHARE\x` works.

#### Expected Behavior

Lock icon is not shown when directory is writable.
```
\\NAS\MY-SHARE
❯
```

#1298

It works correctly when the same network share is mapped to a drive: `
```
Y:
❯ 

```
#### Environment
- Starship version: 0.44.0 (master)
- powershell version: 7.0.2
- Operating system: Windows Windows 10 10.0.19041
- Terminal emulator: Microsoft Terminal <unknown version>

```bash
#!/usr/bin/env pwsh

$ENV:EDITOR="nvim"
$ENV:VISUAL="code --wait"

$ENV:RUSTFLAGS="-Ctarget-cpu=native"

$ENV:PYTHONIOENCODING='UTF-8'

Set-Alias -Name ls -Value "lsd"
function l { lsd -l $args }
function la { lsd -lA $args }
function tree { lsd --tree $args }
Set-Alias -Name vi -Value "nvim"
Set-Alias -Name vim -Value "nvim"

Import-Module PSReadLine
Set-PSReadlineKeyHandler -Key Tab -Function Complete
Set-PSReadlineKeyHandler -Key UpArrow -Function HistorySearchBackward
Set-PSReadlineKeyHandler -Key DownArrow -Function HistorySearchForward

&~/.cargo/bin/lolcow-fortune cowsay -fmoose -L
&starship init powershell | Invoke-Expression
```

#### Starship Configuration

```toml
<unknown config>
```

Aelnor:
#### Description
Do not try to analyze if the current process can write network location on Windows.

#### Motivation and Context
Unfortunately, there's no way on Windows to tell if we can write a network location because it's not being controlled by the OS itself. Thus now the lock symbol is never shown on network locations.
This PR introduces a new unsafe call.
Closes #1476

#### Checklist:

### DIFF
--- a/src/modules/utils/directory_win.rs
+++ b/src/modules/utils/directory_win.rs
@@ -28,6 +28,10 @@ pub fn is_write_allowed(folder_path: &str) -> std::result::Result<bool, &'static
         .collect();
 
     if is_network_path(&folder_name) {
+        log::info!(
+            "Directory '{:?}' is a network drive, unable to check write permissions. See #1506 for details",
+            folder_path
+        );
         return Ok(true);
     }
 
@@ -122,6 +126,7 @@ pub fn is_write_allowed(folder_path: &str) -> std::result::Result<bool, &'static
     Ok(result != 0)
 }
 
+#[link(name = "Shlwapi")]
 extern "system" {
     fn PathIsNetworkPathW(pszPath: LPCWSTR) -> BOOLEAN;
 }
--- a/src/modules/utils/directory_win.rs
+++ b/src/modules/utils/directory_win.rs
@@ -10,21 +10,27 @@ use winapi::um::handleapi;
 use winapi::um::processthreadsapi;
 use winapi::um::securitybaseapi;
 use winapi::um::winnt::{
-    SecurityImpersonation, DACL_SECURITY_INFORMATION, FILE_ALL_ACCESS, FILE_GENERIC_EXECUTE,
-    FILE_GENERIC_READ, FILE_GENERIC_WRITE, GENERIC_MAPPING, GROUP_SECURITY_INFORMATION, HANDLE,
-    OWNER_SECURITY_INFORMATION, PRIVILEGE_SET, PSECURITY_DESCRIPTOR, STANDARD_RIGHTS_READ,
-    TOKEN_DUPLICATE, TOKEN_IMPERSONATE, TOKEN_QUERY,
+    SecurityImpersonation, BOOLEAN, DACL_SECURITY_INFORMATION, FILE_ALL_ACCESS,
+    FILE_GENERIC_EXECUTE, FILE_GENERIC_READ, FILE_GENERIC_WRITE, GENERIC_MAPPING,
+    GROUP_SECURITY_INFORMATION, HANDLE, LPCWSTR, OWNER_SECURITY_INFORMATION, PRIVILEGE_SET,
+    PSECURITY_DESCRIPTOR, STANDARD_RIGHTS_READ, TOKEN_DUPLICATE, TOKEN_IMPERSONATE, TOKEN_QUERY,
 };
 
 /// Checks if the current user has write access right to the `folder_path`
 ///
 /// First, the function extracts DACL from the given directory and then calls `AccessCheck` against
 /// the current process access token and directory's security descriptor.
+/// Does not work for network drives and always returns true
 pub fn is_write_allowed(folder_path: &str) -> std::result::Result<bool, &'static str> {
     let folder_name: Vec<u16> = OsStr::new(folder_path)
         .encode_wide()
         .chain(iter::once(0))
         .collect();
+
+    if is_network_path(&folder_name) {
+        return Ok(true);
+    }
+
     let mut length: DWORD = 0;
 
     let rc = unsafe {
@@ -114,4 +120,12 @@ pub fn is_write_allowed(folder_path: &str) -> std::result::Result<bool, &'static
     }
 
     Ok(result != 0)
+}
+
+extern "system" {
+    fn PathIsNetworkPathW(pszPath: LPCWSTR) -> BOOLEAN;
+}
+
+fn is_network_path(folder_path: &Vec<u16>) -> bool {
+    return unsafe { PathIsNetworkPathW(folder_path.as_ptr()) } == 1;
 }
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -107,7 +107,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             .map(|variable| match variable {
                 "path" => Some(Ok(&final_dir_string)),
                 "read_only" => {
-                    if is_readonly_dir(current_dir.to_str()?) {
+                    if is_readonly_dir(context.current_dir.to_str()?) {
                         Some(Ok(&lock_symbol))
                     } else {
                         None


davidkna:
How about just creating our own binding for `PathIsNetworkPath` instead?

```rust
use winapi::um::winnt::{BOOLEAN, LPCWSTR};
use std::ffi::OsStr;
use std::os::windows::ffi::OsStrExt;

extern "system" {
    fn PathIsNetworkPathW(
        pszPath: LPCWSTR,
    ) -> BOOLEAN;
}

fn is_network_path(path: &str) -> bool {
    let wpath: Vec<u16> = OsStr::new(path)
        .encode_wide()
        .chain(Some(0u16).into_iter())
        .collect();
    
    return unsafe {
        PathIsNetworkPathW(wpath.as_ptr()) == 1
    }
}
```
and `println!("cargo:rustc-link-lib=Shlwapi");` in `build.rs`.

Aelnor:
> 
> 
> How about just creating our own bindings for `PathIsNetworkPath` instead?
> 
> ```rust
> use winapi::um::winnt::{BOOLEAN, LPCWSTR};
> use std::ffi::OsStr;
> use std::os::windows::ffi::OsStrExt;
> use std::iter;
> 
> extern "system" {
>     fn PathIsNetworkPathW(
>         pszPath: LPCWSTR,
>     ) -> BOOLEAN;
> }
> 
> fn is_network_path(path: &str) -> bool {
>     let wpath: Vec<u16> = OsStr::new(path)
>         .encode_wide()
>         .chain(iter::once(0))
>         .collect();
>     
>     return unsafe {
>         PathIsNetworkPathW(wpath.as_ptr()) == 1
>     }
> }
> ```
> 
> and `println!("cargo:rustc-link-lib=Shlwapi");` in `build.rs`.
> 
> Also wouldn't it be better if the whole directory helpers used `Path(Buf)` directly instead of `&str` to avoid duplicate conversions?

Thank you, will rework

chipbuster:
A minor request: can we add a info-level log statement into the network path return path? This way, in case someone is confused when not seeing the symbol on a network location, they could (at least in principle) run starship with logging enabled and immediately see why.

Something like this, though feel free to change it.
```
`log::info!("Cannot check write permissions on network drive, see #1506 for details");
```

davidkna:
It looks like it works right now, but having a `build.rs` with `println!("cargo:rustc-link-lib=Shlwapi");` would be better.
One of the dependencies is likely already linking against `Shlwapi`, but that could change in the future.

src/modules/utils/directory_win.rs
```diff
@@ -115,3 +121,11 @@ pub fn is_write_allowed(folder_path: &str) -> std::result::Result<bool, &'static
 
     Ok(result != 0)
 }
+
+extern "system" {
```
davidkna, line 125:
```suggestion
#[link(name = "Shlwapi")]
extern "system" {
```

Please add this to ensure `Shlwapi` is linked.

Aelnor:
> 
> 
> A minor request: can we add a info-level log statement into the network path return path? This way, in case someone is confused when not seeing the symbol on a network location, they could (at least in principle) run starship with logging enabled and immediately see why.
> 
> Something like this, though feel free to change it.
> 
> ```
> log::info!("Cannot check write permissions on network drive, see #1506 for details");
> ```

Done

davidkna:
I tried to test this and it was working and showing that the directory is readable, but it did not hit the new `is_network_path` and the log statement check at all. I'm not sure why it's working now despite that.

I investigated and instead of using `\\NAS\MY-SHARE` (`--path` arg or `env::current_dir`) the code is working with the logical powershell path `Microsoft.PowerShell.Core\FileSystem::\\NAS\MY-SHARE` (`PWD` env var).

Replacing `if is_readonly_dir(current_dir.to_str()?) {` with `if is_readonly_dir(context.current_dir.to_str()?) {` in `directory.rs` makes it work without any of these changes.

Aelnor:
> 
> 
> I tried to test this and it was working and showing that the directory is readable, but it did not hit the new `is_network_path` and the log statement check at all. I'm not sure why it's working now despite that.
> 
> I investigated and instead of using `\\NAS\MY-SHARE` (`--path` arg or `env::current_dir`) the code is working with the logical powershell path `Microsoft.PowerShell.Core\FileSystem::\\NAS\MY-SHARE` (`PWD` env var).
> 
> Replacing `if is_readonly_dir(current_dir.to_str()?) {` with `if is_readonly_dir(context.current_dir.to_str()?) {` in `directory.rs` makes it work without any of these changes.

Oh, my bad. Thanks for pointing it out.
But the fix you propose doesn't really work on my machine. I've logged `current_dir` and `context.current_dir` on my machine and they're the same and both are wrong (

And if we pass the correct path (for example, '\\server\share'), `AccessCheck` will report the path is read-only

Another issue here is the lock symbol is being printed despite `is_readonly_dir` in directory.rs returning false. And it does not reproduce when I'm using "starship.exe module directory".

davidkna:
> Oh, my bad. Thanks for pointing it out.
> But the fix you propose doesn't really work on my machine. I've logged `current_dir` and `context.current_dir` on my machine and they're the same and both are wrong ("Microsoft.PowerShell.Core\FileSystem::\server\share")

Unlike plain `starship init powershell` sets `--path` to `Convert-Path $PWD`.
Please try again with `starship prompt --path "$(Convert-Path $PWD)"`.

> And if we pass the correct path (for example, '\server\share'), `AccessCheck` will report the path is read-only

Now the access check is always failing for me again on SMB. I must have made a mistake earlier.

> Another issue here is the lock symbol is being printed despite `is_readonly_dir` in directory.rs returning false. And it does not reproduce when I'm using "starship.exe module directory".

Do you have an example to reproduce?

Aelnor:
> 
> 
> > Oh, my bad. Thanks for pointing it out.
> > But the fix you propose doesn't really work on my machine. I've logged `current_dir` and `context.current_dir` on my machine and they're the same and both are wrong ("Microsoft.PowerShell.Core\FileSystem::\server\share")
> 
> Unlike plain `starship prompt` `starship init powershell` sets `--path` to `Convert-Path $PWD`.
> Please try again with `starship prompt --path "$(Convert-Path $PWD)"`.
> 
> > And if we pass the correct path (for example, '\server\share'), `AccessCheck` will report the path is read-only
> 
> Now the access check is always failing for me again on SMB. I must have made a mistake earlier.
> 
> > Another issue here is the lock symbol is being printed despite `is_readonly_dir` in directory.rs returning false. And it does not reproduce when I'm using "starship.exe module directory".
> 
> Do you have an example to reproduce?

Disregard that :) I thought, `starship init powershell` is similar to simple `starship.exe module directory`, but it is not.

Everything is okay, it looks like we need both your and my changes in order to fix the network issue.

Aelnor:
It seems, rebase broke the tests. It is a known issue?

davidkna:
Seems to be something new but not related to this PR. It also popped up in #1518 and #1527.

davidkna:
Please try rebasing on master. #1532 might have fixed the build error.s

chipbuster:
Looks like we're getting smacked with the hammer of (likely) race conditions in our test suite again. I'll see if I can get a successful build through just by repeatedly requesting reruns.

chipbuster:
And we are all green. Thanks @Aelnor and @davidkna for all the work to fix this issue!

PR #1509, opened 2020-07-24
fix: fix compile with battery feature disabled

### LINKED ISSUE #1508: Won't compile without battery feature
This doesn't open in the correct browser and is not specific to my configuration

#### Current Behavior
Building without `battery` feature fails because of an incorrectly sized CONST array. Problem comes from the lines:
```
    #[cfg(feature = "battery")]
    "battery",
```

#### Expected Behavior
You should be able to build without the feature if the feature is going to exist

```
error[E0308]: mismatched types
  --> src/configs/starship_root.rs:14:38
   |
14 |   pub const PROMPT_ORDER: [&str; 44] = [
   |  ______________________________________^
15 | |     "username",
16 | |     "hostname",
17 | |     "singularity",
...  |
62 | |     "character",
63 | | ];
   | |_^ expected an array with a fixed size of 44 elements, found one with 43 elements
```

#### Possible Solution
Set a CONST for size somewhere before this based on the `battery` feature. This will probably be useful in the future if more modules become features.

#### Environment
- Starship version: current master

davidkna:
#### Description
Borrow `PROMPT_ORDER` so the length no longer needs to be set.

#### Motivation and Context

Closes #1508 

#### Checklist:

### DIFF
--- a/src/configs/starship_root.rs
+++ b/src/configs/starship_root.rs
@@ -11,7 +11,7 @@ pub struct StarshipRootConfig<'a> {
 // List of default prompt order
 // NOTE: If this const value is changed then Default prompt order subheading inside
 // prompt heading of config docs needs to be updated according to changes made here.
-pub const PROMPT_ORDER: [&str; 44] = [
+pub const PROMPT_ORDER: &[&str] = &[
     "username",
     "hostname",
     "singularity",


xaocon:
Thanks for the quick work. I'm a novice, didn't know a const could be a borrow.

PR #1511, opened 2020-07-25
feat(package):  Add Maven pom.xml version support

tarka:
### Description

Adds ability to extract version information from a Maven pom.xml if present.

#### Motivation and Context

Starship recognises the existence of Maven projects elsewhere (e.g. Java module), but does not extract the version for the package module. This adds that feature.

Testing is via the inline Rust tests.

#### Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,10 @@ Cargo.lock
 # Vim swap files
 *.swp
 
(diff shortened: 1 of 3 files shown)


andytom:
@tarka thank you for your contribution.

PR #1525, opened 2020-07-27
feat(package): Add support for Helm charts

### LINKED ISSUE #1487: Add Helm package support to the package module

In #1443 we added support for displaying the Helm version. Helm is a package manager and it would be nice to display the version of a chart from the `Chart.yaml`. 
  
This should just be a case of updating [`src/modules/package.rs`](https://github.com/starship/starship/blob/master/src/modules/package.rs) to parse the `Chart.yaml`, if one exists, and return the version along with updating the docs and tests.

simao:
#### Description
Parses a `Chart.yaml` file in the current directory and uses that as the current package version

#### Motivation and Context
Closes #1487

I added a unit test.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1575,7 +1575,7 @@ format = "via [🤖 $version](bold green) "
 
 The `package` module is shown when the current directory is the repository for a
 package, and shows its current version. The module currently supports `npm`, `cargo`,
-`poetry`, `composer`, `gradle`, `julia` and `mix` packages.
+`poetry`, `composer`, `gradle`, `julia`, `mix` and `helm` packages.
 
 - **npm** – The `npm` package version is extracted from the `package.json` present
   in the current directory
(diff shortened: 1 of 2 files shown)


simao:
I only changed the English docs, should I also change all the other languages? That part is only translated into English, even in the other files.

Thanks

andytom:
> I only changed the English docs, should I also change all the other languages? That part is only translated into English, even in the other files.
> 
> Thanks

Hi @simao, thanks you for the contribution. For the docs you just need to update the English version, we use CrowdIn to manage the translations. P.S we are always looking for new translators if you can help out https://translate.starship.rs/

andytom:
The failed CI test seems to be cause by some dependency issues can you give this a rebase to pick up the new changes from dependabot as I think that should fix it.

simao:
Thanks for reviewing my pr @andytom 

Seems the ci error is still not solved, anything else I could try? 

I also changed my code a bit to use `.first()` instead of `[0]`, as I think that could cause a panic.

andytom:
> Thanks for reviewing my pr @andytom
> 
> Seems the ci error is still not solved, anything else I could try?

I don't have any thing else to try yet but I'll have a look and see if I can work out what is going wrong.

simao:
@andytom Thanks a lot for fixing CI. I rebased and now it's green.

andytom:
@simao, thank you for the contribution sorry about the problems with the CI system.

PR #1526, opened 2020-07-27
docs: Add set commands to advanced config doc

### LINKED ISSUE #1524: DEBUG trap not preserved in bash
#### Current Behavior
DEBUG trap set in .bashrc is not being preserved by Starship. At the end of .bashrc the command `eval "$(starship init bash)"` results in the following command being run: `source <("/usr/local/bin/starship" init bash --print-full-init)`

`source` does not inherit the DEBUG trap by default:

> If the -T option is enabled, source inherits any trap on DEBUG; if it is not,  any  DEBUG trap  string  is saved and restored around the call to source, and source unsets the DEBUG trap while it executes.

Since `source` does not inherit the DEBUG trap, `$dbg_trap` is not being set and Starship's trap is set up to use `starship_preexec` instead of `starship_preexec_all`

#### Expected Behavior
Based on documentation [here](https://starship.rs/advanced-config/) I expected the DEBUG trap set before initializing Starship to be preserved.

#### Possible Solution
I've worked around the problem for now by simply copying the code output by `starship init bash --print-full-init` into the end of my .bashrc instead of the line `eval "$(starship init bash)"`

#### Environment
- Starship version: 0.44.0

- Terminal emulator: MATE Terminal 1.20.0

```bash
# for examples

clean() {
  unset HISTFILE
  exit
}

case $- in
    *i*) ;;
      *) return;;
esac

HISTCONTROL=ignoreboth

shopt -s histappend

HISTSIZE=1000
HISTFILESIZE=2000

shopt -s checkwinsize

#shopt -s globstar

fi

case "$TERM" in
esac

#force_color_prompt=yes

	color_prompt=yes
    else
	color_prompt=
    fi
fi

else
fi

case "$TERM" in
xterm*|rxvt*)
    ;;
*)
    ;;
esac

    alias ls='ls --color=auto'
    #alias dir='dir --color=auto'

fi

# some more ls aliases
alias ll='ls -alF'
alias la='ls -A'
alias l='ls -CF'

#   sleep 10; alert

# Alias definitions.

    . ~/.bash_aliases
fi

# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
    . /etc/bash_completion
  fi
fi

# turn on numlock
numlockx

# add Pulumi and Go to the PATH
export PATH=$PATH:$HOME/.pulumi/bin
export GOPATH=$HOME/go
export PATH=$PATH:$GOPATH/bin:/usr/local/go/bin

function settitle() {
  if [[ -z $AT_PROMPT ]]; then
    return
  fi
  unset AT_PROMPT

  echo -ne "\033]2;$(history 1 | sed 's/^[ ]*[0-9]*[ ]*//g')\007"
}
trap settitle DEBUG

#FIRST_PROMPT=1
function blastoff() {
  AT_PROMPT=1

#  if [[ -n $FIRST_PROMPT ]]; then
#    unset FIRST_PROMPT
#    return
#  fi

  INCOGNITO=''
  echo -ne "\033]2;Terminal\007"
  if [[ -z $HISTFILE ]]; then
    INCOGNITO='[incognito]'
  fi
  if [[ $STATUS -ne 0 ]]; then
    echo -en "\n\033[01;31m$STATUS\033[01;32m $INCOGNITO\033[00m"
  else
    if [[ ! -z $INCOGNITO ]]; then
      echo -ne "\033]2;Terminal $INCOGNITO\007"
      echo -ne "\n\033[01;32m$INCOGNITO\033[00m"
    fi
  fi
}
starship_precmd_user_func="blastoff"

eval "$(starship init bash)"
```

#### Starship Configuration

```bash
[aws]
prefix = "in "

[character]
use_symbol_for_status = true

[directory]
truncation_length = 9

[docker_context]
disabled = false
only_with_files = true

[git_branch]
symbol = " "
truncation_length = 32

[git_commit]
commit_hash_length = 10

# Don't want to change anything, but want to see it in action at some point
[git_state]

[git_status]
show_sync_count = true

[golang]
symbol = " "
style = "bold white"

[kubernetes]
disabled = false

[memory_usage]
disabled = false

[python]
python_binary = "python3"

[terraform]
disabled = true
```

jsnider-mtu:
#### Description
DEBUG trap mechanism in bash as described [here](https://starship.rs/advanced-config/) did not work without setting the shell option `functrace` first. This PR updates that example code so that it works correctly.

#### Motivation and Context

Closes #1524 

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -39,7 +39,9 @@ function blastoff(){
     echo "🚀"
 }
 trap blastoff DEBUG     # Trap DEBUG *before* running starship
+set -o functrace
 eval $(starship init bash)
+set +o functrace
 ```
 
 ## Change Window Title


chipbuster (approved):
:+1: LGTM. Dunno how this slipped through the review net.

Something to note is that this is going to break if the user has already `set -o functrace` or launched their shell with `-T`, but at some point we can't actually afford to document every single thing that could possibly go wrong. I think this is a good balance between the two points.

andytom:
Thank you for your contribution @jsnider-mtu and thanks for surfacing this @chipbuster.

PR #1540, opened 2020-07-30
ci: Retry failed CI jobs

andytom:
#### Description

To try to get around our flaky CI tests have added a quick and dirty
retry.

#### Motivation and Context

This is just a quick test to see what happens as part of our investigation into #886 

#### Checklist:


andytom:
Note this causes all tests that run on Windows to be filters out so no tests are run so this should not be merged.

PR #1541, opened 2020-07-31
fix(bash): init breaks if PROMPT_COMMAND includes starship_precmd as an intermediate command

jawang35:
#### Description
If `PROMPT_COMMAND` includes `starship_precmd` as an intermediate command and `eval "$(starship init bash)` is executed again, then `PROMPT_COMMAND` enters an infinite loop of `starship_precmd` executions. This happens when other tools also hook into the shell using `PROMPT_COMMAND`. In particular this happens with [z](https://github.com/rupa/z/). For example,

```sh
❯ echo $PROMPT_COMMAND
_direnv_hook;starship_precmd
_z --add "$(command pwd -P 2>/dev/null)" 2>/dev/null;
```

#### Motivation and Context
Ideally the bash init would be idempotent so that the user can run `eval "$(starship init bash)"` without worrying about crashing the shell. It's possible I'm missing some context and the condition is intentionally checking for `starship_precmd` to be last in `PROMPT_COMMAND`.

#### Checklist:

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -71,7 +71,7 @@ else
     # add multiple instances of the starship function and keep other user functions if any.
     if [[ -z "$PROMPT_COMMAND" ]]; then
         PROMPT_COMMAND="starship_precmd"
-    elif [[ "$PROMPT_COMMAND" != *"starship_precmd" ]]; then
+    elif [[ "$PROMPT_COMMAND" != *"starship_precmd"* ]]; then
         # Appending to PROMPT_COMMAND breaks exit status ($?) checking.
         # Prepending to PROMPT_COMMAND breaks "command duration" module.
         # So, we are preserving the existing PROMPT_COMMAND


andytom:
> It's possible I'm missing some context and the condition is intentionally checking for `starship_precmd` to be last in `PROMPT_COMMAND`.

I think but not 100% sure the logic behind checking if starship is the last command is left over from when we were appending `starship_precmd` to the `PROMPT_COMMAND` which was changed in #1185. As @chipbuster mentioned on that PR we have jumped through a lot of hoops to try to make Starship work well on bash and be a good citizen and not break any other tools. Logically this change makes perfect sense to me but I'm not sure if there are any edge cases I'm missing.

jawang35:
@andytom it looks like this specific check was written in https://github.com/starship/starship/pull/696. This problem is also mentioned specifically and this line of code seems like it was written to address that:

> In addition to that, I added a check to avoid the concatenation of multiple sharship_cmd function ones after the other in the case of multiple calls to eval "$(starship init bash)".

But it looks like it was assuming that `starship_precmd` would always be at the end which has changed since https://github.com/starship/starship/pull/1185.

I've been using this in my bash shell with multiple hooks and precommands with no issues but I can't say for certain all edge cases have been tested. I understand the caution given how fiddly the bash init is compared to what I can see in the other shell init scripts. 🤷‍♂️

andytom:
@jawang35 thank you for your contribution.

PR #1544, opened 2020-08-01
 refactor(directory/lock): fix clippy, replace str with Path

davidkna:
#### Description

I noticed the recent changes to the directory module in #1506 introduced some clippy errors on windows.
When reviewing I didn't run `cargo clippy` because I thought CI would catch clippy errors, but they weren't caught there because clippy isn't running on windows in CI.

I also replaced the use of `str` in the directory lock helpers with `Path` because the module work with `Path`/`OsStr` internally anyway. This avoids coversion to UTF-8 and back and allows it to work on directories that aren't valid unicode.

#### Motivation and Context

Closes #

#### Checklist:


andytom:
Thanks for your contribution @davidkna, I'm going to update the CI to run Clippy on all OSs to catch these in the future.

PR #1546, opened 2020-08-02
docs: Update example configto remove old var

andytom:
#### Description

Have updated the example toml config to remove an old var and replace it
with the updated replacement.

#### Motivation and Context

Keeping the docs updated

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -17,7 +17,7 @@ All configuration for starship is done in this [TOML](https://github.com/toml-la
 
 ```toml
 # Don't print a new line at the start of the prompt
-add_newline = false
+format = "$all"
 
 # Replace the "❯" symbol in the prompt with "➜"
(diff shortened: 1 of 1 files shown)


matchai:
As per the discussion [here](https://github.com/starship/starship/discussions/1116#discussioncomment-7792) I thought we were going to retain the same linebreak config option. Is that not how it works today?

andytom:
> As per the discussion [here](https://github.com/starship/starship/discussions/1116#discussioncomment-7792), I thought we were going to retain the same linebreak config option. Is that not how it works today?

The `add_newline` option was deprecated in #1056, looks like thins might have just been something that slipped through the cracks. Should we add this back in or leave it as a breaking change (note there are a few others listed in #1404)?

matchai:
I think it would be best if we readded it. Removing the linebreak is common enough configuration to justify it being a boolean config option.

PR #1548, opened 2020-08-02
test: sync filesystem before each test is run

davidkna:
#### Description

[Apparently `.sync_all()` might not be enough to ensure new files are visible in the filesystem.](https://github.com/starship/starship/issues/886#issuecomment-667509181)
In the two test module render helpers I found, I added code to sync the filesystem before the modules are rendered. ~~This doesn't work on windows but I don't really remember the failures happening there.~~
I did five ci runs on my fork and they all worked except for unrelated failures reminiscent of #1532  

#### Motivation and Context

#886

#### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,17 @@ attohttpc = { version = "0.15.0", optional = true, default-features = false, fea
 native-tls = { version = "0.2", optional = true }
 
 [target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3", features = ["winuser", "securitybaseapi", "processthreadsapi", "handleapi", "impl-default"]}
+winapi = { version = "0.3", features = [
+  "winuser",
+  "securitybaseapi",
+  "processthreadsapi",
+  "handleapi",
+  "fileapi",
+  "errhandlingapi",
+  "winbase",
+  "winnt",
+  "impl-default",
+] }
 
 [target.'cfg(not(windows))'.dependencies]
 nix = "0.18.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,10 +70,7 @@ winapi = { version = "0.3", features = [
   "securitybaseapi",
   "processthreadsapi",
   "handleapi",
-  "fileapi",
-  "errhandlingapi",
   "winbase",
-  "winnt",
   "impl-default",
 ] }
 
--- a/src/modules/utils/test.rs
+++ b/src/modules/utils/test.rs
@@ -1,7 +1,6 @@
 use crate::config::StarshipConfig;
 use crate::context::{Context, Shell};
 use std::path::Path;
-use std::process::Command;
 
 /// Render a specific starship module by name
 pub fn render_module(
@@ -21,6 +20,7 @@ pub fn render_module(
 /// Syncs directory in filesystem to disk to ensure consistent tests
 #[cfg(not(windows))]
 fn fs_sync(path: &Path) {
+    use std::process::Command;
     Command::new("sync").arg(path).status().unwrap();
 }
 
--- a/src/modules/utils/test.rs
+++ b/src/modules/utils/test.rs
@@ -1,12 +1,7 @@
 use crate::config::StarshipConfig;
 use crate::context::{Context, Shell};
-use std::fs::OpenOptions;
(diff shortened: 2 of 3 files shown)


davidkna:
[I added partial support for windows because the failures happen there too now.](https://github.com/starship/starship/issues/886#issuecomment-669149950)

ATiltedTree:
After a bit of research i think it won't really help us here because were doing exactly the same as `sync_all()`:

`sync_all`'s implementation on linux

```rust
    pub fn fsync(&self) -> io::Result<()> {
        cvt_r(|| unsafe { os_fsync(self.0.raw()) })?;
        return Ok(());


        #[cfg(any(target_os = "macos", target_os = "ios"))]
        unsafe fn os_fsync(fd: c_int) -> c_int {
            libc::fcntl(fd, libc::F_FULLFSYNC)
        }
        #[cfg(not(any(target_os = "macos", target_os = "ios")))]
        unsafe fn os_fsync(fd: c_int) -> c_int {
            libc::fsync(fd)
        }
    }
```

and on windows:

```rust
pub fn fsync(&self) -> io::Result<()> {
        cvt(unsafe { c::FlushFileBuffers(self.handle.raw()) })?;
        Ok(())
    }
```

Reference:
 - [Linux](https://github.com/rust-lang/rust/blob/ec9d5241f749978150a35f82a2358f494c5b9080/library/std/src/sys/unix/fs.rs#L733-L745)
 - [Windows](https://github.com/rust-lang/rust/blob/ec9d5241f749978150a35f82a2358f494c5b9080/library/std/src/sys/windows/fs.rs#L302-L305)

It basically does exactly the same as the implementation in this PR which is also required by the man page you referenced.

> For that an explicit fsync() on a file descriptor for the directory is also needed.

While we can try it, I don't think it will help us much.
And if it does help we should just use `sync_all()` on the directory before rendering and see if that fixes it as well.

Also something I found on the man page for [sync](https://linux.die.net/man/8/sync):

> **Notes**
> On Linux, sync is only guaranteed to schedule the dirty blocks for writing; it can actually take a short time before all the blocks are finally written. 

Maybe we are still to fast for the (probably virtual) disks on github actions to sync.

davidkna:
> After a bit of research i think it won't really help us here because were doing exactly the same as `sync_all()`:
> 
> `sync_all`'s implementation on linux
> 
> ```rust
>     pub fn fsync(&self) -> io::Result<()> {
>         cvt_r(|| unsafe { os_fsync(self.0.raw()) })?;
>         return Ok(());
> 
> 
>         #[cfg(any(target_os = "macos", target_os = "ios"))]
>         unsafe fn os_fsync(fd: c_int) -> c_int {
>             libc::fcntl(fd, libc::F_FULLFSYNC)
>         }
>         #[cfg(not(any(target_os = "macos", target_os = "ios")))]
>         unsafe fn os_fsync(fd: c_int) -> c_int {
>             libc::fsync(fd)
>         }
>     }
> ```
> 
> and on windows:
> 
> ```rust
> pub fn fsync(&self) -> io::Result<()> {
>         cvt(unsafe { c::FlushFileBuffers(self.handle.raw()) })?;
>         Ok(())
>     }
> ```
> 
> Reference:
> 
>     * [Linux](https://github.com/rust-lang/rust/blob/ec9d5241f749978150a35f82a2358f494c5b9080/library/std/src/sys/unix/fs.rs#L733-L745)
> 
>     * [Windows](https://github.com/rust-lang/rust/blob/ec9d5241f749978150a35f82a2358f494c5b9080/library/std/src/sys/windows/fs.rs#L302-L305)
> 
> 
> It basically does exactly the same as the implementation in this PR which is also required by the man page you referenced.
> 
> > For that an explicit fsync() on a file descriptor for the directory is also needed.
> 
> While we can try it, I don't think it will help us much.
> And if it does help we should just use `sync_all()` on the directory before rendering and see if that fixes it as well.

It's true that the implementation here is more or less identical to the one in rust stdlib but as far as I can tell it's not possible to `sync_all` directories with the rust stdlib. (Permission error on windows and EISDIR on linux)

davidkna:
I think I made it work with `sync_all`. Hopefully this works on macos too.

davidkna:
looks like this doesn't work

PR #1563, opened 2020-08-07
feat(directory): fix #1162 add ellipsis to truncated paths

### LINKED ISSUE #1162: Add truncate_symbol for directory as with git_branch

Add `directory.truncate_symbol` when path has more than `directory.truncation_length` parts

### LINKED ISSUE #1626: Add option similar to SPACESHIP_DIR_TRUNC_PREFIX from spaceship prompt
I'm a user of spaceship prompt looking at this project since it seems to be getting more attention and maintenance. I'm looking for an option similar to `SPACESHIP_DIR_TRUNC_PREFIX` from spaceship as I like the directory truncation, but like having a prefix when it is used so as to not be confused by it.

For starship to have an option similar to `SPACESHIP_DIR_TRUNC_PREFIX` from spaceship prompt.

See [spaceship options - Directory (dir)](https://github.com/denysdovhan/spaceship-prompt/blob/d4642e4a4739fed3cf8dcd62f41edf99c84799f4/docs/Options.md#directory-dir)

jhilliker:
#### Description
Adds ellipsis in front of truncated paths: …/
Configurable through new config option: directory.truncation_symbol
~~On~~ Off by default (default is empty string)

#### Motivation and Context
Better visual indication of when the path has been truncated
Closes #1162 
Fixes #1626 

Screenshot from 2020-08-07 00-08-10

~~Updated existing tests~~ No longer needed since default is empty symbol.  Default behaviour is the same as before.
Ran `cargo test && cargo test --tests directory:: -- --ignored`
Added test for new config option: directory.truncation_symbol

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -651,6 +651,7 @@ it would have been `nixpkgs/pkgs`.
 | `disabled`          | `false`                                            | Disables the `directory` module.                                                 |
 | `read_only`         | `"🔒"`                                             | The symbol indicating current directory is read only.                            |
 | `read_only_style`   | `"red"`                                            | The style for the read only symbol.                                              |
+| `truncation_symbol` | `""`                                               | The symbol to prefix to truncated paths. eg: "…/"                                |
 
 <details>
 <summary>This module has a few advanced configuration options that control how the directory is displayed.</summary>
@@ -694,6 +695,7 @@ a single character. For `fish_style_pwd_dir_length = 2`, it would be `/bu/th/ci/
 
 [directory]
 truncation_length = 8
+truncation_symbol = "…/"
 ```
 
 ## Docker Context
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -15,6 +15,8 @@ use crate::config::RootModuleConfig;
 use crate::configs::directory::DirectoryConfig;
 use crate::formatter::StringFormatter;
 
+const HOME_SYMBOL: &str = "~";
+
 /// Creates a module with the current directory
 ///
 /// Will perform path contraction, substitution, and truncation.
@@ -31,8 +33,6 @@ use crate::formatter::StringFormatter;
 /// **Truncation**
 /// Paths will be limited in length to `3` path components by default.
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
-    const HOME_SYMBOL: &str = "~";
-
     let mut module = context.new_module("directory");
     let config: DirectoryConfig = DirectoryConfig::try_load(module.config);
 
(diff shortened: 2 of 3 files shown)


src/modules/directory.rs
```diff
@@ -129,6 +109,38 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     Some(module)
 }
 
+fn is_truncated(path: &str) -> bool {
```
davidkna, line 112: I think it would be better if the `path` was `&Path`

src/modules/directory.rs
```diff
@@ -129,6 +109,38 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    !(path.starts_with(HOME_SYMBOL)
+        || if cfg!(not(target_os = "windows")) {
+            path.starts_with('/')
+        } else {
+            PathBuf::from(path).has_root() || PathBuf::from(String::from(path) + r"\").has_root()
```
davidkna, line 117: Any reason `has_root` is only used on Windows?

docs/config/README.md
```diff
@@ -637,11 +637,12 @@ it would have been `nixpkgs/pkgs`.
...
-| `format`                 | `"[$path]($style)[$lock_symbol]($lock_style) "` | The format for the module.                                                       |
+| `format`                 | `"[$truncated]($style)[$path]($style)[$read_only]($read_only_style) "` | The format for the module.                                |
 | `style`                  | `"bold cyan"`                                   | The style for the module.                                                        |
 | `disabled`               | `false`                                         | Disables the `directory` module.                                                 |
 | `read_only_symbol`       | `"🔒"`                                          | The symbol indicating current directory is read only.                            |
 | `read_only_symbol_style` | `"red"`                                         | The style for the read only symbol.                                              |
+| `truncation_symbol`      | `"…/"`                                            | The symbol indicating path truncation.                                         |
```
davidkna, line 645: Not sure whether this should be shown by default. When the default path separator changes to `\` on windows this will need to become platform dependent too and I don't think this is the default behavior for most shells, at least it's not in spaceship.

src/modules/directory.rs
jhilliker: I agree, but that module is pretty hairy.  The part that does path substitutions (aliases?) all works with strings, so it gets converted to strings somewhere else early on.  By the time it gets to my addition, it's already a string.

I tried to refactor it to always use &Path or &PathBuf, but it ended up a complete rewrite (due to the current version making it a string quite early on), so instead I went with a patch that changed as little as possible.

src/modules/directory.rs
jhilliker: At this point it was a string, so rather than creating an object just to find if it's root, I just looked for the leading slash.

The Windows version has to use the `has_root` because a root path on windows could start with any drive letter, or be a network path.

It's an easy change though, so I will make it and test it.

docs/config/README.md
jhilliker: Will change to just "" and expand the docs.

jhilliker: FWIW, the path's get converted to `/` anyways on windows. See the first few windows tests.

jhilliker:
Requested changes made.  Rebased.  @davidkna please review.

docs/config/README.md
```diff
@@ -631,11 +631,12 @@ it would have been `nixpkgs/pkgs`.
 | ------------------------ | ----------------------------------------------- | -------------------------------------------------------------------------------- |
 | `truncation_length`      | `3`                                             | The number of parent folders that the current directory should be truncated to.  |
 | `truncate_to_repo`       | `true`                                          | Whether or not to truncate to the root of the git repo that you're currently in. |
-| `format`                 | `"[$path]($style)[$lock_symbol]($lock_style) "` | The format for the module.                                                       |
+| `format`                 | `"[$truncated]($style)[$path]($style)[$read_only]($read_only_style) "` | The format for the module.                                |
```
davidkna, line 634: Please document the new `$truncated` variable in the variables section.

Beyond that LGTM.

luislavena:
Hello @jhilliker, was just looking for this feature, thank you!

Something I miss when migrated from powerline-go is that the ellipsis happen for the middle directories instead of the parent:

image

What do you think about something like that?

Either way, thank you for your contribution!
❤️ ❤️ ❤️

docs/config/README.md
jhilliker: Decided to remove the `$truncated` variable as it doesnt make sense to move it.  It's now just included in `$path` as it was before with the fish-style path truncation.

jhilliker:
@luislavena that's interesting, and something that I thought about as well.

I'm onto other projects though, so I'm hoping to wrap this pull request so that anyone else can work on the module with few conflicts.

jhilliker:
Requested changes made. Rebased. @davidkna please review.

matchai:
Thank you for the contribution! I certainly welcome this as a feature to the prompt, though I'm not personally in favor of it being set as the default behavior. The current behavior helps keep things clean and limit visual noise in the directory module when you're in a project directory, as behaves as a would personally prefer.

I could certainly be persuaded to change the default behavior, but I would rather we create a dedicated issue/PR to change the default behavior so be able to get more eyes and opinions on the matter.

davidkna:
> Thank you for the contribution! I certainly welcome this as a feature to the prompt, though I'm not personally in favor of it being set as the default behavior. The current behavior helps keep things clean and limit visual noise in the directory module when you're in a project.

I already had them [disable this by default](https://github.com/starship/starship/pull/1563#discussion_r479633617).

davidkna (approved):
LGTM

jhilliker:
rebased

P.S. your CI system is broken.  It was having intermittent failures.  The last  two forced pushes I made changed literally nothing, but this one passes...

chipbuster:
> your CI system is broken. It was having intermittent failures. The last two forced pushes I made changed literally nothing, but this one passes...

We've been tracking this in #886 for quite a while now, but we still have no idea what breaks. If you have any suspicions, we'd love to hear them!

PR #1576, opened 2020-08-11
feat: Error messaging

### LINKED ISSUE #1075: Implement error messaging for the prompt

ATiltedTree:
#### Description
Implements error messaging.
This take a bit of a different approach than #1416.
This creates a custom logger for the `log` crate which logs everything to a file (`/tmp/starship/session_$STARSHIP_SESSION_KEY.log`) and it logs everything above `Warn` to stderr but only if the log file does not contain the line that should be logged resulting in an error or warning to be only logged at the first starship invocation after opening the shell.
I choose the temp dir for storing logs as they can become quite large and most temp dir are cleared at reboot.

#### Motivation and Context

Closes #1075

Wrong format string:
image

Broken config:
image

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -26,4 +26,4 @@ function global:prompt {
 $ENV:STARSHIP_SHELL = "powershell"
 
 # Set up the session key that will be used to store logs
-$ENV:STARSHIP_SESSION_KEY = -join (('a'..'z') + ('A'..'Z') + ('0'..'9') | Get-Random -Count 16)
+$ENV:STARSHIP_SESSION_KEY = $(::STARSHIP:: session) 
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -26,4 +26,4 @@ function global:prompt {
 $ENV:STARSHIP_SHELL = "powershell"
 
 # Set up the session key that will be used to store logs
-$ENV:STARSHIP_SESSION_KEY = $(::STARSHIP:: session) 
+$ENV:STARSHIP_SESSION_KEY = $(::STARSHIP:: session)
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -26,4 +26,4 @@ function global:prompt {
 $ENV:STARSHIP_SHELL = "powershell"
 
 # Set up the session key that will be used to store logs
-$ENV:STARSHIP_SESSION_KEY = -join ( (65..90) + (97..122) | Get-Random -Count 16 | % {[char]$_})
+$ENV:STARSHIP_SESSION_KEY = -join (('a'..'z') + ('A'..'Z') + ('0'..'9') | Get-Random -Count 16)
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -24,3 +24,6 @@ function global:prompt {
 }
 
 $ENV:STARSHIP_SHELL = "powershell"
+
+# Set up the session key that will be used to store logs
+$ENV:STARSHIP_SESSION_KEY = -join ( (65..90) + (97..122) | Get-Random -Count 16 | % {[char]$_})
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -1,0 +1,87 @@
+use ansi_term::Color;
+use log::{Level, LevelFilter, Metadata, Record};
+use std::{
+    env,
+    fs::{self, File, OpenOptions},
+    io::Write,
+    sync::{Arc, Mutex},
+};
+
+pub struct StarshipLogger {
+    log_file: Arc<Mutex<File>>,
+    log_file_content: Arc<String>,
+}
+
+impl StarshipLogger {
+    fn new() -> Self {
+        let log_dir = env::temp_dir().join("starship");
+
+        fs::create_dir_all(log_dir.clone()).expect("Unable to create log dir!");
+        let session_log_file = log_dir.join(format!(
+            "session_{}.log",
+            env::var("STARSHIP_SESSION_KEY").unwrap_or_default()
+        ));
+
+        Self {
+            log_file_content: Arc::new(
+                fs::read_to_string(session_log_file.clone()).unwrap_or_default(),
+            ),
+            log_file: Arc::new(Mutex::new(
+                OpenOptions::new()
+                    .create(true)
+                    .append(true)
+                    .open(session_log_file)
+                    .unwrap(),
+            )),
+        }
+    }
+}
+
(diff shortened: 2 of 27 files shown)


src/init/starship.ps1
```diff
@@ -24,3 +24,6 @@ function global:prompt {
 }
 
 $ENV:STARSHIP_SHELL = "powershell"
+
+# Set up the session key that will be used to store logs
+$ENV:STARSHIP_SESSION_KEY = -join ( (65..90) + (97..122) | Get-Random -Count 16 | % {[char]$_})
```
davidkna, line 29:
```suggestion
$ENV:STARSHIP_SESSION_KEY = -join ( (65..90) + (97..122) | Get-Random -Count 16 | ForEach-Object {[char]$_})
```
Using aliases like `|%` is discouraged in scripts

src/logger.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::{
+    env,
+    fs::{self, File, OpenOptions},
+    io::Write,
+    sync::{Arc, Mutex},
+};
+
+pub struct StarshipLogger {
+    log_file: Arc<Mutex<File>>,
+    log_file_content: Arc<String>,
```
davidkna, line 12: Should `log_file` be `BufWrite` and `log_file_content` a `HashSet<String>`? I think compressing the logfile could also be worth it.

src/init/starship.ps1
davidkna:
```suggestion
$ENV:STARSHIP_SESSION_KEY = -join (('a'..'z') + ('A'..'Z') + ('0'..'9') | Get-Random -Count 16)
```
even better

src/logger.rs
ATiltedTree: I've changed them to your respective suggestions but how do you think we should compress the log files.
Because we can't just compress all logs that are not our session (e.g. from an old session) because we can have multiple sessions open at the same time.
And compressing our own file does not make much sense to me because we would have to uncompress and recompress the log file every starship invocation which could potentially take quite long.

davidkna: It looks like the file gets flushed with every log message, so no Bufwriter might be better because it's not going to buffer anything.

I didn't think of having to recompress every time, I guess it's better not to compress that in that case.

matchai:
Testing this branch locally, it seems we may have some locale quirks when using `tr` on macOS to account for:
image

ATiltedTree:
> Testing this branch locally, it seems we may have some locale quirks when using `tr` on macOS to account for:
> 
> image

Interesting. What about this:

```sh
cat /dev/urandom | base64 -w 0 | fold -w 16 | head -n 1
```
Does that produce any locale errors?

Or maybe we could use the current time e.g. the milliseconds since the epoch:
```
date +%s%3N
```

matchai:
It looks like on macOS `base64` doesn't accept a `-w` flag. Without it, things seem to work fine.
image

matchai:
The simplest solution would probably be to have an implementation in Rust that is made available by Starship.

ATiltedTree:
Have now refactored it to use a native rust implementation using the `rand` crate. This hopefully should behave the same way everywhere.

matchai:
Unfortunately, it doesn't look like `/tmp/starship` is being created on my system.
I'm about to step away so I can't help in troubleshooting, but I'll try when I get back.

matchai:
With the `log_dir` set to be logged, it looks like it doesn't always create a file in `/tmp/`:

image

Much like how our config is stored in a fixed directory relative to `$HOME`, maybe this should be as well to avoid having differing locations across platforms. 🤔

matchai:
It would be nice if we could get these changes in as part of v0.45.0 to guide users to the migration guide if they're using `prefix` or `suffix`. Do you think it'd be easy to add custom messaging in that particular case?

ATiltedTree:
I haven't really looked into the config deserialization and honestly I don't really understand all of the proc_macro stuff but if we have a way to detect unused config variables then we could log an error every time we encounter a `prefix` variable.

And getting back to your previous question, where do you think we should store this?
I would propose `~/.cache/starship/session_${STARSHIP_SESSION_KEY}.log` since it somewhat matches our config file location.

heyrict:
> Do you think it'd be easy to add custom messaging in that particular case?

I'm not familiar with how the messages are generated in this PR, but if there is a function similar to `log!("...")`, it should be possible to be called in the proc_macro.

matchai:
> And getting back to your previous question, where do you think we should store this?
I would propose ~/.cache/starship/session_${STARSHIP_SESSION_KEY}.log since it somewhat matches our config file location.

I'm in favor of this. 👍

ATiltedTree:
I have now changed the log file location and even made it relocatable and updated the docs about it.
In addition I added logging if we encounter a `prefix` or `suffix` value the config file linking to the migration guide of #1624.

matchai:
One foreseeable issue is that the log file will keep growing as long as the same session is active. This will lead to longer and longer prompt render times.

This can be seen by running the following: `hyperfine "starship prompt" -m 100000`.

This is definitely a contrived example, but perhaps we can limit the deduplication of logs to only some types of errors?
For instance, we could keep logging to stdout, only keeping a history of `ERROR` log types.

ATiltedTree:
Now only warning and up are kept resulting in smaller log files.
I also restored `STARSHIP_LOG` which can be used to view the debug and trace logs.

matchai:
Works great locally! Running the same hyperfine benchmark, it doesn't look like runtime is affected by this change! 😄 

One last note before this is ready to merge:
It looks like setting `STARSHIP_LOG` causes logs to be outputted to both stdout and the session's log file. While this wouldn't normally be an issue, that means that debug logs are also deduplicated similarly to warnings and errors. This can be seen by running `export STARSHIP_LOG=debug`.

ATiltedTree:
Should be fixed now. :smile: It only saves warnings and above now ignoring the `STARSHIP_LOG` variable.

PR #1584, opened 2020-08-13
chore: Remove unused name from Segment and remove some of the misleading underscores

ATiltedTree:
#### Description

As the title says I removed the
  - name property from `Segment` because it wasn't used anywhere
  - misleading underscores in function names and variables because it normally means "this isn't used anywhere but we will in the future" which wasn't the case here

#### Motivation and Context

Improve the code readability

#### Checklist:


matchai:
Looks like a good set of changes. 👍 
Rerunning CI, it looks like the above failure is no longer happening.

PR #1586, opened 2020-08-14
docs: Drop inexistent "number" variable from hostname section

em-:
#### Description

The hostname module does not provide any "number" variable, its presence
in the documentation is likely due to some copy&paste from the jobs
module section, so let's drop it.


andytom:
Good catch @em- thanks for taking the time to fix this.

PR #1592, opened 2020-08-16
fix: restore add_newline configuration

### LINKED ISSUE #1549: Restore the add_newline config option
In #1056 we deprecated the `add_newline` config option. Following a discussion on #1546 it looks like this was a mistake and we should restore this option. 

> I think it would be best if we readded it. Removing the linebreak is common enough configuration to justify it being a boolean config option.

_Originally posted by @matchai in https://github.com/starship/starship/pull/1546#issuecomment-667729821_

matchai:
#### Description

Re-implements the `add_newline` root configuration option, along with documentation and tests.

#### Motivation and Context

Closes #1549 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1,12 +1,5 @@
 # Configuration
 
-::: tip
-
-🔥 Configuration is currently being worked on.
-Many new configuration options will be available in coming releases.
-
-:::
-
 To get started configuring starship, create the following file: `~/.config/starship.toml`.
 
 ```sh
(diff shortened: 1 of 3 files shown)


docs/config/README.md
```diff
@@ -1,23 +1,15 @@
 # Configuration
 
-::: tip
```
chipbuster, line 3: Guess we're finally there, huh? :grin:

matchai: Haha. Yep! 😜

PR #1594, opened 2020-08-17
fix: Add a fallback prompt under TERM=dumb

Dietr1ch:
#### Description

This fallback is the quick-fix for #1588

#### Motivation and Context

See https://github.com/starship/starship/issues/1588#issuecomment-674568040

Ran it under `env`, `env 'TERM=dumb` and `env 'TERM=dumber`.

#### Checklist:

So far this is missing tests and documentation changes.
I'm not sure how to write the test for it (new to rust and this project), nor whether
should this behavior be documented nor where.

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -26,6 +26,15 @@ pub fn get_prompt(context: Context) -> String {
     let config = context.config.get_root_config();
     let mut buf = String::new();
 
+    match std::env::var_os("TERM") {
+        Some(term) if term == "dumb" => {
+            log::error!("Under a 'dumb' terminal (TERM=dumb).");
+            buf.push_str("Starship disabled due to TERM=dumb > ");
+            return buf;
+        }
+        _ => {}
+    }
+
     // A workaround for a fish bug (see #739,#279). Applying it to all shells
     // breaks things (see #808,#824,#834). Should only be printed in fish.
     if let Shell::Fish = context.shell {


chipbuster:
So I tested starship in an Emacs dumb terminal last night (M-x shell) and while the emojis didn't work, the color codes and powerline symbols (!!) did.

Are we sure that the color codes are what's actually causing the observed issues in #1588?

Dietr1ch:
I think that the shell supports more than what's supported on tramp-mode.

Try opening `/ssh:user@machine:some/remote/path` or
`/sudo:user@hostname:/etc/foo`
and then opening a shell, it should silently be a shell on the 'remote'.

(I'm on my phone now and it's kinda late, so I can't try things out now)

Dietr1ch:
Well, [tramp-mode](https://www.emacswiki.org/emacs/TrampMode) defaults its parsing pattern, `tramp-shell-prompt-pattern` to,

```
\\(?:^\\|\r\\)[^]#$%>\n]*#?[]#$%>] *\\(^[\\[[0-9;]*[a-zA-Z] *\\)*"
```

Anything that matches that would be recognized properly by default, which would be really nice, but it rules out using color sequences.

I'm not sure if we should snoop at `$INSIDE_EMACS` for having the "tramp" on its name (`26.2,tramp:2.3.4`)
and let dumb terminals be fancier. I'd expect other programs that try to use "dumb" terminals to also have simple
parsing rules.

Dietr1ch:
Hey, is it possible to merge this?

dumb terminals usually don't handle colours nor fancy symbols properly.


tramp-mode by default assumes the prompt is essentially alphanumeric with a couple of extra visible ascii characters.

https://www.gnu.org/software/emacs/manual/html_node/tramp/Frequently-Asked-Questions.html

andytom (approved):

andytom:
Thank you for your contribution @Dietr1ch

PR #772, opened 2019-12-21
fix: Parse arguments of $VISUAL and $EDITOR

dominikbraun:
#### Description
We thought that Issue https://github.com/starship/starship/issues/769 was a result of empty `$VISUAL` and/or `$EDITOR` variables, which were indeed a problem. However, the same error occurs when an environment variable contains arguments or options. This will cause `starship configure` to panic:

```shell script
export VISUAL="vi -m"
```

While https://github.com/starship/starship/pull/770 fixes the bug for empty environment variables, this PR takes all options like `-m` into account when executing the editor.

#### Motivation and Context
Closes https://github.com/starship/starship/issues/769

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?
- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #1595, opened 2020-08-17
fix $EDITOR argument parsing

jRimbault:
See #1591

Shell word splitting is actually more complicated than splitting by whitespace.


src/configure.rs
```diff
@@ -61,17 +64,12 @@ pub fn update_configuration(name: &str, value: &str) {
 
 pub fn edit_configuration() {
     let config_path = get_config_path();
-    let editor_cmd = get_editor();
+    let editor_cmd = shell_words::split(&get_editor()).unwrap();
```
chipbuster, line 67: `unwrap()` always makes me just a little nervous. Under what conditions can this fail? Is it a realistic failure mode for us? At the very least, it would be good to document when the split might fail (in this case, only when unmatched quotes are found in the result of `get_editor()`)

chipbuster: The self-documenting option for this is `.expect("Unmatched quotes found in editor")`

jRimbault: Yes, I should have use `expect` in the first place here.

jRimbault:
I didn't exactly "intuit" the use case of `vim -m`, I git blamed and looked up the commits and, as it happens, read the corresponding PR #772. To me that seems enough? If you wanted to be even more clever, I'd suggest adding tests and completely separating the logic from the side-effects.

There is still at least one problematic case I can see: if the environment variable contains spaces and is not quoted, which can be documented as a gentle PEBCAK (imo, since git also shares this flaw).

chipbuster:
> I didn't exactly "intuit" the use case of vim -m, I git blamed and looked up the commits and, as it happens, read the corresponding PR #772. To me that seems enough?

Yes, but I think this becomes unreliable as the history grows longer. I'd prefer to add tests at some point so that any changes that break that functionality will be flagged by the testsuite.

> ...which can be documented as a gentle PEBCAK (imo, since git also shares this flaw).

Yeah, shell word splitting is just a pain to try to dance around.

> as a sidenote in my quick cursory looking through the code I noticed a lot of duplicated code, and a few if cfg!(test) which could have been compile time conditionals instead of runtime checks.

We'd be happy to accept a refactor/chore PR cleaning those up, if you have the time! :)

jRimbault:
Completely agree with you :)

I already got the itch, you don't need to push me too much to help around ^^

PR #1612, opened 2020-08-23
fix(zsh): work when `setopt ksh_arrays` is set

### LINKED ISSUE #1610: Starship does not start with `setopt ksh_arrays` in zsh
#### Current Behavior
Starship does not start when there is `setopt ksh_arrays` before starship initialization in `.zshrc`.

#### Expected Behavior
Starship starts and works.

#### Possible Solution
I think this part of `src/init/starship.zsh` need to be fixed. These conditions may work wrong because `setopt ksh_arrays` changes the index of the array.
```sh
# If starship precmd/preexec functions are already hooked, don't double-hook them
# to avoid unnecessary performance degradation in nested shells
if [[ ${precmd_functions[(ie)starship_precmd]} -gt ${#precmd_functions} ]]; then
    precmd_functions+=(starship_precmd)
fi
if [[ ${preexec_functions[(ie)starship_preexec]} -gt ${#preexec_functions} ]]; then
    preexec_functions+=(starship_preexec)
fi
```

#### Environment
- Starship version: 0.44.0
- zsh version: zsh 5.8 (x86_64-apple-darwin19.3.0)
- Operating system: Mac OS 10.15.6

```zsh
setopt ksh_arrays
eval "$(starship init zsh)"
```

#### Starship Configuration

```toml
<unknown config>
```

davidkna:
#### Description

Use a value based instead of an index based method to check if the `starship_pre*` functions are already installed. This works even if the base index of arrays is changed (`KSH_ARRAYS`).

`${list[(r)x}` returns `x` if `x` is in `list` else an empty string.
`${list[(re)x}` ensures string instead of pattern matches are used.

#### Motivation and Context

Closes #1610
The issue is explained in detail there.

#### Checklist:

### DIFF
--- a/src/init/starship.zsh
+++ b/src/init/starship.zsh
@@ -46,10 +46,10 @@ starship_preexec() {
 
 # If starship precmd/preexec functions are already hooked, don't double-hook them
 # to avoid unnecessary performance degradation in nested shells
-if [[ ${precmd_functions[(ie)starship_precmd]} -gt ${#precmd_functions} ]]; then
+if [[ -z ${precmd_functions[(re)starship_precmd]} ]]; then
     precmd_functions+=(starship_precmd)
 fi
-if [[ ${preexec_functions[(ie)starship_preexec]} -gt ${#preexec_functions} ]]; then
+if [[ -z ${preexec_function[(re)starship_preexec]} ]]; then
     preexec_functions+=(starship_preexec)
 fi
 


cappyzawa (approved):
LGTM:+1:

andytom:
@davidkna, thank you for your contribution and @cappyzawa thanks for reviewing.

PR #1616, opened 2020-08-24
chore: fix the double compiling of lib and bin

jRimbault:
#### Description
Rust modules were compiled twice, once as a library and once as a binary.

#### Motivation and Context

This caused the compile time of the _crate_ to be double what it could be (11-16s vs 5-8s on my machine, not really important but still annoying), and caused some analysis to wrongly mark some functions as unused because they are unused in the binary but not in the library (or vice-versa) which caused some confusion.

#### Checklist:


jRimbault:
I've run the tests on Windows, and I guess you already know that the test `modules::directory::tests::root_directory` fails when not running as Admin ?

andytom:
> I've run the tests on Windows, and I guess you already know that the test `modules::directory::tests::root_directory` fails when not running as admin ? It did not fail on my previous PR/branch which is weird.
> 
> ```
> ---- modules::directory::tests::root_directory stdout ----
> thread 'modules::directory::tests::root_directory' panicked at 'assertion failed: `(left == right)`
>   left: `Some("\u{1b}[1;36m/\u{1b}[0m ")`,
>  right: `Some("\u{1b}[1;36m/\u{1b}[0m\u{1b}[31m🔒\u{1b}[0m ")`', src\modules\directory.rs:638:9
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> When running the tests as admin, the root directory is writeable so the lock emoji is not present and the test passes.

Yep, while we have tried to reduce them as much as possible we still have a few tests that depend on the state of the machine that is running the test. This is unfortunately one of those tests, in general we ignore them when running locally and only run them in a machine where we control the state, the CI system.

> Edit: btw do you want me to collapse those two commits into one ?

No we will squash the commits down when we merge the PR.

chipbuster (approved):
And we're up-to-date!

Thanks for the PR @jRimbault, and sorry it took us so long to get this approved!

PR LGTM :+1:

andytom:
@jRimbault thank you for the fix.

PR #1629, opened 2020-08-30
feat: Add timings subcommand

jankatins:
#### Description

Add a `starship timings` subcommand which outputs the timings for all active (=included in the top level format) modules.

Also timings to explain output.

#### Motivation and Context

I wanted to see how much is spend on individual modules to see which was the problem for my long prompt delay. 

Example:
```
~/external/starship on  measure_module_time [!] [!] via 🦀 v1.46.0
[23:42:42] λ  ./target/release/starship timings

 Here are the timings of all active modules in your prompt:
 direct_git_status  -  60ms  -
 git_status         -  55ms  -
 rust               -  23ms  -   via 🦀 v1.46.0
 battery            -   2ms  -
 directory          -   1ms  -   ~/external/starship
 username           -   1ms  -
 git_commit         -  <1ms  -
 git_branch         -  <1ms  -   on  measure_module_time
 time               -  <1ms  -   [23:47:54]
 character          -  <1ms  -   λ
 python             -  <1ms  -
 terraform          -  <1ms  -
 shlvl              -  <1ms  -
 hostname           -  <1ms  -
 conda              -  <1ms  -
 jobs               -  <1ms  -
 docker_context     -  <1ms  -
 git_state          -  <1ms  -
 env_var            -  <1ms  -
 cmd_duration       -  <1ms  -
 line_break         -  <1ms  -   \n
 memory_usage       -  <1ms  -

~/external/starship on  measure_module_time via 🦀 v1.46.0 took 42s
[00:11:29] λ  target/debug/starship explain

 λ   (<1ms)                       -  A character (usually an arrow) beside where the text is entered in your terminal
 ~/external/starship  (3ms)       -  The current working directory
 on  measure_module_time  (1ms)  -  The active branch of the repo in your current directory
 via 🦀 v1.46.0  (30ms)           -  The currently installed version of Rust
 [00:12:15]  (<1ms)               -  The current local time
``` 

I've run the explain command in my environment. I ran the suggested lint and test commands.

**This is my first time I write rust code...**

#### Checklist:

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -7,7 +7,7 @@ use std::time::Duration;
 
 // List of all modules
 // Keep these ordered alphabetically.
-// Default ordering is handled in configs/mod.rs
+// Default ordering is handled in configs/starship_root.rs
 pub const ALL_MODULES: &[&str] = &[
     "aws",
     #[cfg(feature = "battery")]
--- a/src/module.rs
+++ b/src/module.rs
@@ -76,7 +76,7 @@ pub struct Module<'a> {
     pub segments: Vec<Segment>,
 
     /// the time it took to compute this module
-    pub duration: Option<Duration>,
+    pub duration: Duration,
 }
 
 impl<'a> Module<'a> {
@@ -87,7 +87,7 @@ impl<'a> Module<'a> {
             name: name.to_string(),
             description: desc.to_string(),
             segments: Vec::new(),
-            duration: None,
+            duration: Duration::default(),
         }
     }
 
@@ -173,7 +173,7 @@ mod tests {
             name: name.to_string(),
             description: desc.to_string(),
             segments: Vec::new(),
-            duration: None,
+            duration: Duration::default(),
         };
 
         assert!(module.is_empty());
@@ -188,7 +188,7 @@ mod tests {
             name: name.to_string(),
             description: desc.to_string(),
             segments: vec![Segment::new(None, "")],
-            duration: None,
+            duration: Duration::default(),
         };
 
         assert!(module.is_empty());
@@ -203,7 +203,7 @@ mod tests {
             name: name.to_string(),
             description: desc.to_string(),
             segments: vec![Segment::new(None, "\n")],
-            duration: None,
+            duration: Duration::default(),
         };
 
         assert!(!module.is_empty());
@@ -218,7 +218,7 @@ mod tests {
             name: name.to_string(),
             description: desc.to_string(),
             segments: vec![Segment::new(None, " ")],
-            duration: None,
+            duration: Duration::default(),
         };
 
         assert!(!module.is_empty());
--- a/src/module.rs
+++ b/src/module.rs
@@ -110,7 +110,8 @@ impl<'a> Module<'a> {
     pub fn is_empty(&self) -> bool {
         self.segments
             .iter()
-            .all(|segment| segment.value.trim().is_empty())
+            // no trim: if we add spaces/linebreaks it's not "empty" as we change the final output
+            .all(|segment| segment.value.is_empty())
     }
 
     /// Get values of the module's segments
@@ -191,5 +192,35 @@ mod tests {
         };
 
         assert!(module.is_empty());
+    }
+
+    #[test]
+    fn test_module_is_not_empty_with_linebreak_only() {
+        let name = "unit_test";
+        let desc = "This is a unit test";
+        let module = Module {
+            config: None,
+            name: name.to_string(),
+            description: desc.to_string(),
+            segments: vec![Segment::new(None, "\n")],
+            duration: None,
+        };
+
+        assert!(!module.is_empty());
+    }
+
+    #[test]
+    fn test_module_is_not_empty_with_space_only() {
+        let name = "unit_test";
+        let desc = "This is a unit test";
+        let module = Module {
+            config: None,
+            name: name.to_string(),
+            description: desc.to_string(),
+            segments: vec![Segment::new(None, " ")],
+            duration: None,
+        };
+
+        assert!(!module.is_empty());
     }
 }
--- a/src/module.rs
+++ b/src/module.rs
@@ -3,6 +3,7 @@ use crate::segment::Segment;
 use crate::utils::wrap_colorseq_for_shell;
 use ansi_term::{ANSIString, ANSIStrings};
 use std::fmt;
+use std::time::Duration;
 
 // List of all modules
 // Keep these ordered alphabetically.
@@ -73,6 +74,9 @@ pub struct Module<'a> {
 
     /// The collection of segments that compose this module.
     pub segments: Vec<Segment>,
+
+    /// the time it took to compute this module
+    pub duration: Option<Duration>,
 }
 
 impl<'a> Module<'a> {
@@ -83,6 +87,7 @@ impl<'a> Module<'a> {
             name: name.to_string(),
             description: desc.to_string(),
             segments: Vec::new(),
+            duration: None,
         }
     }
 
@@ -167,6 +172,7 @@ mod tests {
             name: name.to_string(),
             description: desc.to_string(),
             segments: Vec::new(),
+            duration: None,
         };
 
         assert!(module.is_empty());
@@ -181,6 +187,7 @@ mod tests {
             name: name.to_string(),
             description: desc.to_string(),
             segments: vec![Segment::new(None, "")],
+            duration: None,
         };
 
         assert!(module.is_empty());
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -63,7 +63,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
     let mut m: Option<Module> = {
         match module {
             // Keep these ordered alphabetically.
-            // Default ordering is handled in configs/mod.rs
+            // Default ordering is handled in configs/starship_root.rs
             "aws" => aws::module(context),
             #[cfg(feature = "battery")]
             "battery" => battery::module(context),
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -128,7 +128,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
     } else {
         // if we took more than 1ms we want to report that and so--in case we have None currently--
         // need to create an empty module just to hold the duration for that case
-        m.get_or_insert(context.new_module(module)).duration = elapsed;
+        m.get_or_insert_with(|| context.new_module(module)).duration = elapsed;
         m
     }
 }
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -120,7 +120,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
         }
     };
 
-    m.get_or_insert(context.new_module(module)).duration = Option::from(start.elapsed());
+    m.get_or_insert(context.new_module(module)).duration = start.elapsed();
     m
 }
 
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -121,6 +121,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
     };
 
     let elapsed = start.elapsed();
+    log::trace!("Took {:?} to compute module {:?}", elapsed, module);
     if elapsed.as_millis() < 1 {
         // If we take less than 1ms to compute a None, then we will not return a module at all
         // if we have a module: default duration is 0 so no need to change it
--- a/src/modules/mod.rs
+++ b/src/modules/mod.rs
@@ -55,65 +55,73 @@ mod battery;
 use crate::config::RootModuleConfig;
 use crate::context::{Context, Shell};
 use crate::module::Module;
+use std::time::Instant;
 
 pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
-    match module {
-        // Keep these ordered alphabetically.
-        // Default ordering is handled in configs/mod.rs
-        "aws" => aws::module(context),
-        #[cfg(feature = "battery")]
-        "battery" => battery::module(context),
-        "character" => character::module(context),
-        "cmake" => cmake::module(context),
-        "cmd_duration" => cmd_duration::module(context),
-        "conda" => conda::module(context),
-        "dart" => dart::module(context),
-        "directory" => directory::module(context),
-        "docker_context" => docker_context::module(context),
-        "dotnet" => dotnet::module(context),
-        "elixir" => elixir::module(context),
-        "elm" => elm::module(context),
-        "erlang" => erlang::module(context),
-        "env_var" => env_var::module(context),
-        "gcloud" => gcloud::module(context),
-        "git_branch" => git_branch::module(context),
-        "git_commit" => git_commit::module(context),
-        "git_state" => git_state::module(context),
-        "git_status" => git_status::module(context),
-        "golang" => golang::module(context),
-        "helm" => helm::module(context),
-        "hg_branch" => hg_branch::module(context),
-        "hostname" => hostname::module(context),
-        "java" => java::module(context),
-        "jobs" => jobs::module(context),
-        "julia" => julia::module(context),
-        "kubernetes" => kubernetes::module(context),
-        "line_break" => line_break::module(context),
-        "memory_usage" => memory_usage::module(context),
-        "nim" => nim::module(context),
-        "nix_shell" => nix_shell::module(context),
-        "nodejs" => nodejs::module(context),
-        "ocaml" => ocaml::module(context),
-        "package" => package::module(context),
-        "perl" => perl::module(context),
-        "php" => php::module(context),
-        "purescript" => purescript::module(context),
-        "python" => python::module(context),
-        "ruby" => ruby::module(context),
-        "rust" => rust::module(context),
-        "shlvl" => shlvl::module(context),
-        "singularity" => singularity::module(context),
-        "swift" => swift::module(context),
-        "terraform" => terraform::module(context),
-        "time" => time::module(context),
-        "crystal" => crystal::module(context),
-        "username" => username::module(context),
-        "zig" => zig::module(context),
-        _ => {
-            eprintln!("Error: Unknown module {}. Use starship module --list to list out all supported modules.", module);
-            None
+    let start: Instant = Instant::now();
+
+    let mut m: Option<Module> = {
+        match module {
+            // Keep these ordered alphabetically.
+            // Default ordering is handled in configs/mod.rs
+            "aws" => aws::module(context),
+            #[cfg(feature = "battery")]
+            "battery" => battery::module(context),
+            "character" => character::module(context),
+            "cmake" => cmake::module(context),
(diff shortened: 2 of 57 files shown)


src/print.rs
```diff
@@ -108,8 +109,11 @@ pub fn explain(args: ArgMatches) {
             let value = module.get_segments().join("");
             ModuleInfo {
                 value: ansi_term::ANSIStrings(&module.ansi_strings()).to_string(),
-                value_len: better_width(value.as_str()),
+                value_len: better_width(value.as_str())
+                    + better_width(module.get_duration().as_str())
+                    + 3,
```
davidkna, line 114: This should be part of `PADDING_WIDTH`

src/modules/mod.rs
```diff
@@ -55,65 +55,76 @@ mod battery;
...
+            "username" => username::module(context),
+            "zig" => zig::module(context),
+            _ => {
+                eprintln!("Error: Unknown module {}. Use starship module --list to list out all supported modules.", module);
+                None
+            }
         }
+    };
+
+    match &mut m {
```
davidkna, line 123: It might be nice to log the duration as debug/trace so it can also be inspected from the log.
Have you tested if this causes any slowdown in the prompt? If so it would be to move it into explain.

src/print.rs
jankatins: done

jankatins:
I've now added a way to get the timings of all computed modules. As my rust knowledge was not good enough to do this via a new mutable hashmap in context, I simply changed it so that instead of None, an empty module is returned. This empty module contains then only the duration.

Example:

```
~/external/starship on  measure_module_time [!] [!] via 🦀 v1.46.0 took 11s
[23:02:44] λ  time ./target/debug/starship timings

 Here are the timings of all active modules in your prompt:
git_status         -  85ms  -   [!]
direct_git_status  -  56ms  -   [!]
rust               -  23ms  -   via 🦀 v1.46.0
directory          -   1ms  -   ~/external/starship
username           -   1ms  -
git_commit         -   1ms  -
git_branch         -   1ms  -   on  measure_module_time
time               -  <1ms  -   [23:02:54]
battery            -  <1ms  -
character          -  <1ms  -   λ
python             -  <1ms  -
terraform          -  <1ms  -
jobs               -  <1ms  -
docker_context     -  <1ms  -
hostname           -  <1ms  -
shlvl              -  <1ms  -
memory_usage       -  <1ms  -
line_break         -  <1ms  -   \n
conda              -  <1ms  -
git_state          -  <1ms  -
env_var            -  <1ms  -
cmd_duration       -  <1ms  -

real    0.190
user    0.140
sys     0.065
```  

Re timing:

This is current master (=without this changes) vs a `cargo build --release` version of this branch:

```
~/external/starship on  measure_module_time via 🦀 v1.46.0 took 7m25s
[23:26:35] λ  time ./target/release/starship explain

 λ   (<1ms)                        -  A character (usually an arrow) beside where the text is entered in your terminal
 ~/external/starship  (1ms)        -  The current working directory
 on  measure_module_time  (<1ms)  -  The active branch of the repo in your current directory
 via 🦀 v1.46.0  (24ms)            -  The currently installed version of Rust
 [23:26:47]  (<1ms)                -  The current local time

real    0.159
user    0.116
sys     0.056


~/external/starship on  measure_module_time via 🦀 v1.46.0
[23:26:47] λ  time starship explain

 λ                          -  A character (usually an arrow) beside where the text is entered in your terminal
 ~/external/starship        -  The current working directory
 on  measure_module_time   -  The active branch of the repo in your current directory
 via 🦀 v1.46.0             -  The currently installed version of Rust
 [23:26:53]                 -  The current local time

real    0.160
user    0.103
sys     0.070
``` 

-> more or less identical.

jankatins:
Seems this also broke lot's of checks: 

``` 
---- modules::username::tests::ssh_wo_username stdout ----
thread 'modules::username::tests::ssh_wo_username' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some("")`', src/modules/username.rs:104:9
``` 

-> I would appreciate if you could comment/give some kind of go-ahead before I clean up these checks...

andytom:
> Seems this also broke lot's of checks:
> 
> ```
> ---- modules::username::tests::ssh_wo_username stdout ----
> thread 'modules::username::tests::ssh_wo_username' panicked at 'assertion failed: `(left == right)`
>   left: `None`,
>  right: `Some("")`', src/modules/username.rs:104:9
> ```
> 
> -> I would appreciate if you could comment/give some kind of go-ahead before I clean up these checks...

Personally I'm not sure how much value there is in having a separate timings sub-command, most if not all of the modules will simply return nothing if they don't detect the required files or folders. From my local testing this produces a lot of modules with `<1ms` timings, see the example below from my checkout of the starship repo, which I found makes it harder to get accurate timing data. I really like having the timings in the `explain` subcommand, this is going to be really useful just helping debug performance issues e.g to see if there are any misbehaving modules.

```
~/s/r/starship on  measure_module_time [$] is 📦 v0.44.0 via 🦀 v1.46.0 
❯ target/debug/starship timings

 Here are the timings of all active modules in your prompt:
 git_status      -  45ms  -   [$] 
 rust            -  43ms  -   via 🦀 v1.46.0 
 battery         -  10ms  -   
 package         -   6ms  -   is 📦 v0.44.0 
 directory       -   3ms  -   ~/s/r/starship 
 username        -   1ms  -   
 git_branch      -   1ms  -   on  measure_module_time 
 hg_branch       -  <1ms  -   
 character       -  <1ms  -   ❯ 
 git_commit      -  <1ms  -   
 python          -  <1ms  -   
 aws             -  <1ms  -   
 dotnet          -  <1ms  -   
 kubernetes      -  <1ms  -   
 dart            -  <1ms  -   
 golang          -  <1ms  -   
 elm             -  <1ms  -   
 ocaml           -  <1ms  -   
 docker_context  -  <1ms  -   
 nodejs          -  <1ms  -   
 terraform       -  <1ms  -   
 perl            -  <1ms  -   
 gcloud          -  <1ms  -   
 java            -  <1ms  -   
 swift           -  <1ms  -   
 cmake           -  <1ms  -   
 erlang          -  <1ms  -   
 time            -  <1ms  -   
 nix_shell       -  <1ms  -   
 jobs            -  <1ms  -   
 purescript      -  <1ms  -   
 shlvl           -  <1ms  -   
 ruby            -  <1ms  -   
 nim             -  <1ms  -   
 php             -  <1ms  -   
 julia           -  <1ms  -   
 line_break      -  <1ms  -   \n
 conda           -  <1ms  -   
 hostname        -  <1ms  -   
 elixir          -  <1ms  -   
 helm            -  <1ms  -   
 crystal         -  <1ms  -   
 cmd_duration    -  <1ms  -   
 memory_usage    -  <1ms  -   
 git_state       -  <1ms  -   
 env_var         -  <1ms  -   
 zig             -  <1ms  -   
 singularity     -  <1ms  -   
```

jankatins:
The problem are commands which output nothing but contribute a lot to the slowness. This is e.g. on a work repo where the original git_status module takes 4secs on a up to date git master branch. If I configure git_status to return a blank instead of the empty `[]`(like I did with the custom `direct_git_status`) then this will not show up in explain:

```
~/projects/bi-etl on  master 🐍 v3.8.5
[13:23:31] λ  starship timings

 Here are the timings of all active modules in your prompt:
 git_status         -  4644ms  -   []
 direct_git_status  -   170ms  -
 python             -     8ms  -   🐍 v3.8.5
 directory          -     2ms  -   ~/projects/bi-etl
 username           -     2ms  -
 git_commit         -     1ms  -
 time               -     1ms  -   [13:24:33]
 battery            -    <1ms  -
 docker_context     -    <1ms  -
 git_branch         -    <1ms  -   on  master
 character          -    <1ms  -   λ
 rust               -    <1ms  -
 line_break         -    <1ms  -   \n
 shlvl              -    <1ms  -
 hostname           -    <1ms  -
 conda              -    <1ms  -
 jobs               -    <1ms  -
 terraform          -    <1ms  -
 git_state          -    <1ms  -
 env_var            -    <1ms  -
 cmd_duration       -    <1ms  -
 memory_usage       -    <1ms  -
```

```
~/projects/bi-etl on  master [] 🐍 v3.8.5 took 4s
[13:24:33] λ  starship explain

 λ   (<1ms)                -  A character (usually an arrow) beside where the text is entered in your terminal
 ~/projects/bi-etl  (4ms)  -  The current working directory
 on  master  (<1ms)       -  The active branch of the repo in your current directory
 []  (4942ms)              -  Symbol representing the state of the repo
 🐍 v3.8.5  (9ms)          -  The currently installed version of Python
 [13:28:38]  (1ms)         -  The current local time
```
-> no direct_git_status despite >100ms

If you are oposed to an extra git timing, my suggestion would be to move this to explain but only keep the entries which have more than one ms runtime (so no `<1ms` entries).

Theoutput would the read:

```
 λ   (<1ms)                -  A character (usually an arrow) beside where the text is entered in your terminal
 ~/projects/bi-etl  (4ms)  -  The current working directory
 on  master  (<1ms)       -  The active branch of the repo in your current directory
 []  (4942ms)              -  Symbol representing the state of the repo
 🐍 v3.8.5  (9ms)          -  The currently installed version of Python
 [13:28:38]  (1ms)         -  The current local time

 Here are the timings of all active modules in your prompt which took more than 1ms:
 git_status         -  4644ms  -   []
 direct_git_status  -   170ms  -
 python             -     8ms  -   🐍 v3.8.5
 directory          -     2ms  -   ~/projects/bi-etl
 username           -     2ms  -
 git_commit         -     1ms  -
 time               -     1ms  -   [13:24:33]
```

andytom:
> The problem are commands which output nothing but contribute a lot to the slowness. This is e.g. on a work repo where the original git_status module takes 4secs on a up to date git master branch. If I configure git_status to return a blank instead of the empty `[]`(like I did with the custom `direct_git_status`) then this will not show up in explain:

Sorry, you are completely correct, I just looked back at my example and noticed the battery modules takes 10ms even though it doesn't display anything. How about we keep the separate timings command but add a filter that removes any modules that take <1ms to render?

jankatins:
Sounds good, will implement or wegen I'm back at my laptop next week.

jankatins:
I've now reduced the output to `>=1ms or produces output`. I also fixed the tests to expect `Some(String::from(""))`. The alternative would be to change the `ModuleRenderer` to return None if it encounteres an empty string. Not sure what you would prefer.

(Interestingly, the current `Module``is_empty()` implementation thinks the linebreak module is empty. Should I fix that?)

davidkna:
> 
> 
> I've now reduced the output to `>=1ms or produces output`. I also fixed the tests to expect `Some(String::from(""))`. The alternative would be to change the `ModuleRenderer` to return None if it encounteres an empty string. Not sure what you would prefer.

I'd prefer `None`.

> (Interestingly, the current `` Module``is_empty() `` implementation thinks the linebreak module is empty. Should I fix that?)

Sure

src/modules/mod.rs
```diff
@@ -55,65 +55,73 @@ mod battery;
 use crate::config::RootModuleConfig;
 use crate::context::{Context, Shell};
 use crate::module::Module;
+use std::time::Instant;
 
 pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
```
davidkna, line 60: This should return `Module<'a>` instead now

jankatins:
> I'd prefer None.

Fair enough: updated

src/modules/mod.rs
jankatins: I tried and it's a bit over my head right now (this is my first rust project): it breaks all over the place as this has than a different return type as the custom module which is handled special. Will try again if you insist but would appreciate a merge without that change if possible...

jankatins:
Added the change to the is_empty: if it contains anything (space, linebreak), it's now not empty.

src/modules/mod.rs
```diff
@@ -55,65 +55,73 @@ mod battery;
...
+            "zig" => zig::module(context),
+            _ => {
+                eprintln!("Error: Unknown module {}. Use starship module --list to list out all supported modules.", module);
+                None
+            }
         }
-    }
+    };
+
+    m.get_or_insert(context.new_module(module)).duration = Option::from(start.elapsed());
```
davidkna, line 123:
```suggestion
    let elapsed = start.elapsed();
    // create empty module with timing information if it takes too long, even if it returns nothing
    if elapsed.as_millis() > 0 {
        m = m.or_else(|| context.new_module(module).duration = Option::from(elapsed));
    }
```
I think with this keeping it as `Option` makes sense.

src/module.rs
```diff
@@ -73,6 +74,9 @@ pub struct Module<'a> {
 
     /// The collection of segments that compose this module.
     pub segments: Vec<Segment>,
+
+    /// the time it took to compute this module
+    pub duration: Option<Duration>,
```
davidkna, line 79: I think this could be plain `Duration`, initiliazed as `Duration::default()` (0) when it is not set yet.

src/test/mod.rs
```diff
@@ -84,7 +84,21 @@ impl<'a> ModuleRenderer<'a> {
 
     /// Renders the module returning its output
     pub fn collect(self) -> Option<String> {
-        crate::print::get_module(self.name, self.context)
+        let ret = crate::print::get_module(self.name, self.context);
+        // all tests rely on the fact that an empty module produces None as output as the
+        // convention was that there would be no module but None. This is nowadays not anymore
+        // the case (to get durations for all modules). So here we make it so, that an empty
+        // module returns None in the tests...
+        match &ret {
```
davidkna, line 92: I'd prefer something like this:
```rust
match ret {
    Some("") => None,
    Some(_) => ret,
    None => None,
}
```

src/module.rs
```diff
@@ -105,7 +110,8 @@ impl<'a> Module<'a> {
     pub fn is_empty(&self) -> bool {
         self.segments
             .iter()
-            .all(|segment| segment.value.trim().is_empty())
+            // no trim: if we add spaces/linebreaks it's not "empty" as we change the final output
+            .all(|segment| segment.value.is_empty())
```
davidkna, line 114:
```suggestion
           .all(|segment| segment.value.chars().any(|c| c != ' '))
```
Since this is mostly about the line_break I think it makes sense to still treat space only segments as empty.

jankatins: Opposing argument is that even a single space shows up on the prompt and is clearly visible and therefore the corresponding module not "empty"

davidkna: Yes that makes sense.

src/test/mod.rs
davidkna: better yet:
```rust
match ret {
    Some("") => None,
    _ => ret,
}
```
or
```rust
match ret {
    Some("") => None,
    ret => ret,
}
```

src/modules/mod.rs
jankatins: wrote it a bit differently because as far as I could understand your code, it only set the duration for the None case.

src/test/mod.rs
jankatins: Tried to do it but rust doesn't like it: `ret` is a `Option<String>` but `Some("")` is a `Option<str>`. And `Some(String::from(""))` is an illegal function call in the match and matching `ret.ref()` doesn't work because ret is an Option  :-(

The best I could come up with is 

``` 
if ret.is_none() {
            return None
        } else {
            let s = ret.unwrap();
            match s.as_str() {
                "" => None,
                _ => Some(s),
            }
        }
``` 

and this doesn't look much better than what is in there now :-(

src/modules/mod.rs
jankatins: One bad side effect of this: if a module is now returning None, this code makes it more or less random if this is then a `None` or a `Some(Module)`, depending if it took more or less than a 1ms. I'm not familiar enough with the tests if this might result in flaky tests :-(

src/test/mod.rs
davidkna: Sorry I just ran `cargo check` and I forgot that doesnt include this function, try: `ret.and_then(|s| if s == "" { None } else { Some(s) })`

davidkna: Wait there's also `filter`: `ret.filter(|s| s != "")`

src/module.rs
jankatins: done

src/test/mod.rs
jankatins: Wow, nice!

jankatins: done

src/modules/mod.rs
```diff
@@ -55,64 +55,81 @@ mod battery;
...
+
+    let elapsed = start.elapsed();
+    if elapsed.as_millis() < 1 {
+        // If we take less than 1ms to compute a None, then we will not return a module at all
+        // if we have a module: default duration is 0 so no need to change it
+        m
+    } else {
+        // if we took more than 1ms we want to report that and so--in case we have None currently--
+        // need to create an empty module just to hold the duration for that case
+        m.get_or_insert(context.new_module(module)).duration = elapsed;
```
davidkna, line 131:
```
m.ok_or(|| context.new_module(module)).map(|m| m.duration = elapsed);`
```
Would be better because otherwise `context.new_module(module)` will be called whether the module exists or not.

src/test/mod.rs
```diff
@@ -89,16 +89,7 @@ impl<'a> ModuleRenderer<'a> {
...
-            Some(value) => {
-                if value == &String::from("") {
-                    None
-                } else {
-                    ret
-                }
-            }
-            None => None,
-        }
+        ret.filter(|s| s != &String::from(""))
```
davidkna, line 92: for me it works without `&String::from("")`
String literals are already `&str` and `&String` ought to be the same as `&str`.
```suggestion
        ret.filter(|s| s != "")
```

src/modules/mod.rs
davidkna: The tests ran successfully with and without that change so it seems to be fine. If it becomes an issues `Module` could be extended with an extra poperty to tell if it was created to provide timing information only.

src/modules/mod.rs
jankatins: Used `get_or_insert_with` to have the same effect but avoid the `ok_or` to take the ownership of `m`...

src/test/mod.rs
jankatins: Changed

davidkna (approved):

src/modules/mod.rs
```diff
@@ -55,64 +55,81 @@ mod battery;
...
+            "username" => username::module(context),
+            "zig" => zig::module(context),
+            _ => {
+                eprintln!("Error: Unknown module {}. Use starship module --list to list out all supported modules.", module);
+                None
+            }
         }
+    };
+
+    let elapsed = start.elapsed();
```
andytom, line 123: Minor suggestion but can we log the duration here?
Something like 
```suggestion
    let elapsed = start.elapsed();
    log::trace!("Took {:?} to compute module {:?}", elapsed, module);
```

andytom (commented):
Just a suggestion to add some extra logging.

src/modules/mod.rs
jankatins: Also added that to the custom module as well

jankatins:
Added a trace to both places and repushed -> all things are green :-)

andytom (approved):

andytom:
@jankatins thank you for your contribution and @davidkna thank you for reviewing.

PR #1632, opened 2020-08-31
feat: Make git_branch and kubernetes prefixes configurable

rbren:
#### Description

I _think_ I did this right.

I want to remove `on` from the the `git_branch` indicator. This will allow me to customize the prefix, but keeps the current behavior as default.

#### Motivation and Context

Closes #

#### Checklist:

let me know what I need to do here...
ditto

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -993,14 +993,15 @@ The `git_branch` module shows the active branch of the repo in your current dire
 
 ### Options
 
-| Option              | Default                          | Description                                                                              |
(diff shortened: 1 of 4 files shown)


rbren:
Figured out how to update the docs, and noticed the forthcoming `format` parameter. Happy to abandon this in favor of `format` if that's preferred.

andytom:
Hi @rbren, Yep with the new format string feature the expected way to change a module's prefix would be to modify the `format` option in the configuration file.

rbren:
👍 sounds good, closing

PR #1643, opened 2020-09-06
docs(readme): add nix installation instructions

dmedinag:
Just adding support for the Nix / NixOS community

#### Checklist:

- [ ] I have updated the tests accordingly. (not relevant?)


chipbuster:
We used to have install instructions for each platform, but we intentionally stripped them out simply because the README was becoming too cluttered.

I personally think that we might have gone too far in that direction, as there are a bunch of platform-specific installation methods (e.g. AUR on arch, using the MUSL static binary) that don't have documentation available anywhere, but we also don't want to clutter up the main page/README with this information. If you look at e.g. ripgrep, it has almost 5 full screen widths dedicated to install instructions--I don't think that's something we really want.

Perhaps a separate installation page with the quirks and alternative methods for each platform could work here?

dmedinag:
Hm, that's an interesting point. I have as well suffered from the never-ending installation section in other projects, and I wouldn't like starship's to look like that either.

> Perhaps a separate installation page with the quirks and alternative methods for each platform could work here?

I like this. Are you thinking of something like an INSTALLATION.md, referenced from the README.md? The README could certainly stick to 1 mainstream installation method per OS (Linux, Mac, Windows) and then offer a link to this separate INSTALLATION.md for smaller distros (?)

chipbuster:
Yeah, something like that. We actually need installation instructions per-shell (because of init script requirements), so the current README only has instructions for cargo, curl-bash, and scoop/brew (technically, only scoop is platform-restricted). I think the current README is fine as it is, and we can just link additional platform-specific instructions to an INSTALL.md (or INSTALLATION.md, or INSTALLING.md...I'm not really fussed about the name).

If you want to edit your PR to be the first set of instructions on that new page, and add a link into the README, I think that would go great!

chipbuster:
Hey @dmedinag 

I'm sorry for falling off the face of the Earth!

I had a chat with @matchai a few days ago and we agreed that this should probably be on the website, so I've pushed a set of small changes that will cause your work to be on starship.rs.

Right now, I just need to figure out what the correct target for the main README link is, and then we'll be able to review and hopefully merge this PR.

dmedinag:
Not an issue! More than happy to see it being investigated. Thanks :)

chipbuster:
Okay, after some thought, we may eventually want to make the extra install instructions dynamic (e.g. only fold out the section the user requests) so I think the correct thing to do is to point the link in the README to the website proper.

I think this is ready for review now.

chipbuster (approved):

chipbuster:
@dmedinag Thanks for your contribution!

PR #1645, opened 2020-09-07
fix: use suffixes for modules before directory

### LINKED ISSUE #1607: Module prefixes are shown at the start of the prompt in v0.45.0
#### Current Behavior

As discussed in #1116, the new format strings will now show prefixes before every module, including the first module in the prompt. This leads to problems when the user is root, for instance:

image

#### Expected Behavior

The first module does not have a prefix.

#### Possible Solution

One possible solution would be to replace module prefixes with appropriate suffixes. That said, we may run into issues at the end of the prompt where suffixes are shown. 🤔

matchai:
#### Description

Previously, all modules would have prefixes, which lead to the first
module having a dangling prefix. This change ensures that the first
few modules would instead have suffixes so that we don't start or
end with a prefix or suffix.

#### Motivation and Context

Closes #1607 

**Before**
image

**After**
image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -7,6 +7,7 @@ mkdir -p ~/.config && touch ~/.config/starship.toml
 ```
 
 All configuration for starship is done in this [TOML](https://github.com/toml-lang/toml) file:
+
 ```toml
 # Don't print a new line at the start of the prompt
 add_newline = false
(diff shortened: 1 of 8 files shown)


docs/config/README.md
```diff
@@ -93,15 +94,15 @@ For example:
 
 - `(@$region)` will show nothing if the variable `region` is `None`, otherwise `@` followed by the value of region.
 - `(some text)` will always show nothing since there are no variables wrapped in the braces.
-- When `$all` is a shortcut for `\[$a$b\] `, `($all)` will show nothing only if `$a` and `$b` are both `None`.
+- When `$all` is a shortcut for `\[$a$b\]`, `($all)` will show nothing only if `$a` and `$b` are both `None`.
   This works the same as `(\[$a$b\] )`.
```
heyrict, line 98: IDK if these changes are introduced by a markdown formatter, but the space is left intentionally here.  If the space is removed above, the corresponding one here should also be removed.

heyrict (approved):
There is a minor issue on the docs. All changes on suffixes LGTM. Thanks for the PR :+1:

docs/config/README.md
matchai: It looks like they were. 😥
I'll go through and fix them up.

PR #1646, opened 2020-09-08
fix(windows): configure console to enable ANSI support

### LINKED ISSUE #1533: Escape sequences are printed instead of colors after running a cygwin gui app
#### Current Behavior

Running a cygwin GUI app breaks the escape sequence parsing
image

#### Expected Behavior

image
Shell should retain colors and custom icons

#### Environment
- Starship version: 0.44.0
- powershell version: 5.1.17134.858
- Operating system: Windows 10 10.0.17134
- Terminal emulator: Windows Powershell

```bash
default configuration
```

#### Starship Configuration

```toml
default configuration
```

davidkna:
#### Description

Consoles on WIndows don't always have support for ANSI sequences enabled.
Call `ansi_term::enable_ansi_support` to configure the current console to enable support for ANSI sequences.

Potential problems:
`enable_ansi_support` only supports Windows 10, not sure what happens on Windows <10 with this change.
Maybe ANSI support should be disabled again after rendering the prompt?

#### Motivation and Context

Might close #1533

I tried testing it in the windows console in legacy mode, but it didn't work with that.
My normal shell already has ANSI support enabled so I wasn't able to test it with a configuration that didn't work previously, but it didn't break anything.

#### Checklist:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,6 +9,9 @@ use starship::module::ALL_MODULES;
 use starship::*;
 
 fn main() {
+    // Configure the current terminal on windows to support ANSI escape sequences.
+    #[cfg(windows)]
+    let _ = ansi_term::enable_ansi_support();
     logger::init();
 
     let status_code_arg = Arg::with_name("status_code")


matchai:
Looks like a good idea to me 👍
I'm in agreement that we should test this on Windows <10 before merging it, if someone has such a system available for testing.

davidkna:
I tested it in a Windows 8.1 VM and the change hasn't caused any additional issues.

andytom (approved):

andytom:
Thank you for your contribution @davidkna .

PR #1648, opened 2020-09-09
fix(pwsh): update init script to propagate $LASTEXITCODE and $?

jpoehls:
Previously the prompt function used in PowerShell would overwrite the $LASTEXITCODE and $? automatic variables that were set by the previous command run the user in the shell. This results in surprising behavior for the user if they inspect those variables looking for the result of the command they last ran.

This fixes the bug reported here: https://github.com/starship/starship/issues/1051
And goes further to also propagate the $? automatic variable which is not mentioned in that bug.

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -3,24 +3,54 @@
 # Starship assumes UTF-8
 [Console]::OutputEncoding = [System.Text.Encoding]::UTF8
 function global:prompt {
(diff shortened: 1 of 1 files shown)


matchai (approved):
Thank you for your contribution @jpoehls! 😄 
Looks like a good, well-documented solution. Great PR! 👍

PR #1649, opened 2020-09-10
fix(nodejs): Change nodejs symbol to the nerd-font nodejs one

whi-tw:
#### Description
Replace unicode hexagon with nodejs-specific nerd-font icon

#### Motivation and Context
The unicode hexagon symbol does not actually fit into a single column with a fixed-width font. [ (nf-mdi-nodejs)](https://www.nerdfonts.com/cheat-sheet?set=nf-mdi-nodejs) does. As starship requires a nerd font as a prerequisite, it's safe to assume that this symbol is available.

Visually checked on Linux, with FiraCode Nerd Font Mono at multiple sizes. 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1814,7 +1814,7 @@ By default the module will be shown if any of the following conditions are met:
 | Option              | Default                              | Description                                        |
 | ------------------- | ------------------------------------ | -------------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"` | The format for the module.                         |
-| `symbol`            | `"⬢ "`                               | A format string representing the symbol of NodeJS. |
+| `symbol`            | `" "`                               | A format string representing the symbol of NodeJS. |
 | `detect_extensions` | `["js", "mjs", "cjs", "ts"]`         | Which extensions should trigger this moudle.       |
 | `detect_files`      | `["package.json", ".node-version"]`  | Which filenames should trigger this module.        |
 | `detect_folders`    | `["node_modules"]`                   | Which folders should trigger this module.          |
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -85,7 +85,7 @@ In the second part, which is enclosed in a `()`, is a [style string](#style-stri
 For example:
 
 - `[on](red bold)` will print a string `on` with bold text colored red.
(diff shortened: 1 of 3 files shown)


andytom:
Hi @whi-tw, sorry for the delay in reviewing this PR, you will also need to update the tests in the [`src/modules/nodejs.rs`](https://github.com/starship/starship/blob/master/src/modules/nodejs.rs#L111) and the [default format string in the docs](https://github.com/starship/starship/tree/master/docs/config#nodejs). You only need to update the English docs the translations will be handled by CrowdIn (https://translate.starship.rs/) P.S. if you are fluent in any other languages and want to help we are always looking for new translators 😀.

chipbuster:
Related: #2270 (though I'm personally in favor of merging this one at the moment).

whi-tw:
@andytom, I've made those changes. I've also added a `wip` commit, which I'm happy to either be reworded into something sensible, or dropped if it's felt unnecessary. If it is reworded and merged, though, I'll drop the relevant text from my other commit message.

andytom (approved):
LGTM

chipbuster:
@whi-tw Thanks for the quick pickup!

docs/config/README.md
```diff
@@ -85,7 +85,7 @@ In the second part, which is enclosed in a `()`, is a [style string](#style-stri
 For example:
 
 - `[on](red bold)` will print a string `on` with bold text colored red.
-- `[⬢ $version](bold green)` will print a symbol `⬢ ` followed by the content of variable `version`, with bold text colored green.
+- `[⌘ $version](bold green)` will print a symbol `⌘ ` followed by the content of variable `version`, with bold text colored green.
```
vladimyr, line 88: @whi-tw Why did you pick ⌘ as a replacement instead of going with _new_ NodeJS symbol?

whi-tw: The new symbol doesn't render in the browser (or in a text editor if using an unpatched font)

Using a symbol that's included in most (all?) Fonts, it ensures that this part of the documentation is readable / understandable for anyone - especially as that section of the docs isn't nodejs-specific.

I thought I'd put this reasoning in the commit message, but it looks like I didn't force push my most recent message amendment before merging, so it still (confusingly) refers to the old symbol.

vladimyr: Ah right, that makes sense! Thanks for the explanation :+1:

PR #1651, opened 2020-09-12
feat(status): add the status module 

### LINKED ISSUE #193: Option to show job exit status in prompt

### LINKED ISSUE #882: Support for status codes

### LINKED ISSUE #1641: show the exit code of the last command executed if it failed
## Feature Request

Screenshot from 2020-09-03 19-13-35

I do this with a custom module #1639

jankatins:
Replacement for the original PR in #1165 by @glehmann updated with all the changes from master and adjusted re best practises  (as I understand them)  for `format`, `symbol` and `style` usage.

#### Description

A "status" module which shows the exit code of the last command

#### Motivation and Context
Closes #193
Closes #882
Closes #1481
Closes #1641 

(and the original PR: Closes: #1140 )

#### Checklist:


docs/config/README.md
```diff
@@ -2082,6 +2083,45 @@ The module will be shown if any of the following conditions are met:
...
 
+## Status
+
+The `status` module displays the exit code of the previous command.
+The module will be shown only if the exit code is not `0`.
+
+### Options
+
+| Variable   | Default                       | Description                                            |
+| ---------- | ----------------------- ------| ------------------------------------------------------ |
```
charego, line 2094: The formatting for "Options" is messed up, probably due to the extra space.

format

status

charego (changes requested):

charego:
Tested on MacOS and it seems to work, thanks! I would really like to see the pipestatus added eventually.

Compare the default fish prompt:

pipestatus

With this feature's behavior:

status

docs/config/README.md
jankatins: Fixed the out-of-space-space and missing space in the end, but the spacing in the symbol line seems to come from the chosen character and how big it is displayed in a specific font:

diff on the commandline:
image

vs 

jetbrains
image

vs 
gitkraken

image

jankatins:
If I understand pipestatus right this needs another commandline argument (to pass in a string for all the exit statuses) and changes to the zsh/bash integrations (to pass in the variable from these two shells which have that feature)? https://unix.stackexchange.com/questions/14270/get-exit-status-of-process-thats-piped-to-another

docs/config/README.md
```diff
@@ -2082,6 +2083,45 @@ The module will be shown if any of the following conditions are met:
 format = "via [🏎  $version](red bold)"
 ```
 
+## Status
+
+The `status` module displays the exit code of the previous command.
+The module will be shown only if the exit code is not `0`.
+
```
andytom, line 2090: Let's add a tip here to make it explicitly obvious that this module is disabled by default
```suggestion

::: tip
                                                              
This module is disabled by default.
To enable it, set `disabled` to `false` in your configuration file.
:::

```

andytom (commented):
Have added a suggestion to the docs. I'm wondering if the changes to the other modules, the `hg_branch` in particular might break anything for our users.

jankatins:
> I'm wondering if the changes to the other modules, the hg_branch in particular might break anything for our users maybe those changes should be done in a separate PR?

In theory, the only case it could break for someone if that person uses the default config (=no changes to the hg module) and is in a hg repo and that person expects a hg output.

The only non-breaking change you could make is setting the default `disabled=true` to `disabled=false` to match the status quo.

It's currently a bug: e.g. after reading the docs, I would have expected that I won't see a hg output even if I have a hg repo clone and I am in that dir.

All the other modules are not effected, I only pasted in the same comment.

Your call: should I tke it out?

jankatins:
For the pipeline status, I opened https://github.com/starship/starship/issues/1669

jankatins:
Fixed the merge conflict

andytom:
> > I'm wondering if the changes to the other modules, the hg_branch in particular might break anything for our users maybe those changes should be done in a separate PR?
> 
> In theory, the only case it could break for someone if that person uses the default config (=no changes to the hg module) and is in a hg repo and that person expects a hg output.
> 
> The only non-breaking change you could make is setting the default `disabled=true` to `disabled=false` to match the status quo.
> 
> It's currently a bug: e.g. after reading the docs, I would have expected that I won't see a hg output even if I have a hg repo clone and I am in that dir.
> 
> All the other modules are not effected, I only pasted in the same comment.
> 
> Your call: should I take it out?

My preference would be to remove the changes to the other modules from this PR and open a new PR for them. Since we auto-generate our release notes based on the squashed and merged PRs commit title, if we leave them in here the changes to the other modules won't show up in the changelog.

jankatins:
> My preference would be to remove the changes to the other modules from this PR and open a new PR for them.

removed the unrelated fixes and will open a new PR for them

andytom (approved):

andytom:
@jankatins Thank you for your contribution and thank you @glehmann for your original work on this.

PR #1652, opened 2020-09-13
feat: Add support for Windows cygwin/msys2 environment with bash/zsh/fish

### LINKED ISSUE #590: Starship cannot be loaded from Fish inside Git-Bash
#### Current Behavior
When executing the Fish shell inside an instance of the Git Bash SDK in Windows, Starship cannot be loaded inside Fish because it doesn't follow the terminal's path syntax. 

`starship init fish` yields:
```fish
source ("C:\Users\Amalia\.cargo\bin\starship.exe" init fish --print-full-init | psub)⏎  
```
which will not work, because MSYS2 (the terminal host) expect Unix-style routes.

#### Expected Behavior
`starship init fish` should work in MSYS2.

Since there are no Starship binaries for Windows, I installed Rust with `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh` inside the Git-Bash SDK window, and then issued `cargo install starship`.

#### Environment
- Starship version: 0.25.2
- Shell type: fish
- Shell version: 3.0.2
- Shell plugin manager: N/A
- Terminal emulator: Git-Bash SDK (https://github.com/git-for-windows/git-sdk-64), mintty 3.0.6 (x86_64-pc-msys)
- Operating system: Windows 10 1903

None, I have just installed the shell using `pacman`.

#### Starship Configuration
None, I have yet to initialize Starship.

#### Possible Solution
Use the plain executable name when constructing the source-able script.

### LINKED ISSUE #634: Support native windows (msys2) environment
Currently it's possible to build starship in native windows environment. Potentially this could be used in some environment like msys2 / cygwin, which allows shell env like zsh / bash.

Immediate problem I have noticed was path resolution to native binary when init shell could not work on windows path, which may need some interop.

- Prebuilt binary for windows native
- Resolving path resolution & additional future fixes to support zsh / bash on msys2

### LINKED ISSUE #1354: Can't make Starship work on a Fish (Cygwin) on Windows environment
## Bug Report
Can't make Starship work on a Fish (Cygwin) on Windows environment

#### Current Behavior
Fish hangs when I start it after configuring `config.fish` to use Starship

gr82WXpH2Y

#### Expected Behavior
Should work

After reading #1310 and #1341 , installing and verifying version with `/usr/local/bin/starship.exe --version`,
I now get another error:
```
fish: Unknown command: C:\\cygwin64\\usr\\local\\bin\\starship.exe
- (line 11):
    "C:\cygwin64\usr\local\bin\starship.exe" prompt --status=$exit_code --keymap=$keymap --cmd-duration=$starship_duration --jobs=(count (jobs -p))
    ^
in function 'fish_prompt'
in command substitution
```

I have contacted the fish team for help and following until [the end](https://github.com/fish-shell/fish-shell/issues/5151#issuecomment-646764438) of [my discussion here](https://github.com/fish-shell/fish-shell/issues/5151), it seems that there is a bug regarding paths.

> faho commented 33 minutes ago
> Okay, starship's windows version prints a broken path using windows-style \. I'm afraid that's a starship bug or at least incompatibility with cygwin.
>
> You'd have to check at what starship init fish --print-full-init prints, replace the broken paths and just put that directly in config.fish instead of going via source.

#### Possible Solution
Fix printing of paths so that the file is read correctly.

#### Environment
- Starship version: 0.42.0
- Shell type: fish
- Shell version: 3.1.0
- Shell plugin manager: None
- Terminal emulator: Cygwin (?) (C:\cygwin64\bin\mintty.exe)

\# ~/.config/fish/config.fish (please see why [here](https://github.com/fish-shell/fish-shell/issues/5151#issuecomment-646758518))
/usr/local/bin/starship.exe init fish --print-full-init | source

#### Starship Configuration
N/A

zoritle:
#### Description

Fix the problem with windows style path and posix style path.

#### Motivation and Context

Using cygpath.exe to convert windows native path to posix style when running on windows and target is bash/zsh/fish.
Other platform/shell did nothing.

fixes #1354
fixes #634
fixes #590
image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1846,6 +1846,7 @@ package, and shows its current version. The module currently supports `npm`, `ca
 - **mix** - The `mix` package version is extracted from the `mix.exs` present
 - **helm** - The `helm` chart version is extracted from the `Chart.yaml` present
 - **maven** - The `maven` package version is extracted from the `pom.xml` present
+- **meson** - The `meson` package version is extracted from the `meson.build` present
 
 > ⚠️ The version being shown is that of the package whose source code is in your
 > current directory, not your package manager.
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -162,6 +162,16 @@ fn extract_maven_version(file_contents: &str) -> Option<String> {
     None
 }
 
+fn extract_meson_version(file_contents: &str) -> Option<String> {
+    let file_contents = file_contents.split_ascii_whitespace().collect::<String>();
+
+    let re = Regex::new(r#"project\([^())]*version:'(?P<version>[^']+)'[^())]*\)"#).unwrap();
+    let caps = re.captures(&file_contents)?;
+
+    let formatted_version = format_version(&caps["version"]);
+    Some(formatted_version)
+}
+
 fn get_package_version(base_dir: &PathBuf, config: &PackageConfig) -> Option<String> {
     if let Ok(cargo_toml) = utils::read_file(base_dir.join("Cargo.toml")) {
         extract_cargo_version(&cargo_toml)
@@ -181,6 +191,8 @@ fn get_package_version(base_dir: &PathBuf, config: &PackageConfig) -> Option<Str
         extract_helm_package_version(&chart_file)
     } else if let Ok(pom_file) = utils::read_file(base_dir.join("pom.xml")) {
         extract_maven_version(&pom_file)
+    } else if let Ok(meson_build) = utils::read_file(base_dir.join("meson.build")) {
+        extract_meson_version(&meson_build)
     } else {
         None
     }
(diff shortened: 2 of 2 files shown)


vladimyr:
Nice work @zoritle 👍

If I'm reading it correctly there is no guard to ensure you are running inside Cygwin. `cygpath` is always invoked and if it is not present or fails there is native path fallback?
If that holds true there is one edge case where this could potentially fail. Imagine running under git bash inside the Cygwin bin directory containing `cygpath.exe`. In that case, the starship would get the wrong path?

src/init/mod.rs
```diff
@@ -33,8 +63,8 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
 
     let shell_basename = Path::new(shell_name).file_stem().and_then(OsStr::to_str);
 
-    let starship = path_to_starship()?.replace("\"", "\"'\"'\"");
-
+    let starship_native = path_to_starship_native()?.replace("\"", "\"'\"'\"");
+    let starship_posix = path_to_starship_posix()?.replace("\"", "\"'\"'\"");
```
vladimyr, line 67: Keep those two together as `native` and `posix` named struct members :bulb:

zoritle: I did this, and also refactored the quote logic into method.

zoritle:
1. Yes,I think invoking "cygpath.exe" not just translating the path, but also determine if you're in a cygwin/msys2 environment.
I also added some checking of environment variable,but I don't find a universal and elegant variable to use.
2. This shouldn't be a problem. According the document of std::process::Command:

https://doc.rust-lang.org/std/process/struct.Command.html
>If program is not an absolute path, the PATH will be searched in an OS-defined way.

So in this situation, the bin dir of cygwin is not inside the PATH of git-bash, the correct cygpath.exe should be invoked,and it's hard to get the absolute (and native) path of it.

seven-linglx:
Is this PR plan to merge into master in next version?

mingwandroid:
Is there any update on this? I'd love to use starship on MSYS2.

PR #1667, opened 2020-09-19
docs(git_status): Update git_status module docs

### LINKED ISSUE #1559: Git Status `show_sync_count` option no longer works with format strings

Following the changes to add the format string feature we no longer have a boolean for enabling the git status sync count, `show_sync_count`, and each one needs to be configured individually. 

Since it was common enough to have a single option we should re-add this as a configuration option.

We could just remove the option from the docs, add this to the migration guide, #1404, and leave this as a breaking change maybe adding an example to the docs for users who want an easy way to enable this.

andytom:
#### Description

Have updated the configuration docs for the `git_status` to remove a
non-existing option and to add an example for how to achieve the same
effect as with the new format strings.

#### Motivation and Context

Closes #1559

#### Checklist:


ATiltedTree (approved):
LGTM! :+1:

PR #1668, opened 2020-09-20
fix: restore the `pyenv_prefix` option to python

andytom:
#### Description

Have restored the `pyenv_prefix` option to the python module. This is
added as a new variable that will only be shown if `pyenv` is being used
to determine the version of python that is being used.

#### Motivation and Context

Fixes a regression introduced in the format-strings release

#### Checklist:


jimleroyer (approved):
Good catch, thanks for the fix.

PR #1672, opened 2020-09-21
fix(git-state): Fix discrepancy between v0.44.0 and master

ATiltedTree:
#### Description

Fixes a discrepancy between v0.44.0 and master:
On 0.44.0 the brackets in the `git_status` module weren't styled but they were on master. 

#### Motivation and Context

Minimize breaking changes

#### Checklist:


matchai (approved):
Good catch! 👍 
Thanks @ATiltedTree!

PR #1677, opened 2020-09-25
fix: actually disable per default disabled modules

jankatins:
#### Description

The default `disabled: true` is actually only available within the module (when the config struct is used and not the user toml) but not all (the hg_branch) modules checked it there again.

Document this in all places and add the check (+ test) to the hg_branch module.

#### Checklist:


cappyzawa (approved):
Thanks for finding this.
LGTM 👍

andytom:
@jankatins thank you for your contribution and @cappyzawa thank you for reviewing.

PR #1678, opened 2020-09-25
fix(kubernetes): Parse stacked kubeconfigs

### LINKED ISSUE #814: [kubernetes] Namespace not shown for context when using multiple kubeconfigs
#### Current Behavior
When using multiple kubeconfig files, the namespace set for a context sometimes does not appear.

#### Expected Behavior
The namespace for a context, if set, should always appear.

This issue occurs when the current context is defined in a kubeconfig file, **that isn't the first one listed in the `$KUBECONFIG` env var**. For example, consider the following `$KUBECONFIG` setting.

```bash
❯ echo $KUBECONFIG
/root/.kube/config:/root/.kube/other_config


❯ cat /root/.kube/config
apiVersion: v1
kind: Config
contexts:
- context:
    namespace: my-namespace
  name: my-context


❯ cat /root/.kube/other_config
apiVersion: v1
kind: Config
contexts:
- context:
    namespace: my-other-namespace
  name: my-other-context
```

If we set the current context to `my-context`, which is defined in the first kubeconfig file, we get the following.

```bash
❯ kubectl config use-context my-context
Switched to context "my-context".

☸ my-context (my-namespace)
❯
```

However, if we try setting the context to `my-other-context`, we get this.

```bash
❯ kubectl config use-context my-other-context
Switched to context "my-other-context".

☸ my-other-context
❯
```

#### Possible Solution
I believe the problem is in [this line](https://github.com/starship/starship/blob/dab27f33342af7711e3c78ab4bd59c6a532e4514/src/modules/kubernetes.rs#L19), where we set the configuration to index 0 of the loaded YAML documents array. This is fine for finding the current context, since this value is always stored in the first document. However, if this value references a context that is defined in a separate document, the namespace configured for that context will not be found.

I propose we supplement the logic in [these lines](https://github.com/starship/starship/blob/dab27f33342af7711e3c78ab4bd59c6a532e4514/src/modules/kubernetes.rs#L27-L36) to collate the `contexts` field of every parsed YAML document, before attempting to find the namespace of the current context.

#### Environment
- Starship version: 0.33.0
- bash version: GNU bash, version 5.0.3(1)-release (x86_64-pc-linux-gnu)

- Operating system: Debian 10
- Terminal emulator: iTerm <unknown version>

```bash

eval "$(starship init bash)"
```

#### Starship Configuration

```toml
prompt_order = [
    "kubernetes",
    "line_break",
    "character",
]

[kubernetes]
disabled = false
```

andytom:
#### Description

Have refactored the kubernetes module to better support stacked
kubeconfigs. Rather than trying to grab both the current-context and
namespace on a single pass we now do two passes. These changes should
fix the problems where the current-context is defined in one file but
the context and the namespace are defined in a different file.

#### Motivation and Context

Closes #814

#### Checklist:

### DIFF
--- a/src/modules/kubernetes.rs
+++ b/src/modules/kubernetes.rs
@@ -9,7 +9,9 @@ use crate::configs::kubernetes::KubernetesConfig;
 use crate::formatter::StringFormatter;
 use crate::utils;
 
-fn get_kube_context(contents: &str) -> Option<(String, String)> {
+fn get_kube_context(filename: path::PathBuf) -> Option<String> {
+    let contents = utils::read_file(filename).ok()?;
+
     let yaml_docs = YamlLoader::load_from_str(&contents).ok()?;
     if yaml_docs.is_empty() {
         return None;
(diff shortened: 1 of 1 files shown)


andytom:
Sorry for the repeated rebases and forced pushes have been trying to get the most recent changes from master so that the CI passes.

cappyzawa (approved):
LGTM

PR #1679, opened 2020-09-26
feat: Add more options for character symbols.

Sciencentistguy:
## Description
Currently, the error_symbol (red by default) is displayed when a command is aborted with control-C (returns code 130).
My change adds cancel_symbol (yellow by default), for when a command returns code 130.

I have also replaced vicmd_symbol with vicmd_success_symbol, vicmd_cancel_symbol, and vicmd_error_symbol. This allows the symbol to stay red or yellow when in vicmd mode.

#### Motivation and Context
This change is one I made because I personally wanted to use it, and I thought I'd upstream.

This change works on all my systems that run Arch Linux.
	- Two tests fail on windows, but they are not related to code that I have modified.

#### Checklist:


Sciencentistguy:
I don't know why the linux tests are failing here, as they pass on my machine.
I also don't have access to a macOS machine to test on, so I cannot test there.

andytom:
This PR also includes some unrelated changes including breaking some of our CI changes so I am going to close this for now. In a more broad idea I am not sure about adding options for individual codes as this is going to be harder to maintain in the future and harder for users to understand too maybe it would be better to have a something more general maybe a map of overrides for different codes and allow the user to configure the colours for differential codes themselves.

PR #1687, opened 2020-09-28
chore: update v0.45.0 removal notice

matchai:
#### Description

- Update the deprecation notice to clarify what was deprecated
- Add a link to docs sidebar with the migration guide

#### Motivation and Context

Closes #

#### Checklist:


jankatins (commented):
I usually use "depreciated" when it still works but a new way is there. In this case the old way does not work anymore, so "deprecated" feels wrong to me. I would find "replaced" more correct.

matchai:
@jankatins Good point 👍 
I've gone ahead and replaced the use of "deprecated" with "removed".

PR #1691, opened 2020-09-29
Fix compiling for iOS targets

aspenluxxxy:
#### Description
`nix` also doesn't provide `getgroups` for iOS, leading to a compile error.

- [N/A] I have tested using **MacOS**
- [N/A] I have tested using **Linux**
- [N/A] I have tested using **Windows**

#### Checklist:

- [N/A] I have updated the documentation accordingly.
- [N/A] I have updated the tests accordingly.


aspenluxxxy:
I've updated the commit to have a semantic naming scheme.

PR #1693, opened 2020-09-30
docs: Remove newline from beginning of example format config.

ericonr:
Having a newline there means that an empty newline is printed everytime
the user runs a command or presses Enter, which for short commands
and/or small terminal windows can noticeably decrease the information
density.


matchai (approved):
Good catch 👍 
Thank you for the quick fix. 😄

PR #1710, opened 2020-10-01
feat: prefix STATUS and PREEXEC_READY shell vars with STARSHIP_

bjornfor:
#### Description
I was surprised to see the STATUS and PREEXEC_READY variables being set in my shell. I think
it's better to have them in the STARSHIP_ "namespace". (Actually, STATUS
got the new name STARSHIP_CMD_STATUS.)

#### Motivation and Context
For cleanliness :-) And it's unfortunate if users try to use the generic name `STATUS` for themself, not knowing starship will overwrite the variable. 

* I ran `cargo test`:
  - on master: `test result: FAILED. 440 passed; 52 failed; 28 ignored; 0 measured; 0 filtered out`
  - on this PR: `test result: FAILED. 440 passed; 52 failed; 28 ignored; 0 measured; 0 filtered out`
  (So no change.)
* I installed starship from this PR onto my system (with bash).

#### Checklist:

- [x] I have updated the documentation accordingly. (I didn't see any documentation regarding to these variables, I think they are an implementation detail. The remaining uses of `STATUS` in the repo seem unrelated to this change, so I left it alone.)
- [x] I have updated the tests accordingly. (AFAICS, no tests need updating.)

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -8,9 +8,9 @@
 # data for commands like `slow | slow | fast`, since the timer starts at the start
 # of the "fast" command.
 
-# To solve this, we set a flag `PREEXEC_READY` when the prompt is drawn, and only
-# start the timer if this flag is present. That way, timing is for the entire command,
-# and not just a portion of it.
+# To solve this, we set a flag `STARSHIP_PREEXEC_READY` when the prompt is
+# drawn, and only start the timer if this flag is present. That way, timing is
+# for the entire command, and not just a portion of it.
 
 # Will be run before *every* command (even ones in pipes!)
 starship_preexec() {
@@ -18,8 +18,8 @@ starship_preexec() {
     local PREV_LAST_ARG=$1
 
     # Avoid restarting the timer for commands in the same pipeline
-    if [ "$PREEXEC_READY" = "true" ]; then
-        PREEXEC_READY=false
+    if [ "$STARSHIP_PREEXEC_READY" = "true" ]; then
+        STARSHIP_PREEXEC_READY=false
         STARSHIP_START_TIME=$(::STARSHIP:: time)
     fi
 
@@ -30,7 +30,7 @@ starship_preexec() {
 starship_precmd() {
     local NUM_JOBS
     # Save the status, because commands in this pipeline will change $?
-    STATUS=$?
+    STARSHIP_CMD_STATUS=$?
 
     # Evaluate the number of jobs before running the preseved prompt command, so that tools
     # like z/autojump, which background certain jobs, do not cause spurious background jobs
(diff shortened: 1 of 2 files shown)


bjornfor:
The macOS test failed in the CI, but I don't think it has anything to do with this PR:

```
test modules::hg_branch::tests::test_hg_disabled_per_default ... ok

failures:

---- modules::directory::tests::truncation_symbol_not_truncated_home stdout ----
thread 'modules::directory::tests::truncation_symbol_not_truncated_home' panicked at 'assertion failed: `(left == right)`
  left: `Some("\u{1b}[1;36m~/.tmpSuxaY7/a/subpath\u{1b}[0m ")`,
 right: `Some("\u{1b}[1;36m…/.tmpiu1BJg/.tmpSuxaY7/a/subpath\u{1b}[0m ")`', src/modules/directory.rs:1330:9


failures:
    modules::directory::tests::truncation_symbol_not_truncated_home

test result: FAILED. 519 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.10s

error: test failed, to rerun pass '--lib'
Error: Process completed with exit code 101.
```

bjornfor:
Seems CI is a bit flaky, it worked on the 2nd try.

chipbuster:
@bjornfor Yeah, our CI has some known issues, but I've never been able to replicate the problems locally. As far as I can tell, the only person who's ever triggered the race condition while on their own hardware was someone who was on an s390x with ~200 hardware threads. For the time being, we just rebuild when a race gets triggered.

src/init/starship.bash
```diff
@@ -8,18 +8,18 @@
...
 starship_preexec() {
     # Save previous command's last argument, otherwise it will be set to "starship_preexec"
     local PREV_LAST_ARG=$1
 
     # Avoid restarting the timer for commands in the same pipeline
-    if [ "$PREEXEC_READY" = "true" ]; then
-        PREEXEC_READY=false
+    if [ "$STARSHIP_PREEXEC_READY" = "true" ]; then
+        STARSHIP_PREEXEC_READY=false
         STARSHIP_START_TIME=$(::STARSHIP:: time)
```
nesara-cmd, line 23: []()

src/init/starship.bash
```diff
@@ -8,18 +8,18 @@
 # data for commands like `slow | slow | fast`, since the timer starts at the start
 # of the "fast" command.
 
-# To solve this, we set a flag `PREEXEC_READY` when the prompt is drawn, and only
-# start the timer if this flag is present. That way, timing is for the entire command,
```
nesara-cmd, line 12: []()

nesara-cmd (changes requested):

src/init/starship.bash
nesara-cmd: 131220

chipbuster (approved):
CI doesn't test changes to the init scripts, so I checked this out and tested it locally. Pleased to report that everything is working.

andytom:
Thank you for your contribution @bjornfor.

bjornfor:
Thanks!

PR #1717, opened 2020-10-01
docs: Add entry for Nushell configuration.

ryanpcmcquen:
#### Description
Add a listing for Nushell, which pairs beautifully with Starship!

#### Motivation and Context
Nushell and Starship go together like peanut butter and honey.

I've been using this configuration for months on my work Mac. It is also listed as the way to incorporate Starship in Nushell's official documentation.

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,14 @@
    starship init fish | source
    ```
 
+   #### Nushell
+
+   Add the following to Nushell's `$(config path)`:
+
+   ```toml
+   prompt = "echo $(starship prompt)"
+   ```
+
    #### Zsh
 
    Add the following to the end of `~/.zshrc`:


andytom:
I am not sure this is actually needed, I thought starship was built into nushell by default see https://www.nushell.sh/blog/2019/11/26/nushell-0_6_0.html.

ryanpcmcquen:
@andytom, that is no longer true:
https://www.nushell.sh/blog/2020/09/01/nushell_0_19.html#starship-is-now-external

andytom:
> @andytom, that is no longer true:
> https://www.nushell.sh/blog/2020/09/01/nushell_0_19.html#starship-is-now-external

Ah you are correct I haven't been been keeping up with the latest releases of Nu. In that case I think this might need more work since as pointed out in #1716, you will need to add the `--status` argument to get the character module to work correctly.

andytom:
I've converted this PR back to a draft as there is still some further work required (see #1716)

PR #1718, opened 2020-10-01
fix: Pass `--no-gpg-sign` to `git commit` in tests

lberrymage:
#### Description
Pass `--no-gpg-sign` to all instances of `git commit` in tests. This prevents the potential need for user interaction for those who have `commit.gpgSign` set to `true`, and in some cases, outright test failure.

#### Motivation and Context

Some users have `commit.gpgSign` set to `true` in their global git config, causing tests which run `git commit` to fail if the configured `user.signingKey` is not present, as is often the case with PGP smart cards or other methods of offline PGP key storage (for smart cards the commit may not outright fail, but bring up a dialog box asking for the user's PIN. However, this is also undesirable because tests shouldn't require user interaction).

I ran `cargo test --all-features -- -Z unstable-options --include-ignored` with a signing PGP subkey configured. This key is on a PGP smart card and is not present on the testing device.

Relevant portion of `$HOME/.gitconfig`:
```
[user]
    signingKey = XXXX
[commit]
    gpgSign = true
```

#### Checklist:

### DIFF
--- a/src/modules/git_state.rs
+++ b/src/modules/git_state.rs
@@ -324,13 +324,17 @@ mod tests {
         // Write a file on master and commit it
         write_file("Version A")?;
         run_git_cmd(&["add", "the_file"], Some(path), true)?;
-        run_git_cmd(&["commit", "--message", "Commit A"], Some(path), true)?;
+        run_git_cmd(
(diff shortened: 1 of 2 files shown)


andytom (approved):

andytom:
@lberrymage, thank you for your contribution.

PR #1735, opened 2020-10-04
fix(git): prevent tests from using git config

### LINKED ISSUE #1734: Git: tests fail due to global git config
#### Current Behavior
I've tried to build the latest version obtained from the Arch User Repositories. Before the actual build, all tests are run but one test related to the git module is failing. The same is happening in the main branch.

```
---- modules::git_state::tests::shows_merging stdout ----
thread 'modules::git_state::tests::shows_merging' panicked at 'assertion failed: `(left == right)`
  left: `Some("(\u{1b}[1;33mMERGING\u{1b}[0m) ")`,
 right: `None`', src/modules/git_state.rs:210:9
```

In my git config, I've set the merge strategy to use fast forward merges only:

```
[merge]
	ff = only
```

#### Expected Behavior
All tests should pass regardless of the user's git configuration.

#### Possible Solution
Don't set the merge strategy to fast forward only. Then all tests will pass.

#### Environment
- Starship version: 0.44.0
- zsh version: zsh 5.8 (x86_64-pc-linux-gnu)
- Operating system: Manjaro 20.1.1
- Terminal emulator: Alacritty 0.5.0

not relevant

rsmidt:
#### Description
Rust's process:Command inherits the current process's environment, hence any git command
will use the user's git config. This can lead to tests failing.

By setting the `HOME` env var to `/dev/null` we can prevent git from using the user's config.

#### Motivation and Context
Failing tests prevent successful package builds, e.g. from the AUR. 

Closes #1734 

#### Checklist:

### DIFF
--- a/src/modules/git_state.rs
+++ b/src/modules/git_state.rs
@@ -267,6 +267,8 @@ mod tests {
         let mut command = Command::new("git");
         command
             .args(args)
+            // By setting the home to /dev/null/ we prevent git from using the user's git config.
+            .env("HOME", "/dev/null")
             .stdout(Stdio::null())
             .stderr(Stdio::null())
             .stdin(Stdio::null());


src/modules/git_state.rs
```diff
@@ -267,6 +267,8 @@ mod tests {
         let mut command = Command::new("git");
         command
             .args(args)
+            // By setting the home to /dev/null/ we prevent git from using the user's git config.
+            .env("HOME", "/dev/null")
```
davidkna, line 271: I believe this needs to be `nul` on windows

PR #1745, opened 2020-10-06
fix: Switched pwsh profile to use dollar hook for a more accurate success/failure handling 

marcos-quezada:
#### Description
./src/init/starship.ps1
Switched the variable `$lastExitCodeForPrompt` to use `$global.?` value instead of `$global.LASTEXITCODE` to create the value to be passed as argument to starship command on the Powershell profile.

#### Motivation and Context
I am an active user of this project, extensively used this on `zsh` and now I'm switching to Powershell and I noticed that the prompt would be incorrectly coloured after execution of a internal `pwsh` command to green even on failure of the command.

Tho this is not a breaking issue, I found it annoying.

Since this is a change directly aimed to the Powershell profile, the kind of tests I performed where mostly installing the modified package and making sure:
`$global:LASTEXITCODE` was modified accordingly, and the prompt was corectly colored; for example, using commands such as `bash -c 'exit 5'`, `gci nonexistinfolder`, `ping nohost`, `Test-Connection nohost -Count 1` and their valid counterparts.
Ran a similar test using zsh and bash, to make sure the change didn't affected those. Further detail can be found on the added comment lines in-code.

#### Checklist:
- [X] I have added explanatory comments in-code.


src/init/starship.ps1
```diff
@@ -14,9 +14,30 @@ function global:prompt {
...
+    # 5
+    # Therefore, for the purpose of the prompt, to avoid coloring the prompt wrongly we should rely
+    # rather on the dollar hook ($?) to check for success/failure of the previous execution.
+    # There's some edge cases like iex where the execution is successful even when the command passed
+    # as parameter returns a failure exit code or fails to execute.
+    # As of Powershell 7 we can leaverage the ternary operator as follows:
+    # $lastExitCodeForPrompt = $origDollarQuestion ? 0 : 1
+    # But for backwards compatibility sake, we can get away with:
+
+    $lastExitCodeForPrompt = @({1}, {0})[$origDollarQuestion]
```
davidkna, line 40:
```suggestion
    $lastExitCodeForPrompt = 0
    if (-not $origDollarQuestion) {
         $lastExitCodeForPrompt = 1
         if ($origLastExitCode -neq $null) {
             $lastExitCodeForPrompt = $origLastExitCode
         }
    }
```
(I haven't tested this code yet)
I'd prefer something like this that preserves the exit code when it's available.

marcos-quezada: The only issue I see about the extra effort to preserve the exit code is that we can get to this problem:
```
> bash -c 'exit 5'
> $Global:LASTEXITCODE
5
> gci inexistentfolder
Get-ChildItem: Cannot find path '/home/my-user/inexistentfolder' because it does not exist
> $Global:LASTEXITCODE
5
``` 
I would check it myself later, but it would be easier if you can point me to a place in the Rust code where the exit code is relevant for some process, then I would totally agree that it is necessary to find to preserve the exit code and a way to tell apart whether the last command was an internal Powershell (exit code is irrelevant) command or an external one (exit code is relevant), if the exit code is only used in the `starship` command to define the color/symbol of the prompt I would argue that a simple 0 or 1 is enough.

davidkna: The exit code value is needed for displaying it in the [`status`  module](https://starship.rs/config/#status).

src/init/starship.ps1
```diff
@@ -14,9 +16,38 @@ function global:prompt {
...
+    # > bash -c 'exit 5'
+    # > $LASTEXITCODE -eq $null
+    # False
+    # > $LASTEXITCODE
+    # 5
+    # Whe start from the premise that the command executed correctly, which covers also the fresh console.
+    $lastExitCodeForPrompt = 0
+
+    # In case we have a False on the Dollar hook, we know there's an error.
+    if( -not $origDollarQuestion){
```
davidkna, line 38: Could this reuse ` if ($lastCmd = Get-History -Count 1) {` from below?

src/init/starship.ps1
```diff
@@ -5,6 +5,8 @@
 function global:prompt {
     $origDollarQuestion = $global:?
     $origLastExitCode = $global:LASTEXITCODE
+    $origCaretValue = ${global:^}
+    $origLastErrorCmdletName = Get-Error |  Where-Object {$_ -ne $null} | Select-Object -expand InvocationInfo | Where-Object {$_.InvocationName -ne $null} | Select-Object -expand InvocationName
```
davidkna, line 9: I ist possible to initialize this when it's needed?

src/init/starship.ps1
```diff
@@ -14,9 +16,38 @@ function global:prompt {
...
+    # 5
+    # Whe start from the premise that the command executed correctly, which covers also the fresh console.
+    $lastExitCodeForPrompt = 0
+
+    # In case we have a False on the Dollar hook, we know there's an error.
+    if( -not $origDollarQuestion){
+    # We check if the caret (^) original value is null aka. this is the first command executed.
+      if ($origCaretValue -eq $null){
+    # If there's a $LASTEXITCODE we use it, otherwise, we can safely return the standard error.
+        $lastExitCodeForPrompt = @($origLastExitCode,{1})[$origLastExitCode -eq $null]
```
davidkna, line 42: Please use `if/else` instead
```suggestion
        $lastExitCodeForPrompt = if ($origLastExitCode -eq $null) { 1 } else { $origLastExitCode } 
```

src/init/starship.ps1
```diff
@@ -14,9 +16,38 @@ function global:prompt {
...
+      }else{
+    # Finally if the last error origin is the same as the last command run, we know it was an internal
+    # Powershell command and we can return 1, otherwise, we use the $LASTEXITCODE
+        $lastExitCodeForPrompt = @({1},$origLastExitCode)[$origCaretValue -eq $origLastErrorCmdletName]
```
davidkna, line 46: please use `if` here too.

src/init/starship.ps1
marcos-quezada: Done, also check previous comment for clarification about moving this login to $lastCmd if block.

src/init/starship.ps1
marcos-quezada: Moved the logic to andle $LASTEXITCODE inside the $lastCmd if block as suggested to take advantage of that variable.

src/init/starship.ps1
marcos-quezada: Done, see previous comment as this logic was moved inside $lastCmd if block.

src/init/starship.ps1
marcos-quezada: Done, also see previous comments of the relocation of this logic inside the $lastCmd if block.

davidkna (approved):

andytom:
Thank you for your contribution @marcos-quezada and thank you for reviewing @davidkna.

PR #1746, opened 2020-10-06
feat: Adds a Disk Used module

### LINKED ISSUE #1589: Add a storage module
I would like to be able to see the free / used space of the current / specified / all directories while using the terminal.
When I am developing with docker over time my system gets clogged with old containers, networks and images. It will really be helpful if I can keep track of the used space from the shell (especially when I am working on a remote machine over ssh).

I create a feature request as per https://github.com/starship/starship/issues/209#issuecomment-631135400.

A new module that will display these values. I could look like:
`💾 (<drive mount point | drive name | empty>: <used percentage | free percentage>%) | (another drive)`
Example:
`💾 sda1: 60% | sda2: 20%` or just `💾 60%` and `💾 20%` depending on the drive the user is in.

I was using Powerlevel10k which has a plugin for it but I am trying to move from powerlevel10k to starship.
I could also keep checking the free space manually but the problem builds over time and I usually forget to check.

## Misc.
I looked into implementing this myself, and it looks pretty straight forward with the [sysinfo](https://crates.io/crates/sysinfo) crate which the memory module already uses.

SarthakSingh31:
#### Description
Adds a module that lets the user see the amount of disk used on the disk corresponding to their current directory or any other disk in their computer.

#### Motivation and Context
Closes #1589 

Screenshot from 2020-10-06 13-49-25
Screenshot from 2020-10-06 13-50-35
Screenshot from 2020-10-06 13-53-11
Screenshot from 2020-10-06 13-53-56

I tested the prompt on my linux computer but I only have drive. I need help testing this on other OS and computers with more drives.

Update: Also tested it on windows, the current path to disk resolution also works for it now.

- **MacOS**
  - [ ] Current path to disk resolution
  - [ ] Multiple disk rendering
- **Linux**
  - [X] Current path to disk resolution
  - [ ] Multiple disk rendering
- **Windows**
  - [X] Current path to disk resolution
  - [ ] Multiple disk rendering

#### Checklist:
I don't know how to write tests for this prompt, I couldn't find a way to mock the Disk objects without which it makes testing this very hard.

#### Benchmarks:
The first batch is with the module enabled and the second batch is with the module disabled:
Screenshot from 2020-10-07 15-09-09


src/modules/disk_used.rs
```diff
@@ -0,0 +1,323 @@
...
+            if volume_path_name == disk.get_mount_point() {
+                return Ok(disk);
+            }
+        }
+        Err(Box::new(DiskNotFoundError::new(PathBuf::from(
+            volume_path_name,
+        ))))
+    } else {
+        let last_error = unsafe { GetLastError() };
+        Err(Box::new(WinAPIVolumeError::new(format!(
```
davidkna, line 169: How about using `std::io::last_os_error()` instead of the custom error and logging the failure? Anyways the speed on windows appears to be unusable so it might make sense to just use sysinfo or not supporting windows at all.

docs/config/README.md
```diff
@@ -698,6 +699,76 @@ truncation_length = 8
...
+
+\*: This variable can only be used as a part of a style string
+\*\*: Disabled if `all_threshold` is not defined
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[memory_usage]
```
davidkna, line 762:
```suggestion
[disk_used]
```

davidkna:
On windows on my SMB share I get:
```
\\NAS\SHARE 💾
❯
```
Maybe some failures are not properly handled?

src/modules/disk_used.rs
SarthakSingh31: I didn't know that `std::io::last_os_error()` existed will switch to it. I will also look into if I can do anything about the performance

docs/config/README.md
SarthakSingh31: Good catch will fix

SarthakSingh31:
> On windows on my SMB share I get:
> 
> ```
> \\NAS\SHARE 💾
> ❯
> ```
> 
> Maybe some failures are not properly handled?

I was indeed missing a case in the formatting

src/modules/disk_used.rs
SarthakSingh31: I am going to leave the windows part in with the hopes that sysinfo gets faster on windows with future builds, since the module is disabled by default I don't think it should cause any issues.

docs/config/README.md
```diff
@@ -698,6 +699,76 @@ truncation_length = 8
 truncation_symbol = "…/"
 ```
 
+## Disk Used
+
+The `disk_used` module shows disk used in current directory or any disk specified.
```
davidkna, line 704: Please rename the module to `disk_usage` to align with the `memory_usage` module

src/modules/disk_used.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+fn format_disk_used(
+    disk: &Disk,
+    config: &DiskUsedConfig,
+    show_disk_name: bool,
+    add_separator: bool,
+    percentage_char: &'_ str,
+) -> Result<Vec<Segment>, Box<dyn Error>> {
+    let used_space = (disk.get_total_space() - disk.get_available_space()) as f64;
```
davidkna, line 39: I'd prefer if these were kept as `u64` like in `memory_usage` and converted to `f64` on demand

src/modules/disk_used.rs
```diff
@@ -0,0 +1,307 @@
...
+    )
+    .and_then(|formatter| {
+        formatter
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(style)),
+                "default_style" => Some(Ok(config.default_style)),
+                _ => None,
+            })
+            .map(|var| match var {
+                "name" if show_disk_name => match get_disk_name(disk) {
```
davidkna, line 72: For me the name always ends up empty on windows.

src/modules/disk_used.rs
```diff
@@ -0,0 +1,307 @@
...
+        // Strip out the nulls from the end
+        while let Some(last) = volume_path_name.last() {
+            if *last == 0 {
+                volume_path_name.pop();
+            } else {
+                break;
+            }
+        }
+        let volume_path_name = OsString::from_wide(&volume_path_name);
+        log::info!("Got path name: {:?}", volume_path_name);
```
davidkna, line 138: The log level for this is too high.
```suggestion
        log::trace!("Got path name: {:?}", volume_path_name);
```

src/modules/disk_used.rs
```diff
@@ -0,0 +1,307 @@
...
+            .iter()
+            .filter(|disk| match current_disk {
+                Some(current) => disk.get_name() != current.get_name(),
+                None => true,
+            })
+            .filter(|disk| should_display_disk(disk, threshold))
+            .collect();
+        let mut all_segments = Vec::new();
+
+        for (i, disk) in display_disks.iter().enumerate() {
```
davidkna, line 246: You don't need to change this but `filter_map` would work well here and avoid `collect`ing

docs/config/README.md
nesara-cmd: 👍

davidkna (changes requested):

alex-treebeard:
@davidkna I'm keen for this feature. How much work do you think this needs to get merged?

davidkna:
@davidkna I'd very much like this PR to properly work on Windows before merging, but beyond this code quality is good enough. `sysinfo` has since been replaced with `sys-info` (for the future I'm also looking at `systemstat`) in starship because it has been very slow on Windows in general, so I would prefer if this module used one of those libraries.

Closing this PR as stale.

PR #1751, opened 2020-10-08
perf: use built-in EPOCHREALTIME envvar to compute cmd_duration for ZSH5+

### LINKED ISSUE #1750: [Perf] Use Zsh's built-in time instead of sub-shell invocation to compute cmd_duration
We are currently invoking `starship time` in a sub-shell to compute `STARSHIP_START_TIME` for cmd_duration in the [init script](https://github.com/starship/starship/blob/master/src/init/starship.zsh#L62). The sub-shell takes a non-trivial toll on startup time. By avoiding sub-shell command invocation and using built-ins, we could see much better improvement in latency.

Since the computation happens in the shell->Rust hooks, any optimization might need to be shell-specific.

##### ZSH
ZSH provides access to [$EPOCHREALTIME](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#index-EPOCHREALTIME), which can be used to compute cmd_duration instead. This env-var has been present [since Zsh-5](https://github.com/zsh-users/zsh/blob/e27142d45686cacc6ed155e5045b97dd6243d44c/Src/Modules/datetime.c#L230-L237), which was released [in 2012](http://zsh.sourceforge.net/News/). Hence, relying on this should be a safe option and not break compatibility for existing clients.

###### ZSH - POC
Replace `STARSHIP_START_TIME=$("/usr/local/bin/starship" time)` in the [zsh-init script](https://github.com/starship/starship/blob/master/src/init/starship.zsh#L62) with `((STARSHIP_START_TIME = int(rint($EPOCHREALTIME * 1000))));`. This requires the addition of two zsh modules (`zsh/datetime` and `zsh/mathfunc`).

Benchmarks:

```
~ ❯ echo "No Prompt" && benchmark_zsh
No Prompt
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      58.8 ms ±   3.2 ms    [User: 29.7 ms, System: 22.9 ms]
  Range (min … max):    55.6 ms …  70.2 ms    50 runs

~ ❯ echo "Starship with Standard Time Measurement" && benchmark_zsh
Starship with Standard Time Measurement
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      69.5 ms ±   3.8 ms    [User: 34.7 ms, System: 27.7 ms]
  Range (min … max):    66.4 ms …  89.6 ms    50 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

~ ❯ echo "Starship with Time Measurement using ZSH var" && benchmark_zsh
Starship with Time Measurement using ZSH var
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      65.7 ms ±   2.9 ms    [User: 32.3 ms, System: 25.4 ms]
  Range (min … max):    63.0 ms …  76.9 ms    50 runs
```

As can be seen, this reduced the startup time of Spaceship from ~10ms to ~6ms (a win of 4ms). Note that the current 10ms launch time is due to optimizations I proposed [here](https://github.com/starship/starship/issues/1748).

##### Bash
Bash-5.0, released in 2019 Jan, also provides support for `$EPOCHREALTIME`. The solution for ZSH can be applied here as well. However, as pointed out by @matchai , older MacOS releases (older than Catalina, launched in 2019) ship with a very old version of bash (bash-3.2, I think). This is due to licensing issues. While users can opt to install newer versions of bash or an alternate shell, the default is old and there might still be a non-negligible userbase still using it. Hence, adopting it here would require a fallback to the current method for older bash-versions.

I don't use other shells (and even bash rarely) so don't have any narrative for them.

For ZSH, [prompt expansion](https://github.com/starship/starship/issues/1748) is also an option. `print -P %D{%s%3.}` can return the same output as well. However, the costliest component is still the sub-shell invocation so this method does not have high ROI.

I've tested the changes locally and they work fine. I'd be happy to raise a PR if this makes sense to the team. Thanks!

Gautham:
#### Description
We currently invoke `starship time` in a sub-shell to compute time, which is non-performant. By using $EPOCHREALTIME,
which is an inbuilt env-var in ZSH5, we can get better performance. This commit only targets ZSH5+ and maintains the
old behaviour.

Ref [docs on EPOCHREALTIME](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#index-EPOCHREALTIME).

#### Motivation and Context
The current approach is not very performant and this provides a reliable optimization to prompt startup for virtually all ZSH users. (ZSH5 released in 2012 so userbase not in ZSH5 will be small).
 
Closes #1750 

```
~ ❯ echo "No Prompt" && benchmark_zsh
No Prompt
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      58.8 ms ±   3.2 ms    [User: 29.7 ms, System: 22.9 ms]
  Range (min … max):    55.6 ms …  70.2 ms    50 runs

~ ❯ echo "Starship with Standard Time Measurement" && benchmark_zsh
Starship with Standard Time Measurement
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      69.5 ms ±   3.8 ms    [User: 34.7 ms, System: 27.7 ms]
  Range (min … max):    66.4 ms …  89.6 ms    50 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

~ ❯ echo "Starship with Time Measurement using ZSH var" && benchmark_zsh
Starship with Time Measurement using ZSH var
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      65.7 ms ±   2.9 ms    [User: 32.3 ms, System: 25.4 ms]
  Range (min … max):    63.0 ms …  76.9 ms    50 runs
```

I sourced the updated zsh-init directly (after replacing `::STARSHIP::`) and verified that the output from both definitions of the new command `__starship_get_time` return the same output.

#### Checklist:
- [*] I have updated the documentation accordingly.
- [*] I have updated the tests accordingly.

### DIFF
--- a/src/init/starship.zsh
+++ b/src/init/starship.zsh
@@ -16,6 +16,20 @@ starship_render() {
     PROMPT="$(::STARSHIP:: prompt --keymap="${KEYMAP-}" --status=$STATUS --cmd-duration=${STARSHIP_DURATION-} --jobs="$NUM_JOBS")"
 }
 
+# Defines a function `__starship_get_time` that sets the time since epoch in millis in STARSHIP_CAPTURED_TIME.
+if [[ $ZSH_VERSION == ([1-4]*) ]]; then
+    # ZSH <= 5; Does not have a built-in variable so we will rely on Starship's inbuilt time function.
+    __starship_get_time() {
+        STARSHIP_CAPTURED_TIME=$(::STARSHIP:: time)
+    }
+else
+    zmodload zsh/datetime
+    zmodload zsh/mathfunc
+    __starship_get_time() {
+        ((STARSHIP_CAPTURED_TIME = int(rint($EPOCHREALTIME * 1000))))
+    }
+fi
+
 # Will be run before every prompt draw
 starship_precmd() {
     # Save the status, because commands in this pipeline will change $?
@@ -24,8 +38,8 @@ starship_precmd() {
     # Compute cmd_duration, if we have a time to consume, otherwise clear the
     # previous duration
     if [[ -n "${STARSHIP_START_TIME+1}" ]]; then
-        STARSHIP_END_TIME=$(::STARSHIP:: time)
-        STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
+        __starship_get_time && STARSHIP_END_TIME=$STARSHIP_CAPTURED_TIME
+        ((STARSHIP_DURATION = STARSHIP_END_TIME - STARSHIP_START_TIME))
         unset STARSHIP_START_TIME
     else
         unset STARSHIP_DURATION
@@ -35,7 +49,7 @@ starship_precmd() {
     starship_render
 }
 starship_preexec() {
-    STARSHIP_START_TIME=$(::STARSHIP:: time)
+    __starship_get_time && STARSHIP_START_TIME=$STARSHIP_CAPTURED_TIME
 }
 
 # If precmd/preexec arrays are not already set, set them. If we don't do this,
@@ -74,7 +88,8 @@ else
     zle -N zle-keymap-select starship_zle-keymap-select-wrapped;
 fi
 
-STARSHIP_START_TIME=$(::STARSHIP:: time)
+__starship_get_time && STARSHIP_START_TIME=$STARSHIP_CAPTURED_TIME
+
 export STARSHIP_SHELL="zsh"
 
 # Set up the session key that will be used to store logs


src/init/starship.zsh
```diff
@@ -16,6 +16,20 @@ starship_render() {
...
+if [[ $ZSH_VERSION == ([1-4]*) ]]; then
+    # ZSH <= 5; Does not have a built-in variable so we will rely on Starship's inbuilt time function.
+    __starship_get_time() {
+        STARSHIP_CAPTURED_TIME=$(::STARSHIP:: time)
+    }
+else
+    zmodload zsh/datetime
+    zmodload zsh/mathfunc
+    __starship_get_time() {
+        ((STARSHIP_CAPTURED_TIME = int(rint($EPOCHREALTIME * 1000))))
```
davidkna, line 29: I think this will break [`show_milliseconds`](https://starship.rs/config/#command-duration) in the  command duration module.

Gautham: The output is time since epoch in millis as an integer, the same as `starship time`. This should be a drop-in replacement. Am I missing something? I used this approach because `show_milliseconds` is an option we support for `cmd_duration`. If not, there are simpler and more-widely supported options like `$SECONDS`.

Sample Output:
```
~ ❯ starship time
1602149653499
~ ❯ ((currentTimeInEpoch = int(rint($EPOCHREALTIME * 1000)))) && echo $currentTimeInEpoch
1602149677644
```

davidkna: Sorry the `* 1000` part mislead me.

Gautham:
Gentle reminder for this :)

davidkna (approved):

andytom:
Thanks for your contribution @Gautham and thanks for reviewing @davidkna

PR #1759, opened 2020-10-08
docs: Update defaults in config docs

andytom:
#### Description

Have updated some of the config docs to correct the git_status default
that was incorrect. Have also taken the time to switch some of the
format strings to use toml literal strings rather than normal strings to
simplify the escape strings.

#### Motivation and Context

Related to #1741

#### Checklist:


Snuggle (approved):
LGTM!

matchai (commented):
Is there a reason we've changed many of the double-quotes for single-quotes?
The rest of the docs appear to use double-quotes in TOML strings.

andytom:
> Is there a reason we've changed many of the double-quotes for single-quotes?
> The rest of the docs appear to use double-quotes in TOML strings.

Yep, so I was trying to avoid having to double escape some for the format strings (got the idea from https://starship.rs/config/#escapable-characters) but it looks like I might have updated a few that didn't need to be updated.

PR #1762, opened 2020-10-09
fix(build): add serde_derive to dependency list

davidkna:
#### Description

There's been a few CI failures with errors about missing `serde_derive` (#1756)
Add it to the dependency list to hopefully prevent this kind of test failure in the future.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1198,6 +1198,7 @@ dependencies = [
  "rayon",
  "regex",
  "rust-ini",
+ "serde_derive",
  "serde_json",
  "shell-words",
  "starship_module_config_derive",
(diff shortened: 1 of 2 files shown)


davidkna:
It looks like these failures still happen with the change (#1761).

andytom:
> But not to #1769. thinking

I mentioned this is Discord but I vaguely remember something like this happening before, I think it was something to do with the GH Actions cache getting messed up. Since it seemed to go away once we bumped a version and the `Cargo.lock` got updated and cleared the cache. I'm not sure but that could explain why it is passing on the Dependency updates.

PR #1777, opened 2020-10-12
perf(git_status): Improved git_status module performance

SarthakSingh31:
#### Description
There is a race condition that happens between https://github.com/starship/starship/blob/9044f9e14ca36a70d748810a0c4eda89d3f18391/src/modules/git_status.rs#L171
and 
https://github.com/starship/starship/blob/9044f9e14ca36a70d748810a0c4eda89d3f18391/src/modules/git_status.rs#L185
because
https://github.com/starship/starship/blob/9044f9e14ca36a70d748810a0c4eda89d3f18391/src/modules/git_status.rs#L184
takes some time to process. While line 184 is executing line 171 executes of the other maps and returns a `None`. To avoid this I moved the write lock to the earliest point possible. **This is not a long term solution** a race condition can still happen if the processor switches context to a different thread at the wrong point. The real solution would be to pre initialize the `RepoStatus` object.

#### Motivation and Context
Fixes bug with `RepoStatus` being created more than once.

Preformace diff: (the first one is with the fix)
Screenshot from 2020-10-12 15-19-15

Doesn't change any platform dependent code.
- [x] Ran the tests on linux

### DIFF
--- a/src/formatter/string_formatter.rs
+++ b/src/formatter/string_formatter.rs
@@ -229,7 +229,7 @@ impl<'a> StringFormatter<'a> {
                     StyleElement::Variable(name) => {
                         let variable = variables.get(name.as_ref()).unwrap_or(&None);
                         match variable {
-                            Some(style_string) => style_string.clone().map(|string| string),
+                            Some(style_string) => style_string.clone(),
                             None => Ok("".into()),
                         }
                     }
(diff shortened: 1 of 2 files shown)


matchai (approved):
LGTM! 👍 
Thank you for the perf boost!

PR #1782, opened 2020-10-13
fix(directory): preserve substitution order

### LINKED ISSUE #1765: [config] directory-module substitutions segment randomly chooses less specific replacement
#### Current Behavior

I'm using directory substitutions to shorten the path of some of my directories (see below for the configuration).

I try to replace `~/` with a home glyph (sticking to `H ` for this bug report) and `~/Sources` with another glyph (sticking to `Src ` for this bug report).

Every random call of the starship prompt, I get one of the two above mentioned substitutions (see the pictures below). I couldn't find pattern in the occurrence.

This only happens if there are more specific substitutions available.

#### Expected Behavior

The configured substitution is always used--even if one is a more specific substitution to the other.

- wrong substitution: wrong substitution
- correct substitution: correct substitution

#### Environment
- Starship version: 0.46.0
- Shell: fish 3.1.2
- Operating system: macOS 10.15.7
- Terminal emulator: iTerm.app 3.4.0beta9

I'm currently using a custom function in order to configure starship without messing up my prompt ;) Therefore the below fish configuration is the content of the function

```fish
	switch "$fish_key_bindings"
		case fish_hybrid_key_bindings fish_vi_key_bindings
			set keymap "$fish_bind_mode"
		case '*'
			set keymap insert
	end
	set --export STARSHIP_SHELL fish
	set --export STARSHIP_SESSION_KEY ("/usr/local/bin/starship" session)
	set --local --export STARSHIP_CONFIG ~/.config/starship.toml
	"/usr/local/bin/starship" prompt --status=$status --keymap=$keymap --cmd-duration=$CMD_DURATION --jobs=(count (jobs -p))
```

#### Starship Configuration

_(stripped to the relevant part--IMHO)_

```toml
# … lines deleted …
[directory]
style = "italic bg:#6F6A70 fg:#EEEEEE"
read_only_symbol = " "
read_only_style = "bold italic bg:#6F6A70 fg:#ff3322"
format = "[[ ](bg:#565257 fg:#6F6A70)$path[](bg:#565257 fg:#6F6A70)]($style)[$read_only]($read_only_style)"

[directory.substitutions]
"~/Sources" = "Src "
"~/" = "H "
# … lines deleted …
```

davidkna:
#### Description

Use `IndexMap` instead of `HashMap` to store directory substitutions in order to preserve the insertion order.

The extra crate dependency should be fine because it's already an indirect dependency.

#### Motivation and Context

Closes #1765

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -438,15 +438,6 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.8.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e91b62f79061a0bc2e046024cb7ba44b08419ed238ecbd9adbd787434b9e8c25"
-dependencies = [
- "autocfg",
-]
-
-[[package]]
-name = "hashbrown"
 version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d7afe4a420e3fe79967a00898cc1f4db7c8a49a9333a29f8a4bd76a253d5cd04"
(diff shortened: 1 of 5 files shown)


matchai (approved):
Nice catch!
Thank you for the fix, David 😄

PR #1787, opened 2020-10-13
fix: replace emoji for CMake and Erlang with NerdFonts

matchai:
#### Description

The emoji used for the CMake and Erlang modules appear to have little font support, as is shown on the emojipedia entries for the respective emoji:
https://emojipedia.org/triangle-with-rounded-corners/
https://emojipedia.org/three-networked-computers/

This PR replaces the emoji used with NerdFont icons for "Triangle Outline" and "Erlang".

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -452,7 +452,7 @@ The `cmake` module shows the currently installed version of CMake if:
 | Option     | Default                            | Description                                  |
 | ---------- | ---------------------------------- | -------------------------------------------- |
 | `format`   | `"via [$symbol$version]($style) "` | The format for the module.                   |
-| `symbol`   | `"🛆 "`                             | The symbol used before the version of cmake. |
+| `symbol`   | `"喝 "`                             | The symbol used before the version of cmake. |
 | `style`    | `"bold blue"`                      | The style for the module.                    |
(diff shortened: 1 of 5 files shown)


ryanpcmcquen:
Thank you @matchai!

ATiltedTree (approved):

wilbeibi:
Ha, this is why my cpp project show chinese character "喝" ("drink") in prompt.

DCjanus:
Hi, "喝" in Chinese means "drink", that's really weird for me, a Chinese native speaker.

matchai:
The character occupies the same Unicode codepoint as the symbol we are aiming to use in the Nerd Fonts version of fonts.

If you are seeing the Chinese character for "drink" try changing your current font to a Nerd Font patched version.

mattn:
This PR break the prompt for Japanese too. I think this have better to be reverted.

image

matchai:
@mattn, @DCjanus
It looks like the Nerd Fonts collision is caused by https://github.com/ryanoasis/nerd-fonts/issues/365.
We will be updating the prompt to use a new symbol until this issue is closed. 👍

PR #1795, opened 2020-10-15
feat: Add CMakeCache.txt detection to CMake module

cschlosser:
#### Description

For CMake it also makes sense to check if we're currently in the build folder.
This is done by checking for the existence of `CMakeCache.txt`.

#### Motivation and Context

Now it also marks a `build` folder as `CMake` related.
This is actually where must users will spend most of their time as it's the location where the actual build system is called.

#### Checklist:

### DIFF
--- a/docs/de-DE/config/README.md
+++ b/docs/de-DE/config/README.md
@@ -426,10 +426,9 @@ vicmd_symbol = "[V](bold green) "
 
 ## CMake
 
-The `cmake` module shows the currently installed version of CMake if any of the following conditions are met:
+The `cmake` module shows the currently installed version of CMake if:
 
 - The current directory contains a `CMakeLists.txt` file
-- The current directory contains a `CMakeCache.txt` file
 
 ### Optionen
 
--- a/docs/de-DE/config/README.md
+++ b/docs/de-DE/config/README.md
@@ -426,9 +426,10 @@ vicmd_symbol = "[V](bold green) "
 
 ## CMake
 
-The `cmake` module shows the currently installed version of CMake if:
+The `cmake` module shows the currently installed version of CMake if any of the following conditions are met:
 
 - The current directory contains a `CMakeLists.txt` file
+- The current directory contains a `CMakeCache.txt` file
 
 ### Optionen
 
--- a/src/modules/cmake.rs
+++ b/src/modules/cmake.rs
@@ -11,7 +11,7 @@ use crate::utils;
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let is_cmake_project = context
         .try_begin_scan()?
-        .set_files(&["CMakeLists.txt"])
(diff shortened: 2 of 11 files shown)


src/modules/cmake.rs
```diff
@@ -77,7 +77,17 @@ mod tests {
         let dir = tempfile::tempdir()?;
         File::create(dir.path().join("CMakeLists.txt"))?.sync_all()?;
         let actual = ModuleRenderer::new("cmake").path(dir.path()).collect();
-        let expected = Some(format!("via {} ", Color::Blue.bold().paint("喝 v3.17.3")));
+        let expected = Some(format!("via {} ", Color::Blue.bold().paint("喝 v3.17.3")));
```
andytom, line 80: Not sure what has happened here but it looks like the symbol used here has changed and that is why the tests are failing.
```suggestion
        let expected = Some(format!("via {} ", Color::Blue.bold().paint("喝 v3.17.3")));
```

src/modules/cmake.rs
```diff
@@ -77,7 +77,17 @@ mod tests {
...
+        assert_eq!(expected, actual);
+        dir.close()
+    }
+
+    #[test]
+    fn buildfolder_with_cmake_cache() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        File::create(dir.path().join("CMakeCache.txt"))?.sync_all()?;
+        let actual = ModuleRenderer::new("cmake").path(dir.path()).collect();
+        let expected = Some(format!("via {} ", Color::Blue.bold().paint("喝 v3.17.3")));
```
andytom, line 90: As above
```suggestion
        let expected = Some(format!("via {} ", Color::Blue.bold().paint("喝 v3.17.3")));
```

andytom (commented):
I've left a few suggestions to fix the changes can you also update the [config docs](https://github.com/starship/starship/blob/master/docs/config/README.md#cmake) to match these changes.

cschlosser:
Thanks for taking a look at it and figuring this out.
Even Github seems to have problems with it:
`Suggestion cannot be identical to original text. Please refresh and try again.`
Seems like I have to do this locally.

I will update the docs once the tests are working.

docs/de-DE/config/README.md
```diff
@@ -426,9 +426,10 @@ vicmd_symbol = "[V](bold green) "
```
andytom, line 426: Sorry I should have mentioned this earlier but can you remove the changes to non English documentation. All our translations are done using Crowdin, https://translate.starship.rs/, and it gets confused if the translations are changed outside of Crowdin. P.S we are always looking for people to help translate if you want to help out.

cschlosser: No problem. Updated the PR

andytom (approved):

andytom:
Thanks for your contribution @cschlosser

PR #1802, opened 2020-10-17
chore(context): Use monotonic clock for timeout

### LINKED ISSUE #1530: Use a monotonic clock for calculating the timeout in starship::context
We should switch to use a monotonic increasing clock when calculating the timeout in `starship::context`.

> That code should be using `Instant::now()`, not `SystemTime`. `SystemTime` is not monotonic, meaning the time could go backwards. The `unwrap()` in that code could panic if the system syncs with an NTP server, or daylight savings time happens, or the user is on a train traveling west, etc. [See docs](https://doc.rust-lang.org/std/time/struct.Instant.html)
> 
> And since this issue is about perf, I would guess that `Instant::now` is faster than `SystemTime::now`, since it doesn't have to sync with the system clock.

_Originally posted by @whatisaphone in https://github.com/starship/starship/pull/1499#issuecomment-664707458_

andytom:
#### Description

Have switched to use a monotonic clock for calculating the timeout when
indexing the current directory for the context to avoid any issues with
calculating the timeout when the systems clock might change.

#### Motivation and Context

Closes #1530

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -11,7 +11,7 @@ use std::ffi::OsString;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::string::String;
-use std::time::{Duration, SystemTime};
+use std::time::{Duration, Instant};
 
 /// Context contains data or common methods that may be used by multiple modules.
 /// The data contained within Context will be relevant to this particular rendering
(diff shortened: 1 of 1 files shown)


jimleroyer (approved):
Looks good!

PR #1810, opened 2020-10-19
fix(pwsh): preserve original console encondings

davidkna:
#### Description

Instead of forcing the whole shell to use utf-8 only change the `[Console]::OutputEncoding` before `starship prompt` is run and restore the old value afterwards.

#### Motivation and Context

Closes #1801

#### Checklist:


matchai (approved):
Good call! 👍

PR #1815, opened 2020-10-21
feat: Add a Lua module

cappyzawa:
#### Description

I added an module for [The Programming Language Lua](https://www.lua.org/).

#### Motivation and Context

Closes #

スクリーンショット 2020-10-22 3 41 04

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -228,6 +228,7 @@ $crystal\
 $cmd_duration\
 $custom\
 $line_break\
+$lua\
 $jobs\
 $battery\
 $time\
@@ -1487,6 +1488,42 @@ The `line_break` module separates the prompt into two lines.
 
 [line_break]
 disabled = true
+```
+
+## Lua
+
+The `lua` module shows the currently installed version of Lua.
+The module will be shown if any of the following conditions are met:
+
+- The current directory contains a `.lua-version` file
+- The current directory contains a file with the `.lua` extension
+
+### Options
+
+| Option     | Default                            | Description                                     |
+| ---------- | ---------------------------------- | ----------------------------------------------- |
+| `format`   | `"via [$symbol$version]($style) "` | The format for the module.                      |
+| `symbol`   | `"🌙 "`                            | A format string representing the symbol of Lua. |
+| `style`    | `"bold blue"`                      | The style for the module.                       |
+| `disabled` | `false`                            | Disables the `lua` module.                      |
+
+### Variables
+
+| Variable | Example   | Description                          |
+| -------- | --------- | ------------------------------------ |
+| version  | `v5.4.0`  | The version of `lua`                 |
+| symbol   |           | Mirrors the value of option `symbol` |
+| style\*  |           | Mirrors the value of option `style`  |
+
+\*: This variable can only be used as a part of a style string
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[lua]
+format = "via [🌕 $version](bold blue) "
 ```
 
 ## Memory Usage
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1501,12 +1501,13 @@ The module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Option     | Default                            | Description                                     |
-| ---------- | ---------------------------------- | ----------------------------------------------- |
-| `format`   | `"via [$symbol$version]($style) "` | The format for the module.                      |
-| `symbol`   | `"🌙 "`                            | A format string representing the symbol of Lua. |
-| `style`    | `"bold blue"`                      | The style for the module.                       |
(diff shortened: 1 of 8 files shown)


src/modules/lua.rs
```diff
@@ -0,0 +1,115 @@
...
+    Some(module)
+}
+
+fn format_lua_version(lua_stdout: &str) -> Option<String> {
+    // lua -v output looks like this:
+    // Lua 5.4.0  Copyright (C) 1994-2020 Lua.org, PUC-Rio
+
+    let version = lua_stdout
+        // split into ["", "5.4.0  Copyright (C) 1994-2020 Lua.org, PUC-Rio"]
+        .splitn(2, "Lua ")
```
davidkna, line 61: I'm not a lua dev, but is it possible for someone to have `luajit` as `lua` in their path? In that line of thinking a `lua_binary` option like in the `python module` might be useful.

This line wouldn't work with `luajit` because the version string is `LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/` on my system

cappyzawa: Thanks for tips!
I'll check it out 👍

cappyzawa:
@davidkna I've fixed based on https://github.com/starship/starship/pull/1815#discussion_r509579155 🙆‍♂️

src/modules/lua.rs
```diff
@@ -0,0 +1,170 @@
...
+    #[test]
+    fn lua_binary_is_luajit() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        File::create(dir.path().join("main.lua"))?.sync_all()?;
+
+        check_luajit_renders(&dir, None);
+        dir.close()
+    }
+
+    fn check_luajit_renders(dir: &tempfile::TempDir, starship_config: Option<toml::Value>) {
```
davidkna, line 143:
```suggestion
    #[test]
    fn check_luajit_renders(dir: &tempfile::TempDir, starship_config: Option<toml::Value>) {
```
I think this is missing a `#[test]` annotation.

cappyzawa: I referred to https://github.com/starship/starship/blob/master/src/modules/python.rs#L358.

I think this fn doesn't require annotation because this is called by other tests which has `#[test]` annotation.

davidkna: Sorry I didn't see that. The `assert_eq!(expected, actual);` tripped me up.

davidkna (approved):
LGTM

docs/config/README.md
```diff
@@ -1489,6 +1490,44 @@ The `line_break` module separates the prompt into two lines.
...
+- The current directory contains a file with the `.lua` extension
+
+### Options
+
+| Option       | Default                            | Description                                                                   |
+| ------------ | ---------------------------------- | ----------------------------------------------------------------------------- |
+| `format`     | `"via [$symbol$version]($style) "` | The format for the module.                                                    |
+| `symbol`     | `"🌙 "`                            | A format string representing the symbol of Lua.                               |
+| `style`      | `"bold blue"`                      | The style for the module.                                                     |
+| `lua_binary` | `"lua"`                            | Configures the python binary that Starship executes when getting the version. |
```
andytom, line 1509: Should this be lua not python?

src/modules/lua.rs
andytom: So the reason that this was done in a separate function in the python module was because we reused the functions across all the tests so we would test all the different scenarios worked correctly, I am not sure if that is what is being done here. Would it make sense to test against both lua binaries in all the tests here too?

docs/config/README.md
cappyzawa: Yes 😢 
I fixed it.

src/modules/mod.rs
```diff
@@ -94,6 +95,7 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
             "julia" => julia::module(context),
             "kubernetes" => kubernetes::module(context),
             "line_break" => line_break::module(context),
+            "lua" => lua::module(context),
             "memory_usage" => memory_usage::module(context),
             "nim" => nim::module(context),
             "nix_shell" => nix_shell::module(context),
```
andytom, line 101: Sorry just realised I missed this earlier can you add a description for the lua module to the description function below.

cappyzawa: I added it 👍

src/modules/lua.rs
cappyzawa: I merge this function in `lua_binary_is_luajit` because it only is called by `lua_binary_is_luajit`.

andytom (approved):

andytom:
Thank you for your contribution @cappyzawa  and thank you for reviewing @davidkna .

PR #1823, opened 2020-10-23
docs: Updated the Preset page to include more configurations

mickimnet:
#### Description

* made the Nerd Font Symbol preset downloadable & added information how to use it
* added the new Powerline style, based on the default configuration
* updated the preview picture, so that all current preset previews have one unified style
* added the M.y.t.h. presets

#### Motivation and Context

Show of the new `format` feature ;)

Preview-Preset-Page


docs/presets/README.md
```diff
@@ -3,91 +3,76 @@
 Here is a collection of community-submitted configuration presets for Starship.
 If you have a preset to share, please [submit a PR](https://github.com/starship/starship/edit/master/docs/presets/README.md) updating this file! 😊
 
-## Nerd Font Symbols
+## Default configuration with Nerd Font Symbols
 
 This preset doesn't change anything except for the symbols used for each module.
 If emojis aren't your thing, this might catch your eye!
```
jankatins, line 9: Btw: this sentence is ambiguous: i never figured out if this means the normal default need a nerdfont or the normal default work with a normal font (my guess, as i never installed a nerd font and it works).

mickimnet: @jankatins, most of the modules use emojis by default and the are not handled by the font. The above configuration is not using any emojis, but Nerd Font symbols instead.

The configuration contains a lot more settings, than mere colors & symbols. What phrasing would you suggest to clarify that only symbols and colors are changed in the first two configuration presets?

mickimnet:
Closed, because #1829 is a better replacement, IMHO.

nukopy:
Hi, Mick(@mickimnet).
Cc: @matchai 

I just have finished the implementation of the Gallery.
Let me explain to you the points of implementation.

## Demo

At first, please see the demonstration below.
I displayed the Gallery, visited GitHub to see source code of `starship.toml`, and download `starship.toml` of the prompt.

starship-gallery

## Points

There are two points.


- **Made UI very simple**
- **Added download button so that we don't have to right-click**

### Made UI very simple

IMHO, we just wanna show prompts so I displayed only the theme name and its image in Gallery, although creators information is important such as GitHub names, descriptions.

This is a reference to [the gallery of Visual Studio Code themes](https://vscodethemes.com/dark). This web site is very beautiful.

### Added download button so that we don't have to right-click

HTML `<a download>` is unavailable in some browsers, so I change this so that visitors just click "Download the starship.toml" button as shown in demo in any browser(I checked by Chrome, Firefox, Safari).

I'm looking forward to your constructive feedback.

Take care
Yosuke

PR #1836, opened 2020-10-27
feat(logger): don't create empty log files

davidkna:
#### Description

I noticed that the logger created lots of empty files on my system. This PR prevents that by only opening the log file for writing when needed.

I also removed `Arc` from `log_file` and `log_file_content` for simplicity. @ATiltedTree is there any reason these needed to be `Arc`?

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -1,17 +1,19 @@
 use ansi_term::Color;
 use log::{Level, LevelFilter, Metadata, Record};
+use once_cell::sync::OnceCell;
 use std::{
     collections::HashSet,
     env,
     fs::{self, File, OpenOptions},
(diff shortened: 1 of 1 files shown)


ATiltedTree:
There is not particular that these were `Arc`'s. I really only followed [this example](https://doc.rust-lang.org/std/sync/struct.Mutex.html#examples) which used `Arc`'s and so I did too.

ATiltedTree (approved):
LGMT! :+1:

andytom:
Thanks for the contribution @davidkna  and thank you for reviewing @ATiltedTree .

PR #1839, opened 2020-10-28
fix(bug-report): verify exit code of open, always print url

### LINKED ISSUE #1838: starship bug-report: does not work on remote systems
#### Current Behavior

Running `starship bug-report` on a remote system says it is loading detailed info into my web browser but it does not, since the system doesn't have a web browser or a way to start one on my local system.

#### Expected Behavior

I would expect it to detect that it cannot start a browser and provide an alternative (albeit less convenient) method to gather detailed information.

#### Possible Solution

Maybe just dump the text that would go to GitHub for easy cut-and-paste?

#### Environment
- Starship version: starship 0.46.2
- Shell type: zsh
- Shell version: zsh 5.4.2 (x86_64-ubuntu-linux-gnu)
- Shell plugin manager: zplug
- Terminal emulator: N/A

davidkna:
#### Description

`open::that` returns an `io::Result<ExitCode>` but only the `Result` was checked, not if the `ExitCode` indicated failure. This PR makes it verify both.

To prevent issues like #1838 in the future it might make sense to always print the URL. Should I add this to this PR too?

#### Motivation and Context

Closes #1838

#### Checklist:

### DIFF
--- a/src/bug_report.rs
+++ b/src/bug_report.rs
@@ -19,19 +19,19 @@ pub fn create() {
     };
 
     let link = make_github_issue_link(crate_version!(), environment);
+    let short_link = shorten_link(&link);
 
     if open::that(&link)
         .map(|status| status.success())
         .unwrap_or(false)
     {
-        print!("Take a look at your browser. A GitHub issue has been populated with your configuration")
+        println!("Take a look at your browser. A GitHub issue has been populated with your configuration.");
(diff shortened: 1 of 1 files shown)


andytom:
> To prevent issues like #1838 in the future it might make sense to always print the URL. Should I add this to this PR too?

I can't think of any downside to this other than we might print out a big wall of text which might be a pain on a small screen but that should be fine.

PR #1843, opened 2020-10-30
feat: Add i686-unknown-linux-musl build

### LINKED ISSUE #1835: Starship support for ISH on IOS
## Feature Request
It would be nice to be able to use Starship on ISH on a iOS device: https://ish.app
ISH is a linux emulator APP on iOS - it's emulating a x86 environment and runs a default Alpine Linux.

There is no i386 Build available.

Possibility to run Starship on ISH (Bash on Alpine Linux - usermode x86 emulator) 

#### Notes
**Get ISH ready:**

1. Run `wget http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86/apk-tools-static-2.10.5-r1.apk`
2. Run `tar xf apk-tools-static-2.10.5-r1.apk sbin/apk.static`
4. Run `./sbin/apk.static add apk-tools` to install apk
7. After this you can install `apk add curl` , bash, git, ...

equt:
#### Description

Add `i686-unknown-linux-musl` target into build matrix.

Currently, only the target itself is added, so there might be a good chance the cross-compile would fail.

#### Motivation and Context

Closes #1835

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,6 +44,7 @@ jobs:
           - x86_64-unknown-linux-musl
           - x86_64-apple-darwin
           - x86_64-pc-windows-msvc
+          - i686-unknown-linux-musl
         include:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -80,15 +84,15 @@ jobs:
           target: ${{ matrix.target }}
 
       - name: Setup | musl tools
-        if: matrix.target == 'x86_64-unknown-linux-musl'
+        if: matrix.target == 'x86_64-unknown-linux-musl' || matrix.target == 'i686-unknown-linux-musl'
```
davidkna, line 87: Seems to build fine for me locally so this might not break CI.

For the future I think rewriting the condition to something like this is better (I hope this syntax is correct):
```suggestion
        if:  ${{ endsWith(matrix.target, 'musl') }}
```

PR #1846, opened 2020-10-31
fix(gcloud): respect overrides of the active cloud config

### LINKED ISSUE #1845: Google Cloud module should respect the `CLOUDSDK_ACTIVE_CONFIG_NAME` variable
When set this variable overrides the active configuration name; it is not currently respected meaning that the shell may display incorrect gcloud details

jkp:
The active gclould configuration name can be overridden by setting the shell variable `CLOUDSDK_ACTIVE_CONFIG_NAME`. This patch ensures that a manual override is respected by the gcloud module: previous behaviour was to always look at the globally active config name which is lower in precedence than the contents of this variable. Fixes #1845.

### DIFF
--- a/src/modules/gcloud.rs
+++ b/src/modules/gcloud.rs
@@ -53,23 +53,26 @@ fn get_gcloud_region_from_config(current_config: &PathBuf) -> Option<Region> {
     Some(region.to_string())
 }
 
-fn get_active_config(config_root: &PathBuf) -> Option<String> {
-    let path = config_root.join("active_config");
-    let file = File::open(&path).ok()?;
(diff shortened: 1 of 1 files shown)


andytom (commented):
@jkp thanks for the contribution, this change LGTM but can you add a test case for this.

jkp:
yeah will do...just haven't had a chance to get back to it - busy week at work so far.

jkp:
@andytom pushed with a test case now.

andytom (approved):

andytom:
@jkp, thank you for the contribution.

PR #1847, opened 2020-11-01
feat: check node engines version in package.json

### LINKED ISSUE #1605: Show node version in red when incompatible with the current package
Starship already shows the node.js version, and the version of the module in the current dir from the package.json. Package.json can contain an [engines](https://docs.npmjs.com/files/package.json#engines) property that defines which versions of node it is compatible with. It'd be handy if the node.js version number were shown in red if it was not capable of running the current package.

t-mangoe:
#### Description

I added codes which check node.js engines version written in package.json

#### Motivation and Context

Closes #1605

#### Checklist:

 -> I added two tests.


Cargo.toml
```diff
@@ -30,16 +30,16 @@ tls-vendored = ["native-tls/vendored"]
 [dependencies]
 clap = "2.33.3"
 ansi_term = "0.12.1"
-dirs-next = "2.0.0"
```
andytom, line 33: Can you revert the downgrades to these libraries?

src/modules/nodejs.rs
```diff
@@ -59,12 +73,45 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
 }
 
+fn get_engines_version(base_dir: &PathBuf) -> Option<String> {
+    if let Ok(json_str) = utils::read_file(base_dir.join("package.json")) {
+        let package_json: json::Value = json::from_str(&json_str).ok()?;
+        let raw_version = package_json.get("engines")?.get("node")?.as_str()?;
+        Some(raw_version.to_string())
+    } else {
+        None
+    }
```
andytom, line 83: Not sure if it matters but you could simplify this a little bit to:
```suggestion
    let json_str = utils::read_file(base_dir.join("package.json")).ok()?;
    let package_json: json::Value = json::from_str(&json_str).ok()?;
    let raw_version = package_json.get("engines")?.get("node")?.as_str()?;
    Some(raw_version.to_string())
```

src/modules/nodejs.rs
```diff
@@ -59,12 +73,45 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+}
+
+fn check_engines_version(_nodejs_version: &str, _engines_version: Option<String>) -> bool {
```
andytom, line 86: It's generally better practice not to prefix variables with a `_` as it will stop the compiler from warning if the variable is not used.

```suggestion
fn check_engines_version(nodejs_version: &str, engines_version: Option<String>) -> bool {
```

src/modules/nodejs.rs
```diff
@@ -31,14 +37,22 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("nodejs");
     let config = NodejsConfig::try_load(module.config);
     let nodejs_version = utils::exec_cmd("node", &["--version"])?.stdout;
+    let _engines_version = get_engines_version(&context.current_dir);
+    let in_engines_range = check_engines_version(&nodejs_version, _engines_version);
```
andytom, line 41: As below
```suggestion
    let engines_version = get_engines_version(&context.current_dir);
    let in_engines_range = check_engines_version(&nodejs_version, engines_version);
```

andytom (commented):
Hi @t-mangoe thank you for your contribution. I have added a few suggestions hopefully nothing major.

Can you also update the Docs for the [NodeJS module](https://github.com/starship/starship/tree/master/docs/config#nodejs) to include this additional configuration (you only need to update the English docs all translations are done via [translate.starship.rs](https://translate.starship.rs/)).

t-mangoe:
@andytom Thank you for your review.
 Following your suggestions, I have fixed my codes.
 Also, I updated the Docs for the [NodeJS module](https://github.com/starship/starship/tree/master/docs/config#nodejs).

andytom:
> @andytom Thank you for your review.
> Following your suggestions, I have fixed my codes.
> Also, I updated the Docs for the [NodeJS module](https://github.com/starship/starship/tree/master/docs/config#nodejs).

Hi @t-mangoe, Sorry for the delay in replying. Thanks for fixing the issues unfortunately it looks like there are still some conflicts with the `Cargo.toml` if you can fix those conflicts I think we should be good to go here.

t-mangoe:
@andytom 
Oh, sorry. I overlooked the conflicts in Cargo.toml.
I resolved these conflicts.

t-mangoe:
I have resolved conflicts.

andytom (approved):

andytom:
Thank you for your contribution @t-mangoe

PR #1872, opened 2020-11-13
fix: one broken emphasis format in zh-CN guide.

xr9:
Just a quick markdown format fix.

### DIFF
--- a/docs/zh-CN/guide/README.md
+++ b/docs/zh-CN/guide/README.md
@@ -218,7 +218,7 @@
 
 ## 🤝 贡献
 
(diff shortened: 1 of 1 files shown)


andytom:
Hi @xr9, Thank you for this contribution, unfortunately since we use Crowdin for managing our translation, via https://translate.starship.rs we need all changes to the translated docs to go via Crowdin otherwise the get reverted. Would it be possible for you to add these change via https://translate.starship.rs?

matchai:
Thanks @xr9! The change has been accepted on Crowdin.
It should be visible on the docs once #1705 is merged.

PR #1879, opened 2020-11-14
perf(memory_usage): reduce the latency of memory usage fetching in linux.

iosmanthus:
#### Description

This pull request uses [`procfs`](https://docs.rs/procfs/0.8.1) to fetch memory usage in Linux, instead of [`sysinfo`](https://docs.rs/sysinfo/0.15.3) which will refresh some other fields while we just try to refresh memory info.

Here is a [benchmark](https://gist.github.com/iosmanthus/75d192ba0cb2ed45b2f3c04a4c24973e) of these two crates comparing with meminfo fetching:

```rust
    Finished bench [optimized] target(s) in 0.10s
     Running target/release/deps/meminfo-33685aca30a53811

running 2 tests
test benches::bench_procfs  ... bench:      31,920 ns/iter (+/- 7,817)
test benches::bench_sysinfo ... bench:     261,627 ns/iter (+/- 51,030)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured; 0 filtered out

```

#### Motivation and Context

In my environment, the fish prompt runs really slow while using `sysinfo`:
```sh
❯ starship timings
 memory_usage  -  435ms  -   using 🐏 56GiB/126GiB = 45% <--- about 500ms
 username      -    4ms  -   🎃 iosmanthus
 hostname      -   <1ms  -   ip-172-16-5-32 in
 directory     -   <1ms  -   ~
 line_break    -   <1ms  -   \n
 time          -   <1ms  -   🕙 19:59:42
 character     -   <1ms  -   ❯
```

While using `procfs` instead, the latency drops tremendously:
```sh
 username      -   5ms  -   🎃 iosmanthus
 hostname      -  <1ms  -   ip-172-16-5-32 in
 directory     -  <1ms  -   ~
 memory_usage  -  <1ms  -   using 🐏 56GiB/126GiB = 45% <--- less than 1ms
 line_break    -  <1ms  -   \n
 time          -  <1ms  -   🕙 20:18:56
 character     -  <1ms  -   ❯
```

This pull request also introduces a `meminfo` module in `modules/utils` which use a trait `MemoryUsgae` and a struct `MemInfo` to abstract the behaviour of different memory information fetching approaches which should work with other pull requests like https://github.com/starship/starship/pull/1573

Works fine in Windows, passed compilation:
image

#### Checklist:


davidkna:
I think it might make sense to just replace `sysinfo` with another library. For example [`sys-info`](https://crates.io/crates/sys-info) looks like a decent alternative with good platform support.
Beyond that I'm not sure if adding a library for procfs is necessary, parsing procfs seems to amount to `split_whitespace` and maybe handling the units.

PR #1887, opened 2020-11-15
feat(config): Add support for config composition using `$STARSHIP_CONFIG`

### LINKED ISSUE #2684: Add support for merging (or overriding) via 2 config files
I keep most of my dotfiles in a public Github repository, and although most of the Starship config is fine to share, there are some things like Kubernetes aliases that might refer to semi-sensitive information like AWS ARNs.

I would love to be able to keep most of my config is the standard location and then in the same directory have a file like `starship-override.toml` that could add/override the values provided in the standard config. This would allow someone to keep andything that might be considered sensitive in a separate file that is not kept in public source control.

TOML does not have a native way to pull in an external file, so any solution that isn't built into Starship would likely end up being pretty hacky.

fsommar:
Hi👋

This PR introduces support for config composition. By setting `$STARSHIP_CONFIG` to multiple paths separated by whatever your OS uses for PATH-like separation (most commonly `:`), the first mention of a value wins. This mimics the resolution logic as seen in [kubectl](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#merging-kubeconfig-files).

#### Motivation and Context

The motivation for config composition is to allow for conditional prompt modules while reusing the same basic config.

Conditional prompt setup is already possible by changing `$STARSHIP_CONFIG` via some external mechanism - such as dotenv. That, however, does not allow for reusing default configuration, and as such the config would need to be replicated across each separate config file.

By allowing `$STARSHIP_CONFIG` to point towards multiple config files, the problem of having repeated config is solved. This follows the precedence of other tools, maybe most noteworthy of which is kubectl, that reads all files in a config variable and concatenates them into one effective config.

There are effectively no changes to the regular code path of having just the one config in `$STARSHIP_CONFIG` and as such there should be no performance impact for that use case. Having multiple config files does not noticeably affect performance in my testing (20 files on the path).

While this doesn't directly solve any of the many tickets on conditional configuration, it may probably be used in conjunction with shell-based scripting (like hooking into the `fish_prompt` event in fish) to conditionally change only a small set of variables and might be good enough for some of the use cases. Here are some of the tickets I found on this topic:
#607 #680 #1001 #1081 #1235

Closes #2684

N/A

I've tested it with fish on WSL2 running Ubuntu on Windows 10. Even pointing `$STARSHIP_CONFIG` to files on the Windows disk (which is infamously slow for WSL2), there is still no noticeable performance impact with 20 files. And realistically I don't see people using more than a couple files for their config.

I've tested it with Powershell on Windows 10 where the separator for PATH-like variables is `;`, and that works as expected.

I've tested the same things with fish in MacOS and it works as expected there as well.

#### Checklist:

I've updated the documentation in English. What's the process is for documentation in other languages?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -10,9 +10,9 @@ use serde::{
 use std::borrow::Cow;
 use std::clone::Clone;
 use std::collections::HashMap;
-use std::io::ErrorKind;
-
 use std::env;
+use std::ffi::OsStr;
+use std::io::ErrorKind;
 use toml::Value;
 
 /// Root config of a module.
@@ -108,6 +108,7 @@ pub struct StarshipConfig {
     pub config: Option<Value>,
 }
 
+/// Get the contents of the starship config environment variable
 pub fn get_config_path() -> Option<String> {
     if let Ok(path) = env::var("STARSHIP_CONFIG") {
         // Use $STARSHIP_CONFIG as the config path if available
@@ -123,25 +124,10 @@ pub fn get_config_path() -> Option<String> {
     }
 }
 
-impl StarshipConfig {
-    /// Initialize the Config struct
-    pub fn initialize() -> Self {
-        if let Some(file_data) = Self::config_from_file() {
-            Self {
-                config: Some(file_data),
-            }
-        } else {
-            Self {
-                config: Some(Value::Table(toml::value::Table::new())),
-            }
-        }
-    }
-
-    /// Create a config from a starship configuration file
-    fn config_from_file() -> Option<Value> {
-        let file_path = get_config_path()?;
-
-        let toml_content = match utils::read_file(&file_path) {
+/// Reads config(s) from a path by separating values by the OS's path separator (usually `:`).
+pub fn read_configs(path: impl AsRef<OsStr>) -> Vec<String> {
+    env::split_paths(&path)
+        .filter_map(|path| match utils::read_file(&path) {
             Ok(content) => {
                 log::trace!("Config file content: \"\n{}\"", &content);
                 Some(content)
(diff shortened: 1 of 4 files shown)


00dani:
Is there a good chance of this being merged soon? I found myself wanting the ability to use multiple TOML files for Starship configuration yesterday, because I wanted to install the [Nerd Font symbols preset](https://starship.rs/presets/#nerd-font-symbols) and keep my own configuration separate from it. This sounds like exactly what I need.

fsommar:
I can rebase this PR if there is interest from maintainers in getting it merged

AlanGreene:
I've been looking for a way to customise certain features on different machines without having to duplicate all of the base config so this would be perfect for my use case. Happy to help out with testing or anything else needed to get this merged.

asymmetric:
I think a simpler approach would be to provide env vars to toggle a certain module on off.

So your base config would e.g. have `gcloud.disabled = true`, and then in an `.envrc` or similar you'd set `GCLOUD_DISABLED=false`.

jankatins:
I also looked for such a feature and wanted to use it to overwrite "only" all the symbols because I don't want to install a nerd font. That would enable putting two versions of this on the website to simply simply switch to non-nerdfont symbols without going through your own config and integrating it there.  

Re env vars: having them for enable/disable for each module would be very welcome and a really great combination with e.g. direnv :-)

zzamboni:
This would be a great addition, would make much easier to load presets without having to modify your main config.

@asymmetric your suggestion would be nice (e.g. allowing to conditionally disable modules by manipulating the environment, which could happen automatically based on the current directory, for example), but I don't think it offers the same abilities, e.g. as mentioned above to load presets.

cappyzawa (commented):
This looks like so a cool feature ✨ 

Could you resolve conflicts, please?
Sorry for my late review.

ctran:
Ping @fsommar to resolve conflicts.  Thanks.

fsommar:
@cappyzawa @ctran I've rebased over master.

cappyzawa (commented):
Thanks for responding.

I have only one comment.
Other than that LGTM.

src/config.rs
```diff
@@ -137,36 +143,69 @@ impl StarshipConfig {
         }
     }
 
-    /// Create a config from a starship configuration file
-    fn config_from_file() -> Option<Value> {
-        let file_path = get_config_path()?;
```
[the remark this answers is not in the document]
fsommar: Done!

src/configure.rs
```diff
@@ -320,7 +320,14 @@ fn get_editor_internal(visual: Option<String>, editor: Option<String>) -> String
 
 fn get_config_path() -> OsString {
     if let Some(config_path) = env::var_os("STARSHIP_CONFIG") {
-        return config_path;
+        let mut paths = env::split_paths(&config_path);
```
fsommar, line 323: I noticed that there was one more place where `$STARSHIP_CONFIG` was read: in editing or toggling values when invoked as a CLI. Does this look sensible? I could bubble up the error but it'd require more invasive changes as this function is called from multiple entrypoints that do not return a Result themselves.

src/config.rs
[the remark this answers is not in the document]
fsommar: Turns out it's still used in the git status module on WSL

fsommar:
I decided to align all the places where `$STARSHIP_CONFIG` is read. I abort from within `crate::configure::get_config_path` to keep the function signatures in that crate the same as they're exported outside of the crate – the alternative being to return a Result from that function.

fsommar:
If you're happy with the direction I can add more tests later

src/config.rs
```diff
@@ -437,10 +459,31 @@ fn get_palette<'a>(
     }
 }
 
+/// Merge two TOML configs by mutating the first argument
+fn merge_toml(base: &mut Value, value: &Value) {
```
davidkna, line 463: This could avoid cloning by taking `value` directly by value and using `into_iter()` instead of `iter()`:
```suggestion
fn merge_toml(base: &mut Value, value: Value) {
```

src/config.rs
```diff
@@ -437,10 +459,31 @@ fn get_palette<'a>(
...
+            // Recurse further into base table if keys match to continue merging.
+            if let Some(base) = base.get_mut(key) {
+                merge_toml(base, value)
+            } else {
+                let _ = base.insert(key.clone(), value.clone());
+            }
+        }
+    } else {
+        log::warn!(
+            "Cannot merge TOML config of type {} into table, ignoring value",
```
davidkna, line 475: `base` may not be a table I think.

cappyzawa:
@fsommar 
> I decided to align all the places where $STARSHIP_CONFIG is read. I abort from within crate::configure::get_config_path to keep the function signatures in that crate the same as they're exported outside of the crate – the alternative being to return a Result from that function.

Sounds good for me 👍

mjperrone:
I'm interested in this being merged. Wondering what's left in case someone else wanted to pick up where you left off @fsommar

PR #1897, opened 2020-11-17
fix(bash): Fix number of background jobs

### LINKED ISSUE #1776: jobs module always displays symbol ✦
#### Current Behavior

The jobs module always shows ✦ for me even when there are no jobs. When I have N jobs, it shows ✦(N+1).

Screenshot_2020-10-12_00-11-25

This is caused by the fact that my `$_PRESERVED_PROMPT_COMMAND` includes [`autojump_add_to_database`](https://github.com/wting/autojump/blob/master/bin/autojump.bash#L38) which creates a subshell and backgrounds. At the time that `jobs -p | wc -l` is run, there in fact is a background job. I believe this to also be the case with [z](https://github.com/rupa/z/).

#### Expected Behavior

I expect there to be no ✦ when I have no jobs and ✦N when I have N jobs. Even more, I would like to be able to hide ✦ when N &leq; threshold which would tidy this up nicely.

I can see how this might just be construed as an an incompatibility with things like autojump and z, but I believe those tools are commonly used.

#### Possible Solution

I can see any of three paths forward:

1. Filter out specific jobs from the `jobs -p | wc -l` call. This filter could be user-defined, Starship-defined, or user-augmentable.
2. Allow configuration of hiding the symbol ✦ when number of jobs &leq; threshold, possibly by introducing a `symbol_threshold`.
3. Return none instead of "" for `$number` when the number of jobs &leq; threshold so that I could use `"(✦$number)" as my format string. (Say, what's `$symbol` for anyway?)

Options 1 and 2 are debatable and more complex in nature. Option 3 to me seems to be a clear improvement without much downside or foreseen issues with existing users. I have implemented it at https://github.com/aduong/starship/tree/conditional-jobs for my own use.

#### Environment

- Starship version: 0.46.0
- bash version: GNU bash, version 5.0.3(1)-release (x86_64-pc-linux-gnu)
- Operating system: Ubuntu 19.10
- Terminal emulator: xfce4-terminal

I believe the relevant parts of my bashrc are as follows

```bash
test -e /usr/share/autojump/autojump.bash && . /usr/share/autojump/autojump.bash

...

eval "$(starship init bash)"
```

but [this is the full bashrc](https://github.com/aduong/dotfiles/blob/starship/bashrc).

#### Starship Configuration

The relevant part of my starship configuration is

```toml
[jobs]
threshold = 1
format = "(✦$number )"
```

but [this is the full starship.toml](https://github.com/aduong/dotfiles/blob/starship/starship.toml).

haraldh:
Without this patch, I always get the indicator for background jobs.
Also it does not rely on `wc` to count.

#### Motivation and Context

Closes #1776 

#### Checklist:

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -28,8 +28,11 @@ starship_preexec() {
 
 # Will be run before the prompt is drawn
 starship_precmd() {
+    local NUM_JOBS
     # Save the status, because commands in this pipeline will change $?
     STATUS=$?
+
+    NUM_JOBS=$(n=0; while read line; do [[ $line ]] && n=$((n+1));done <<< $(jobs -p) ; echo $n)
 
     # Run the bash precmd function, if it's set. If not set, evaluates to no-op
     "${starship_precmd_user_func-:}"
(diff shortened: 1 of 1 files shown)


chipbuster:
@haraldh I think not using wc alone would make this worth merging, but I'm curious to know what you mean by "always get the indicator for background jobs." This almost never happens when using starship alone--are you using other shell frameworks as well (e.g. z or autojump)?

haraldh:
I am using [`zoxide`](https://github.com/ajeetdsouza/zoxide) which is basically `z`, yes.

chipbuster:
I'll do some testing later when I have time. I can't find the open issue for this, but I remember most people specifically reporting that zoxide did *not* suffer from that problem (or at least that when they switched from z.sh to zoxide, the problem went away).

chipbuster (approved):
Well I will be damned...I've done some testing on this patch and can confirm that it fixes #1776 for both autojump and z.sh.

Gotta be honest: I don't even understand why this bug was showing up in the first place, but I'll take it. Code has been verified to work on the MacOS default bash.

I'll modify the PR to note that it closes that issue.

chipbuster:
@haraldh Thank you so much for your contribution!

PR #1903, opened 2020-11-17
fix: bg:none overwrites foreground colour

ocdy1001:
#### Description
I changed the parsing of styles a bit so that bg:none will not override other fields.

#### Motivation and Context
Background does not have to have a colour. You can do that with `red` or `fg:red` for example. It is confusing if `bg:none` will override foreground just like a standalone `none`.

This has been tested with some new test cases. I also tested it in a Fish shell in a suckless st terminal on Artix Linux 5.9.6.

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -99,7 +99,7 @@ Style strings are a list of words, separated by whitespace. The words are not ca
 
 where `<color>` is a color specifier (discussed below). `fg:<color>` and `<color>` currently do the same thing , though this may change in the future. The order of words in the string does not matter.
 
-The `none` token overrides all other tokens in a string if it is not part of a `bg:` specifier, so that e.g. `fg:red none fg:blue` will still create a string with no styling. `bg:none` does nothing so `fg:red bg:none` is equivalent to `red` or `fg:red`. It may become an error to use `none` in conjunction with other tokens in the future.
+The `none` token overrides all other tokens in a string if it is not part of a `bg:` specifier, so that e.g. `fg:red none fg:blue` will still create a string with no styling. `bg:none`  sets the background to the default color so `fg:red bg:none` is equivalent to `red` or `fg:red` and `bg:green fg:red bg:none` is also equivalent to `fg:red` or `red`. It may become an error to use `none` in conjunction with other tokens in the future.
 
 A color specifier can be one of the following:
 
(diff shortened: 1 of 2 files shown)


nesara-cmd:
[]()

ocdy1001:
Little extra fix: `big:none` now also overwrites the previous background colour.

docs/advanced-config/README.md
```diff
@@ -99,7 +99,7 @@ Style strings are a list of words, separated by whitespace. The words are not ca
 
 where `<color>` is a color specifier (discussed below). `fg:<color>` and `<color>` currently do the same thing , though this may change in the future. The order of words in the string does not matter.
 
-The `none` token overrides all other tokens in a string, so that e.g. `fg:red none fg:blue` will still create a string with no styling. It may become an error to use `none` in conjunction with other tokens in the future.
+The `none` token overrides all other tokens in a string if it is not part of a `bg:` specifier, so that e.g. `fg:red none fg:blue` will still create a string with no styling. `bg:none` does nothing so `fg:red bg:none` is equivalent to `red` or `fg:red`. It may become an error to use `none` in conjunction with other tokens in the future.
```
chipbuster, line 102: Is the second change here correct? My impression was that `bg:none` explicitly sets a "transparent" background (or rather, a background that matches the one in the rest of the terminal).

src/config.rs
```diff
@@ -684,6 +694,37 @@ mod tests {
         assert!(<Style>::from_config(&config).is_none());
     }
 
+    #[test]
```
chipbuster, line 697: Woowee, love me a good test function :+1:

src/config.rs
```diff
@@ -375,16 +375,26 @@ pub fn parse_style_string(style_string: &str) -> Option<ansi_term::Style> {
...
-                    // Try to see if this token parses as a valid color string
-                    color_string => parse_color_string(color_string).map(|ansi_color| {
-                        if col_fg {
-                            style.fg(ansi_color)
+                    color_string => {
+                        if color_string == "none" && col_fg {
+                            None
                         } else {
-                            style.on(ansi_color)
+                            let parsed = parse_color_string(color_string);
```
chipbuster, line 382: I find this block a little hard to understand on quick scan. Could you add a comment here (briefly) explaining what the goal here is?

chipbuster (commented):
Just a few small fixes and I think this'll be good to merge!

docs/advanced-config/README.md
ocdy1001: This is indeed incorrect, I forgot to update it in my second fix I did. I have now updated it to match the current behavior.

src/config.rs
ocdy1001: Here we decide what to do with the string when it is supposed to be a color. Either yield a None style, reset background color or set the appropriate fg/bg color. Comments added in the code.

chipbuster (approved):
Thanks for your contribution! This is looking ready to merge to me.

PR #1755, opened 2020-10-08
perf(init): Use random for session-key instead of invoking starship session

Gautham:
#### Description
We currently invoke `starship session` in a sub-shell to generate the session-key, which is non-performant. By using $RANDOM instead, we can avoid this and generate the session-key quicker. This commit provides this optimization for ZSH (tested on 5.8 and 4.3) and bash-3.2.

#### Motivation and Context
Improves the performance of shell startup.

```
### With `starship session`
~ ❯ benchmark_zsh
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      69.0 ms ±   4.6 ms    [User: 33.7 ms, System: 27.2 ms]
  Range (min … max):    65.1 ms …  88.7 ms    50 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

### With `$RANDOM`
~ ❯ benchmark_zsh
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      60.2 ms ±   2.3 ms    [User: 30.5 ms, System: 23.4 ms]
  Range (min … max):    57.6 ms …  69.2 ms    50 runs

### With starship not enabled.
~ ❯ benchmark_zsh
Benchmark #1: zsh -i -c exit;
  Time (mean ± σ):      58.4 ms ±   3.1 ms    [User: 29.7 ms, System: 22.2 ms]
  Range (min … max):    56.2 ms …  75.5 ms    50 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
```

#### How Has This Been Tested?
Verified that the snippet in question works on zsh-5.8, zsh-4.3 and bash-3.2. Also verified that the prompt works correctly using zsh-5.8

- [x] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #1904, opened 2020-11-17
perf(init): Use random for session-key in fish

Glandos:
#### Description
Same as #1755 but for fish users

#### Motivation and Context
Improves the performance of shell startup

With `starship session`:
```
❯ hyperfine -r500 "fish -i -c exit"
Benchmark #1: fish -i -c exit
  Time (mean ± σ):      13.3 ms ±   3.9 ms    [User: 9.8 ms, System: 3.8 ms]
  Range (min … max):     5.5 ms …  20.4 ms    500 runs
 
```

With `random`:
```
❯ hyperfine -r500 "fish -i -c exit"
Benchmark #1: fish -i -c exit
  Time (mean ± σ):      10.2 ms ±   2.9 ms    [User: 7.7 ms, System: 2.7 ms]
  Range (min … max):     4.1 ms …  15.6 ms    500 runs
 
  Warning: Command took less than 5 ms to complete. Results might be inaccurate.
```

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -18,4 +18,4 @@ function fish_mode_prompt; end
 export STARSHIP_SHELL="fish"
 
 # Set up the session key that will be used to store logs
-export STARSHIP_SESSION_KEY=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)
+export STARSHIP_SESSION_KEY=(random 10000000000000 9999999999999999)
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -18,4 +18,4 @@ function fish_mode_prompt; end
 export STARSHIP_SHELL="fish"
 
 # Set up the session key that will be used to store logs
-export STARSHIP_SESSION_KEY=(::STARSHIP:: session)
+export STARSHIP_SESSION_KEY=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)


src/init/starship.fish
```diff
@@ -18,4 +18,4 @@ function fish_mode_prompt; end
 export STARSHIP_SHELL="fish"
 
 # Set up the session key that will be used to store logs
-export STARSHIP_SESSION_KEY=(::STARSHIP:: session)
+export STARSHIP_SESSION_KEY=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)
```
davidkna, line 21:
```suggestion
export STARSHIP_SESSION_KEY=(random 1000 9999)(random 1000 9999)(random 1000 9999)(random 1000 9999)'
```
How about this to get rid of the need to cut the string?

davidkna (approved):
It seems to work fine, but the `random` approach is barely faster for me. My solution without `string sub` appears to be a bit faster:
| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `fish -c 'export X=(starship session)'` | 6.2 ± 1.1 | 4.8 | 10.7 | 1.06 ± 0.26 |
| `fish -c 'export X=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)'` | 6.0 ± 0.8 | 4.9 | 9.1 | 1.03 ± 0.22 |
| `fish -c 'export X=(random 1000 9999)(random 1000 9999)(random 1000 9999)(random 1000 9999)'` | 5.8 ± 1.0 | 4.6 | 9.7 | 1.00 |

matchai:
If the performance payoff is within a margin of 1-2ms, I'm not sure it's worth complicating the setup scripts further.

src/init/starship.fish
Glandos:
```suggestion
export STARSHIP_SESSION_KEY=(random 10000000000000 9999999999999999)
```

Let's go one step further. Since it's not meant to be secure, use a 16-digit number, and we're done.
I first copy/paste-and-adapt the code from Bash, but fish can do better, so let's do it. And it's lot easier to read.

Glandos:
You're right, it's not the goal of performance, but more about simplicity. So I updated the code.

I also like the idea of avoiding forking subshell for simple tasks.

src/init/starship.fish
davidkna: Is that going to work on <64 Bit systems? 

New benchmarks:
| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `fish -c 'export X=(starship session)'` | 17.4 ± 9.2 | 5.6 | 33.9 | 1.00 |
| `fish -c 'export X=(random 10000000000000 9999999999999999)'` | 18.4 ± 6.9 | 3.9 | 30.1 | 1.06 ± 0.68 |

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `fish -c 'export X=(starship session)'` | 6.5 ± 0.4 | 5.9 | 9.4 | 1.35 ± 0.15 |
| `fish -c 'export X=(random 10000000000000 9999999999999999)'` | 4.8 ± 0.4 | 4.1 | 5.9 | 1.00 |

davidkna:
I also think given the mixed speed improvement and that `starshp session` generates better session keys that the old `starship session` is preferable.
I still think it might be worth to do some mac and windows (does fish support windows?) benchmarks because iirc they start binaries more slowly than linux.

src/init/starship.fish
Glandos: Yes it should. Fish [parses arguments as `long long`](https://github.com/fish-shell/fish-shell/blob/master/src/builtin_random.cpp#L63), and standard C++ says that it is [at least 64 bits](https://en.cppreference.com/w/cpp/language/types#Properties)

rouge8:
I ran the benchmarks above on macOS 10.15.7 on a 2020 Intel MacBook Pro  and the speed improvement is more significant here -- 55% faster!:

```console
❯ hyperfine "fish -c 'export X=(starship session)'" "fish -c 'export X=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)'" "fish -c 'export X=(random 1000 9999)(random 1000 9999)(random 1000 9999)(random 1000 9999)'"
Benchmark #1: fish -c 'export X=(starship session)'
  Time (mean ± σ):      18.8 ms ±   0.7 ms    [User: 11.5 ms, System: 5.9 ms]
  Range (min … max):    16.8 ms …  20.4 ms    139 runs
 
Benchmark #2: fish -c 'export X=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)'
  Time (mean ± σ):      12.2 ms ±   0.6 ms    [User: 8.2 ms, System: 3.6 ms]
  Range (min … max):    11.2 ms …  17.1 ms    171 runs
 
Benchmark #3: fish -c 'export X=(random 1000 9999)(random 1000 9999)(random 1000 9999)(random 1000 9999)'
  Time (mean ± σ):      12.3 ms ±   0.7 ms    [User: 8.3 ms, System: 3.6 ms]
  Range (min … max):    11.0 ms …  16.0 ms    198 runs
 
Summary
  'fish -c 'export X=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)'' ran
    1.01 ± 0.08 times faster than 'fish -c 'export X=(random 1000 9999)(random 1000 9999)(random 1000 9999)(random 1000 9999)''
    1.55 ± 0.10 times faster than 'fish -c 'export X=(starship session)''
```

matchai:
Running the same test, I am seeing similar results on a 2018 Macbook Pro.
(FWIW, setting `$HOME` to a tempdir stops fish from resolving to your personal config in `~/.config/fish`)

```
❯ env HOME=/tmp hyperfine "fish -c 'export X=(starship session)'" "fish -c 'export X=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)'" "fish -c 'export X=(random 1000 9999)(random 1000 9999)(random 1000 9999)(random 1000 9999)'"
Benchmark #1: fish -c 'export X=(starship session)'
  Time (mean ± σ):      14.5 ms ±   0.6 ms    [User: 7.6 ms, System: 5.5 ms]
  Range (min … max):    13.7 ms …  17.0 ms    139 runs

Benchmark #2: fish -c 'export X=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)'
  Time (mean ± σ):       8.4 ms ±   0.5 ms    [User: 4.9 ms, System: 3.2 ms]
  Range (min … max):     7.5 ms …  10.4 ms    232 runs

Benchmark #3: fish -c 'export X=(random 1000 9999)(random 1000 9999)(random 1000 9999)(random 1000 9999)'
  Time (mean ± σ):       7.9 ms ±   0.5 ms    [User: 4.7 ms, System: 2.9 ms]
  Range (min … max):     7.0 ms …  11.0 ms    248 runs

Summary
  'fish -c 'export X=(random 1000 9999)(random 1000 9999)(random 1000 9999)(random 1000 9999)'' ran
    1.06 ± 0.09 times faster than 'fish -c 'export X=(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)''
    1.83 ± 0.14 times faster than 'fish -c 'export X=(starship session)''
```

andytom:
@Glandos thank you for your contribution.

PR #1910, opened 2020-11-20
feat(git_branch): add 'only_attached' config bool

### LINKED ISSUE #1905: Option to disable git_branch when in detached HEAD state
The git_commit module has an option (`only_detached`) to disable it if the repository isn't currently in a detached-HEAD state. This makes sense: if you're on a branch, that information is more likely to be useful than the commit sha1.

But the git_branch module displays no matter what. When you're in a detached-HEAD state, it always prints `HEAD`, which is not particularly useful information.

The git_branch module should have a corresponding option (maybe `only_attached`?) which disabled output if the repo is in a detached HEAD state.

This can pretty easily be accomplished via a custom command, but having this symmetry between git_branch and git_commit built into starship directly seems nice.

aarongable:
#### Description
This adds a new boolean, `only_attached`, to the configuration for
the `git_branch` module. This mirrors the `only_detached` config
value for the `git_commit` module: setting the value to true causes
the module to suppress its output when the user is not on a branch.
For the sake of backwards compatibility, this new config value
defaults to false.

#### Motivation and Context
This allows users to have either a branch name or a commit hash in
their prompt, as opposed to having either a branch name or the
overly-wordy "HEAD (sha1)", using a config such as:

```toml
[git_branch]
  symbol = ""
  only_attached = true
[git_commit]
  format = "(at [$hash]($style) )"
  only_detached = true
```

Closes #1905

Adds two new tests to the `git_branch` module, which confirm that
the output of the module is unchanged when the config value is set
to true and the repository has a branch checked out, and that the
output of the module is suppressed when the config value is set
to true and the repository is in a detached HEAD state.

I have also tested manually by replacing my local starship binary
with one produced from this version of the code.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1030,6 +1030,7 @@ The `git_branch` module shows the active branch of the repo in your current dire
 | `style`             | `"bold purple"`                  | The style for the module.                                                                |
 | `truncation_length` | `2^63 - 1`                       | Truncates a git branch to X graphemes.                                                   |
 | `truncation_symbol` | `"…"`                            | The symbol used to indicate a branch name was truncated. You can use `""` for no symbol. |
+| `only_attached`      | `false`                                         | Only show the branch name when not in a detached HEAD state. |
 | `disabled`          | `false`                          | Disables the `git_branch` module.                                                        |
 
 ### Variables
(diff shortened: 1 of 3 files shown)


andytom (approved):

andytom:
Thank you for your contribution @aarongable.

PR #1919, opened 2020-11-22
fix(git_commit): show last created tag on current commit

miguelfito:
### Description
If option to show tag is enabled, it now shows last created tag on current commit.

#### Motivation and Context
Fix / improve behavior of `git_commit` module. It makes sense to show latest created tag pointing to current commit instead of oldest one. 

#### Checklist:

### DIFF
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -339,4 +339,71 @@ mod tests {
         assert_eq!(expected, actual);
         Ok(())
     }
+
+    #[test]
+    fn test_latest_tag_shown_with_tag_enabled() -> io::Result<()> {
+        let repo_dir = fixture_repo(FixtureProvider::GIT)?;
+
+        let mut git_commit = Command::new("git")
+            .args(&["rev-parse", "HEAD"])
+            .current_dir(&repo_dir.path())
+            .output()?
+            .stdout;
+        git_commit.truncate(7);
+        let commit_output = str::from_utf8(&git_commit).unwrap().trim();
+
+        Command::new("git")
+            .args(&["tag", "v1", "-m", "Testing tags v1"])
+            .current_dir(&repo_dir.path())
+            .output()?;
+
+        Command::new("git")
+            .args(&["tag", "v0", "-m", "Testing tags v0", "HEAD~1"])
+            .current_dir(&repo_dir.path())
+            .output()?;
+
+        Command::new("git")
+            .args(&["tag", "v2", "-m", "Testing tags v2"])
+            .current_dir(&repo_dir.path())
+            .output()?;
+
+        let git_tag = Command::new("git")
+            .args(&[
+                "for-each-ref",
+                "--contains",
+                "HEAD",
+                "--sort=-taggerdate",
+                "--count=1",
+                "--format",
+                "%(refname:short)",
+                "refs/tags",
+            ])
+            .current_dir(&repo_dir.path())
+            .output()?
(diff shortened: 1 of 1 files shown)


chipbuster (commented):
Is there a way we can add a test for this change? Maybe do a test where we create a commit + multiple tags?

miguelfito:
> Is there a way we can add a test for this change? Maybe do a test where we create a commit + multiple tags?

Done, @chipbuster! Thanks for the suggestion! 😉

src/modules/git_commit.rs
```diff
@@ -335,4 +339,76 @@ mod tests {
...
+        git_commit.truncate(7);
+        let commit_output = str::from_utf8(&git_commit).unwrap().trim();
+
+        Command::new("git")
+            .args(&["tag", "v2", "-m", "Testing tags v2"])
+            .current_dir(&repo_dir.path())
+            .output()?;
+
+        // Wait one second between tags
+        thread::sleep(time::Duration::from_millis(1000));
```
chipbuster, line 363: I'm a little bit suspicious of this line because it seems to rely on wallclock properties (e.g. if a daylight-savings or multiple leap-second timechange occur while the sleep is happening, does the test still work?) but I don't see a better way of testing it at the moment since you can't artifically specify creation times in git.

Nothing to be done for the moment, just something to remember if we see this test start failing in GH Actions for no apparent reason.

src/modules/git_commit.rs
```diff
@@ -335,4 +339,76 @@ mod tests {
...
+            .args(&["tag", "v0", "-m", "Testing tags v0", "HEAD~1"])
+            .current_dir(&repo_dir.path())
+            .output()?;
+
+        Command::new("git")
+            .args(&["tag", "v1", "-m", "Testing tags v1"])
+            .current_dir(&repo_dir.path())
+            .output()?;
+
+        let git_tag = Command::new("git")
```
chipbuster, line 375: Just to confirm, the output format of this shouldn't change with newer git versions right? Git generally doesn't consider the output text of commands to be an API contract, but it looks to me like there's enough specifiers in here that there's not really a way git could change the output without rendering the flags broken?

chipbuster (commented):
Just one question to answer and we should be ready to merge! (Probably no changes needed, I just needed to make sure of something)

src/modules/git_commit.rs
miguelfito: Totally agree with you! It's a little bit _dirty_ solution for generate two tags in a row and then make sure starship is showing the latest one, but I didn't found another way of testing it...

I hope this _trick_ won't broke test, but honestly, I don't know how GitHub Actions will react in those all scenarios... 😓

src/modules/git_commit.rs
miguelfito: Yes, this output shouldn't change with newer git releases. If so, I think we should find another way to get the latest tag on a commit...

chipbuster: Okay, sounds good!

src/modules/git_commit.rs
chipbuster: Yeah, I don't think we really have a choice but to hope that this keeps working (it definitely works for now, since the GH Actions build is passing). Just making a note here in case it breaks in a future release so that we have an easier time tracking it down.

chipbuster (approved):
Based on our latest set of conversations, I think this is good to merge :+1:

PR #1938, opened 2020-11-28
feat(rust): Support new rust-toolchain format

### LINKED ISSUE #1935: Support the new format for `rust-toolchain`

Rustup 1.23.0 was released and accounced today.

The announcement contains [a new format for `rust-toolchain`](https://blog.rust-lang.org/2020/11/27/Rustup-1.23.0.html#new-format-for-rust-toolchain).

```toml
[toolchain]
channel = "nightly-2020-07-10"
components = ["rustfmt", "clippy"]
targets = ["wasm32-unknown-unknown"]
```

Of cource the current `rust` module does not work well with the new format while the target toolchain is not installed.

```console
foo on  master [?] is 📦 v0.1.0 via ⚙ [toolchain]
❯
```

Here is change of the parser in Rustup since we implemented #426.

https://github.com/rust-lang/rustup.rs/blob/d84e6e50126bccd84649e42482fc35a11d019401/src/config.rs#L320-L358

↓

https://github.com/rust-lang/rustup/blob/master/src/config.rs#L546-L616

Support the new format.

https://github.com/qryxip/starship/commit/7795993d397fc189678c953559b98e089b24bdf2

```diff
-foo on  master [?] is 📦 v0.1.0 via ⚙ [toolchain]
+foo on  master [?] is 📦 v0.1.0 via ⚙ nightly-2020-07-10-x86_64-unknown-linux-gnu
 ❯
```

Find `'^channel\s*=\s*"(.*)"'` instead of using `toml::from_str` when parsing a `rust-toolchain`?

dnaka91:
#### Description

This changes the extraction of the currently used Rust version from the `rust-toolchain` file to support the [new format](https://blog.rust-lang.org/2020/11/27/Rustup-1.23.0.html#new-format-for-rust-toolchain) as well. It reads the first line as before but checks for `[toolchain]` in the resulting string to detect the new TOML format. In that case it parses the whole file as TOML file and extracts the version from the `toolchain.channel` value. In case the first line is empty the TOML parsing is triggered as well, as a TOML file can contain potential empty lines before the `[toolchain]` section.

I added test cases for both versions as well.

Old format (still valid):
```txt
1.34.0
```

New format:
```toml
[toolchain]
channel = "1.34.0"
```

#### Motivation and Context

A new version of rustup was released yesterday, introducing an optional new TOML format for the `rust-toolchain` file that allows to provide additional _components_ and _targets_ that rustup will auto-install if missing.

The current implementation in starship simply reads the first line to extract the currently used Rust version. This works fine with the old format but creates an empty version output (in my case) with the new format.

Closes #1935

#### Checklist:

#### Considerations & ideas

Currently the extraction will fail if there are some other sections in the TOML version before the `[toolchain]` section as it will fail to trigger the TOML parsing. There is in my opinion no reason to put anything else in there as it has no use to `rustup` but it is technically possible or people maybe like to put a comment at the top of the file.

Therefore, it may be a good choice to try TOML parsing first and then fall back to reading the first line. I left the current order intact as there are probably not many TOML versions of `rust-toolchain` around yet and it is simply faster than trying to parse the TOML format first.

### DIFF
--- a/src/modules/rust.rs
+++ b/src/modules/rust.rs
@@ -144,7 +144,7 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
 
         match contents.lines().count() {
             0 => None,
-            1 => Some(contents.trim().to_owned()),
+            1 => Some(contents),
             _ => Some(
                 toml::from_str::<OverrideFile>(&contents)
                     .ok()?
@@ -152,6 +152,8 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
                     .channel?,
             ),
         }
+        .filter(|c| !c.trim().is_empty())
+        .map(|c| c.trim().to_owned())
     }
 
     if let Ok(true) = context
--- a/src/modules/rust.rs
+++ b/src/modules/rust.rs
@@ -125,24 +125,25 @@ fn extract_toolchain_from_rustup_override_list(stdout: &str, cwd: &Path) -> Opti
 
 fn find_rust_toolchain_file(context: &Context) -> Option<String> {
     // Look for 'rust-toolchain' as rustup does.
-    // https://github.com/rust-lang/rustup.rs/blob/d84e6e50126bccd84649e42482fc35a11d019401/src/config.rs#L320-L358
+    // https://github.com/rust-lang/rustup/blob/89912c4cf51645b9c152ab7380fd07574fec43a3/src/config.rs#L546-L616
 
-    fn read_target(path: &Path) -> Option<String> {
-        let content = fs::read_to_string(path).ok()?;
-        let line = content.lines().next()?.trim();
+    fn read_channel(path: &Path) -> Option<String> {
+        let contents = fs::read_to_string(path).ok()?;
 
-        if line.is_empty() || line == "[toolchain]" {
-            let value = toml::from_str::<toml::Value>(&content).ok()?;
-            let target = value
-                .as_table()?
-                .get("toolchain")?
-                .as_table()?
-                .get("channel")?
-                .as_str()?;
-
-            Some(target.trim().to_owned())
-        } else {
-            Some(line.to_owned())
+        match contents.lines().count() {
+            0 => None,
+            1 => Some(contents.trim().to_owned()),
+            _ => Some(
+                toml::from_str::<toml::Value>(&contents)
+                    .ok()?
+                    .as_table()?
+                    .get("toolchain")?
+                    .as_table()?
+                    .get("channel")?
+                    .as_str()?
+                    .trim()
+                    .to_owned(),
+            ),
         }
     }
 
@@ -150,14 +151,14 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
         .dir_contents()
         .map(|dir| dir.has_file("rust-toolchain"))
     {
-        if let Some(toolchain) = read_target(Path::new("rust-toolchain")) {
+        if let Some(toolchain) = read_channel(Path::new("rust-toolchain")) {
             return Some(toolchain);
         }
     }
 
     let mut dir = &*context.current_dir;
     loop {
-        if let Some(toolchain) = read_target(&dir.join("rust-toolchain")) {
+        if let Some(toolchain) = read_channel(&dir.join("rust-toolchain")) {
             return Some(toolchain);
         }
         dir = dir.parent()?;
--- a/src/modules/rust.rs
+++ b/src/modules/rust.rs
@@ -2,6 +2,8 @@ use std::fs;
 use std::path::Path;
 use std::process::{Command, Output};
 
+use serde::Deserialize;
+
 use super::{Context, Module, RootModuleConfig};
 
 use crate::configs::rust::RustConfig;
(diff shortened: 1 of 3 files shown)


src/modules/rust.rs
```diff
@@ -127,24 +127,37 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
...
     // https://github.com/rust-lang/rustup.rs/blob/d84e6e50126bccd84649e42482fc35a11d019401/src/config.rs#L320-L358
 
-    fn read_first_line(path: &Path) -> Option<String> {
+    fn read_target(path: &Path) -> Option<String> {
         let content = fs::read_to_string(path).ok()?;
-        let line = content.lines().next()?;
-        Some(line.trim().to_owned())
+        let line = content.lines().next()?.trim();
+
+        if line.is_empty() || line == "[toolchain]" {
```
davidkna, line 134: I checked how rustup parses the file and they treat all files with no lines as invalid, a single line as the old format and anything longer as TOML ([source](https://github.com/rust-lang/rustup/blob/89912c4cf51645b9c152ab7380fd07574fec43a3/src/config.rs#L591-L616)). Would you mind changing the code to match rustup behaviour?

src/modules/rust.rs
```diff
@@ -127,24 +127,37 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
     // Look for 'rust-toolchain' as rustup does.
     // https://github.com/rust-lang/rustup.rs/blob/d84e6e50126bccd84649e42482fc35a11d019401/src/config.rs#L320-L358
```
davidkna, line 128: Please update this comment to match a commit with the newer behaviour.

src/modules/rust.rs
```diff
@@ -127,24 +127,37 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
...
+    fn read_target(path: &Path) -> Option<String> {
         let content = fs::read_to_string(path).ok()?;
-        let line = content.lines().next()?;
-        Some(line.trim().to_owned())
+        let line = content.lines().next()?.trim();
+
+        if line.is_empty() || line == "[toolchain]" {
+            let value = toml::from_str::<toml::Value>(&content).ok()?;
+            let target = value
+                .as_table()?
```
davidkna, line 137: I think it might be better to define two `struct`s here that `#[derive(Deserialize)]`, but this is fine too.

davidkna (changes requested):

src/modules/rust.rs
dnaka91: Oh I see, I'll update it to stay as close as possible to rustup's implementation :+1:

src/modules/rust.rs
dnaka91: Missed that, will update it.

src/modules/rust.rs
dnaka91: I wanted to keep it very simple, but in general I like to have extra structs with derives deserializers for anything involving serde as well. Just wasn't sure what is preferred in starship.

I'll add separate structs for this :+1:

dnaka91: I just checked and saw that starship is not using `serde`, so adding it might impact compilation time.

What would you prefer @davidkna?
- Use structs with `#[derive(Deserialize)]` but have to add `serde` as dependency.
- Keep the current version.

davidkna: `serde_derive` is a dependency. Isn't that crate enough to `#[derive(Deserialize)]`? Anyway `toml` already uses `serde` internally and starship also depends on `serde_json` so I don't see any issues with adding `serde` to the dependency list.

If you do need to add a direct dependecy to `serde` maybe remove the `serde_derive` dependency in exchange because it looks like `serde` has superseeded `serde_derive`.

src/modules/rust.rs
```diff
@@ -125,26 +125,40 @@ fn extract_toolchain_from_rustup_override_list(stdout: &str, cwd: &Path) -> Opti
...
-        let line = content.lines().next()?;
-        Some(line.trim().to_owned())
+    // https://github.com/rust-lang/rustup/blob/89912c4cf51645b9c152ab7380fd07574fec43a3/src/config.rs#L546-L616
+
+    fn read_channel(path: &Path) -> Option<String> {
+        let contents = fs::read_to_string(path).ok()?;
+
+        match contents.lines().count() {
+            0 => None,
+            1 => Some(contents.trim().to_owned()),
```
davidkna, line 135: This still needs the `is_empty` check.
```suggestion
            1 => Some(contents.trim()).filter(|s| !s.is_empty()).map(|s| s.to_owned()),
```

src/modules/rust.rs
dnaka91: You're right, I checked with `cargo tree -i -p serde` and there are several other crates that already include `serde` as dependency.

I just updated the PR using the exact same structs as rustup.

src/modules/rust.rs
dnaka91: I think this would apply to the TOML version as well, as there could be just an empty string like `channel = ""`.
So I will apply the logic to both of them.

src/modules/rust.rs
```diff
@@ -125,26 +127,47 @@ fn extract_toolchain_from_rustup_override_list(stdout: &str, cwd: &Path) -> Opti
...
+
+        match contents.lines().count() {
+            0 => None,
+            1 => Some(contents),
+            _ => Some(
+                toml::from_str::<OverrideFile>(&contents)
+                    .ok()?
+                    .toolchain
+                    .channel?,
+            ),
```
davidkna, line 153: Nitpick:
```suggestion
            _ => toml::from_str::<OverrideFile>(&contents)
                    .ok()?
                    .toolchain
                    .channel,
```

davidkna (approved):

PR #1947, opened 2020-11-30
feat(python): Smarter python binary usage

### LINKED ISSUE #1670: Switch default python python_binary to python3

[Python2 is EOL](https://www.python.org/doc/sunset-python-2/) and most linux distros have decided to not switch /usr/bin/python to a python3 version. That means, at least on some newer linux systems, `python` is gone and you only have `python3`. 

See e.g 
* debian: https://www.debian.org/doc/packaging-manuals/python-policy/ch-python.html#s-interpreter_loc -> `The python binary package must also ensure that /usr/bin/python is provided, as a symlink to the current python2.Y executable.` -> not a py3 one!
* redhat: https://developers.redhat.com/blog/2019/05/07/what-no-python-in-red-hat-enterprise-linux-8/ -> not completely clear if it has /usr/bin/python out of the box or if it needs explicitly installed
* Ubuntu: https://wiki.ubuntu.com/Python/3 -> ` /usr/bin/python and /usr/bin/python2 will point to Python 2.7 and /usr/bin/python3 will point to the latest supported Python 3 version.`
* The python.org installer for unix like systems (which include mac) seem to not link python to a py3 version: https://www.python.org/dev/peps/pep-0394/#application-to-the-cpython-reference-interpreter

It seems windows installer still distribute `python.exe`

Unfortunately, the default command to look for in the `python`module is `python`and not `python3` ([default: `python_binary: "python"`](https://github.com/starship/starship/blob/master/src/configs/python.rs#L20). 
 
Change the default to python3 on linux + mac

Search for both python3 and python.

andytom:
#### Description

Have updated the python module to try multiple python binaries when
determining the version. With the new logic if starship doesn't find
`python` on the `PATH`, which is the default for some Linux Distros, it
will fallback to `python3` and then `python2`.

#### Motivation and Context

Closes #1670 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2009,16 +2009,25 @@ The module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Option               | Default                                                                 | Description                                                                   |
-| -------------------- | ----------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
-| `format`             | `'via [${symbol}${pyenv_prefix}${version}( \($virtualenv\))]($style) '` | The format for the module.                                                    |
-| `symbol`             | `"🐍 "`                                                                 | A format string representing the symbol of Python                             |
-| `style`              | `"yellow bold"`                                                         | The style for the module.                                                     |
-| `pyenv_version_name` | `false`                                                                 | Use pyenv to get Python version                                               |
-| `pyenv_prefix`       | `pyenv `                                                                | Prefix before pyenv version display, only used if pyenv is used               |
-| `scan_for_pyfiles`   | `true`                                                                  | If false, Python files in the current directory will not show this module.    |
-| `python_binary`      | `python`                                                                | Configures the python binary that Starship executes when getting the version. |
-| `disabled`           | `false`                                                                 | Disables the `python` module.                                                 |
+| Option               | Default                                                                 | Description                                                                            |
+| -------------------- | ----------------------------------------------------------------------- | -------------------------------------------------------------------------------------- |
+| `format`             | `'via [${symbol}${pyenv_prefix}${version}( \($virtualenv\))]($style) '` | The format for the module.                                                             |
+| `symbol`             | `"🐍 "`                                                                 | A format string representing the symbol of Python                                      |
+| `style`              | `"yellow bold"`                                                         | The style for the module.                                                              |
+| `pyenv_version_name` | `false`                                                                 | Use pyenv to get Python version                                                        |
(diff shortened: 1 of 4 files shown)


docs/config/README.md
```diff
@@ -2009,16 +2009,25 @@ The module will be shown if any of the following conditions are met:
...
+| `symbol`             | `"🐍 "`                                                                 | A format string representing the symbol of Python                                      |
+| `style`              | `"yellow bold"`                                                         | The style for the module.                                                              |
+| `pyenv_version_name` | `false`                                                                 | Use pyenv to get Python version                                                        |
+| `pyenv_prefix`       | `pyenv `                                                                | Prefix before pyenv version display, only used if pyenv is used                        |
+| `scan_for_pyfiles`   | `true`                                                                  | If false, Python files in the current directory will not show this module.             |
+| `python_binary`      | `["python", "python3, "python2"]`                                       | Configures the python binaries that Starship should executes when getting the version. |
```
jankatins, line 2020: It would be nice to have a small sentence why this default order was chosen: 

```
::: tip

The default values and order for `python_binary` was chosen to first identify the Python version in a virtualenv/conda environments (which currently still add a `python`, no matter if it points to `python3` or `python2`). This has the side effect that if you have still a system Python verison 2 installed, it will be picked up before any Python version 3 (at least on Linux, as Distros do not setup a `/usr/bin/python` symlink for Python 3). If you do not work with Python 2 anymore but cannot remove the system Python version 2, changing this to `["python3"]` will hide any Python version 2. 
```

jankatins (approved):
LGTM

docs/config/README.md
andytom: Good suggestion, I've expanded the new tip to include this.

docs/config/README.md
```diff
@@ -2009,16 +2009,34 @@ The module will be shown if any of the following conditions are met:
...
+
+The `python_binary` variable accepts either a string or a list of strings.
+Starship will try executing each binary until it gets a result. Note you can
+only change the binary that Starship executes to get the version of Python not
+the arguments that are used.
+
+The default values and order for `python_binary` was chosen to first identify
+the Python version in a virtualenv/conda environments (which currently still
+add a `python`, no matter if it points to `python3` or `python2`). This has the
+side effect that if you still have a system Python verison 2 installed, it may
```
jankatins, line 2033: `verison -> version` 

(I fixed a few problematic things in my comment above, e.g. `Python version 2` -> `Python 2` and a better (...) for the symlink on linux. Sorry that I did this after the initial "add comment" :-()

PR #1948, opened 2020-11-30
feat(status): Convert known status code to their meaning

### LINKED ISSUE #1821: status module should show if the process was killed by a signal or other reason communicated by exit code

Not really.

In unix, if a signal kills the process then it will have an exit code > 128.  From the [Bash reference manual](https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html):

> For the shell’s purposes, a command which exits with a zero exit status has succeeded. A non-zero exit status indicates failure. This seemingly counter-intuitive scheme is used so there is one well-defined way to indicate success and a variety of ways to indicate various failure modes. When a command terminates on a fatal signal whose number is N, Bash uses the value 128+N as the exit status.
>
> If a command is not found, the child process created to execute it returns a status of 127. If a command is found but is not executable, the return status is 126.

The Linux Documentation project has [a more technical document](https://tldp.org/LDP/abs/html/exitcodes.html) explaining it from the kernel's point of view.

And just for fun, here is [a collection of quasi-standard exit codes](https://krueger.report/programming/exit-status-codes/).

I would ask that:

- A different symbol and color could be used for terminated-by-signal and 126 and 127.
- That signals could be configured to be one of: exit-code, signal (e.g., `-9`), short name (e.g., `INT`), or maybe even long name (e.g., `interrupt`).

Can't think of any.  Starship would have to do this.

See also: This would also apply to #1669

izissise:
#### Description
User is able to choose if their want to display the meaning of
known status code in place of number.

Inspired by  go-powerline
https://github.com/justjanne/powerline-go/blob/master/segment-exitcode.go#L9

#### Motivation and Context
Having a number doesn't give a lot of information, status number of above 128 give the corresponding signal number of why a program exited 
Closes #1821

There is one unit tests, I successfully use it on my prompt on linux

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2211,20 +2211,14 @@ To enable it, set `disabled` to `false` in your configuration file.
 | Option                    | Default                     | Description                                            |
 | ------------------------- | --------------------------- | ------------------------------------------------------ |
 | `format`                  | `[$symbol$status]($style) ` | The format of the module                               |
-| `program_error_symbol`    | `"🔴"`                       | The symbol displayed on program error                  |
+| `symbol`                  | `"✖"`                       | The symbol displayed on program error                  |
 | `not_executable_symbol`   | `"🚫"`                       | The symbol displayed when file isn't executable        |
 | `not_found_symbol`        | `"🔍"`                       | The symbol displayed when the command can't be found   |
 | `sigint_symbol`           | `"🧱"`                       | The symbol displayed on SIGINT (Ctrl + c)              |
 | `signal_symbol`           | `"⚡"`                       | The symbol displayed on any signal |
 | `style`                   | `"bold red"`                | The style for the module.                              |
+| `map_symbol`              | `false`                     | Enable symbols mapping from exit code                  |
 | `disabled`                | `true`                      | Disables the `status` module.                          |
-
-
-                    "status" => Some(Ok(exit_code)),
-                    "status_int" => Some(Ok(exit_code)),
-                    "status_common_meaning" => Ok(status_common_meaning(exit_code_int)).transpose(),
-                    "status_signal_number" => Ok(status_signal_int(exit_code_int)).transpose(),
-                    "status_signal_name" => Ok(status_signal_name(exit_code_int)).transpose(),
 
 ### Variables
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2242,8 +2242,9 @@ To enable it, set `disabled` to `false` in your configuration file.
 
 [status]
 style = "bg:blue"
-symbol = "💣 "
-format = '[\[$symbol$status\]]($style) '
+symbol = "🔴"
+format = '[\[$symbol $status_common_meaning$status_signal_name\]]($style) '
+map_symbol = true
 disabled = false
 
 ```
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2229,6 +2229,7 @@ To enable it, set `disabled` to `false` in your configuration file.
 | status_common_meaning   | `ERROR` | Meaning of the code if not a signal                                     |
 | status_signal_number    | `9`     | Signal number corresponding to the exit code, only if signalled         |
 | status_signal_name      | `KILL`  | Name of the signal corresponding to the exit code, only if signalled    |
+| status_maybe_int        | `7`     | Contains the exit code number when no meaning has been found            |
 | symbol                  |         | Mirrors the value of option `symbol`                                    |
 | style\*                 |         | Mirrors the value of option `style`                                     |
 
@@ -2243,7 +2244,7 @@ To enable it, set `disabled` to `false` in your configuration file.
 [status]
 style = "bg:blue"
 symbol = "🔴"
-format = '[\[$symbol $status_common_meaning$status_signal_name\]]($style) '
+format = '[\[$symbol $status_common_meaning$status_signal_name$status_maybe_int\]]($style) '
 map_symbol = true
 disabled = false
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2217,6 +2217,7 @@ To enable it, set `disabled` to `false` in your configuration file.
 | `sigint_symbol`           | `"🧱"`                       | The symbol displayed on SIGINT (Ctrl + c)              |
 | `signal_symbol`           | `"⚡"`                       | The symbol displayed on any signal |
 | `style`                   | `"bold red"`                | The style for the module.                              |
+| `recognize_signal_code`   | `true`                      | Enable signal mapping from exit code                   |
 | `map_symbol`              | `false`                     | Enable symbols mapping from exit code                  |
 | `disabled`                | `true`                      | Disables the `status` module.                          |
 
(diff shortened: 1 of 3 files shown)


chipbuster:
So possibly a dumb question, but what happens if this is running under powershell on Windows? Or what if it's running in fish--does fish also use the `128+N` convention used by bash?

izissise:
Actually bash doesn't modify the exit code for non builtin commands, if the program is terminated by a signal kernel set the exit code to be 128 + signal number (signal table here https://man7.org/linux/man-pages/man7/signal.7.html)

More links on signal and exit numbers:
- wait manpage https://linux.die.net/man/2/waitpid
- kernel wait macro definition https://kernel.googlesource.com/pub/scm/linux/kernel/git/nico/archive/+/v0.01/include/sys/wait.h

For other codes below 129 they might be specific to bash, I'm not sure more testing is needed on other shell, we need the ouput of the following (or equivalent):
```
For bash
$ /dev/null; echo $?
126
$ cmd_unknown; echo $?
127
```
I believe MacOs respect the same convention, I don't known about Windows

Bash actually a fine doc on exit codes https://tldp.org/LDP/abs/html/exitcodes.html

izissise:
After reading the associated issue, I implemented some of the ideas there.

chipbuster:
Okay, just so that I understand the overall changes here, this basically adds new format string variables to the status module that correspond to (if applicable) exit signal names, numbers, and meanings, but doesn't affect existing format strings?

My primary concern is that we cannot force-enable any of these because POSIX only specifies that processes terminated by a signal should return `>=128`, but does not specify what that number should be (not to mention we also support non-POSIX shells like fish and pwsh). So it must be possible (with configuration) to retain the old output, which only checks for zero and nonzero output.

I think the format variables currently do that, but is there any way to disable the symbols you've put in for non-executable or file-not-found?

izissise:
I'll add a flag to disable modified symbols.

src/modules/status.rs
```diff
@@ -52,6 +71,60 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     }
 }
 
+fn status_common_meaning(ex: u32) -> Option<String> {
```
davidkna, line 74: I'd prefer if the below functions all used `&str` instead of `String` and always returned `None` if no match is found instead of either `None` or and empty string.

src/modules/status.rs
```diff
@@ -36,6 +51,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 })
                 .map(|variable| match variable {
                     "status" => Some(Ok(exit_code)),
+                    "status_int" => Some(Ok(exit_code)),
```
davidkna, line 54: What's the differnce between `status_int` and `status` supposed to be?

mickimnet:
> So possibly a dumb question, but what happens if this is running under powershell on Windows? Or what if it's running in fish--does fish also use the `128+N` convention used by bash?

I did try to find out, if _Fish Shell_ aheres to the `128+N`—didn‘t found anything specific—but I looks like it does. On my newly setup MacBook Pro with the default _Fish Shell_ prompt, I do also get  `SIGINT` displayed in the prompt, if I CTRL-c a command.

davidkna:
I think the 128+N exit code for signals might be a POSIX thing. But windows doesn't appear to follow this convention.

src/modules/status.rs
izissise: `status` exist for backward compatibility

src/modules/status.rs
izissise: Done I introduced a new variable for the case where no meaning is found

src/modules/status.rs
```diff
@@ -3,6 +3,9 @@ use super::{Context, Module, RootModuleConfig};
 use crate::configs::status::StatusConfig;
 use crate::formatter::StringFormatter;
 
+type ExitCode = u32;
```
davidkna, line 6: Exit codes can be negative on Windows.
```suggestion
type ExitCode = i64;
```
or
```suggestion
type ExitCode = i32;
```

docs/config/README.md
```diff
@@ -2208,20 +2208,30 @@ To enable it, set `disabled` to `false` in your configuration file.
...
-| Variable | Example | Description                          |
-| -------- | ------- | ------------------------------------ |
-| status   | `127`   | The exit code of the last command    |
-| symbol   |         | Mirrors the value of option `symbol` |
-| style\*  |         | Mirrors the value of option `style`  |
+| Variable                | Example | Description                                                             |
+| ----------------------- | ------- | ----------------------------------------------------------------------- |
+| status                  | `127`   | The exit code of the last command                                       |
+| status_int              | `127`   | The exit code of the last command                                       |
+| status_common_meaning   | `ERROR` | Meaning of the code if not a signal                                     |
```
davidkna, line 2229: I think `status_common_meaning` and `status_signal_name` should be combined (they are mutually exclusive anyway). Please also add an option to disable recognizing signal exit codes. Chances are high those are false positives on windows and it should help alleviate the need for `status_signal_name` .

davidkna (changes requested):

docs/config/README.md
izissise: Indeed, there are mutually exclusive, but one might want to display text for common meanings and signal numbers for signals, combining both would prevent this.

I'll add an option to ignore signal matching

davidkna (changes requested):

docs/config/README.md
```diff
@@ -2208,20 +2208,31 @@ To enable it, set `disabled` to `false` in your configuration file.
...
+| status_signal_number    | `9`     | Signal number corresponding to the exit code, only if signalled         |
+| status_signal_name      | `KILL`  | Name of the signal corresponding to the exit code, only if signalled    |
+| status_maybe_int        | `7`     | Contains the exit code number when no meaning has been found            |
```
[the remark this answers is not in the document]
izissise: Indeed, it's done

src/modules/status.rs
```diff
@@ -24,10 +27,47 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
                 .map_meta(|var, _| match var {
-                    "symbol" => Some(config.symbol),
+                    "symbol" => match exit_code_int {
+                        126 if config.map_symbol => Some(config.not_executable_symbol),
+                        127 if config.map_symbol => Some(config.not_found_symbol),
+                        130 if config.recognize_signal_code && config.map_symbol => {
+                            Some(config.sigint_symbol)
+                        }
+                        x if x > 128
+                            && x < 256
```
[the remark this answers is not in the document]
izissise: Done

src/modules/status.rs
```diff
@@ -24,10 +27,47 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+        let common_meaning = status_common_meaning(exit_code_int);
+
+        let raw_signal_number = match config.recognize_signal_code {
+            true => status_to_signal(exit_code_int),
+            false => None,
+        };
+        let signal_number = raw_signal_number.map(|sn| sn.to_string());
+        let signal_number_name = raw_signal_number
+            .map(|sn| status_signal_name(sn).or_else(|| signal_number.as_deref()))
+            .flatten();
```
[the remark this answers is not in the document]
izissise: No more  flatten :)

davidkna (approved):
LGTM

chipbuster:
@izissise Thanks for the contribution!
@davidkna Thank you for all the review effort on the PR!

PR #1963, opened 2020-12-05
Disable AWS by default

vladinator1000:
#### Description
Disables the AWS config

#### Motivation and Context
Not everybody uses AWS so it makes sense to me to have this be an opt-in feature.

Ran the tests on windows, there were only 2 failures that look unrelated
```
---- modules::directory::tests::root_directory stdout ----
thread 'modules::directory::tests::root_directory' panicked at 'assertion failed: `(left == right)`
  left: `Some("\u{1b}[1;36m/\u{1b}[0m ")`,
 right: `Some("\u{1b}[1;36m/\u{1b}[0m\u{1b}[31m🔒\u{1b}[0m ")`', src\modules\directory.rs:721:9

---- modules::custom::tests::command_returns_right_string stdout ----
thread 'modules::custom::tests::command_returns_right_string' panicked at 'assertion failed: `(left == right)`
  left: `Some("?????\r\n")`,
 right: `Some("강남스타일\r\n")`', src\modules\custom.rs:284:9
```

#### Checklist:

- [ ] I have updated the tests accordingly (no tests for that config).

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -254,7 +254,7 @@ is read from the `AWS_VAULT` env var.
 | `symbol`         | `"☁️ "`                                        | The symbol used before displaying the current AWS profile.      |
 | `region_aliases` |                                                | Table of region aliases to display in addition to the AWS name. |
 | `style`          | `"bold yellow"`                                | The style for the module.                                       |
-| `disabled`       | `false`                                        | Disables the `AWS` module.                                      |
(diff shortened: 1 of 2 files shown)


andytom:
Hi @vladinator1000, the AWS module should only be triggered for users who use AWS, e.g they have the `AWS_PROFILE` env var set or they have an `~/.aws/config` file is this not the case?

vladinator1000:
Ahhh, I totally forgot about that. Makes sense 👍

PR #1964, opened 2020-12-06
feat(username): Try harder to guess if inside ssh

yangsheng6810:
#### Description
In addition to `SSH_CONNECTION`, also try other common environment variables set in an ssh connection: `SSH_CLIENT`, `SSH_TTY`.

#### Motivation and Context
Not all servers set `SSH_CONNECTION` in an ssh connection, therefore testing for the existence of `SSH_CONNECTION` is not reliable enough. In fact, testing for `SSH_CLIENT` and `SSH_TTY` is still not good enough, but should be better. Other means (e.g. [link](https://unix.stackexchange.com/questions/9605/how-can-i-detect-if-the-shell-is-controlled-from-ssh)) are beyond my capability.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2399,9 +2399,11 @@ The module will be shown if any of the following conditions are met:
 - The user is currently connected as an SSH session
 - The variable `show_always` is set to true
 
-Note: SSH connection is detected by checking environment variables
+::: tip
+SSH connection is detected by checking environment variables
 `SSH_CONNECTION`, `SSH_CLIENT`, and `SSH_TTY`. If your SSH host does not set up
 these variables, one workaround is to set one of them with a dummy value.
+:::
 
 ### Options
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2399,6 +2399,10 @@ The module will be shown if any of the following conditions are met:
 - The user is currently connected as an SSH session
 - The variable `show_always` is set to true
 
+Note: SSH connection is detected by checking environment variables
+`SSH_CONNECTION`, `SSH_CLIENT`, and `SSH_TTY`. If your SSH host does not set up
+these variables, one workaround is to set one of them with a dummy value.
+
 ### Options
 
 | Option        | Default                 | Description                           |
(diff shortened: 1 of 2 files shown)


src/modules/username.rs
```diff
@@ -129,6 +136,18 @@ mod tests {
...
+        let actual = ModuleRenderer::new("username")
+            .env("USER", "astronaut")
+            .env("SSH_TTY", "/dev/pts/0")
+            .collect();
+        let expected = Some(format!("{} in ", Color::Yellow.bold().paint("astronaut")));
+
+        assert_eq!(expected, actual);
+        Ok(())
+    }
+
```
andytom, line 150: Can you add a test for the `SSH_CLIENT` env var?

andytom (commented):
Hi @yangsheng6810, thank you for your contribution, I've just left a little suggestion but other than that this LGTM.

src/modules/username.rs
yangsheng6810: Test added. I also included some documentation on the behavior, and a workaround on servers that do not set these environment variables correctly.

docs/config/README.md
```diff
@@ -2399,6 +2399,10 @@ The module will be shown if any of the following conditions are met:
 - The user is currently connected as an SSH session
 - The variable `show_always` is set to true
 
+Note: SSH connection is detected by checking environment variables
+`SSH_CONNECTION`, `SSH_CLIENT`, and `SSH_TTY`. If your SSH host does not set up
+these variables, one workaround is to set one of them with a dummy value.
+
```
andytom, line 2405: You can make this stand out a bit more by marking this as a tip (hopefully I've got the syntax right here). 

```suggestion
:::: tip
SSH connection is detected by checking environment variables
`SSH_CONNECTION`, `SSH_CLIENT`, and `SSH_TTY`. If your SSH host does not set up
these variables, one workaround is to set one of them with a dummy value.
::::

```

andytom (approved):
LGTM, have left an additional suggestion but it's not a blocker.

andytom:
@yangsheng6810 thank you for your contribution

PR #1965, opened 2020-12-06
docs(conda): Fix default format in Conda module's config doc

kesyog:
#### Description
Correct the config docs to add "via" to the start of the Conda module's default `format` string such that it matches the string's actual default value.

#### Motivation and Context
 The default `format` string for the Conda module starts with "via", as seen [here](https://github.com/starship/starship/blob/2e763b5e7c41dd556c9515bdc36297bc177d7f46/src/configs/conda.rs#L19).  However, this is not reflected in the config docs.

Many similar Starship modules include "via" in their default format string so I assume including it is the intended behavior, and that the doc should be updated, rather than updating the actual default value to match the docs.

n/a

Viewed the changes in Github's rich diff viewer

#### Checklist:


andytom (changes requested):
Hi @kesyog, thank you for your contribution. When updating the docs you only need to update the English docs since all our translations are handled by Crowdin (https://translate.starship.rs/) and if we modify the files outside of Crowdin it causes problems, would it be possible for you to revert the changes all other language files, other than that this LGTM.

kesyog:
No problem, I've gone ahead and reverted changes to all the non-English files

andytom (approved):

andytom:
Thank you for  your contribution @kesyog

PR #1992, opened 2020-12-11
refactor: a few low-hanging optimizations

alexmaco:
#### Description
A few low hanging optimization to avoid allocating `Vec`s that are always the same, and make openstack config reading lazy and more robust.

#### Checklist:

### DIFF
--- a/src/modules/git_branch.rs
+++ b/src/modules/git_branch.rs
@@ -49,11 +49,14 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     }
 
     // Truncate fields if need be
-    for e in vec![
+    for e in [
         &mut graphemes,
         &mut remote_branch_graphemes,
         &mut remote_name_graphemes,
-    ] {
+    ]
+    .iter_mut()
+    {
+        let e = &mut **e;
         let trunc_len = len.min(e.len());
         if trunc_len < e.len() {
             // The truncation symbol should only be added if we truncate
--- a/src/modules/openstack.rs
+++ b/src/modules/openstack.rs
@@ -9,23 +9,29 @@ use crate::utils;
 type Cloud = String;
 type Project = String;
 
-fn get_osp_project_from_config(context: &Context, osp_cloud: Option<&str>) -> Option<Project> {
+fn get_osp_project_from_config(context: &Context, osp_cloud: &str) -> Option<Project> {
     // Attempt to follow OpenStack standards for clouds.yaml location:
     // 1st = $PWD/clouds.yaml, 2nd = $HOME/.config/openstack/clouds.yaml, 3rd = /etc/openstack/clouds.yaml
-    let config = vec![
-        utils::read_file(context.get_env("PWD").unwrap() + "/clouds.yaml"),
-        utils::read_file(dirs_next::home_dir()?.join(".config/openstack/clouds.yaml")),
-        utils::read_file("/etc/openstack/clouds.yaml"),
+    let config = [
+        context.get_env("PWD").map(|pwd| pwd + "/clouds.yaml"),
+        dirs_next::home_dir().map(|home| {
+            home.join(".config/openstack/clouds.yaml")
+                .display()
+                .to_string()
+        }),
+        Some(String::from("/etc/openstack/clouds.yaml")),
     ];
-    let clouds =
-        YamlLoader::load_from_str(&config.into_iter().find_map(Result::ok).unwrap()).ok()?;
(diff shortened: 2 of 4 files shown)


src/modules/git_branch.rs
```diff
@@ -49,11 +49,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     }
 
     // Truncate fields if need be
-    for e in vec![
+    for e in &mut [
```
davidkna, line 52: How about just removing the `vec!` and adding an `iter_mut()`?

src/modules/openstack.rs
```diff
@@ -9,23 +9,32 @@ use crate::utils;
...
 
-fn get_osp_project_from_config(context: &Context, osp_cloud: Option<&str>) -> Option<Project> {
+fn get_osp_project_from_config(context: &Context, osp_cloud: &str) -> Option<Project> {
     // Attempt to follow OpenStack standards for clouds.yaml location:
     // 1st = $PWD/clouds.yaml, 2nd = $HOME/.config/openstack/clouds.yaml, 3rd = /etc/openstack/clouds.yaml
-    let config = vec![
-        utils::read_file(context.get_env("PWD").unwrap() + "/clouds.yaml"),
-        utils::read_file(dirs_next::home_dir()?.join(".config/openstack/clouds.yaml")),
-        utils::read_file("/etc/openstack/clouds.yaml"),
+    let files: &[Box<dyn Fn(&Context) -> Option<String>>] = &[
```
davidkna, line 15: I'd prefer something simpler, more similiar to the old solution like this here. I am not sure how zero-cost closures are:
```rust
    let config = [
        utils::read_file(context.get_env("PWD").unwrap() + "/clouds.yaml"),
        utils::read_file(dirs_next::home_dir()?.join(".config/openstack/clouds.yaml")),
        utils::read_file("/etc/openstack/clouds.yaml"),
    ];
    let clouds =
        YamlLoader::load_from_str(&config.iter().find_map(|i| i.as_ref().ok()).unwrap()).ok()?;
```
If you want to avoid reading all the files maybe just put the filenames into the array instead of closures. I feel like closures would overcomplicate this function:
```rust
    let config = [
        context.get_env("PWD").unwrap() + "/clouds.yaml",
        dirs_next::home_dir()?.join(".config/openstack/clouds.yaml"),
        "/etc/openstack/clouds.yaml",
    ];
````
But I like the way you simplified the function futher down.

File: src/modules/username.rs [hunk 3]
```diff
@@ -56,10 +56,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn is_ssh_connection(context: &Context) -> bool {
-    let ssh_env: Vec<&str> = vec!["SSH_CONNECTION", "SSH_CLIENT", "SSH_TTY"];
-    ssh_env
-        .into_iter()
-        .any(|env| context.get_env(env).is_some())
+    static SSH_ENV: &[&str] = &["SSH_CONNECTION", "SSH_CLIENT", "SSH_TTY"];
+    SSH_ENV.iter().any(|env| context.get_env(env).is_some())
```
davidkna, line 60: I somewhat prefer this, but your solution is good too.
```suggestion
    let ssh_env = ["SSH_CONNECTION", "SSH_CLIENT", "SSH_TTY"];
    ssh_env
        .iter()
        .any(|env| context.get_env(env).is_some())
```

davidkna (changes requested):

src/modules/openstack.rs
alexmaco: Indeed, I could avoid the closures and just use plain function pointers instead, which avoids one more indirection.
However, I think the strings shouldn't be constructed up front, since some of the operations can fail, like `home_dir` returning `None`. In this case the function should not return, but instead try the other files it can, in order.

davidkna: If you want to avoid early exits you could just use `Option`, although none of these are very likely to fail:
```rust
    let config = [
        context.get_env("PWD").map(|pwd| pwd + "/clouds.yaml"),
        dirs_next::home_dir().map(|home| home.join(".config/openstack/clouds.yaml")),
        Some("/etc/openstack/clouds.yaml"),
    ];
```

But your solution is decent too I suppose.

davidkna (approved):
LGTM, but I don't really know openstack, so somebody else should maybe have a look at this before it gets merged.

andytom:
Thank you for your contribution @alexmaco and thank you @davidkna for reviewing.

PR #1993, opened 2020-12-11
fix(config): log as error if failure to read config wasn't caused by NotFound

davidkna:
#### Description

Until now all log messages related to failure to read config file contents as a string were not visible by default. The case of a missing config file is innocuous but I think other errors like invalid unicode are worth showing to the user.

This PR increases the log level from `debug` to `error` if the source of the error is not `ErrorKind::NotFound`.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,6 +5,7 @@ use indexmap::IndexMap;
 
 use std::clone::Clone;
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


andytom (approved):

andytom:
Great Idea, thanks for the fix @davidkna.

PR #1995, opened 2020-12-12
feat(shlvl): Add `repeat` option

### LINKED ISSUE #1878: Repeat string option for SHLVL
Not really a problem, just an extension of an existing feature

My current prompt written in fish has a great feature for showing `$SHLVL` where instead of showing a boring old number, it actually makes my prompt bigger as the level goes up. It's a great visual indicator of how "deep" I am in shells.

Examples:
```
user@hostname:~
~> fish
user@hostname:~
~~> fish
user@hostname:~
~~~> fish
user@hostname:~
~~~~> 
```

I'd love for something like this to be possible in starship. The way I implemented it in fish script is by using `printf` to format a string with a variable number of spaces equal to `$SHLVL`, then replacing the spaces with the "fish tail" `~`:
```fish
printf '%*s>' $SHLVL '' | tr ' ' '~'
```

The implementation of this in Rust should be pretty straightforward, the only thing I'm wondering about is how to provide configuration for it. Maybe you could define a "repeat" option that's a string that gets multiplied by the current `$SHLVL`?

I could probably implement this if we think it's a good idea.

Not having this is fine, I just like it and think it's a better visual representation of `$SHLVL` than a number.

dbeckwith:
#### Description

Adds an option to the `shlvl` module which causes the `symbol` string to be repeated by the current SHLVL. This enables a nice a visual indicator for SHLVL that in my opinion is more intuitive than just displaying a number.

#### Motivation and Context

Closes #1878

##### Alternative Implementation

A different approach to adding this, instead of repeating the existing `symbol` string, would be to make the `repeat` option itself a string which is repeated, and you have to include that in your format for it to show up. If messing with `symbol` is something that's generally discouraged in Starship then this would be a good alternative.

#### Checklist:

### DIFF
--- a/src/modules/shlvl.rs
+++ b/src/modules/shlvl.rs
@@ -22,8 +22,13 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let shlvl_str = &shlvl.to_string();
 
-    let symbol = if config.repeat {
-        Cow::Owned(config.symbol.repeat(shlvl.try_into().unwrap_or(usize::MAX)))
+    let repeat_count = if config.repeat {
+        shlvl.try_into().unwrap_or(1)
+    } else {
+        1
+    };
+    let symbol = if repeat_count != 1 {
+        Cow::Owned(config.symbol.repeat(repeat_count))
     } else {
         Cow::Borrowed(config.symbol)
     };
--- a/src/modules/shlvl.rs
+++ b/src/modules/shlvl.rs
@@ -3,6 +3,8 @@ use super::{Context, Module};
 use crate::config::RootModuleConfig;
 use crate::configs::shlvl::ShLvlConfig;
 use crate::formatter::StringFormatter;
+use std::borrow::Cow;
+use std::convert::TryInto;
 
 const SHLVL_ENV_VAR: &str = "SHLVL";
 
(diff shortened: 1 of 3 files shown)


src/modules/shlvl.rs
```diff
@@ -20,10 +22,21 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    let repeat_count = if config.repeat {
+        shlvl.try_into().unwrap_or(1)
+    } else {
+        1
+    };
+    let symbol = if repeat_count != 1 {
+        Cow::Owned(config.symbol.repeat(repeat_count))
+    } else {
+        Cow::Borrowed(config.symbol)
+    };
```
andytom, line 34: I might be missing something but could this not be
```suggestion
    let symbol = config.symbol.repeat(repeat_count);
```

andytom (commented):
Have left a suggestion/question but in general LGTM.

src/modules/shlvl.rs
dbeckwith: I figured it would be a slight optimization to not have to allocate and copy the symbol string in the common case where `repeat_count` is 1 by using `Cow::Borrowed`.

andytom (approved):

PR #2002, opened 2020-12-17
refactor: specify all dependencies up to the patch/x.y.z level

davidkna:
#### Description

This PR replaces all dependency versions that were specified at the `x.y` (minor level) to the `x.y.z` (patch level).

#### Motivation and Context

~This will cause [deps.rs](https://deps.rs/) not to flag  starship as insecure on the front page.~

Additionaly this will allow dependabot to send PRs for patch level updates for all direct dependencies. These more up to dependency packages could cause issues for out-of-date crates.io mirror users, but likely not much more than is already the case.

#### Checklist:


chipbuster:
Just to be clear, I think the deps.rs issue stems from a dependency range that's not under our direct control (we would need yaml-rust to move its minimum `linked-hash-map` dependency to `>=0.2.1`, and given that that crate is in maintenance mode, I don't know if they'd be willing to do it).

davidkna:
It looks like the version requirement for `linked-hash-map` in the ` yaml-rust ` repo was upped 5 months ago, but there hasn't been a release sine then. There's also a pr there to replace the `linked-hash-map` crate altogether. It looks like it's troublesome in general.

chipbuster (approved):
Looks fine to me. If we start getting too many dependabot PRs we might need to revert, but I highly doubt that will be an issue.

PR #2008, opened 2020-12-18
docs: Fix typo in README

adrian5:


matchai (approved):
Good catch! Thank you for the fix! 😄

PR #2014, opened 2020-12-20
fix(user): Fix username detection on Windows

### LINKED ISSUE #1350: Powershell: Username is not shown
#### Current Behavior
Username is not printed despite of ```show_always = true``` for [username]

#### Expected Behavior
Username should be shown

image

#### Environment
- Starship version: 0.42.0
- powershell version: 7.0.2

```bash
none
```

#### Starship Configuration

```toml
[username]
show_always = true
```

### LINKED ISSUE #1554: username module: Not hiding username on Termux
#### Current Behavior
I'm using starship + fish shell on [Termux](https://termux.com/), but the `username` module isn't hiding my username.

This happens because the `username` module ensures that both `$LOGNAME` and `$USER` match, otherwise it assumes we're on a remote session. But the problem is that `$LOGNAME` is _not_ set on Termux when I'm on a local session. Here's the offending code:

https://github.com/starship/starship/blob/edc62f45188b742a4d20b601170ba8b9bdb5c803/src/modules/username.rs#L15-L16
https://github.com/starship/starship/blob/edc62f45188b742a4d20b601170ba8b9bdb5c803/src/modules/username.rs#L25

#### Expected Behavior
Username to be hidden when I'm on a local shell.

Screenshot_20200803-095818

#### Possible Solution
PR #1157 apparently fixes this by falling back to `$USER` when `$LOGNAME` is not set (I haven't _actually_ tested it, though):

https://github.com/starship/starship/pull/1157/files#diff-41209ab340992e5aa3ca03ae5516f372R30-R33

A local workaround I found, is to manually set `$LOGNAME` as an environment variable in fish (`~/.config/fish/config.fish`):

```bash
set -gx LOGNAME $USER
```

And that didn't seem to break anything else, which is good. :)

#### Environment
- Starship version: 0.44.0
- Operating system: Android 9
- Terminal emulator: Termux 0.95

```bash
starship init fish | source
```

#### Starship Configuration

```toml
add_newline = false

[character]
symbol = "➜"

[directory]
truncate_to_repo = false
```

### LINKED ISSUE #1579: Username and Hostname modules don't work in Powershell
#### Current Behavior
In Powershell on Windows 10 Username and Hostname modules don't display anything

#### Expected Behavior
Username module shows the username and Hostname module shows the hostname

image

#### Possible Solution
* Usename module uses $USER to read the username, whereas in Powershell it's $Env:USERNAME
* Hostname uses gethostname() which should produce the right result in Windows, but for some reason it's not

#### Environment
```
~ on ☁️  ap-southeast-2 starship --version
starship 0.44.0
```

- Shell type: [fish, zsh]
```
Powershell
```

```
~ on ☁️  ap-southeast-2 $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.0.3
PSEdition                      Core
GitCommitId                    7.0.3
OS                             Microsoft Windows 10.0.19042
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

- Shell plugin manager: [e.g. oh-my-fish, oh-my-zsh, fisher, antigen]
none

- Terminal emulator: [e.g. iTerm, Hyper, Terminator]
image

- Operating system: [e.g. macOS 10.13.4, Windows 10]
Microsoft Windows 10.0.19042

```
~ on ☁️  ap-southeast-2 gc $PROFILE
# Replace more with less
Set-Alias more less
Set-Alias edit notepad++

# Set environment variables
$Env:PAGER='less'
$Env:EDITOR='notepad++'

# Set-PSReadlineOption -EditMode Emacs
Set-PSReadlineKeyHandler -Key ctrl+d -Function ViExit

# Import-Module posh-git
# Import-Module oh-my-posh
# Set-Theme MySorin
# $ThemeSettings.Options.ConsoleTitle = $false

# $GitPromptSettings.DefaultPromptPrefix.Text = '$Env:COMPUTERNAME '
# $GitPromptSettings.DefaultPromptPrefix.ForegroundColor = [ConsoleColor]::Magenta

$Env:AWS_DEFAULT_REGION='ap-southeast-2'
```
#### Starship Configuration
```
~ on ☁️  ap-southeast-2 gc C:\Users\alexv\.config\starship.toml
add_newline = true
prompt_order = [
    "username", "hostname", "directory",
    "git_branch", "git_commit", "git_state", "git_status",
    "nodejs", "python",
    "aws"
]

[username]
show_always = true
```

andytom:
#### Description

This switches Windows to use the env var `USERNAME` rather than `USER`
for getting the username of the current user. I have also done some minor
refactoring to simplify some of the code and make the checks to see if
the user is root or another user a bit more robust.

#### Motivation and Context

Closes #1554
Closes #1579
Closes #1350
Closes #1157

**Note** - I have not tested this on Windows the changes were made after checking the docs and looking at the related issues.

#### Checklist:

### DIFF
--- a/src/modules/username.rs
+++ b/src/modules/username.rs
@@ -4,6 +4,13 @@ use crate::configs::username::UsernameConfig;
 use crate::formatter::StringFormatter;
 use crate::utils;
 
+const ROOT_UID: Option<u32> = Some(0);
+#[cfg(not(target_os = "windows"))]
+const USERNAME_ENV_VAR: &str = "USER";
+
+#[cfg(target_os = "windows")]
+const USERNAME_ENV_VAR: &str = "USERNAME";
+
 /// Creates a module with the current user's username
 ///
 /// Will display the username if any of the following criteria are met:
@@ -11,25 +18,26 @@ use crate::utils;
 ///     - The current user is root (UID = 0)
 ///     - The user is currently connected as an SSH session (`$SSH_CONNECTION`)
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
-    let user = context.get_env("USER");
+    let username = context.get_env(USERNAME_ENV_VAR)?;
     let logname = context.get_env("LOGNAME");
 
-    const ROOT_UID: Option<u32> = Some(0);
     let user_uid = get_uid();
+
+    let is_not_login = logname.is_some() && username != logname.unwrap_or_default();
+    let is_root = user_uid == ROOT_UID;
 
     let mut module = context.new_module("username");
     let config: UsernameConfig = UsernameConfig::try_load(module.config);
 
-    if user != logname || is_ssh_connection(&context) || user_uid == ROOT_UID || config.show_always
-    {
-        let username = user?;
+    if is_not_login || is_ssh_connection(&context) || is_root || config.show_always {
         let parsed = StringFormatter::new(config.format).and_then(|formatter| {
             formatter
                 .map_style(|variable| match variable {
                     "style" => {
-                        let module_style = match user_uid {
-                            Some(0) => config.style_root,
-                            _ => config.style_user,
+                        let module_style = if is_root {
+                            config.style_root
+                        } else {
+                            config.style_user
                         };
                         Some(Ok(module_style))
                     }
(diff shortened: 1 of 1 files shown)


davidkna (approved):
I prefer the approach in #1157, but for the purpose of fixing this quickly LGTM.

src/modules/username.rs
```diff
@@ -4,32 +4,40 @@ use crate::configs::username::UsernameConfig;
...
 ///     - The user is currently connected as an SSH session (`$SSH_CONNECTION`)
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
-    let user = context.get_env("USER");
+    let username = context.get_env(USERNAME_ENV_VAR)?;
     let logname = context.get_env("LOGNAME");
 
-    const ROOT_UID: Option<u32> = Some(0);
     let user_uid = get_uid();
 
+    let is_not_login = logname.is_some() && username != logname.unwrap_or_default();
```
davidkna, line 26: Nitpick:
I'd prefer
```suggestion
        let is_not_login = logname.map(|logname| username != logname).unwrap_or(false);
```
or
```suggestion
    let is_not_login = logname.is_some() && username != logname.unwrap();
```
here

andytom:
> I prefer the approach in #1157, but for the purpose of fixing this quickly LGTM.

I think #1157 has a lot going for it but since the PR is from before the format string release I don't think it is going to be easy to merge, my goal here is just to fix the current problems and then we can look at this again to see what improvements can be made.

PR #2020, opened 2020-12-21
feat: Add support for cygwin/msys2/git-bash evironment

### LINKED ISSUE #590: Starship cannot be loaded from Fish inside Git-Bash

### LINKED ISSUE #634: Support native windows (msys2) environment

### LINKED ISSUE #1354: Can't make Starship work on a Fish (Cygwin) on Windows environment

zoritle:
#### Description

see #1652 
Refactored the init logic.
Clean commit tree.
#### Motivation and Context

fixes #1354
fixes #634
fixes #590 

#### Checklist:

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -47,7 +47,6 @@ impl StarshipPath {
             .output();
         let output = match res {
             Ok(output) => output,
-            // Failed to execute cygpath.exe means there're not inside cygwin evironment,return directly.
             Err(e) => {
                 if e.kind() != io::ErrorKind::NotFound {
                     log::warn!("Failed to convert \"{}\" to unix path:\n{:?}", str_path, e);
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -1,5 +1,5 @@
 use std::ffi::OsStr;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::{env, io};
 
 /* We use a two-phase init here: the first phase gives a simple command to the
@@ -17,12 +17,63 @@ using whatever mechanism is available in the host shell--this two-phase solution
 has been developed as a compatibility measure with `eval $(starship init X)`
 */
 
-fn path_to_starship() -> io::Result<String> {
-    let current_exe = env::current_exe()?
-        .to_str()
-        .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?
-        .to_string();
-    Ok(current_exe)
+struct StarshipPath {
+    native_path: PathBuf,
+}
+impl StarshipPath {
+    fn init() -> io::Result<Self> {
+        Ok(Self {
+            native_path: env::current_exe()?,
+        })
+    }
+    fn str_path(&self) -> io::Result<&str> {
+        let current_exe = self
+            .native_path
+            .to_str()
+            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "can't convert to str"))?;
+        Ok(current_exe)
+    }
+    fn sprint(&self) -> io::Result<String> {
+        self.str_path().map(|s| s.replace("\"", "\"'\"'\""))
+    }
+    fn sprint_posix(&self) -> io::Result<String> {
+        // On non-Windows platform, return directly.
+        if cfg!(not(target_os = "windows")) {
+            return self.sprint();
+        }
+        let str_path = self.str_path()?;
+        let res = std::process::Command::new("cygpath.exe")
+            .arg(str_path)
+            .output();
+        let output = match res {
+            Ok(output) => output,
+            // Failed to execute cygpath.exe means there're not inside cygwin evironment,return directly.
+            Err(e) => {
+                if e.kind() != io::ErrorKind::NotFound {
+                    log::warn!("Failed to convert \"{}\" to unix path:\n{:?}", str_path, e);
+                }
+                // Failed to execute cygpath.exe means there're not inside cygwin evironment,return directly.
+                return self.sprint();
+            }
+        };
+        let res = String::from_utf8(output.stdout);
+        let posix_path = match res {
+            Ok(ref cygpath_path) if output.status.success() => cygpath_path.trim(),
+            Ok(_) => {
+                log::warn!(
+                    "Failed to convert \"{}\" to unix path:\n{}",
+                    str_path,
+                    String::from_utf8_lossy(&output.stderr),
+                );
+                str_path
+            }
+            Err(e) => {
+                log::warn!("Failed to convert \"{}\" to unix path:\n{}", str_path, e);
+                str_path
+            }
+        };
+        Ok(posix_path.replace("\"", "\"'\"'\""))
+    }
 }
 
 /* This prints the setup stub, the short piece of code which sets up the main
@@ -33,7 +84,7 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
 
     let shell_basename = Path::new(shell_name).file_stem().and_then(OsStr::to_str);
 
-    let starship = path_to_starship()?.replace("\"", "\"'\"'\"");
+    let starship = StarshipPath::init()?;
 
     let setup_stub = match shell_basename {
         Some("bash") => {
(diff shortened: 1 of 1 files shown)


nesara-cmd:
0e826b5d21

src/init/mod.rs
```diff
@@ -143,32 +189,31 @@ fi"#,
...
-        "powershell" => Some(PWSH_INIT),
-        "ion" => Some(ION_INIT),
+    let starship_path = StarshipPath::init()?;
+
+    match shell_name {
+        "bash" => print_script(BASH_INIT, &starship_path.sprint_unix()?),
+        "zsh" => print_script(ZSH_INIT, &starship_path.sprint_unix()?),
+        "fish" => print_script(FISH_INIT, &starship_path.sprint_unix()?),
+        "powershell" => print_script(PWSH_INIT, &starship_path.sprint()?),
+        "ion" => print_script(ION_INIT, &starship_path.sprint()?),
```
davidkna, line 199: I think `powershell` and `ion` should both use `sprint_unix`. PowerShell has a linux version and ion doesn't even run on windows.

nesara-cmd:
https://github.com/starship/starship/commit/c5a543ec51ce705fc8add733592d027442defa1a

src/init/mod.rs
zoritle: No. `sprint` directly return the platform native path string, which is the normal case.

`sprint_unix` return converted(on Windows)/native(on other platforms)  path string, is only used for some shells(bash/zsh/fish/etc) that runs on Windows through posix compatibility layer provided by cygwin.To be more precise, I renamed it to `sprint_posix`.

These shells can only accept posix path string (`/c/Windows/`) not native (`C:\Windows`), which cause the issues.

PowerShell is natively cross-platform and can handle native path on every platform. Ion though have not yet supported Windows, we shouldn't do additional thing except they decide to support Windows via cygwin runtime.

By the way, bash have been additionally patched to handle both Windows and posix path string, that's why people used git-bash with starship without problem, but zsh/fish not. To be consistent, we don't rely on this behavior and also apply conversion to it.

davidkna: Wouldn't PowerShell running via the Cygwin compatibly layer on windows still need the posix-style path? Also ion might not officially support windows, but I expect running it with Cygwin is possible.

zoritle: Cygwin is a compatibility layer specifically designed for C programs written against the posix api.
Powershell is built on .NET platform, which is intrinsic cross platform. I don't know if there're any way or any reason to build .NET runtime against cygwin...
Rust [did not support cygwin](https://users.rust-lang.org/t/does-rust-work-in-cygwin-if-so-how-can-i-get-it-working/25735) as a target, nor need to. Porting Ion to Windows is more possible than building it against cygwin since rust is a modern programming language.
Anyhow, since Ion currently unable to run on Windows, whether patch it or not makes no sense. I think keep it intact is more intuitive and avoid unexpected behavior in the future...

nesara-cmd:
CMS100648-1)

nesara-cmd:
CMS100648-1)

davidkna:
I tried this in git bash and `starship.exe init bash` still returned the Windows path. Is this expected?

zoritle:
> 
> 
> I tried this in git bash and `starship.exe init bash` still returned the Windows path. Is this expected? Do I need to use the GNU windows toolchain?

No, it should return converted path, and which rust toolchain you use does not matter.
For troubleshooting:
1. Ensure your are building from `feat/support_cygwin` branch, not master.
2. Ensure the `starship.exe` you called was built from source, not the official release installed by other tools (for example, scoop).
3.  Ensure you can call `cygpath.exe` inside git-bash, and `which cygpath` should return `/usr/bin/cygpath`.
4. Ensure git config `core.autocrlf` is not set to `true` before checkout, otherwise may cause line ending issues in embedded init script.

davidkna:
Well seems to work now, must have done something wrong the last time.

src/init/mod.rs
```diff
@@ -17,12 +17,55 @@ using whatever mechanism is available in the host shell--this two-phase solution
...
+            // Failed to execute cygpath.exe means there're not inside cygwin evironment,return directly.
+            Err(_) => {
+                return self.sprint();
+            }
+        };
+        let res = String::from_utf8(output.stdout);
+        let posix_path = match res {
+            Ok(ref cygpath_path) if output.status.success() => cygpath_path.trim(),
+            Ok(output) => {
+                eprintln!("Warning: getting unix path failure:{}, fallback.", output);
```
davidkna, line 59: Please use `log::warn!` (or better `log::error!`), add spaces and maybe reword the message to: `Failed to convert "{}" to unix path:\n{}`.

src/init/mod.rs
```diff
@@ -17,12 +17,55 @@ using whatever mechanism is available in the host shell--this two-phase solution
...
+                str_path
+            }
+            Err(e) => {
+                eprintln!("Warning: invalid cygpath:{},fallback.", e);
```
davidkna, line 63: Same here

davidkna (changes requested):

src/init/mod.rs
```diff
@@ -17,12 +17,59 @@ using whatever mechanism is available in the host shell--this two-phase solution
...
+            }
+        };
+        let res = String::from_utf8(output.stdout);
+        let posix_path = match res {
+            Ok(ref cygpath_path) if output.status.success() => cygpath_path.trim(),
+            Ok(output) => {
+                log::warn!(
+                    "Failed to convert \"{}\" to unix path:\n{}",
+                    str_path,
+                    output
```
davidkna, line 62: cypaths error messages are on stderr.

davidkna, line 62:
```suggestion
            Ok(_) => {
                log::warn!(
                    "Failed to convert \"{}\" to unix path:\n{}",
                    str_path,
                    String::from_utf8_lossy(&output.stderr),
```

src/init/mod.rs
```diff
@@ -17,12 +17,59 @@ using whatever mechanism is available in the host shell--this two-phase solution
...
+        let str_path = self.str_path()?;
+        let res = std::process::Command::new("cygpath.exe")
+            .arg(str_path)
+            .output();
+        let output = match res {
+            Ok(output) => output,
+            // Failed to execute cygpath.exe means there're not inside cygwin evironment,return directly.
+            Err(_) => {
+                return self.sprint();
+            }
```
davidkna, line 53: Might also make sense to combine with the match below.

```suggestion
            Err(e) if e.kind() == ErrorKind::NotFound => {
                return self.sprint();
            }
```

src/init/mod.rs
```diff
@@ -17,12 +17,63 @@ using whatever mechanism is available in the host shell--this two-phase solution
...
+        if cfg!(not(target_os = "windows")) {
+            return self.sprint();
+        }
+        let str_path = self.str_path()?;
+        let res = std::process::Command::new("cygpath.exe")
+            .arg(str_path)
+            .output();
+        let output = match res {
+            Ok(output) => output,
+            // Failed to execute cygpath.exe means there're not inside cygwin evironment,return directly.
```
davidkna, line 50:
```suggestion
```

davidkna (approved):
LGTM, but please remove that duplicate comment.

Also please use regular commits for your progress. PRs are squashed when merging.

andytom:
Thank you for you contribution @zoritle and thank you for reviewing @davidkna.

PR #2025, opened 2020-12-22
feat(status): Add hex_status & text_status variables

mika-fischer:
#### Description

Add hex_status (exit status formatted as hex) and text_status (textual representation of exit code if available) to status module.

text_status is only non-empty on Windows, as UNIX does not really have predefined exit codes, using signals instead.

#### Motivation and Context
Windows exit codes are very large integers which are hard to interpret (see example below).

#### Screenshots
Before:
💥 -1073741510❯
After:
💥 0xC000013A (CONTROL_C_EXIT)❯

#### Checklist:

### DIFF
--- a/src/modules/status.rs
+++ b/src/modules/status.rs
@@ -48,6 +48,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             false => None,
         };
 
+        let hex_status = exit_code
+            .parse::<i32>()
+            .ok()
+            .map(|code| format!("0x{:X}", code));
+
         let parsed = StringFormatter::new(config.format).and_then(|formatter| {
             formatter
                 .map_meta(|var, _| match var {
@@ -78,6 +83,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                     "common_meaning" => Ok(common_meaning.as_deref()).transpose(),
                     "signal_number" => Ok(signal_number.as_deref()).transpose(),
                     "signal_name" => Ok(signal_name.as_deref()).transpose(),
+                    "hex_status" => Some(Ok(hex_status.as_ref().map_or(exit_code, String::as_str))),
+                    "text_status" => to_text_status(exit_code).map(|s| Ok(s)),
                     _ => None,
                 })
                 .parse(None)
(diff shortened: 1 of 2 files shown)


src/modules/status.rs
```diff
@@ -52,6 +59,55 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+#[cfg(windows)]
+fn to_text_status(exit_code: &str) -> &str {
+    use winapi::shared::ntdef::NTSTATUS;
+    use winapi::shared::ntstatus::*;
+
+    exit_code
+        .parse::<NTSTATUS>()
+        .ok()
+        .map_or("", |code| match code {
+            STATUS_ACCESS_VIOLATION => " (ACCESS_VIOLATION)",
```
davidkna, line 76: I'd prefer if adding the braces and space was handled in the format string.

davidkna:
It looks like this PR has quite a bit of overlap with #1948. It might be a good idea to wait on this until that's merged.

src/modules/status.rs
mika-fischer: Is there a way to do conditional formating in starship? The issue here is that in most normal cases, the text_status will be empty and I'd like to avoid the " ()" in that common case.

mika-fischer:
Ah, I missed #1948. We'll have to think about how we merge text_status and status_signal_name. Maybe the easiest would be for this patch to also use status_signal_name?

Also regarding the hex code, it seems a bit inflexible to always introduce new variables, but I did not find a way to specify the number formatting in starship. Maybe it would be nice to provide that? `${status_int:x}` or something?

src/modules/status.rs
davidkna: Format strings have [conditional formatting](https://starship.rs/config/#conditional-format-strings). Enclosing the variable in braces ensures that it'll only get printed if the varible has a value (the other braces would need to escaped though).

davidkna:
I think using `status_common_meaning` would be best for this PR.

src/modules/status.rs
mika-fischer: Great, thanks for the pointer!. Removed the formatting.

mika-fischer:
I'll rebase this once #1948 is merged.

davidkna:
You can rebase this now.

mika-fischer:
Sorry for the delay. This is a straight rebase of the old branch.

I see two issues:
- status, int, maybe_int, hex_status seems a bit redundant. But I'm not sure if there are any formatting possibilities so that for instance the hex format could be configured by the user in the format string. Otherwise I see no easy way to avoid it.
- There is some overlap between text_status and signal_name. One possibility would be to use signal_name (and maybe also common_meaning) on non-windows platforms for text_status. This way one could just use text_status on all platforms.

What do you think?

davidkna:
I think it's best to just put the `text_status` code windows-gated into `common_meaning`. I think the error codes there should also apply to windows for the most part? Using `text_status` as a meta-variable could work too.

Formatting can't deal with stuff like hex formatting yet, so all the variables are unavoidable for now.

mika-fischer:
@davidkna Sorry for neglecting this for a while...

Since the branch has diverged quite a bit I think I'll split this up into two new pull requests:
1. Introduce hex_status since that should be simple and relatively uncontroversial.
2. How to handle ntstatus exit codes on Windows.

Regarding two, I strongly disagree with using common_meaning for ntstatus exit codes on Windows since they serve completely different purposes IMO.

common_meaning is for conventions regarding exit codes controlled by the application. These can be wrong at times. For instance diff returns 1 when the files differ, hardly an "ERROR", which is the common_meaning of exit code 1. Therefore I would not want to use this in my config due to the potentialy misleading status names...

OTOH the NTSTATUS exit codes on Windows (at least the interesting ones) come from the operating system. When the application returns 0xC0000005 (STATUS_ACCESS_VIOLATION) either the application manually returned this code (very unlikely), or more likely the operating system set the code after killing the process due to the access violation. So this is not a convention, but enforced by the operating system.

In that respect the NTSTATUS exit codes are pretty similar to signals on Unix. And I'm actually pretty sure that the Unix signal -> exit code translation does not happen at all on Windows (I'll try to verify this), so the corresponding code in starship should be flagged out for Windows.

So for the NTSTATUS codes I would prefer either:
 - re-use signal_name / signal_number on Windows for this, since they're effectively unused on Windows (makes it simple for cross-platform configs but the name is a bit confusing)
 - introduce ntstatus_name / ntstatus_number (cross-platform configs can just use both signal_name & ntstatus_name since only one of them is ever non-empty).

If you have a preference, let me know.

I'll post the links to the new pull requests here and then close this one.

mika-fischer:
Linked the new pull requests, so closing this one

PR #2053, opened 2021-01-03
fix(docker_context): enable if either yml or yaml is found

benwaffle:
#### Description
If `only_with_files = true` is set for the docker_context module, then this prompt is only shown in case a `Dockerfile` or `docker-compose.yml` file is found.
This PR also enables it for `docker-compose.yaml` (notice `yml` vs `yaml`).

#### Motivation and Context
Fixes the prompt for a valid docker-compose yaml file.

1. Created a docker-compose.yaml file
2. Ran `~/path/to/dev/build/target/starship prompt docker_context`

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -710,13 +710,13 @@ The `docker_context` module shows the currently active
 
 ### Options
 
-| Option            | Default                            | Description                                                                             |
-| ----------------- | ---------------------------------- | --------------------------------------------------------------------------------------- |
-| `format`          | `"via [$symbol$context]($style) "` | The format for the module.                                                              |
-| `symbol`          | `"🐳 "`                            | The symbol used before displaying the Docker context.                                   |
-| `style`           | `"blue bold"`                      | The style for the module.                                                               |
-| `only_with_files` | `false`                            | Only show when there's a `docker-compose.yml` or `Dockerfile` in the current directory. |
-| `disabled`        | `true`                             | Disables the `docker_context` module.                                                   |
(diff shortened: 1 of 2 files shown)


andytom (commented):
LGTM, can you also update the docs in https://github.com/starship/starship/tree/master/docs/config#docker-context.

chipbuster (dismissed):
Seems like a reasonable change. Thanks for your contribution!

matchai:
Along with an addition to docs, it would be good if we had a test to cover this new code path.

benwaffle:
I'm not sure how to write the test.

chipbuster:
I was going to suggest following the existing tests for the docker_context module, but....we have no tests in there.

Why don't we have tests for the docker module? I'm *very* tired right now so it's possible I'm just forgetting where they are, but if it turns out that we don't have tests for the docker module, I'd be in favor of merging this PR as-is and then adding tests in a separate PR (or as part of another issue).

andytom:
> I was going to suggest following the existing tests for the docker_context module, but....we have no tests in there.
> 
> Why don't we have tests for the docker module? I'm _very_ tired right now so it's possible I'm just forgetting where they are, but if it turns out that we don't have tests for the docker module, I'd be in favor of merging this PR as-is and then adding tests in a separate PR (or as part of another issue).

It looks like there are no test for this module. Let's get this merged in and I'll create a separate issue for adding tests.

andytom (approved):

andytom:
Thank you for your contribution @benwaffle

PR #2054, opened 2021-01-03
ci(deploy): Fix failing Linux GNU builds

andytom:
#### Description

This fixes the failing Linux 64 bit builds that are linking to GNU libc
that are failing because cross doesn't compile openssl-sys. The fix is
simply to enable the vendored-tls feature so that we get the rust
vendored SSL rather than the system SSL. Have also done a little bit of
refactoring to the deploy script. Now rather than have a matrix of
targets have just included each target individually (this may be
something we want to revisit in the future as the number of platforms
increases). I have also removed the installation of the musl tools since
this is no longer needed when using cross for compilation.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -39,28 +39,32 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target:
-          - x86_64-unknown-linux-gnu
-          - x86_64-unknown-linux-musl
-          - i686-unknown-linux-musl
-          - x86_64-apple-darwin
-          - x86_64-pc-windows-msvc
         include:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
             name: starship-x86_64-unknown-linux-gnu.tar.gz
+            flags: "--features tls-vendored"
+
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
             name: starship-x86_64-unknown-linux-musl.tar.gz
+            flags: "--features tls-vendored"
+
           - target: i686-unknown-linux-musl
             os: ubuntu-latest
             name: starship-i686-unknown-linux-musl.tar.gz
+            flags: "--features tls-vendored"
+
           - target: x86_64-apple-darwin
             os: macOS-latest
             name: starship-x86_64-apple-darwin.tar.gz
+            flags: ""
+
           - target: x86_64-pc-windows-msvc
             os: windows-latest
             name: starship-x86_64-pc-windows-msvc.zip
+            flags: ""
+
     runs-on: ${{ matrix.os }}
     steps:
       - name: Setup | Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -39,28 +39,32 @@ jobs:
...
-          - x86_64-unknown-linux-gnu
-          - x86_64-unknown-linux-musl
-          - i686-unknown-linux-musl
-          - x86_64-apple-darwin
-          - x86_64-pc-windows-msvc
         include:
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
             name: starship-x86_64-unknown-linux-gnu.tar.gz
+            flags: "--features tls-vendored"
```
davidkna, line 46: I think every target can just use `--features tls-vendored` because openssl isn't used on windows and macos.

.github/workflows/deploy.yml
```diff
@@ -83,43 +87,30 @@ jobs:
...
-        run: sudo apt install -y musl-tools
-
       - name: Build | Build
-        if: "! contains(matrix.target, 'musl')"
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --release --target ${{ matrix.target }}
+          args: --release ${{ matrix.flags }} --target ${{ matrix.target }}
           use-cross: true
```
davidkna, line 95: Something like this might speed release builds up a bit:
```suggestion
          use-cross: contains(matrix.target, 'musl')
```

```suggestion
          use-cross: contains(matrix.target, 'linux')
```

.github/workflows/deploy.yml
andytom: Interesting I didn't know that, in that case would be better to make `tls-vendored` the default in the `Cargo.toml`?

.github/workflows/deploy.yml
andytom: I think given that we will probably have to use cross for the Apple M1 builds and other ARM builds in the future I think it make sense to leave this as it is.

.github/workflows/deploy.yml
davidkna: Not sure. It's not very relevant outside of getting distributable builds and it bloats the binary quite a bit. It would prevent issues with OpenSSL upgrades though.

.github/workflows/deploy.yml
davidkna: Cross doesn't really support cross compiling to MacOS/Windows as far as I can tell.
On my M1 Mac in can cross compile starship to x86_64 without any setup with plain cargo (but cross works too, but it doesn't look like it's doing anything). Windows should also cross-compile without extra setup to ARM64 once sys-info supports it.

But I guess it would make any feature *BSD builds easier.

.github/workflows/deploy.yml
andytom: For now I've set it so that every target used the `tls-vendored` feature. We can look into moving this into `Cargo.toml` in the future if we need to.

davidkna (approved):

PR #2056, opened 2021-01-03
build: use shadow-rs with build-time information stored in your binary

### LINKED ISSUE #2047: Add more than information with build-time ？
I use `starship -V` with output below:
```bash
➜  ~ starship -V
starship 0.47.0
```
But sometime with starship run target OS environment,it's need to report `feature` or `bug`,maybe need other information, e.g.:
* starship pkg_version
* starship build artifact (binary) git commit hash
* starship with which git branch
* starship build_time
* starship build with rust version ,also build target OS 
* ...

Are there any plans to consider [shadow-rs](https://github.com/baoyachi/shadow-rs), if necessary, I can help.
just like this:

```bash
➜  ~ starship -V
starship 0.47.0
branch:master
commit_hash:d2958725
build_time:2021-01-01 22:16:41
build_env:rustc 1.48.0 (7eac88abb 2020-11-16),stable-x86_64-apple-darwin (default)
```

baoyachi:
#### Description
fixes #2047

#### Motivation and Context

it would help a lot with bug reports

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,8 +479,6 @@ dependencies = [
  "libc",
  "libgit2-sys",
  "log",
- "openssl-probe",
- "openssl-sys",
  "url",
 ]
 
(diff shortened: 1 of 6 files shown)


matchai:
I think we would greatly benefit by having these changes also included in the generated output of `starship bug-report`.
Would you mind updating the `Starship version:` under "Environment" to include build details?

chipbuster:
Tacking on, I think it's appropriate in this case to have both short and long version flags, since sometimes you really do only want the short version number. Would you mind making it so that the short version is the old string while the long version is the long info? I believe you can do this with the [`long_version` and `version_short` methods on `Clap::App`](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.long_version)

baoyachi:
detail link move to:[https://github.com/starship/starship/pull/2124](https://github.com/starship/starship/pull/2124)
issue:[https://github.com/starship/starship/issues/2047](https://github.com/starship/starship/issues/2047)

@marblenix @chipbuster ch

PR #2067, opened 2021-01-04
docs: Update package default format to replace /via/is

skvale:
https://github.com/starship/starship/blob/master/src/configs/package.rs#L17

#### Motivation and Context

Closes #

#### Checklist:


matchai (approved):
Thank you for the quick fix, @skvale! 😊

PR #551, opened 2019-10-16
fix: Disable venv custom prompt when starship is in use

matiaskotlik:
#### Description
Set the `VIRTUAL_ENV_DISABLE_PROMPT` variable in `starship.fish`, venv status is already shown in the prompt via the python module.

#### Motivation and Context

Closes #549 

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Screenshots (if appropriate):
DeepinScreenshot_select-area_20191016173222

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #2090, opened 2021-01-06
fix(pwsh): disable builtin venv prompt

davidkna:
#### Description

To prevent duplicate venv entries in the prompt, disable the builtin venv prompt. The `python` module can serve as the replacement.

#### Motivation and Context

Closes #2087

Context #551

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -63,6 +63,9 @@ function global:prompt {
     }
 }
 
+# Disable virtualenv prompt, it breaks starship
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM. Do we need to apply this to any other shells?

davidkna:
the other shells seem fine. I don't think ion supports any of the venv activation scripts.

andytom:
Thanks for the quick fix @davidkna

PR #2091, opened 2021-01-07
fix: bash-preexec detection and implementation

chipbuster:
#### Description
Change detection and implementation for bash-preexec: now checks for a preexec variable in case bash-preexec is installed at a system level (and so necessarily sourced before anything the user does).

#### Motivation and Context

Fixes an issue found in discussion on #2039 

#### Checklist:


andytom (approved):
LGTM

PR #2095, opened 2021-01-07
fix(directory): Remove re-implementation of "current_dir" resolution logic

deadalusai:
#### Description
This PR removes the `get_current_dir` function from the directory module and instead relies on the `current_dir` value already written to the Context. I've ported the "powershell" prefix cleanup code into `Context::new_with_dir`, and updated the unit tests.

#### Motivation and Context
I discovered this issue while testing `starship` with the [Nu shell](https://github.com/nushell/nushell). It's a WIP shell, so still has some bugs. I switched to passing `--path` to work around one such issue and found that `starship` was inconsistent with how it handled the argument.

This screenshot demonstrates the problem:
starship_current_dir_resolution
Specifically, invoking the command `starship prompt --path C:\Windows`:
- When called with no `PWD` environment variable, prints `C:/Windows {lock symbol}` ✔️
- When called with a `PWD` environment variable, prints `{the PWD env variable} {lock symbol}` ❌

The issue is caused by the `directory` module function `get_current_dir` which has a subtly different implementation of the same routine in `Context::new`. The difference in behavior leads it to ignore the `--path` argument (sort of - the "is_readonly_dir" test _was_ being run against the `context.current_dir` value).

The difference is in the priority order of the three sources checked for the "current directory":
- `context::Context::new`
    - Try `--path` argument
    - Or try `PWD` environment variable
    - Or try `env::current_dir` routine
- `directory::get_current_dir`
    - Check `use_logical_path` config
        - If true, try `PWD` environment variable
        - If false, try `env::current_dir` routine
    - Or try `--path` argument

The same command as above, but with the fix applied:
image

#### Checklist:


deadalusai:
Repro of the issue on Linux:
image

And after the fix is applied:
image

davidkna:
The code doesn't really make this clear, but I think the difference was intentional.
`--path` is supposed to override the logical path (the logical physcial path?) starship does operations in and `$PWD` is supposed to used to set the logical path, but only for display purposes.  I believe PowerShell has more special paths beyond `Microsoft.PowerShell.Core\FileSystem::` that make paths unaccesible to the starship binary.
#1209 was supposed to improve this, but it never happened.

deadalusai:
> `--path` is supposed to override the logical path (the logical physcial path?) starship does operations in and `$PWD` is supposed to used to set the logical path, but only for display purposes.

My question is, why would you want to override the logical path, but not for display purposes? `starship` seems like it is interested in the path specifically _for_ display purposes.

Re: the PowerShell stuff, I have tried to maintain the existing behavior as it was written.

davidkna:
ok I looked at the code and gathered my thoughts. That path of the code can be somewhat confusing:
`--path` sets the path starship operates in. PowerShell on linux and macos need that to work properly and it's used by the tests.

`PWD` sets the path the `directory` module uses for display if `use_logical_path` is enabled. This can have a different value than `--path` mainly on powershell. For more uses (logical drives) and details see #1114.

deadalusai:
So, if I understand correctly - given the code in `init/starship.ps1`, `PWD` and `--path` are bound approximately thus:

```
$env:PWD = $PWD
starship prompt --path (Convert-Path -LiteralPath $PWD)
```

E.g. if I `cd` into a "Registry" drive, then the values are bound as:
```
PWD = "HKEY_LOCAL_MACHINE\SOFTWARE"
--path = "HKLM:\SOFTWARE"
```

Most modules render based on the path `HLKM:\SOFTWARE` and presumably don't do much.
The Directory module however renders based on the path `HKEY_LOCAL_MACHINE\SOFTWARE`.



I think I can maintain this behavior and also address the issue I set out to fix. @davidkna what do you think about the following approach (in pseudocode):

```python
directory::module(context, config):
    let display_path = resolve_display_path(context, config)
    # note using same path for readonly check
    return "{display_path} {is_readonly_dir(display_path)}"
    
resolve_display_path(context, config):
    if config.use_logical_path and context.env.PWD:
        return context.env.PWD
    else:
        return context.current_dir
```

This tries to resolve two issues that I can see:
1. Avoid re-inventing the wheel re: resolving the physical path by falling back to `current_dir` if `use_logical_path` is not set.
2. Ensure that the "readonly" flag is set based on the display directory, rather than the `current_dir`.

The main issue with this approach is that it relies on `PWD` being in sync with `--path` (the effect is that the path argument may appear to be ignored if they are not). This isn't really a new issue though, as I guess #1209 was attempting to address this limitation.

A better solution might be to crib from #1209 by introducing a `logical_dir` (display_dir?) argument and doing the following:
```
resolve_display_path(context, config):
    if context.logical_dir:
        return context.logical_dir
    else:
        return context.current_dir
```

This has some benefits - namely two different shells, one with "virtual path" features and one with issues keeping `PWD` in sync can co-exist on the same machine, using the same starship configuration:
    - Any shell which wishes to invoke the "logical display path" functionality can do so explicitly by passing `--logical-dir`.
    - Any shell which wants to just override `current_dir` at all levels can safely do so by just passing `--logical-dir`.

davidkna:
I think I'd prefer your second solution, but both options sound good.

deadalusai:
@davidkna Closed in favor of https://github.com/starship/starship/pull/2104

PR #2098, opened 2021-01-07
fix(nix_shell): remove extra whitespace of symbol

EMayej:
Documents don't need to change because they have only one whitespace.

#### Motivation and Context

Closes #

#### Checklist:


chipbuster:
This looks like a duplicate of https://github.com/starship/starship/pull/1364, which was closed because those spaces are there to work around an issue with emoji widths.

Perhaps the thing to do is to add a comment to that line of code explaining what's going on, since we've gotten two PRs addressing this within as many weeks.

PR #2104, opened 2021-01-09
refactor(directory): Introduce `logical-path` argument which allows a shell to explicitly specify both a logical and physical filesystem path

deadalusai:
#### Description
Introduce `logical-path` argument which allows a shell to explicitly specify both a logical and physical filesystem path.

Fix `directory::module` to consume both path and logical-path (if provided).  The "logical" path is preferred when rendering the "display path", while the "physical" path is used to resolve the "read only" flag. Repo- and home-directory contraction is based on the logical path if it is set, or the physical path if it is not.

The custom "get_current_dir" logic has been removed entirely, and the `directory` module now relies on `context.current_dir` / `context.logical_dir` directly.

Changes have been made to `init/starship.ps1` to work with this new flag:
- Calculate and pass "physical" and "logical" paths explicitly (as other shells do not pass `--logical-path` they always rendering the based on the physical path)
- Moved the "powershell provider prefix" cleanup code to the PowerShell prompt script itself - this code _should_ now support any kind of powershell path prefix.

#### Motivation and Context
I discovered this issue while testing `starship` with the [Nu shell](https://github.com/nushell/nushell). It's a WIP shell, so still has some bugs. I switched to passing `--path` to work around one such issue and found that `starship` was inconsistent with how it handled the argument.

This screenshot demonstrates the problem:
starship_current_dir_resolution
Specifically, invoking the command `starship prompt --path C:\Windows`:
- When called with no `PWD` environment variable, prints `C:/Windows {lock symbol}` ✔️
- When called with a `PWD` environment variable, prints `{the PWD env variable} {lock symbol}` ❌

The issue is caused by the `directory` module function `get_current_dir` which looks to the `PWD` environment variable to select the path it is going to display to the user, overriding the provided `--path` argument.

To resolve this, I've introduced a new `prompt` argument `--logical-path` - rather than setting the PWD environment variable, a shell with "logical" path display requirements can instead provide `--logical-path` to set the logical path.

The same command as above, but with the fix applied:
image

See discussion on #2095 for more context.
This (abandoned?) PR #1209 also looked to introduce `--logical-path` as an alternative to PWD.

I tested manually using the following scripts:

Windows/PowerShell:
```powershell
function test ($Path, $LogicalPath) {
    "Testing: starship prompt --path '$Path' --logical-path '$LogicalPath'"
    target\debug\starship.exe prompt --path $Path --logical-path $LogicalPath
    "`n"
}

New-PSDrive -Name Sys32 -PSProvider FileSystem -Root 'C:\Windows\System32' -ErrorAction SilentlyContinue | Out-Null
New-PSDrive -Name Dev -PSProvider FileSystem -Root 'C:\Users\Ben Fox\Dev' -ErrorAction SilentlyContinue | Out-Null

test 'C:\Users\Ben Fox\Dev\starship' 'C:\Users\Ben Fox\Dev\starship'
test 'C:\Users\Ben Fox\Dev\starship' 'Dev:\starship'
test 'C:\Users\Ben Fox\Dev\starship' 'whatever\basically'
test 'C:\Windows\System32\' 'C:\Windows\System32\'
test 'C:\Windows\System32\' 'Sys32:\'
test 'HKEY_LOCAL_MACHINE\SOFTWARE' 'HKLM:\SOFTWARE'
```
(results)
image

Linux/bash:
```bash
function test() {
    echo Testing: starship prompt --path "$1" --logical-path "$2"
    target/debug/starship prompt --path "$1" --logical-path "$2"
    echo
}

test '~/Dev/starship' '~/Dev/starship'
test '~/Dev/starship' 'logical/path'
test '/' '/'
test '/' 'logical/path'
```
(results)
image

#### Checklist:


deadalusai:
Something I noted during testing is that navigating to a repository directory with a non-physical logical directory "breaks" the repo truncation behavior:

image

I don't think this is a regression, but something we could improve.

src/modules/directory.rs
```diff
@@ -33,43 +31,52 @@ const HOME_SYMBOL: &str = "~";
...
-        Some(repo_root) if config.truncate_to_repo && (repo_root != &home_dir) => {
-            log::debug!("Repo root: {:?}", repo_root);
-            // Contract the path to the git repo root
-            contract_repo_path(current_dir, repo_root)
-                .unwrap_or_else(|| contract_path(current_dir, &home_dir, HOME_SYMBOL))
-        }
-        // Contract the path to the home directory
-        _ => contract_path(current_dir, &home_dir, HOME_SYMBOL),
+    let physical_dir = &context.current_dir;
+    let logical_dir = context.logical_dir.as_ref().unwrap_or(physical_dir);
```
davidkna, line 44: I think `std::fs::canonicalize` should be run on if `use_logical_dir` is false.

src/configs/directory.rs
```diff
@@ -9,7 +9,6 @@ pub struct DirectoryConfig<'a> {
     pub truncate_to_repo: bool,
     pub substitutions: IndexMap<String, &'a str>,
     pub fish_style_pwd_dir_length: i64,
-    pub use_logical_path: bool,
```
davidkna, line 12: Why did you remove this option?

src/init/starship.ps1
```diff
@@ -1,16 +1,50 @@
...
+        LogicalPath =
+            if ($cwd.Path.StartsWith($provider_prefix)) {
+                $cwd.Path.Substring($provider_prefix.Length)
+            } else {
+                $cwd.Path
+            };
+    }
+}
+
+function Fix-Arg ($arg) {
```
davidkna, line 23: Manual quoting doesn't work? And backslashes aren't even supposed to be escape characters in powershell. Also why would the shell need to add literal quotes?

davidkna (changes requested):

davidkna:
> 
> 
> Something I noted during testing is that navigating to a repository directory with a non-physical logical directory "breaks" the repo truncation behavior:
> 
> image
> 
> I don't think this is a regression, but something we could improve.
> 
> Edit:
> A small change produces this result
> 
> ```diff
> diff --git a/src/modules/directory.rs b/src/modules/directory.rs
> index 85da1c3..80a9a82 100644
> --- a/src/modules/directory.rs
> +++ b/src/modules/directory.rs
> @@ -56,7 +56,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
>      let dir_string = repo
>          .and_then(|r| r.root.as_ref())
>          .filter(|root| *root != &home_dir)
> -        .and_then(|root| contract_repo_path(&logical_dir, root))
> +        .and_then(|root| contract_repo_path(&physical_dir, root))
>          .unwrap_or_else(|| contract_path(&logical_dir, &home_dir, HOME_SYMBOL));
> 
>      log::debug!("Dir string: {}", dir_string);
> ```
> 
> image

But this could cause unexpected behavious outside powershell (symlinks etc). Maybe a third `--display-path` option is needed after all.

src/init/starship.ps1
davidkna: Edit: looks like this is an actual issue but you could use `.TrimEnd("\")`

src/configs/directory.rs
deadalusai: Apologies, I should have dropped a PR comment explaining this.

So I spent a bit of time considering this option and decided that it had the potential to be a bit of a footgun. As I understand it there's currently only one shell which supports logical paths (PowerShell) and I found that it was very easy to get "stuck" in a virtual path... here's an example of what I mean in GIF form:

use_logical_path_false_behavior

src/init/starship.ps1
deadalusai: This isn't an issue with escaping characters in PowerShell code, but rather in the way PS generates the [Windows command-line argument string](https://docs.microsoft.com/en-us/previous-versions//17w5ykft(v=vs.85)?redirectedfrom=MSDN) when it invokes an external program, and there seems to be a bug in this routine.

In the documentation I linked above, note the final three rules:
- Backslashes are interpreted literally, unless they immediately precede a double quotation mark.
- If an even number of backslashes is followed by a double quotation mark, one backslash is placed in the argv array for every pair of backslashes, and the double quotation mark is interpreted as a string delimiter.
- If an odd number of backslashes is followed by a double quotation mark, one backslash is placed in the argv array for every pair of backslashes, and the double quotation mark is "escaped" by the remaining backslash, causing a literal double quotation mark (") to be placed in argv.

E.g. for the following external command invocation:
```powershell
> $a = 'hello "world" goodbye "moon"'
> $b = 'path\without\whitespace\'
> $c = 'path\with whitespace\'
> externalprogram $a $b $c
```
PS generates the following Windows command-line string:
```cmd
externalprogram "hello \"world\" goodbye \"moon\"" path\without\whitespace\ "path\with whitespace\"
```

The first argument is quoted correctly including escaped quote marks. ✔️ 
The second argument is not quoted at all as it does not contain any whitespace. ✔️ 
The third argument however has whitespace _and_ a trailing, odd number of backslashes which ends up escaping the closing quote mark. ❌

PowerShell could handle this by detecting a trailing backslash and escaping it as part of the quoted command-line string serialisation, but for some reason it doesn't. I'm not sure why not - it seems an obvious oversight.
I'm using PowerShell 5.0 - I haven't tested with the cross-platform PowerShell version yet so maybe they've fixed it there.

deadalusai:
> But this could cause unexpected behavious outside powershell (symlinks etc).

Other shells do not pass `--logical-path` at all, which means that `contract_repo_path` will be called against the physical path anyway.

src/init/starship.ps1
deadalusai: In reading through that PowerShell issue thread I learnt about the the [stop-parsing symbol (`--%`)](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parsing?view=powershell-7.1). This was introduced in PowerShell 3.0 (released in 2012) so I think we can rely on it being available.

With the stop-parsing symbol, we can manually construct a command string which we know will be correctly parsed and PS will pass it unchanged to the Windows API.

deadalusai: I've been experimenting with an `Invoke-Native` function over here:
https://github.com/deadalusai/printargs/blob/master/test.ps1

This function works around the argument escaping issue by quoting/escaping arguments manually and using `--%` to pass the argument string through to the rust executable directly. The string is quoted using the Windows command-line argument string rules, but the code _appears_ to work on Linux (Ubuntu/Mint) correctly too.

deadalusai: Ok, it works correctly on Linux/Unix because the .NET runtime on Linux itself [parses the argument string into the `argv` collection](https://github.com/dotnet/runtime/blob/9c41693194e3b005c91ee471ab3e649d2479d34d/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.Unix.cs#L430) before forking the new process, so we don't need to rely on distro-specific argv parsing behavior.

["This follows the rules outlined in "Parsing C++ Command-Line Arguments" at https://msdn.microsoft.com/en-us/library/17w5ykft.aspx."](https://github.com/dotnet/runtime/blob/9c41693194e3b005c91ee471ab3e649d2479d34d/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.Unix.cs#L786-L789).

deadalusai: I've updated the PR with a more robust solution for invoking native executables (`Invoke-Native`) which ensures that the strings passed as arguments make their way into the Rust argv list in the same shape as they started, trailing backslashes and all.

There is also a PowerShell library called [Native](https://github.com/mklement0/Native) available on PSGallery which supports all platforms and works by forking out to `/bin/bash` or `cmd`, but that would be an external dependency which may not be desirable.

davidkna:
In traditional shells:

The physical path is the path on disk with all symlinks resolved. (`std::env::current_dir`, `std::fs::canonicalize($PWD)`)
The logical path is the path without any symlinks resolved. If you cd into a symlinked directory, the parent hierachy stays the same. (`$PWD`) 

You can't just remove the logical handling for the traditional shells.

I think something like this might work for `logical_dir`:

- `--logical-path` (if `use_logical_path` is false `std::fs::canonicalize`)
- `--path` (if `use_logical_path` is false `std::fs::canonicalize`)
- if `use_logical_path` `$PWD` else `std::env::current_dir()`
- fall back to `current_dir`,  (if `use_logical_path` is false `std::fs::canonicalize`)

deadalusai:
> In traditional shells:
> 
> The physical path is the path on disk with all symlinks resolved. (`std::env::current_dir`, `std::fs::canonicalize($PWD)`)
> The logical path is the path without any symlinks resolved. If you cd into a symlinked directory, the parent hierachy stays the same. (`$PWD`)
> 
> You can't just remove the logical handling for the traditional shells.

Ok, I think I understand what you're getting at here now.

It seems like we should aim to standardize what data `current_dir` and `logical_dir` hold, and populate them explicitly as part of the context population step. Would the following work?

- `current_dir` - holds the phyisical current directory.
     Initialised from `--path`, `$PWD` or `std::env::current_dir` and canonicalized (if it is possible to canonicalize).
     Could be renamed `canonical_dir`?
- `logical_dir` - holds the logical current directory.
    Initialised from `--logical-path`, `--path`, '$PWD', or `std::env::current_dir` and _not_ canonicalized.

`use_logical_path` would just select between those two values. Other modules which depend on `current_dir` should still be able to do so safely?

davidkna:
> `current_dir` - holds the phyisical current directory.
> Initialised from `--path`, `$PWD` or `std::env::current_dir` and canonicalized (if it is possible to canonicalize).

That would work, but `current_dir` doesn't really need to be canonicalized unless `use_logical_dirs` is false, because it only matters in the `directory` module. (`std::env::current_dir` is always canonicalized as far as I know.) Not sure how expensive `canonicalize` is.

> `logical_dir` - holds the logical current directory.
> Initialised from `--logical-path`, `$PWD`, or `std::env::current_dir` and _not_ canonicalized.

I think part of the aim for this pr for `logical_dir` to prefer `--path` unless `--logical-path` is also given? Do you intend to keep that behaviour?

deadalusai:
> That would work, but `current_dir` doesn't really need to be canonicalized unless `use_logical_dirs` is false, because it only matters in the `directory` module. (`std::env::current_dir` is always canonicalized as far as I know.) Not sure how expensive `canonicalize` is.

Looking at the existing code, I can only see one call to `canonicalize` as part of [real_path](https://github.com/starship/starship/blob/4de9e43cff46c834bd340d24a02fc95d85310a33/src/modules/directory.rs#L227-L243) which is used by `contract_real_path`. I haven't touched any of that code - are we sure we need to canonicalize?

> I think part of the aim for this pr for `logical_dir` to prefer `--path` unless `--logical-path` is also given? Do you intend to keep that behaviour?

My initial goal at least was just to make it possible to override the logical path from the command line and the addition of `--logical-path` solves that issue. But I think you're right in that it might fit the principal of least surprise for `--path` to imply `--logical-path` by default as that's definitely what confused me initially.

deadalusai:
> That would work, but `current_dir` doesn't really need to be canonicalized unless `use_logical_dirs` is false, because it only matters in the `directory` module. (`std::env::current_dir` is always canonicalized as far as I know.) Not sure how expensive `canonicalize` is.

Looking at the existing code, I can only see one call to `canonicalize` as part of [real_path](https://github.com/starship/starship/blob/4de9e43cff46c834bd340d24a02fc95d85310a33/src/modules/directory.rs#L227-L243) which is used by `contract_real_path`. I haven't touched any of that code - I'm not sure we need extra `canonicalize` calls here?

> I think part of the aim for this pr for `logical_dir` to prefer `--path` unless `--logical-path` is also given? Do you intend to keep that behaviour?

My initial goal at least was just to make it possible to override the logical path from the command line and the addition of `--logical-path` solves that issue. But I think you're right in that it might fit the principal of least surprise for `--path` to imply `--logical-path` by default as that's definitely what confused me initially.

src/configs/directory.rs
deadalusai: I've restored this configuration flag, though I think clearing it (at least under PowerShell) would be a bad idea.

davidkna:
> Looking at the existing code, I can only see one call to `canonicalize` as part of [real_path](https://github.com/starship/starship/blob/4de9e43cff46c834bd340d24a02fc95d85310a33/src/modules/directory.rs#L227-L243) which is used by `contract_repo_path`. I haven't touched any of that code - I'm not sure we need extra `canonicalize` calls here?

With the old code `std::env::current_dir` would have taken care of that, unless a fallback was used, which should almost never happen. All the other path sources need calls to `canonicalize`  unless `use_logical_path` is false.

> My initial goal at least was just to make it possible to override the logical path from the command line and the addition of `--logical-path` solves that issue. But I think you're right in that it might fit the principal of least surprise for `--path` to imply `--logical-path` by default as that's definitely what confused me initially.

I think having `--path` imply `--logical-path` would be best.

deadalusai:
@davidkna On further investigation it may not be safe to always `canonicalize` paths [on Windows](https://doc.rust-lang.org/std/fs/fn.canonicalize.html):

> On Windows, this converts the path to use extended length path syntax

`std::env::current_dir` does not return extended length path syntax under Windows.

davidkna:
> @davidkna On further investigation it may not be safe to always `canonicalize` paths [on Windows](https://doc.rust-lang.org/std/fs/fn.canonicalize.html):
> 
> > On Windows, this converts the path to use extended length path syntax
> 
> `std::env::current_dir` does not return extended length path syntax under Windows.

This might have the upside of getting starship to support long paths on windows! I think removing the `\\?\` prefix on windows for display should work.

deadalusai:
> This might have the upside of getting starship to support long paths on windows! (I think it might be best to run that on `current_dir` after all) I think removing the `\\?\` prefix on windows for display should work.

Ok, I've implemented this now and I think it works pretty well, given how I understand what you're aiming for. Tested on Linux and Windows and added test coverage.

The `use_logical_path` flag now switches between these two routines for resolving the physical/logical paths:
- `use_logical_path = true`
    - Physical path: `current_dir`
    - Logical path: `logical_dir`  or `current_dir` if it's not set
- `use_logical_path = false`
    - Physical path `current_dir` canonicalized (which resolves symlinks)
    - Logical path: `current_dir` canonicalized (and cleaned up when on Windows)
    
Example of the path canonicalization with `use_logical_path = false` on Windows:
image

And the same on Linux:
image

davidkna (changes requested):

src/modules/directory.rs
```diff
@@ -242,6 +236,28 @@ fn real_path<P: AsRef<Path>>(path: P) -> PathBuf {
...
+    #[cfg(windows)]
+    {
+        path.iter()
+            .enumerate()
+            .map(|(i, segment)| match (i, segment.to_str()) {
+                // Match and strip the Windows extended-path prefix segment if it is there
+                (0, Some(s)) if s.starts_with("\\\\?\\") => std::ffi::OsStr::new(&s[4..]),
+                _ => segment,
+            })
+            .collect::<PathBuf>()
```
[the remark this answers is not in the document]
deadalusai: That's much simpler than what I was doing 🤦 
According to https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation there are only two variations of the extended path prefix, so I'll just attempt to strip them both.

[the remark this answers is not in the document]
deadalusai: Hmm, `strip_prefix` doesn't appear to work with extended path prefixes - something to do with how the path components are calculated.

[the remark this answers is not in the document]
deadalusai: I've worked around this issue by doing the extended-path prefix stripping later in the process.

src/modules/directory.rs
```diff
@@ -33,43 +31,72 @@ const HOME_SYMBOL: &str = "~";
...
+        let logical_dir = context.logical_dir.as_ref().unwrap_or(physical_dir);
+        (physical_dir.clone(), logical_dir.clone())
+    } else {
+        // Explicitly canonicalize the current path to expand symlinks.
+        // For the logical path, use a cleaned version of the canonical path.
+        let p = &context.current_dir;
+        let physical_dir = p.canonicalize().unwrap_or_else(|_| p.clone());
+        let logical_dir = canonical_path_cleaned_for_display(&physical_dir);
+        (physical_dir, logical_dir)
+    };
```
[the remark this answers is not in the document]
deadalusai: No longer an issue after the `Context.current_dir`/`Context.logical_dir` refactor.

src/modules/directory.rs
```diff
@@ -33,43 +31,72 @@ const HOME_SYMBOL: &str = "~";
...
+        // Map the configured physical and logical paths directly.
+        // If no logical path is configured, fall back to current_dir.
+        let physical_dir = &context.current_dir;
+        let logical_dir = context.logical_dir.as_ref().unwrap_or(physical_dir);
+        (physical_dir.clone(), logical_dir.clone())
+    } else {
+        // Explicitly canonicalize the current path to expand symlinks.
+        // For the logical path, use a cleaned version of the canonical path.
+        let p = &context.current_dir;
+        let physical_dir = p.canonicalize().unwrap_or_else(|_| p.clone());
```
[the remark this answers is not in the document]
deadalusai: Ok, so I've run with this suggestion and made the following changes:

- `Context::current_dir` is now initialised from `--path` or `std::env::current_dir`
- Additionally `current_dir` is now always canonicalized (on *nix and windows), which expands symlinks.
   On Windows, this means that `current_dir` is now always using extended-path prefixes.
- `Context::logical_dir` is now initialised from `--logical-path` or `PWD` or `current_dir` (before canonicalization)

The `directory::module` routine now just picks between rendering `current_dir` and `logical_dir` depending on the `use_logical_dir` config setting.

davidkna:
WIth this PR on `\\NAS\Share` I get `\\NAS\Shar` instead from the directory module when `use_logical_paths` is `false`.

davidkna:
It looks like this also breaks `$HOME` -> `~` on Windows when `use_logical_dir` is `false`.

deadalusai:
> WIth this PR on `\\NAS\Share` I get `\\NAS\Shar` instead from the directory module when `use_logical_paths` is `false`. `\\NAS\Share/subdir` is fine.
 
This looks like a bug in `path-slash`:
https://github.com/rhysd/path-slash/blob/master/src/lib.rs#L147
It's always popping the last character in the buffer, but that character isn't a `/` when only the "prefix" match case has been hit [in the code above](https://github.com/rhysd/path-slash/blob/master/src/lib.rs#L136).
 
> It looks like this also breaks `$HOME` -> `~` on Windows when `use_logical_dir` is `false`.

This is happening because I'm mistakenly doing `starts_with` checks using paths that are or are not canonicalized (depending on your config). I could fix it by canonicalizing the inputs before doing the starts_with check, but that would have unintended side effects as canonicalization would resolve symlinks in your home directory. I've got some ideas for things to try.

davidkna:
Same issue with git repo contraction I think. I think it should be fine either to assume the logical path exists or using the physical path if the logical path doesn't exist.

deadalusai:
@davidkna I think I've found a robust solution for dealing with with comparisons between verbatim and non-verbatim paths (_verbatim_ is the Rust std lib's name for the Windows extended-path prefixes).

I've added two routines for doing `equals` and `starts_with` comparisons with Paths which knows how to compare verbatim and non-verbatim path prefixes. The Rust Path library is pretty robust here and does most of the heavy lifting.

If there are any other modules doing "starts with" checks on `context.current_dir` we would need to fix them.

deadalusai:
I've raised a PR to fix path-slash: https://github.com/rhysd/path-slash/pull/5

deadalusai:
Updated path-slash to latest. This fixes the issue with the trailing character of some Windows paths being truncated, e.g. `\\server\share` and `C:`.

src/init/starship.ps1
```diff
@@ -1,38 +1,79 @@
...
+    $Arguments = $Arguments | ForEach-Object {
+        $_ = $_ -Replace '(\\+)"','$1$1"' # Escape backslash chains immediately preceeding quote marks.
+        $_ = $_ -Replace '(\\+)$','$1$1'  # Escape backslash chains immediately preceeding the end of the string.
+        $_ = $_ -Replace '"','\"'         # Escape internal quote marks.
+        "`"$_`""                          # Quote the argument.
+    }
+    # Use the stop-parsing symbol (--%) to ensure the argument string is passed along unchanged 
+    # See: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parsing?view=powershell-5.1
+    Invoke-Expression "& `"$Executable`" --% $($Arguments -Join ' ')"
+}
```
davidkna, line 36: Some of the stuff that broke long paths appears to be happening in the powershell part. Here's a fix:
```suggestion
function Invoke-Native {
    param($Executable, $Arguments)

    # Build an arguments string which follows the C++ command-line argument quoting rules
    # See: https://docs.microsoft.com/en-us/previous-versions//17w5ykft(v=vs.85)?redirectedfrom=MSDN
    $Arguments = $Arguments | ForEach-Object {
        $t = $_ -Replace '(\\+)"','$1$1"' # Escape backslash chains immediately preceeding quote marks.
        $t = $t -Replace '(\\+)$','$1$1'  # Escape backslash chains immediately preceeding the end of the string.
        $t = $t -Replace '"','\"'         # Escape internal quote marks.
        "`"$t`""                          # Quote the argument.
    }

    $process = New-Object System.Diagnostics.Process
    $process.startInfo.FileName = $Executable
    $process.startInfo.Arguments = $Arguments
    
    # Use UTF-8
    $process.startInfo.StandardOutputEncoding = [System.Text.Encoding]::UTF8
    # Required for UTF-8
    $process.startInfo.RedirectStandardOutput = $true

    $process.startInfo.CreateNoWindow = $true
    $process.startInfo.UseShellExecute = $false

    $process.Start() > $null

    $process.StandardOutput.ReadToEnd()
}
```

src/init/starship.ps1
```diff
@@ -1,38 +1,79 @@
...
+    $arguments = @(
+        "prompt"
+        "--path=$($cwd.Path)",
+        "--logical-path=$($cwd.LogicalPath)",
+        "--jobs=$($jobs)"
+    )
 
     # Save old output encoding and set it to UTF-8
     $origOutputEncoding = [Console]::OutputEncoding
     [Console]::OutputEncoding = [System.Text.Encoding]::UTF8
```
davidkna, line 55: Whit the fix above this isn't needed anymore.
```suggestion
```

src/init/starship.ps1
```diff
@@ -1,38 +1,79 @@
...
-        $out = @(&::STARSHIP:: prompt "--path=$current_directory" --status=$lastExitCodeForPrompt --jobs=$jobs --cmd-duration=$duration)
-    } else {
-        $out = @(&::STARSHIP:: prompt "--path=$current_directory" --status=$lastExitCodeForPrompt --jobs=$jobs)
+        
+        $arguments += "--cmd-duration=$($duration)"
     }
+
+    $arguments += "--status=$($lastExitCodeForPrompt)"
+
+    $out = Invoke-Native -Executable ::STARSHIP:: -Arguments $arguments
```
davidkna, line 75: Just this is enough to get the prompt with the above changes:
`Invoke-Native -Executable ::STARSHIP:: -Arguments $argument`

Please get rid of `$out` and the `[Console]::OutputEncoding` stuff.

src/modules/utils/path.rs
```diff
@@ -0,0 +1,320 @@
...
+        Prefix::Verbatim(segment) => NormalizedPrefix::Verbatim(segment),
+        Prefix::VerbatimUNC(server, share) => NormalizedPrefix::UNC(server, share),
+        Prefix::VerbatimDisk(disk) => NormalizedPrefix::Disk(disk),
+        Prefix::DeviceNS(device) => NormalizedPrefix::DeviceNS(device),
+        Prefix::UNC(server, share) => NormalizedPrefix::UNC(server, share),
+        Prefix::Disk(disk) => NormalizedPrefix::Disk(disk),
+    }
+}
+
+fn normalize_path(path: &Path) -> (NormalizedPrefix, &std::path::Path) {
```
davidkna, line 31: Can you make these two functions no-op when not on windows? `#[cfg(windows)]`?

src/init/starship.ps1
davidkna: Edit: with this approach to command execution, instead of your escape code, this worked:
```ps1
    foreach ($arg in $Arguments) {
        $process.startInfo.ArgumentList.Add($arg)
    }
```

src/init/starship.ps1
```diff
@@ -1,38 +1,79 @@
 #!/usr/bin/env pwsh
 
+function Get-Cwd {
```
davidkna, line 3: Would you please prefix these with `starship-` or find some way to avoid getting these functions into the shell scope?

src/init/starship.ps1
davidkna: I tested your version of argument escaping, because it looks like `ArgumentList` doesn't work on PS5, but it looks like it doesn't work. Maybe try  implementing a solution based on something like https://github.com/PowerShell/PowerShell/pull/13482#issuecomment-678522514? (I'd still prefer using `ArgumentList` for PS 6.1 and up)

Because this all looks very broken without `ArgumentList` I think it might be best to just drop support for PS < 6.1 if you don't feel like doing that.

deadalusai: `ProcessStartInfo.ArgumentList` was only added in .NET 5.0, so this code will only work in PowerShell 6+

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.argumentlist?view=net-5.0&viewFallbackFrom=netframework-4.8#System_Diagnostics_ProcessStartInfo_ArgumentList

(which sucks because I'd much rather not be doing this manual escaping)

src/init/starship.ps1
deadalusai: I've moved these function declarations into the `global:prompt` function itself so they no longer leak into the user's shell scope.

src/init/starship.ps1
deadalusai: I've adjusted the `Invoke-Native` function to use `System.Diagnostics.ProcessStartInfo` as in your suggested change above. If `ArgumentList` is available we use it, otherwise we fall back to building an escaped argument string.

Do you have any examples of argument strings which break the manual escaping process? All my test sample strings work correctly:

```
$a = 'hello "world" \" \\" goodbye "moon"' 
$b = 'path\without\whitespace\'
$c = 'path\with whitespace\'
$d = '--flag=argument passed to flag'
$e = 'five trailing backslashes \\\\\'
$f = '\\\\\five leading backslashes'
$g = '\\?\C:\Windows\Extended Path Syntax\'
```

davidkna: It looks like your solution is sometimes appending `;` to the argument.
When I got rid of all the `;` in `ForEach-Object` it appeared to work fine. You also might want to avoid assigning to `$_`. The VSCode PowerShell extension complains about that.
Beyond that: `ArgumentList` was added in PowerShell 6.1 with .Net Core 2.1. I think only PowerShell 7.1 has .Net 5.

deadalusai: Ah, there was a mistyped `;` I thought I'd already fixed, but I'd only fixed it in my testing script 🤦‍♂️.

Re: `ArgumentList` that makes sense. I was following the docs but I think I was in the "framework" docs not the "core" docs. I'm testing for the property, not the PowerShell version so it will work on PS 6 & 7.

davidkna (approved):
LGTM. But because this PR changes some of the fundamentals I'd prefer a second review.

deadalusai:
@davidkna Is there still interest in merging this?

davidkna:
I agree that this should be merged soon. Unless someone finds any issues or objects to merging I think I'll merge this tomorrow.

PR #2105, opened 2021-01-09
docs: Add a comment describing double-space in nix config

chipbuster:
#### Description
We have had two attempts now to "fix" the two spaces in the nixshell module's default config, which are needed to provide a workaround for rendering of multiwidth emojis. Adds a comment to the code there describing why the spaces are present so that any would-be contributors know that it is not an accident.

#### Checklist:


chipbuster:
What in the actual heck happened with this commit history?

I guess my fork's master and this repo's master have somehow come unsynced?

chipbuster:
This is ridiculous. I'm going to reopen another PR in a way that doesn't require us to review the entire codebase to merge.

PR #2121, opened 2021-01-11
fix(install): improve supported target detection

davidkna:
#### Description

Improve detection of supported platforms. Previously there was only a check to see if the install was on `i386` (this still isn't supported but `i686` now is), instead check against a list of supported install targets.
In addition:
- Detect fake `x64` as `i686` instead of `i386`
- Replace `\n` in `info` and `error` with `printf "\n"`

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -33,6 +33,8 @@ BLUE="$(tput setaf 4 2>/dev/null || echo '')"
 MAGENTA="$(tput setaf 5 2>/dev/null || echo '')"
 NO_COLOR="$(tput sgr0 2>/dev/null || echo '')"
 
+SUPPORTED_TARGETS="x86_64-unknown-linux-gnu x86_64-unknown-linux-musl i686-unknown-linux-musl x86_64-apple-darwin x86_64-pc-windows-msvc"
+
 info() {
   printf "%s\n" "${BOLD}${GREY}>${NO_COLOR} $*"
 }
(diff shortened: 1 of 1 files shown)


install/install.sh
```diff
@@ -242,6 +242,33 @@ check_bin_dir() {
...
 
+is_build_available() {
+  local arch="$1"
+  local platform="$2"
+
+  local target="${arch}-${platform}"
+  local good
+  
+  good=$(
+    for t in x86_64-unknown-linux-gnu x86_64-unknown-linux-musl i686-unknown-linux-musl x86_64-apple-darwin x86_64-pc-windows-msvc; do
```
chipbuster, line 253: Could we move this to some array (or pseudo-array) so that there's a single obvious source of truth for which platforms are supported by the install script?

If that's too much of a pain in POSIX I'm not too attached to this request.

chipbuster (commented):
Just one request. I don't know how easy this is in POSIX (I think using a space-delimited string should work, but there may be some splitting nonsense I haven't thought of).

install/install.sh
davidkna: Done!

chipbuster (approved):
LGTM! :+1:

andytom:
Thank you for your contribution @davidkna.

PR #2146, opened 2021-01-14
perf(git_status): add option to use windows starship.exe to render in wsl

davidkna:
#### Description

Because I/O is slow on Windows paths with WSL, add the option to render `git_status` with a Windows `starship.exe`. This is 10x faster. This still uses the linux starship config.

#### Motivation and Context

Closes #

- [x] I have tested using **WSL1**
- [x] I have tested using **WSL2**

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1457,25 +1457,33 @@ format = '[+$added]($added_style)/[-$deleted]($deleted_style) '
 The `git_status` module shows symbols representing the state of the repo in your
 current directory.
 
+::: tip
+
+The Git Status module is very slow in Windows directories (for example under `/mnt/c/`) when in a WSL environment.
+You can disable the module or use the `windows_starship` option to use a Windows-native Starship executable to compute `git_status` for those paths.
+
+:::
+
 ### Options
 
-| Option              | Default                                       | Description                         |
-| ------------------- | --------------------------------------------- | ----------------------------------- |
-| `format`            | `'([\[$all_status$ahead_behind\]]($style) )'` | The default format for `git_status` |
-| `conflicted`        | `"="`                                         | This branch has merge conflicts.    |
-| `ahead`             | `"⇡"`                                         | The format of `ahead`               |
-| `behind`            | `"⇣"`                                         | The format of `behind`              |
-| `diverged`          | `"⇕"`                                         | The format of `diverged`            |
-| `up_to_date`        | `""`                                          | The format of `up_to_date`          |
-| `untracked`         | `"?"`                                         | The format of `untracked`           |
-| `stashed`           | `"$"`                                         | The format of `stashed`             |
-| `modified`          | `"!"`                                         | The format of `modified`            |
-| `staged`            | `"+"`                                         | The format of `staged`              |
-| `renamed`           | `"»"`                                         | The format of `renamed`             |
-| `deleted`           | `"✘"`                                         | The format of `deleted`             |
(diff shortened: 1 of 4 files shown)


BatmanAoD:
I think this looks like the best possible solution for getting git status for Windows directories from within WSL, which I think is a pretty reasonable use-case.

docs/config/README.md
```diff
@@ -1132,6 +1132,13 @@ cherry_pick = "[🍒 PICKING](bold red)"
 The `git_status` module shows symbols representing the state of the repo in your
 current directory.
 
+::: tip
+
+The Git Status module is very slow in Windows directories (anything in `/mnt/c/`) when in a WSL environment.
```
BatmanAoD, line 1137: The phrasing here made me concerned that the implementation would actually be parsing the Linux path and assuming that `/mnt/c` itself indicates whether something is on the host file system (which would not be quite correct), but since that's not the case, this comment should probably make it clear that this is merely a common example. 

```suggestion
The Git Status module is very slow in Windows directories (under `/mnt/c/`, for example) when in a WSL environment.
```

jsco2t:
@BatmanAoD I noticed this hasn't been merged, is it because of your pending comment above? If that's the case - if @davidkna happens to be busy I would be willing to make the requested change and submit a new PR. I'm using WSL daily and am constantly hitting this same issue. `git_status` is always timing out.

BatmanAoD:
@jsco2t I'm not actually a maintainer, and my comment was already addressed, so I don't know what (if anything) is blocking this.

src/modules/git_status.rs
```diff
@@ -354,6 +361,107 @@ fn format_symbol(format_str: &str, config_path: &str, context: &Context) -> Opti
...
+                log::Level::Error
+            };
+
+            log::log!(level, "Failed to get Windows path:\n{:?}", e);
+
+            return None;
+        }
+    };
+
+    let winpath = match std::str::from_utf8(&winpath.stdout) {
```
chipbuster, line 404: Just checking: there's no way this should ever return a UNC path right?

docs/config/README.md
```diff
@@ -1457,25 +1457,33 @@ format = '[+$added]($added_style)/[-$deleted]($deleted_style) '
 The `git_status` module shows symbols representing the state of the repo in your
 current directory.
 
+::: tip
+
+The Git Status module is very slow in Windows directories (for example under `/mnt/c/`) when in a WSL environment.
+You can disable the module or use the `windows_starship` option to use a Windows-native Starship executable to compute `git_status` for those paths.
```
chipbuster, line 1463: Do we want to add an example here or would that clutter things too much? (e.g. for a system-wide install, `/mnt/c/Program Files/starship/starship.exe` or something of that nature)

chipbuster (approved):
LGTM. I have one minor question about WSL paths (having to do with the recent `dunce` merge) and one about examples, but they're pretty minor--if desired, we can merge as-is and come back to those later.

src/modules/git_status.rs
davidkna: It does not seem to return an UNC path for long paths anyway. In any case, `dunce::canoncialize` is just `std::fs::canoncialize`, followed by `dunce::simplified` to remove the UNC prefix from `std::fs::canoncialize` if possible, so even a pre-prefixed path would be simplified.

docs/config/README.md
davidkna: I think the tips already look fairly cluttered, and I already added an example below.

PR #2151, opened 2021-01-15
test: enable logging in tests

davidkna:
#### Description

This PR enables logging during testing with `cargo test`. I put the log initialisation into `ModuleRender` because most tests should hit that. The log is shown on test failure. This might help investigate the cause behind #886 and help investigate the cause of test failures in general.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -18,7 +18,7 @@ pub struct StarshipLogger {
 }
 
 impl StarshipLogger {
-    fn new() -> Self {
+    pub fn new() -> Self {
         let log_dir = env::var_os("STARSHIP_CACHE")
             .map(PathBuf::from)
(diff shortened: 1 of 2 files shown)


chipbuster (approved):
LGTM.

There is a very small possibility that I'll want to add an envar to enable/disable this later on since the AUR package does run the (non-ignored) tests during the install procedure, but let's worry about that if someone complains. Native-built Electron apps also dump a lot of text to terminal, and I've yet to hear anyone complain about that.

davidkna:
This shouldn't be an issue because the log messages should only become visible once a test fails.

andytom:
Thank you for your contribution @davidkna

PR #1233, opened 2020-05-19
fix: Allow configuration of root keys from cli

arawnx:
#### Description
This fork resolves bug #1199; the bug was as such: If no table is specified (using a period) holding the value in the command `starship config [value] [key]`, an error is returned. This bug-fix permits changing table-less values. Additionally, the function `update_configuration(name: &str, value: &str)` is made somewhat more verbose.

#### Motivation and Context
Closes #1199 

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### How Has This Been Tested?
I have run a test as follows:

1. Execute `cargo test`; no failed runs
2. Execute `starship config key value`; creates a table-less index "key" with value "value" in config file starship.toml
3. Execute `starship config table.key value`; works as above, with "key" under the table "table"
4. Perform the above two on already-existing TOML variables; correctly replaces the current value without adding a new one.
- [x] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.
(Neither are applicable)

PR #2154, opened 2021-01-16
perf(configure): Don't allocate when unnecessary

hbina:
Signed-off-by: Hanif Ariffin <hanif.ariffin.4326@gmail.com>

#### Description

Just use the iterator instead.

#### Motivation and Context

Closes #

#### Checklist:


davidkna:
I think this PR wouldn't work well with future changes to the `config` command like to allow changes to top-level items (#1233), so I think it would be better not to merge this.

hbina:
I didn't know just. Feel free to close if you feel like it

PR #2155, opened 2021-01-16
test: avoid setting $HOME

### LINKED ISSUE #886: Intermittent Test Failures in GH Actions
# Bug Report

Our CI appears to be intermittently failing. Last week I got a failure in the `go` tests which went away on rerun. Just now, we got a failure in `nodejs` for a commit that only added comments.

I'm not sure what is causing this (a race condition in our test suite, a bug in GH Actions, UB in one of our dependencies) but I'm making an issue to track these. Hopefully we'll end up seeing some sort of pattern that we can exploit.

If you think you see an intermittent failure in Actions, please post the failed test here along with the following info:

- Related PR
- What the change was that triggered the failure
- The `cargo test` output for the failed tests.

davidkna:
#### Description

Add a method to `Context` to get the home directory that first attempts to read `$HOME` with mocked environment variables  from `self.get_env` during testing.

#### Motivation and Context

Related to #886 might fix to some extent
https://github.com/starship/starship/issues/886#issuecomment-761162673

Beyond what's described in that comment `$HOME` was also being set in `test_get_config_path`.

#### Checklist:

### DIFF
--- a/src/bug_report.rs
+++ b/src/bug_report.rs
@@ -254,10 +254,10 @@ mod tests {
     #[test]
     #[cfg(not(windows))]
     fn test_get_config_path() {
-        env::set_var("HOME", "/test/home");
-
         let config_path = get_config_path("bash");
-        assert_eq!("/test/home/.bashrc", config_path.unwrap().to_str().unwrap());
-        env::remove_var("HOME");
+        assert_eq!(
+            dirs_next::home_dir().unwrap().join(".bashrc"),
+            config_path.unwrap()
+        );
     }
 }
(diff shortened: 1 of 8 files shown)


andytom:
This might also fix #1734.

davidkna:
I don't think so because `get_home` has a fallback to `dirs_next::home_dir()`. It might help make #1735 work though.

andytom (approved):
LGTM

andytom:
Thanks for implementing this hopefully this helps

PR #2131, opened 2021-01-12
feat(task_warrior): introduce task_warrior module

Aelnor:
#### Description
Introduce 'task_warrior' module. It shows various icons depending on due tasks (overdue/due today/this week/no due) and current active task.
The module is disabled by default.

#### Motivation and Context
Closes #1296

#### Screenshots (if appropriate):
Screenshot from 2021-01-12 21-04-23

#### How Has This Been Tested?
- [ ] I have tested using **MacOS**
- [X] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:
- [X] I have updated the documentation accordingly.
- [X] I have updated the tests accordingly.

PR #2156, opened 2021-01-16
feat(timewarrior): Timewarrior module

Rasmus-Bertell:
#### Description

Add a timewarrior module. Still needs testing and tag support.
Also it's possible to add the symbol to taskwarrior and forget about timewarrior tags since they are usually used together, but I believe it's better to have separate modules.

#### Motivation and Context

I'm waiting for Taskwarrior module to be added and noticed the lack of Timewarrior module.
I tend to forget my timetracking on so this would help with that.

- [ ] Tested using **MacOS**
- [x] Tested using **Linux**
- [ ] Tested using **Windows**

#### Checklist:

- [x] Add support for tags
- [x] Update the documentation
- [x] Update the tests

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2454,14 +2454,14 @@ To enable it, set `disabled` to `false` in your configuration file.
 :::
 
 ### Options
-| Option           | Default | Description                            |
-| ---------------- | ------- | -------------------------------------- |
-| `symbol`         | ``     | Displayed symbol                       |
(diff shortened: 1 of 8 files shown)


RasmusBertell:
By the way, I'm a total rust newbie, so any improvements and tips are more than welcome.

RasmusBertell:
This pairs nicely with #2131 in my opinion and I consider this to be done(ish) now.

PR #2160, opened 2021-01-17
perf(nodejs): evaluate nodejs format string lazily

HansAuger:
#### Description
Update Format string. Does not do lazy loading since the version is also needed for determining the style, so version has to be evaluated at least once.

#### Motivation and Context
Relates to #2111

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1760,7 +1760,7 @@ The module will be shown if any of the following conditions are met:
 
 | Option     | Default                            | Description                                        |
 | ---------- | ---------------------------------- | -------------------------------------------------- |
-| `format`   | `"via [$symbol$version]($style) "` | The format for the module.                         |
+| `format`   | `"via [$symbol($version )]($style)"` | The format for the module.                         |
 | `symbol`   | `"⬢ "`                             | A format string representing the symbol of NodeJS. |
 | `style`    | `"bold green"`                     | The style for the module.                          |
 | `disabled` | `false`                            | Disables the `nodejs` module.                      |
--- a/src/configs/nodejs.rs
+++ b/src/configs/nodejs.rs
@@ -14,7 +14,7 @@ pub struct NodejsConfig<'a> {
 impl<'a> RootModuleConfig<'a> for NodejsConfig<'a> {
     fn new() -> Self {
         NodejsConfig {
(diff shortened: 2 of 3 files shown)


andytom (commented):
Hi @HansAuger, thanks for making this change. I think you will also need to update the tests in `src/modules/nodejs.rs` but other than that LGTM.

HansAuger:
True. done!

HansAuger:
Fetching the nodejs version lazily is a bit more cumbersome here since it's needed in two places in the `match`, not only for the version string itself but also to determine the style. 

I tried to have a mutable reference to the version string around at the level of the `module` function and then have a function that would do the `node --version` call and set the value only if this reference is empty. But the compiler didn't seem happy about it in a way my rust exp wasn't apt enough to deal with. I  am open to suggestion on how to tackle this.

Although it's also safe to say that the style is probably evaluated for most users, so it might not be worth the trouble. In that case I could also add a comment why the version is not evaluated lazily for nodejs, since that would differ from all other/most modules.

davidkna:
I made it work:

At the top of the file add `use once_cell::sync::Lazy;`
and use this to get the node version:
```rust
    let nodejs_version =
        Lazy::new(|| utils::exec_cmd("node", &["--version"]).map(|cmd| cmd.stdout));
        formatter
            .map_meta(|var, _| match var {
                "symbol" => Some(config.symbol),
                _ => None,
            })
            .map_style(|variable| match variable {
                "style" => {
                    let engines_version = get_engines_version(&context.current_dir);
                    let in_engines_range =
                        check_engines_version(((*nodejs_version).as_ref())?, engines_version);
                    if in_engines_range {
                        Some(Ok(config.style))
                    } else {
                        Some(Ok(config.not_capable_style))
                    }
                }
                _ => None,
            })
                "version" => (*nodejs_version)
                    .as_ref()
                    .map(|version| version.trim())
                    .map(Ok),
                _ => None,
            })
            .parse(None)
    });
```

HansAuger:
@davidkna Amazing, thanks for the input! I implemented your suggestion.

andytom (approved):
LGTM

andytom:
Thanks for fixing this @HansAuger and thanks for the review @davidkna

PR #2166, opened 2021-01-19
perf(dart): Lazy eval dart

### LINKED ISSUE #2111: Refactor modules to lazily execute commands

Many of our modules will execute a command to retrieve a variable even if it is not used, for example, in the [crystal module](https://github.com/starship/starship/blob/master/src/modules/crystal.rs#L23) we execute the `crystal --version` command even if the `version` variable is not used in the format string.

Where possible I would like us to lazily execute these commands only when the module uses the variable. An example of a module that does this is the [ruby module](https://github.com/starship/starship/blob/master/src/modules/ruby.rs#L35-L39). This should give two benefits the first is that for users who are not using the variable in the format string starship will be faster since it will not execute the command (see #2084), the second is that if the command fails that module will fail gracefully and still display parts that are available

```
/tmp/tmp.6sbhJnyHpX via 💎  
❯ ls
test.cr  test.rb
/tmp/tmp.6sbhJnyHpX via 💎  
❯ ruby --version
bash: ruby: command not found
/tmp/tmp.6sbhJnyHpX via 💎  
❯ crystal --version
bash: crystal: command not found
```

Modules that need to be updated:
* [x] ~cmake (format string only) - #2135~
* [x] ~crystal - #2129~
* [x] ~dotnet  (format string only) - #2165~
* [x] ~dart - #2166~
* [x] ~elixir - #2172~
* [x] ~elm - #2167~
* [x] ~erlang  (format string only) - #2165~
* [x] ~golang (format string only) - #2165~
* [x] ~helm  (format string only) - #2165~
* [x] ~java - #2168~
* [x] ~julia  (format string only) - #2165~
* [x] ~kotlin - #2186~
* [x] ~lua - #2185~
* [x] ~nim  (format string only) - #2165~
* [x] ~nodejs - #2160~ 
* [x] ~ocaml - #2188~
* [x] ~perl - #2189~
* [x] ~php - #2190~
* [x] ~purescript - #2191~
* [x] ~python - #2158~
* [x] ~ruby  (format string only) - #2165~
* [x] ~rust (format string only) -  #2165~
* [x] ~swift - #2159~ 
* [x] ~zig - #2187~
* [x] ~custom - #2173~

HansAuger:
#### Description
Lazy format evaulation and format string change for dartlang

#### Motivation and Context

Related to: #2111

#### Checklist:

### DIFF
--- a/src/configs/dart.rs
+++ b/src/configs/dart.rs
@@ -13,7 +13,7 @@ pub struct DartConfig<'a> {
 impl<'a> RootModuleConfig<'a> for DartConfig<'a> {
     fn new() -> Self {
         DartConfig {
-            format: "via [$symbol$version]($style) ",
+            format: "via [$symbol$version ]($style)",
             symbol: "🎯 ",
             style: "bold blue",
             disabled: false,
--- a/src/configs/dart.rs
+++ b/src/configs/dart.rs
@@ -13,7 +13,7 @@ pub struct DartConfig<'a> {
 impl<'a> RootModuleConfig<'a> for DartConfig<'a> {
     fn new() -> Self {
         DartConfig {
-            format: "via [$symbol$version ]($style)",
+            format: "via [$symbol($version )]($style)",
             symbol: "🎯 ",
             style: "bold blue",
             disabled: false,
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -604,12 +604,12 @@ The module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Option     | Default                            | Description                                     |
-| ---------- | ---------------------------------- | ----------------------------------------------- |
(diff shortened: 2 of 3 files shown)


src/configs/dart.rs
```diff
@@ -13,7 +13,7 @@ pub struct DartConfig<'a> {
 impl<'a> RootModuleConfig<'a> for DartConfig<'a> {
     fn new() -> Self {
         DartConfig {
-            format: "via [$symbol$version]($style) ",
+            format: "via [$symbol$version ]($style)",
```
davidkna, line 16: You need to wrap this in braces so when `$version` is missing the extra space doesn't get printed.

```suggestion
            format: "via [$symbol($version )]($style)",
```

andytom (approved):
LGTM

andytom:
@HansAuger Thanks for updating this too.

PR #2171, opened 2021-01-19
perf(utils): Add timeout to `utils::exec_cmd`

andytom:
#### Description

This adds a timeout to any command executed using the `utils::exec_cmd`.
The initial time limit is hard coded to 500ms but if required we can
make this configurable. Have also switched the tests to be a bit more
granular on which systems they are ignored.

#### Motivation and Context

This should fix some issues with slow commands. 

Related to #312

#### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ indexmap = "1.6.1"
 notify-rust = { version = "4.2.2", optional = true }
 semver = "0.11.0"
 which = "4.0.2"
+process_control = { version = "3.0.0", features = ["crossbeam-channel"] }
 
 # Optional/http:
 attohttpc = { version = "0.16.1", optional = true, default-features = false, features = ["tls", "form"] }
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -282,7 +282,7 @@ fn internal_exec_cmd(cmd: &str, args: &[&str]) -> Option<CommandOutput> {
         }
     };
 
-    match process.with_output_timeout(time_limit).wait() {
+    match process.with_output_timeout(time_limit).terminating().wait() {
         Ok(Some(output)) => {
             let stdout_string = String::from_utf8(output.stdout).unwrap();
             let stderr_string = String::from_utf8(output.stderr).unwrap();
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,8 +1,9 @@
+use process_control::{ChildExt, Timeout};
 use std::fs::File;
 use std::io::{Read, Result};
 use std::path::Path;
-use std::process::Command;
-use std::time::Instant;
+use std::process::{Command, Stdio};
+use std::time::{Duration, Instant};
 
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -64,6 +64,7 @@ indexmap = "1.6.1"
 notify-rust = { version = "4.2.2", optional = true }
 semver = "0.11.0"
 which = "4.0.2"
+process_control = "3.0.0"
```
davidkna, line 67: This features is supposed to speed the crate up and it's already an indirect dependency.

```suggestion
process_control = { version = "3.0.0", features = ["crossbeam-channel"] }
```

src/utils.rs
```diff
@@ -258,15 +259,31 @@ fn internal_exec_cmd(cmd: &str, args: &[&str]) -> Option<CommandOutput> {
...
+        .spawn()
+    {
+        Ok(process) => process,
+        Err(error) => {
+            log::info!("Unable to run {:?}, {:?}", cmd, error);
+            return None;
+        }
+    };
+
+    match process.with_output_timeout(time_limit).wait() {
```
davidkna, line 285: I think terminating the process on timeout makes sense.

```suggestion
    match process.with_output_timeout(time_limit).terminating().wait() {
```

src/utils.rs
```diff
@@ -287,6 +304,10 @@ fn internal_exec_cmd(cmd: &str, args: &[&str]) -> Option<CommandOutput> {
                 stderr: stderr_string,
             })
         }
+        Ok(None) => {
+            log::warn!("Executing command {:?} timedout", cmd);
```
davidkna, line 308: I think this needs to separated with a space:
```suggestion
            log::warn!("Executing command {:?} timed out", cmd);
```

Cargo.toml
andytom: Good idea, I never thought to add while working on the poc.

src/utils.rs
andytom: I'm not sure about this the [process_control docs](https://docs.rs/process_control/3.0.0/process_control/struct.Terminator.html#safety) seem to indicate that `terminating` isn't safe to use in production.

davidkna: [terminat**ing**](https://docs.rs/process_control/3.0.0/process_control/trait.Timeout.html#tymethod.terminating) should be safe. [It's supposed what to be one of the things that sets it apart from other crates.](https://docs.rs/process_control/3.0.0/process_control/index.html#comparable-crates)

If you don't want to add `terminating` the crate isn't required for the timeout functionality.

```rust
let (sender, receiver) = channel();

thread::spawn(move|| {
    sender.send(process.output()).unwrap();
});

receiver.recv_timeout(duration)
```

andytom: Ah nice I didn't see that.

davidkna (approved):
LGTM

PR #2172, opened 2021-01-19
perf(elixir): evaluate version lazily

davidkna:
#### Description

This updates the module to lazily execute the `elixir --version` command only when the `$version` or `$otp_version` variables are used in the format string.

Because the output of the command is used to provide both `$version` and `$otp_version` I wrapped the call to `get_elixir_version` with `once_cell::sync::Lazy`.

I also updated the format string to better accommodate a missing `elixir` command.

#### Motivation and Context

Related to: #2111

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -808,12 +808,12 @@ The module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Option     | Default                                                 | Description                                                     |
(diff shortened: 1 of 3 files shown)


davidkna:
I think I've seen a few failures with installing mercurial lately. Maybe the install should be switched to `brew`?

andytom (approved):

andytom:
Thanks for your contribution @davidkna

PR #2174, opened 2021-01-19
ci(macos): install mercurial with brew

davidkna:
#### Description

On macOS in CI Install `mercurial` with `brew` instead of `pip3`, because installing with `pip3` was brittle lately. `brew` is also a bit faster to install `mercurial`

#### Motivation and Context

Closes #

#### Checklist:


andytom:
Thanks for the quick fix @davidkna

PR #2182, opened 2021-01-21
feat(custom): allow using bools for when

### LINKED ISSUE #2066: Add `custom` that does not require a shell

I want to have a fixed text on my prompt that shouldn't require any shell, but for now I am using `echo text` the timings on that is about 5ms.  I'd like it to be as fast as line break/character which is <1ms

By default this should be enough
```toml
[custom.z_bg]
format = "[\u001B\\[K]($style)"
style = "#101010 bg:#202020"
```

This is what I am doing right now

```toml
[custom.z_bg]
command = "echo "
format = "[$output\u001B\\[K]($style)"
shell = "sh"
# shell = ["bash", "--noprofile", "--norc"]
style = "#101010 bg:#202020"
when = "true"
```

### LINKED ISSUE #3601: New OS condition for custom module needs additional condition to succeed
#### Current Behavior
The OS condition for the custom module (implemented in #2750 and improved in #2843) needs an additional condition to succeed (`when` or `file` for instance)

#### Expected Behavior
The OS condition for the custom module should succeed when the OS condition is met but without additional conditions

In the below minimal configuration the prompt should display `with-second-condition without-second-condition` but in fact it shows only `with-second-condition`.

#### Environment
- Starship version: 1.3.0
- pwsh version: PowerShell 7.3.0-preview.1
- Operating system: Windows 10.0.22000
- Git Commit Hash: e6a921b5
- Branch/Tag: v1.3.0
- Build Time: 2022-02-07 15:08:03 +00:00

#### Starship Configuration

```toml
format = '${custom.with-second-condition} ${custom.without-second-condition}'

[custom.with-second-condition]
format = 'with-second-condition'
os     = 'windows'
when   = ''

[custom.without-second-condition]
format = 'without-second-condition'
os     = 'windows'
```

davidkna:
#### Description

This PR allows users to set `when` for custom modules to a boolean value that can permanently enable or disable the module without executing any shell commands.

- I added an `Either<A, B>` type and implemented `ModuleConfig` support for it
- I set the default value of `when` to from `None` to `false`
- I added support for custom modules in `ModuleRenderer`

#### Motivation and Context

Closes #2066
Closes #3601

#### Checklist:


docs/config/README.md
```diff
@@ -2561,7 +2561,7 @@ If you have an interesting example not covered there, feel free to share it ther
 | Option        | Default                       | Description                                                                                                                |
 | ------------- | ------------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
 | `command`     |                                 | The command whose output should be printed. The command will be passed on stdin to the shell.                              |
-| `when`        |                                 | A shell command used as a condition to show the module. The module will be shown if the command returns a `0` status code. |
+| `when`        | `false`                         | Either a boolean to permanently enable or disable the module or a string shell command used as a condition to show the module. The module will be shown if the command returns a `0` status code. |
```
jankatins, line 2564: At least for me just "boolean" makes it not completed clear if case matter or not (true, True, TRUE).

davidkna: It refers to TOML booleans. They aren't really explained for `disabled`/`enabled` either but I suppose it might make sense to mention it here.

jankatins: Fun, so `"true"` would still refer to the `true` command?

davidkna: Yes.

docs/config/README.md
```diff
@@ -2561,7 +2561,7 @@ If you have an interesting example not covered there, feel free to share it ther
 | Option        | Default                       | Description                                                                                                                |
 | ------------- | ------------------------------- | -------------------------------------------------------------------------------------------------------------------------- |
 | `command`     |                                 | The command whose output should be printed. The command will be passed on stdin to the shell.                              |
-| `when`        |                                 | A shell command used as a condition to show the module. The module will be shown if the command returns a `0` status code. |
+| `when`        | `false`                         | Either a boolean value (`true` or `false`) or a string shell command used as a condition to show the module. In case of a string, the module will be shown if the command returns a `0` status code. |
```
mickimnet, line 2564: @davidkna, do you think adding _"(`true` or `false`, without quotes)_" would make it more obvious, especially in regards to the comment about `"true"` would still run the command?

PR #2187, opened 2021-01-21
perf(zig): Lazy eval zig

HansAuger:
#### Description
Evaluate format call of zig lazily, update format string

#### Motivation and Context

Relates to: #2111

#### Checklist:

### DIFF
--- a/src/modules/zig.rs
+++ b/src/modules/zig.rs
@@ -33,11 +33,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             })
             .map(|variable| match variable {
                 "version" => {
-                    let zig_version_output = utils::exec_cmd("zig", &["version"])?
-                        .stdout
-                        .trim()
-                        .to_string();
-                    let zig_version = format!("v{}", zig_version_output);
+                    let zig_version_output = utils::exec_cmd("zig", &["version"])?.stdout;
+                    let zig_version = format!("v{}", zig_version_output.trim());
                     Some(Ok(zig_version))
                 }
                 _ => None,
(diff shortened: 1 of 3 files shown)


src/modules/zig.rs
```diff
@@ -38,7 +37,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 _ => None,
             })
             .map(|variable| match variable {
-                "version" => Some(Ok(zig_version.clone())),
+                "version" => {
+                    let zig_version = format!("v{}", zig_version_output);
```
andytom, line 41: I think you also need to move the `exec_cmd ` command from line 21 to here.

HansAuger: True. Done!

Also fixed the clippy complaint.

src/modules/zig.rs
```diff
@@ -38,7 +32,14 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 _ => None,
             })
             .map(|variable| match variable {
-                "version" => Some(Ok(zig_version.clone())),
+                "version" => {
+                    let zig_version_output = utils::exec_cmd("zig", &["version"])?
+                        .stdout
+                        .trim()
+                        .to_string();
+                    let zig_version = format!("v{}", zig_version_output);
```
davidkna, line 40: Tiny optimisation:

```suggestion
                    let zig_version_output = utils::exec_cmd("zig", &["version"])?.stdout;
                    let zig_version = format!("v{}", zig_version_output.trim());
```

andytom (approved):

PR #2198, opened 2021-01-23
feat(directory): Add home directory symbol

### LINKED ISSUE #1174: Allow us to modify home directory symbol (~) in [directory] options

toti1212:
#### Description

As another prompt libraries, could be good if we can modify and/or add a symbol when the current directoy is home (~). I added the option to enable/disable the symbol (`show_home_symbol`) and also an option to customize it (`home_symbol`), 🏠   by default 

Fixes #1174

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -656,6 +656,8 @@ it would have been `nixpkgs/pkgs`.
 | `read_only`         | `"🔒"`                                             | The symbol indicating current directory is read only.                            |
 | `read_only_style`   | `"red"`                                            | The style for the read only symbol.                                              |
 | `truncation_symbol` | `""`                                               | The symbol to prefix to truncated paths. eg: "…/"                                |
+| `home_symbol_enable`  | `true`                                           | Enable to use a symbol when the path is home instead of "~"                      |
+| `home_symbol`       | `"🏠"`                                             | The symbols indicating home directory.                                           |
 
 <details>
 <summary>This module has a few advanced configuration options that control how the directory is displayed.</summary>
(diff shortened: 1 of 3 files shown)


mickimnet:
Did you take a look at #1209, which was based of #1174 ?

toti1212:
Sorry, but I hadn't seen #1209  😔 . This approach is simpler and doesn't take subdirectories into account, only the ~ directory. But, since there is some work in progress, feel free to close this RP if you want and we can continue to contribute on the other two.

andytom (commented):
I think there might be a simpler way to implement this. We currently have [`HOME_SYMBOL`](https://github.com/starship/starship/blob/master/src/modules/directory.rs#L19) defined as a constant, I think if you just make that configurable with the default being `~` and it would be enough to resolve this.

toti1212:
You are right @andytom . I took your comments into account and I updated the solution. But, as I said, if this PR looks good, welcome! But if not, let's continue with the other two! 🙂

andytom (approved):
LGTM.

andytom:
Thank you for your contribution @toti1212

toti1212:
You are welcome! Keep it up guys!

PR #2200, opened 2021-01-23
refactor(memory_usage): replace sys-info with custom code

davidkna:
#### Description

This PR gets rid of `sys-info` and instead uses custom code to query the system about memory information. The code uses some `unsafe` but it should be safer than `sys-info` which is mostly implemented in C.

The code for Linux and Windows is relatively straightforward.
The macOS code turned out to be more trouble than expected because the `mach` crate doesn't have `mach_host_self` or `host_statistics64`, so I added a `bindgen` step to `build.rs` to generate bindings on the fly. If preferred I could also switch to static bindings for that instead, but `bindgen` doesn't appear to have much of an impact on build times.

This PR drops support for FreeBSD/OpenBSD for this module but I'll likely add support for it later because it should be similar to the macOS code. I think this PR should solve the Windows aarch64 build and the FreeBSD build in cross.

#### Motivation and Context

Closes #

#### Checklist:


davidkna:
It looks like `process_control` doesn't support FreeBSD so I'll postpone that work until that's fixed.

PR #2205, opened 2021-01-25
test(docker_context): Add basic tests to module

andytom:
#### Description

This adds some basic tests to the docker_context module. Have also fixed the docs to match the default config. 

#### Motivation and Context

Closes #2133

#### Checklist:

### DIFF
--- a/src/modules/docker_context.rs
+++ b/src/modules/docker_context.rs
@@ -77,3 +77,186 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         _ => None,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::test::ModuleRenderer;
+    use ansi_term::Color;
+    use std::fs::File;
+    use std::io::{self, Write};
+
+    #[test]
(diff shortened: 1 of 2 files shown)


src/modules/docker_context.rs
```diff
@@ -77,3 +77,186 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+#[cfg(test)]
+mod tests {
+    use crate::test::ModuleRenderer;
+    use ansi_term::Color;
+    use std::fs::File;
+    use std::io::{self, Write};
+
+    #[test]
+    #[ignore]
+    fn only_trigger_when_docker_config_exists() -> io::Result<()> {
```
davidkna, line 90: Is there any reason you can't un-ignore this test and do it in an empty temporary directory?

andytom: The only reason is that I forgot when I was testing locally, have fixed that now.

davidkna (approved):

andytom:
Thanks for reviewing @davidkna

PR #2211, opened 2021-01-26
docs: fix the format of status.format

matoruru:
#### Description

I edited based on this:
https://github.com/starship/starship/blob/3c31c4b99d509ca0ca06bf9b4addbea1259cdac9/src/configs/status.rs#L22

#### Motivation and Context

This was not working when I just copy and paste this part in my `starship.toml`.

#### Checklist:


mickimnet:
LGTM.

andytom:
@matoruru thank you for the fix and @mickimnet thanks for reviewing.

matoruru:
Not at all!

PR #2228, opened 2021-01-28
feat(install): make install script posix compliant

### LINKED ISSUE #2226: feat(install): install script should not require bash
The install.sh script currently requires bash. Not all linux distributions include bash by default and there is no reason the script could not be posix compliant (it looks like it mostly is already).

Make the install script posix complaint and use /usr/bin/env sh for the shebang

Installing bash as a prerequisite works, but should not be necessary

NOTE:

I am happy to take on this work and will submit a PR if the community / maintainers agree.

dmccaffery:
#### Description

* update shebang to support posix-compliant shells (like dash)
* replace all usages of echo with printf (to normalise)
* remove locals, which are not posix complaint
* rename `command` variable to `fetch_cmd` as it overloads a posix builtin
* rename `complete` function to `completed` as it overloads the bash builtin for completion

#### Motivation and Context

Bash should not be necessary to run the install script on systems that would otherwise not require it. Ubuntu uses dash by default and I use zsh, so no reason to install bash before installing starship. :)

Closes #2226 

Ran install scripts on macOS natively, as well as the following docker containers for linux distributions (not the same as native, but as close as I can get with ease):

* debian
* fedora
* centos
* linuxmint
* ubuntu

#### Checklist:

- [ ] (not relevant) I have updated the documentation accordingly.
- [ ] (not relevant) I have updated the tests accordingly.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -23,7 +23,6 @@
 #     Override the base URL used for downloading releases
 
 set -eu
-set -o pipefail 2>/dev/null || true
 printf '\n'
 
 BOLD="$(tput bold 2>/dev/null || printf '')"
--- a/install/install.sh
+++ b/install/install.sh
@@ -199,7 +199,7 @@ detect_platform() {
     freebsd) platform="unknown-freebsd" ;;
   esac
 
-  printf '%s\n' "${platform}"
+  printf '%s' "${platform}"
 }
 
 # Currently supporting:
--- a/install/install.sh
+++ b/install/install.sh
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env sh
+
+# shellcheck disable=SC2039
 
 # Options
 #
@@ -20,18 +22,19 @@
 #   -B, --base-url
 #     Override the base URL used for downloading releases
 
-set -euo pipefail
-printf "\n"
+set -eu
+set -o pipefail 2>/dev/null || true
+printf '\n'
 
-BOLD="$(tput bold 2>/dev/null || echo '')"
-GREY="$(tput setaf 0 2>/dev/null || echo '')"
-UNDERLINE="$(tput smul 2>/dev/null || echo '')"
-RED="$(tput setaf 1 2>/dev/null || echo '')"
-GREEN="$(tput setaf 2 2>/dev/null || echo '')"
-YELLOW="$(tput setaf 3 2>/dev/null || echo '')"
-BLUE="$(tput setaf 4 2>/dev/null || echo '')"
-MAGENTA="$(tput setaf 5 2>/dev/null || echo '')"
-NO_COLOR="$(tput sgr0 2>/dev/null || echo '')"
+BOLD="$(tput bold 2>/dev/null || printf '')"
+GREY="$(tput setaf 0 2>/dev/null || printf '')"
+UNDERLINE="$(tput smul 2>/dev/null || printf '')"
+RED="$(tput setaf 1 2>/dev/null || printf '')"
+GREEN="$(tput setaf 2 2>/dev/null || printf '')"
+YELLOW="$(tput setaf 3 2>/dev/null || printf '')"
+BLUE="$(tput setaf 4 2>/dev/null || printf '')"
+MAGENTA="$(tput setaf 5 2>/dev/null || printf '')"
+NO_COLOR="$(tput sgr0 2>/dev/null || printf '')"
 
 SUPPORTED_TARGETS="x86_64-unknown-linux-gnu x86_64-unknown-linux-musl \
                   i686-unknown-linux-musl aarch64-unknown-linux-musl \
@@ -41,19 +44,19 @@ SUPPORTED_TARGETS="x86_64-unknown-linux-gnu x86_64-unknown-linux-musl \
                   x86_64-unknown-freebsd"
 
 info() {
-  printf "%s\n" "${BOLD}${GREY}>${NO_COLOR} $*"
+  printf '%s\n' "${BOLD}${GREY}>${NO_COLOR} $*"
 }
 
 warn() {
-  printf "%s\n" "${YELLOW}! $*${NO_COLOR}"
+  printf '%s\n' "${YELLOW}! $*${NO_COLOR}"
 }
 
 error() {
-  printf "%s\n" "${RED}x $*${NO_COLOR}" >&2
+  printf '%s\n' "${RED}x $*${NO_COLOR}" >&2
 }
 
-complete() {
-  printf "%s\n" "${GREEN}✓${NO_COLOR} $*"
+completed() {
+  printf '%s\n' "${GREEN}✓${NO_COLOR} $*"
 }
 
 # Gets path to a temporary file, even if
@@ -82,17 +85,17 @@ test_writeable() {
 }
 
 fetch() {
-  local command
+  local fetch_cmd
   if hash curl 2>/dev/null; then
     set +e
-    command="curl --silent --fail --location $1"
+    fetch_cmd="curl --silent --fail --location $1"
     curl --silent --fail --location "$1"
     rc=$?
     set -e
   else
     if hash wget 2>/dev/null; then
       set +e
-      command="wget -O- -q $1"
+      fetch_cmd="wget -O- -q $1"
       wget -O- -q "$1"
       rc=$?
       set -e
@@ -104,12 +107,12 @@ fetch() {
 
   if [ $rc -ne 0 ]; then
     printf "\n" >&2
-    error "Command failed (exit code $rc): ${BLUE}${command}${NO_COLOR}"
+    error "Command failed (exit code $rc): ${BLUE}${fetch_cmd}${NO_COLOR}"
     printf "\n" >&2
     info "This is likely due to Starship not yet supporting your configuration." >&2
     info "If you would like to see a build for your configuration," >&2
     info "please create an issue requesting a build for ${MAGENTA}${ARCH}-${PLATFORM}${NO_COLOR}:" >&2
-    info "${BOLD}${UNDERLINE}https://github.com/starship/starship/issues/new/${NO_COLOR}\n" >&2
+    info "${BOLD}${UNDERLINE}https://github.com/starship/starship/issues/new/${NO_COLOR}" >&2
     exit $rc
   fi
 }
@@ -189,7 +192,7 @@ detect_platform() {
     freebsd) platform="unknown-freebsd" ;;
   esac
 
-  echo "${platform}"
+  printf '%s\n' "${platform}"
 }
 
 # Currently supporting:
@@ -212,7 +215,7 @@ detect_arch() {
     arch=arm
   fi
 
-  echo "${arch}"
+  printf '%s' "${arch}"
 }
 
 detect_target() {
@@ -224,7 +227,7 @@ detect_target() {
     target="${target}eabihf"
   fi
 
-  echo "${target}"
+  printf '%s' "${target}"
 }
 
 
@@ -261,7 +264,7 @@ check_bin_dir() {
     IFS=:
     for path in $PATH; do
       if [ "${path}" = "${bin_dir}" ]; then
-        echo 1
+        printf 1
         break
       fi
     done
@@ -278,12 +281,12 @@ is_build_available() {
   local target="$3"
 
   local good
-  
+
   good=$(
     IFS=" "
     for t in $SUPPORTED_TARGETS; do
-      if [ "${t}" == "${target}" ]; then
-        echo 1
+      if [ "${t}" = "${target}" ]; then
+        printf 1
         break
       fi
     done
@@ -395,7 +398,7 @@ else
   VERBOSE=
 fi
 
-echo
+printf '\n'
 
 EXT=tar.gz
 if [ "${PLATFORM}" = "pc-windows-msvc" ]; then
@@ -408,9 +411,9 @@ confirm "Install Starship ${GREEN}latest${NO_COLOR} to ${BOLD}${GREEN}${BIN_DIR}
 check_bin_dir "${BIN_DIR}"
 
 install
-complete "Starship installed"
+completed "Starship installed"
 
-echo
+printf '\n'
 info "Please follow the steps for your shell to complete the installation:
 
   ${BOLD}${UNDERLINE}Bash${NO_COLOR}
--- a/install/install.sh
+++ b/install/install.sh
@@ -58,11 +58,15 @@ completed() {
   printf '%s\n' "${GREEN}✓${NO_COLOR} $*"
 }
 
+has() {
+  command -v "$1" 1>/dev/null 2>&1
+}
+
 # Gets path to a temporary file, even if
 get_tmpfile() {
   local suffix
   suffix="$1"
-  if hash mktemp; then
+  if has mktemp; then
     printf "%s.%s" "$(mktemp)" "${suffix}"
   else
     # No really good options here--let's pick a default + hope
@@ -85,14 +89,14 @@ test_writeable() {
 
 fetch() {
   local fetch_cmd
-  if hash curl 2>/dev/null; then
+  if has curl; then
     set +e
     fetch_cmd="curl --silent --fail --location $1"
     curl --silent --fail --location "$1"
     rc=$?
     set -e
   else
-    if hash wget 2>/dev/null; then
+    if has wget; then
       set +e
       fetch_cmd="wget -O- -q $1"
       wget -O- -q "$1"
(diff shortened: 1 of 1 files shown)


davidkna:
I don't really see the value of removing `local` just for the sake of POSIX compliance, because I can't think of any shell that doesn't support it. Even `busybox` has support for it. ([More elaborate comment on stackoverflow](https://stackoverflow.com/a/18600920)).

install/install.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env sh
```
davidkna, line 1: Is the space here intentional? I think some systems could have issues with that.

install/install.sh
```diff
@@ -20,18 +20,18 @@
 #   -B, --base-url
 #     Override the base URL used for downloading releases
 
-set -euo pipefail
```
davidkna, line 23: Isn't pipefail also nearly universally supported? It might be a bit worse than `local`, I guess.

davidkna: I think `set -o pipefail 2>/dev/null || true` in a separate line should be POSIX compliant.

dmccaffery:
Agreed. I removed them because I don't think dash supports it (default shell in Ubuntu), but will verify.

That being said, they are only global in process and the install script is invoked; not evaluated.

install/install.sh
dmccaffery: It was intentional and I always do it, but I never knew it could be an issue. Apparently it's a great debate: https://web.archive.org/web/20170215070707/http://www.in-ulm.de/~mascheck/various/shebang/#details

I have no issue with removing the space; and will do!

install/install.sh
dmccaffery: I can add this; it should be okay and will verify. It does fail in dash.

install/install.sh
dmccaffery: done

install/install.sh
dmccaffery: done

install/install.sh
dmccaffery: `set -o pipefail` immediately exists regardless of any attempt to ignore the exit code with `|| true` in dash... simple test:

```sh
temp=$(mktemp -d)
cat << EOT > "$temp"/with-pipefail.sh
#! /usr/bin/env sh

set +e
printf "%s\n" "attempting to set pipefail"
set -o pipefail 2>/dev/null || true
printf "%s\n" "we got here"
EOT

chmod +x "$temp"/with-pipefail.sh

docker run --volume="$temp":/test debian /test/with-pipefail.sh
rm -rf "$temp"
```

dmccaffery:
I just realised that modern versions of curl will fail with binary files when not used with a pipe or `--output`. I tried to use `--output -` to no avail; not sure how to resolve without dramatically modifying the fetch logic to not write to the stdout buffer.

vladimyr:
I guess that is time to close #1315

davidkna:
@vladimyr Maybe. I recently duplicated a lot of your changes from there so merge conflicts are likely significant. But if you still want to get any of the other changes from that PR merged please rebase it.

@dmccaffery Would you please port over `download` and `unpack` from #1315 to make sure error handling without `pipefail` still works?

dmccaffery:
@davidkna thanks and will do -- apologies for the delay; I've been busy.

@vladimyr -- I'm so sorry I didn't see your PR before opening this one. I did a search for an issue but didn't think to scan current PRs!

dmccaffery:
@davidkna I made the requested changes and also added a new GitHub Actions workflow for testing the install across a wide array of environments, including Windows WSL. This is isolated on the last commit in case you do not want to incorporate it.

Let me know if there is anything else you need!

.github/workflows/install.yml
```diff
@@ -0,0 +1,116 @@
...
+jobs:
+  check:
+    name: check
+    runs-on: ubuntu-20.04
+    container: koalaman/shellcheck-alpine:stable
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: shell
+        run: shellcheck -e SC1090,SC1091,SC1071 install/install.sh
```
davidkna, line 19: There's a comment in `install.sh` to disable `SC2039`. Why are these warning disabled  only here?

install/build.sh
```diff
@@ -0,0 +1,76 @@
...
+	--build-arg IMAGE="$IMAGE" \
+	--label=org.opencontainers.image.name=starship \
+	--label=org.opencontainers.image.revision="$GITHUB_SHA" \
+	--label=org.opencontainers.image.version="$GITHUB_REF" \
+	--label=org.opencontainers.image.source="https://github.com/$GITHUB_REPOSITORY/starship" \
+	--tag "ghcr.io/$GITHUB_REPOSITORY/starship:$TAG" \
+	"${PWD}/install"
+
+    # this can be added if images are ever published
+	# --cache-from=type=registry,ref="ghcr.io/$GITHUB_REPOSITORY/starship:$IMAGE-latest" \
```
davidkna, line 75: Is there a use-case to distribute Starship in a container?

davidkna (changes requested):

davidkna:
I think it might be better to move the testing changes to a separate PR so this can be merged quickly.

install/build.sh
dmccaffery: Not that I can think of except to be able to quickly pull one to reproduce an issue someone reports with a specific flavour combination. I can't think of any reason beyond that.

.github/workflows/install.yml
dmccaffery: 1090 and 1091 have to do with spellcheck being able to follow sourced or exec'd scripts, which it can't do when things are executed from a url, download, or uncompressed asset. These could be moved into the script, but it's fairly common to disable these for scripts that bootstrap other scripts over the wire.

1071 isn't actually necessary here, but it "ignores" scripts that shellcheck can't lint based on the shebangs (like zsh, for example). We don't have any of those in this use case.

It's a style choice and I'm happy to go either way.

install/install.sh
```diff
@@ -104,41 +110,69 @@ fetch() {
...
+    return 1
   fi
+
+  $cmd && return 0 || rc=$?
+
+  error "Command failed (exit code $rc): ${BLUE}${cmd}${NO_COLOR}"
+  printf "\n" >&2
+  info "This is likely due to Starship not yet supporting your configuration."
+  info "If you would like to see a build for your configuration,"
+  info "please create an issue requesting a build for ${MAGENTA}${arch}-${platform}${NO_COLOR}:"
```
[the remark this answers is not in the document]
dmccaffery: Yes. Will do.

.github/workflows/install.yml
```diff
@@ -0,0 +1,116 @@
...
+          - ubuntu-16.04
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: install
+        run: ./install/install.sh --force
+      - name: test
+        run: |
+          starship --version
+          command -v starship
```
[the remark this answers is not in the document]
dmccaffery: Can do. Prefer this to all be containerised, or on Ubuntu?

dmccaffery:
Agreed. I'll pull the testing out and open a separate PR.

.github/workflows/install.yml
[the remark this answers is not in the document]
davidkna: Just ubuntu is fine but macOS would be nice too, especially given the old bash version it ships with.

.github/workflows/install.yml
davidkna: In that case a comment in your future PR would be nice.

install/install.sh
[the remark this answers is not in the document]
dmccaffery: done

dmccaffery:
@davidkna : removed the install tests, which have been reopened in a WIP/draft PR for now: #2406 -- also made the requested change to the download function, so should be ready for a final review. Thanks!

install/install.sh
```diff
@@ -189,7 +199,7 @@ detect_platform() {
     freebsd) platform="unknown-freebsd" ;;
   esac
 
-  echo "${platform}"
+  printf '%s\n' "${platform}"
```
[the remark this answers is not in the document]
dmccaffery: echo emits a newline, which is why I kept it in printf -- has no side effects to change, though so accepted recommendation.

davidkna (approved):
LGTM

chipbuster:
@dmccaffery Thank you for your contributions, and thanks @davidkna for reviewing!

vladimyr:
> @vladimyr -- I'm so sorry I didn't see your PR before opening this one. I did a search for an issue but didn't think to scan current PRs!

@dmccaffery No worries I didn't post my comment to criticize but to spare you some time dealing with conversion. I hope it was helpful and you found useful stuff over there :crossed_fingers: 

> @vladimyr Maybe. I recently duplicated a lot of your changes from there so merge conflicts are likely significant.

@davidkna Let's put it this way. I'm in no way emotionally attached to my code but what is the point of duplicating? You could simply continue on top of it or at least copy it completely. Doing neither leads to the omission of important stuff like:

> - support `PREFIX` environment variable
> 
> Added support for `PREFIX` environment variable which enables the following _standardish_ invocation style:
> ```sh
> curl -fsSL https://starship.rs/install.sh | PREFIX=$HOME/.local sh
> # starship ends up inside $HOME/.local/bin
> ```

discussed in detail here: https://github.com/starship/starship/pull/1315#discussion_r439212699

But yeah, you gave me an _interesting_ option:

> But if you still want to get any of the other changes from that PR merged please rebase it.

:zipper_mouth_face: ...

Let's get back to the technical track...

@davidkna said:
> I don't really see the value of removing `local` just for the sake of POSIX compliance, because I can't think of any shell that doesn't support it. Even `busybox` has support for it. ([More elaborate comment on stackoverflow](https://stackoverflow.com/a/18600920)).

1. SO user literally mentioned `ksh` and non-GNU/non-BSD systems as potential deal-breakers:
	  > So unless you're aiming to support **non-GNU non-BSD systems** like Solaris, or those using **standard ksh**, etc., **you could get away with using local**. (Might want to put some comment right at the start of the script, below the shebang line, noting that it is not strictly a POSIX sh script. Just to be not evil.)
	  
	  Even though I do agree it is very unlikely to get a bug report from the latter I wouldn't be so surprised to see a `ksh` user around.

2. The fact that you _could get away with it_ doesn't mean you should! To prove my point let me show you Docker's install script: https://get.docker.com Try to find some `local`s inside :wink: 

3. There are reasons why `local` isn't standardized by POSIX and while following that [debian.org discussion link](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=381091) gives you some insight here is the link to the actual Open Group bug tracker https://www.austingroupbugs.net/bug_view_page.php?bug_id=767 
	Once you scratch under the surface you figure out that the answer to _Are there any shells out there that don't support `local`?_ is not that important because that is not the right question to ask in the first place. As this SE user pointed out:
	> It's not as simple as supporting `local` or not.
	
	https://unix.stackexchange.com/questions/493729/list-of-shells-that-support-local-keyword-for-defining-local-variables
	
4. Regardless of what you end up doing disabling SC2039 is **wrong** thing to do! First of all, it is an umbrella warning so you just disabled a whole bunch of other helpful warnings plus that is the exact reason why it is deprecated:
	> _Note: This warning has been retired in favor of individual SC3xxx warnings for each individual issue._
	
	https://github.com/koalaman/shellcheck/wiki/SC2039 Unfortunately, that change is relatively new (https://github.com/koalaman/shellcheck/wiki/SC2039/_compare/e241394...6c9fa14 & https://github.com/koalaman/shellcheck/wiki/SC2039/6c9fa14) so those `SC3xxx` rules haven't been implemented yet. So you either live with shellcheck yelling at you or you trade safety for commodity.
	
Finally, regardless of all that thanks both of you for your contribution and for moving things forward! :+1:

vladimyr:
FWIW Once the new ShellCheck build gets released we could disable `SC3043` only :tada: 
https://github.com/koalaman/shellcheck/blob/81acea0/src/ShellCheck/Checks/ShellSupport.hs#L339-L345

vladimyr:
> I wouldn't be so surprised to see a `ksh` user around.

To further strengthen my argument about `ksh`'s `local` support let's take a look at the following. I have Termux installed on my phone and I'm using `zsh` with starship OFC:
```
~
❯ echo $ZSH_VERSION
5.8

~
❯ type local
local is a reserved word
```

Switching quickly to `bash` gives the following output:
```
~ $ echo $BASH_VERSION
5.1.4(1)-release
~ $ type local
local is a shell builtin
```

But Android (11) ships with [`mksh` (MirBSD Korn Shell)](http://www.mirbsd.org/mksh.htm) (R57): https://android.googlesource.com/platform/external/mksh/+/refs/heads/android11-release/src/sh.h#187
So let's switch again to `/system/bin/sh` to check its `local` support:
```
P2a42:/ $ echo $KSH_VERSION
@(#)MIRBSD KSH R52 2016/01/20
P2a42:/ $ type local
local is an alias for '\typeset'
```

Apart from the fact it reveals I'm using a bit older Android OS/device :upside_down_face:, this is still a fairly modern `ksh` clone and yet `local` is defined as an alias! :warning:

davidkna:
> What is the point of duplicating?
 
I wasn't aware of your PR.

> I wouldn't be so surprised to see a `ksh` user around.

For the use here (locals in functions) it looks like the implementation differences don't matter at all. The only issue is standard ksh which doesn't support `local` in posix functions and I agree it would be worthwhile to support that shell.

dmccaffery:
@davidkna / @vladimyr :

Should I remove the locals again? I don't think it will have any side effects. No one should be sourcing the install script. When executed, the scope is the forked process of the sub shell, so there is no need for locals anyway (as far as I know).

vladimyr:
> I wasn't aware of your PR.

Ah, too bad you missed it. However, one way or another installer is finally (almost) POSIX compliant :tada: 

> The only issue is standard ksh which doesn't support `local` in posix functions and I agree it would be worthwhile to support that shell.


> Should I remove the locals again? I don't think it will have any side effects. No one should be sourcing the install script. When executed, the scope is the forked process of the sub shell, so there is no need for locals anyway (as far as I know).

It seems we have a consensus here. `ksh` is a worthwhile target to support and, as @dmccaffery nicely explained, ensuring local scope isn't that crucial under terms of preferred/advertised use case. So, @dmccaffery I'm sorry for expanding your todo list :see_no_evil: but feel free to remove locals :wink:

dmccaffery:
... sure, but imma smack you. 😈

vladimyr:
As of writing this comment Starship's install instructions (https://github.com/starship/starship/blob/4925764/docs/README.md#install-latest-version) suggest the following:
```sh
curl -fsSL https://starship.rs/install.sh | bash
```

First, this needs to get changed to (at least):
```sh
curl -fsSL https://starship.rs/install.sh | sh
```
in order to advertise reaching POSIX compliance to the end-users.

IMHO there are 4 likely types of workflows users may follow to install Starship using the installer script:

1. **simple (c/p) workflow** (most likely)
	In this workflow, the value of shebang is ignored as script content is piped directly to the (hopefully POSIX compatible) `sh`.

2. **advanced workflow**
	In this workflow, the user makes sure that the script is fully downloaded before starting its execution to prevent possible [timing attacks](https://unix.stackexchange.com/a/339276).
	```sh
	sh -c "$(curl -fsSL https://starship.rs/install.sh)"
	```
	Shebang is ignored in this case because the desired shell (`sh`) is explicitly invoked.
	**It is worth noting that this form should be preferred over the currently advertised one because it is equally convenient (c/p-able) but more secure.** For example, take a look at Homebrew installation docs: https://brew.sh 

3. **cautious workflow**
	In this workflow, the user first downloads the installer script:
	```sh
	curl -fsSL https://starship.rs/install.sh -o install.sh
	```
	They then verify its contents, and eventually run it:
	
	- by passing it as shell parameter
		```sh
		sh install.sh
		```
		Shebang is ignored in this case because the desired shell (`sh`) is explicitly invoked.
		**Again, it is worth noting that some projects prefer writing out these steps in detail instead of advertising simple c/p-able shell snippets thus encouraging good safety practices!** For example, here is how Docker folks do it: https://docs.docker.com/engine/install/debian/#install-using-the-convenience-script
	
	- as if it's an executable
		```sh
		chmod +x install.sh
		./install.sh
		```
		In this case, the program loader parses the shebang and uses it to spawn a shell that'll be used for executing script contents.
	
So, only in 1 out 4 possible workflows shebang actually comes into the play. In all other workflows, it serves as a mere indicator of POSIX compliance so the user knows which shell to pick/invoke. In the former case, loader decides which shell will get invoked by parsing the shebang which brings me to my point about @dmccaffery's shebang choice https://github.com/starship/starship/blob/4925764/install/install.sh#L1:
```sh
!#/usr/bin/env sh
```
IMO it is somewhat superfluous. There is hardly any advantage in using `env` and looking up `sh` from the `PATH` instead of going with plain old `/bin/sh`. Yes, I'm aware that Solaris' `/bin/sh` is actually a Bourne Shell which is not POSIX compliant and what [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html#:~:text=Applications%20should%20note%20that%20the%20standard,(for%20example%2C%20with%20sed).%20For%20example%3A) says about it but bear in mind that:

1. it is part of **informative** section
2. it is safe to assume Solaris folks are typically advanced users and well aware of that quirk and they'll probably invoke the correct (POSIX compliant) shell directly anyway

Honestly, I never saw anyone suggesting the use of `env` for POSIX `sh` lookup but apparently [Wikipedia authors think it is a good idea](https://en.wikipedia.org/wiki/Shebang_(Unix)#Program_location:~:text=A%20similar%20problem%20exists%20for%20the,Character). I definitely don't share their opinion because while it makes it more foolproof it also makes it less secure too. There is no security without knowledge. If the user treats shebang as some sort of magic and rejoices at the perceived flexibility from `env` usage they are certainly also not aware that a simple tweak to their shell profile (specifically `PATH` value) together with the introduction of fake `sh` will end up with them having a very very bad day.

So, long story short, I'm proposing the following:

1. **Let's simplify things by using simple `/bin/sh` shebang.** (As we did before with `/bin/bash` too.)
2. **Let's strike a balance between safety and convenience** by advertising the following form instead:
	```sh
	sh -c "$(curl -fsSL https://starship.rs/install.sh)"
	```

PR #2237, opened 2021-01-29
ci: continue release on error

matchai:
#### Description

During the release process, if a build fails, the whole release process is thrown out the window.
This will cause the release to proceed with the artifacts that are correctly built.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -64,10 +64,11 @@ jobs:
             os: macOS-latest
             name: starship-x86_64-apple-darwin.tar.gz
           
-          # TODO: switch to macOS-latest once it reaches 11.0
-          - target: aarch64-apple-darwin
-            os: macos-11.0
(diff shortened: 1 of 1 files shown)


davidkna:
How did the build end up failing? Can't see the logs for it. I also just noticed the build cache is still enabled in the release workflow. Could that have been the cause?

davidkna (approved):

andytom:
> How did the build end up failing? Can't see the logs for it. I also just noticed the build cache is still enabled in the release workflow. Could that have been the cause?

It was a problem with the runner, macOS 11 is in private preview so only some organisations or users can use it and it doesn't seem like we are one of the lucky ones :disappointed:

davidkna:
Looks like I tested it when enrolment was open and they've closed it now.

matchai:
I've updated the PR to temporarily disable aarch64-apple-darwin builds.

andytom (approved):
LGTM

PR #2249, opened 2021-01-30
refactor: Changes to how String ownership is acquired in native types for faster compile times

wsb1994:
#### Description

All instances of to_string() operating on &str were changed to .to_owned() which is highly optimized for that use case; The compiler may *choose* to optimize to_string() in some cases, but this removes that choice and ensures that optimization.

#### Motivation and Context

to_owned() can in many cases become highly optimized compared to to_string(). It's an example of "compilers choice, vs always optimize". It used to *always* be better. I think this is no longer the case and made these changes under that assumption.

Below is some evidence of this being the case in the past, and I have also personally observed speed increases from using to_owned(), but it seems like that's not guaranteed to be the case anymore and is probably just a headache left over from yesteryear. I wouldn't worry too much about it. The changes still definitely improve compile time slightly, so just take that as you will as the consideration here.

https://medium.com/@ericdreichert/converting-str-to-string-vs-to-owned-with-two-benchmarks-a66fd5a081ce

TL;DR: Merge this if you like, but it seems like the newest update may render the changes moot, as the compiler seems to now prefer to generate the optimized assembly in many cases if not all of them.  You'll probably get about two seconds off your compile time at best with this change so that might still make it worth it if you value that sort of thing.

Results may vary.

All changes are drop in equivalent. They change a reference into a String. All tests still pass, no new tests are needed. The only difference is the increased compile-time performance, and possible micro-optimizations being guaranteed rather than possibly gained. 

In other words, nothing of importance was changed. Small benefits were gained. 

#### Checklist:

### DIFF
--- a/src/bug_report.rs
+++ b/src/bug_report.rs
@@ -144,22 +144,22 @@ fn get_shell_info() -> ShellInfo {
     let shell = std::env::var("STARSHIP_SHELL");
     if shell.is_err() {
         return ShellInfo {
-            name: UNKNOWN_SHELL.to_string(),
-            version: UNKNOWN_VERSION.to_string(),
-            config: UNKNOWN_CONFIG.to_string(),
+            name: UNKNOWN_SHELL.to_owned(),
+            version: UNKNOWN_VERSION.to_owned(),
+            config: UNKNOWN_CONFIG.to_owned(),
         };
     }
 
     let shell = shell.unwrap();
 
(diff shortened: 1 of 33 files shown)


wsb1994:
Leaving my opinion on this here:

This optimization no longer exists. The compile time benefit is still there, but it's too small to even notice. Therefore I'm closing this since it's not really that big of a deal in the grand scheme of things!

PR #2264, opened 2021-02-03
feat(python): Configure when the module is shown

andytom:
#### Description

This makes it possible to configure when the python module is shown
    
This includes a breaking change since we are removing the
`scan_for_pyfiles` configuration option in favour of setting the
`detect_extensions` to an empty array.

#### Motivation and Context

Related to #1977 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -388,7 +388,7 @@ can do this in two ways:
 
 By default it only changes color. If you also want to change it's shape take a
 look at [this example](#with-custom-error-shape).
- 
+
 ::: warning
 `error_symbol` is not supported on elvish shell.
 :::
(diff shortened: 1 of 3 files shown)


davidkna (approved):
LGTM

IevgenSobko:
Is it possible to update https://starship.rs/config/#python only with released features, not one in master but not released yet?
I found myself in the situation when documentation says to use this 
```
detect_extensions = []
```
but it is actually not released and in my 0.50.0 version doesn't work.
Thanks to other languages documentation I've found an old way of excluding extensions for python.

andytom:
> Is it possible to update https://starship.rs/config/#python only with released features, not one in master but not released yet?
> I found myself in the situation when documentation says to use this
> 
> ```
> detect_extensions = []
> ```
> 
> but it is actually not released and in my 0.50.0 version doesn't work.
> Thanks to other languages documentation I've found an old way of excluding extensions for python.

Thanks for bringing this up @IevgenSobko, this has been something I've been sort of aware of but not really thought about too much. I'll open a new issue for this.

PR #2268, opened 2021-02-04
feat(cmd_duration): Add notify display time

### LINKED ISSUE #2194: Increase cmd_duration notification duration

The duration of the cmd_duration notification is too short, so I barely have enough time to read it.

There should be an option to set how long the notification is up for.

Don't force a specific duration and just let the user's notification daemon keep it for the default notification duration.

Lucky1313:
#### Description
Adds a notify display duration configuration option to the cmd_duration module. 

#### Motivation and Context

Closes #2194
Closes #3515

It turns out this is actually difficult to test, because, to quote the `notify-rust` docs, notification timeout is [often not respected by server, sorry.](https://docs.rs/notify-rust/4.2.2/notify_rust/struct.Notification.html#structfield.timeout) (which it turns out my system doesn't)
I think this is straightforward enough, but would appreciate others who can test.

#### Checklist:

### DIFF
--- a/src/modules/cmd_duration.rs
+++ b/src/modules/cmd_duration.rs
@@ -101,6 +101,7 @@ fn undistract_me<'a, 'b>(
 ) -> Module<'a> {
     use ansi_term::{unstyle, ANSIStrings};
     use notify_rust::{Notification, Timeout};
+    use std::convert::TryFrom;
 
     if config.show_notifications && config.min_time_to_notify as u128 <= elapsed {
         let body = format!(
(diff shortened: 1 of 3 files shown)


src/modules/cmd_duration.rs
```diff
@@ -101,22 +101,27 @@ fn undistract_me<'a, 'b>(
...
-        let mut notification = Notification::new();
-        notification
-            .summary("Command finished")
-            .body(&body)
-            .icon("utilities-terminal")
-            .timeout(Timeout::Milliseconds(750));
-
-        if let Err(err) = notification.show() {
-            log::trace!("Cannot show notification: {}", err);
+        if let Ok(display_time) = u32::try_from(config.notify_display_time) {
```
andytom, line 112: Would it be better to try to parse the timeout and if it fails log an error/warning and then fall back to the default of 750?

D4N: Depends, not showing the notification at all might prompt the user to go looking for issues sooner.

PR #2270, opened 2021-02-05
perf(nodejs): Change `⬢` to `🐢`

LitoMore:
#### Motivation and Context

1. Other languages use emoji as icons, only Node.js does not.
2. The `⬢` is smaller than other emoji icons, and they are not in the same baseline, it looks so weird.
3. The `🐢` came from [Node.js GitHub description](https://github.com/nodejs/node).

#### Screenshots:

Before (iTerm on macOS)

image

After (iTerm on macOS)

image

Before (Windows Terminal with WSL)

image

After (Windows Terminal with WSL)

image

#### Checklist:


chipbuster:
As far as I can tell, the GitHub repo description for nodejs is the only place where the turtle emoji shows up--it certainly doesn't look like it's associated with the language normally (unlike the crab for Rust and the camel for perl)

I'm certainly not in the nodejs community, but I certainly wouldn't guess NodeJS for a turtle, and I'm worried some users might think we're mocking NodeJS's speed.

Is there somewhere else where the association of the turtle and Node are well-established? Alternatively, we've started to explore using Nerd Fonts for language symbols--is there a NerdFont symbol that might work for this?

andytom:
I am going to close this since #1649 has been merged, so I am not sure this is still needed.

PR #2271, opened 2021-02-05
feat: abbreviate package.json semantic versions

tbjgolden:
#### Description

npm packages often use semantic release as a tool to automate versioning, by having a program look at conventional commit messages, like those on this repo.

This PR abbreviates long versions like `v0.0.0-semantically-released` to `semantic`

#### Motivation and Context

For npm packages, a somewhat common pattern is using semantic release, which allows a ci tool (e.g. github actions) to automatically determine versions based on (a conventional) commit history. This means that the correct version isn't actually in the package.json file.
Normally, you'll see a version like 0.0.0-semantically-released in the package.json, to indicate that the package version is derived elsewhere. In it's current implementation, this is awkward because it conveys little meaning, uses quite a bit of space, and doesn't actually let you know the version

#### Screenshots:

Screenshot_20210205_155513

#### Checklist:

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -68,7 +68,8 @@ fn extract_package_version(file_contents: &str, display_private: bool) -> Option
     };
 
     let formatted_version = format_version(raw_version);
-    if formatted_version == "v0.0.0-development" || formatted_version.find("semantic").is_some() {
+    if formatted_version == "v0.0.0-development" || formatted_version.find("semantic").is_some()
+    {
         return Some("semantic".to_string());
     };
 
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -68,7 +68,7 @@ fn extract_package_version(file_contents: &str, display_private: bool) -> Option
     };
 
     let formatted_version = format_version(raw_version);
-    if formatted_version.ends_with("-semantically-released") {
+    if formatted_version == "v0.0.0-development" || formatted_version.find("semantic").is_some() {
         return Some("semantic".to_string());
     };
 
@@ -348,17 +348,47 @@ mod tests {
     }
 
     #[test]
-    fn test_package_version_with_semantic_version() -> io::Result<()> {
+    fn test_package_version_semantic_development_version() -> io::Result<()> {
         let config_name = "package.json";
         let config_content = json::json!({
             "name": "starship",
-            "version": "0.0.0-semantically-released"
+            "version": "0.0.0-development"
         })
         .to_string();
 
         let project_dir = create_project_dir()?;
         fill_config(&project_dir, config_name, Some(&config_content))?;
         expect_output(&project_dir, Some("semantic"), None)?;
+        project_dir.close()
+    }
+
+    #[test]
+    fn test_package_version_with_semantic_other_version() -> io::Result<()> {
+        let config_name = "package.json";
+        let config_content = json::json!({
+            "name": "starship",
+            "version": "v0.0.0-semantically-released"
+        })
+        .to_string();
+
+        let project_dir = create_project_dir()?;
+        fill_config(&project_dir, config_name, Some(&config_content))?;
+        expect_output(&project_dir, Some("semantic"), None)?;
+        project_dir.close()
+    }
+
+    #[test]
+    fn test_package_version_with_non_semantic_tag() -> io::Result<()> {
+        let config_name = "package.json";
+        let config_content = json::json!({
+            "name": "starship",
+            "version": "v0.0.0-alpha"
+        })
+        .to_string();
+
+        let project_dir = create_project_dir()?;
+        fill_config(&project_dir, config_name, Some(&config_content))?;
+        expect_output(&project_dir, Some("v0.0.0-alpha"), None)?;
         project_dir.close()
     }
 
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -68,6 +68,10 @@ fn extract_package_version(file_contents: &str, display_private: bool) -> Option
     };
 
     let formatted_version = format_version(raw_version);
+    if formatted_version.ends_with("-semantically-released") {
+        return Some("semantic".to_string());
+    };
+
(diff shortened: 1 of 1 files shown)


vladimyr (approved):
This is nice!

> Normally, you'll see a version like 0.0.0-semantically-released in the package.json, to indicate that the package version is derived elsewhere. In it's current implementation, this is awkward because it conveys little meaning, uses quite a bit of space, and doesn't actually let you know the version

With that being said now you replaced the actual _version_ coming from `package.json` with a version hint but it lacks some visual indication to be easily differentiated from the regular version IMO

May I propose to emulate semantic release [badge](https://github.com/semantic-release/semantic-release#badge) by adding that rocket emoji before `semantic` 💡

src/modules/package.rs
```diff
@@ -68,6 +68,10 @@ fn extract_package_version(file_contents: &str, display_private: bool) -> Option
     };
 
     let formatted_version = format_version(raw_version);
+    if formatted_version == "v0.0.0-development" || formatted_version.find("semantic").is_some() {
```
vladimyr, line 71: Let's make it bit more strict because we might encounter actual prerelase version (non-zero) containing `semantic` identifier:
```suggestion
    if formatted_version == "v0.0.0-development" || formatted_version.starts_with("v0.0.0-semantic") {
```

vladimyr (changes requested):

src/modules/package.rs
matchai: The current implementation is aiming to identify any of the following common semantic release placeholders:
- `0.0.0-semantically-released`
- `0.0.0-semantic-versioning`
- `0.0.0-semantic-release`

Do you know of any packages that currently have the `semantic` identifier in their version, and aren't using semantic-release?

vladimyr: > The current implementation is aiming to identify any of the following common semantic-release placeholders:

That's perfectly fine I just wanted to make it more strict to prevent something like `1.4.21-foo-semantic-bar` being incorrectly matched. And yes, I'm not aware of anyone out there using such ridiculous prerelease identifiers but let's not underestimate people's creativity :stuck_out_tongue_winking_eye:

tbjgolden: Was thinking about this when I wrote this originally actually

My thinking was if they had `semantic` in a version like this then it's arguably still more likely (albeit speculative) that  they were using semantic versioning than not with something like `1.4.21-foo-semantic-bar` - though not sure how we'd be able to explain the `1.4.21`

Open to tweak it or keep the same, what do you think @matchai ?

matchai: I think the safer option is to prefix with v0.0.0, and keep an eye out for issues opening up with exceptional edge cases 😊

src/modules/package.rs
```diff
@@ -343,6 +347,51 @@ mod tests {
...
+        let config_name = "package.json";
+        let config_content = json::json!({
+            "name": "starship",
+            "version": "0.0.0-development"
+        })
+        .to_string();
+
+        let project_dir = create_project_dir()?;
+        fill_config(&project_dir, config_name, Some(&config_content))?;
+        expect_output(&project_dir, Some("semantic"), None)?;
```
vladimyr, line 361:
```suggestion
        expect_output(&project_dir, Some("semantic"), None);
```

src/modules/package.rs
```diff
@@ -343,6 +347,51 @@ mod tests {
...
+        let config_name = "package.json";
+        let config_content = json::json!({
+            "name": "starship",
+            "version": "v0.0.0-semantically-released"
+        })
+        .to_string();
+
+        let project_dir = create_project_dir()?;
+        fill_config(&project_dir, config_name, Some(&config_content))?;
+        expect_output(&project_dir, Some("semantic"), None)?;
```
vladimyr, line 376:
```suggestion
        expect_output(&project_dir, Some("semantic"), None);
```

src/modules/package.rs
```diff
@@ -343,6 +347,51 @@ mod tests {
...
+        let config_name = "package.json";
+        let config_content = json::json!({
+            "name": "starship",
+            "version": "v0.0.0-alpha"
+        })
+        .to_string();
+
+        let project_dir = create_project_dir()?;
+        fill_config(&project_dir, config_name, Some(&config_content))?;
+        expect_output(&project_dir, Some("v0.0.0-alpha"), None)?;
```
vladimyr, line 391:
```suggestion
        expect_output(&project_dir, Some("v0.0.0-alpha"), None);
```

vladimyr (changes requested):

matchai:
Hi @tbjgolden. It looks like this PR may have unfortunately fallen through the cracks. 😅 
I'd be happy to have it merged if you can go ahead and run `rustfmt` on the project.

tbjgolden:
> Hi @tbjgolden. It looks like this PR may have unfortunately fallen through the cracks. 😅 
> I'd be happy to have it merged if you can go ahead and run `rustfmt` on the project

Ah yes - will fix up now

tbjgolden:
Should be good to go

As for the rocket emoji to indicate semantic versioning - I like the idea though haven't included it here

Seems that the output currently generates a space between the package and the version, so might need some tweaking to get it showing `📦🚀` a la that badge

vladimyr:
> Seems that the output currently generates a space between the package and the version, so might need some tweaking to get it showing `📦🚀` a la that badge

Using `\b` is certainly possible but hardly advisable. 🙃

vladimyr (approved):
LGTM 🚀

matchai:
I think that the PR as-is serves the purpose of clarifying the version number well. I think that adding more emoji will start to clutter the prompt and confuse the user. I'm personally happy to merge it in its current state 😊

vladimyr:
Sure, sure, that was just me rambling about possibilities. It is perfectly fine in its current form!

matchai:
Thank you Tom and Dario for getting this PR though! 😊

PR #2280, opened 2021-02-08
feat(cmake): Configure when the module is shown

andytom:
#### Description

This makes it possible to configure when the cmake module is shown

#### Motivation and Context

Related to #1977 

#### Checklist:


davidkna (approved):
LGTM

PR #2281, opened 2021-02-08
feat: Move 'v' version prefix to module config, allowing a user to remove it

### LINKED ISSUE #2236: Option to remove 'v' from version number
## Feature Request

instead of 
`🐍v3.9.1`

```rust
    format!(
        "v{}",
```

I'm not aware of any

eugenesvk:
#### Description
I removed the hardcoded `v` version prefix from a module's main file to its config file for most of the modules (see below details on which modules are affected), allowing any user to remove such prefix from their prompt
(though I'd also suggest removing it from the config file altogether as it doesn't add any informational value)

#### Motivation and Context
It allows a user to save valuable prompt space
Closes #2236

After making a change to each module I ran `cargo test -- modules::<ModuleName>::tests` to make sure all module tests were passing and later ran this updated version of starship in a Windows Powershell session (though only testing the `rust`, `nodejs`, and `python` modules this way)

#### Checklist:

#### Modules affected:
2. Moved `v` from the module file to the config file:
  + cmake
  + crystal
  + dart
  + dotnet
  + elm
  + go
  + helm: original string has `v`, so added an extra `.trim_start_matches("v")`
  + java
  + julia
  + kotlin
  + lua
  + nim
  + nodejs: original string has `v`, so added an extra `.trim_start_matches('v')`
  + ocaml
  + package
  + perl
  + php
  + purescript
  + python: split python and pyenv versions to allow `v3.8.2` for the former while avoding `pyenv vsystem` for the latter (though that would also preclude `pyenv v3.8.2`)
  + ruby
  + rust
  + swift
  + vagrant
  + zig
2. Untouched, module already has no `v`
  + elixir
  + erlang
3. Untouched, module has no `$version`
  + aws
  + battery
  + character
  + cmd_duration
  + conda
  + custom
  + directory
  + docker_context
  + env_var
  + gcloud
  + git_branch, git_commit, git_state, git_status
  + hg_branch
  + hostname
  + jobs
  + kubernetes
  + memory_usage
  + mod
  + nix_shell
  + openstack
  + shlvl
  + singularity
  + starship_root
  + status
  + terraform
  + time
  + username

### DIFF
--- a/src/configs/python.rs
+++ b/src/configs/python.rs
@@ -4,8 +4,6 @@ use starship_module_config_derive::ModuleConfig;
 
 #[derive(Clone, ModuleConfig)]
 pub struct PythonConfig<'a> {
-    pub pyenv_version_name: bool,
-    pub pyenv_prefix: &'a str,
     pub python_binary: VecOr<&'a str>,
     pub format: &'a str,
     pub style: &'a str,
@@ -19,10 +17,8 @@ pub struct PythonConfig<'a> {
 impl<'a> RootModuleConfig<'a> for PythonConfig<'a> {
     fn new() -> Self {
         PythonConfig {
-            pyenv_version_name: false,
-            pyenv_prefix: "pyenv ",
             python_binary: VecOr(vec!["python", "python3", "python2"]),
-            format: "via [${symbol}${pyenv_prefix}(${pyenv_version} )(v${version} )(\\($virtualenv\\))]($style)",
+            format: "via [${symbol}(v${version} )(\\($virtualenv\\))]($style)",
             style: "yellow bold",
             symbol: "🐍 ",
             disabled: false,
--- a/src/configs/python.rs
+++ b/src/configs/python.rs
@@ -22,7 +22,7 @@ impl<'a> RootModuleConfig<'a> for PythonConfig<'a> {
             pyenv_version_name: false,
             pyenv_prefix: "pyenv ",
             python_binary: VecOr(vec!["python", "python3", "python2"]),
-            format: "via [${symbol}${pyenv_prefix}(${version} )(\\($virtualenv\\))]($style)",
+            format: "via [${symbol}${pyenv_prefix}(${pyenv_version} )(v${version} )(\\($virtualenv\\))]($style)",
             style: "yellow bold",
             symbol: "🐍 ",
             disabled: false,
--- a/src/modules/helm.rs
+++ b/src/modules/helm.rs
@@ -69,6 +69,8 @@ fn format_helm_version(helm_stdout: &str) -> Option<String> {
             // return "v3.1.1" or " v3.1.1"
             .trim_start_matches("Client: ")
             // return "v3.1.1"
+            .trim_start_matches('v')
+            // return "3.1.1"
             .trim()
             .to_owned(),
     )
@@ -121,7 +123,7 @@ mod tests {
     fn test_format_helm_version() {
         let helm_2 = "Client: v2.16.9+g8ad7037";
         let helm_3 = "v3.1.1+ggit afe7058";
-        assert_eq!(format_helm_version(helm_2), Some("v2.16.9".to_string()));
-        assert_eq!(format_helm_version(helm_3), Some("v3.1.1".to_string()));
+        assert_eq!(format_helm_version(helm_2), Some("2.16.9".to_string()));
+        assert_eq!(format_helm_version(helm_3), Some("3.1.1".to_string()));
     }
 }
--- a/src/modules/julia.rs
+++ b/src/modules/julia.rs
@@ -66,7 +66,7 @@ fn format_julia_version(julia_stdout: &str) -> Option<String> {
         .split_whitespace()
         .next()?;
 
-    Some(format!("v{}", version))
+    Some(version.to_string())
 }
 
 #[cfg(test)]
@@ -127,6 +127,6 @@ mod tests {
     #[test]
     fn test_format_julia_version() {
         let input = "julia version 1.4.0";
-        assert_eq!(format_julia_version(input), Some("v1.4.0".to_string()));
+        assert_eq!(format_julia_version(input), Some("1.4.0".to_string()));
     }
 }
--- a/src/modules/kotlin.rs
+++ b/src/modules/kotlin.rs
@@ -77,7 +77,7 @@ fn format_kotlin_version(kotlin_stdout: &str) -> Option<String> {
     let re = Regex::new(KOTLIN_VERSION_PATTERN).ok()?;
     let captures = re.captures(kotlin_stdout)?;
     let version = &captures["version"];
-    Some(format!("v{}", version))
+    Some(version.to_string())
 }
 
 #[cfg(test)]
@@ -162,7 +162,7 @@ mod tests {
         let kotlin_input = "Kotlin version 1.4.21-release-411 (JRE 14.0.1+7)";
         assert_eq!(
             format_kotlin_version(kotlin_input),
-            Some("v1.4.21".to_string())
+            Some("1.4.21".to_string())
         );
     }
 
@@ -171,7 +171,7 @@ mod tests {
         let kotlin_input = "info: kotlinc-jvm 1.4.21 (JRE 14.0.1+7)";
         assert_eq!(
             format_kotlin_version(kotlin_input),
-            Some("v1.4.21".to_string())
+            Some("1.4.21".to_string())
         );
     }
 }
--- a/src/modules/lua.rs
+++ b/src/modules/lua.rs
@@ -80,7 +80,7 @@ fn format_lua_version(lua_stdout: &str) -> Option<String> {
     let re = Regex::new(LUA_VERSION_PATERN).ok()?;
     let captures = re.captures(lua_stdout)?;
     let version = &captures["version"];
-    Some(format!("v{}", version))
+    Some(version.to_string())
 }
 
 #[cfg(test)]
@@ -156,13 +156,13 @@ mod tests {
     #[test]
     fn test_format_lua_version() {
         let lua_input = "Lua 5.4.0  Copyright (C) 1994-2020 Lua.org, PUC-Rio";
-        assert_eq!(format_lua_version(lua_input), Some("v5.4.0".to_string()));
+        assert_eq!(format_lua_version(lua_input), Some("5.4.0".to_string()));
 
         let luajit_input =
             "LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/";
         assert_eq!(
             format_lua_version(luajit_input),
-            Some("v2.1.0-beta3".to_string())
+            Some("2.1.0-beta3".to_string())
         );
     }
 }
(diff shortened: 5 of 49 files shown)


src/modules/helm.rs
```diff
@@ -69,6 +69,8 @@ fn format_helm_version(helm_stdout: &str) -> Option<String> {
             // return "v3.1.1" or " v3.1.1"
             .trim_start_matches("Client: ")
             // return "v3.1.1"
+            .trim_start_matches("v")
```
andytom, line 72:
```suggestion
            .trim_start_matches('v')
```

src/modules/julia.rs
```diff
@@ -66,7 +66,7 @@ fn format_julia_version(julia_stdout: &str) -> Option<String> {
         .split_whitespace()
         .next()?;
 
-    Some(format!("v{}", version))
+    Some(format!("{}", version))
```
andytom, line 69:
```suggestion
    Some(version.to_string())
```

src/modules/lua.rs
```diff
@@ -80,7 +80,7 @@ fn format_lua_version(lua_stdout: &str) -> Option<String> {
     let re = Regex::new(LUA_VERSION_PATERN).ok()?;
     let captures = re.captures(lua_stdout)?;
     let version = &captures["version"];
-    Some(format!("v{}", version))
+    Some(format!("{}", version))
```
andytom, line 83:
```suggestion
    Some(version.to_string())
```

src/modules/nim.rs
```diff
@@ -37,7 +37,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "version" => utils::exec_cmd("nim", &["--version"])
                     .map(|command_output| command_output.stdout)
                     .and_then(|nim_version_output| {
-                        Some(format!("v{}", parse_nim_version(&nim_version_output)?))
+                        Some(format!("{}", parse_nim_version(&nim_version_output)?))
```
andytom, line 40:
```suggestion
                        Some(parse_nim_version(&nim_version_output)?.to_string())
```

src/modules/ocaml.rs
```diff
@@ -50,7 +50,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                     } else {
                         utils::exec_cmd("ocaml", &["-vnum"])?.stdout
                     };
-                    Some(Ok(format!("v{}", &ocaml_version.trim())))
+                    Some(Ok(format!("{}", &ocaml_version.trim())))
```
andytom, line 53:
```suggestion
                    Some(Ok(ocaml_version.trim().to_string()))
```

src/modules/package.rs
```diff
@@ -201,9 +201,9 @@ fn get_package_version(base_dir: &Path, config: &PackageConfig) -> Option<String
 fn format_version(version: &str) -> String {
     let cleaned = version.replace('"', "").trim().to_string();
     if cleaned.starts_with('v') {
-        cleaned
+        cleaned.trim_start_matches('v').to_string()
     } else {
-        format!("v{}", cleaned)
+        cleaned
     }
```
andytom, line 207: I think these few lines can be replaced with 
```rust
 version.replace('"', "").trim().trim_start_matches('v').to_string()
```

src/modules/perl.rs
```diff
@@ -46,7 +46,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "version" => {
                     let perl_version =
                         utils::exec_cmd("perl", &["-e", "printf q#%vd#,$^V;"])?.stdout;
-                    Some(Ok(format!("v{}", perl_version)))
+                    Some(Ok(format!("{}", perl_version)))
```
andytom, line 49:
```suggestion
                    Some(Ok(perl_version.to_string()))
```

src/modules/php.rs
```diff
@@ -61,7 +61,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn format_php_version(php_version: &str) -> String {
-    format!("v{}", php_version)
+    format!("{}", php_version)
 }
```
andytom, line 65: This whole function can be removed.

src/modules/purescript.rs
```diff
@@ -36,7 +36,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             .map(|variable| match variable {
                 "version" => {
                     let purs_version = utils::exec_cmd("purs", &["--version"])?.stdout;
-                    Some(Ok(format!("v{}", purs_version.trim())))
+                    Some(Ok(format!("{}", purs_version.trim())))
```
andytom, line 39:
```suggestion
                    Some(Ok(purs_version.trim().to_string()))
```

src/modules/kotlin.rs
```diff
@@ -77,7 +77,7 @@ fn format_kotlin_version(kotlin_stdout: &str) -> Option<String> {
     let re = Regex::new(KOTLIN_VERSION_PATTERN).ok()?;
     let captures = re.captures(kotlin_stdout)?;
     let version = &captures["version"];
-    Some(format!("v{}", version))
+    Some(format!("{}", version))
```
andytom, line 80:
```suggestion
    Some(fversion.to_string())
```

src/modules/ruby.rs
```diff
@@ -64,7 +64,6 @@ fn format_ruby_version(ruby_version: &str) -> Option<String> {
         .next()?;
 
     let mut formatted_version = String::with_capacity(version.len() + 1);
-    formatted_version.push('v');
     formatted_version.push_str(version);
     Some(formatted_version)
```
andytom, line 68: This can be replaced with
```rust
version.to_string()
```

src/modules/rust.rs
```diff
@@ -209,7 +209,7 @@ fn format_rustc_version(mut rustc_stdout: String) -> String {
     let offset = &rustc_stdout.find('(').unwrap_or_else(|| rustc_stdout.len());
     let formatted_version: String = rustc_stdout.drain(..offset).collect();
 
-    format!("v{}", formatted_version.replace("rustc", "").trim())
+    format!("{}", formatted_version.replace("rustc", "").trim())
```
andytom, line 212:
```suggestion
    formatted_version.replace("rustc", "").trim().to_string()
```

src/modules/swift.rs
```diff
@@ -61,7 +61,7 @@ fn parse_swift_version(swift_version: &str) -> Option<String> {
     // return "5.2.2" or "5.3-dev"
     let version = splited.next()?;
 
-    Some(format!("v{}", version))
+    Some(format!("{}", version))
```
andytom, line 64:
```suggestion
    Some(version.to_string())
```

src/modules/vagrant.rs
```diff
@@ -61,7 +61,6 @@ fn format_vagrant_version(vagrant_stdout: &str) -> Option<String> {
         .nth(1)?;
 
     let mut formatted_version = String::with_capacity(version.len() + 1);
-    formatted_version.push('v');
     formatted_version.push_str(version);
     Some(formatted_version)
```
andytom, line 65: I think this can be replaced with 
```rust
Some(version.to_string())
```

src/modules/zig.rs
```diff
@@ -34,7 +34,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             .map(|variable| match variable {
                 "version" => {
                     let zig_version_output = utils::exec_cmd("zig", &["version"])?.stdout;
-                    let zig_version = format!("v{}", zig_version_output.trim());
+                    let zig_version = format!("{}", zig_version_output.trim());
                     Some(Ok(zig_version))
```
andytom, line 38: I think this can now be
```rust
Some(Ok(utils::exec_cmd("zig", &["version"])?.stdout.trim().to_string()))
```

andytom (commented):
I've added some suggestions to simplify some of the codes and fix some of the clippy issues. I've just bin doing this via the Github UI so have there might be some typos or style issues that need to be fixed.

src/configs/python.rs
```diff
@@ -22,7 +22,7 @@ impl<'a> RootModuleConfig<'a> for PythonConfig<'a> {
             pyenv_version_name: false,
             pyenv_prefix: "pyenv ",
             python_binary: VecOr(vec!["python", "python3", "python2"]),
-            format: "via [${symbol}${pyenv_prefix}(${version} )(\\($virtualenv\\))]($style)",
+            format: "via [${symbol}${pyenv_prefix}(${pyenv_version} )(v${version} )(\\($virtualenv\\))]($style)",
```
andytom, line 25: For python maybe this would be a good time to make a bigger change and deprecate the `pyenv_prefix` and `pyenv_version_name` and just use the format string with the new `pyenv_version` variable. This would simplify the code and make the configuration a bit more consistent. This would be a breaking change at the moment to use the pyenv version you would have to set the following
```toml
[python]
pyenv_version_name = true
```
after this you would need to do
```toml
[python]
format = 'via [${symbol}pyenv ${pyenv_version} (\($virtualenv\))]($style)'
```
What do you think?

eugenesvk: I'm fine with with removing the `pyenv_prefix` since I don't see any benefits to it vs. inserting its value directly in the format string since I'm always editing the format string anyway, though there might be users who are fine with the format and only need to change a prefix? No clue how common it is
Re. the `pyenv_version_name` — if it's removed would this mean that every prompt refresh would need to make calls to both python **and** pyenv to update even if the user doesn't have/use pyenv? That might be a performance issue?

Otherwise I've incorporated all the other code simplifications and will push shortly to see how this affects all those CI checks

andytom: The python and pyenv commands are lazily evaluated, so they would only be called when the `$version` and/or `$pyenv_version` variables are used in the format string.

eugenesvk: Ok, but the default format string does include both of these variables in the format string, so this doesn't solve this issue. If it is an issue, that is
Then this gets us back to my first point — how useful this extra bool var is for other users:
- is it that they simply edit the `format` string in their configs (like I do since I also need to get rid of all those useless `via `) and then it's largely irrelevant since you'd just as well remove the `$pyenv_version` if you don't need it
- or is it that many don't touch the `format` string, but only add `symbol`/`pyenv_version_name` vars

andytom: Ah sorry, I didn't explain that side of things. I would just change the default format string to add the `v` so that it is  something like `via [${symbol}(v${version} )(\($virtualenv\) )]($style)'`. So if a user wants to use the pyenv version they would just modify the format string.

eugenesvk: Oh, yeah, that's fine, I'll try to change this later
Would it be better, maybe, to make it a separate PR given it's a breaking change?

eugenesvk: Ok, added it as a separate commit, but still within this PR (and sismplified per the next review comment)

andytom:
I think this has been superseded by #2611, so I am going to close this, please feel free to reopen it if this is not the case. @eugenesvk I am sorry this did not get merged.

eugenesvk:
What about the other modules (e.g.  `package`) that weren't touched in the superseding PR? Do you plan to also cover them there or maybe take a few commits from this PR?

cymruu:
@eugenesvk https://github.com/starship/starship/issues/2236#issuecomment-831474650 this comment suggests the change is wanted, so I think you should create new PR with changes to `package` module

PR #2288, opened 2021-02-10
feat: add support for tcsh

### LINKED ISSUE #1515: Can starship support tcsh?
I received this message when trying to use starship on tcsh:
"printf "\ntcsh is not yet supported by starship.\nFor the time being, we support bash, zsh, fish, and ion.\nPlease open an issue in the starship repo if you would like to see support for tcsh:\nhttps://github.com/starship/starship/issues/new"\n\n"

Unfortunately, tcsh is what my company uses, and their is no way around it

Any chance you will be offering this capability?

vivekmalneedi:
#### Description

adds support for the tcsh shell

#### Motivation and Context

Closes #1515

#### Checklist:


src/init/starship.tcsh
```diff
@@ -0,0 +1,14 @@
+setenv STARSHIP_SHELL tcsh;
+setenv STARSHIP_SESSION_KEY `(::STARSHIP:: session)`;
+alias precmd 'set STARSHIP_CMD_STATUS = $status;set NUM_JOBS=`(jobs | wc -l)`;set STARSHIP_END_TIME = `::STARSHIP:: time`;set STARSHIP_DURATION = 0;if ( $?STARSHIP_START_TIME ) @ STARSHIP_DURATION = $STARSHIP_END_TIME - $STARSHIP_START_TIME;set prompt = "`(::STARSHIP:: prompt --status $STARSHIP_CMD_STATUS --jobs $NUM_JOBS --cmd-duration $STARSHIP_DURATION)`";set prompt = "\n$prompt[1] \n$prompt[2]";unset STARSHIP_START_TIME;';
```
davidkna, line 3: From what I've seen disabling newlines is among the most popular things people change in their configs. Is a `for` loop not possible here?

src/init/mod.rs
```diff
@@ -212,6 +219,11 @@ pub fn init_main(shell_name: &str) -> io::Result<()> {
         "powershell" => print_script(PWSH_INIT, &starship_path.sprint()?),
         "ion" => print_script(ION_INIT, &starship_path.sprint()?),
         "elvish" => print_script(ELVISH_INIT, &starship_path.sprint_posix()?),
+        // print_script() adds quotes around starship_path which breaks starship.tcsh
+        "tcsh" => print!(
+            "{}",
+            TCSH_INIT.replace("::STARSHIP::", &starship_path.sprint_posix()?)
```
davidkna, line 225: I think you should still add something here to properly escape the path to starship.

src/init/starship.tcsh
vivekmalneedi: a for loop is unfortunately not possible there, but I realized all I had to do was escape the newlines from within starship. So there are no longer any limitations on newlines in my latest revision.

src/init/mod.rs
vivekmalneedi: fixed, it is now consistent with the other shells.

davidkna:
Please list `tcsh` in [`bug_report.rs`](https://github.com/starship/starship/blob/master/src/bug_report.rs#L207) along with the other shells.

vivekmalneedi:
@davidkna fixed, added there and in a main.rs help message

src/init/mod.rs
```diff
@@ -173,6 +173,13 @@ fi"#,
             );
             Some(script)
         }
+        Some("tcsh") => {
+            let script = format!(
+                r#"eval "`({} init tcsh --print-full-init)`""#,
```
davidkna, line 178: Please quote this path:

```suggestion
                r#"eval "`("{}" init tcsh --print-full-init)`""#,
```

davidkna (changes requested):
This mostly looks good to merge, but there's still something to do beyond the path quoting comment I made.

1. You need to rebase this on master and add compatibility with the new shell module.
1. For me the number of jobs isn't shown correctly with your code. After starting a job `jobs | wc -l` still prints 0.

vivekmalneedi:
thank you for catching the issue with jobs, it is now fixed and I have added support for the shell module

src/init/starship.tcsh
```diff
@@ -0,0 +1,6 @@
+setenv STARSHIP_SHELL tcsh;
+setenv STARSHIP_SESSION_KEY `(::STARSHIP:: session)`;
+alias precmd 'set STARSHIP_CMD_STATUS = $status;set STARSHIP_PATH = ::STARSHIP::;jobs > $STARSHIP_TCSH_TMPFILE;set NUM_JOBS=`(cat $STARSHIP_TCSH_TMPFILE | wc -l)`;rm $STARSHIP_TCSH_TMPFILE;set STARSHIP_END_TIME = `$STARSHIP_PATH time`;set STARSHIP_DURATION = 0;if ( $STARSHIP_START_TIME != -1 ) @ STARSHIP_DURATION = $STARSHIP_END_TIME - $STARSHIP_START_TIME;set prompt = "`($STARSHIP_PATH prompt --status $STARSHIP_CMD_STATUS --jobs $NUM_JOBS --cmd-duration $STARSHIP_DURATION)`";set STARSHIP_START_TIME = -1;';
```
davidkna, line 3: I don't really like the jobs workaround here. I'd prefer if you removed it and added a warning in the docs about it.

Also: The other shell inits preserve the old `precmd` and `postcmd` hooks. Is this also possible in tcsh?

vivekmalneedi: 1. I agree, it's a messy workaround.
2. I have added this functionality to the script.

davidkna (approved):
LGTM

docs/config/README.md
```diff
@@ -1399,6 +1399,10 @@ The module will be shown only if there are background jobs running.
 The module will show the number of jobs running if there is more than 1 job, or
 more than the `threshold` config value, if it exists.
 
+::: warning
+This module is not supported on tcsh.
+:::
```
andytom, line 1404: I can't remember the exact reasons but I think we need to have additional line breaks around the `:::` to work better with CrowdIn.
```suggestion
::: warning

This module is not supported on tcsh.

:::
```

andytom (approved):

andytom:
Thank you for your contribution @vivekmalneedi

install/install.sh
```diff
@@ -432,4 +432,9 @@ info "Please follow the steps for your shell to complete the installation:
   Add the following to the end of ${BOLD}~/.config/ion/initrc${NO_COLOR}:
 
       eval \$(starship init ion)
+
+  ${BOLD}${UNDERLINE}Tcsh${NO_COLOR}
+  Add the following to the end of ${BOLD}~/.tcshrc${NO_COLOR}:
+
+      eval "`starship init tcsh`"
```
vladimyr, line 439: Here is what `shellcheck` thinks about it:
```
Line 439:
      eval "`starship init tcsh`"
            ^-- SC2046: Quote this to prevent word splitting.
            ^-- SC2006: Use $(...) notation instead of legacy backticked `...`.
```

First of all AFAIK, you don't need to quote _backticked_ portion so it can be simplified to:
```tcsh
eval `starship init tcsh`
```

and to incorporate that nicely into the existing script use the following form:
```
  ${BOLD}${UNDERLINE}Tcsh${NO_COLOR}
  Add the following to the end of ${BOLD}~/.tcshrc${NO_COLOR}:

      eval \`starship init tcsh\`
```

Escaping backticks prevents SC2006 false positive.

docs/README.md
```diff
@@ -115,3 +115,13 @@ description: Starship is the minimal, blazing fast, and extremely customizable p
...
    ```
 
+   #### Tcsh
+
+   Add the following to the end of `~/.tcshrc`:
+
+   ```sh
+   # ~/.tcshrc
+
+   eval "`starship init tcsh`"
```
vladimyr, line 125: Quoting this is unnecessary.

install/install.sh
vivekmalneedi: Thanks, I have submitted a PR: #2429, to fix these issues

PR #2291, opened 2021-02-10
docs: Update Nix installation documentation

pnotequalnp:
This changes the documentation to show the provided Home Manager module instead of manual Home Manager installation instructions. Also fixes a typo in the NixOS instructions and removes the unfinished and now unnecessary section on zsh integration via Home manager (since that is included in the provided module).

#### Checklist:

### DIFF
--- a/docs/installing/README.md
+++ b/docs/installing/README.md
@@ -32,19 +32,16 @@ Enable the `programs.starship` module in your `home.nix` file, and add your sett
   programs.starship = {
     enable = true;
     enableZshIntegration = true;
+    # Configuration written to ~/.config/starship.toml
     settings = {
-      add_newline = false;
-      format = lib.concatStrings [
-        "$line_break"
-        "$package"
-        "$line_break"
-        "$character"
-      ];
-      scan_timeout = 10;
-      character = {
-        success_symbol = "➜";
-        error_symbol = "➜";
-      };
+      # add_newline = false;
+
+      # character = {
+      #   success_symbol = "[➜](bold green)";
+      #   error_symbol = "[➜](bold red)";
+      # };
+
+      # package.disabled = true;
     };
   };
 }
--- a/docs/installing/README.md
+++ b/docs/installing/README.md
@@ -25,7 +25,32 @@ nix-env -iA nixos.starship
 
 #### Declarative, single user, via [home-manager](home-manager)
 
-Add `pkgs.starship` to your `home.packages` in your `home.nix` file, then run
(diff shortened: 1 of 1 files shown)


docs/installing/README.md
```diff
@@ -25,28 +25,42 @@ nix-env -iA nixos.starship
...
 
-Add `pkgs.starship` to your `home.packages` in your `home.nix` file, then run
+Enable the `programs.starship` module in your `home.nix` file, and add your settings
+
+```nix
+{
+  programs.starship = {
+    enable = true;
+    enableZshIntegration = true;
+    settings = {
```
chipbuster, line 35: I'm not familiar with how Nix works here. Do these settings eventually become part of the starship config file? If so, are they necessary to set here, or is this just a sane default set?

pnotequalnp: They're just the example config [that HM gives](https://rycee.gitlab.io/home-manager/options.html#opt-programs.starship.settings). The HM module [generates the Starship config file](https://github.com/nix-community/home-manager/blob/master/modules/programs/starship.nix#L96) from the `settings` entry.

pnotequalnp: Here's an adapted version of [the example on the Starship docs](https://starship.rs/config/) with every option commented.

```nix
{
  settings = {
    # add_newline = false;

    # character = {
    #   success_symbol = "[➜](bold green)";
    #   error_symbol = "➜(bold red)";
    # };

    # package.disabled = true;
  };
}
```

Would this be better for the HM example?

chipbuster: I believe 
```
#   error_symbol = "➜(bold red)";
```

needs to be

```
#   error_symbol = "[➜](bold red)";
```

but otherwise yes, I think that would be a good config. It might also be nice to add a comment in that block explaining that this is for the config/settings and can be set to change the generated starship.toml? Not sure how to phrase that in a concise way.

pnotequalnp: Indeed, that's a typo. And using the description in the HM manual for the settings entry.

```nix
{
  # Configuration written to ~/.config/starship.toml
  settings = {
    # add_newline = false;

    # character = {
    #   success_symbol = "[➜](bold green)";
    #   error_symbol = "[➜](bold red)";
    # };

    # package.disabled = true;
  };
}
```

chipbuster: That looks good!

chipbuster (approved):
Looks good to me

andytom:
Thank you for your contribution @pnotequalnp.

PR #2292, opened 2021-02-11
perf(init): Prefer set -x over export for fish users

msanders:
#### Motivation and Context

In Fish, `export` is a function that wraps the built-in `set` command for compatibility with POSIX shells. It’s more idiomatic and has slightly less overhead to use the built-in `set -x` instead. See https://github.com/fish-shell/fish-shell/issues/2704#issuecomment-178749261.

With `export`:

```
$ hyperfine "fish -ic exit"
Benchmark #1: fish -ic exit
  Time (mean ± σ):      41.8 ms ±   1.1 ms    [User: 29.1 ms, System: 12.7 ms]
  Range (min … max):    40.2 ms …  45.5 ms    57 runs
```

With `set -x`:

```
$ hyperfine "fish -ic exit"
Benchmark #1: fish -ic exit
  Time (mean ± σ):      38.7 ms ±   1.5 ms    [User: 26.6 ms, System: 11.7 ms]
  Range (min … max):    35.4 ms …  43.2 ms    60 runs
```

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -15,7 +15,7 @@ end
 set VIRTUAL_ENV_DISABLE_PROMPT 1
 
 function fish_mode_prompt; end
-export STARSHIP_SHELL="fish"
+set -gx STARSHIP_SHELL="fish"
 
 # Set up the session key that will be used to store logs
-export STARSHIP_SESSION_KEY=(random 10000000000000 9999999999999999)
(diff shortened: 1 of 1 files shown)


andytom (approved):

andytom:
Thank you for your contribution @msanders

PR #2296, opened 2021-02-11
fix(init): fix error message when trying to 'starship init' with an unsupported shell

bloatoo:
```shell
-> starship init unknown_shell
printf "\nunknown_shell is not yet supported by starship.\nFor the time being, we support bash, zsh, fish, and ion.\nPlease open an issue in the starship repo if you would like to see support for unknown_shell:\nhttps://github.com/starship/starship/issues/new"\n\n
```

#### Checklist:

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -185,7 +185,7 @@ fi"#,
         }
         Some(shell_basename) => {
             println!(
-                "\"{0}\" is not yet supported by starship.\n\
+                "{0} is not yet supported by starship.\n\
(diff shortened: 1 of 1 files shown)


matchai:
The reason that we escape the escape characters is so that the error will print correctly when the output is sourced, as should be the case when users copy the init script:

CleanShot 2021-02-11 at 14 10 17@2x

davidkna:
You should be able to add a check for eval with something like [this](https://docs.rs/atty/0.2.14/atty/).

PR #2307, opened 2021-02-13
feat(test): allow dynamic mocking of commands

### LINKED ISSUE #942: Refactor `utils::exec_cmd` to make testing easier

We are currently working on writing better unit test for modules and moving away from having tests that depend on having the development machine setup in a particular way. We have a simple abstraction around executing commands `utils::exec_cmd` and a crude way of mocking the output. 

> ... while it made the initial implementation easy the current way of doing things does have it's limitations. In addition to keeping the tests and the mocks closer together this would allow us to set different responses for on a per test basis, this would be useful for Java and Python where some versions print the version strings on `stdout` and others use `stderr`. Also the case statement is getting a bit unwieldy and we haven't even migrated all the other commands in.

_Originally posted by @andytom in https://github.com/starship/starship/pull/936#issuecomment-586584265_

I've not really considered anything else but interested in any suggestions.

davidkna:
#### Description

I added a new method `cmd` to `ModuleRenderer` to allow dynamic mocking of commands and implemented it in `Context`.
I added two additional tests for the java module that make use of this new feature.
In addition, for a small perf boost I gated the `Context.env` `HashMap` with `#[cfg(test)]`.

#### Motivation and Context

Closes #942

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -43,7 +43,12 @@ pub struct Context<'a> {
     pub shell: Shell,
 
     /// A HashMap of environment variable mocks
+    #[cfg(test)]
     pub env: HashMap<&'a str, String>,
+
+    /// A HashMap of command mocks
+    #[cfg(test)]
+    pub cmd: HashMap<&'a str, Option<CommandOutput>>,
 
     /// Timeout for the execution of commands
     cmd_timeout: Duration,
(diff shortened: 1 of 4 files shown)


andytom (approved):
LGTM

PR #2308, opened 2021-02-13
docs: Clarify wording on add_newline option

### LINKED ISSUE #1987: Clarify what `add_newline` and `line_break` do

We've had a number of questions about how disabling `add_newline` wasn't working when what the user really wanted to do was disable the `line_break` module. The most recently issue is #1982, but there are other issues, discussions, and Discord questions around the same subject.

Clarify in the documentation that `add_newline` is not responsible for splitting the prompt into two lines.

I don't know what the best way to do this is: adding a lot of text to the description of `add_newline` on the configuration page seems like it could add noise to our docs. An FAQ entry might be nice, but I don't think many people will look at the FAQ page if they think that `add_newline` does what they want.

Maybe an FAQ entry with a link and short notice on the config page would be good? I'm also open to other ways of formatting this,

Changing the name of the option, but I think that ship has sailed.

adaschevici:
#### Description
Change wording for the confusing `add_newline` sentence to something that hopefully makes more sense.

#### Motivation and Context
It aims to prevent users from being confused by the subtle difference between `add_newline` and `line_break` options

Closes #1987 

Tested by checking the docs local build.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -9,7 +9,7 @@ mkdir -p ~/.config && touch ~/.config/starship.toml
 All configuration for starship is done in this [TOML](https://github.com/toml-lang/toml) file:
 
 ```toml
-# Inserts a newline between shell prompts
+# Inserts a blank line between shell prompts
 add_newline = true
 
 # Replace the "❯" symbol in the prompt with "➜"
@@ -155,7 +155,7 @@ This is the list of prompt-wide configuration options.
 | -------------- | ------------------------------ | ----------------------------------------------------- |
 | `format`       | [link](#default-prompt-format) | Configure the format of the prompt.                   |
 | `scan_timeout` | `30`                           | Timeout for starship to scan files (in milliseconds). |
-| `add_newline`  | `true`                         | Inserts new line between shell prompts.               |
+| `add_newline`  | `true`                         | Inserts blank line between shell prompts.               |
 
 ### Example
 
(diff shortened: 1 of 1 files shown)


docs/config/README.md
```diff
@@ -9,8 +9,8 @@ mkdir -p ~/.config && touch ~/.config/starship.toml
 All configuration for starship is done in this [TOML](https://github.com/toml-lang/toml) file:
 
 ```toml
-# Don't print a new line at the start of the prompt
-add_newline = false
+# Inserts a newline between shell prompts
```
chipbuster, line 12: Do you think clarifying that this is  a "blank line" instead of newline will help?

To me, a newline is just a line break, whereas a "blank line" is specifically a whole line with nothing on it--perhaps this will help further clear up some of the confusion? (It doesn't make much sense to insert a blank line in the middle of the prompt).

chipbuster (commented):
Just a suggestion. If you think "newline" is better, then we should be good to merge this as-is.

docs/config/README.md
adaschevici: makes sense, updating

chipbuster (approved):
LGTM but I'd like to get approval from at least one other astronaut before merging.

cappyzawa (approved):
LGTM

andytom:
Thank you for your contribution @adaschevici.

PR #2311, opened 2021-02-13
feat(zig): Configure when the module is shown

davidkna:
#### Description

This makes it possible to configure when the zig module is shown based on the contents of a directory.

#### Motivation and Context

Related to #1977 

#### Checklist:

### DIFF
--- a/src/modules/zig.rs
+++ b/src/modules/zig.rs
@@ -8,17 +8,19 @@ use crate::formatter::StringFormatter;
 /// Will display the Zig version if any of the following criteria are met:
 ///     - The current directory contains a file with extension `.zig`
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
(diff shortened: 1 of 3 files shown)


src/modules/zig.rs
```diff
@@ -8,18 +8,20 @@ use crate::formatter::StringFormatter;
 /// Will display the Zig version if any of the following criteria are met:
```
andytom, line 8: Not a blocker but I think we can remove this comment.
```suggestion
```

src/modules/zig.rs
```diff
@@ -8,18 +8,20 @@ use crate::formatter::StringFormatter;
...
 ///     - The current directory contains a file with extension `.zig`
```
andytom, line 9:
```suggestion
```

andytom (approved):
LGTM

PR #2322, opened 2021-02-16
ci(docs): only publish site on release

### LINKED ISSUE #2298: Is it possible to update https://starship.rs/config/ only with released features

> Is it possible to update https://starship.rs/config/#python only with released features, not one in master but not released yet?
> I found myself in the situation when documentation says to use this 
>  ```toml
> detect_extensions = []
> ```
> but it is actually not released and in my 0.50.0 version doesn't work.
> Thanks to other languages documentation I've found an old way of excluding extensions for python.

_Originally posted by @IevgenSobko in https://github.com/starship/starship/issues/2264#issuecomment-776852086_

We should look into how we update our docs to make it obvious when a change is available in a released version rather than just being merged into master. This could be that we version the docs (similar to https://docs.rs/) or maybe we only update the website on release of a new version or maybe there is another alternative?

I'm not sure what the best way to do this is so opening this up as a discussion please add suggestions, pros/cons of different solutions etc

matchai:
#### Description

This PR triggers a Netlify deployment of our docs as part of our release pipeline. When this PR is merged, I'll go ahead and disable automated deployments in the Netlify dashboard:

CleanShot 2021-02-16 at 16 43 22

#### Motivation and Context

Closes #2298

This PR ensures that docs aren't updated ahead of the release of the associated features.

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -33,6 +33,20 @@ jobs:
         env:
           COMMITTER_TOKEN: ${{ secrets.HOMEBREW_GITHUB_API_TOKEN }}
 
+  publish_docs:
+    name: Publish docs to Netlify
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Setup | Checkout
+      uses: actions/checkout@master
+
+    - name: Publish
+      uses: netlify/actions/build@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
+
   # Build sources for every OS
   github_build:
     name: Build release binaries


.github/workflows/deploy.yml
```diff
@@ -33,6 +33,20 @@ jobs:
...
+
+    steps:
+    - name: Setup | Checkout
+      uses: actions/checkout@master
+
+    - name: Publish
+      uses: netlify/actions/build@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
```
matchai, line 48: This required secret has now been added to the project 🔑

andytom (approved):
LGTM

PR #2324, opened 2021-02-17
feat(terraform): Configure when the module is shown

cappyzawa:
#### Description

This makes it possible to configure when the terraform module is shown

#### Motivation and Context

Related to #1977 

#### Checklist:


davidkna (approved):

andytom:
Thanks for updating this module @cappyzawa.

PR #2329, opened 2021-02-18
feat: create shell module

### LINKED ISSUE #1380: Easily identify current shell as bash, fish, nushell, etc.
## Feature Request

Thank you for a really awesome prompt. Starship is fast and informative.

I usually use `fish`, but I've been experimenting with `nushell` lately, and the two shells use very different syntax. But I use Starship with both. This leaves me with no easy way to tell which shell a terminal is running.

Ideally, I'd like to be able to change my prompt to "🐟❯" or "nu❯", making it easily visible which shell syntax I should use. There are several ways that this could be done, in the section below.

1. This could be implemented using some kind of `[shell.nushell.character]` section containing `success = "nu❯"`. Essentially, allow shell specific configuration overrides.
2. I can use the `env_var` module to display `SHELL`, but this only allows me to insert `/usr/bin/fish` somewhere in the prompt line.
3. I could configure each of my shells to set a custom `SHELL_SYMBOL` variable, and then configure `env_var` to display that variable at the far left of the prompt.
4. Somebody could implement a `shell` plugin that appeared towards the left of the prompt, and that included an emoji or short string appropriate to common shells.

(3) already works, I think, so I should probably just do that. But if anybody else is having trouble telling shells apart, it might be nice to have an easier way to tell shells apart.

rosowskimik:
#### Description
A "shell" module, which allows to quickly identify which shell is being used.

#### Motivation and Context
Closes #1380

Ran cargo test modules::shell, also tested it with powershell on Windows and zsh & bash on Linux.
- [x] have tested using **Windows**

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -236,6 +236,7 @@ $jobs\
 $battery\
 $time\
 $status\
+$shell\
 $character"""
 ```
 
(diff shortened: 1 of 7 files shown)


cappyzawa (approved):
LGTM 👍

andytom:
Thank you for the contribution @rosowskimik and thank you for reviewing @cappyzawa.

PR #1725, opened 2020-10-02
feat: add support for elvish shell

davidkna:
#### Description

Adds support for the elvish shell.
This does not yet work with the currently released version of elvish because only `HEAD` (v0.15) supports ANSI color escapes in the prompt.
elvish has an error system and does not expose exit codes.

#### Motivation and Context

Closes #1115

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #2334, opened 2021-02-19
feat: Add nushell installation instructions

ibraheemdev:
This PR adds nushell installation instructions to the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,14 @@ shown below. Can't see yours? Have a look at the [extra platform instructions](h
    eval $(starship init ion)
    ```
 
+   #### Nushell
+
+   Run the following command:
+
+   ```sh
+   config set prompt "starship prompt"
+   ```
+
    #### Elvish
 
    **Warning** Only elvish v0.15 or higher is supported.
--- a/docs/README.md
+++ b/docs/README.md
@@ -91,6 +91,14 @@ description: Starship is the minimal, blazing fast, and extremely customizable p
    ```sh
    Invoke-Expression (&starship init powershell)
    ```
+   
+   #### Nushell
+
+   Run the following command:
+
+   ```sh
+   config set prompt "starship prompt"
+   ```
 
    #### Ion
 
@@ -114,4 +122,3 @@ description: Starship is the minimal, blazing fast, and extremely customizable p
 
    eval (starship init elvish)
    ```
-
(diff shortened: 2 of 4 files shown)


davidkna:
There's already for this at #1717. This wasn't merged yet because the level of integration between starship and nushell with just these instructions is rather low (#1716,  nushell/nushell#2703, nushell/nushell#2733).

wycats:
@davidkna we'd be happy to prioritize improving integration with Starship so that you would be willing to merge the PR. What exactly would we need to do?

davidkna:
Things that might are missing from the nu side:
1) I think command duration (`--cmd-duration`) should be supported. I just recently made a more elaborate feature request on that (nushell/nushell#3080)
1) Exit status of the last command (`--status`) would be very nice to have exposed, but elvish doesn't support this either so I think  it should suffice to list nu along with elvish in the warning in the docs

Things that are missing from this PR:
1) I'm not sure the current `starship init` approach like for all the other shells would work with nu. I think it would be best if there was something like a `$STARSHIP_INIT_VERSION` env var to notify users on init updates, because I expect they'll eventually happen.
1) I remember nu had some issues with unset PWD? Is that fixed or does this need a workaround with `--path` and `--logical-path`?
1) Does nu re-render the prompt on vi input mode switch now? If so the current input mode (default/input) should forwarded to starship (`--keymap`).
1) Does nu support jobs yet? If so the number of jobs should be forwarded to starship (`--jobs`)
1) This  PR should also handle setting the `$STARSHIP_SESSION_KEY` (a random session key to avoid duplicating log messages)  and `$STARSHIP_SHELL` env vars. There's also a few rust files that need to be changed to properly support a new shell too (see #2288 and #1725 for reference)

davidkna:
I looked around a bit and I think it will be possible to use the regular `starship init` style for nu once it supports sourcing from stdin or eval:

```toml
startup = [
    # Todo: figure out sourcing from starship init
    """
    def starship_prompt [] {
        starship prompt
    }
    """
]

prompt = "starship_prompt"
```

jonathandturner:
@davidkna - looking at that list, looks like folks are already starting to add some of those features. The exit code/status shouldn't be too bad to add also.

We set PWD when the path changes. Would be nice to confirm this meets what starship needs.

In theory, when the user configures the prompt, they should be able to get the input mode and pass that to the keymap flag.

For your `source` command example, we're currently treating `source` like an import or `use` statement, so we need to know statically where we're loading from. Need to think a bit more about that to see if we can be a little more flexible to allow for dynamically sourcing.

davidkna:
@jonathandturner It looks like the issues concerning PWD have already been fixed in nu for a while, so the current behaviour should be fine.

PR #2372, opened 2021-02-23
docs: Updated commands for chocolatey

### LINKED ISSUE #1945: Update documentation for installing starship on Windows
No. The installation guidelines in the website mention a very hacky way of installation starship on windows (poweshell). The README.md also states the same.

> Add the following to the end of Microsoft.PowerShell_profile.ps1. You can check the location of this file by querying the >$PROFILE variable in PowerShell. Typically the path is ~\Documents\PowerShell\Microsoft.PowerShell_profile.ps1 or >~/.config/powershell/Microsoft.PowerShell_profile.ps1 on -Nix.

>`Invoke-Expression (&starship init powershell)`

Instead, I noticed that [starship is present on chocolatey](https://chocolatey.org/packages/starship.portable). The installation on powershell would then simply be:

```powershell
choco install starship.portable
```
I would be interested in resolving this issue! 👍

nimishbongale:
#### Description
Adds the prerequisites and starship installation command for the chocolatey community package under "Advanced Installation". 

#### Motivation and Context
For the major purpose of including all possible ways to download and install starship.

Closes #1945 

star

#### Checklist:


docs/installing/README.md
```diff
@@ -12,13 +12,31 @@ There are so many platforms out there that they didn't fit into the main
 README.md file, so here are some installation instructions for other platforms
 from the community. Is yours not here? Please do add it here if you figure it
 out!
+
+## [Chocolatey](https://chocolatey.org)
+
+### Prerequisites
+
+```powershell
+Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
```
chipbuster, line 21: This is a chocolatey install instruction right?

I'd prefer not to have that directly in our install instructions, since that means that if the instructions ever change in the future, we'd need to modify our code to keep up. I'd much rather prefer a link to https://chocolatey.org/install and a note saying that you need to make sure chocolatey is installed.

nimishbongale: I've made the necessary changes @chipbuster. Do have a look! Thanks.

chipbuster (commented):
Just to confirm, if you install with chocolatey, does the installation tell you how to add the appropriate lines to the shell init file?

If not, then we need to communicate that here somehow, but I'm not sure what the best way to do that would be (since it would be the default instructions on the main page)

nimishbongale:
> Just to confirm, if you install with chocolatey, does the installation tell you how to add the appropriate lines to the shell init file?

@chipbuster 

The starship files download and adding to the path happens, but the starship init command must be levied at the end of the shell init, you're right. Just tagging @saanuregh to confirm the same.

Maybe I link back to the part which talks about the addition of the shell init command?

saanuregh:
The installation tells you how to add the appropriate lines to the shell init file at the end.

vladimyr:
@nimishbongale Why is this placed under advanced installation? There is hardly any difference between Homebrew, Scoop, and Chocolatey powered installation apart from package manager being used?

davidkna:
Advanced installation was introduced for package-manger specific installation instructions.

vladimyr:
@davidkna Then someone should move all those PM instructions there to avoid inconsistency. Otherwise, it is really hard to guess what's behind calling scoop less advanced than chocolatey 😀

vladimyr:
Furthermore, if this is a new home for PM powered installation instructions it does seem pretty awkward to link to install instructions of Chocolatey itself. It is safe to assume that folks already have their PM correctly installed before they try installing starship using it. Alternatively, we end up with a cluttered list linking back to Homebrew, Scoop, Nix, Termux, _you name it_ install instructions...

nimishbongale:
@vladimyr 

Your points do make sense. I initially proposed putting in all the instructions under the normal installation section itself, but I think the discussion swayed to having it as a part of the advanced installation.

I'm okay with either, and would love to know if anyone has any strong opinions about the same. Thanks!

chipbuster:
If the concern is about the semantics of "Advanced," I'd be happy to entertain a PR about changing the title of the section.

We've had a discussion about placing installation instructions for all platforms in the main README, and the decision was that we're not going to do that, so it's either a secondary page like this or not having these instructions at all.

chipbuster:
@nimishbongale Thank you for the contribution!

@saanuregh Thanks for the quick response and for maintaining the chocolatey package!

vladimyr:
@chipbuster My concern is not so much about the title of the section or its placement on the secondary page. It is about introducing an (unintended?) distinction between Homebrew, Scoop, and the rest of PMs. Former being placed as a prominent part of README versus latter on a secondary page may result in uninformed readers get the impression that they are split by some kind of support tier boundary.

vladimyr:
Again not only do we end up having inconsistency as a result of scattering PM-powered installation instructions around but with this particular PR, there is a new layer of inconsistency added on top of it by linking to Chocolatey install instructions. We either follow the suite by linking to Homebrew, Scoop, Termux install instructions (which is frankly completely unnecessary and just clutters docs) or we don't do it all including Chocolatey.

PR #2374, opened 2021-02-23
fix(python): Handle PyPy python version correctly

Tyilo:
#### Description

Add support for parsing PyPy python version.

#### Motivation and Context

Before this fix, I get the following prompt:

```
... via 🐍 v3.7.9 (7e6e2bb30ac5fbdbd443619cae28c51d5c162a02, Nov 24 2020, 10:03:59)
[PyPy 7.3.3-beta0 with GCC 10.2.0]
```

With the fix:

```
... via 🐍 v3.7.9
```

#### Checklist:

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -69,31 +69,30 @@ fn get_python_version(context: &Context, config: &PythonConfig) -> Option<String
     if config.pyenv_version_name {
         return Some(context.exec_cmd("pyenv", &["version-name"])?.stdout);
     };
-    let version = config.python_binary.0.iter().find_map(|binary| {
-        match context.exec_cmd(binary, &["--version"]) {
-            Some(output) => {
-                if output.stdout.is_empty() {
-                    Some(output.stderr)
-                } else {
-                    Some(output.stdout)
-                }
+    let version = config
+        .python_binary
+        .0
+        .iter()
+        .find_map(|binary| context.exec_cmd(binary, &["--version"]))
+        .map(|output| {
+            if output.stdout.is_empty() {
+                output.stderr
+            } else {
+                output.stdout
             }
-            None => None,
-        }
-    })?;
-    Some(format_python_version(&version))
+        })?;
+
+    format_python_version(&version)
 }
 
-fn format_python_version(python_stdout: &str) -> String {
-    format!(
-        "v{}",
-        python_stdout
-            .trim_start_matches("Python ")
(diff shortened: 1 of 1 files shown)


Tyilo:
Maybe the PyPy version should also be specified so that the output is `🐍 v3.7.9 (PyPy 7.3.3-beta0)` instead of `🐍 v3.7.9`?

chipbuster:
> Maybe the PyPy version should also be specified so that the output is `🐍 v3.7.9 (PyPy 7.3.3-beta0)` instead of `🐍 v3.7.9`?

I think this is fine as it is. The prompt already gets long enough without us trying to specify every single piece of version information--we can add a feature later to specify the exact python interpreter if needed, but that should probably be a separate feature PR.

src/modules/python.rs
```diff
@@ -89,8 +89,10 @@ fn format_python_version(python_stdout: &str) -> String {
         "v{}",
         python_stdout
             .trim_start_matches("Python ")
-            .trim_end_matches(":: Anaconda, Inc.")
             .trim()
+            .split_whitespace()
+            .next()
+            .unwrap_or("?")
```
vladimyr, line 95: Ideally, this shouldn't be more complex than:
```rust
    let version = python_version
        // split into ["Python", v3.8.6", ...]
        .split_whitespace()
        // return "v3.8.6"
        .nth(1)?;
```
but that is tricky to achieve due to inherited surroundings.

vladimyr (changes requested):
@Tyilo You did a fine job ensuring PyPy version gets correctly formatted applying minimal changes to the existing codebase but let's please improve the complete Python version extraction & formatting logic instead.

To help you with that here is a patch for you to apply: https://github.com/vladimyr/starship/commit/34b0302.patch (as seen here https://github.com/vladimyr/starship/commit/34b0302)

src/modules/python.rs
Tyilo: @vladimyr Done. I modified your patch slightly.

vladimyr: > @vladimyr Done. I modified your patch slightly.

Thank you for fixing my _documentation_ errors :+1:

vladimyr (approved):
LGTM
Well done @Tyilo!

davidkna:
Thanks for contribution @Tyilo and thanks for reviewing @vladimyr.

PR #2382, opened 2021-02-25
build: add more architectures to Snap package

### LINKED ISSUE #2381: Snap packages for arm64

I'm trying to install Starship on my shiny new Ubuntu VM on my M1 MacBook Air, but the snap packages aren't available for arm64 😱 

Screenshot 2021-02-25 at 17 38 29

The solution I would want to see would be for the snap to include arm64 binaries as well 🚀 

none 😅 

-----

I started looking into the snap yaml file, and it seems like we could just add arm64 here:

https://github.com/starship/starship/blob/4888ef3e0515655db0c3f360b41c60e0857aebbb/snapcraft.yaml#L78-L80

It seems like these are being built in the cloud somewhere, and while it would probably work it would be even cooler if we could use Rust's excellent ability to cross compile rather than spawning three VMs 😎

LinusU:
#### Description

I added `arm64` as a build platform to the snapcraft yaml. I also tried building this with `snapcraft build` on my arm64 Ubuntu installation, and it seemed to work 🎉 

#### Motivation and Context
Closes #2381 

I tested by running `snapcraft build`.

#### Checklist:

#### edit:

This now brings in all supported snap platforms, with an allowed failure for `s390x` which currently fails to build:

```text
	arch=arm64	state=Successfully built
	arch=armhf	state=Successfully built
	arch=i386	state=Successfully built
	arch=amd64	state=Successfully built
	arch=ppc64el	state=Successfully built
	arch=s390x	state=Failed to build
```

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -81,5 +81,6 @@ architectures:
   - build-on: armhf
   - build-on: i386
   - build-on: ppc64el
+    build-error: ignore
   - build-on: s390x
     build-error: ignore
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -76,5 +76,10 @@ apps:
       - removable-media
 
 architectures:
-  - build-on: i386
   - build-on: amd64
+  - build-on: arm64
+  - build-on: armhf
+  - build-on: i386
+  - build-on: ppc64el
+  - build-on: s390x
+    build-error: ignore


snapcraft.yaml
```diff
@@ -78,3 +78,4 @@ apps:
 architectures:
   - build-on: i386
   - build-on: amd64
+  - build-on: arm64
```
davidkna, line 81: Please also add `armhf` as starship also distributes binaries for that architecture.
```suggestion
  - build-on: arm64
  - build-on: armhf
```


I don't know if they build right now but you could also add `s390x` and `ppc64el`, but with `build-error: ignore`?
I think that's the syntax:
```suggestion
  - build-on: arm64
  - build-on: armhf
  - build-on: s390x
    build-error: ignore
  - build-on: ppc64el
    build-error: ignore
```

LinusU: Tried them all, this is the results:

```text
	arch=arm64	state=Successfully built
	arch=armhf	state=Successfully built
	arch=i386	state=Successfully built
	arch=amd64	state=Successfully built
	arch=ppc64el	state=Successfully built
	arch=s390x	state=Failed to build
```

Will update accordingly 🚀

LinusU:
@davidkna ready for another review ✔️

snapcraft.yaml
```diff
@@ -76,5 +76,10 @@ apps:
       - removable-media
 
 architectures:
-  - build-on: i386
   - build-on: amd64
+  - build-on: arm64
+  - build-on: armhf
+  - build-on: i386
+  - build-on: ppc64el
```
davidkna, line 83: Please ignore build errors for ppc too, as it is not actively tested.
```suggestion
  - build-on: ppc64el
    build-error: ignore
```

LinusU:
@davidkna sorry for the delay, I've updated the file now 🚀

vladimyr (approved):

davidkna:
Thanks for the contribution @LinusU and thanks for reviewing @vladimyr.

PR #2397, opened 2021-03-02
fix(package): fix meson version extraction if 'meson_version' is present

EHfive:
#### Description

Fix meson version extraction if 'meson_version' is present.

Previously, if `meson_version` is present after `version`, the meson project version will be extracted from `meson_version`.

#### Checklist:


andytom (approved):

andytom:
Thank you for your contribution @EHfive.

PR #2406, opened 2021-03-03
[WIP] feat: add tests and workflow for install script

### LINKED ISSUE #2405: feat: add tests / GitHub Actions workflows for install script
## Feature Request

Add tests for the install/bootstrap script across well-known shell environments, including:
* busybox (sh)
* mksh
* ksh
* dash
* ash
* zsh
* bash (ubuntu + macOS)

This will ensure compatibility of any change to the installation/bootstrap script across a wide range of shell environments in which starship is likely to be installed.

dmccaffery:
#### Description

add tests for the install/bootstrap script across well-known shell environments, including:
* busybox (sh)
* mksh
* ksh
* dash
* ash
* zsh
* bash (ubuntu + macOS)

#### Motivation and Context

Verify the installation across a wide-range of shells in which starship may be installed.

Closes #2405

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -1,0 +1,74 @@
+name: Installation
+on:
+  push:
+    paths:
+      - "install/**"
+  pull_request:
+    paths:
+      - "install/**"
+
+jobs:
+  check:
+    name: check
+    runs-on: ubuntu-20.04
+    container: koalaman/shellcheck-alpine:stable
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: shell
+        run: shellcheck install/install.sh
+
+  shell:
+    name: "Install in ${{ matrix.shell }}"
+    runs-on: ubuntu-20.04
+    needs:
+      - check
+    strategy:
+      matrix:
+        shell:
+          - dash
+          - bash
+          - zsh
+          - ash
+          - ksh
+          - mksh
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: install-shell
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y curl grep ${{ matrix.shell }}
+      - name: install
+        shell: ${{ matrix.shell }} -lc {0}
+        run: ./install/install.sh --force
+      - name: test
+        shell: ${{ matrix.shell }} -lc {0}
+        run: |
+          echo $0 | grep ${{ matrix.shell }} || exit 1
+          starship --version
+          command -v starship
+
+  macos:
+    name: "Install in bash on ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    needs:
+      - check
+    strategy:
+      matrix:
+        os:
+          - macos-11.0
+          - macos-10.15
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: install
+        shell: bash
+        run: ./install/install.sh --force
+      - name: test
+        shell: bash
+        run: |
+          echo $0 | grep bash || exit 1
+          $0 --version
+          starship --version
+          command -v starship
--- a/install/install.sh
+++ b/install/install.sh
@@ -23,7 +23,6 @@
 #     Override the base URL used for downloading releases
 
 set -eu
-set -o pipefail 2>/dev/null || true
 printf '\n'
 
 BOLD="$(tput bold 2>/dev/null || printf '')"
--- a/install/install.sh
+++ b/install/install.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env sh
 
-# shellcheck disable=SC2039
+# shellcheck disable=SC1090 # can't follow non-constant source
+# shellcheck disable=SC1091 # not following due to SC1090 (or other reasons)
+# shellcheck disable=SC2039 # something is undefined in posix, in this case the local keyword
 
 # Options
 #
--- a/install/install.sh
+++ b/install/install.sh
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env sh
+
+# shellcheck disable=SC2039
 
 # Options
 #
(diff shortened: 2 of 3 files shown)


install/install.sh
```diff
@@ -1,4 +1,8 @@
-#!/bin/bash
```
davidkna, line 1: Please rebase this.

davidkna (changes requested):

.github/workflows/install.yml
```diff
@@ -0,0 +1,74 @@
...
+    name: "Install in ${{ matrix.shell }}"
+    runs-on: ubuntu-20.04
+    needs:
+      - check
+    strategy:
+      matrix:
+        shell:
+          - dash
+          - bash
+          - zsh
```
davidkna, line 31: IIRC zsh doesn't work, and it isn't POSIX

.github/workflows/install.yml
```diff
@@ -0,0 +1,74 @@
+name: Installation
+on:
+  push:
+    paths:
+      - "install/**"
```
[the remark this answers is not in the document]
dmccaffery: It didn't run because the PR is in draft node. The matrix context object isn't available within the script block, so I have to write these as separate jobs -- didn't get time to do it yet.

.github/workflows/install.yml
```diff
@@ -0,0 +1,74 @@
...
+  macos:
+    name: "Install in bash on ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    needs:
+      - check
+    strategy:
+      matrix:
+        os:
+          - macos-11.0
+          - macos-10.15
```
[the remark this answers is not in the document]
dmccaffery: it is in preview; latest points to 10.15 at the moment -- thought we would want to check big sur anyway as it is out in the wild.

dmccaffery:
re: superlinter -- that's a 
massive (~2GB) toolchain. The only thing it has for shell outside of shellcheck is shfmt, which is a formatter. If you're in agreement with the default style of shfmt; I am happy to include. 

That being said, I don't personally recommend superlinter in situations where what is being linted / formatted is well-known -- it's designed as more of a generic "scan this monorepo for all the things" when you aren't sure what's going to be in there type of solution.

Do we want to use superlinter for everything, including for rust, markdown, etc? If so, then it makes a bit more sense.

Also, re: shellcheck container -- the container I'm using is the  container directly from the maintainer of shellcheck. I tend to avoid third-party wrappers of other third-party code as much as possible as there is less visibility into what is going on underneath. I'm happy to go either way as it's not my repo; just sharing the reasons behind the path I chose.

.github/workflows/install.yml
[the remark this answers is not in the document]
matchai: For the time being, it seems that macos-11.0 is in private preview: https://github.com/actions/virtual-environments/issues/2486
I'll look into seeing if I can get us in, but for the time being, it's available to us.

davidkna:
I mostly suggested super-linter because it is an official GitHub project, but it could certainly also be used elsewhere in the project. shfmt seems fine.

But the shellcheck container is fine too.

.github/workflows/install.yml
```diff
@@ -0,0 +1,74 @@
...
+      - check
+    strategy:
+      matrix:
+        shell:
+          - dash
+          - bash
+          - zsh
+          - ash
+          - ksh
+          - mksh
```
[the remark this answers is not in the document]
davidkna: Maybe you could do `ln -s $(which ${{ matrix.shell }}) /usr/local/bin/sh` and rerun the test with `sh`?

.github/workflows/install.yml
mirabilos: zsh has an `emulate sh` mode that brings it closely enough to POSIX.

See https://github.com/MirBSD/mksh/blob/463cce09ddaa8b11f839d2e5747f85ee412b1a76/Build.sh#L34-L41 how to portably enable it (this is `mksh`’s build script, which basically has to run with any `sh`, POSIX or not/before)…

.github/workflows/install.yml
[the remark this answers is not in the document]
mirabilos: For `mksh` as packaged in Debian and derivatives, this should be `/bin/lksh` though, which is the “more POSIXy” one.

install/install.sh
```diff
@@ -20,18 +24,18 @@
 #   -B, --base-url
 #     Override the base URL used for downloading releases
 
-set -euo pipefail
```
mirabilos, line 23: `set -o pipefail`, on shells that _do_ support it, is _very_ useful, though.

Many shells have `test ±o` but you can probably enable this portably by doing:

    set -eu
    (set -o pipefail) >/dev/null 2>&1 && set -o pipefail || :

install/install.sh
```diff
@@ -395,7 +407,7 @@ else
   VERBOSE=
 fi
 
-echo
+printf '\n'
```
mirabilos, line 410: This is nonsense, a simple `echo` will always work here, and spurious `printf` is penalising the shells that don’t have it as built-in utility.

`echo` is portable iff…
 - its first argument does not begin with a hyphen-minus `-`
 - there is no backslash in any of its arguments (not even escaped)

install/install.sh
```diff
@@ -224,7 +236,7 @@ detect_target() {
     target="${target}eabihf"
   fi
 
-  echo "${target}"
+  printf '%s' "${target}"
```
mirabilos, line 239: As for penalising shells with no built-in `printf`, there’s a nice way you can work around this for Korn shells at least:

```sh
if test -n "${KSH_VERSION:-}"; then
    write() {
        print -nr -- "$*"
    }
    writeln() {
        print -r -- "$*"
    }
else
    write() {
        printf '%s' "$*"
    }
    writeln() {
        printf '%s\n' "$*"
    }
fi
```

Then use the `write` and `writeln` helper functions everywhere a simple `echo` (see my previous comment) won’t do. This will be fast on all Korn shell variants (that set `$KSH_VERSION`; some set it as alias to `${.sh.version}` but you cannot check *that* portably) and on all shells with `printf` built in, which covers most (including many ash variants such as busybox and dash).

install/install.sh
```diff
@@ -261,7 +273,7 @@ check_bin_dir() {
     IFS=:
     for path in $PATH; do
       if [ "${path}" = "${bin_dir}" ]; then
-        echo 1
+        printf 1
```
mirabilos, line 276: This isn’t even the same thing! (`printf 1` omits the newline afterwards)

`1` does not start with `-` and does not contain `\` so it’s safe to use with `echo`.

install/install.sh
dmccaffery: This is an install script; not something that is executed constantly; optimising for performance is not the goal here. The goal is to be portable and posix compliant. If we mix echos and printf based on the exact conditions under which we believe are portable, then the next person that comes along and once to make a change will need to be armed with the same knowledge.

I'd vote to keep things consistent and simple; especially for an installation script that is run once per version (at most). I don't think the milliseconds saved per release with a builtin is worth the complexity.

install/install.sh
dmccaffery: The newline is not necessary.

dmccaffery: I don't care for the rudeness, by the way. I'm all for constructive feedback, but this isn't it. 

I made the changes due to actual failing scenarios where bash needed to be installed prior to using this install script. It was not portable. I made tests to prove it and made the script posix-compliant to resolve. It wasn't based on zealotry and it wasn't intended to be optimal.

I will abandon the PR; someone else is welcome to take a crack at it, but I don't have time to argue unnecessarily over mundane minutiae.

install/install.sh
dmccaffery: "Probably" doesn't cut it. Need to be sure.

mirabilos:
Deavon M. McCaffery dixit:

>"Probably" doesn't cut it. Need to be sure.

Right.

Just checked POSIX, it’ll indeed do (an unknown option causes
an errorlevel of >0, and the test in a subshell, with errors
redirected, first has precedent as well).

bye,
//mirabilos (mksh developer)

mirabilos:
Deavon M. McCaffery dixit:

>I don't care for the rudeness, by the way. I'm all for constructive
>feedback, but this isn't it.

Sorry, this wasn’t meant to be rude. A user pointed me to it in IRC
happy to see mksh support beginning, so I thought I’d point out a
few possible optimisations and things.

bye,
//mirabilos
-- 
„Cool, /usr/share/doc/mksh/examples/uhr.gz ist ja ein Grund,
mksh auf jedem System zu installieren.“
	-- XTaran auf der OpenRheinRuhr, ganz begeistert
(EN: “[…]uhr.gz is a reason to install mksh on every system.”)

PR #2409, opened 2021-03-03
feat(scala): Add scala module

### LINKED ISSUE #651: Add Scala / SBT on Starship
No, it is a new module

If inside an sbt project folder (can be identified by `build.sbt` in the project root), I would like to see the current scala (and maybe sbt version). Currently only the Java version is shown, although it is way less interesting compared to the sbt/scala version

#### Possible Pitfalls
* `sbt sbtVersion`, `sbt scalaVersion `sbt --version` are slow, they start an sbt server just to figure out the versions
* `sbt <command>` might lead to initiate the download of the specified version of sbt to the `.ivy2` folder if it is not on the system, which should not be triggered by a prompt tool.

Faster solution (but less reliable)
* Something like `grep 'scalaVersion' <projectRoot>/build.sbt` for a specified scala version, *usually* there is something like `scalaVersion := "2.13.0"`
* Something like `grep 'scalaVersion' <projectRoot>/project/build.properties` for a specified sbt version as in `sbt.version=1.3.0`

bbenzikry:
#### Description

Add a module for identifying scala version. 
Tested on scala 2 + dotty/scala 3

#### Motivation and Context

As a starship user, I'd like to be able to see my current scala version while happily traversing projects. 
Closes #651 ( at least partially ) 

It's important to note that: 
* Scala tooling is a bit on the slow side, so it may require changing ``command_timeout``, especially when using version shims like asdf

```
Benchmark #1: ./scala -version
  Time (mean ± σ):     536.7 ms ±  73.6 ms    [User: 703.9 ms, System: 101.9 ms]
  Range (min … max):   504.6 ms … 745.4 ms    10 runs

Benchmark #1: ./scalac -version
  Time (mean ± σ):     397.4 ms ±   5.2 ms    [User: 536.2 ms, System: 86.3 ms]
  Range (min … max):   390.3 ms … 408.0 ms    10 runs

Benchmark #1: starship module scala
  Time (mean ± σ):     402.6 ms ±   4.7 ms    [User: 540.2 ms, System: 90.5 ms]
  Range (min … max):   397.4 ms … 411.3 ms    10 runs
```
* This PR does not support sbt / project versions, as versions can be defined in a lot of different ways and sbt itself is slow ( at least during my tests ) to the point where I feel it won't be usable. 

```
Benchmark #1: sbt sbtVersion | tail -1 | awk '{print $2}'
  Time (mean ± σ):      4.730 s ±  0.137 s    [User: 14.113 s, System: 1.405 s]
  Range (min … max):    4.514 s …  4.984 s    10 runs
```

It may be possible to check ``project/build.properties`` but I believe it should be in a seperate PR that deals with the perf issues more thoroughly. 

image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2289,7 +2289,7 @@ By default the module will be shown if any of the following conditions are met:
 | `detect_extensions` | `["sbt", "scala"]`                         | Which extensions should trigger this module.     |
 | `detect_files`      | `[".scalaenv", ".sbtenv", "build.sbt"]`    | Which filenames should trigger this module.      |
 | `detect_folders`    | `[".metals"]`                              | Which folders should trigger this modules.       |
-| `symbol`            | `"🆂 "`                                    | format string representing the symbol of Scala   |
+| `symbol`            | `"🆂 "`                                    | A format string representing the symbol of Scala.   |
 | `style`             | `"red dimmed"`                             | The style for the module.                        |
 | `disabled`          | `false`                                    | Disables the `scala` module.                     |
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2277,7 +2277,7 @@ The `scala` module shows the currently installed version of Scala.
 By default the module will be shown if any of the following conditions are met:
 
 - The current directory contains a `build.sbt`, `.scalaenv` or `.sbtenv` file
-- The current directory contains a file with the `.scala` or `.scalarc` extension
+- The current directory contains a file with the `.scala` or `.sbt` extension
 - The current directory contains a directory named `.metals`
 
 ### Options
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -216,6 +216,7 @@ $purescript\
 $python\
 $ruby\
 $rust\
+$scala\
 $swift\
 $terraform\
 $vagrant\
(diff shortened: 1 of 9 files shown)


src/modules/scala.rs
```diff
@@ -0,0 +1,172 @@
+use std::fs;
```
vladimyr, line 1: This is unused. as Clippy already reported:
```suggestion
```

src/modules/scala.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use crate::configs::scala::ScalaConfig;
+use crate::formatter::StringFormatter;
+
+use super::{Context, Module, RootModuleConfig};
+
+use regex::Regex;
+
+const SCALA_VERSION_PATTERN: &str = "version[\\s](?P<version>[0-9\\.\\-A-z]+)";
```
vladimyr, line 10: You can rely on a fact that version can't contain whitespace:
```suggestion
const SCALA_VERSION_PATTERN: &str = "version[\\s](?P<version>[^\\s]+)";
```

src/modules/scala.rs
```diff
@@ -0,0 +1,172 @@
...
+            return None;
+        }
+    });
+
+    Some(module)
+}
+
+fn get_scala_version(context: &Context) -> Option<String> {
+    let scala_command = String::from("scalac");
+    let output = context.exec_cmd(&scala_command.as_str(), &["-version"])?;
```
vladimyr, line 59: This can be simplified to:
```suggestion
    let output = context.exec_cmd("scalac", &["-version"])?;
```

src/modules/scala.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    Some(format!("v{}", &version))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test::ModuleRenderer;
+    use ansi_term::Color;
+    use std::fs::File;
```
vladimyr, line 82:
```suggestion
    use std::fs::{self, File};
```

vladimyr (changes requested):
Looks good! :+1:

vladimyr (approved):

docs/config/README.md
```diff
@@ -2269,6 +2270,49 @@ The module will be shown if any of the following conditions are met:
...
+### Options
+
+
+| Option              | Default                                    | Description                                      |
+| ------------------- | -------------------------------------------| -----------------------------------------------  |
+| `format`            | `"via [${symbol}(${version} )]($style)"`   | The format for the module.                       |
+| `detect_extensions` | `["sbt", "scala"]`                         | Which extensions should trigger this module.     |
+| `detect_files`      | `[".scalaenv", ".sbtenv", "build.sbt"]`    | Which filenames should trigger this module.      |
+| `detect_folders`    | `[".metals"]`                              | Which folders should trigger this modules.       |
+| `symbol`            | `"🆂 "`                                    | format string representing the symbol of Scala   |
```
davidkna, line 2292:
```suggestion
| `symbol`            | `"🆂 "`                                    | A format string representing the symbol of Scala.   |
```

davidkna (approved):

davidkna:
Thanks for the contribution @bbenzikry and thanks for reviewing @vladimyr.

PR #2411, opened 2021-03-04
chore(github-actions): Remove AUR deploy

ATiltedTree:
#### Description

The AUR package was adopted as a community package so we don't need to run the AUR deploy action anymore:
https://archlinux.org/packages/community/x86_64/starship/

Also some small YAML formatting changes.

#### Motivation and Context

I just saw it was adopted so I created this PR

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -38,14 +38,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Setup | Checkout
-      uses: actions/checkout@master
+      - name: Setup | Checkout
+        uses: actions/checkout@master
 
-    - name: Publish
-      uses: netlify/actions/build@master
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


chipbuster:
I guess this is the best place to have this discussion: should we also remove `starship-bin` in that case? I doubly don't see the point of that on Arch now, since you can get pretty much the exact same product via the official repos.

Also, do you know what the standard procedure is for providing an AUR package that shadows the official repo, but is available on multiple architectures?

ATiltedTree:
Going after the [wiki](https://wiki.archlinux.org/index.php/AUR_submission_guidelines#Rules_of_submission), there shouldn't be any AUR package shadowing repo packages. And since `starship` [is already packaged in Archlinux ARM](https://archlinuxarm.org/packages?q=starship), I don't think that there is a need for a `-bin` package, since that are all the architectures we support anyways.

chipbuster:
I think I got a request for PPC support at some point...regardless, I think `starship-git` (which is not under my control) should be enough for anyone who really needs to build from source.

chipbuster (approved):
:+1:

PR #2413, opened 2021-03-04
feat(install): add code-signing for macOS

davidkna:
#### Description

Sign the starship binary for the local machine when using the install script to install macOS binaries when running macOS on an ARM processor. This allows running the starship binary installed by the install script directly without further workarounds or security warnings.
Intel macOS might also need this in the future.

I only tested the script with the Intel binaries on ARM, but that seemed to work. If you remove the ARM architecture check the script can also be tested on Intel macOS.

#### Motivation and Context

Related #1774

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -152,6 +152,56 @@ elevate_priv() {
   fi
 }
 
+# This functions handles signing the starship binary locally on macOS
+sign() {
+  local sudo
+
+  local bin
+  local rc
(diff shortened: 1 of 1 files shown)


vladimyr:
@davidkna Please take a look at my [comment](https://github.com/starship/starship/issues/1774#issuecomment-792409660) back inside the original issue thread and let's continue the discussion there.

PR #2425, opened 2021-03-07
feat: Right Prompt

rickvanprim:
#### Description
Proof of concept for generic right prompt support.  The logic is as follows:
1. Divide left/right prompts up on a per line basis.
2. Calculate the column widths for left and right lines.
3. Output left line unconditionally.
4. If the terminal width is greater than the left line width, right line width, and minimum spacing between them, insert padding spaces to right align the right line, then output the right line.
5. If there are more left lines remaining, or there was a right line, emit a line break.  A right line necessitates a line break as it doesn't appear that we can reset the cursor position without the shell clearing all the text that appears after it.

If this feature and approach are something Starship is interested in, then I'll update documentation, add tests, etc.

#### Motivation and Context
This adds an approximation of the right prompt functionality that exists in Zsh.

Notable difference is Zsh actually has the concept of a right prompt, which allows it to exist on the same line as cursor, and the right prompt text is able to hide/show depending on whether the cursor's position overlaps with it.

image

The golden path has been tested on:
 * Windows with Windows Terminal in Nushell (manually invoked, as the prompt doesn't seem to be via TTY and as such `term_size::dimensions()` returns None).
 * WSL with Windows Terminal/VSCode Terminal in Zsh.

#### Checklist:

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -1,4 +1,5 @@
 use crate::context::Shell;
+use crate::print::UnicodeWidthGraphemes;
 use crate::segment::Segment;
 use crate::utils::wrap_colorseq_for_shell;
 use ansi_term::{ANSIString, ANSIStrings};
@@ -123,6 +124,14 @@ impl<'a> Module<'a> {
             .iter()
             // no trim: if we add spaces/linebreaks it's not "empty" as we change the final output
             .all(|segment| segment.value.is_empty())
+    }
+
+    /// Get values of the module's segments
+    pub fn get_segments_width(&self) -> usize {
+        self.segments
+            .iter()
+            .map(|segment| segment.value.width_graphemes())
+            .sum()
     }
 
     /// Get values of the module's segments
--- a/src/print.rs
+++ b/src/print.rs
@@ -79,50 +79,106 @@ pub fn get_prompt(context: Context) -> String {
         buf.push_str("\x1b[J"); // An ASCII control code to clear screen
     }
 
-    let formatter = if let Ok(formatter) = StringFormatter::new(config.format) {
-        formatter
-    } else {
-        log::error!("Error parsing `format`");
-        buf.push('>');
-        return buf;
-    };
-    let modules = formatter.get_variables();
-    let formatter = formatter.map_variables_to_segments(|module| {
-        // Make $all display all modules
-        if module == "all" {
-            Some(Ok(PROMPT_ORDER
-                .par_iter()
-                .flat_map(|module| {
-                    handle_module(module, &context, &modules)
-                        .into_iter()
-                        .flat_map(|module| module.segments)
-                        .collect::<Vec<Segment>>()
-                })
-                .collect::<Vec<_>>()))
-        } else if context.is_module_disabled_in_config(&module) {
-            None
-        } else {
-            // Get segments from module
-            Some(Ok(handle_module(module, &context, &modules)
-                .into_iter()
-                .flat_map(|module| module.segments)
-                .collect::<Vec<Segment>>()))
+    let left_lines = match config
+        .format
+        .lines()
+        .map(|line| StringFormatter::new(line))
+        .collect::<Result<Vec<StringFormatter>, _>>()
+    {
+        Ok(left_lines) => left_lines,
+        Err(_) => {
+            log::error!("Error parsing `format`");
+            buf.push('>');
+            return buf;
         }
-    });
+    };
 
-    // Creates a root module and prints it.
-    let mut root_module = Module::new("Starship Root", "The root module", None);
-    root_module.set_segments(
-        formatter
-            .parse(None)
-            .expect("Unexpected error returned in root format variables"),
-    );
+    let right_lines = match config
+        .format_right
+        .lines()
+        .map(|line| StringFormatter::new(line))
+        .collect::<Result<Vec<StringFormatter>, _>>()
+    {
+        Ok(right_lines) => right_lines,
+        Err(_) => {
+            log::error!("Error parsing `format_right`");
+            buf.push('>');
+            return buf;
+        }
+    };
 
-    let module_strings = root_module.ansi_strings_for_shell(context.shell);
     if config.add_newline {
         writeln!(buf).unwrap();
     }
-    write!(buf, "{}", ANSIStrings(&module_strings)).unwrap();
+
+    let mut left_lines = left_lines.into_iter();
+    let mut right_lines = right_lines.into_iter();
+    loop {
+        let left_line = left_lines.next();
+        let right_line = right_lines.next();
+        if left_line.is_none() && right_line.is_none() {
+            break;
+        }
+
+        let left_line_module = left_line.map(|formatter| compute_format(&context, formatter));
+        let right_line_module = right_line.map(|formatter| compute_format(&context, formatter));
+        let add_trailing_newline = right_line_module.is_some() || left_lines.len() > 0;
+
+        match (left_line_module, right_line_module) {
+            (Some(left_line_module), Some(right_line_module)) => {
+                let left_line_width = left_line_module.get_segments_width();
+                let right_line_width = right_line_module.get_segments_width();
+                if left_line_width + config.split_padding + right_line_width <= context.width {
+                    // To end cursor in correct position, we print right, carriage return, print left.
+                    let padding = context.width - (left_line_width + right_line_width);
+                    write!(
+                        buf,
+                        "{}{pad_char:padding$}{}",
+                        ANSIStrings(&left_line_module.ansi_strings_for_shell(context.shell)),
+                        ANSIStrings(&right_line_module.ansi_strings_for_shell(context.shell)),
+                        pad_char = ' ',
+                        padding = padding,
+                    )
+                    .unwrap();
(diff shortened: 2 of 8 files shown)


src/print.rs
```diff
@@ -42,50 +42,110 @@ pub fn get_prompt(context: Context) -> String {
         buf.push_str("\x1b[J"); // An ASCII control code to clear screen
     }
 
-    let formatter = if let Ok(formatter) = StringFormatter::new(config.format) {
```
rickvanprim, line 45: Extracted into a function below.

andytom:
> If this feature and approach are something Starship is interested in, then I'll update documentation, add tests, etc.

Hi @rickvanprim, thank you for starting work on this, I have only had a quick look but this seems to be heading in the right direction and how we have discussed implementing this in the past. I think this is definitely something we want to add so :+1: from me.

src/module.rs
```diff
@@ -120,6 +121,14 @@ impl<'a> Module<'a> {
             .all(|segment| segment.value.is_empty())
     }
 
+    /// Get values of the module's segments
+    pub fn get_segments_width(&self) -> usize {
+        self.segments
+            .iter()
+            .map(|segment| segment.value.width())
```
rickvanprim, line 128: Should this use `print::better_width()`?

rickvanprim:
What is the expected level of testing for this.  I see there's a mocked `ModuleRenderer`, but there doesn't appear to be anything mocking the whole process around generating a prompt.  Is adding that a requirement for this PR?  Or what would be an acceptable way to cover this change?

syphar:
I would love to have this landing in starship, and I'm happy to help if I can (at the least with testing on mac , zsh/bash)

With this configurat

ion, how would a future addition of `RPROMT` support look like? I mean to have right-side elements next to the cursor?

rickvanprim:
@syphar I whipped something up locally which lets you specify `--prompt-side=[both|left|right]` to split things up for Zsh.  However this doesn't actually work with multi-line stuff, as covered by https://superuser.com/questions/974908/multiline-rprompt-in-zsh

There's a behavior difference between Zsh and non-Zsh for right prompt, which is in non-Zsh you need to put a linebreak after it to reset the cursor.  That makes me think maybe this is less about being able to split up left/right, and maybe just having a bonus field for Zsh, as if I made it so that the bottom right prompt line got treated differently and passed to Zsh RPROMPT, then the behavior would be different between Zsh and non-Zsh for the same config.  Any thoughts/suggestions?

Milo123459:
Is there anyway I can help with the development of this? I'd love to see this in starship.

rickvanprim:
@Milo123459 I don't know.  I've never contributed to Starship before and thus far haven't heard anything from the maintainers.  Maybe because the review is just a Draft.  I'll hit `Ready for review` and see if that gets more attention.

Milo123459:
> @Milo123459 I don't know. I've never contributed to Starship before and thus far haven't heard anything from the maintainers. Maybe because the review is just a Draft. I'll hit `Ready for review` and see if that gets more attention.

Cool. Just mention me if I can help in anyway!

sim1:
@rickvanprim this looks great, are you using this? would you be able to rebase onto master? I see your branch is a bit behind and I'd love to try it out locally myself.

rickvanprim:
@sim1 PR is rebased, please give it a shot and let me know what you think 🙂  I haven't been using it lately as I've mainly been on my Mac with Zsh, but it was working when I was last playing with it on my Windows machine.

You may also want to check out https://github.com/rickvanprim/starship/tree/right-prompt-zsh .  This is an extension of this PR that splits generation into a left prompt and a right prompt for Zsh (and any other shells that support `RPROMPT`).  I never finished this as I started running into design uncertainty around inconsistent behavior between shells and the problem that left prompt needs to include both sides for everything except the text entry line.  I think this existing PR is more ideal behavior because it's consistent, though maybe people really like being able to have right prompt stuff on the same line as the input, in which case this other branch is worth looking at.

matchai (commented):
This is looking good to me so far! 👍
This seems like an elegant way to support left/right prompts in a shell-agnostic way. 😊 

Unfortunately, I'm having no luck on my end getting it to work locally. Seemingly some part of root config loading isn't pulling in `format_right`, despite being set. I'll try to do some digging on my end when I've got a little more time to spare.

As for testing, I we don't need anything too rigorous, but a few tests to make sure that we're correctly handling some basic cases would be appreciated:
- Correctly rendering left/right prompts on multi-line prompts
- Left-only and right-only prompts

davidkna:
> My impression is that `better_width()` would be more accurate when measuring character width. @davidkna, is that correct?

The width should be more correct, assuming that the terminal is rendering them correctly. But this will almost never be the case. But for the default config it should almost always be the same value.
I think something like the `width` options in fish or some kind of terminal quirk handling is needed to handle this correctly.

src/module.rs
rickvanprim: `print::better_width()` doesn't seem to exist anymore, so I'm assuming that `UnicodeWidthGraphemes::width_graphemes()` is the preferred approach, and will update to use that.

rickvanprim:
@sim1 please see https://github.com/starship/starship/pull/2786.

rickvanprim:
> This is looking good to me so far! 👍
> This seems like an elegant way to support left/right prompts in a shell-agnostic way. 😊
> 
> Unfortunately, I'm having no luck on my end getting it to work locally. Seemingly some part of root config loading isn't pulling in `format_right`, despite being set. I'll try to do some digging on my end when I've got a little more time to spare.
> 
> As for testing, I we don't need anything too rigorous, but a few tests to make sure that we're correctly handling some basic cases would be appreciated:
> 
> * Correctly rendering left/right prompts on multi-line prompts
> * Left-only and right-only prompts
> * Terminal width can't be identified (if it can be easily tested)

I added some tests if you don't mind taking a look.  I didn't do a terminal width can't be identified one, though I think that's covered by the too narrow test.  Happy to add another test if you prefer.

Also please take a look at https://github.com/starship/starship/pull/2787.  I just hacked on the `prompt()` method to `ModuleRenderer` for these tests, and I clean that up in this downstream PR.

matchai:
Hey @rickvanprim. Sorry for the wait! 😅
Giving this a shot locally, it looks like it's working wonderfully!

Two quick concerns I'd like to raise, which are both pretty fixable, and would make right-prompt a whole lot more convenient to use:

#### 1. Duplication of modules in `$all`

The most common thing I foresee people trying to do when trying out right-prompt is to quickly throw in a `format_right` to show the time or an environment variable:

```toml
format_right = "$time"

[time]
disabled = false
```

In fish shell, this is what happens... 😅 

CleanShot 2021-06-29 at 20 23 59@2x

So that's not great.
If a user is familiar enough with Starship and read through our docs, they may know that the `time` module exists in the format string. Okay! Let's just strip that out then.

```toml
format = """
$directory\
$git_branch\
$package\
$rust\
$line_break\
$character
"""
format_right = "$time"

[time]
format = "[$time](black-white)"
disabled = false
```

CleanShot 2021-06-29 at 20 31 20@2x

Huh... It's still buggy.
That brings us to-

#### 2. Treating `$line_break` as a newline

The PR works wonderfully if you hand-write a multi-line `format`, but doesn't seem to play nicely with the `line_break` module in `$all` or on its own in `format`. Ideally, `line_break` should be treated similarly to a newline in `format`.

This is looking fantastic so far, and I'm really looking forward to getting this merged!
Sorry for the long wait between reviews. I'll try to stay on top of this one until we get it through.😄

Milo123459:
Sorry for disturbing:

But there is a conflict. Might be good to get that solved!

Milo123459:
Oh also, rebase the PR. This pr needs to be rebased because there are a few random commits/changes. @rickvanprim

matchai:
@Milo123459 I appreciate the sentiment, but there's no need to comment or ping contributors when their PRs are in need of conflict resolution or rebasing. The GitHub UI and git CLI make that clear enough. 😊

rickvanprim:
@Milo123459 I've rebased my 3 open PR's.

@matchai I tried briefly to better handle `$all` and `$line_break`, but I don't see a path forward without considerable additional work.  I think `$all` would work best by starting with a duplicate of `PROMPT_ORDER` and removing any variables found in the prompt strings from it.  However I don't have access to the variables without parsing which is either happening multiple times or interferes with the current line splitting approach.  Likewise I realize this would fail if someone put a line break in the middle of a group or style specifier.  This would be a breaking change, but I think the best form of this would be to have `left_format` and `right_format` as arrays of strings, delete `$line_break`, and return a load/validation error if any of the format line strings contain line breaks.

I'm no longer feeling motivated to continue working on this feature, so for the time being this is as far as I'm taking it.  If someone else would like to take over, they're welcome to.  Otherwise feel free to close the PR.

Milo123459:
I'm happy to take over if needed.

mike-lloyd03:
I'd like to voice my support for this feature. I'm still a Rust newb but it would be great to have an RPROMPT option in Starship.

matchai:
I'll be closing this PR since it has reached a standstill.
This feature is free to be picked up if anyone wants to give it a go. 😊

PR #2428, opened 2021-03-07
fix(zsh): Set PROMPT just once

ericbn:
#### Description

No need to forcefully set the `PROMPT` variable every time the prompt is to be shown. Just set it once, leaving the command to be evaluated every time the prompt is to be shown, by enabling the `promptsubst` option.

Setting it once is also friendlier to users that want to experiment with another prompt theme by temporarily setting `PROMPT` to something else.  This would currently not be possible, because the variable is always reset before every prompt draw (precmd) and keymap change (zle-keymap-select).

Some other updates to take better advantage of the zsh script dialect:

  * `$` is not required to read variables inside `(( ))` arithmetic expressions.
  * The zsh dialect to check if a variable is set is `${+var}`. Better than `${var+1}`, which substitutes 1 if var is set, which is intended for more general substitutions, not just to check if var is set.
  * The number of jobs can be read using the `%j` escape sequence, which is expanded when the `promptpercent` option is set.

Also simplified a couple of code lines by avoiding a temporary `STARSHIP_START_TIME` variable, since we already have `STARSHIP_CAPTURED_TIME`.

#### Motivation and Context

My initial motivation was to avoid forcefully setting the `PROMPT` variable every time the prompt is to be shown, as this does not allow for overwriting it for, let's say for example, working on another prompt and testing how it looks like. I had to disable the whole starship initialization because it would not let me overwrite `PROMPT` not even temporarily.

Then I saw some refactoring could be done...

I didn't see any documentation or test that could be updated related to this. More than happy to update any, if I missed something.

#### Checklist:

### DIFF
--- a/src/init/starship.zsh
+++ b/src/init/starship.zsh
@@ -7,15 +7,6 @@
 # after drawing the prompt. This ensures that the timing for one command is only
 # ever drawn once (for the prompt immediately after it is run).
 
-zmodload zsh/parameter  # Needed to access jobstates variable for NUM_JOBS
-
-starship_render() {
-    # Use length of jobstates array as number of jobs. Expansion fails inside
-    # quotes so we set it here and then use the value later on.
-    NUM_JOBS=$#jobstates
-    PROMPT="$(::STARSHIP:: prompt --keymap="${KEYMAP-}" --status=$STARSHIP_CMD_STATUS --cmd-duration=${STARSHIP_DURATION-} --jobs="$NUM_JOBS")"
-}
-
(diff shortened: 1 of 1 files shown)


ericbn:
This working in Zsh 4 and 5. Now, I'm wondering, do we want to keep compatibility with Zsh 4? (And not even sure if this is compatible with versions earlier than that.) Zsh 5 is almost 9 years old already!

cappyzawa (approved):
LGTM

andytom:
Thank you for your contribution @ericbn and thank you for reviewing @cappyzawa

PR #2444, opened 2021-03-11
style(init): Cleanup the unknown shell message

andytom:
#### Description

Have make a small change to the message that is printed when an unknown
shell is used. This correct the placement of the trailing `"` so that
the two training new lines are correctly printed and updates the list of
supported shells.

#### Motivation and Context

Before:
```
❯ eval $(starship init fakesh)

fakesh is not yet supported by starship.
For the time being, we support bash, zsh, fish, and ion.
Please open an issue in the starship repo if you would like to see support for fakesh:
https://github.com/starship/starship/issues/newnn
```

After:
```
❯ eval $(target/debug/starship init fakesh)

fakesh is not yet supported by starship.
For the time being, we support the following shells:
* bash
* elvish
* fish
* ion
* powershell
* tcsh
* zsh

Please open an issue in the starship repo if you would like to see support for fakesh:
https://github.com/starship/starship/issues/new

```

#### Checklist:

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -193,9 +193,17 @@ fi"#,
         Some(shell_basename) => {
             println!(
                 "printf \"\\n{0} is not yet supported by starship.\\n\
-                 For the time being, we support bash, zsh, fish, and ion.\\n\
+                 For the time being, we support the following shells:\\n\
+                 * bash\\n\
+                 * elvish\\n\
+                 * fish\\n\
+                 * ion\\n\
+                 * powershell\\n\
+                 * tcsh\\n\
+                 * zsh\\n\
+                 \\n\
                  Please open an issue in the starship repo if you would like to \
-                 see support for {0}:\\nhttps://github.com/starship/starship/issues/new\"\\n\\n",
+                 see support for {0}:\\nhttps://github.com/starship/starship/issues/new\\n\\n\"",
                 shell_basename
             );
             None
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -81,12 +81,15 @@ init code. The stub produces the main init script, then evaluates it with
 pub fn init_stub(shell_name: &str) -> io::Result<()> {
     log::debug!("Shell name: {}", shell_name);
 
-    let shell_basename = Path::new(shell_name).file_stem().and_then(OsStr::to_str);
+    let shell_basename = Path::new(shell_name)
+        .file_stem()
+        .and_then(OsStr::to_str)
+        .unwrap_or(shell_name);
 
     let starship = StarshipPath::init()?;
 
     let setup_stub = match shell_basename {
-        Some("bash") => {
+        "bash" => {
             /*
              * The standard bash bootstrap is:
              *      `source <(starship init bash --print-full-init)`
@@ -132,14 +135,14 @@ fi"#,
 
             Some(script)
         }
-        Some("zsh") => {
+        "zsh" => {
             let script = format!(
                 "source <(\"{}\" init zsh --print-full-init)",
                 starship.sprint_posix()?
             );
             Some(script)
         }
-        Some("fish") => {
+        "fish" => {
             // Fish does process substitution with pipes and psub instead of bash syntax
             let script = format!(
                 "source (\"{}\" init fish --print-full-init | psub)",
@@ -147,7 +150,7 @@ fi"#,
             );
             Some(script)
         }
-        Some("powershell") => {
+        "powershell" => {
             // Explanation of syntax:
             // &: Explicitly tells powershell to execute path with starship executable.
             //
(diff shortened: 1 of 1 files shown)


vladimyr (approved):

src/init/mod.rs
```diff
@@ -193,9 +193,17 @@ fi"#,
         Some(shell_basename) => {
```
davidkna, line 193: Since you're touching this it might make sense to also combine the `None` and `Some(shell_basename)` match branches.

vladimyr:
Hmmm, now we end up in a situation where unprocessed user input possibly becomes part of an error message. I'm not sure that is really desirable 🤔 Is there a way to put a constraint on the shell arg using clap?

src/init/mod.rs
vladimyr: > Is there a way to put a constraint on the shell arg using clap?

Seems there is https://docs.rs/clap/2.33.3/clap/struct.Arg.html#examples-38

vladimyr:
> Is there a way to put a constraint on the shell arg using clap?

Seems there is https://docs.rs/clap/2.33.3/clap/struct.Arg.html#examples-38

vladimyr:
Nah, ignore me, I forgot that shell value might be both name and path to the binary.

davidkna:
`shell_words::quote` could be used for escaping here (on the other hand that might not work well with non-posix shells).

src/init/mod.rs
```diff
@@ -180,23 +180,21 @@ fi"#,
...
+                 * powershell\\n\
+                 * tcsh\\n\
+                 * zsh\\n\
+                 \\n\
                  Please open an issue in the starship repo if you would like to \
-                 see support for {0}:\\nhttps://github.com/starship/starship/issues/new\"\\n\\n",
-                shell_basename
+                 see support for {0}:\\nhttps://github.com/starship/starship/issues/new\\n\\n\"",
+                shell_name
             );
```
davidkna, line 198: I think this will use for escaping (this will also get rid of any format string vulnerabilities).
```rust
            let quoted_arg = shell_words::quote(shell_name);
            println!(
                "printf \"\\n%s is not yet supported by starship.\\n\
                 For the time being, we support the following shells:\\n\
                 * bash\\n\
                 * elvish\\n\
                 * fish\\n\
                 * ion\\n\
                 * powershell\\n\
                 * tcsh\\n\
                 * zsh\\n\
                 \\n\
                 Please open an issue in the starship repo if you would like to \
                 see support for %s:\\nhttps://github.com/starship/starship/issues/new\\n\\n\" {0} {0}",
                 quoted_arg
            );
```

Re: combining `None` and `Some(unknown)`, you could also just ` let shell_basename = Path::new(shell_name).file_stem().and_then(OsStr::to_str).unwrap_or(shell_name);` to get rid of the `Option<_>`.

davidkna (approved):

src/init/mod.rs
```diff
@@ -162,41 +165,40 @@ fi"#,
...
+                 * powershell\\n\
+                 * tcsh\\n\
+                 * zsh\\n\
+                 \\n\
                  Please open an issue in the starship repo if you would like to \
-                 see support for {0}:\\nhttps://github.com/starship/starship/issues/new\"\\n\\n",
-                shell_basename
+                 see support for %s:\\nhttps://github.com/starship/starship/issues/new\\n\\n\" {0} {0}",
+                quoted_arg
             );
```
vladimyr, line 202: There is this wonderful macro crate https://github.com/dtolnay/indoc we could leverage to make it a bit more readable IMO :tada: 
```rust
printdoc! {
    r#"printf "\n%s is not yet supported by starship.
    For the time being, we support the following shells:
    * bash
    * elvish
    * fish
    * ion
    * powershell
    * tcsh
    * zsh
    
    Please open an issue in the starship repo if you would like to
    see support for %s:
    https://github.com/starship/starship/issues/new\n\n" {0} {0}"#,
    quoted_arg
};
```

davidkna: I think this would still need literal `\n` everywhere.

vladimyr: Nope, it is just one large string spanning multiple lines. You can do this in shell, just tried.

vladimyr: Here is the code: https://gist.github.com/vladimyr/16701e99f95c3ec1c0b7b489a2e9d3eb

And the output:
```
$ cargo run 2>/dev/null | sh 

wsh is not yet supported by starship.
For the time being, we support the following shells:
* bash
* elvish
* fish
* ion
* powershell
* tcsh
* zsh

Please open an issue in the starship repo if you would like to
see support for wsh:
https://github.com/starship/starship/issues/new


```

vladimyr: I obviously put an extra line break after _would like to_ :facepalm:

davidkna: Well this does actually seem to break it for at least tcsh and elvish.

vladimyr: Hmm, right, just got an unmatched quote error in `tcsh`. :disappointed: 

With that being said this shouldn't ever get print inside tcsh or elvsih but more importantly how do we even know that undetected shell supports `printf` in this form? Should we try to invoke `sh` here? :thinking:

davidkna: I think `printf` is reasonably common both as a builtin and as a binary in `$PATH` so there shouldn't be much need for that. This would also require escaping the whole command again.

vladimyr: Ah, yeah, forgot about POSIX `print` binary 👍
What can I say - I tried 🙃

PR #2451, opened 2021-03-11
feat(aws): add support for getting profile from awsu

piksel:
#### Description
Adds another override, like the `AWS_VAULT` one, that uses another environment variable to resolve the current AWS profile.

#### Motivation and Context

When using [awsu](https://github.com/kreuzwerker/awsu) the currently assumed role/profile is a helpful addition to the prompt.

#### Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -47,14 +47,14 @@ fn get_aws_region_from_config(context: &Context, aws_profile: Option<&str>) -> O
 }
 
 fn get_aws_profile_and_region(context: &Context) -> (Option<Profile>, Option<Region>) {
-    match (
-        context
-            .get_env("AWS_VAULT")
-            .or_else(|| context.get_env("AWS_PROFILE")),
-        context
-            .get_env("AWS_DEFAULT_REGION")
-            .or_else(|| context.get_env("AWS_REGION")),
-    ) {
+    let profile_env_vars = vec!["AWSU_PROFILE", "AWS_VAULT", "AWS_PROFILE"];
+    let profile = profile_env_vars
+        .iter()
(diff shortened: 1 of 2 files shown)


vladimyr (approved):

src/modules/aws.rs
```diff
@@ -48,9 +48,11 @@ fn get_aws_region_from_config(context: &Context, aws_profile: Option<&str>) -> O
...
 fn get_aws_profile_and_region(context: &Context) -> (Option<Profile>, Option<Region>) {
     match (
-        context
-            .get_env("AWS_VAULT")
-            .or_else(|| context.get_env("AWS_PROFILE")),
+        context.get_env("AWSU_PROFILE").or_else(|| {
+            context
+                .get_env("AWS_VAULT")
+                .or_else(|| context.get_env("AWS_PROFILE"))
+        }),
```
andytom, line 55: Rather than nesting the profile look up here maybe it would be better to iterate through a list of env vars using a find_map, something like

```rust
let profile_env_vars = vec!["AWSU_PROFILE", "AWS_VAULT", "AWS_PROFILE"];  
let profile = profile_env_vars.iter().find_map(|env_var| context.get_env(env_var));
```

piksel: Probably!  My rust-fu is pretty much at beginner level.

andytom: No worries we all have to start somewhere, I think writing this module might have been my first meaningful bit of rust 😀.

andytom (approved):

andytom:
Thank you for your contribution @piksel and thank you for the review @vladimyr.

PR #2455, opened 2021-03-12
fix(java): use consistent separators for java path

### LINKED ISSUE #2427: Java version missing on Windows 10
#### Current Behavior
For Java directories on Windows 10, the Java portion of the starship prompt works, but is missing the Java **version**. 

#### Expected Behavior
For Java directories on Windows 10, the current Java version should be displayed as part of the starship prompt.

Confirmed that `java` is on my path. Also, this works fine on Linux.

#### Possible Solution
On Windows 10, specifying `java.exe` *might* be necessary. Regardless, a test should be added for Windows 10.

#### Environment
- Starship version: 0.50.0
- powershell version: 7.1.2
- Operating system: Windows 10.0.19042
- Terminal emulator: Windows Terminal Preview Version: 1.7.572.0
- Git Commit Hash: 0558b38e
- Branch/Tag: v0.50.0
- Rust Version: rustc 1.49.0 (e1884a8e3 2020-12-29)
- Build Time: 2021-02-02 20:05:59
- JAVA_HOME: "C:\Program Files\AdoptOpenJDK\jdk-11.0.10.9-hotspot"
- Java Version: OpenJDK 64-Bit Server VM (11.0.10+9) for windows-amd64 JRE (11.0.10+9), built on Jan 20 2021 06:07:07 by "" with MS VC++ 15.9 (VS2017)

```powershell
# --- PowerShell Profile ---
#

# path {{{
#

# GNU Core Utils
$env:Path = $env:Path + ";C:\Program Files (x86)\GnuWin32\bin"

# rust / cargo
$env:Path = $env:Path + ";C:\Users\JUnderwo\.cargo\bin"

#
# end path }}}

# command-line {{{
#

# import some powershell modules
Import-Module PSReadLine
Import-Module PSFzf

# Chocolatey profile
$ChocolateyProfile = "$env:ChocolateyInstall\helpers\chocolateyProfile.psm1"
if (Test-Path($ChocolateyProfile)) {
    Import-Module "$ChocolateyProfile"
}

# vi-mode
Set-PSReadlineOption -EditMode vi

# zoxide
Invoke-Expression (& {
    $hook = if ($PSVersionTable.PSVersion.Major -lt 6) { 'prompt' } else { 'pwd' }
    (zoxide init --hook $hook powershell) -join "`n"
})

# starship prompt

#
# end command-line }}}

# misc {{{
#

#$env:PAGER = bat
$env:LESS = "-eiFMR"
$env:EDITOR= "nvim"

#
# end misc }}}

# aliases {{{
#

Set-Alias h Get-History | bat
Set-Alias m bat
Set-Alias lf ls
Set-Alias ll ls
Set-Alias vi nvim

#
# end aliases }}}

# functions {{{
#

# new guid, useful for MS Windows Terminal profiles
function guid {
    [guid]::NewGuid()
}

# unix-like "ln -s"
function lns ($target, $link) {
    New-Item -Path $link -ItemType SymbolicLink -Value $target
}

# windows-like "mklink"
function mklink {
    cmd /c mklink $args
}

function update {
    sudo choco upgrade all
}

function which($cmd) {
    (Get-Command $cmd).Definition
}

function whoami {
    (get-content env:\userdomain) + "\" + (get-content env:\username)
}

#
# end functions }}}

#
# vim: set ft=ps1 fdm=marker fdl=1:
```

#### Starship Configuration

```toml
# --- starship prompt config file ----------------------------------------------

# --- performance issues -------------------------------------------------------

# - if there are performance issues, remove the following line
#   $git_status\
# - if performance issues have been addressed, the following line should appear after $git_state
#   $git_status\

# --- prompt format ------------------------------------------------------------

format = """
$username\
$hostname\
$shlvl\
$kubernetes\
$directory\
$git_branch\
$git_commit\
$git_state\
$hg_branch\
$docker_context\
$package\
$cmake\
$dart\
$dotnet\
$elixir\
$elm\
$erlang\
$golang\
$helm\
$java\
$julia\
$nim\
$nodejs\
$ocaml\
$perl\
$php\
$purescript\
$python\
$ruby\
$rust\
$swift\
$terraform\
$zig\
$nix_shell\
$conda\
$memory_usage\
$aws\
$gcloud\
$env_var\
$crystal\
$custom\
$cmd_duration\
$line_break\
$jobs\
$battery\
$time\
$status\
$character"""

# --- prompt overrides and additions -------------------------------------------

[java]
symbol = " "
style = "red"

[custom.grails]
description = "The version of Grails used in this project"
command = "rg grailsVersion gradle.properties | cut -d'=' -f2"
files = ["grailsw", "gradle.properties"]
symbol = " "
style = "green"

[time]
disabled = false
```

andytom:
#### Description

This switches us from just appending `/bin/java` to `$JAVA_HOME` to
treating `$JAVA_HOME` as a path. This should fix any issues on Windows
where $JAVA_HOME might use `\` rather than `/`.

#### Motivation and Context

Related to #2427

#### Checklist:

### DIFF
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -57,9 +57,9 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 fn get_java_version(context: &Context) -> Option<String> {
     let java_command = context
         .get_env("JAVA_HOME")
-        .and_then(|java_home| {
-            Path::new(&java_home)
-                .join("bin")
+        .map(PathBuf::from)
+        .and_then(|path| {
+            path.join("bin")
                 .join("java")
                 .into_os_string()
                 .into_string()
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -88,6 +88,7 @@ mod tests {
     use ansi_term::Color;
     use std::fs::File;
     use std::io;
+    use std::path::PathBuf;
 
     #[test]
     fn test_parse_java_version_openjdk() {
@@ -264,6 +265,27 @@ mod tests {
         File::create(dir.path().join(".java-version"))?.sync_all()?;
         let actual = ModuleRenderer::new("java").path(dir.path()).collect();
         let expected = Some(format!("via {}", Color::Red.dimmed().paint("☕ v13.0.2 ")));
+        assert_eq!(expected, actual);
+        dir.close()
+    }
+
+    #[test]
+    fn test_java_home() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        File::create(dir.path().join("Main.java"))?.sync_all()?;
+        let java_home: PathBuf = ["a", "b", "c"].iter().collect();
+        let java_bin = java_home.join("bin").join("java");
+
+        let actual = ModuleRenderer::new("java")
+            .env("JAVA_HOME", java_home.to_str().unwrap())
+            .cmd(&format!("{} -Xinternalversion", java_bin.to_str().unwrap()),
+            Some(CommandOutput {
+                stdout: "OpenJDK 64-Bit Server VM (11.0.4+11-LTS-sapmachine) for linux-amd64 JRE (11.0.4+11-LTS-sapmachine), built on Jul 17 2019 08:58:43 by \"\" with gcc 7.3.0".to_owned(),
+                stderr: String::new(),
+            }))
+            .path(dir.path())
+            .collect();
+        let expected = Some(format!("via {}", Color::Red.dimmed().paint("☕ v11.0.4 ")));
         assert_eq!(expected, actual);
         dir.close()
     }
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -1,6 +1,6 @@
 use crate::configs::java::JavaConfig;
 use crate::formatter::StringFormatter;
-use std::path::Path;
+use std::path::PathBuf;
 
 use super::{Context, Module, RootModuleConfig};
 
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -61,8 +61,9 @@ fn get_java_version(context: &Context) -> Option<String> {
             Path::new(&java_home)
                 .join("bin")
                 .join("java")
-                .to_str()
-                .map(str::to_owned)
+                .into_os_string()
+                .into_string()
+                .ok()
         })
         .unwrap_or_else(|| String::from("java"));
 
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -92,7 +92,6 @@ mod tests {
     use ansi_term::Color;
     use std::fs::File;
     use std::io;
-    use std::path::PathBuf;
 
     #[test]
     fn test_parse_java_version_openjdk() {
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -55,15 +55,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn get_java_version(context: &Context) -> Option<String> {
-    let java_command = match context.get_env("JAVA_HOME") {
-        Some(java_home) => {
-            let java_path = Path::new(&java_home).join("bin").join("java");
-            java_path.to_str().unwrap_or("java").to_owned()
-        }
-        None => String::from("java"),
-    };
+    let java_command = context
+        .get_env("JAVA_HOME")
+        .and_then(|java_home| {
+            Path::new(&java_home)
+                .join("bin")
+                .join("java")
+                .to_str()
+                .map(str::to_owned)
+        })
+        .unwrap_or_else(|| String::from("java"));
 
-    let output = context.exec_cmd(&java_command.as_str(), &["-Xinternalversion"])?;
+    let output = context.exec_cmd(&java_command, &["-Xinternalversion"])?;
     let java_version = if output.stdout.is_empty() {
         output.stderr
     } else {
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -1,5 +1,6 @@
 use crate::configs::java::JavaConfig;
 use crate::formatter::StringFormatter;
+use std::path::Path;
 
 use super::{Context, Module, RootModuleConfig};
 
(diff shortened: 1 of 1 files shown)


davidkna:
I think this is better, but I think the issue in #2427 is that `JAVA_HOME` is set to `C:\Program Files\AdoptOpenJDK\jdk-11.0.10.9-hotspot"` (trailing single quote). Windows should support both `/` and `\` equally.

davidkna (approved):
LGTM, but please add this test because `JAVA_HOME` is currently untested:

```rust
    #[test]
    fn test_java_home() -> io::Result<()> {
        File::create(dir.path().join("Main.java"))?.sync_all()?;
        let java_home: PathBuf = ["a", "b", "c"].iter().collect();
        let java_bin = java_home.join("bin").join("java");

        let actual = ModuleRenderer::new("java")
            .env("JAVA_HOME", java_home.to_str().unwrap())
            .cmd(&format!("{} -Xinternalversion", java_bin.to_str().unwrap()),
            Some(CommandOutput {
                stdout: "OpenJDK 64-Bit Server VM (11.0.4+11-LTS-sapmachine) for linux-amd64 JRE (11.0.4+11-LTS-sapmachine), built on Jul 17 2019 08:58:43 by \"\" with gcc 7.3.0".to_owned(),
                stderr: String::new(),
            }))
            .path(dir.path())
            .collect();
            let expected = Some(format!("via {}", Color::Red.dimmed().paint("☕ v11.0.4 ")));
        assert_eq!(expected, actual);
        dir.close()
    }
```

src/modules/java.rs
```diff
@@ -55,7 +56,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
 fn get_java_version(context: &Context) -> Option<String> {
     let java_command = match context.get_env("JAVA_HOME") {
```
vladimyr, line 58: Can you remove duplicated `java` fallback by using `map` instead?

andytom:
> I think this is better, but I think the issue in #2427 is that `JAVA_HOME` is set to `C:\Program Files\AdoptOpenJDK\jdk-11.0.10.9-hotspot"` (trailing single quote). Windows should support both `/` and `\` equally.

Looking at the issue again I think you might be correct. I'll update the description so that this PR doesn't close that issue. Maybe I'm remembering this wrong but I thought Windows had issues with mixed separators e.g `C:/some/example/path/` is fine and `C:\some\example\path` is fine but `C:/some\example/path` causes issues.

src/modules/java.rs
andytom: I've had a quick try but not seeing a clean way to do this, do you have a suggestion?

davidkna: How about this?
```rust
    let java_command = context
        .get_env("JAVA_HOME")
        .and_then(|java_home| {
            Path::new(&java_home)
                .join("bin")
                .join("java")
                .to_str()
                .map(str::to_owned)
        })
        .unwrap_or_else(|| "java".to_owned());
```

andytom: 👍🏻 much nicer than what I had

src/modules/java.rs
```diff
@@ -54,12 +55,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    let java_command = context
+        .get_env("JAVA_HOME")
+        .and_then(|java_home| {
+            Path::new(&java_home)
+                .join("bin")
+                .join("java")
+                .to_str()
+                .map(str::to_owned)
+        })
+        .unwrap_or(String::from("java"));
```
vladimyr, line 67: I just tested something similar locally:
```
    let java_command = context
        .get_env("JAVA_HOME")
        .map(PathBuf::from)
        .and_then(|path| {
            path.join("bin")
                .join("java")
                .into_os_string()
                .into_string()
                .ok()
        })
        .unwrap_or("java".to_string());
```

Honestly, I'm not that proficient in Rust to figure out what's better so just pasting it here for you to decide :shrug:

src/modules/java.rs
vladimyr: > I've had a quick try but not seeing a nice way to do this, do you have a suggestion?

Sorry, I missed your question. I was deep in the zone trying to figure it out :upside_down_face:

src/modules/java.rs
davidkna: I think using `into` might avoid extra allocations.

src/modules/java.rs
```diff
@@ -1,5 +1,6 @@
 use crate::configs::java::JavaConfig;
 use crate::formatter::StringFormatter;
+use std::path::Path;
```
vladimyr, line 3:
```suggestion
use std::path::PathBuf;
```

src/modules/java.rs
andytom: 🤷🏻 I've take a bit of both of your suggestions, thanks for your help.

src/modules/java.rs
```diff
@@ -54,12 +55,19 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-        Some(java_home) => format!("{}/bin/java", java_home),
-        None => String::from("java"),
-    };
-
-    let output = context.exec_cmd(&java_command.as_str(), &["-Xinternalversion"])?;
+    let java_command = context
+        .get_env("JAVA_HOME")
+        .and_then(|java_home| {
+            Path::new(&java_home)
+                .join("bin")
```
vladimyr, line 62:
```suggestion
        .map(PathBuf::from)
        .and_then(|path| {
            path.join("bin")
```

src/modules/java.rs
```diff
@@ -84,6 +92,7 @@ mod tests {
     use ansi_term::Color;
     use std::fs::File;
     use std::io;
+    use std::path::PathBuf;
```
vladimyr, line 95:
```suggestion
```

src/modules/java.rs
vladimyr: I've added a few more suggestions but I guess they are mostly cosmetic at this point so feel free to ignore them. :upside_down_face:

vladimyr (approved):

andytom:
Thanks for all your help @davidkna and @vladimyr.

PR #2456, opened 2021-03-12
fix(cmake): Change cmake symbol to unicode triangle

### LINKED ISSUE #2009: NerdFonts Codepoint Leakage
[Issue 365 on NerdFonts](https://github.com/ryanoasis/nerd-fonts/issues/365) affects the CMake symbol that we currently use in our module. There are two issues here:

- Users who are not using nerd-patched fonts will see a CJK symbol instead of however their terminal emulator displays an unknown symbol
- Users who use that section of the Unicode standard may have to choose between using our prompt and using the system in their language

We should keep this in mind moving forward and potentially pick a new CMake module symbol and avoid using glyphs located in the affected Unicode block until the issue is fixed by NerdFonts. Unfortunately, it looks like it's very much a breaking change from them, so it might be a while before that fix comes out.

### LINKED ISSUE #2446: CMake incorrect CMake symbol
#### Current Behavior

The 喝-symbol is currently used for CMake, however this gives me a Chinese valid symbol in my terminal, but not the intended triangle. 

#### Expected Behavior

Instead, I expect a triangle symbol, so use the correct unicode symbol for a triangle. Instead try to use the following symbol '△'.

image

### Solution

I created a Pull request to get this fixed. 

https://github.com/starship/starship/pull/2447

danger89:
Fixes #2446
Fixes #2009

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -454,7 +454,7 @@ the module will be activated if any of the following conditions are met:
 | Option              | Default                                | Description                                  |
 | ------------------- | -------------------------------------- | -------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"`   | The format for the module.                   |
-| `symbol`            | `"喝 "`                                | The symbol used before the version of cmake. |
+| `symbol`            | `"△ "`                                | The symbol used before the version of cmake. |
 | `detect_extensions` | `[]`                                   | Which extensions should trigger this moudle  |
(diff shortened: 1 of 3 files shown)


vladimyr (approved):

vladimyr:
@danger89 Please change the commit message from:
> Change cmake symbol to unicode triangle

to:
> fix(cmake): change cmake symbol to unicode triangle

so commit linter stops complaining about non-semantic commit title.

andytom:
Thank you for your contribution @danger89 and thank you for reviewing @vladimyr

PR #2464, opened 2021-03-16
feat(aws): temporary credentials countdown prompt

### LINKED ISSUE #1978: AWS credential "countdown"
The business I work for uses AWS STS tokens with one hour expiry. Countless times I am working on the command line and in the midst of a long running script, the token expires. I want to know at-a-glance if I should refresh my token before a long running operation.

`expiration = 2020-12-07 15:22:20-05:00` is stored in ~/.aws/credentials. With configurable time periods (e.g. 20m, 10m, 5m left) display a green, yellow, red key icon next to the AWS cloud icon. Alternatively, it could be just how many minutes or seconds are left like cmd_duration but inverse.

Writing a shell script to do it. I'm assuming Rust would be faster.

clvanbrandt:
#### Description
This PR adds the possibility to display the remaining duration of temporary AWS credentials in the prompt. The remaining duration is put between brackets after the region in seconds if the credentials are still valid or a customizable symbol if they are expired. 

The duration is retrieved in two different ways :
- by looking at the `AWS_SESSION_EXPIRATION` env variable which is set by `aws-vault` or the `AWSUME_EXPIRATION` which is set by AWsume.
- by looking for a `expiration` line in the profile section in the AWS credentials file (#1978).

#### Motivation and Context
Display remaining duration of the temporary credentials associated to the current profile.
Closes #1978 

#### Screenshots
Valid credentials
image

Expired credentials (obviously the symbol could be better)
image

Added some tests to test the new functionality. I ran `cargo test modules::aws` succesfully.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -277,12 +277,13 @@ date is read from the `AWSUME_EXPIRATION` env var.
 
 ### Variables
 
-| Variable | Example          | Description                          |
-| -------- | ---------------- | ------------------------------------ |
-| region   | `ap-northeast-1` | The current AWS region               |
-| profile  | `astronauts`     | The current AWS profile              |
-| symbol   |                  | Mirrors the value of option `symbol` |
-| style\*  |                  | Mirrors the value of option `style`  |
+| Variable | Example          | Description                                 |
+| -------- | ---------------- | ------------------------------------------- |
+| region   | `ap-northeast-1` | The current AWS region                      |
+| profile  | `astronauts`     | The current AWS profile                     |
+| duration | `2h27m20s`       | The temporary credentials validity duration |
+| symbol   |                  | Mirrors the value of option `symbol`        |
+| style\*  |                  | Mirrors the value of option `style`         |
 
 \*: This variable can only be used as a part of a style string
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -251,7 +251,7 @@ $character"""
 The `aws` module shows the current AWS region and profile. This is based on
 `AWS_REGION`, `AWS_DEFAULT_REGION`, and `AWS_PROFILE` env var with
 `~/.aws/config` file. This module also shows an expiration timer when using temporary
-credentials. w
+credentials.
 
 When using [aws-vault](https://github.com/99designs/aws-vault) the profile
 is read from the `AWS_VAULT` env var and the credentials expiration date
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -250,26 +250,30 @@ $character"""
 
 The `aws` module shows the current AWS region and profile. This is based on
 `AWS_REGION`, `AWS_DEFAULT_REGION`, and `AWS_PROFILE` env var with
-`~/.aws/config` file.
+`~/.aws/config` file. This module also shows an expiration timer when using temporary
+credentials. w
 
 When using [aws-vault](https://github.com/99designs/aws-vault) the profile
-is read from the `AWS_VAULT` env var.
+is read from the `AWS_VAULT` env var and the credentials expiration date
+is read from the `AWS_SESSION_EXPIRATION` env var.
 
 When using [awsu](https://github.com/kreuzwerker/awsu) the profile
 is read from the `AWSU_PROFILE` env var.
 
 When using [AWSume](https://awsu.me) the profile
-is read from the `AWSUME_PROFILE` env var.
+is read from the `AWSUME_PROFILE` env var and the credentials expiration
+date is read from the `AWSUME_EXPIRATION` env var.
 
 ### Options
 
-| Option           | Default                                           | Description                                                     |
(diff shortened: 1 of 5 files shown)


matchai:
I think that there may be a more user-friendly way to represent the countdown. Having such high precision as to present the "X thousand seconds remaining" isn't actionable most of the time.

Similarly to how, by default, the battery module only appears when it's actionable, perhaps we could show a symbol when there's < 2 mins left, and another when it has expired?

clvanbrandt:
I agree that the current version is a bit raw and your suggestion makes sense. 

How about making the threshold an option ? I feel like finding a definitive value might be hard.

andymac4182:
It would be great to support the AWsume expiration as well if possible. `AWSUME_EXPIRATION=2021-04-19T17:53:53` I just added a simple PR for adding profile support for AWsume. https://github.com/starship/starship/pull/2609

Not sure if this would be easy to add for aws-vault as well.

andymac4182:
@matchai What would you like done to get this in? It would be a big help so happy to give a hand here with this.

clvanbrandt:
@andymac4182 I added support for AWsume expiration variable

I also kind of like your suggestion about the time display. I think there are a lot of possibilities in that regard and I'd be happy to work on it if we can agree on a specific implementation.

andymac4182:
Thanks for that @clvanbrandt 😄 

Thinking out loud it would be good to look at adding an array configuration item. Ordered by size the largest one is the display one then it goes to the less than ones.

andymac4182:
@matchai Can you let us know what your preferred option would be? 

* Matching `cmd_duration` https://github.com/starship/starship/blob/master/src/modules/cmd_duration.rs#L54-L81
* Bracketing times

matchai:
@andymac4182 Both sound like good approaches to me. 👍 
Extracting the time formatter from `cmd_duration` for reuse will definitely come in handy for future modules.

clvanbrandt:
Now using the `cmd_duration` render time function. I changed its visibility to pub for now because I wasn't sure where to move it (or if it was necessary to move it at all). Example screenshot :

starship_credentials

andymac4182:
That looks great to me! Thanks heaps for the work. Does it still go to an `[X]` when it has expired?

andymac4182:
@matchai Can you please approve the workflows on this so they can run.

clvanbrandt:
> That looks great to me! Thanks heaps for the work. Does it still go to an `[X]` when it has expired? 

It does. The symbol used is also configurable. Anything to be changed about this ?

andymac4182:
This looks great to me :) I believe its just a review that is all that's required now.

src/modules/cmd_duration.rs
```diff
@@ -51,7 +51,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 // Render the time into a nice human-readable string
-fn render_time(raw_millis: u128, show_millis: bool) -> String {
+pub fn render_time(raw_millis: u128, show_millis: bool) -> String {
```
andymac4182, line 54: @matchai Should this be pulled into a shared file?

matchai: Yeah, I think it would make sense to move into https://github.com/starship/starship/blob/master/src/utils.rs 😊

docs/config/README.md
```diff
@@ -250,26 +250,30 @@ $character"""
...
+| ------------------- | ---------------------------------------------------------------- | ----------------------------------------------------------------- |
+| `format`            | `'on [$symbol($profile )(\($region\) )(\[$duration\])]($style)'` | The format for the module.                                        |
+| `symbol`            | `"☁️ "`                                                           | The symbol used before displaying the current AWS profile.        |
+| `region_aliases`    |                                                                  | Table of region aliases to display in addition to the AWS name.   |
+| `style`             | `"bold yellow"`                                                  | The style for the module.                                         |
+| `expiration_symbol` | `X`                                                              | The symbol displayed when the temporary credentials have expired. |
+| `disabled`          | `false`                                                          | Disables the `AWS` module.                                        |
 
 ### Variables
```
davidkna, line 279: `duration` needs to documented in the Variables section here.

matchai (approved):
Looks good to me! 😊 
Thank you Carl-Louis and Andrew for your contributions!

andymac4182:
Big thanks to @clvanbrandt This will help our team a lot.

andymac4182:
@clvanbrandt Not sure if you have a simple solution but I just tested this with `awsume` with the latest release and noticed this requires the `+10:00` on the end which it doesn't set as it just assumes local timezone. 

I haven't used chrono before so not sure if there is a simple fix for this.

clvanbrandt:
There is probably a way by adding an option to specify the time format (something else than `%Y-%m-%dT%H:%M:%S%z` that is currently assumed) but this might be too specific I don't know

PR #2465, opened 2021-03-17
perf(git_status): replace git2 in git status module with git cli

### LINKED ISSUE #1446: Replace the use of git2 with calls to the Git CLI
## Feature Request
Our biggest performance bottleneck has always been git2, performing considerably worse than calling `git` directly.

This task is to replace our existing use of git2 with calls to `git`, allowing us to remove the dependency altogether and improving performance across the board. If `git_status` improves in speed, the whole prompt improves in speed.

This is already implemented as a proof-of-concept in https://github.com/matchai/starship-poc/blob/master/src/vcs/git.rs, where we measured a considerable increase in prompt render time:

Replace the use of git2 functions with calls to `git status --porcelain` directly. This would also mean creating our own data structures for git-related data.

Fixes #891 
Fixes #597 
Fixes #1551

dodgez:
#### Description
The git_status module has been rewritten to directly call the git cli using `git status -b --porcelain=2 -uall (etc.)` and `git stash list`.  Specifically, the Repository and Status structs from git2 are no longer used (and therefore git2 isn't used in this module).  They have been "replaced" by passing the repo directory directly to the git commands and the porcelain info is extracted directly to determine necessary information.  The `RepoStatus` struct has been modified with this new method and the `GitStatusInfo` struct has also been slightly tweaked.

**Important** in the contributing guidelines [testing section](https://github.com/starship/starship/blob/master/CONTRIBUTING.md#testing), it mentions placing code in `src/utils.rs` when an external command is being run.  Given the dynamic nature of `git` and the tests of this module, it seemed to me mocking this command when testing would nullify the behavior of the tests in this module.  _I am not 100% sure about this and I want to make sure this gets addressed._

#### Motivation and Context
After using starship on WSL, I noticed substantial performance issues when the `git_status` module was enabled.  I really liked this feature and after some digging I found the below issue that mentions potentially migrating this module from libgit2 - which is slow - to the git cli directly.

This allows one to symlink git.exe in linux and get much better performance in WSL and it sounds like there are also improvements in general.

Closes #1446 

None

I tested my changes by running `cargo check`, `cargo fmt`, and `cargo test` many times.

I ran the tests many times throughout making these changes on WSL and Windows.  Windows fails but as mentioned below, there were a couple issues that seemed limited to emoji and superuser permissions.  I don't have MacOS, so I cannot test it there :(

Only `src/modules/git_status.rs` has been modified but the issue linked above mentions some git structures already in place.  It may be useful to merge that code with this new code.

- [X] I have tested using **Linux** (through WSL)
- [X] I have tested using **Windows** (caveat: some tests failed but these seemed to be related to emoji and superuser permissions)

#### Checklist:
I don't think there are any updates to documentation (that isn't auto-generated) and the tests seem to be working still.  However, please let me know if I am blatantly missing something.
- [X] I have updated the documentation accordingly. (none to update afaik)
- [X] I have updated the tests accordingly.  (none to update afaik but see above note about mocking external git commands)

I haven't submitted any issues, PRs, etc for this project before, I am also new to rust, and this is some of my first activity in an open source project.  Please let me know if there is anything I can improve on and if I have missed anything.

### DIFF
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -203,26 +203,6 @@ fn get_repo_status(context: &Context, repo_root: &PathBuf) -> Result<RepoStatus>
     log::debug!("New repo status created");
 
     let mut repo_status = RepoStatus::default();
-    let config_output = context
-        .exec_cmd(
-            "git",
-            &[
-                "-C",
-                &repo_root.to_string_lossy(),
-                "config",
-                "--get",
-                "--local",
-                "status.showUntrackedFiles",
-            ],
-        )
-        .map_or("all".to_owned(), |output| output.stdout);
-
-    let untracked = "--untracked-files=".to_owned()
-        + if !config_output.trim().is_empty() {
-            config_output.trim()
-        } else {
-            "all"
-        };
     let status_output = context
         .exec_cmd(
             "git",
@@ -233,7 +213,6 @@ fn get_repo_status(context: &Context, repo_root: &PathBuf) -> Result<RepoStatus>
                 "--porcelain=2",
                 "--renames",
                 "--branch",
-                &untracked,
             ],
         )
         .ok_or_else(|| Error::new(ErrorKind::NotFound, "git status failed"))?;
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -129,7 +129,15 @@ impl<'a> GitStatusInfo<'a> {
     pub fn check_repository(&self) -> Option<bool> {
         let repo_root = self.repo.root.as_ref()?;
         self.context
-            .exec_cmd("git", &["-C", &repo_root.to_string_lossy(), "--no-optional-locks", "status"])
+            .exec_cmd(
+                "git",
+                &[
+                    "-C",
+                    &repo_root.to_string_lossy(),
+                    "--no-optional-locks",
+                    "status",
+                ],
+            )
             .map(|output| output.stderr.trim().is_empty())
     }
 
@@ -234,7 +242,13 @@ fn get_stashed_count(context: &Context, repo_root: &PathBuf) -> std::io::Result<
     let stash_output = context
         .exec_cmd(
             "git",
-            &["-C", &repo_root.to_string_lossy(), "--no-optional-locks", "stash", "list"],
+            &[
+                "-C",
+                &repo_root.to_string_lossy(),
+                "--no-optional-locks",
+                "stash",
+                "list",
+            ],
         )
         .ok_or_else(|| Error::new(ErrorKind::NotFound, "git stash failed"))?;
 
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -7,7 +7,6 @@ use crate::configs::git_status::GitStatusConfig;
 use crate::context::Repo;
 use crate::formatter::StringFormatter;
 use crate::segment::Segment;
-use std::io::{Error, ErrorKind, Result};
 use std::path::PathBuf;
 use std::sync::Arc;
 
@@ -114,7 +113,6 @@ struct GitStatusInfo<'a> {
     repo: &'a Repo,
     repo_status: OnceCell<Option<RepoStatus>>,
     stashed_count: OnceCell<Option<usize>>,
-    valid_repo: OnceCell<Option<bool>>,
 }
 
 impl<'a> GitStatusInfo<'a> {
@@ -124,26 +122,7 @@ impl<'a> GitStatusInfo<'a> {
             repo,
             repo_status: OnceCell::new(),
             stashed_count: OnceCell::new(),
-            valid_repo: OnceCell::new(),
         }
-    }
-
-    pub fn check_repository(&self) -> &Option<bool> {
-        self.valid_repo.get_or_init(|| {
-            let repo_root = self.repo.root.as_ref()?;
-            self.context
-                .exec_cmd(
-                    "git",
-                    &[
-                        "-C",
-                        &repo_root.to_string_lossy(),
-                        "--no-optional-locks",
-                        "rev-parse",
-                        "HEAD",
-                    ],
-                )
-                .map(|output| output.stderr.trim().is_empty())
-        })
     }
 
     pub fn get_ahead_behind(&self) -> Option<(usize, usize)> {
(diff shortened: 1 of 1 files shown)


dodgez:
Also, I apologize for the lack of git history.  I made most of these changes in a WSL clone but needed to bring it over to the Windows side and ended up pasting all of the changes I had made at once.

src/modules/git_status.rs
```diff
@@ -207,62 +205,64 @@ impl<'a> GitStatusInfo<'a> {
...
-        return Err(git2::Error::from_str("Repo has no status"));
-    }
-
-    statuses
-        .iter()
-        .map(|s| s.status())
-        .for_each(|status| repo_status.add(status));
+    let output = Command::new("git")
+        .current_dir(repo_root)
+        .args(&["config", "--get", "--local", "status.showUntrackedFiles"])
```
jankatins, line 215: Just fur my curiosity: Why make this a git config and not a starship one? Means one external call more...

davidkna:
There's also an unfinished version of this [in the repo](https://github.com/starship/starship/compare/libgit-to-git-cli). Using `context.exec_cmd` also enables timeouts so it would be best if you used that either way. 
Regarding WSL performance #2146 might also interest you.

jankatins:
Nice: 

``` 
[11:47:16] λ  starship timings

 direct_git_status  -  72ms  -   [!]
 docker_nas         -   7ms  -   [🐳: NAS]
 git_status         -   3ms  -   [!]
 username           -   1ms  -
 directory          -   1ms  -   ~/external/local-nas
 git_commit         -   1ms  -
 character          -  <1ms  -   λ
 git_branch         -  <1ms  -   on  main
 line_break         -  <1ms  -   \n
 time               -  <1ms  -   [11:47:22]
``` 

(direct_git_status is a custom module derived from spaceships git prompt -> essentially the same as this PR)

src/modules/git_status.rs
dodgez: The previous code used pulled this same git config option using git2 so I figured I would maintain that for compatibility.  I also thought about that extra external call.  It also makes sense to me for starship to use the git config option, but I can see arguments for both sides.

davidkna: Does `git` not respect its own option here? And if not, it might be faster to read the option directly in starship like before.

dodgez:
@davidkna this code doesn't need much git infrastructure so I wonder if it would be worth merging that rewrite once all of git2 is dealt with.  I am not entirely sure though.  Also, that is a fair point about timeouts and I think that will be my next step then in this PR (to use `context.exec_cmd`).  I listed a concern above about adding the faux output above, is that a problem here?  Namely, the contributing guide mentions putting mock output for a command if an external command is used; however, this would nullify the tests relying on git actually running afaik.  What do you think?

Thanks for linking that; I saw that issue just after I submitted the PR and I think it is a great option.  I also think this PR can solve some more general performance issues but with an added bonus to me personally.  I will keep an eye on it.

@jankatins I am not familiar with the previous timings, but if you say nice that sounds like a good thing!  I have replaced my WSL binary with this branch and it is working wonderfully.  The only concern is in cases where libgit2 was actually faster than executing git, but from the linked issue it sounds like the consensus is that this is not the case.

src/modules/git_status.rs
dodgez: That is a great point @davidkna.  I would believe git respects its option.  The reason I put this code here was before I made any changes the code was using git2 to pull the same option (so I believe at some point it read it from the repo config).  It may be worth just removing this altogether.

The starship configuration for showing the untracked files in the prompt applies independently.  The previous code pulled in all the information and then selectively pulled that information to put into the prompt, which is the same thing I am doing here.

Actually, now that I think about it, the way I am getting the current status removes the need for this command.  I think the old code had to supply this because it used git2 which may not have respected the current configuration.  I will remove this.

dodgez:
Hmm, when I run the timings of this specific PR on Windows, `git_status` takes upwards of 200ms.  This is compared to a baseline of about 55ms on branch `master`.

This branch:
image

On `master`:
image

I think this may be the process overhead when spinning up `git` cli.  However WSL performance is on par when not in windows land, but it is much better when on the Windows filesystem because I can use git.exe which generally performs faster.  For me, this is a show stopper performance issue for Windows.  I am not sure where time could be made up (if any).  Unless this has significant performance increases on Native Linux and MacOS (neither I can test), this may be dead in the water.

davidkna:
The main issue seems to be `check_repository` using `git status` thereby calling it multiple times.

dodgez:
I didn't even think about that, I will work on removing the redundant calls.  Thanks for the pointer @davidkna!

dodgez:
Although after looking at the code, the `get_repo_status` and `get_stashed` are the only two functions that call `check_repository` and they are `OnceCell`s so further calls to them will not induce any further calls on `check_repository`, correct?

davidkna:
Is `check_repository` needed at all?

dodgez:
Currently if `check_repository` were not executed, `get_repo_status` and `get_stashed` would throw an error (if the cd is not a git repo).  I may be able to remove it and change those functions to use options.

The one thing I can see that may be an issue is `exec_cmd` returns `None` if the command was not a success so we couldn't capture any output from `git status` and `git stash list` for debug purposes unless `log::trace!` were captured.  Do you think it is worth moving to options and ignoring errors that occur unless someone turned on trace logging?  We could still log that the command failed but not the output it gives with `log::debug!`.

Before I push those changes, I will see what performance increase we get by moving to options to see if it would even significantly decrease the timing.

dodgez:
It looks like removing `check_repository` brings the timing (on my machine) down from ~230ms to ~120ms after several tests.  That is a noticeable improvement.

However, this is still ~2x what the current implementation is but ~3x smaller than the initial commit implementation.  Shaving off ~60ms would bring it down to what the current implementation is on Windows.

jankatins:
My understanding of the problem with the current implementation is not that it's slow on average but that it blows up in certain repos (at least that's why I switched to a custom git status; unfortunately, I do not have such a repo at hand anymore :-/). So if the average goes up a bit but the multi seconds in certain repos do not happen anymore, that's IMO an improvement.

dodgez:
Ah, that would make sense, thanks!  Are there specific next steps for this PR?  At the moment I am happy with the current state, but if there are any requested changes I am more than willing to work on them.

matchai (approved):
Looks good to me!

Major perf improvements on larger repos with negligible regressions on smaller repos is a win in my books! 🚀

PR #2479, opened 2021-03-19
feat(ocaml): add switch variable to ocaml format

cdaringe:
#### Description

- support rendering the opam `switch`

#### Motivation and Context

- `switch` to me is more useful than ocaml `version`. i suspect users of local switches also would desire to see their switch vs ocaml version

image

1. installed my binary locally, ran it
2. unit test

#### Checklist:

### DIFF
--- a/src/modules/ocaml.rs
+++ b/src/modules/ocaml.rs
@@ -235,7 +235,7 @@ mod tests {
                format = "$switch"
             })
             .collect();
-        let expected = Some("test-switch-name".to_string());
+        let expected = Some("my-test.switch".to_string());
         assert_eq!(expected, actual);
         dir.close()
     }
--- a/src/modules/ocaml.rs
+++ b/src/modules/ocaml.rs
@@ -31,16 +31,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             })
             .map(|variable| match variable {
                 "switch" => {
-                    let stdout = context.exec_cmd("opam", &["switch", "-s", "show"])?.stdout;
+                    let stdout = context
+                        .exec_cmd("opam", &["switch", "-s", "--safe", "show"])?
+                        .stdout;
                     let switch_raw = stdout.trim();
-                    let path = Path::new(switch_raw);
-                    let normalized = if path.has_root() {
-                        // use named local switch, often prefixed with full path by opam
-                        path.file_name()?.to_str()?
-                    } else {
-                        switch_raw
-                    };
-                    Some(Ok(normalized.to_owned()))
+                    Some(Ok(Path::new(switch_raw).file_name()?.to_str()?.to_owned()))
                 }
                 "version" => {
                     let is_esy_project = context
--- a/src/modules/ocaml.rs
+++ b/src/modules/ocaml.rs
@@ -1,4 +1,5 @@
 use super::{Context, Module, RootModuleConfig};
+use std::path::Path;
 
 use crate::configs::ocaml::OCamlConfig;
 use crate::formatter::StringFormatter;
@@ -29,6 +30,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 _ => None,
             })
             .map(|variable| match variable {
+                "switch" => {
+                    let stdout = context.exec_cmd("opam", &["switch", "-s", "show"])?.stdout;
+                    let switch_raw = stdout.trim();
+                    let path = Path::new(switch_raw);
+                    let normalized = if path.has_root() {
+                        // use named local switch, oft prefixed with full path by opam
+                        path.file_name()?.to_str()?
+                    } else {
+                        switch_raw
+                    };
+                    Some(Ok(normalized.to_owned()))
+                }
                 "version" => {
                     let is_esy_project = context
                         .try_begin_scan()?
@@ -206,6 +219,23 @@ mod tests {
 
         let actual = ModuleRenderer::new("ocaml").path(dir.path()).collect();
         let expected = Some(format!("via {}", Color::Yellow.bold().paint("🐫 v4.10.0 ")));
+        assert_eq!(expected, actual);
+        dir.close()
+    }
+
+    #[test]
+    fn folder_with_ml_file_and_uses_switch_variable() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        File::create(dir.path().join("any.ml"))?.sync_all()?;
+
+        let actual = ModuleRenderer::new("ocaml")
+            .path(dir.path())
+            .config(toml::toml! {
+               [ocaml]
+               format = "$switch"
+            })
+            .collect();
+        let expected = Some("test-switch-name".to_string());
         assert_eq!(expected, actual);
         dir.close()
     }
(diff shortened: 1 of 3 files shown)


andytom (approved):
LGTM

src/modules/ocaml.rs
```diff
@@ -29,6 +30,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 _ => None,
             })
             .map(|variable| match variable {
+                "switch" => {
+                    let stdout = context.exec_cmd("opam", &["switch", "-s", "show"])?.stdout;
+                    let switch_raw = stdout.trim();
```
vladimyr, line 35: Just invoke trim on stdout directly. There is no need to introduce an additional variable.

src/modules/ocaml.rs
```diff
@@ -29,6 +30,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                    let path = Path::new(switch_raw);
+                    let normalized = if path.has_root() {
+                        // use named local switch, oft prefixed with full path by opam
+                        path.file_name()?.to_str()?
+                    } else {
+                        switch_raw
+                    };
```
vladimyr, line 42: I may be wrong about this but couldn't you just take the filename portion unconditionally? 🤔

src/modules/ocaml.rs
```diff
@@ -29,6 +30,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 _ => None,
             })
             .map(|variable| match variable {
+                "switch" => {
+                    let stdout = context.exec_cmd("opam", &["switch", "-s", "show"])?.stdout;
+                    let switch_raw = stdout.trim();
+                    let path = Path::new(switch_raw);
+                    let normalized = if path.has_root() {
+                        // use named local switch, oft prefixed with full path by opam
```
vladimyr, line 38: Don't abbreviate comments 🙃
```suggestion
                        // use named local switch, often prefixed with full path by opam
```

vladimyr (changes requested):

src/modules/ocaml.rs
cdaringe: i did this originally, but:

```
temporary value dropped while borrowed
consider using a `let` binding to create a longer lived value (rustc E0716)
```

the extra var is consistent w/ the other impls as well

src/modules/ocaml.rs
cdaringe: switches are absolute paths, or not path based. this check protects against mangling a non-path based switch name

src/modules/ocaml.rs
cdaringe: oh, oft is a word :) uncommon, but not abbreviated :)

src/modules/ocaml.rs
vladimyr: I understand your motivation but I don't think there is a way to mangle named switch by extracting filename from it. It seems that OPAM authors agree with me: https://github.com/ocaml/opam/blob/9f9e9297/doc/pages/Tricks.md#display-the-current-opam-switch-in-the-prompt-

cdaringe: ya, you're right. i did some tests locally, and i think it's all zen. updated in 279adb6

vladimyr:
@cdaringe I took your code and did the following:
1. Prefixed local switches with an asterisk as suggested inside [Opam docs](https://github.com/ocaml/opam/blob/9f9e9297/doc/pages/Tricks.md#display-the-current-opam-switch-in-the-prompt-)
2. Extracted switch parsing so it can be tested independently
3. Replaced global `opam switch show --safe` mock with per-test mocking so both global & local switch scenario can be emulated/tested

You can inspect it here: https://github.com/vladimyr/starship/commit/2d8620b
Patch download: https://github.com/vladimyr/starship/commit/2d8620b.patch

Please apply proposed changes unless you have some concerns or objections. :tada:

cdaringe:
@vladimyr, the `*` doesn't make sense to me. i think that is more appropriately used in situations like `git branch` to show the selected entity. i'd prefer to see my switch name vs "<junk-char><my-switch-name>".

i'd prefer to keep it as is.

vladimyr:
> @vladimyr, the `*` doesn't make sense to me. i think that is more appropriately used in situations like `git branch` to show the selected entity. i'd prefer to see my switch name vs `<junk-char><switch-name>`.

Junk char aka prefix is there so one could easily distinguish between global and local switch name because in reality there is no such thing as a local name - it is a (shortened) path instead. By removing prefix you are making it ambiguous.

And that's just a single bullet point, let's start with the rest we hopefully agree on?

andytom:
Personally I would prefer to leave the switch version without the prefix `*` since we will at some point want to allow users to customise it and I think it would be better to use the format string to do that
Something like
```toml
[ocaml] 
format = "*$switch" 
```

vladimyr:
> ```toml
> [ocaml] 
> format = "*$switch" 
> ```

This would work if the asterisk was mere decoration but it should be an indicator instead.

Let me take a step back and describe what switches really are. The switch is simply put a folder, subtree, prefix where OCaml binaries and libraries get installed. As such it can be global (shared between projects) or local (project-specific). Global switches are stored as named folders inside `$HOME/.opam` (opam's installation dir). Local on the other end, are just project subfolders named `_opam`. For example:
- global switch `foo` points to `$HOME/.opam/foo`
- local switch `bar` points to `/path/to/bar/_opam` (where `bar` is a _project name_)

To give you a better impression I made some switches locally, take a look:
image

As you can see opam by convention strips `$HOME/.opam` part when listing global switches while retaining full path for local ones. Otherwise, you wouldn't be able to tell the difference between global `foo` and local `foo` from the earlier screenshot!
Obviously listing the full path as a local switch name would make the prompt egregiously long so opam authors suggest the following smart trick https://github.com/ocaml/opam/blob/9f9e9297/doc/pages/Tricks.md#display-the-current-opam-switch-in-the-prompt-
> You can use e.g. the command `opam switch show --safe 2>/dev/null | sed 's|.*/|*|'` to return a (shortened) switch name.

The important part is the use of `sed` which roughly translates to strip all path segments from the start and swap them with an asterisk thus making `/home/vladimyr/scratchpad/ocaml/foo` become `*foo` while global `foo` stays the same.
Again, if you just take basename you are making it ambiguous!

So what we really need here is something along these lines:
```toml
[ocaml]
local_switch_indicator = "*" # or whatever you'd like to put here I've just c/p opam's authors choice
global_switch_indicator = ""
format = "$switch_indicator$switch_name"
```
By introducing these new config vars it becomes a truly configurable (and optional) indicator thus addressing @cdaringe concerns without any information losses.

The best thing is that this is relatively simple to implement on top of my earlier proposal:

> You can inspect it here: [vladimyr@2d8620b](https://github.com/vladimyr/starship/commit/2d8620b)
> Patch download: [vladimyr/starship@`2d8620b`.patch](https://github.com/vladimyr/starship/commit/2d8620b.patch)

andytom:
Thanks for explaining that @vladimyr, that makes a lot more sense to me. I think your suggestion of global and relative symbols makes the most sense going forward.

cdaringe:
hey guys. i'm AFK for a bit. i'm good w/ whatever. feel free to patch or close as seen fit

PR #2490, opened 2021-03-24
docs(readme): add missing link to home-manager

kidonng:
#### Description

The link to `home-manager` seems to be missing when the section was added.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/docs/installing/README.md
+++ b/docs/installing/README.md
@@ -49,7 +49,7 @@ curl -fsSL https://starship.rs/install.sh | bash -s -- -b /data/data/com.termux/
 nix-env -iA nixos.starship
 ```
 
-#### Declarative, single user, via [home-manager](home-manager)
+#### Declarative, single user, via [home-manager](https://github.com/nix-community/home-manager)
 
 Enable the `programs.starship` module in your `home.nix` file, and add your settings
 


docs/installing/README.md
```diff
@@ -49,7 +49,7 @@ curl -fsSL https://starship.rs/install.sh | bash -s -- -b /data/data/com.termux/
 nix-env -iA nixos.starship
 ```
 
-#### Declarative, single user, via [home-manager](home-manager)
+#### Declarative, single user, via [home-manager](https://github.com/nix-community/home-manager)
```
vladimyr, line 52: Linking to the Github repo is fine but other links point either to frontpages or wikis. In this case, we could follow the suit by linking to the manual https://nix-community.github.io/home-manager
@kidonng What do you think?

kidonng: I'm aware of the manual, but it directly jumps to installation and usage without explaining what is it and "Words of warning" like the README on GitHub does. This is friendlier for those who may not be familiar with what the `home-manager` is.

vladimyr: > This is friendlier for those who are not familiar with `home-manager.

That is exactly what I wanted to make sure, thanks!

vladimyr (approved):

chipbuster:
@kidonng Thanks for the fix!

PR #2513, opened 2021-03-26
feat(vcsh): Add new module for VCSH

alerque:
[VCSH](https://github.com/RichiH/vcsh) is a widely used dotfile management system. One of the most common usage patterns is to switch into a nested shell to work on a specific config repository, make changes, then drop back out. These nested shells are spawned with the appropriate GIT variables set such that all `git` actions act on the VCSH repository, not whatever directry you might happen to be in. It's very helful to have a prominant alert that you are acting on one of these config repos so you don't forget and try to do work on the current directory.

While many people will have one such repository (and hence a binary on/off flag would work), it's setup by default to handle many repositories. In my case I have a dozen or so. Hence the most obvious choice for noting the condition is to include the name of the currently active repository if any.

This *could* be rigged up with the `env_var` module, but that requires understanding the internals and tinkering with it quite a bit. This is a pretty straightforward low-overhead module that can be enabled by default and "just do the right thing" for the vast majority of cases.

### DIFF
--- a/src/modules/vcsh.rs
+++ b/src/modules/vcsh.rs
@@ -40,5 +40,35 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             return None;
         }
     });
+
     Some(module)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::test::ModuleRenderer;
+    use ansi_term::Color;
+
+    #[test]
+    fn not_in_env() {
+        let actual = ModuleRenderer::new("vcsh").collect();
+
+        let expected = None;
+
+        assert_eq!(expected, actual);
+    }
+
+    #[test]
+    fn env_set() {
+        let actual = ModuleRenderer::new("vcsh")
+            .env("VCSH_REPO_NAME", "astronauts")
+            .collect();
+
+        let expected = Some(format!(
+            "vcsh {} ",
+            Color::Yellow.bold().paint("astronauts")
+        ));
+
+        assert_eq!(expected, actual);
+    }
 }
--- a/src/modules/vcsh.rs
+++ b/src/modules/vcsh.rs
@@ -1,0 +1,44 @@
+use super::{Context, Module};
+
+use crate::config::RootModuleConfig;
+use crate::configs::vcsh::VcshConfig;
+use crate::formatter::StringFormatter;
+
+/// Creates a module that displays VCSH repository currently in use
+///
+/// Will display the name of the current VCSH repository if one is active.
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let repo = context.get_env("VCSH_REPO_NAME").unwrap_or_default();
+    if repo.trim().is_empty() {
+        return None;
+    }
+
+    let mut module = context.new_module("vcsh");
+    let config: VcshConfig = VcshConfig::try_load(module.config);
+
+    let parsed = StringFormatter::new(config.format).and_then(|formatter| {
+        formatter
+            .map_meta(|var, _| match var {
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "repo" => Some(Ok(&repo)),
+                _ => None,
+            })
+            .parse(None)
+    });
+
+    module.set_segments(match parsed {
+        Ok(segments) => segments,
+        Err(error) => {
+            log::warn!("Error in module `vcsh`:\n{}", error);
+            return None;
+        }
+    });
+    Some(module)
+}
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -189,6 +189,7 @@ $hostname\
 $shlvl\
 $kubernetes\
 $directory\
+$vcsh\
 $git_branch\
 $git_commit\
 $git_state\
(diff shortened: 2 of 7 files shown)


vladimyr:
First of all, I understand and appreciate your motivation. Codewise this is getting clear approval and :+1: However, I'm not really sure that adding a whole new module for the sake of displaying the environment variable is the right thing to do. There is already `env_var` module, as you already mentioned:


> This _could_ be rigged up with the `env_var` module, but that requires understanding the internals and tinkering with it quite a bit. This is a pretty straightforward low-overhead module that can be enabled by default and "just do the right thing" for the vast majority of cases.

and to be honest I'm not sure what do you mean by _understanding the internals and tinkering with it quite a bit_? Internals of what, VCSH, Starship, both? :thinking: 

I agree that this is pretty straightforward but it also sets a kinda controversial precedence I'd rather try to avoid.

/cc @starship/astronauts

andytom:
I think we have already set that precedent. We have a few modules that started out as something that could just be done using the env_var module (`SHLVL` and the `AWS` module spring to mind). I like the idea of having ready to use modules to make it easier for the users to just install and go without having to configure it and it avoids the current limitation of only having one `env_var` module at a time. However something we do need to think about for the future rewrite is how to reduce the boilerplate in these show an env var style modules.

vladimyr:
> I think we have already set that precedent. We have a few modules that started out as something that could just be done using the env_var module (`SHLVL` and the `AWS` module spring to mind).

Technically we did although I'd argue that `SHLVL` is something that really deserves special treatment due to its ubiquitous nature and `AWS` might started off as an `env_var` clone but evolved into a fairly complex beast.

> I like the idea of having ready to use modules to make it easier for the users to just install and go without having to configure it and it avoids the current limitation of only having one `env_var` module at a time.

I guess we agree on the ready-to-use part but I'd rather offer it as an optional ready-to-use config snippet enabled by _multi `env_var`_ feature than having it baked in. OFC that means we need to figure out #2426 first but I'd rather wait for it without introducing stuff like this into the codebase because once it gets in chances are we won't be able to easily remove/migrate it later on.

> However something we do need to think about for the future rewrite is how to reduce the boilerplate in these show an env var style modules.

My perspective is that they shouldn't be modules at all but easily reused config snippets instead.

alerque:
While this initial pass ta the module is pretty simple, there are more advanced things it could do that would not be possible via the current `env_var` module design (even if the limitation on only being able to use it once was removed). My hope would be after this basic starting point is in place that other VCSH users might have input on how to tweak the feature set. There are several possibilities:

* VCSH repo names don't always match the directories where they are managed, if they differ it might make sense to show the mapping.
* The status of the VCSH ignore file is important, and adding a warning flag if the file needs rewriting might be a useful reminder.
* Entering a VCSH repo and looking at the `git_status` module would give some clues asa to the state, but showing a dirty flag when *any* VCSH repo needs updating could also be quite useful, but a more costly operation to run that would require opt-in.

There may be others, the point is none of these are going to play nicely in the confines of an `env_var` snippet.

You've noted the `AWS` module *started* as a simple ENV var, this isn't ever going to be that involved but I expect it may outgrow this humble starting point.

vladimyr:
> > I think we have already set that precedent. We have a few modules that started out as something that could just be done using the env_var module (`SHLVL` and the `AWS` module spring to mind).
> 
> Technically we did although I'd argue that `SHLVL` is something that really deserves special treatment due to its ubiquitous nature and `AWS` might started off as an `env_var` clone but evolved into a fairly complex beast.

I forgot to mention one thing. `SHLVL` and `AWS_` vars are pretty much commonly respected by all kinds of different tools (_competing_ AWS CLIs and stuff) while this isn't a general dotfile management thing but rather VCSH-specific. And as much as I understand that VCSH is really a popular project it is far from becoming the de-facto standard in a crowded field populated by all kinds of different implementations like [Stow](https://www.gnu.org/software/stow/), [yadm](https://yadm.io/) and [rcm](https://github.com/thoughtbot/rcm) to name just a few...

That being said if others are in favor of merging it, I'm fine with it.

vladimyr:
> While this initial pass ta the module is pretty simple, there are more advanced things it could do that would not be possible via the current `env_var` module design (even if the limitation on only being able to use it once was removed).

> There may be others, the point is none of these are going to play nicely in the confines of an `env_var` snippet.
> 
> You've noted the `AWS` module _started_ as a simple ENV var, this isn't ever going to be that involved but I expect it may outgrow this humble starting point.

And that is a completely different discussion :smile: I'm sorry if I misunderstood your original comment but I didn't get the impression it is supposed to grow in functionality thus making it a bad fit for `env_var` in the long-term.

Regarding AWS, seems like we are repeating the history here: https://github.com/starship/starship/pull/419 (see the comments)

Dunno, I have mixed feelings about it. On the one hand, you could/should be able to fulfill your current needs by building upon `env_var` so I'm not really in favor of merging it as is. If there is a current need that can't be served by `env_var` (not talking about the multi-var issue) then it should be presented as part of this PR and I wouldn't object to merging it. On the other hand, I don't want to be that guy who will prevent growing it as a part of Starship itself just because at the moment it is not complex enough to become part of it in the first place. Letting smarter folks decide...

alerque:
To my knowledge none of `stow`, `yadm`, or `rcm` use anything equivalent to an ENV var to track state at all--because none of them spawn a shell for the user to work inside of. They are all CLI tools that take a one-shot approach to having dotfiles in a singe repo managed by the tool. There isn't as much call for those to influence the prompt since the currently running shell will never be in a mode specific to those tools.

Also I fail to see the logic in excluding a module because is for something more niche rather than a de facto standard. Why include Mercurial module if Git is the de facto standard? Why have a dozen different language modules (Rust, Python, Elm, Scala, OCaml, Zig, etc.)?

The simple reason is that because Starship's architecture is a single precompiled binary end users can't just add on their own modules the way they could with say Powerline9k/10k or many of the other scripting language based prompts out there. The `env_var` module is simply not flexible enough to cover the range of things people are going to want modules for, and shelling out from a `custom` module defeats the purpose of having a prompt generated by a single fast binary. First class support means getting a Rust module upstreamed.

vladimyr:
> Also I fail to see the logic in excluding a module because is for something more niche rather than a de facto standard. Why include Mercurial module if Git is the de facto standard? Why have a dozen different language modules (Rust, Python, Elm, Scala, OCaml, Zig, etc.)? 

I guess I poorly worded my reasoning here, I give you that. To follow your comparison you could also list other VCSes that aren't included and I wouldn't have any complaints regarding including, for instance, the Fossil module; while in the case of Bazaar I'd probably be more on the opposing side; CVS or TFS - I'd be reserved about adding those. So it is not about whether it is a de facto standard or not it is a classical question of where to draw the line. To draw a fair one, one needs to be educated on the subject, and honestly, my knowledge of dotfile management tools is rather scarce.

To give you more background on my comment, I knew about stow, rcm, and yadm already and just learned about VCSH's very existence. At a first glance, it seemed like it leans more toward the _Bazaar bucket_ rather than the _git-hg-fossil league_. Now you can say sure that is just my subjective and uneducated guess and you'd be totally right. And because of that, and because I hold my strong opinions weekly, I did a second pass and found this nice list https://dotfiles.github.io/utilities/. This places it somewhere in the middle between mainstream and exotic so I obviously need to rethink my stance here.

Without further ado, you successfully convinced me to change my opinion. Thank you for taking the time to patiently present your case and participating in the constructive discussion. I'm eager to see VCSH listed on our config page 🚀

src/modules/vcsh.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::configs::vcsh::VcshConfig;
+use crate::formatter::StringFormatter;
+
+/// Creates a module that displays VCSH repository currently in use
+///
+/// Will display the name of the current VCSH repository if one is active.
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let repo = context
+        .get_env("VCSH_REPO_NAME")
+        .unwrap_or_else(|| "".into());
```
vladimyr, line 13: Just one small nitpick.
```suggestion
        .unwrap_or_default();
```

vladimyr (approved):

vladimyr (approved):
LGTM 👍

andytom:
Thank you for your contribution @alerque and thank you for reviewing @vladimyr

PR #2521, opened 2021-03-27
feat(config): allow printing default and computed config

### LINKED ISSUE #2309: Make it easier to create initial starship.toml without jumping back&forth between config/docs
## Feature Request
Add a commented out config file `starship.toml`, prepopulated with all the main options  on the first installation to make it potentially easier for the users to configure

Currently, to create a custom `starship.toml` config file users need to go back and forth between their config file and the [official docs](https://starship.rs/config/#prompt) and copy&paste&edit the settings they need
This is rather inconvenient

Add a commented out config file on the first install that allows the user to uncomment/edit only the config lines he needs while having most of the explanation in the same file (and only needing to read up on the general explanation of how the configuration is done in the docs)

For example, the default `starship.toml` would have this section for the rust module
```toml
[rust]
# Option  	  Default                           	Description
# format  	= "via [$symbol($version )]($style)"	The format for the module.
# symbol  	= "🦀 "                              	A format string representing the symbol of Rust
# style   	= "bold red"                        	The style for the module.
# disabled	= false                             	Disables the rust module.
```
(since the variables are mostly self-evident, their description can be left only in the docs. Or maybe even the Description can be skipped as it's too repetitive)

If a user then needs to remove the `via` from his __rust__ string he'd just <kbd>ctrl</kbd><kbd>f</kbd> `rust`, uncomment the format string and delete the `via ` text without leaving his text editor

The are at least a few obvious downsides:

- The wall of comments can be frightening
- Useless modules would have to be manually cleaned up
- Documentation of the options is duplicated

A better solution would be to include these chunks of texts only for those modules a user needs as part of some onboarding process (and ideally get them programmatically from the docs to have "one source of truth"), but then it'd be also a much more complicated endeavor

davidkna:
#### Description

This PR adds `starship print-config` to print the computed config and `starship print-config --default` to print the default config.

I made a new struct in `src/configs/mod.rs` that contains all the module configs and implements `serve::Serialize` and `ModuleConfig`.
Options that have a `None` value need to be skipped because toml doesn't support `None` values and a warning comment about this is added at the top of the generated config files.
Before formatting as toml the struct needs to be converted into a `toml::Value` because otherwise the toml crate won't be able to properly handle printing the top-level root options.
I also got rid of the wrapper struct in `custom.rs` because the other modules don't use this for `extensiosn`, `folders` and `extensions` either.

#### Motivation and Context

Closes #2309
Related to #1921

#### Checklist:

- [x] I have updated the documentation accordingly. (not in-depth)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1628,7 +1628,7 @@ dependencies = [
 
 [[package]]
 name = "starship_module_config_derive"
-version = "0.1.2"
+version = "0.1.3"
 dependencies = [
  "proc-macro2",
  "quote 1.0.9",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -694,6 +694,7 @@ checksum = "824845a0bf897a9042383849b02c1bc219c2383772efcd5c6f9766fa4b81aef3"
 dependencies = [
  "autocfg",
  "hashbrown",
+ "serde",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ term_size = "0.3.2"
 quick-xml = "0.22.0"
 rand = "0.8.3"
 serde = { version = "1.0.125", features = ["derive"] }
-indexmap = "1.6.2"
+indexmap = { version ="1.6.2", features = ["serde"] }
 notify-rust = { version = "4.3.0", optional = true }
 semver = "0.11.0"
 which = "4.0.2"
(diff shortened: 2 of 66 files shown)


davidkna:
I ended up having to modify the `ModuleConfig` macro to use `default()` to make printing custom modules possible, but I think this aligns better with the general expectations for its behaviour.

andytom (approved):
LGTM

PR #2527, opened 2021-03-28
feat(config): warn about unknown config key names

### LINKED ISSUE #1213: Better feedback for invalid/broken configuration

There have been a few issues recently (e.g #1204 and #1179) that have been caused by errors in the configuration (missing quotes/typos/etc) when looking at the debug logs it is not clear why the configuration is incorrect. 

This is going to need to wait until #1075 has been completed but it would be nice to display some friendly error messages if there is anything that is wrong when parsing the config. Initially this could just be invalid toml but long term it would be nice to find keys that are incorrectly named e.g `symblo` not `symbol` or of the wrong type e.g using the string `'false'` rather than `false` for a bool but that will require further work to the config system. 

This is just some initial ideas but if there are any alternatives or quick wins we can implement now to help with configuration problems, please add a comment below.

davidkna:
#### Description

This PR prints a warning when an unknown config key is encountered:
`[WARN] - (starship::configs::starship_root): Unknown config key 'docker' `

This PR builds on the changes to `starship_module_config_derive` in #2521 to use `default` (https://github.com/starship/starship/pull/2521#issuecomment-808752340).

`load_config(&self, config: &'a Value) -> Self` is replaced by `load_config(&mut self, config: &'a Value)` and it loops over the config table now instead of checking every key individually.

`RootModuleConfig` implements `ModuleConfig` manually now to avoid issues with other modules not being known by the module.

#### Motivation and Context

Fixes #1213

#### Checklist:

### DIFF
--- a/src/configs/starship_root.rs
+++ b/src/configs/starship_root.rs
@@ -1,9 +1,8 @@
-use crate::config::ModuleConfig;
+use crate::{config::ModuleConfig, module::ALL_MODULES};
 
 use serde::Serialize;
-use starship_module_config_derive::ModuleConfig;
 
-#[derive(Clone, ModuleConfig, Serialize)]
+#[derive(Clone, Serialize)]
 pub struct StarshipRootConfig<'a> {
     pub format: &'a str,
     pub scan_timeout: u64,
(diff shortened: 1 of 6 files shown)


src/configs/starship_root.rs
```diff
@@ -87,3 +85,27 @@ impl<'a> Default for StarshipRootConfig<'a> {
...
+    fn load_config(&mut self, config: &'a toml::Value) {
+        if let toml::Value::Table(config) = config {
+            config.iter().for_each(|(k, v)| match k.as_str() {
+                "format" => self.format.load_config(v),
+                "scan_timeout" => self.scan_timeout.load_config(v),
+                "command_timeout" => self.command_timeout.load_config(v),
+                "add_newline" => self.add_newline.load_config(v),
+                unknown => {
+                    if !ALL_MODULES.contains(&unknown) && unknown != "custom" {
+                        log::warn!("Unknown config key '{}'", unknown);
```
vladimyr, line 99: Would this log both config key and module name or just the key?

davidkna: This only logs the config key. But the logger adds `(starship::configs::module_name)` to the message.

vladimyr: Thought so, wanted to confirm though. 👍

vladimyr (approved):
LGTM

PR #2535, opened 2021-03-31
build: Update snapcraft.yaml

popey:
build-error is not a valid syntax in snapcraft.yaml
The build-error lines are causing _all_ architecture builds to fail.

Screenshot_20210331_115239

```
Running pull phase...
Issues while validating snapcraft.yaml: The 'architectures[4]' property does not match the required schema: additional properties are not allowed ('build-error' was unexpected) or orderedDict([('build-on', 'ppc64el'), ('build-error', 'ignore')]) is not of type 'string'
Build failed
```

Sounds like an interesting option, but not one that exists right now, as far as I can tell! :D

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -81,6 +81,4 @@ architectures:
   - build-on: armhf
   - build-on: i386
   - build-on: ppc64el
-    build-error: ignore
   - build-on: s390x
-    build-error: ignore


vladimyr (approved):
LGTM

vladimyr:
@popey If you don't mind would you please reword your commit so it passes semantic pull request check. Basically prefixing the current title with `build: ` should do the trick.

andytom:
Thanks for your contribution @popey and thanks for reviewing @vladimyr

LinusU:
Interesting, as I remember I added these and they worked together with the local `snapcraft` command. 🤔 

Sorry for causing the build to fail 🙈

PR #2539, opened 2021-04-01
docs: quote `$PWD` in window title example

twm:
#### Description

Quote expansion of `$PWD` to prevent word-splitting, which would give an incorrect result if there is a single space and an error message if there is more than one space. Without the quoting it functions like this:

    ❯ basename /tmp/foo
    foo

    ❯ basename /tmp/foo bar
    foo

    ❯ basename /tmp/foo bar baz
    basename: extra operand ‘baz’
    Try 'basename --help' for more information.

If there are spaces in the directory path the result was often nonsensical.

#### Checklist:

- [ ] I have updated the tests accordingly (N/A).

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -80,7 +80,7 @@ add the following snippet to your `~/.bashrc` or `~/.zshrc`:
 
 ```bash
 function set_win_title(){
-    echo -ne "\033]0; $(basename $PWD) \007"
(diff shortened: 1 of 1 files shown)


vladimyr (approved):
Oh, right. Good catch @twm! 👍

andytom:
Thanks for fixing this @twm and thank for reviewing @vladimyr

PR #2548, opened 2021-04-03
refactor(init): cleanup init stub

vladimyr:
#### Description

Reformat init stub generation code and alter `bash` version.

#### Motivation and Context

Improve readability of the `init_stub` procedure.

#### Checklist:

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -147,23 +147,7 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
             starship.sprint_posix()?
         ),
         "powershell" => print!(
-            /*
-             * Explanation of syntax:
-             *
-             *      iex: alias of Invoke-Expression
-             *
-             *      &: Explicitly tells PowerShell to execute path with Starship
-             *         executable.
-             *
-             *      @: multi-line stdout is returned as an array, but a single
-             *         line or no lines are returned as-is. @ ensures it's always
-             *         an array.
-             *
-             *      -join "`n": Joins the stdout array together as a string with
-             *                  newlines. PowerShell escapes with ` instead of \
-             *                  thus `n translates to a newline.
-             */
-            r#"iex (@(&"{}" init powershell --print-full-init) -join "`n")"#,
+            r#"Invoke-Expression (& "{}" init powershell --print-full-init | Out-String)"#,
             starship.sprint()?
         ),
         "ion" => print!("eval $({} init ion --print-full-init)", starship.sprint()?),
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -145,15 +145,21 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
         "powershell" => print!(
             /*
              * Explanation of syntax:
-             * &: Explicitly tells powershell to execute path with starship executable.
              *
-             * @: multi-line stdout is returned as an array, but a single line or no lines
-             *    are returned as-is. @ ensures it's always an array.
+             *      iex: alias of Invoke-Expression
              *
-             * -join "`n": Joins the stdout array together as a string with newlines.
-             *             Powershell escapes with ` instead of \ thus `n translates to a newline.
+             *      &: Explicitly tells PowerShell to execute path with Starship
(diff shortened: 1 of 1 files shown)


src/init/mod.rs
```diff
@@ -121,68 +121,50 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
...
+            r#"
+            major="${{BASH_VERSINFO[0]}}"
+            minor="${{BASH_VERSINFO[1]}}"
+
+            if ((major > 4)) || {{ ((major == 4)) && ((minor >= 1)); }}; then
+                source <("{0}" init bash --print-full-init)
+            else
+                source /dev/stdin <<<"$("{0}" init bash --print-full-init)"
+            fi
+            "#,
```
vladimyr, line 133: I've tested the following snippet using versions `bash-3.2`, `bash-4.4`, and `bash-5.1`. It passes shellcheck, uses arithmetic comparisons, and avoids using subshell for the right `OR` branch (check for version `>=4.1`).
Note that Bash isn't indentation sensitive so indenting this doesn't really change anything except it doesn't stick out anymore.

src/init/mod.rs
```diff
@@ -121,68 +121,50 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
...
+            r#"Invoke-Expression (@(&"{}" init powershell --print-full-init) -join "`n")"#,
+            starship.sprint()?
+        ),
+        "ion" => print!("eval $({} init ion --print-full-init)", starship.sprint()?),
+        "elvish" => print!(
+            r#"eval ("{}" init elvish --print-full-init | slurp)"#,
+            starship.sprint_posix()?
+        ),
+        "tcsh" => print!(
+            r#"eval `("{}" init tcsh --print-full-init)`"#,
```
vladimyr, line 165: Previously it looked like this:
```rust
r#"eval "`("{}" init tcsh --print-full-init)`""#,
```
I removed quotes around (back)ticks because they are redundant.

src/init/mod.rs
vladimyr: `IIFE`-like `__main` is used to allow sourcing this stub without polluting scope.

src/init/mod.rs
```diff
@@ -121,68 +121,60 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
...
+             *
+             *      @: multi-line stdout is returned as an array, but a single
+             *         line or no lines are returned as-is. @ ensures it's always
+             *         an array.
+             *
+             *      -join "`n": Joins the stdout array together as a string with
+             *                  newlines. PowerShell escapes with ` instead of \
+             *                  thus `n translates to a newline.
+             */
+            r#"iex (@(&"{}" init powershell --print-full-init) -join "`n")"#,
```
davidkna, line 166: The PowerShell linter recommends against using shorthands in PowerShell scripts.

But for simplification using `Out-String` over ``-join "`n"`` might make sense.

vladimyr: I've just swapped `Invoke-Expression` with `iex` here because I got used to seeing it around inside various installation instructions. Everything else is inherited.
So you are suggesting going back to `Invoke-Expression`, and replacing ``-join "`n"`` with `Out-String`? Whatever you like, just please write it down cause I'm doing blind coding here. I've never used posh longer than few minutes and don't have an installation at hand for testing :upside_down_face:

davidkna: This will work:

```suggestion
            r#"Invoke-Expression (&"{}" init powershell --print-full-init | | Out-String)"#,
```

vladimyr: Now it makes sense to completely lose the above comment? Names are pretty descriptive and I would guess that `&` operator is a well-known thing in the PowerShell world?

davidkna: I think it's fine to remove the comment.

vladimyr: Done, thanks for your help! :+1:

chipbuster (approved):
This looks good to me. I'm not particularly qualified to review the powershell changes, but it looks like @davidkna has that one handled.

PR #2551, opened 2021-04-03
feat(python): Remove parentheses from pyvenv.cfg prompt parameter

### LINKED ISSUE #1844: Python virtualenv part of prompt now has double brackets
#### Current Behavior
I just updated this morning and now the virtualenv part of the prompt is displayed with 2 sets of parentheses

`via 🐍 v3.7.2 ((src)) `

Previously the virtualenv name only had one set of brackets

I am using pipenv and "src" is the short name of the virtualenv (`/Users/wardy/.local/share/virtualenvs/src-PnVlxuxe`)

I'm unaware of where it picks up the "virtualenv" variable but perhaps the update has added parens around this value now??

#### Expected Behavior

`via 🐍 v3.7.2 (src) `

#### Possible Solution

The toml file can be updated to revert this to the old behaviour. Perhaps this needs to change in the default install?

```toml
[python]
format = 'via [${symbol}${pyenv_prefix}${version}( $virtualenv)]($style) '
```

Note the removal of `\(` and `\)` around `$virtualenv`

#### Environment
- Starship version: 0.46.2
- Operating system: Mac OS 10.15.7

```bash
# Aliases from cygwin

alias cp='cp -i'
alias mv='mv -i'
alias df='df -h'
alias du='du -h'
#alias less='less -r'                          # raw control characters
alias more=less
alias whence='type -a'                        # where, of a sort
alias grep='grep --color'                     # show differences in colour
alias egrep='egrep --color=auto'              # show differences in colour
alias fgrep='fgrep --color=auto'              # show differences in colour
alias ls='ls -hBFG'                           # classify files in colour
alias dir='ls --color=auto --format=vertical'
alias vdir='ls --color=auto --format=long'
alias ll='ls -l'                              # long list
alias la='ls -A'                              # all but . and ..
alias l='ls -CF'                              #
alias where=which


# pyenv - https://github.com/pyenv/pyenv#installation
pyenv init - | source

# Starship fancy prompt 
starship init fish | source
```

#### Starship Configuration

```toml
# #[time]
# #disabled = false

[battery]
charging_symbol = "⚡️"
discharging_symbol = "💀"
full_symbol = "🔋"

[[battery.display]]
style = "bold red"
threshold = 75

[directory]
truncate_to_repo = false

#[python]
#format = 'via [${symbol}${pyenv_prefix}${version}( $virtualenv)]($style) '
# [memory_usage]
# disabled = false
# separator = "/"
# show_percentage = true
# show_swap = true
# style = "bold dimmed green"
# #symbol = " "
# threshold = -1

# [time]
# disabled = false
# format = "🕙[ %T ]"
# time_range = "10:00:00-23:00:00"
# utc_time_offset = "-5"

```

cswartzvi:
#### Description

Updated the retrieval of the the `prompt` parameter from `pyvenv.cfg` so that it removes extra parentheses. pipenv adds the `prompt` parameter with explicit parentheses when using a local `.venv` or setting the `PIPENV_VENV_IN_PROJECT` flag. This creates the 'double parentheses' in the issue below. Removing the additional parentheses when they are read from `pyvenv.cfg` does not affect the display of other python virtual environments. 

#### Motivation and Context

Closes #1844 

Example `pyvenv.cfg` with `prompt` parameter as created by pipenv:
image

Example `pyvenv.cfg` as created by the standard venv and `prompt` CLI option (`python3 -m venv venv --prompt venv-demo`):
image

Example of double parentheses (before fix):
image

After the fix:
image

#### Checklist:

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -373,7 +373,7 @@ prompt = 'foo'
         venv_cfg.write_all(
             br#"
 home = something
-prompt = '(foo) '
+prompt = '(foo)'
         "#,
         )?;
         venv_cfg.sync_all()?;
@@ -390,7 +390,7 @@ prompt = '(foo) '
 
         assert_eq!(actual, expected);
         dir.close()
-    }    
+    }
 
     fn check_python2_renders(dir: &tempfile::TempDir, starship_config: Option<toml::Value>) {
         let config = starship_config.unwrap_or(toml::toml! {
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -109,7 +109,7 @@ fn get_prompt_from_venv(venv_path: &Path) -> Option<String> {
         .ok()?
         .general_section()
         .get("prompt")
-        .map(String::from)
+        .map(|prompt| String::from(prompt.trim_matches(&['(', ')'] as &[_])))
 }
 
 #[cfg(test)]
@@ -364,6 +364,33 @@ prompt = 'foo'
         assert_eq!(actual, expected);
         dir.close()
     }
+
+    #[test]
+    fn with_active_venv_and_dirty_prompt() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        create_dir_all(dir.path().join("my_venv"))?;
+        let mut venv_cfg = File::create(dir.path().join("my_venv").join("pyvenv.cfg"))?;
+        venv_cfg.write_all(
+            br#"
+home = something
+prompt = '(foo) '
+        "#,
+        )?;
+        venv_cfg.sync_all()?;
+
+        let actual = ModuleRenderer::new("python")
+            .path(dir.path())
+            .env("VIRTUAL_ENV", dir.path().join("my_venv").to_str().unwrap())
+            .collect();
+
+        let expected = Some(format!(
+            "via {}",
+            Color::Yellow.bold().paint("🐍 v3.8.0 (foo) ")
+        ));
+
+        assert_eq!(actual, expected);
+        dir.close()
+    }    
 
     fn check_python2_renders(dir: &tempfile::TempDir, starship_config: Option<toml::Value>) {
         let config = starship_config.unwrap_or(toml::toml! {


src/modules/python.rs
```diff
@@ -365,6 +365,33 @@ prompt = 'foo'
...
+            .collect();
+
+        let expected = Some(format!(
+            "via {}",
+            Color::Yellow.bold().paint("🐍 v3.8.0 (foo) ")
+        ));
+
+        assert_eq!(actual, expected);
+        dir.close()
+    }
```
vladimyr, line 393: `rustfmt` check fails due to trailing whitespace:
```suggestion
    }
```

src/modules/python.rs
```diff
@@ -365,6 +365,33 @@ prompt = 'foo'
...
 
+    #[test]
+    fn with_active_venv_and_dirty_prompt() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        create_dir_all(dir.path().join("my_venv"))?;
+        let mut venv_cfg = File::create(dir.path().join("my_venv").join("pyvenv.cfg"))?;
+        venv_cfg.write_all(
+            br#"
+home = something
+prompt = '(foo) '
```
vladimyr, line 376: Is whitespace after right paren intentional?

vladimyr (changes requested):
Looks good overall! :+1:

src/modules/python.rs
vladimyr: I guess not :upside_down_face: 
```suggestion
prompt = '(foo)'
```

cswartzvi: > I guess not 🙃

Oops! Good catch, that was unintentional. pipenv adds a trailing space to the prompt and I wanted make sure it was handled properly. Forgot to remove it 😕

vladimyr (approved):
LGTM 👍

andytom:
Thank you for your contribution @cswartzvi  and thank you for reviewing @vladimyr .

PR #2483, opened 2021-03-21
docs(faq): make uninstall instructions shell-independent

heyajulia:
#### Description

The uninstall instructions assume that the user is running Bash, which is a safe bet, but isn't always the case.

#### Motivation and Context

This change wraps the uninstall command in `bash -c`, making it shell-independent in most cases. It will close issue #2484.

Closes #2484

#### Screenshots (if appropriate):

N/A

#### How Has This Been Tested?

I ran `bash -c 'rm "$(which starship)"'` to uninstall Starship, and it worked, even though I'm using Fish as my shell (this is good).

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #2552, opened 2021-04-03
docs(faq): change uninstall snippet to use `sh`

vladimyr:
#### Description

Use POSIX `sh` instead of Bash to make uninstall instructions shell agnostic.

#### Motivation and Context

As suggested here: https://github.com/starship/starship/pull/2483#issuecomment-803616815

#### Checklist:


chipbuster (approved):
:+1:

As discussed :)

PR #2556, opened 2021-04-04
fix(configure): fix print-config issues

davidkna:
#### Description

This PR fixes issues for `print-config` related to the new `FullConfig` struct.

The `golang` module was incorrectly listed as `go`.

For `RootConfig` `#[serde(flatten)]` wasn't enough to make `print-config` (non `--default`) work. I inlined `RootConfig` instead and added a comment in `starship_root.rs` to ensure both are always changed together.

#### Motivation and Context

Closes #

#### Checklist:


vladimyr (approved):
LGTM

PR #2565, opened 2021-04-08
feat(deno): create module

Milo123459:
#### Description

Added a new deno module
#### Motivation and Context

Closes #

img

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -662,7 +662,7 @@ By default the module will be shown if any of the following conditions are met:
 | Option              | Default                                           | Description                                     |
 | ------------------- | ------------------------------------------------- | ----------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"`              | The format for the module.                      |
-| `symbol`            | `"🦕 "`                                           | A format string representing the symbol of Dart |
+| `symbol`            | `"🦕 "`                                           | A format string representing the symbol of Deno |
 | `detect_extensions` | `[]`                                        | Which extensions should trigger this module.    |
 | `detect_files`      | `["mod.ts", "mod.js", "deps.ts", "deps.js"]` | Which filenames should trigger this module.     |
 | `detect_folders`    | `[]`                                  | Which folders should trigger this module.       |
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -199,6 +199,7 @@ $docker_context\
 $package\
 $cmake\
 $dart\
+$deno\
 $dotnet\
 $elixir\
 $elm\
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -662,7 +662,7 @@ By default the module will be shown if any of the following conditions are met:
 | Option              | Default                                           | Description                                     |
 | ------------------- | ------------------------------------------------- | ----------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"`              | The format for the module.                      |
-| `symbol`            | `"🦕 "`                                           | A format string representing the symbol of Deno |
+| `symbol`            | `"🦕  "`                                           | A format string representing the symbol of Deno |
 | `detect_extensions` | `[]`                                        | Which extensions should trigger this module.    |
 | `detect_files`      | `["mod.ts", "mod.js", "deps.ts", "deps.js"]` | Which filenames should trigger this module.     |
 | `detect_folders`    | `[]`                                  | Which folders should trigger this module.       |
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -651,6 +651,41 @@ By default the module will be shown if any of the following conditions are met:
 format = "via [🔰 $version](bold red) "
 ```
 
+## Deno
+
+The `deno` module shows you your currently installed version of Deno.
+By default the module will be shown if any of the following conditions are met:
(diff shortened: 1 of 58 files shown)


Milo123459:
Currently, if Deno is not installed, it looks like this:
img

Is there something I need to do to make sure it doesn't look like this? Or can I just add a check and do a return None?

vladimyr (changes requested):
It looks good overall :+1: 
Please address the comments I've left you and add tests for the remaining two files that trigger module display.

wyze:
@Milo123459 You could move the version check up, right below the `is_deno_project`, variable. Then you can return None if the version comes back as None.

Milo123459:
> @Milo123459 You could move the version check up, right below the `is_deno_project`, variable. Then you can return None if the version comes back as None.

I've spoken to people in the discord and they said leave it as is. Thanks though.

src/modules/deno.rs
```diff
@@ -0,0 +1,108 @@
...
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "version" => context
+                    .exec_cmd("deno", &["-V"])
+                    .and_then(|output| parse_deno_version(output.stdout.trim()))
```
davidkna, line 32: This calls `.trim()`, but later in the test you don't call `.trim()` before passing to `parse_deno_version`, is it needed?

src/modules/deno.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+fn parse_deno_version(deno_version: &str) -> Option<String> {
+    let version = deno_version
+        // split into ["deno 1.8.3", ""]
+        .split('\n')
+        // get first line
+        .next()
+        // unwrap
+        .unwrap()
```
davidkna, line 57: I'd prefer this (no `unwrap`, `split('\n')` -> `lines()`):
```suggestion
        // split into ["deno 1.8.3", ""]
        .lines()
        // get first line
        .next()?
```

But it even works without these lines for me:
```suggestion
```

docs/config/README.md
```diff
@@ -650,6 +651,41 @@ By default the module will be shown if any of the following conditions are met:
...
+The `deno` module shows you your currently installed version of Deno.
+By default the module will be shown if any of the following conditions are met:
+- The current directory contains a `mod.ts`, `mod.js`, `deps.ts` or `deps.js` file
+
+### Options
+
+| Option              | Default                                           | Description                                     |
+| ------------------- | ------------------------------------------------- | ----------------------------------------------- |
+| `format`            | `"via [$symbol($version )]($style)"`              | The format for the module.                      |
+| `symbol`            | `"🦕 "`                                           | A format string representing the symbol of Dart |
```
davidkna, line 665: Emojis need to have two instead of one space after them to avoid rendering bugs:
```suggestion
| `symbol`            | `"🦕  "`                                           | A format string representing the symbol of Dart |
```

davidkna (changes requested):

src/modules/deno.rs
Milo123459: No, but I mean I'll add it.

src/modules/deno.rs
Milo123459: For me it didn't work without the long lines of code, but, if it works for you then I suppose I can try again.

Okay, it worked without it. Wasn't working yesterday so must of been a fluke.

vladimyr:
> ... and add tests for the remaining two files that trigger module display.

@Milo123459 Please apply the following patch: https://github.com/vladimyr/starship/commit/eb09433.patch

Milo123459:
> > ... and add tests for the remaining two files that trigger module display.
> 
> @Milo123459 Please apply the following patch: [vladimyr/starship@`eb09433`.patch](https://github.com/vladimyr/starship/commit/eb09433.patch)

Oh my bad, I forgot. Adding now.

Milo123459:
Added those tests @vladimyr

docs/config/README.md
```diff
@@ -650,6 +651,41 @@ By default the module will be shown if any of the following conditions are met:
...
+The `deno` module shows you your currently installed version of Deno.
+By default the module will be shown if any of the following conditions are met:
+- The current directory contains a `mod.ts`, `mod.js`, `deps.ts` or `deps.js` file
+
+### Options
+
+| Option              | Default                                           | Description                                     |
+| ------------------- | ------------------------------------------------- | ----------------------------------------------- |
+| `format`            | `"via [$symbol($version )]($style)"`              | The format for the module.                      |
+| `symbol`            | `"🦕  "`                                           | A format string representing the symbol of Deno |
```
vladimyr, line 665:
```suggestion
| `symbol`            | `"🦕 "`                                           | A format string representing the symbol of Deno |
```

vladimyr (changes requested):

docs/config/README.md
vladimyr: I know @davidkna requested this earlier https://github.com/starship/starship/pull/2565#discussion_r610603565 but all other emoji symbols have a single trailing space (except gcloud and git tag that have two). Plus this got changed but the code stayed the same.

vladimyr (approved):
LGTM :rocket:

davidkna:
Thanks for the contribution @Milo123459 and thanks for reviewing @vladimyr.

PR #2566, opened 2021-04-09
refactor: propagate version formatting errors

vladimyr:
#### Description

Propagate errors originating from `VersionFormatter#format_version` to the upper level.

#### Motivation and Context

Continuation of #2499
(It got closed before I managed to leave my review.)

#### Checklist:

---

/cc @offbyone

### DIFF
--- a/src/formatter/version.rs
+++ b/src/formatter/version.rs
@@ -17,10 +17,16 @@ impl<'a> VersionFormatter<'a> {
         Ok(Self { formatter })
     }
 
+    /// Format version string using provided format
+    pub fn format_version(
+        version: &'a str,
+        format: &'a str,
+    ) -> Result<String, StringFormatterError> {
+        Self::new(format).and_then(|formatter| formatter.format(version))
+    }
+
     /// Formats a version structure into a readable string
-    ///
-    /// No matter what comes in, this will return some usable string
-    pub fn format_version(self, version: &str) -> Result<String, StringFormatterError> {
+    pub fn format(self, version: &'a str) -> Result<String, StringFormatterError> {
         let parsed = Versioning::new(version);
         let formatted = self
             .formatter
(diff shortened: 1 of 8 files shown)


src/modules/java.rs
```diff
@@ -74,19 +71,24 @@ fn get_java_version(context: &Context, config: &JavaConfig) -> Option<String> {
...
-        VersionFormatter::new(version_format)
-            .ok()?
-            .format_version(version),
-    )
+    let formatted = VersionFormatter::new(version_format)
+        .and_then(|formatter| formatter.format_version(version));
+
+    match formatted {
+        Ok(formatted) => Some(formatted),
+        Err(error) => {
```
offbyone, line 87: What conditions would lead to this warning? I would rather just silently format the raw version than log a warning when this happens.

offbyone (approved):
I'm mostly curious about the error handling rationale; I personally would prefer that version formatting be fairly silent, if possible, but that might be a "just me" philosophy. Otherwise, it looks like you reordered parameters in a few places, but ...  ¯\_(ツ)_/¯ different strokes for different folks :)

src/modules/java.rs
vladimyr: > I would rather just silently format the raw version than log a warning when this happens.

This is _silent_ unless you explicitly opt-in by setting `STARSHIP_LOG` environment variable to the `warn` level.

> What conditions would lead to this warning?

The simplest case that would cause an error is a malformed version format like:
```toml
version = "v${raw" # missing right curly
```

This results in (with logging enabled):
```
[WARN] - (starship::modules::java): Error formating `java` version:
 --> 1:3
  |
1 | v${raw
  |   ^---
  |
  = expected variable_name
```

vladimyr:
> I'm mostly curious about the error handling rationale; I personally would prefer that version formatting be fairly silent, if possible, but that might be a "just me" philosophy.

As I explained above https://github.com/starship/starship/pull/2566#discussion_r610245095 this is still silent. Error handling is moved to the upper level because version formatter doesn't know who called it thus can't print contextually rich warnings.

> Otherwise, it looks like you reordered parameters in a few places, but ... ¯_(ツ)_/¯ different strokes for different folks :)

This speaks for itself ie I find the original version so good I didn't feel a need to change things around :wink:
I did reordering to achieve the following form: `format_version(version /*[1]*/, format /*[2]*/)` where version is followed by format so it reads like: _version formatter formats **version [1]** using provided **format [2]**_.

vladimyr:
> Otherwise, it looks like you reordered parameters in a few places

Actually, there is more than just that. I did 3 additional things there:
1. ensured that every `format_{module}_version` function has following signature `format_{module}_version(version, version_string)` (no context, config, just format string)
2. simplified version parsing in a few of those
3. renamed a bunch of java tests from `test_parsing_*` to `test_format_*` in order to follow already established convention

offbyone (approved):

davidkna (approved):

src/formatter/version.rs
```diff
@@ -17,81 +19,85 @@ impl<'a> VersionFormatter<'a> {
...
+
+        formatted.map(|segments| {
+            segments
                 .iter()
                 .map(|segment| segment.value.as_str())
                 .collect::<Vec<&str>>()
-                .join(""),
-            Err(_) => version.to_string(),
-        }
+                .join("")
```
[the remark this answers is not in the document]
vladimyr: Done, thanks for the tip! :+1:

PR #2586, opened 2021-04-15
feat(typescript): create module

Milo123459:
#### Description

Adds a ts module
#### Motivation and Context

Closes #

img

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -220,6 +220,7 @@ $rust\
 $scala\
 $swift\
 $terraform\
+$typescript\
 $vagrant\
 $zig\
 $nix_shell\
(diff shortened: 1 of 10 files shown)


matchai:
Thank you for taking the time to work on this feature, Milo!

I'm, unfortunately, a little uncertain whether this should be added as a module. Adding a JS project dependency opens the door for countless other runtime dependency modules (e.g. React, Next, etc.)

We had an abandoned PR long ago for a single module to achieve all them all with a single implementation, so it may be best if we end up doing that.

That said, creating a `js_dependencies` module now opens the door for dependency versions for _all other runtimes_. 🤔

Milo123459:
I think TypeScript doesn't really compare to things like React and Next personally. TypeScript is a superset of JS, and react is a framework or library or whatever. 
Maybe, a js_dependencies parent module? For example:
js_dependencies/
   next: ...
   react: ...
   vue: ...
But then, what if someone wants to display next's version, but not Reacts?

Thanks for the fast response Matan.

matchai:
Another important consideration as that the prompt, in JS projects, is already very long with git_branch, git_status, node, and package. Additionally adding TS would lengthen it, and other js dependencies would be much too long.

> I think TypeScript doesn't really compare to things like React and Next personally. TypeScript is a superset of JS, and react is a framework or library or whatever.

The same justification for having the TS version number would apply to React: there are very meaningful functional and API changes between versions that change how I solve the same problem.

> Maybe, a js_dependencies parent module?

That was my thinking as well. Similarly to how we support custom modules:

```toml
[js_dependency.react]
symbol = "⚛️ "

[js_dependency.vue]
symbol = "🏔 "
```

> But then, what if someone wants to display next's version, but not Reacts?

That's a good question. Since Next contains a fixed version of React, I assume that would be a solved problem – the user would always want to see the Next version. 😛

Milo123459:
The node version can take upwords of 300 seconds and I've tested this module and using the package.json deserialization it can take 200 ms at most, this shouldn't affect speed.
But I do suppose that this might not be the best contribution and that it isn't helpful or whatever, it does have flaws.

*or* we could just add modules for everything, like a react, next or vue module? Probably won't happen, also, if there is a js_dependencies one, then there will have to be a deno_dependencies one, and it should be node_dependencies. 

Overall:
Shouldn't affect speed that much
It works on 3 different types of platforms
node_dependencies and js_dependencies

And who's to not say that js_dependencies can have whatever dependencies you want? Nothing fixed in place?

matchai:
To clarify, by "that the prompt, in JS projects, is already very long", I meant the actual horizontal space occupied by the prompt, not the speed. 😄

Milo123459:
> To clarify, by "that the prompt, in JS projects, is already very long", I meant the actual horizontal space occupied by the prompt, not the speed. 😄

My bad!

Milo123459:
So uh, @matchai to sum up this isn't going to be merged?

matchai:
Probably not as-is. We'll need to discuss it with the team to figure out what the best way to proceed is.

Milo123459:
> Probably not as-is. We'll need to discuss it with the team to figure out what the best way to proceed is.

Cool. Sounds good.

kidonng:
I'd also like to point out `tsc` (i.e. TypeScript compiler) is usually something project-wide rather than system wide, to my knowledge. Therefore it has to be found inside `node_modules/.bin`.
But there comes another question that if one is using PnP, there won't be `node_modules`. You have to call `yarn run tsc` (and IIRC pnpm support PnP too, the rabbit hole is deep...)

vladimyr:
> But there comes another question that if one is using [PnP](https://yarnpkg.com/features/pnp/), there won't be `node_modules`. You have to call `yarn run tsc` (and IIRC pnpm support PnP too, the rabbit hole goes deep...)

Now I don't have any data to back my statements but it seems to me that yarn2 usage isn't that wide exactly because of PnP issues people are experiencing. Pnpm on another hand uses symlinks so you'll find (pointer to) tsc exactly where you would expect it.

matchai:
Closing this PR for the time being. We'll revisit the inclusion of TS once we sort out the story for JS dependencies.

PR #2587, opened 2021-04-15
feat(gcloud): Shorten google cloud iam service account addresses

### LINKED ISSUE #1737: Shorten Google Cloud IAM service account addresses
Google Cloud Platform service accounts have lengthy names in the `<name>@<project-id>.iam.gserviceaccount.com` format. This takes up a lot of space with information that is not useful.

Only the leading part of the service account e-mail is interesting - `<name>@<project-id>`. It would be great if the rest could be chopped off whenever ending in `.iam.gserviceaccount.com`. Additionally, since the project identifier is usually part of the prompt, we could go one further and remove the `@<project-id>` part too, only leaving the name.

In general, being able to run regular expressions on format variables might be a universal solution to these types of requests, so that every user may further tweak the values to their liking. 

Screenshot demonstrating the verboseness:
Image

cymruu:
Hello, thanks everyone for you hard work on this project. The experience of contributing was great! It was my first time doing anything in rust but I hope it will be somehow helpful. 

#### Description
Parses google account string into 2 variables so user can how he sees the information in prompt.

#### Motivation and Context
Closes #1737

I ran `cargo test` on MacOS and Linux.

#### Checklist:

### DIFF
--- a/src/configs/gcloud.rs
+++ b/src/configs/gcloud.rs
@@ -15,7 +15,7 @@ pub struct GcloudConfig<'a> {
 impl<'a> Default for GcloudConfig<'a> {
     fn default() -> Self {
         GcloudConfig {
-            format: "on [$symbol$account$account_at(\\($region\\))]($style) ",
+            format: "on [$symbol$account@$domain(\\($region\\))]($style) ",
             symbol: "☁️ ",
             style: "bold blue",
             disabled: false,
--- a/src/configs/gcloud.rs
+++ b/src/configs/gcloud.rs
@@ -15,7 +15,7 @@ pub struct GcloudConfig<'a> {
 impl<'a> Default for GcloudConfig<'a> {
     fn default() -> Self {
         GcloudConfig {
-            format: "on [$symbol$account(\\($region\\))]($style) ",
+            format: "on [$symbol$account$account_at(\\($region\\))]($style) ",
             symbol: "☁️ ",
             style: "bold blue",
             disabled: false,
--- a/src/modules/gcloud.rs
+++ b/src/modules/gcloud.rs
@@ -99,8 +99,9 @@ fn alias_region(region: String, aliases: &HashMap<String, &str>) -> String {
 fn map_account(account: String) -> (Option<String>, Option<String>) {
     match account.find('@') {
         Some(index) => {
-            let (account, at) = account.split_at(index);
-            (Some(account.to_owned()), Some(at.to_owned()))
+            let (account, mut domain) = account.split_at(index);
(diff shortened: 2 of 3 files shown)


src/modules/gcloud.rs
```diff
@@ -132,7 +147,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 _ => None,
             })
             .map(|variable| match variable {
-                "account" => gcloud_account.as_ref().map(Ok),
+                "account" => account.as_ref().map(Ok),
+                "account_at" => at.as_ref().map(Ok),
```
matchai, line 151: Rather than `account_at`, perhaps `domain` would be clearer?

src/configs/gcloud.rs
```diff
@@ -15,7 +15,7 @@ pub struct GcloudConfig<'a> {
 impl<'a> Default for GcloudConfig<'a> {
     fn default() -> Self {
         GcloudConfig {
-            format: "on [$symbol$account$account_at(\\($region\\))]($style) ",
+            format: "on [$symbol$account@$domain(\\($region\\))]($style) ",
```
matchai, line 18: Don't forget to update the format string in the docs! 😊

src/modules/gcloud.rs
```diff
@@ -96,6 +96,17 @@ fn alias_region(region: String, aliases: &HashMap<String, &str>) -> String {
...
+fn map_account(account: String) -> (Option<String>, Option<String>) {
+    match account.find('@') {
+        Some(index) => {
+            let (account, mut domain) = account.split_at(index);
+            domain = domain.strip_prefix('@').unwrap();
+            (Some(account.to_owned()), Some(domain.to_owned()))
+        }
+        None => (Some(account), None),
+    }
+}
```
vladimyr, line 108:
```suggestion
fn map_account(account: &str) -> (Option<String>, Option<String>) {
    let mut segments = account.splitn(2, '@');
    (segments.next().map(String::from), segments.next().map(String::from))
}
```

src/modules/gcloud.rs
```diff
@@ -121,6 +132,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         None
     };
 
+    let (account, domain) = match gcloud_account {
+        Some(account) => map_account(account),
+        None => (None, None),
+    };
```
vladimyr, line 138:
```suggestion
    let (account, domain) = if let Some(gcloud_account) = gcloud_account {
        map_account(&gcloud_account)
    } else {
        (None, None)
    };
```

vladimyr (changes requested):

PR #2589, opened 2021-04-16
feat(config): Add support for `inverted` token in style strings

### LINKED ISSUE #2390: Add keyword to invert foreground and background color in style strings
I want my prompt to be very easy to find even when buried among tons of output, and the best way to do that is to give it a bright and solid background:
screenshot-1614507630
To stay legible the text has to be as dark as possible, i.e. the color of the terminal background.

A keyword like `inverse` or `inverted` that would have the same effect as `\e[7m`, i.e. swap the foreground and background colors.

A working alternative is to manually set the foreground color to the HEX color code of the terminal background, e.g. `[$path](bold #161821 bg:blue)`. But this isn't very flexible and needs to be manually updated every time the terminal colorscheme is changed.

cheap-glitch:
#### Description

This PR adds support for the `inverted` keyword in style strings. This keyword swaps the foreground and background colors, similar to the `\e[7m` ANSI code.

~~(I chose `reverse` instead of `inverted` as I first proposed, because that's the terminology `ansi_term` uses).~~

#### Motivation and Context

Closes #2390 

image

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -98,9 +98,9 @@ Style strings are a list of words, separated by whitespace. The words are not ca
   - `<color>`
   - `none`
 
-where `<color>` is a color specifier (discussed below). `fg:<color>` and `<color>` currently do the same thing , though this may change in the future. The order of words in the string does not matter.
+where `<color>` is a color specifier (discussed below). `fg:<color>` and `<color>` currently do the same thing, though this may change in the future. `inverted` swaps the background and foreground colors. The order of words in the string does not matter.
 
-The `none` token overrides all other tokens in a string if it is not part of a `bg:` specifier, so that e.g. `fg:red none fg:blue` will still create a string with no styling. `bg:none`  sets the background to the default color so `fg:red bg:none` is equivalent to `red` or `fg:red` and `bg:green fg:red bg:none` is also equivalent to `fg:red` or `red`. It may become an error to use `none` in conjunction with other tokens in the future.
+The `none` token overrides all other tokens in a string if it is not part of a `bg:` specifier, so that e.g. `fg:red none fg:blue` will still create a string with no styling. `bg:none` sets the background to the default color so `fg:red bg:none` is equivalent to `red` or `fg:red` and `bg:green fg:red bg:none` is also equivalent to `fg:red` or `red`. It may become an error to use `none` in conjunction with other tokens in the future.
 
 A color specifier can be one of the following:
 
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -28,7 +28,7 @@ function blastoff(){
 starship_precmd_user_func="blastoff"
 ```
 
-- To run a custom function right before a command runs, you can use the 
+- To run a custom function right before a command runs, you can use the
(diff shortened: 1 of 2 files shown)


docs/advanced-config/README.md
```diff
@@ -92,6 +92,7 @@ Style strings are a list of words, separated by whitespace. The words are not ca
   - `bold`
   - `underline`
   - `dimmed`
+  - `reverse`
```
chipbuster, line 95: I think it'd be good to have a short paragraph below describing what `reverse` does. It's not immediately obvious if this is color inversion or some sort of right-to-left token, but also the docs currently don't mention "inverted," which I think is what most people would be searching for.

I think after the paragraph on line 103 would be a good location (I'd put this comment there, but the GitHub interface won't let me)

chipbuster (commented):
I didn't realize our library already supported this. Nice find.

I'd like to see a little more explanation for what the keyword does. Would be especially nice if the word "invert" or "inverted" showed up, since I think that's how most people would search for this feature.

src/config.rs
```diff
@@ -675,20 +677,22 @@ mod tests {
     }
 
     #[test]
-    fn table_get_styles_bold_italic_underline_green_dimmy_silly_caps() {
-        let config = Value::from("bOlD ItAlIc uNdErLiNe GrEeN diMMeD");
+    fn table_get_styles_bold_italic_underline_green_dimmed_reverse_silly_caps() {
```
vladimyr, line 680: Yeah, I know that this covers both regular and inverted styles but I'd rather separate those by c/p. Having dedicated tests seems easier on the eyes IMHO.

vladimyr (changes requested):

src/config.rs
cheap-glitch: Sorry, could you clarify what you mean by "c/p"? Do mean splitting the function to have one test case for each keyword?

vladimyr: My bad, I should've used clearer language. Anyway, I meant copy-pasting the existing test and extending a copy rather than the original, so it results in having both existing `table_get_styles_bold_italic_underline_green_dimmy_silly_caps` and a newly added `table_get_styles_bold_italic_underline_green_dimmy_reverse_silly_caps`.

docs/advanced-config/README.md
cheap-glitch: Actually, how about making an exception and using a different keyword than `ansi_term`? I agree that `reverse` is ambiguous. Using `inverted` is more explicit and self-documenting.

vladimyr (commented):
LGTM but as @chipbuster noted (https://github.com/starship/starship/pull/2589/files#r615217074) docs need to get updated.

docs/advanced-config/README.md
chipbuster: Yeah, I wouldn't mind using `inverted`.

cheap-glitch: Done. Should I still add a little blurb in the doc or do you think the keyword is sufficiently self-documenting?

chipbuster: I think a single sentence in the previous paragraph would be sufficient now (e.g. ""inverted" swaps the background and foreground colors")

vladimyr (approved):
I guess the right mark for this one is:

W┴פ˥

😀

PR #2590, opened 2021-04-16
feat(toolbox): Support Toolbox 

### LINKED ISSUE #3104: Add toolbox container indicator
## Feature Request

[Toolbox](https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/) has a feature where upon entering a container, a diamond is prefixed in the PS1 as a visual indicator that the user is working inside a container. `toolbox` will source [/etc/profile.d/toolbox.sh](https://github.com/containers/toolbox/blob/main/profile.d/toolbox.sh) automatically. This container indicator would be nice to have when using starship as well.

The script seems to check for the presence of `/run/.containerenv` and `/run/.toolboxenv`. Starship could check for the presence of the files to mimic the behavior, or somehow lean on `toolbox.sh` directly in case that functionality ever changes in the future.

alphacc:
#### Description
WIP to support toolbox from https://github.com/containers/toolbox

#### Motivation and Context
Have a better experience when using Fedora Silverblue

Fixes #3104

Heavily Inspired by the singularity module

I'd like to see if there is interest from other user, and I also would like to talk with toolbox upstream to see how we can better use env variable (TOOLBOX_ namespace ? etc...)

I just use the plugin daily for my own laptop.

#### Checklist:

### DIFF
--- a/src/configs/toolbox.rs
+++ b/src/configs/toolbox.rs
@@ -1,0 +1,23 @@
+use crate::config::ModuleConfig;
+
+use serde::Serialize;
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig, Serialize)]
+pub struct ToolboxConfig<'a> {
+    pub symbol: &'a str,
+    pub format: &'a str,
+    pub style: &'a str,
+    pub disabled: bool,
+}
+
+impl<'a> Default for ToolboxConfig<'a> {
+    fn default() -> Self {
+        ToolboxConfig {
+            format: "[$symbol\\[$env\\]]($style) ",
+            symbol: "",
+            style: "blue bold dimmed",
+            disabled: false,
+        }
+    }
+}
--- a/src/configs/toolbox.rs
+++ b/src/configs/toolbox.rs
@@ -16,7 +16,7 @@ impl<'a> Default for ToolboxConfig<'a> {
         ToolboxConfig {
             format: "[$symbol\\[$env\\]]($style) ",
             symbol: "",
-            style: "blue bold dimmed",
+            style: "bright-blue bold dimmed",
             disabled: false,
         }
     }
--- a/src/modules/toolbox.rs
+++ b/src/modules/toolbox.rs
@@ -1,0 +1,69 @@
+use super::{Context, Module, RootModuleConfig};
+
+use crate::configs::toolbox::ToolboxConfig;
+use crate::formatter::StringFormatter;
+
+/// Creates a module with the current toolbox image
+///
+/// Will display the toolbox image if `$toolbox_NAME` is set.
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let toolbox_env = context.get_env("TOOLBOX_PATH")?;
+    let toolbox_name = context.get_env("NAME")?;
+
+    let mut module = context.new_module("toolbox");
+    let config: ToolboxConfig = ToolboxConfig::try_load(module.config);
+
+    let parsed = StringFormatter::new(config.format).and_then(|formatter| {
+        formatter
+            .map_meta(|variable, _| match variable {
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "env" => Some(Ok(&toolbox_name)),
+                _ => None,
+            })
+            .parse(None)
+    });
+
+    module.set_segments(match parsed {
+        Ok(segments) => segments,
+        Err(error) => {
+            log::warn!("Error in module `toolbox`: \n{}", error);
+            return None;
+        }
+    });
+
+    Some(module)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::test::ModuleRenderer;
+    use ansi_term::Color;
+
+    #[test]
+    fn no_env_set() {
+        let actual = ModuleRenderer::new("toolbox").collect();
+
+        let expected = None;
+        assert_eq!(expected, actual);
+    }
+    #[test]
+    fn env_set() {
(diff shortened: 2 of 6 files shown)


matchai:
Hey Thomas 👋 

With less than 1000 GitHub stars, I'm afraid that Toolbox may be a little too obscure to qualify to be added to Starship.
Perhaps it's better suited as a [custom module](https://github.com/starship/starship/discussions/1252)?

jankatins:
I've this as custom module (the checked files seem to be also used by the bash scripts to identify "in toolbox"): 

```toml
[custom.toolbox]
description = "The activated toolbox container"
command = "echo ${DISTTAG%container}"
when = """ [[ -f /run/.containerenv && -f /run/.toolboxenv ]] """
style = "bold red"
format = "[\\[Toolbox: $output\\]]($style) "
disabled = false
```

```bash
[22:01:31] [1] ✗  cat /etc/profile.d/toolbox.sh
[...]
if [ -f /run/.containerenv ] \
   && [ -f /run/.toolboxenv ]; then
    PS1=$(printf "\[\033[35m\]⬢\[\033[0m\]%s" "[\u@\h \W]\\$ ")
[...]
```

I would love to have this out of the box. We use this as $work to run our development env in there and fedora and GNOME seems to put a lot of effort into it, so my guess is that this will become biggish. A lot of news items on gnome 40 or fedora 34 mention it:

https://www.phoronix.com/scan.php?page=news_item&px=Fedora-Workstation-34-Excites
https://help.gnome.org/misc/release-notes/40.0/developers.html
https://www.theregister.com/2021/03/17/fedora_34_preview/

vladimyr:
I second this. Toolbox and all the work on Fedora Silverblue have a high potential of producing the next big thing. It is definitely a niche but not obscure IMHO. :+1: from me.

alphacc:
Ok thanks for the feedback I'll work on improving it and then we can decide.

One additional point is that FC34 will be the base for CentOS Stream 9 / RHEL 9 so it's likely it could gain more popularity soon.

sclel016:
I'd definitely love to see this merge to Starship in some capacity.

src/configs/toolbox.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub format: &'a str,
+    pub style: &'a str,
+    pub disabled: bool,
+}
+
+impl<'a> Default for ToolboxConfig<'a> {
+    fn default() -> Self {
+        ToolboxConfig {
+            format: "[$symbol\\[$env\\]]($style) ",
+            symbol: "",
```
jankatins, line 18: the bash one uses a red dot

jkonecny12:
The toolbox repository is pretty close to 1000 :star:

HarryMichal:
> The toolbox repository is pretty close to 1000 star

Already over :)

src/modules/toolbox.rs
```diff
@@ -0,0 +1,69 @@
+use super::{Context, Module, RootModuleConfig};
+
+use crate::configs::toolbox::ToolboxConfig;
+use crate::formatter::StringFormatter;
+
+/// Creates a module with the current toolbox image
+///
+/// Will display the toolbox image if `$toolbox_NAME` is set.
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let toolbox_env = context.get_env("TOOLBOX_PATH")?;
```
HarryMichal, line 10: While `TOOLBOX_PATH` is set in every toolbx container, it does not hold information important for this particular module. It is a path to the internal Toolbx binary used for container initialization. Instead I'd look for the file `/run/.toolboxenv`. It's presence signifies that the container is in fact a toolbx container.

src/modules/toolbox.rs
```diff
@@ -0,0 +1,69 @@
...
+    let toolbox_name = context.get_env("NAME")?;
```
HarryMichal, line 11: Is set in our [Containerfile](https://github.com/containers/toolbox/blob/cfffb72fb0c275bfe32318c044a408603ca769f9/images/fedora/f36/Containerfile#L3) but we already have [reports](https://github.com/containers/toolbox/issues/188) of these variables interfering with other tools. While it is not possible to not use them for the foreseeable future due to the way Fedora OCI image machinery works, it is only specific to the Fedora images and they are not Toolbx-specific.

To read the name of the container, I'd read the file `/run/.containerenv` created by Podman (which is a dependency of Toolbx) and get the name from there. The content of the file can look like this:

```
engine="podman-3.4.2"
name="toolbox-dev"
id="236e7a414947759564bb2bc2d5cb3fc4a5999c7e8457bc763aaf5361a3f700c8"
image="registry.fedoraproject.org/fedora-toolbox:36"
imageid="3ef9a4eaf734b79fcafc0f846efb8a8e8850cfc81c66a9e8aea10c6637623a04"
rootless=1
graphRootMounted=1⏎
```

HarryMichal (changes requested):
First of all, thank you for your work on this! I started using starship recently and it makes me happy that I don't have to add support for our tool to it myself.

I left some comments regarding the implementation. It mainly touches on the use of the different env vars. Also note that Toolbox has been [renamed](https://debarshiray.wordpress.com/2021/11/10/toolbox-is-now-toolbx/) to Toolbx (despite the fact that the rename has not occured yet in the implementation and most of the docs).

If you have any further questions, ping me :).

src/modules/toolbox.rs
alphacc: Thanks, yes it's better. I'll work on that.

src/modules/toolbox.rs
alphacc: noted, it's more elegant indeed.

src/configs/toolbox.rs
alphacc: Ok I'll add that !

jankatins:
Another more generic "container" PR: https://github.com/starship/starship/pull/3304

PR #2597, opened 2021-04-17
docs: Replace the section title "Gcloud" with "gcloud"

shuuji3:
#### Description

This is a PR for the documentation just replacing the section title "Gcloud" with "gcloud" here: https://starship.rs/config/#gcloud.

#### Motivation and Context

The official representation of the `gcloud` command-line tool is `gcloud` not `Gcloud` and there is no chance the word is written as `Gcloud`.  (ref. https://cloud.google.com/sdk/gcloud) The starship module name is also `gcloud`. I understand all section names in the config page start with the capital case but I feel `Gcloud` weired a little and the `gcloud` is more natural.

None

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1020,7 +1020,7 @@ By default the module will be shown if any of the following conditions are met:
 format = "via [e $version](bold red) "
 ```
 
-## gcloud
+## Google Cloud (`gcloud`)
 
 The `gcloud` module shows the current configuration for [`gcloud`](https://cloud.google.com/sdk/gcloud) CLI.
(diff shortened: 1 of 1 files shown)


vladimyr:
FWIW, Google doesn't abbreviate it at all. They use Google Cloud (Platform). So maybe we can go with:
```md
## Google Cloud (`gcloud`)
```
That lets users search for both Google and gcloud (lowercased as the law mandates 😜).

shuuji3:
I think it's better than the only "gcloud". I'll update the commit.

vladimyr (approved):
Let's have nice (sub)titles! 🚀

andytom:
Thanks for the fix @shuuji3 and thanks for reviewing @vladimyr.

shuuji3:
Thank you for merging and the good suggestion for @vladimyr!

PR #1418, opened 2020-06-30
fix!: remove the haskell module

matchai:
#### Description

Given how slow the Haskell module is (#1240), it is slowing down the entire prompt
from rendering when the module is active. This commit removes the module until we
can find a faster way to retreive the Haskell version.

#### Motivation and Context

Closes #1240 
We will open a new issue to reimplement the Haskell module in a more performant way.

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #2601, opened 2021-04-17
feat(haskell): create module

### LINKED ISSUE #1458: Support for Haskell
My feature request isn't really regarding a problem, but perhaps if [spaceship prompt](https://github.com/denysdovhan/spaceship-prompt) parity is a goal then it is? I noticed that there is no icon for haskell / ghc, and I think it would be great if there was!

A Haskell module could be added, and it could look for the presence of a `stack.yaml | package.yaml` file (which would contain the corresponding lts / ghc version) or a `*.cabal` file, and if you REALLY want to support all projects a `default.nix` file. The Haskell module could be represented by the vanilla unicode lambda λ, as it was in spaceship prompt, or (if starship is committed to nerdfonts) could use the nerdfont haskell symbol `nf-dev-haskell`)

You could just not support haskell, which is fair, but I know a lot of the haskell community feels strongly tied to the rust community due to our shared reverence for type safety, and I think a lot of haskell users would be keen to adopt this project for their prompts.

NOTE: I would love to help implement, I have 0 rust experience, but if someone is willing to give very clear instructions (I am sure the pattern for adding a module is well established) and a bit of mentorship I could take a run at this.

Milo123459:
#### Description

Adds a haskell module
#### Motivation and Context

Closes #1458

img

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1328,6 +1328,46 @@ By default the module will be shown if any of the following conditions are met:
 format = "via [🏎💨 $version](bold cyan) "
 ```
 
+## Haskell
+
+The `haskell` module shows the currently installed version of Haskell.
+By default the module will be shown if any of the following conditions are met:
+
+- The current directory has a file with the extension of `.hs`, `.hsc` or `.hs-boot`.
+
+### Options
+
+| Option              | Default                              | Description                                      |
+| ------------------- | ------------------------------------ | ------------------------------------------------ |
+| `format`            | `"via [$symbol($version )]($style)"` | The format for the module.                       |
+| `detect_extensions` | `["hs", "hsc", "hs-boot"]`                                 | Which extensions should trigger this module.     |
+| `detect_files`      | `[]`    | Which filenames should trigger this module.      |
+| `detect_folders`    | `[]`                                 | Which folders should trigger this modules.       |
+| `symbol`            | `" "`                               | A format string representing the symbol of Haskell. |
+| `style`             | `"bold purple"`                       | The style for the module.                        |
+| `disabled`          | `false`                              | Disables the `haskell` module.                      |
+
+### Variables
+
+| Variable | Example  | Description                          |
+| -------- | -------- | ------------------------------------ |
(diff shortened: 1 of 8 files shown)


davidkna:
Actually, previously there was a haskell module, but it was removed because getting the version was too slow (#1418). Speed might be less of a concern now that command timeouts exist though. I think invocation via `stack` like in the old module is necessary to avoid side-effects (downloading and installing ghc?).

andytom:
It might also be worth taking a look at the linked ticket #1458 which explains some additional things to consider in this modules and some possible quicker ways to speed up the fetching of the version.

fosskers:
It would also be very helpful if the Stackage LTS resolver named in the project's `stack.yaml` was shown (if available).

docs/config/README.md
```diff
@@ -1328,6 +1328,44 @@ By default the module will be shown if any of the following conditions are met:
 format = "via [🏎💨 $version](bold cyan) "
 ```
 
+## Haskell
+
+The `haskell` module shows the currently installed version of Haskell.
+By default the module will be shown if any of the following conditions are met:
+
+- The current directory has a file with the extension of `.hs`, `.hsc` or `.hs-boot`.
```
fosskers, line 1336: I believe the criteria should be if there is a `stack.yaml` or  `foo.cabal` present at the project root.

Otherwise yes, if I go deeper into source directories, I do see the Haskell logo pop up (but no GHC version).

matdsoupe: yea, I agree with it!

Milo123459: This PR has kinda been ignored / neglected. If there is demand for it I can rewrite to fix it, but, I'm not sure right now.

fosskers: Oh certainly I was looking forward to it. That's not a demand, of course, we've all got lives.

fosskers: The main thing I was interested in was the detection and display of the `resolver` field in a `stack.yaml`, if such a file is present in the project.

Milo123459: Alright

PR #2609, opened 2021-04-19
feat(aws): Add support for profile from awsume

andymac4182:
#### Description
We use AWESUME to assume into different accounts. It stores the profile in its own environment variable and sets the environment variables for the access key and secret key. 

#### Motivation and Context
Show same info as aws-vault

#### Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -47,7 +47,7 @@ fn get_aws_region_from_config(context: &Context, aws_profile: Option<&str>) -> O
 }
 
 fn get_aws_profile_and_region(context: &Context) -> (Option<Profile>, Option<Region>) {
-    let profile_env_vars = vec!["AWSU_PROFILE", "AWS_VAULT", "AWS_PROFILE"];
+    let profile_env_vars = vec!["AWSU_PROFILE", "AWS_VAULT", "AWSUME_PROFILE", "AWS_PROFILE"];
     let profile = profile_env_vars
         .iter()
         .find_map(|env_var| context.get_env(env_var));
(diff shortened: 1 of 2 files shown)


src/modules/aws.rs
```diff
@@ -47,7 +47,7 @@ fn get_aws_region_from_config(context: &Context, aws_profile: Option<&str>) -> O
 }
 
 fn get_aws_profile_and_region(context: &Context) -> (Option<Profile>, Option<Region>) {
-    let profile_env_vars = vec!["AWSU_PROFILE", "AWS_VAULT", "AWS_PROFILE"];
+    let profile_env_vars = vec!["AWSU_PROFILE", "AWS_VAULT", "AWSUME_PROFILE", "AWS_PROFILE"];
```
vladimyr, line 50: Why this specific order? Are they intentionally ordered this way?

andytom:
Thanks for adding this @andymac4182, can you also update the English version of the docs.

src/modules/aws.rs
andymac4182: I used the existing order to determine this.

In my case if there is an AWSUME_PROFILE defined that means that AWSUME has ran and set the credentials to use. I would assume that will be the same for AWS_VAULT and AWSU_PROFILE which do similar things to AWSUME.

andymac4182:
I will update the docs for this. :)

src/modules/aws.rs
vladimyr: I get that my question was more like why `AWSU_PROFILE` comes before `AWSUME_PROFILE` and why `AWS_VAULT` has higher search priority than the latter. I guess we are safe to assume that users won't have multiple AWS credential managers installed so we are good even the search order seems pretty random.

andymac4182:
Thanks for reviewing @andytom I have updated the docs.

src/modules/aws.rs
andymac4182: Ahh. I would say the ordering is probably based on the order added. Personally it should be any credential manager then the AWS one. Most people will only use a single credential manager so it won't matter but if you have more than one then I don't think the ordering will help either as you would need to know which one was executed last.

vladimyr (commented):
Please add a new test modeled after  `profile_set_from_awsu`.

vladimyr (approved):
LGTM 👍

andytom:
Thank you for your contribution @andymac4182 and thank you for reviewing @vladimyr.

PR #2611, opened 2021-04-19
feat: Add version formating for modules

### LINKED ISSUE #2576: Use the version format string for all versions

In #2499 and #2566 we added the ability to customise how we format the versions this has been added to a few modules (`python`, `node`, `java`, `rust` and `ruby`) but it would be good to add it to all the other modules that display a version.

All modules that display some kind of version can use the format version.

cymruu:
#### Description
Add support for version formatting for the rest of modules. 
 
#### Motivation and Context
Closes #2576

#### Checklist:

### DIFF
--- a/src/formatter/version.rs
+++ b/src/formatter/version.rs
@@ -1,6 +1,5 @@
 use super::string_formatter::StringFormatterError;
 use super::StringFormatter;
-use crate::module::Module;
 use once_cell::sync::Lazy;
 use std::ops::Deref;
 use versions::Versioning;
@@ -63,18 +62,14 @@ impl<'a> VersionFormatter<'a> {
     }
 
     pub fn format_module_version(
-        module: &Module,
+        module_name: &str,
         version: &str,
         version_format: &str,
     ) -> Option<String> {
         match VersionFormatter::format_version(version, version_format) {
             Ok(formatted) => Some(formatted),
             Err(error) => {
-                log::warn!(
-                    "Error formatting `{}` version:\n{}",
-                    module.get_name(),
-                    error
-                );
+                log::warn!("Error formatting `{}` version:\n{}", module_name, error);
                 Some(format!("v{}", version))
             }
         }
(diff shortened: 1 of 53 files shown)


src/formatter/version.rs
```diff
@@ -60,6 +61,24 @@ impl<'a> VersionFormatter<'a> {
                 .collect::<String>()
         })
     }
+
+    pub fn format_module_version(
+        module: &Module,
```
vladimyr, line 66: Well, you only need the module name, why not pass that instead?

andytom:
Hi @cymruu, can you fix the merge conflicts and I think we should be about ready to merge this.

offbyone:
Oh, _very_ nice! Thanks for coming in and doing all the work I was too lazy to finish! :D

andytom (approved):
LGTM, but would like a second opinion since this touches a lot of modules and I might have missed something.

offbyone:
It all looks good to me, FWIW. I think we should let the workflow run to make sure it builds cleanly.

andytom:
> It all looks good to me, FWIW. I think we should let the workflow run to make sure it builds cleanly.

Huh, I'd not seen that before must have been a new change for Github.

offbyone:
It is: https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/

Mainly working to prevent abuse by cryptocurrency miners.

offbyone:
(I'd have hit "go" on that myself, but ... 
image

:D

andytom:
Thank you for your contribution @cymruu and thanks for reviewing @offbyone.

andytom:
> It is: https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/
> 
> Mainly working to prevent abuse by cryptocurrency miners.

Ah ok that explains it. Annoyingly it doesn't show anything in the app.

vladimyr:
This was truly a Herculean effort! :heart: :100: 

I have one small suggestion though. Let's agree on naming. Either we decide to go with `get_something_version` or with `parse_something_version` (personally, I believe the latter is a better name) and then we should ensure that's consistent across all modules.

PR #2612, opened 2021-04-19
fix(install): if glibc >= 2.18, use the dynamically linked binary

### LINKED ISSUE #5361: `install.sh` detects Linux Mint as musl instead of glibc distro
#### Current Behavior

Using `curl -sS https://starship.rs/install.sh | sh`  suggests to install the `musl` version of starship instead of `glibc`

```
# curl -sS https://starship.rs/install.sh | sh
  Configuration
> Bin directory: /usr/local/bin
> Platform:      unknown-linux-musl
> Arch:          x86_64

> Tarball URL: https://github.com/starship/starship/releases/latest/download/starship-x86_64-unknown-linux-musl.tar.gz
? Install Starship latest to /usr/local/bin? [y/N] N
x Aborting (please answer "yes" to continue)
```

#### Expected Behavior

The installer should have downloaded `https://github.com/starship/starship/releases/download/v1.16.0/starship-x86_64-unknown-linux-gnu.tar.gz`

```
# uname -a
Linux forza-virtual-machine 6.4.4-forza-mint #1 SMP PREEMPT Thu Jul 20 13:53:41 CEST 2023 x86_64 x86_64 x86_64 GNU/Linux

# cat /etc/os-release
NAME="Linux Mint"
VERSION="21.2 (Victoria)"
ID=linuxmint
ID_LIKE="ubuntu debian"
PRETTY_NAME="Linux Mint 21.2"
VERSION_ID="21.2"
HOME_URL="https://www.linuxmint.com/"
SUPPORT_URL="https://forums.linuxmint.com/"
BUG_REPORT_URL="http://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/"
PRIVACY_POLICY_URL="https://www.linuxmint.com/"
VERSION_CODENAME=victoria
UBUNTU_CODENAME=jammy

# bash --version
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)

# ldd /bin/bash
        linux-vdso.so.1 (0x00007ffd662da000)
        libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007f3ad3a9e000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3ad3800000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f3ad3c4a000)
```
#### Environment
- Starship version: 1.16.0
- Operating system: Linux Mint 21.2
- Git Commit Hash: 
- Branch/Tag: master
- Rust Version: rustc 1.71.0 (8ede3aae2 2023-07-12)
- Rust channel:  release
- Build Time: 2023-07-30 13:36:32 +00:00

```bash
<unknown config>
```

#### Starship Configuration

```toml
<unknown config>
```

michel-slm:
#### Description
`install.sh` previously uses the statically linked `unknown-linux-musl` build
for every Linux installation. If the host's `libc` is a recent enough
GNU libc, however,  it should use this rather than the statically linked build.

Signed-off-by: Michel Alexandre Salim <michel@michel-slm.name>

#### Motivation and Context
The dynamic build is smaller, and in case there are security vulnerabilities affecting the `libc` implementation, users who don't upgrade to the latest static build will get the fix for free when their distribution releases a security update.

Closes #5361

Tested by:
- short-circuiting `install.sh` to print the result of `detect_platform` and exiting
- testing on Vagrant on the following distributions and verifying that the `unknown-linux-gnu` build works:
  - Arch Linux
  - CentOS 8
  - CentOS Stream 8
  - Debian testing
  - Fedora
  - Ubuntu 20.04

#### Checklist:

I don't think there are tests that cover this yet. I've updated the inline docs in `install.sh`

---
This change is [Reviewable](https://reviewable.io/reviews/starship/starship/2612)

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -193,8 +193,34 @@ detect_platform() {
     cygwin_nt*) platform="pc-windows-msvc";;
     # mingw is Git-Bash
     mingw*) platform="pc-windows-msvc" ;;
-    # use the statically compiled musl bins on linux to avoid linking issues.
-    linux) platform="unknown-linux-musl" ;;
+    linux)
+      # default to the statically compiled musl bins on linux
+      # to avoid linking issues.
+      platform="unknown-linux-musl"
+      # musl's ldd from Alpine prints to stderr
+      ldd_verstr="$(ldd --version 2>&1 | head -n1)"
+      # glibc version string is inconsistent across distributions
+      # instead check if the string does not contain musl (case insensitive)
+      if echo "${ldd_verstr}" | grep -iv musl >/dev/null; then
+        glibc_version=$(echo "${ldd_verstr}" | awk 'NR==1 { print $NF }')
+        # on installations with a new enough glibc, use the dynamically compiled binary
+        #
+        # use this to see potential max versions
+        # strings $(which starship) | grep -i glibc
+        #
+        # then use https://gist.github.com/fasterthanlime/17e002a8f5e0f189861c with
+        # `MAX_VER` set to the version from earlier e.g. GLIBC_2.33 => 2.33
+        glibc_maj_ver=$(echo "${glibc_version}" | cut -d. -f1)
+        glibc_min_ver=$(echo "${glibc_version}" | cut -d. -f2)
+        min_maj_ver=2
+        min_min_ver=18
+        if [ "${glibc_maj_ver}" -gt "${min_maj_ver}" ] \
+          || { [ "${glibc_maj_ver}" -eq "${min_maj_ver}" ] \
+          && [ "${glibc_min_ver}" -ge "${min_min_ver}" ]; }; then
+          platform="unknown-linux-gnu"
+        fi
+      fi
+      ;;
     darwin) platform="apple-darwin" ;;
     freebsd) platform="unknown-freebsd" ;;
   esac


install/install.sh
```diff
@@ -193,8 +193,20 @@ detect_platform() {
...
     mingw*) platform="pc-windows-msvc" ;;
-    # use the statically compiled musl bins on linux to avoid linking issues.
-    linux) platform="unknown-linux-musl" ;;
+    linux)
+      # default to the statically compiled musl bins on linux
+      # to avoid linking issues.
+      platform="unknown-linux-musl"
+      if [ -f /etc/os-release ]; then
+	# on known-good distributions, use the dynamically compiled bins
+	distro=$(cat /etc/os-release | grep '^ID=' | sed -e 's/^ID=//')
```
vladimyr, line 202: Just source it and work directly with `$DISTRIB_ID`?

davidkna:
I think the main issue wasn't whether a distro uses glibc, but whether the system glibc version was compatible.

install/install.sh
vladimyr: I pointed you towards neofetch for a good reason :upside_down_face: Here is how they do it:
```bash
            elif [[ -f /etc/os-release || \
                    -f /usr/lib/os-release || \
                    -f /etc/openwrt_release || \
                    -f /etc/lsb-release ]]; then

                # Source the os-release file
                for file in /etc/lsb-release /usr/lib/os-release \
                            /etc/os-release  /etc/openwrt_release; do
                    source "$file" && break
                done

                # Format the distro name.
                case $distro_shorthand in
                    on)   distro="${NAME:-${DISTRIB_ID}} ${VERSION_ID:-${DISTRIB_RELEASE}}" ;;
                    tiny) distro="${NAME:-${DISTRIB_ID:-${TAILS_PRODUCT_NAME}}}" ;;
                    off)  distro="${PRETTY_NAME:-${DISTRIB_DESCRIPTION}} ${UBUNTU_CODENAME}" ;;
                esac
```
Also, you're gonna need lowercase it because here is what I get on Linux Mint 19.2:
```sh
# from /etc/lsb-release
DISTRIB_ID=LinuxMint

# from /etc/os-release
ID=linuxmint
```

vladimyr: And you can extend that (allow)list with Linux Mint because it is fairly popular Ubuntu derivate :wink:

vladimyr:
@davidkna Do you happen to know what's the minimal supported glibc version?

davidkna:
It's more about incompatible ABI versions with any other version. Cross uses v2.15, but it might also use whatever ubuntu is at from the CI env.

vladimyr:
Reading through https://github.com/rust-lang/libc/issues/1412 `2.28` seems like a safe bet.
Searching for `glibc >= 2.28` on DistroWatch yields a lot of distros: https://distrowatch.com/search.php?pkg=glibc&relation=greaterequal&pkgver=2.28&distrorange=InLatest#pkgsearch

install/install.sh
davidkna: I think I prefer the old check https://github.com/starship/starship/commit/2e14d1af5a3b90aedab39180e0851e8f945b9f55

vladimyr:
```sh
if [ "$platform" = "linux" ] && ldd /bin/sh | grep -iq musl; then
  platform=unknown-linux-musl
fi
```

vladimyr:
I'm not sure how to interpret this but I'm pasting it here for the convenience:
```
~ 
❯ strings $(which starship) | grep -i glibc
GLIBC_2.2.5
GLIBC_2.3.3
GLIBC_2.3.2
GLIBC_2.15
GLIBC_2.33
GLIBC_2.17
GLIBC_2.14
GLIBC_2.9
GLIBC_2.18
GLIBC_2.3.4
GLIBC_2.3
GLIBC_2.32
```

install/install.sh
michel-slm: so, something like the old check, but flipping it to detect glibc, would be ideal if we also add a check for the glibc version that works in a cross platform way, right?

michel-slm: this seems to work. I'll rework the PR to use it
```
$ ldd --version | head -n1 | sed -e 's/ldd (GNU libc) //'
2.28
```

(first detecting if the string contains GNU libc, of course)

vladimyr:
> @vladimyr yeah, so now the problem is... older versions of a given distro might have a glibc that is too old. Especially for long-lived distros like Debian oldstable or older CentOS releases

So current Debian oldstable is Debian 9 (stretch) and it ships with glibc 2.24 which might be a problem but the current stable Debian 10 (buster) ships with 2.28 which should work. Debian 9 isn't EOLed AFAIK but there is a newer stable version and supporting older versions shouldn't be Starship's priority anyway.

CentOS 7 ships with glibc 2.17 and the latest CentOS 8 ships with 2.28 but there is a whole CentOS situation with that abrupt EOL notice so I guess we shouldn't be concerned too much about it.

I understand your point and IMHO we should support only popular and regularly updated (non-rolling) distros by supporting their latest stable versions.

install/install.sh
vladimyr: You can simplify it by replacing `head` plus `sed` combo with awk:
```sh
ldd --version | awk 'NR==1 { print $NF }'
# or invoking /usr/lib/libc.so.6 directly
/usr/lib/libc.so.6 | awk 'NR==1 { print $NF }'
```

vladimyr:
According to Amos's checker https://gist.github.com/fasterthanlime/17e002a8f5e0f189861c with `MAX_VER` set to `2.28` latest Starship (v0.51) depends on:
```
~/scratchpad 
❯ ./glibc-check.sh ./starship                                                      
Found version 2.14
Found version 2.15
Found version 2.17
Found version 2.18
Found version 2.2.5
Found version 2.3
Found version 2.3.2
Found version 2.3.3
Found version 2.3.4
Found version 2.4
Found version 2.7
Found version 2.9
```

I guess this means the minimum supported glibc version is 2.18?

vladimyr:
:warning: Make sure to check it using `shellcheck`:
```
Line 202:
      if [[ "${ldd_verstr}" == "${ldd_glibc_prefix}"* ]]; then
         ^-- SC3010: In POSIX sh, [[ ]] is undefined.
 
Line 205:
        echo $glibc_version
             ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean: (apply this, apply all SC2086)
        echo "$glibc_version"
 
Line 207:
        if [[ "${glibc_version}" > "${min_version}" ]] \
           ^-- SC3010: In POSIX sh, [[ ]] is undefined.
 
Line 208:
           || [[ "${glibc_version}" == "${min_version}" ]]; then
              ^-- SC3010: In POSIX sh, [[ ]] is undefined.
```

install/install.sh
vladimyr: Well, in that case you can use following checks:
```sh
ldd --version | awk '/GNU libc/ { print $NF }'
# or invoking /usr/lib/libc.so.6 directly
/usr/lib/libc.so.6 | awk '/GNU libc/ { gsub(/\.$/,""); print $NF }'
```

vladimyr:
> It's more about incompatible ABI versions with any other version. Cross uses v2.15, but it might also use whatever ubuntu is at from the CI env.

Apparently, it used glibc `v2.18` and our [build runner](https://github.com/starship/starship/runs/2177112874) for `v0.51` release was powered by Ubuntu 20.04.2 which ships with glibc `v2.31`.

install/install.sh
vladimyr: > I can use `awk` if that's preferable

The sole reason why I prefer `awk` here is to minimize calls to external binaries.

michel-slm:
ah, didn't notice we're using /bin/sh. Will recheck and drop the requirement to 2.18

michel-slm:
odd, Fedora's shellcheck 0.7.1 does not detect SC3010 issues when I pass `shellcheck -o all` or `shellcheck -o SC3010`, but pasting in `shellcheck.net` does bring up that issue as well as some pre-existing issues.

michel-slm:
@vladimyr fixed the `SC3010` warnings, dropped the min version to 2.18, and added instructions for computing the minimum version based on your comments. Thanks!

install/install.sh
```diff
@@ -193,8 +193,28 @@ detect_platform() {
...
     # mingw is Git-Bash
     mingw*) platform="pc-windows-msvc" ;;
-    # use the statically compiled musl bins on linux to avoid linking issues.
-    linux) platform="unknown-linux-musl" ;;
+    linux)
+      # default to the statically compiled musl bins on linux
+      # to avoid linking issues.
+      platform="unknown-linux-musl"
+      ldd_verstr="$(ldd --version | head -n1)"
+      ldd_glibc_prefix="ldd (GNU libc) "
```
vladimyr, line 201: The wording varies between vendors:
```sh
# linuxmint
$ ldd --version
ldd (Ubuntu GLIBC 2.27-3ubuntu1.4) 2.27
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.

# ubuntu
ldd (Ubuntu EGLIBC 2.19-0ubuntu6.15) 2.19
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.

# arch
ldd (GNU libc) 2.33
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

install/install.sh
```diff
@@ -193,8 +193,28 @@ detect_platform() {
...
+      if test "${ldd_verstr#*$ldd_glibc_prefix}" != "$ldd_verstr"; then
```
vladimyr, line 202: Why `test` instead of `[]`?

install/install.sh
```diff
@@ -193,8 +193,28 @@ detect_platform() {
...
+        glibc_version=$(echo "${ldd_verstr}" | sed -e "s/^${ldd_glibc_prefix}//")
+        # on installations with a new enough glibc, use the dynamically compiled binary
+        #
+        # use this to see potential max versions
+        # strings $(which starship) | grep -i glibc
+        #
+        # then use https://gist.github.com/fasterthanlime/17e002a8f5e0f189861c with
+        # `MAX_VER` set to the version from earlier e.g. GLIBC_2.33 => 2.33
+        min_version="2.18"
+        if [ "${glibc_version}" > "${min_version}" ] \
```
vladimyr, line 212: This is redirection unless escaped: `\>` :warning: 
Even then it is lexical comparison between strings rather than numerical one:
```sh
~ $ [ 3.22 \> 3.3 ] && echo 'true' || echo 'false'
false
```

install/install.sh
```diff
@@ -313,6 +333,8 @@ is_build_available() {
 # defaults
 if [ -z "${PLATFORM-}" ]; then
   PLATFORM="$(detect_platform)"
+  echo ${PLATFORM}
+  # exit 0
```
vladimyr, line 337: Debugging leftovers I guess? :upside_down_face:

vladimyr (changes requested):

install/install.sh
michel-slm: Woops

install/install.sh
michel-slm: We can't do numerical comparison, sadly, since glibc version numbers might sometimes be of x.y.z form

I guess the cleanest way is to split based on dot and explicitly compare the parts as integers

vladimyr: > I guess the cleanest way is to split based on dot and explicitly compare the parts as integers

Right, plus making sure that missing semver segments default to zero. :+1:

install/install.sh
michel-slm: fixed

install/install.sh
michel-slm: fixed. Right now we can just compare major and minor, if we ever need to compare patch versions we can add that comparison then.

install/install.sh
michel-slm: oh dear. not good. might be worth figuring out how to detect the musl version string instead.

install/install.sh
michel-slm: no particular reason, I can switch it back

michel-slm:
@vladimyr 
- used your awk code snippet to extract the version
- since there are so many different glibc version strings, I checked whether `musl` is missing from the first line instead. Also noticed that musl's ldd outputs the version to stderr, not stdout (ugh!) so I redirected stderr to stdout before testing

PR #2613, opened 2021-04-19
fix(jobs): Handle zero jobs with zero thresholds

### LINKED ISSUE #2610: [jobs] Cannot take negative value (zero jobs not shown)
#### Current Behavior

```toml
[jobs]
threshold = -1
```

does not show 0 (zero) in the output if there are no background jobs.

#### Expected Behavior

`0` (zero) is printed for zero background jobs

#### Environment

```
$ starship --version
starship 0.51.0
branch:
commit_hash:
build_time:2021-03-29 15:35:00
build_env:rustc 1.50.0,
```

- Shell type: bash
- Shell version: 4.4.23
- Terminal emulator: Konsole
- Operating system: NixOS Linux
-

Anthuang:
First PR here, let me know if anything can be improved!

#### Description
Currently, if there are 0 jobs running then the jobs module doesn't show anything. However, if a negative threshold is specified (user wants to explicitly see that there are 0 jobs running), then it should show 0.

Closes #2610

#### Checklist:

### DIFF
--- a/src/modules/jobs.rs
+++ b/src/modules/jobs.rs
@@ -15,6 +15,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .trim()
         .parse::<i64>()
         .ok()?;
+
+    if config.threshold < 0 {
+        log::warn!(
+            "threshold in [jobs] ({}) was less than zero",
+            config.threshold
+        );
+        return None;
+    }
+
     if num_of_jobs == 0 && config.threshold > 0 {
         return None;
     }
--- a/src/modules/jobs.rs
+++ b/src/modules/jobs.rs
@@ -28,11 +28,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let module_number = if num_of_jobs > config.threshold {
+    let module_number = if num_of_jobs > config.threshold || config.threshold == 0 {
         num_of_jobs.to_string()
-    } else if num_of_jobs == 0 && config.threshold == 0 {
-        // Explicitly show when there are 0 background jobs
-        "0".to_string()
     } else {
         "".to_string()
     };
--- a/src/modules/jobs.rs
+++ b/src/modules/jobs.rs
@@ -15,7 +15,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .trim()
         .parse::<i64>()
         .ok()?;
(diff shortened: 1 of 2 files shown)


src/modules/jobs.rs
```diff
@@ -15,12 +15,24 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+
+    if config.threshold < 0 {
+        log::warn!(
+            "threshold in [jobs] ({}) was less than zero",
+            config.threshold
+        );
+        return None;
+    }
+
+    if num_of_jobs == 0 && config.threshold > 0 {
```
vladimyr, line 27:
```suggestion
    if num_of_jobs < config.threshold {
```

src/modules/jobs.rs
```diff
@@ -15,12 +15,24 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
         return None;
     }
 
     let module_number = if num_of_jobs > config.threshold {
         num_of_jobs.to_string()
+    } else if num_of_jobs == 0 && config.threshold == 0 {
+        // Explicitly show when there are 0 background jobs
+        "0".to_string()
```
vladimyr, line 35: You don't need separate branch. If threshold is zero number will always get printed.
```suggestion
    let module_number = if num_of_jobs > config.threshold || config.threshold == 0 {
        num_of_jobs.to_string()
    } else {
        "".to_string()
```

src/modules/jobs.rs
Anthuang: The previous behavior was that if there are _any_ background jobs running, then the module would show a symbol, regardless of the threshold. Should we try to keep this same behavior?

vladimyr: Oh right, good catch! :+1: Let's keep existing behavior.

vladimyr (approved):
> First PR here, let me know if anything can be improved!

Sure, do more great PRs like this one! :heart: 

LGTM

vladimyr:
Actually, I forgot, docs need to get updated too :upside_down_face:

Anthuang:
> > First PR here, let me know if anything can be improved!
> 
> Sure, do more great PRs like this one! ❤️
> 
> LGTM

Thank you for the reviews!! And I just added the docs, lmk if it sounds good or if I should change the wording

andytom:
Thank you @Anthuang for your contribution and thank you @vladimyr for the review.

PR #2614, opened 2021-04-20
fix(build): respect the lockfile

vladimyr:
#### Description

`cargo build` lacks `--locked` argument resulting in `Cargo.lock` being ignored and the latest version of dependencies being used instead.

#### Motivation and Context

Ensure build pipeline is deterministic.

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -70,7 +70,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --all-features -- -D clippy::all
+          args: --workspace --locked --all-targets --all-features -- -D clippy::all
 
   # Ensure that the project could be successfully compiled
   cargo_check:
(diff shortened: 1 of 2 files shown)


offbyone:
This looks obviously correct. I can't +1 it formally, but +1 anyway.

vladimyr:
@offbyone Thanks for verifying it! :+1:

davidkna:
I think you should also add `--locked` to `cargo test`

vladimyr:
@davidkna As usual, you are absolutely right! Thanks!

.github/workflows/workflow.yml
```diff
@@ -98,7 +98,7 @@ jobs:
           override: true
 
       - name: Build | Check
-        run: cargo check --all
+        run: cargo check --workspace --locked
```
vladimyr, line 101: Option `--all` is deprecated and replaced with `--workspace`.

davidkna (approved):
LGTM

PR #2626, opened 2021-04-22
feat(nodejs): Check for `.nvmrc`

kidonng:
#### Description

Check for `.nvmrc` file presence in `nodejs` module.

#### Motivation and Context

Starship already checks for `.node-version`, and [`.nvmrc`](https://github.com/nvm-sh/nvm#nvmrc) is another popular alternative that worth considering.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1863,6 +1863,7 @@ By default the module will be shown if any of the following conditions are met:
 
 - The current directory contains a `package.json` file
 - The current directory contains a `.node-version` file
+- The current directory contains a `.nvmrc` file
 - The current directory contains a `node_modules` directory
 - The current directory contains a file with the `.js`, `.mjs` or `.cjs` extension
 - The current directory contains a file with the `.ts` extension
--- a/src/configs/nodejs.rs
+++ b/src/configs/nodejs.rs
@@ -26,7 +26,7 @@ impl<'a> Default for NodejsConfig<'a> {
             disabled: false,
             not_capable_style: "bold red",
             detect_extensions: vec!["js", "mjs", "cjs", "ts"],
-            detect_files: vec!["package.json", ".node-version"],
+            detect_files: vec!["package.json", ".node-version", ".nvmrc"],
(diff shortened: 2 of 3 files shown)


vladimyr (approved):
LGTM

---

Btw, it would be nice to also take `.node_version`/`.nvmrc` into account when _coloring_ node version, talking specifically about the `not_capable_style` option. But that is a matter for another PR.

kidonng:
> Btw, it would be nice to also take `.node_version`/`.nvmrc` into account when _coloring_ node version, talking specifically about the `not_capable_style` option. But that is a matter for another PR.

That's not trivial to implement. Both file don't have a fixed format, [more info](https://github.com/tj/n#specifying-nodejs-versions).
However, if someone wants to try, `lts`, `latest` and `current` besides version numbers should be supported at least.

vladimyr:
> That's not trivial to implement.

A smart man once told me: if it was easy I would do it myself. 😀

> Both file don't have a fixed format, making it hard to parse and handle, [more info](https://github.com/tj/n#specifying-nodejs-versions).

Jokes aside I absolutely love informed responses! You made my day ❤️ 

Here is how I see it. If it is a numeric version, complete, incomplete, w, or w/o prefix it should be supported by the existing semver parser crate we already use. Support for lts, latest and current is IMO, out of scope. Those are moving targets and supporting them requires constant changing. Argon, boron, carbon, and friends are easier to support because those names are known upfront https://github.com/nodejs/Release/blob/main/CODENAMES.md

If you think about it is a sort of progressive enhancement. You obviously can't support everything but in those cases, you don't lose anything because right now users also don't get that actionable info - like in cases having `.nvmrc` with lts inside, for example. On the other hand, we could give them valuable feedback over a subset of possibilities most importantly in cases where a concrete version is specified, being probably most widely used. That seems like a clear win to me.

vladimyr:
Also if you agree we can create a new issue to move the convo there.

kidonng:
> Also if you agree we can create a new issue to move the convo there.

Sure, we need a issue to track this.

PR #2633, opened 2021-04-23
docs(dart): use double quotes for consistency

vladimyr:
#### Description

Alter Dart's configuration defaults to use double quotes like the rest of the modules.

#### Motivation and Context

Keeping things consistent plus single quote strings in TOML have a different meaning. FWIW, they are treated as raw strings thus ignoring escape sequences.

#### Checklist:


andytom:
Good catch @vladimyr

PR #2636, opened 2021-04-23
fix(fish): explicitly set scope in fish init script

kidonng:
#### Description

This PR consists of ~~two~~ three changes:

- Remove unnecessary `-l` argument to be consistent with the `set keymap` above, and prefix variable names with Starship to avoid conflicts.

  In fish, variables are automatically scoped to parent function if no scope is specified, so `-l` makes no difference for the two changed lines.
  (However, this is not the same for `set keymap` case, as `-l/--local` will scope it to `switch`.)

  This change is purely for styling.

- Explicitly set `$VIRTUAL_ENV_DISABLE_PROMPT`'s scope to global

  In fish, variables have a default scope of global while sourcing scripts. But user may try different things like putting Starship load scripts into a function.

  image

  Notice how `$VIRTUAL_ENV_DISABLE_PROMPT` is actually not set if Starship is loaded via a function.

- Erase `fish_mode_prompt` function instead of overwriting it with an empty function.

  This should be theoretically faster (though don't actually matter at all), and less confusing than an empty function.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -1,14 +1,14 @@
 function fish_prompt
     switch "$fish_key_bindings"
         case fish_hybrid_key_bindings fish_vi_key_bindings
-            set keymap "$fish_bind_mode"
+            set STARSHIP_KEYMAP "$fish_bind_mode"
         case '*'
-            set keymap insert
+            set STARSHIP_KEYMAP insert
     end
-    set exit_code $status
+    set STARSHIP_CMD_STATUS $status
     # Account for changes in variable name between v2.7 and v3.0
-    set starship_duration "$CMD_DURATION$cmd_duration"
-    ::STARSHIP:: prompt --status=$exit_code --keymap=$keymap --cmd-duration=$starship_duration --jobs=(count (jobs -p))
+    set STARSHIP_DURATION "$CMD_DURATION$cmd_duration"
+    ::STARSHIP:: prompt --status=$STARSHIP_CMD_STATUS --keymap=$STARSHIP_KEYMAP --cmd-duration=$STARSHIP_DURATION --jobs=(count (jobs -p))
 end
 
 # disable virtualenv prompt, it breaks starship
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -14,7 +14,9 @@ end
 # disable virtualenv prompt, it breaks starship
 set -g VIRTUAL_ENV_DISABLE_PROMPT 1
 
-function fish_mode_prompt; end
+# Remove default mode prompt
+builtin functions -e fish_mode_prompt
+
 set -gx STARSHIP_SHELL "fish"
 
 # Set up the session key that will be used to store logs
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -11,7 +11,7 @@ function fish_prompt
     ::STARSHIP:: prompt --status=$STARSHIP_CMD_STATUS --keymap=$STARSHIP_KEYMAP --cmd-duration=$STARSHIP_DURATION --jobs=(count (jobs -p))
 end
 
-# disable virtualenv prompt, it breaks starship
+# Disable virtualenv prompt, it breaks starship
 set -g VIRTUAL_ENV_DISABLE_PROMPT 1
 
 # Remove default mode prompt
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -5,14 +5,14 @@ function fish_prompt
         case '*'
             set keymap insert
     end
-    set -l exit_code $status
+    set exit_code $status
     # Account for changes in variable name between v2.7 and v3.0
-    set -l starship_duration "$CMD_DURATION$cmd_duration"
+    set starship_duration "$CMD_DURATION$cmd_duration"
     ::STARSHIP:: prompt --status=$exit_code --keymap=$keymap --cmd-duration=$starship_duration --jobs=(count (jobs -p))
 end
 
 # disable virtualenv prompt, it breaks starship
-set VIRTUAL_ENV_DISABLE_PROMPT 1
+set -g VIRTUAL_ENV_DISABLE_PROMPT 1
 
 function fish_mode_prompt; end
 set -gx STARSHIP_SHELL "fish"


src/init/starship.fish
```diff
@@ -5,14 +5,14 @@ function fish_prompt
         case '*'
             set keymap insert
     end
-    set -l exit_code $status
+    set exit_code $status
```
davidkna, line 8: If I understand this correctly `-l` would prevent this from setting a previously defined variable with the same name? It might be better to use `-l` everywhere in that case.

src/init/starship.fish
```diff
@@ -5,14 +5,14 @@ function fish_prompt
...
     # Account for changes in variable name between v2.7 and v3.0
-    set -l starship_duration "$CMD_DURATION$cmd_duration"
+    set starship_duration "$CMD_DURATION$cmd_duration"
     ::STARSHIP:: prompt --status=$exit_code --keymap=$keymap --cmd-duration=$starship_duration --jobs=(count (jobs -p))
 end
 
 # disable virtualenv prompt, it breaks starship
-set VIRTUAL_ENV_DISABLE_PROMPT 1
+set -g VIRTUAL_ENV_DISABLE_PROMPT 1
```
davidkna, line 15: I think it might be better to use `-gx` for consistency here.

src/init/starship.fish
kidonng: Yes, but we should use a more specific name like `starship_exit_code` from the beginning.

src/init/starship.fish
kidonng: I'm not sure, but this seems to be used by fish itself, i.e. it doesn't need to be exported.

src/init/starship.fish
davidkna: I would prefer `-l` everywhere, but prefixing everything with `starship_` is also fine.

davidkna (approved):
LGTM

src/init/starship.fish
```diff
@@ -1,20 +1,22 @@
...
-    set -l exit_code $status
+    set STARSHIP_CMD_STATUS $status
     # Account for changes in variable name between v2.7 and v3.0
-    set -l starship_duration "$CMD_DURATION$cmd_duration"
-    ::STARSHIP:: prompt --status=$exit_code --keymap=$keymap --cmd-duration=$starship_duration --jobs=(count (jobs -p))
+    set STARSHIP_DURATION "$CMD_DURATION$cmd_duration"
+    ::STARSHIP:: prompt --status=$STARSHIP_CMD_STATUS --keymap=$STARSHIP_KEYMAP --cmd-duration=$STARSHIP_DURATION --jobs=(count (jobs -p))
 end
 
 # disable virtualenv prompt, it breaks starship
```
vladimyr, line 14: I'm sorry but I can't help it :see_no_evil: 
```suggestion
# Disable virtualenv prompt, it breaks starship
```

vladimyr (approved):
LGTM :+1:

andytom:
Thank you for your contribution @kidonng and thanks for reviewing @vladimyr and @davidkna

vladimyr:
Strong :+1: for uppercasing stuff to keep it consistent with other init scripts. 

Going slightly OT, I found this listing interesting:
```
❯ rg -io 'STARSHIP\w+' src/init/starship.* | sort -u 
src/init/starship.bash:STARSHIP_CMD_STATUS
src/init/starship.bash:STARSHIP_DURATION
src/init/starship.bash:STARSHIP_END_TIME
src/init/starship.bash:starship_precmd
src/init/starship.bash:starship_precmd_user_func
src/init/starship.bash:starship_preexec
src/init/starship.bash:starship_preexec_all
src/init/starship.bash:STARSHIP_PREEXEC_READY
src/init/starship.bash:STARSHIP_SESSION_KEY
src/init/starship.bash:STARSHIP_SHELL
src/init/starship.bash:STARSHIP_START_TIME
src/init/starship.elv:STARSHIP_SESSION_KEY
src/init/starship.elv:STARSHIP_SHELL
src/init/starship.fish:starship_duration
src/init/starship.fish:STARSHIP_SESSION_KEY
src/init/starship.fish:STARSHIP_SHELL
src/init/starship.ion:STARSHIP_ION_CMD_DURATION
src/init/starship.ion:STARSHIP_ION_STATUS
src/init/starship.ion:STARSHIP_SESSION_KEY
src/init/starship.ion:STARSHIP_SHELL
src/init/starship.ps1:STARSHIP_SESSION_KEY
src/init/starship.ps1:STARSHIP_SHELL
src/init/starship.tcsh:STARSHIP_CMD_STATUS
src/init/starship.tcsh:STARSHIP_DURATION
src/init/starship.tcsh:STARSHIP_END_TIME
src/init/starship.tcsh:STARSHIP_PATH
src/init/starship.tcsh:STARSHIP_POSTCMD
src/init/starship.tcsh:STARSHIP_PRECMD
src/init/starship.tcsh:STARSHIP_SESSION_KEY
src/init/starship.tcsh:STARSHIP_SHELL
src/init/starship.tcsh:STARSHIP_START_TIME
src/init/starship.zsh:STARSHIP_CAPTURED_TIME
src/init/starship.zsh:STARSHIP_CMD_STATUS
src/init/starship.zsh:STARSHIP_DURATION
src/init/starship.zsh:starship_get_time
src/init/starship.zsh:STARSHIP_JOBS_COUNT
src/init/starship.zsh:starship_precmd
src/init/starship.zsh:starship_preexec
src/init/starship.zsh:STARSHIP_SESSION_KEY
src/init/starship.zsh:STARSHIP_SHELL
src/init/starship.zsh:STARSHIP_START_TIME
src/init/starship.zsh:starship_zle
```

Here is what I get after the filename prefix gets removed, and sorting becomes case insensitive:
```
❯ rg --no-filename -io 'STARSHIP\w+' src/init/starship.* | sort -uf                         
STARSHIP_CAPTURED_TIME
STARSHIP_CMD_STATUS
STARSHIP_DURATION
STARSHIP_END_TIME
starship_get_time
STARSHIP_ION_CMD_DURATION
STARSHIP_ION_STATUS
STARSHIP_JOBS_COUNT
STARSHIP_PATH
STARSHIP_POSTCMD
STARSHIP_PRECMD
starship_precmd_user_func
starship_preexec
starship_preexec_all
STARSHIP_PREEXEC_READY
STARSHIP_SESSION_KEY
STARSHIP_SHELL
STARSHIP_START_TIME
starship_zle
```

It might be a good idea to remove that `ION` part to unify variable names and standardize on either `CAPTURED_TIME` or `DURATION`. :thinking:

PR #2638, opened 2021-04-23
fix(zsh): Reverts changes to background jobs retaining improvements around START_TIME

### LINKED ISSUE #2505: Symbol for "jobs" not being shown.
#### Current Behavior
When I background a vim session, I see no indication of it in my terminal.

#### Expected Behavior
As per configuration below, I expect to see bold red `!!` at the beginning of my prompt when vim is in the background.

I'm not 100% sure if this started happening with 0.51 or even 0.50. I just noticed it this morning 🤷 

#### Environment
- Starship version: 0.51.0
- zsh version: zsh 5.8 (x86_64-apple-darwin20.1.0)
- Operating system: Mac OS 11.2.3
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.50.0
- Rust channel:  release
- Build Time: 2021-03-23 17:59:38

```bash
export ZDOTDIR=$HOME
export LANG="en_US.UTF-8"
export EDITOR=vim
export CLICOLOR=1
export HISTFILE=~/.histfile
export HISTSIZE=10000
export SAVEHIST=1000
export WORDCHARS=''
export ENHANCD_DISABLE_DOT=1
export RIPGREP_CONFIG_PATH=~/.ripgrep
export BAT_CONFIG_PATH=~/.bat
export FZF_DEFAULT_COMMAND="rg --files"
export FZF_CTRL_T_COMMAND="rg --files"
export ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE='fg=5'

# Experimental
export ZPLUG_HOME=/usr/local/opt/zplug
source $ZPLUG_HOME/init.zsh

zplug "willghatch/zsh-cdr"
zplug "zsh-users/zsh-syntax-highlighting"
zplug "b4b4r07/enhancd"
zplug "olets/zsh-abbr"

# Install plugins if there are plugins that have not been installed
if ! zplug check --verbose; then
    printf "Install? [y/N]: "
    if read -q; then
        echo; zplug install
    fi
fi

zplug load

GO_BINARIES=$HOME/go/bin
HOMEBREW_BINARIES="/usr/local/bin:/usr/local/sbin"
CARGO_BINARIES=$HOME/.cargo/bin
PERSONAL_BINARIES=$HOME/bin

# Local binaries
export PATH=$PATH:$HOMEBREW_BINARIES:$CARGO_BINARIES:$PERSONAL_BINARIES:$GO_BINARIES

fpath=(/usr/local/share/zsh-completions $fpath)

autoload -Uz compinit && compinit -u
autoload -Uz colors  && colors

zstyle ':completion:*:default' list-colors ${(s.:.)LS_COLORS}
zstyle ':completion:*:*:*:*:*' menu select

source "$HOME/.dotfiles/zsh-minimal/options.zsh"          # Core settings
eval "$(starship init zsh)"                               # Themeing with starship

zstyle :compinstall filename "$HOME/.zshrc"

bindkey -e
bindkey "^[[1;5C" forward-word  # CTRL right_arrow to move a word forward
bindkey "^[[1;5D" backward-word # CTRL left_arrow to move a word backward
bindkey "[3~" delete-char
bindkey "^[[3~" delete-char
bindkey "^[3;5~" delete-char
bindkey "\e[3~" delete-char

alias vim="nvim"
alias vi="nvim"
alias ls="exa -l"
alias tf="terraform"
alias bex="bundle exec"
alias typora="open -a typora"

# Node env
eval "$(nodenv init -)"

# Ruby
eval "$(rbenv init -)"

# Direnv hooking
eval "$(direnv hook zsh)"

# Kiex
[ -f ~/.kiex/scripts/kiex ] && source "$HOME/.kiex/scripts/kiex"

rg() {
  if [ -t 1 ]; then
    command rg -p "$@" | less -RFX
  else
    command rg "$@"
  fi
}

c() {
  IN_REPO=$(command git rev-parse --is-inside-work-tree 2&> /dev/null)
  if [ "$IN_REPO" = "true" ]; then
    builtin cd -q $(command git root)
  fi
}

set_title() {
    local title=$*
    if [ -n "${PREFERRED_TITLE}" ]; then
      title=$PREFERRED_TITLE
    fi

    if [ -z "$TMUX" ]; then
      echo -ne "\033]0;"$title"\007"
    else
      tmux rename-window $title
    fi
}

title() {
 IN_REPO=$(command git rev-parse --is-inside-work-tree 2&> /dev/null)
  if [ "$IN_REPO" = "true" ]; then
    PATH_IN_GIT=$(command git root)
    set_title $(basename $PATH_IN_GIT)
  else
    set_title $(basename $PWD)
  fi
}

chpwd_functions+=title

title

font_size() {
  sd 'size: [\d\.]+' "size: $1" ~/.config/alacritty/alacritty.yml
}

presentation_mode() {
  font_size 24
}

normal_mode() {
  font_size 16
}

laptop_mode() {
  font_size 14
}

dots() {
  cd ~/.dotfiles
}

watchy() {
  watchexec --clear -- "if $@; then noti -m ✅; else noti -m ❌; fi"
}

unaws() {
  unset AWS_ACCESS_KEY_ID AWS_ACCOUNT_ID AWS_DEFAULT_REGION AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN DP_ENV
}

if [[ -f "$HOME/.custom-client-functions" ]]; then
  source "$HOME/.custom-client-functions";
fi
```

#### Starship Configuration

```toml
format="""
$directory$terraform$env_var$git_branch$git_state$git_status
$jobs$character\
"""

add_newline = true

[character]
success_symbol = "[❯](bold blue)"

[directory]
style = "blue"

[git_branch]
style = "bold yellow"

[git_status]
format = "[($staged)](yellow)[($untracked)](blue)[($deleted)](red)[($modified)](green) "
conflicted = "✘"
untracked  = "●"
modified   = "●"
staged     = "●"
deleted    = "●"

[jobs]
symbol = "!! "
style = "bold red"

[env_var]
variable = "DP_ENV"
style = "fg:6"

[terraform]
symbol = "🪐 "
style = "green"

```

felipesere:
The zsh expansion of `%j` was not working. As such, the symbol and count jobs never showed up in the prompt.
I have not really figured out why it didn't work, even though this feature of ZSH seems well documented.

#### Motivation and Context
Brings back the background jobs
Closes #2505 

Built it locally and ran in a new shell.

#### Checklist:

No documentation or test updates were necessary.


src/init/starship.zsh
```diff
@@ -7,6 +7,15 @@
 # after drawing the prompt. This ensures that the timing for one command is only
 # ever drawn once (for the prompt immediately after it is run).
 
+zmodload zsh/parameter  # Needed to access jobstates variable for NUM_JOBS
+
+starship_render() {
+    # Use length of jobstates array as number of jobs. Expansion fails inside
+    # quotes so we set it here and then use the value later on.
+    NUM_JOBS=$#jobstates
```
davidkna, line 15: How about getting `NUM_JOBS` (or better `STARSHIP_NUM_JOBS`) in `starship_precmd()` like `STARSHIP_CMD_STATUS` and keeping the current approach of only setting `PROMPT` once?

src/init/starship.zsh
```diff
@@ -82,8 +95,3 @@ export STARSHIP_SHELL="zsh"
 STARSHIP_SESSION_KEY="$RANDOM$RANDOM$RANDOM$RANDOM$RANDOM"; # Random generates a number b/w 0 - 32767
 STARSHIP_SESSION_KEY="${STARSHIP_SESSION_KEY}0000000000000000" # Pad it to 16+ chars.
 export STARSHIP_SESSION_KEY=${STARSHIP_SESSION_KEY:0:16}; # Trim to 16-digits if excess.
-
-VIRTUAL_ENV_DISABLE_PROMPT=1
```
davidkna, line 86: I think this needs to be kept either way.

src/init/starship.zsh
felipesere: From a couple of tests, I don't think you can set it just once:
Screenshot 2021-04-25 at 19 33 59

The updated value of `$STARSHIP_NUM_JOBS` is never read again...

vladimyr:
@felipesere Please apply the following patch https://github.com/vladimyr/starship/commit/145d337d.patch

It does what @davidkna [suggested](https://github.com/starship/starship/pull/2638#discussion_r619844002):
> How about getting `NUM_JOBS` (or better `STARSHIP_NUM_JOBS`) in `starship_precmd()` like `STARSHIP_CMD_STATUS` and keeping the current approach of only setting `PROMPT` once?

while ensuring the prompt gets updated.

src/init/starship.zsh
vladimyr: It is definitely possible to set `PROMPT` once and keep it updated by updating the variable holding `jobstates` length. See the comment below for the patch doing exactly that. :arrow_down:

vladimyr:
> The zsh expansion of `%j` was not working. As such, the symbol and count jobs never showed up in the prompt.
> I have not really figured out why it didn't work, even though this feature of ZSH seems well documented.

You can apply this diff to see what is really happening there:
```diff
diff --git a/src/modules/jobs.rs b/src/modules/jobs.rs
index e045ec0f..86804564 100644
--- a/src/modules/jobs.rs
+++ b/src/modules/jobs.rs
@@ -9,10 +9,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let config = JobsConfig::try_load(module.config);
 
     let props = &context.properties;
-    let num_of_jobs = props
+    let num_of_jobs_str = props
         .get("jobs")
-        .unwrap_or(&"0".into())
-        .trim()
+        .map(String::as_str)
+        .unwrap_or("0")
+        .trim();
+
+    eprintln!("jobs: {}", num_of_jobs_str);
+
+    let num_of_jobs = num_of_jobs_str
         .parse::<i64>()
         .ok()?;


```

The problem is in order of operations. Currently, the `PROMPT` setting looks like this:
```zsh
PROMPT='$(::STARSHIP:: prompt --keymap=${KEYMAP} --status=${STARSHIP_CMD_STATUS} --cmd-duration=${STARSHIP_DURATION} --jobs=%j)'
```

What this means is that `$(...)` part gets executed first so `%j` gets sent straight to Starship. Expansion of prompt (`%`) variables happens last and can't be used here because the output of Starship is subject to expansion. It could work in theory if Starship decides to produce a string containing `%j` but that is not the case.

Hope this clears things a bit.

felipesere:
Wow! Now that I see it, it feels obvious. Thank you for also posting the rust snippet that makes the issue visible.

src/init/starship.zsh
```diff
@@ -63,12 +69,12 @@ starship_zle-keymap-select() {
 local existing_keymap_select_fn=$widgets[zle-keymap-select];
 # zle-keymap-select is a special widget so it'll be "user:fnName" or nothing. Let's get fnName only.
 existing_keymap_select_fn=${existing_keymap_select_fn//user:};
-if [[ -z ${existing_keymap_select_fn} ]]; then
+if [[ -z $existing_keymap_select_fn ]]; then
```
jankatins, line 72: Is it really needed to remove the braces (here and elsewhere)? At least I consider using braces for shell vars (`${VAR}` instead of ~`$VAR`~) a best practise.

src/init/starship.zsh
```diff
@@ -86,4 +92,4 @@ export STARSHIP_SESSION_KEY=${STARSHIP_SESSION_KEY:0:16}; # Trim to 16-digits if
 VIRTUAL_ENV_DISABLE_PROMPT=1
 
 setopt prompt{percent,subst}
```
davidkna, line 94: `promptpercent` should no longer be needed with these changes.

```suggestion
setopt promptsubst
```

src/init/starship.zsh
felipesere: I don't have a strong opinion either way. Most variables in that file don't use `$`?

vladimyr: @jankatins I removed those for sake of consistency. You can either always use curlies or only where needed and here was the latter case so I simply followed the suit to make it easier for future maintainers.

Regarding whether the extensive use of curlies is a best practice well, that's certainly a matter of personal preference but I'd argue it is not. Personally, I'm using them only for interpolations and stuff like length calculating, indexing, expansions otherwise I avoid them to minimize the clutter. I pretty much do what Mr. Shellharden _preaches_ https://github.com/anordal/shellharden/blob/master/how_to_do_things_safely_in_bash.md#should-i-use-curly-braces

vladimyr (approved):
LGTM

@felipesere Thanks for taking the time to fix it!

felipesere:
🎉  This was fun! I'll see if there are other interesting issues to pick up

vladimyr:
Glad you liked the ride 😀 Our issue tracker is a never-ending stream of wonders I'm sure you'll find something interesting to work on. 😉

davidkna:
Thanks for the fix @felipesere and thanks for reviewing again @vladimyr.

PR #2649, opened 2021-04-26
feat(shell): add unknown_indicator parameter

### LINKED ISSUE #2541: Allow setting some default value for Shell prompt
Sometimes I use Starship with the Command Prompt on Windows or the `sh` shell on Linux. The Shell prompt doesn't display anything there.

Simple solution: allow setting a `default` option in the config file.

_None_

#### Additional context
I see that the Shell prompt is also missing the `style` option. That can be added too!

rashil2000:
This adds a default parameter for the shell module, which is displayed in case Starship is not able to identify the shell.

#### Description
Fairly simple addition of a 'unknown-indicator' element in the ShellConfig struct. 
Please note that I'm a complete beginner in Rust, so if there's any mistakes, I'll be glad to rectify them 😅

#### Motivation and Context

Closes #2541 

Running `cargo test` passes all existing tests. Let me know if I should add one more test case (like `test_default_if_unknown_shell`).

#### Checklist:

### DIFF
--- a/src/configs/shell.rs
+++ b/src/configs/shell.rs
@@ -13,6 +13,7 @@ pub struct ShellConfig<'a> {
     pub ion_indicator: &'a str,
     pub elvish_indicator: &'a str,
     pub tcsh_indicator: &'a str,
+    pub default: &'a str,
     pub disabled: bool,
 }
 
@@ -27,6 +28,7 @@ impl<'a> Default for ShellConfig<'a> {
             ion_indicator: "ion",
             elvish_indicator: "esh",
             tcsh_indicator: "tsh",
+            default: "",
             disabled: true,
         }
     }
--- a/src/configs/shell.rs
+++ b/src/configs/shell.rs
@@ -13,7 +13,7 @@ pub struct ShellConfig<'a> {
     pub ion_indicator: &'a str,
     pub elvish_indicator: &'a str,
     pub tcsh_indicator: &'a str,
-    pub default: &'a str,
+    pub unknown_indicator: &'a str,
     pub disabled: bool,
 }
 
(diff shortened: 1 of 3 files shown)


src/configs/shell.rs
```diff
@@ -13,6 +13,7 @@ pub struct ShellConfig<'a> {
     pub ion_indicator: &'a str,
     pub elvish_indicator: &'a str,
     pub tcsh_indicator: &'a str,
+    pub default: &'a str,
```
andytom, line 16: Personally I would prefer to use a more specific name for this option maybe `unknown_indicator`?

rashil2000: Sure, that certainly makes more sense! I'll change it.

src/configs/shell.rs
```diff
@@ -27,6 +28,7 @@ impl<'a> Default for ShellConfig<'a> {
             ion_indicator: "ion",
             elvish_indicator: "esh",
             tcsh_indicator: "tsh",
+            unknown_indicator: "",
```
vladimyr, line 31: Does it make sense to use the question mark here as default? :thinking:

rashil2000: Leaving it empty would be, in my opinion, slightly less jarring than a question mark.

vladimyr: I guess you are right, especially because this is a new option.

vladimyr (approved):
LGTM

rashil2000:
Not sure why the test is failing, I haven't touched the `src/configs/mod.rs` file. Do I need to do something?

vladimyr:
Oh that's not your fault, can you please merge latest master?

andytom (approved):

andytom:
Thank you for your contribution @rashil2000 and thank you reviewing @vladimyr

PR #2653, opened 2021-04-26
feat(character): Add option for different symbol when root user [Draft]

### LINKED ISSUE #2421: Ability to set different character module symbol for normal and root user
Would be nice to have this feature to reflect more clear difference of prompt that you're in. For example when user is root then set `#` symbol as character and when is not then something else. I don't think that there are needed any default values I guess it makes sense for this feature to be disabled by default and override current behavior by user's explicit choice.

DavidSmith166:
#### Description

Adds ability to set different character module symbol for normal and root user as described in #2421 

I'm still very much learning Rust so I don't know if my approach follows Starship's design philosophy. Any and all feedback would be greatly appreciated.

- Added root_symbol option in character config
- Added fn is_root_euid which determines whether the user's euid is 0 (root)
- Edited existing tests such that their behavior does not change with my changes

#### Motivation and Context

Closes #2421 

starship_demo

#### Checklist:

### DIFF
--- a/src/configs/character.rs
+++ b/src/configs/character.rs
@@ -9,19 +9,22 @@ pub struct CharacterConfig<'a> {
     pub success_symbol: &'a str,
     pub error_symbol: &'a str,
     pub vicmd_symbol: &'a str,
-    pub root_symbol: &'a str,
+    pub root_success_symbol: &'a str,
+    pub root_error_symbol: &'a str,
+    pub root_symbol_enabled: bool,
     pub disabled: bool,
 }
 
 impl<'a> Default for CharacterConfig<'a> {
     fn default() -> Self {
-        const DEFAULT_SUCCESS_SYMBOL: &str = "[❯](bold green)";
         CharacterConfig {
             format: "$symbol ",
-            success_symbol: DEFAULT_SUCCESS_SYMBOL,
+            success_symbol: "[❯](bold green)",
             error_symbol: "[❯](bold red)",
             vicmd_symbol: "[❮](bold green)",
-            root_symbol: DEFAULT_SUCCESS_SYMBOL,
+            root_success_symbol: "[#](bold green)",
+            root_error_symbol: "[#](bold red)",
+            root_symbol_enabled: false,
             disabled: false,
         }
     }
--- a/src/configs/character.rs
+++ b/src/configs/character.rs
@@ -9,6 +9,7 @@ pub struct CharacterConfig<'a> {
     pub success_symbol: &'a str,
     pub error_symbol: &'a str,
     pub vicmd_symbol: &'a str,
+    pub root_symbol: &'a str,
     pub disabled: bool,
 }
 
@@ -19,6 +20,7 @@ impl<'a> Default for CharacterConfig<'a> {
             success_symbol: "[❯](bold green)",
             error_symbol: "[❯](bold red)",
             vicmd_symbol: "[❮](bold green)",
+            root_symbol: "[#](bold green)",
             disabled: false,
         }
     }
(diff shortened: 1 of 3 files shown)


src/modules/character.rs
```diff
@@ -69,6 +82,25 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+            }
+        },
+        None => None,
+    };
+
+    match euid {
+        Some(euid) => euid == root_uid,
+        None => false,
+    }
+}
```
davidkna, line 102: You should reuse `is_root_user()` from the `username` module here (`src/modules/username.rs`).

src/modules/character.rs
```diff
@@ -113,6 +145,7 @@ mod test {
                 .config(toml::toml! {
                     [character]
                     success_symbol = "[➜](bold green)"
+                    root_symbol = "[➜](bold green)"
```
daniel-white, line 148: can we have `root_success_symbol` and `root_error_symbol`?

daniel-white: also for convention sake - can we change this to `#`

src/modules/character.rs
DavidSmith166: Thank you for your feedback! I've implemented your suggestion

src/modules/character.rs
DavidSmith166: Good call, how does it look now?

daniel-white:
Is this restricted to Linux only?

src/configs/character.rs
```diff
@@ -19,6 +22,9 @@ impl<'a> Default for CharacterConfig<'a> {
             success_symbol: "[❯](bold green)",
             error_symbol: "[❯](bold red)",
             vicmd_symbol: "[❮](bold green)",
+            root_success_symbol: "[#](bold green)",
+            root_error_symbol: "[#](bold red)",
+            root_symbol_enabled: false,
```
matchai, line 27: I'm a little concerned by the amount of added configuration options something like this adds.

Ideally, we would have `success_style`, `error_style`, `symbol` and `root_symbol`; similarly to how we have `style`, `not_capable_style`, and `symbol` for the Node.js module. The trade-off there being that folks wouldn't be able to change the error symbol. 🤔

I'd be curious to hear the thoughts of other @starship/astronauts about how we could go about handling these cases.

src/modules/character.rs
```diff
@@ -28,14 +32,19 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let keymap = props.get("keymap").unwrap_or(&keymap_default);
     let exit_success = exit_code == "0";
 
+    let assumed_mode = match config.root_symbol_enabled && cfg!(target_os = "linux") && is_root_user() {
```
daniel-white, line 35: does this imply linux only - i'd love to see this on mac

kidonng: macOS's root user is disabled by default, `sudo` gives your **current user** root privileges, so this doesn't make much sense on macs.

https://support.apple.com/guide/directory-utility/about-the-root-user-dirub32398f1/mac

daniel-white: but on macs if i `sudo bash` i get a prompt with a `#`

kidonng: Root privileges ≠root user

davidkna: Even if using the root user is discouraged in macOS, I don't think the prompt should artificially restrict displaying that. The current implementation also seems to detect root even if the user is only acting with root privileges (`sudo bash` ).

Furthermore this check  shouldn't exclude all the other BSDs. `is_root_user()` is already handling different operating systems.

aw1cks:
I would love to see this implemented, is there anything I can do to help with this?

PR #2655, opened 2021-04-27
fix(init): always use lf line endings

### LINKED ISSUE #2449: cygwin zsh generated init script line endings
#### Current Behavior
when starship tries to start I get this in my terminal
```
/proc/self/fd/12:5: command not found: ^M
/proc/self/fd/12:9: command not found: ^M
/proc/self/fd/12:11: command not found: ^M
/proc/self/fd/12:29: parse error near `^M'
```
then the default zsh prompt

#### Expected Behavior
Starship prompt starts successfully

Since I recognize that ^M is the cr in crlf, I figured that something is going wrong and is interpreting ^M as a command, so I ran
```sh
starship init zsh --print-full-init >> starship.sh
```
in my terminal and changed all line endings to lf instead of crlf. I then replaced
```sh
eval "$(starship init zsh)"
```
in my .zshrc with
```sh
source ~/starship.sh
```
and that works, so that is the solution I am currently using

#### Possible Solution
ensure that only lf line endings are generated when generating the script for zsh (and bash as well, I do get very similar errors, and the same solution of replacing the generated script line endings works)

#### Environment
- Starship version: 0.50.0
- zsh version: zsh 5.8 (x86_64-pc-cygwin)
- Operating system: Windows 10.0.18362
- Terminal emulator: alacritty v0.6.0
- Git Commit Hash: 0558b38e
- Branch/Tag: v0.50.0
- Rust Version: rustc 1.49.0 (e1884a8e3 2020-12-29)
- Build Time: 2021-02-02 20:05:59

```sh
#
# ~/.zshrc
#

# set default editor
export EDITOR=nvim

# aliases
alias ls='ls --color=auto'

HISTFILE=~/.zsh_history
HISTSIZE=1000
SAVEHIST=1000
setopt nomatch globdots
unsetopt autocd beep extendedglob notify
bindkey -v

zstyle :compinstall filename '/cygdrive/c/Users/cdiekvos/.zshrc'
zstyle ':completion:*' matcher-list '' '+m:{a-zA-Z}={A-Za-z}' '+r:|[._-]=* r:|=*' '+l:|=* r:|=*'

autoload -Uz compinit colors
compinit
colors

eval "$(starship init zsh)"
```

#### Starship Configuration

```toml
add_newline = false

# Make prompt a single line instead of two lines
[line_break]
disabled = true
```

davidkna:
#### Description

This change adds a `.gitattributes` file that ensures all text files are checked out with lf line endings, to improve combability with linux executables on windows.

This does not seem to break the old PowerShell 5.1 or PowerShell 7.

#### Motivation and Context

Closes #2449

#### Checklist:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
-* text=auto eol=lf
+/src/init/starship.* text eol=lf
+/src/init/mod.rs text eol=lf
+*.sh text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,2 @@
-/src/init/starship.* text eol=lf
-/src/init/mod.rs text eol=lf
+/src/init/ text eol=lf
 *.sh text eol=lf


andytom (approved):

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
vladimyr, line 1: Let's make it specific instead?
```suggestion
/src/init/starship.* text eol=lf
```

davidkna: Thanks for the suggestion. I also added `*.sh` for any other shell scripts and the `mod.rs` file (the bash init has multiple lines).

vladimyr: > and the `mod.rs` file (the bash init has multiple lines).

Isn't this ensured by rustc itself? :thinking:

davidkna: Seems not:
```
❯ cargo run -- init bash | coreutils cat -A
M-oM-;M-?^M$
            __main() {^M$
                local major="${BASH_VERSINFO[0]}"^M$
                local minor="${BASH_VERSINFO[1]}"^M$
^M$
                if ((major > 4)) || { ((major == 4)) && ((minor >= 1)); }; then^M$
                    source <("C:\Users\david\Documents\GitHub\starshippppp\target\debug\starship.exe" init bash --print-full-init)^M$
                else^M$
                    source /dev/stdin <<<"$("C:\Users\david\Documents\GitHub\starshippppp\target\debug\starship.exe" init bash --print-full-init)"^M$
                fi^M$
            }^M$
            __main^M$
            unset -f __main^M$
            ^M$
```

vladimyr: Hm, I guess I was too optimistic 🤷

.gitattributes
```diff
@@ -0,0 +1,3 @@
+/src/init/starship.* text eol=lf
+/src/init/mod.rs text eol=lf
```
vladimyr, line 2: Init is basically mod plus a bunch of scripts so I guess we can fold it back to:

```suggestion
/src/init/ text eol=lf
```

Obviously, this is nitpicking do as you think it's the best.

vladimyr (approved):
LGTM

andytom:
Thanks everyone

PR #2658, opened 2021-04-28
refactor(tcsh): set `prompt` once

vladimyr:
#### Description

Refactor `tcsh` init script to set `prompt` variable only once.

#### Motivation and Context

Make init script more readable by utilizing multi-line aliases that resemble functions and undo changes made by #2429 (I was wrong, quoting is in fact required)

/cc @vivekmalneedi

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ shown below. Can't see yours? Have a look at the [extra platform instructions](h
    ```sh
    # ~/.tcshrc
 
-   eval `starship init tcsh`
+   eval "`starship init tcsh`"
    ```
 
 ## 🤝 Contributing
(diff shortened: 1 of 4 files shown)


vivekmalneedi:
I don't think tcsh supports command substitutions in the prompt variable.
Also the eval command errors out when it sees a `#`, so I don't think you can put comments before code in `starship.tcsh`. In an earlier version I put some comments at the bottom of the script and that worked fine since there's no error message when eval fails anyway.

Additionally the if condition uses the `-1` check to fix a [bug](https://github.com/starship/starship/pull/2429/commits/9d968d0e06580e0c4f04f6190b9fd5306774bae4). Regardless of whether the condition is true or false tcsh still tried to do the variable substitution on `STARSHIP_START_TIME` which caused an error when it was unset. I'm not sure if switching to a multiline if changes how tcsh interprets this.

As a final comment, I attempted multiline aliases when I wrote `starship.tcsh` originally but I didn't manage to get it to work.

PR #2662, opened 2021-04-28
perf: skip wrapping ansi escape for ignored shells

davidkna:
#### Description

There is no need to shells that don't have their ANSI escape codes wrapped to rebuild the ansi string.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -246,6 +246,10 @@ pub fn wrap_seq_for_shell(
     escape_begin: char,
     escape_end: char,
 ) -> String {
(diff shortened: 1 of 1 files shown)


andytom (approved):

vladimyr:
Am I missing something or is this already handled by https://github.com/starship/starship/blob/f59e85f5378e980bf646cca837cdc723eef1e969/src/module.rs#L135-L154? :thinking:

davidkna:
I guess you're right

PR #2670, opened 2021-04-30
refactor(print): introduce `UnicodeWidthGraphemes` trait

vladimyr:
#### Description

This introduces a new `Grapheme` type (wrapper for `str`) together with `UnicodeWidthGraphemes` trait that allows calculating grapheme aware widths of strings and graphemes.

#### Motivation and Context

I wanted to abstract away grapheme aware width calculations behind the simple `.width_grapheme()` interface. In order to do that completely, this should probably get moved elsewhere, any suggestions? :thinking: 

#### Checklist:

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -16,6 +16,43 @@ use crate::module::ALL_MODULES;
 use crate::modules;
 use crate::segment::Segment;
 
+pub struct Grapheme<'a>(&'a str);
+
+impl<'a> Grapheme<'a> {
+    pub fn width(&self) -> usize {
+        self.0
+            .chars()
+            .filter_map(UnicodeWidthChar::width)
+            .max()
+            .unwrap_or(0)
+    }
+}
+
+pub trait UnicodeWidthGraphemes {
+    fn width_graphemes(&self) -> usize;
+}
+
+impl<T> UnicodeWidthGraphemes for T
+where
+    T: AsRef<str>,
+{
+    fn width_graphemes(&self) -> usize {
+        self.as_ref()
+            .graphemes(true)
+            .map(Grapheme)
+            .map(|g| g.width())
+            .sum()
+    }
+}
+
+#[test]
+fn test_grapheme_aware_width() {
+    // UnicodeWidthStr::width would return 8
+    assert_eq!(2, "👩‍👩‍👦‍👦".width_graphemes());
+    assert_eq!(1, "Ü".width_graphemes());
+    assert_eq!(11, "normal text".width_graphemes());
+}
+
 pub fn prompt(args: ArgMatches) {
     let context = Context::new(args);
     let stdout = io::stdout();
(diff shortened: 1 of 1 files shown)


src/print.rs
```diff
@@ -16,6 +16,44 @@ use crate::module::ALL_MODULES;
...
+
+pub trait UnicodeWidthGraphemes {
+    fn width_graphemes(&self, is_extended: bool) -> usize;
+}
+
+// Assume that graphemes have width of the first character in the grapheme
+impl<'a> UnicodeWidthGraphemes for Grapheme<'a> {
+    #[inline]
+    fn width_graphemes(&self, _is_extended: bool) -> usize {
+        self.0.chars().next().and_then(|c| c.width()).unwrap_or(0)
```
davidkna, line 29: I think might `max` be slightly more correct:
```suggestion
        self.0.chars().max(|c| c.width()).unwrap_or(0)
```

src/print.rs
```diff
@@ -16,6 +16,43 @@ use crate::module::ALL_MODULES;
...
+    #[inline]
+    fn width_graphemes(&self, is_extended: bool) -> usize {
+        self.graphemes(is_extended)
+            .map(Grapheme)
+            .map(|g| g.width())
+            .sum()
+    }
+}
+
+impl UnicodeWidthGraphemes for String {
```
davidkna, line 41: If you implement `UnicodeWidthGraphemes` for `AsRef<str>` instead, you should be able to get rid of the extra `impl` here.

davidkna (commented):
Overall approach LGTM!

src/print.rs
vladimyr: I was looking for that! Thank you!

src/print.rs
```diff
@@ -16,6 +16,43 @@ use crate::module::ALL_MODULES;
 use crate::modules;
 use crate::segment::Segment;
 
+pub struct Grapheme<'a>(&'a str);
+
+impl<'a> Grapheme<'a> {
+    pub fn width(&self) -> usize {
+        self.0.chars().flat_map(|c| c.width()).max().unwrap_or(0)
```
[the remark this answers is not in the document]
vladimyr: Indeed :+1:

src/print.rs
```diff
@@ -16,6 +16,43 @@ use crate::module::ALL_MODULES;
...
+            .map(Grapheme)
+            .map(|g| g.width())
+            .sum()
+    }
+}
+
+impl UnicodeWidthGraphemes for String {
+    #[inline]
+    fn width_graphemes(&self, is_extended: bool) -> usize {
+        self.as_str().width_graphemes(is_extended)
```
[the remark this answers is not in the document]
vladimyr: I wanted to leave it configurable but you are probably right, we only use extended anyway.

vladimyr:
FWIW I'm still feelin' meh about keeping it inside the print module :shrug:

davidkna (approved):

davidkna:
`utils.rs` maybe? But it doesn't feel much better.

vladimyr:
> `utils.rs` maybe? But it doesn't feel much better.

You've said it all 🤷 Let's leave it as is and move it later on when/if someone comes with a better idea.

andytom:
Thank you as always @vladimyr and thanks for reviewing @davidkna.

PR #2676, opened 2021-05-01
ci: publish on successful release

matchai:
#### Description

We've had some cases where the release process failed mid-way, but the version was still rolled out to Brew and Cargo.
This updates the release process to depend on a successful build before being published on those repos.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,48 +5,6 @@ on:
       - "v*"
 
 jobs:
-  # Publish starship to Crates.io
-  cargo_publish:
-    name: Publish Cargo Package
-    runs-on: ubuntu-latest
-    steps:
-      - name: Setup | Checkout
-        uses: actions/checkout@v2
-
-      - name: Setup | Rust
-        uses: actions-rs/toolchain@v1
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -188,3 +146,47 @@ jobs:
           body_path: RELEASE.md
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  # Publish starship to Crates.io
+  cargo_publish:
```
matchai, line 151: The jobs were reordered, with the only other change being made is that they all now have `needs: github_release`.

vladimyr (approved):
Hmm, with cargo registry this is a clear win but homebrew builds things on their own so you don't get much by requiring a successful build (which is a prerequisite of a successful Github release). At best you'll gonna prevent that one PR that's going to fail on their end anyway. I mean it is not an objection just an observation :shrug:

This is definitely a step in the right direction but we need to ensure that both test suite and build go green before publishing releases. If I'm reading our manifests right, currently there is a chance (unlikely though) of distributing the release that compiles successfully but fails to pass all our tests?

All in all, LGTM

PR #2677, opened 2021-05-01
docs: fix raw string format

jonathf:
#### Description
The docs list `v{raw}` as the format for version number.
This doesn't work, and after a little bit of experimentation, I found the right string: `"v${raw}"`

It also seems a table in the Perl section has some superfluous characters, messing it up in the output.

#### Motivation and Context

I use way to long to figure this out. I assume other have made the same mistake. Lets avoid doing it again.

I've only tested it for Python and CMake, as that is what I use.
I assume that the interpolation here is the same in all instances, but I will not be testing them all.

I updated my config with the updated string format, and it work for raw, minor, major, patch.

#### Checklist:


andytom (approved):

andytom:
Good catch @jonathf thank you for taking the time to fix this.

PR #2679, opened 2021-05-02
refactor(init): tidy up init scripts

vladimyr:
#### Description

Namespace all shell functions and variables under `__starship` namespace, use idiomatic shell language forms, and optimize performance by leveraging shell builtins where applicable.

#### Motivation and Context

Minimize env pollution, standardize variable/function names and optimize init speed.

#### Checklist:

### DIFF
--- a/src/init/starship.zsh
+++ b/src/init/starship.zsh
@@ -3,93 +3,90 @@
 # can cause timing issues, as a user who presses "ENTER" without running a command
 # will see the time to the start of the last command, which may be very large.
 
-# To fix this, we create STARSHIP_START_TIME upon preexec() firing, and destroy it
+# To fix this, we create STARSHIP_CMD_START_TIME upon preexec() firing, and destroy it
 # after drawing the prompt. This ensures that the timing for one command is only
 # ever drawn once (for the prompt immediately after it is run).
 
-zmodload zsh/parameter  # Needed to access jobstates variable for STARSHIP_JOBS_COUNT
+autoload is-at-least
+zmodload zsh/parameter # Needed to access jobstates variable for STARSHIP_JOBS_COUNT
 
-# Defines a function `__starship_get_time` that sets the time since epoch in millis in STARSHIP_CAPTURED_TIME.
-if [[ $ZSH_VERSION == ([1-4]*) ]]; then
-    # ZSH <= 5; Does not have a built-in variable so we will rely on Starship's inbuilt time function.
-    __starship_get_time() {
-        STARSHIP_CAPTURED_TIME=$(::STARSHIP:: time)
-    }
-else
+if is-at-least 4.3.13; then
     zmodload zsh/datetime
     zmodload zsh/mathfunc
-    __starship_get_time() {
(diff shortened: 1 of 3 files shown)


Milo123459 (approved):
LGTM

src/init/starship.zsh
```diff
@@ -3,93 +3,90 @@
...
-}
-
-## Check for existing keymap-select widget.
-local existing_keymap_select_fn=$widgets[zle-keymap-select];
-# zle-keymap-select is a special widget so it'll be "user:fnName" or nothing. Let's get fnName only.
-existing_keymap_select_fn=${existing_keymap_select_fn//user:};
-if [[ -z $existing_keymap_select_fn ]]; then
-    zle -N zle-keymap-select starship_zle-keymap-select;
+# zle-keymap-select is a special widget named like "user:handler-name"
+zle_keymap_select=${${widgets[zle-keymap-select]}//user:}
```
MunifTanjim, line 60: consider prefixing it too: `__starship_preserved_zle_keymap_select`

vladimyr: Well, this is very much a work in progress. I haven't even pushed all my local commits. FWIW, my latest iteration looks like this:
```sh
__zle_keymap_select=${widgets[zle-keymap-select]#user:}
```

I didn't choose `__starship` prefix because it has been used exclusively for things **added** by Starship itself.

MunifTanjim: The reason recommended it because, `__zle_keymap_select` is too generic. Another tool can try to do the same thing and use the same variable name, and that could mess with starship's intialization.

But prefixing it with `__starship` brings down that possibility to almost zero.

vladimyr: That's a good point! Honestly, I'm no fan of verbose identifiers but you are right it is better to be safe than sorry.

src/init/starship.zsh
```diff
@@ -3,93 +3,90 @@
...
 # If starship precmd/preexec functions are already hooked, don't double-hook them
 # to avoid unnecessary performance degradation in nested shells
-if [[ -z ${precmd_functions[(re)starship_precmd]} ]]; then
-    precmd_functions+=(starship_precmd)
-fi
-if [[ -z ${preexec_function[(re)starship_preexec]} ]]; then
-    preexec_functions+=(starship_preexec)
-fi
+(( ! ${precmd_functions[(Ie)__starship_precmd]} )) && precmd_functions+=(__starship_precmd)
+(( ! ${preexec_functions[(Ie)__starship_preexec]} )) && preexec_functions+=(__starship_preexec)
```
CoelacanthusHex, line 56: Maybe it can use [`add-zsh-hook` module](https://zsh.sourceforge.io/Doc/Release/User-Contributions.html#Manipulating-Hook-Functions) of zsh, this method is also recommended by [zsh official theme document](https://zsh.sourceforge.io/Doc/Release/User-Contributions.html#Writing-Themes). 
Using this method is good for implementation of https://github.com/starship/starship/issues/1994 as well.

PR #2680, opened 2021-05-02
feat: Network namespace module for Linux

rustn00b:
This PR introduces a module that displays the current network namespace (for Linux)

#### Description
The module attempts to identify if the shell (or more precisely the starship process) is running inside a non-root network
namespace. When this is found to be the case, the module displays a name for this namespace.

The name is determined as described below:
1. If the namespace has an explicit name (e.g. created using the `ip netns add` command), then this name is used. The name
    is derived from the name of mountpoint binding of the namespace. Since it is possible to bind the namespace multiple times,
    it can effectively have several "names". The module only returns the first one encountered. 
2. When no explicit name exists, the module walks up the process tree in order to determine if on of the process' ancestors
    is running under a different namespace, this is indicative of that the process is not running under the root namespace (in the
    latter case no namespace transitions are expected to be encountered). If a transition is encountered a name is generated
    for the namespace (hex-encoded version of its numeric ID).
    Arguably, it would be more accurate to walk up the entire tree to the root instead of stopping at the first transition.
    The aim here was to make this as fast as possible - but this could be revisited.
3. When it isn't possible to walk up the entire tree (e.g. when the parent process belongs to a different user), the namespace
    is assumed to be the root one, so nothing is displayed by the module. Leaning towards being quiet instead of flooding the
    user with data he/she may not care about.

This module is Linux specific and won't be compiled in on other platforms.

#### Motivation and Context
When working with network namespaces it is fairly common to setup complex topologies that can be tricky to keep track
of when you are running a set of shells in each of the involved namespaces. Keeping tabs on these namespaces
most often involves repetitive calls to `ip net id` which you are bound to forget at some (crucial...) point and 
end up confused as a result.

My solution to this problem was using something similar to the module presented in this PR but as a Zsh RPROMPT. 
This module however is significantly faster.

Testing of the module is through daily use and verification of the situation it is functionally intended to recognize.
1. Create a named network namespace and run `starship module linux_netns` to confirm output
    1a. starship executed as root
    1b. starship executed as non-root user
2. Execute a shell in an "unnamed" network namespace. Done by running a podman container and then executing
    starship in its network namsepace using `nsenter -n`
    2a. again this is done as root
    2b. and as a non-root user (in this case, the process tree cannot be walked so no namespace is returned by the module) 
3. Confirm that the module produces no output in "normal" (non-network-namespaced operation)

As seen elsewhere, I ran `hyperfine` on the module and it comes out at 0.9±0.3 ms (T470 laptop) which seems in line with (if not less than) the other modules on my machine. 

I'm not quite sure how this could be unit-tested without a framework to create (unshared) children with a 
backchannel to inform the parent (testcase) process of the execution result in the target context. It must be pointed out
that I have basically no Rust experience, so I may just be too unaware of the "right" way to do this.

#### Checklist:

I poked around in other modules and tried to document as seen in those, but it's quite likely I stumbled past something
obvious and failed to write up something I should have.

Anyway, thanks for your time and I'm really happy to get your feedback. Don't be too harsh this is my very first foray into
the world of Rust after finishing the Rust book. So I fully expect to see many stupid mistakes pointed out to me!


matchai:
Unfortunately, since this module is specific to an operating system, I'm afraid it can't qualify to be added to Starship. If we can expand it to have cross-platform support, it could be a good candidate.

rustn00b:
Understood, unfortunately I actually have no idea if a similar system exists on Windows and/or MacOS. Should that be the case, there may be interest in providing something akin to this module for these OSs.
I was surprised to see that the bot reported build issues on MacOS ("Some checks were not successful") since I thought I had cordoned off all the Linux-specific namespace and ProcFS bits with a "target_os=linux" guard. I obviously have not understood how to make these work properly.
Anyway many thanks for your work on the project!

matchai:
Closing this PR for the time being. Feel free to reopen it if you manage to make it cross-platform. 😊

PR #2687, opened 2021-05-04
fix: add missing project filter rules

vladimyr:
#### Description

Add missing project filter rules.

#### Motivation and Context

Without necessary wiring respective config sections are ignored. :warning:   

#### Checklist:

### DIFF
--- a/src/modules/deno.rs
+++ b/src/modules/deno.rs
@@ -11,6 +11,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let is_deno_project = context
         .try_begin_scan()?
         .set_files(&config.detect_files)
+        .set_extensions(&config.detect_extensions)
+        .set_folders(&config.detect_folders)
         .is_match();
 
     if !is_deno_project {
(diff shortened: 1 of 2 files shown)


andytom (approved):

andytom:
Good catch @vladimyr, thanks for the quick fix.

Milo123459:
Oh god, these are all modules I added, my bad!

If I may ask, why are we always adding these checks? Or should I just remove the ones that default to nothing and do nothing from my modules I've added?

Milo123459:
Oh and, why don't we just have a universal .scan function that just calls all 3 under-the-hood? Surely this'd clean up code?

andytom:
> Oh god, these are all modules I added, my bad!
> 
> If I may ask, why are we always adding these checks? 

We are adding these to make it possible for uses to configure when the module is triggered. For example I might want to have it so that any directory that contains a file called `some_rust_project` activates the `rust` module. Having these settings allows users to do that.

>Or should I just remove the ones that default to nothing and do nothing from my modules I've added?
 
No these options should be consistent across the all modules.

PR #2696, opened 2021-05-06
feat(directory): Directory substitution styling

Jason-S-Ross:
#### Description

This allows directory substitutions to include style specifiers.

##### Compatibility
This will cause issues with any users that already have a style string in their substitutions table - they will have to escape those style strings if they want the e.g. literal string `[foo](bold)` to appear in their prompt.

##### Implementation
This results in some extensive changes to the implementation of the `directory` module. Once substitutions are applied, we're no longer working with a string, but something that stores style information as well. I defined a `StyleString` type that is a wrapper for a vec of `{ char, Style }` structs in order to simplify things a bit. This styled directory string still needs to be truncated, so I just inlined the implementation of that into `module`.

#### Motivation and Context

This is beneficial for a couple of reasons:
1. This allows the user to change the color of the path separation character, which can help readability.
2. This allows the user to specify special colors for special directories, which can help warn when they are about to do something dangerous.

This was requested in #1837.

Here's a simple example and the result
```toml
[directory.substitutions]
"/" = "[╱](red bold)"
"rust" = "[rust](fg:161)"
```

With Substitution Styling

I ran the Linux test suite. This should only affect the directory module.

#### Checklist:

I have not updated the documentation because I'm not sure if it is necessary for this change.

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -392,7 +392,10 @@ impl StyleString {
         let mut last_end = 0;
         let str_repr = self.to_string();
         for (start, part) in str_repr.match_indices(from) {
-            result.extend_from_slice(&self.chars[last_end..start]);
+            // Each part of the chars vector is a char so we have to convert from bytes to chars
+            let start = str_repr[..start].chars().count();
+            let slice = &self.chars[last_end..start];
+            result.extend_from_slice(slice);
             result.extend_from_slice(to);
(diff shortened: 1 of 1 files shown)


Jason-S-Ross:
I've identified a bug with the root directory not being correctly identified if the path separator is styled. I'm closing this until I can resolve.

PR #2747, opened 2021-05-19
docs: New preset "Bracketed Segments"

zzamboni:
This preset changes all the built-in formats to show their segment in brackets instead of using the default Starship wording ("via", "is", "on", etc.).

Before (default):
image

After:
image


andytom (approved):
LGTM

PR #2749, opened 2021-05-22
docs(preset): New preset "Plain Text Symbol"

### LINKED ISSUE #2550: Add a plain text alternative to Nerd Fonts

As nifty as they are, I see three problems with Nerd Fonts:

1. They do not render correctly when copy/pasted into an environment that uses a non Nerd Font. This also means that using Starship will result in “tofu” when using tools like asciinema.
2. Using Nerd Fonts as an icon set introduces accessibility issues with screen readers.
3. Strange rendering issues like #2546 crop up.

Add a configuration to render Nerd Font icons as text when they’re essential and not at all when they merely augment existing text. So, e.g., instead of this:

```
my-repo on [branch icon] my-branch [!] is [version icon] v0.1.0 via [node icon] v12.19.1
```

The new configuration option would render this:

```
my-repo on my-branch [!] is v0.1.0 via nodejs v12.19.1
```

1. **Add a configuration to just turn off the Nerd Font icons.** This would be simpler and terser, but a bit ambiguous in some contexts where the icons are the only device conveying semantic information.
2. **Add a configuration to use unicode-only characters.** I’m not confident this is a particularly workable solution. It sounds like from the issue #1649 was trying to resolve that not all monospace fonts actually render _all_ their characters (especially the dingbat-ish ones) monospace.

rumia-san:
#### Description
This PR adds a preset "Plain Text Symbol" so that users could use starship with terminal/fonts that doesn't support the NerdFonts/emojis/unicodes.

#### Motivation and Context

Closes #2550

I love bitmap fonts a lot (Fixedsys, Dina, etc.). However the NerdFonts/emojis/unicodes could not be rendered with bitmap font. I found there is a "good first issue" #2550, so I decided to write a preset for this case.

Before (Without this preset):
plain-text-symbols-before

After (With this preset):
plain-text-symbols-after

#### Checklist:


andytom (approved):
LGTM

andytom:
@rumia-san thank you for your contribution.

PR #2751, opened 2021-05-23
feat: Add Operating System condition to custom commands

### LINKED ISSUE #2750: Add Operating System condition to custom commands
I'm trying to create a config that works both on Linux and Windows. I created a custom command that shows if sudo is cached, like so:

```toml
[custom.sudo]
when = "sudo -n true"
style = "bold red"
format = "[!]($style)"
```

But there is no `sudo` on Windows, and this makes the prompt take a long time to show if enabled on Windows.

I'd like to be able to restrict it to Linux only. Something like adding an "SO" constraint. Like so:

```toml
[custom.sudo]
when = "sudo -n true"
style = "bold red"
format = "[!]($style)"
os = "linux" # could be "windows" or "mac", too
```

Or maybe a shell constraint:

```toml
[custom.sudo]
when = "sudo -n true"
style = "bold red"
format = "[!]($style)"
shell = "bash" # could be "pwsh", too
```

Both would work, but the first one is preferred, as I can have bash on Windows and PowerShell on Linux. But I can imagine the shell constraint being useful as well.

Not having a single config file, but two separate ones. It is suboptimal.

jgoday:
#### Description

Tries to implement #2750.
Add optional so to custom configs to allow filter them by Operating System.
Internally, it uses [env::consts::OS](https://doc.rust-lang.org/std/env/consts/constant.OS.html) to match against different Operating Systems.

#### Motivation and Context

Closes #2750.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3012,7 +3012,7 @@ If you have an interesting example not covered there, feel free to share it ther
 | `style`       | `"bold green"`                  | The style for the module.                                                                                                  |
 | `format`      | `"[$symbol($output )]($style)"` | The format for the module.                                                                                                 |
 | `disabled`    | `false`                         | Disables this `custom` module.                      |
-| `os`          |                                 | Operating System name on which the module will be shown. |
+| `os`          |                                 | Operating System name on which the module will be shown (linux, macos, windows, ... ) [See possible values](https://doc.rust-lang.org/std/env/consts/constant.OS.html). |
 
 ### Variables
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3012,7 +3012,7 @@ If you have an interesting example not covered there, feel free to share it ther
 | `style`       | `"bold green"`                  | The style for the module.                                                                                                  |
 | `format`      | `"[$symbol($output )]($style)"` | The format for the module.                                                                                                 |
 | `disabled`    | `false`                         | Disables this `custom` module.                      |
-| `os`          |                                 | Operating System name on which the module will be shown (linux, macos, windows, ... ) [See possible values](https://doc.rust-lang.org/std/env/consts/constant.OS.html). |
+| `os`          |                                 | Operating System name on which the module will be shown (unix, linux, macos, windows, ... ) [See possible values](https://doc.rust-lang.org/std/env/consts/constant.OS.html). |
 
 ### Variables
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2974,6 +2974,7 @@ These modules will be shown if any of the following conditions are met:
 - The current directory contains a directory whose name is in `directories`
 - The current directory contains a file whose extension is in `extensions`
 - The `when` command returns 0
+- The current Operating System (std::env::consts::OS) matchs with `so` field if defined.
 
 ::: tip
 
@@ -3010,7 +3011,8 @@ If you have an interesting example not covered there, feel free to share it ther
 | `symbol`      | `""`                            | The symbol used before displaying the command output.                                                                      |
 | `style`       | `"bold green"`                  | The style for the module.                                                                                                  |
 | `format`      | `"[$symbol($output )]($style)"` | The format for the module.                                                                                                 |
-| `disabled`    | `false`                         | Disables this `custom` module.                                                                                             |
+| `disabled`    | `false`                         | Disables this `custom` module.                      |
+| `so`          |                                 | Operating System name on which the module will be shown. |
 
 ### Variables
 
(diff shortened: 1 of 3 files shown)


giggio:
@jgoday I noticed you used `so`, as I had on the issue description of #2750. `so` is operating system in my native language, portuguese, I believe `os` would be more adequate, as in english. I also fixed the issue description.

jgoday:
> @jgoday I noticed you used `so`, as I had on the issue description of #2750. `so` is operating system in my native language, portuguese, I believe `os` would be more adequate, as in english. I also fixed the issue description.

You are right :), I had not noticed. I have changed the field name and the docs in later commit. Thanks.

docs/config/README.md
```diff
@@ -3010,7 +3011,8 @@ If you have an interesting example not covered there, feel free to share it ther
 | `symbol`      | `""`                            | The symbol used before displaying the command output.                                                                      |
 | `style`       | `"bold green"`                  | The style for the module.                                                                                                  |
 | `format`      | `"[$symbol($output )]($style)"` | The format for the module.                                                                                                 |
-| `disabled`    | `false`                         | Disables this `custom` module.                                                                                             |
+| `disabled`    | `false`                         | Disables this `custom` module.                      |
+| `os`          |                                 | Operating System name on which the module will be shown. |
```
andytom, line 3015: I think it might be worth adding some examples of the possible values here since it is not obvious what a user should use on a Mac e.g osx, macOS or darwin are all reasonable guesses.

src/modules/custom.rs
```diff
@@ -32,6 +33,10 @@ pub fn module<'a>(name: &str, context: &'a Context) -> Option<Module<'a>> {
             is_match = exec_when(when, &config.shell.0);
         }
 
+        if let Some(os) = config.os {
+            is_match = os == env::consts::OS;
```
andytom, line 37: I think this might cause some false positives since if `is_match` is already false e.g. the when command returned false this would override that and always display the module if the os matches.

andytom (changes requested):

src/modules/custom.rs
jgoday: You're right, should be fixed now.

docs/config/README.md
jgoday:
```
(linux, macos, windows, ... ) [See possible values](https://doc.rust-lang.org/std/env/consts/constant.OS.html)
```
Should this be enough in the options description ?

src/modules/custom.rs
```diff
@@ -37,6 +38,12 @@ pub fn module<'a>(name: &str, context: &'a Context) -> Option<Module<'a>> {
         }
     }
 
+    if let Some(os) = config.os {
+        if os != env::consts::OS {
```
davidkna, line 42: I think it would be nice to allow setting `os` to `unix` to exclude windows and only allow unixlike systems like with `cfg!(unix)`.

jgoday: Would the following condition be sufficient ?
```rust
if os != env::consts::OS && !(os == "unix" && env::consts::OS != "windows") {
      return None;
}
```

davidkna: I would prefer `os == "unix" && !cfg!(unix)`.

jgoday: Nice, you're right, changed to 
```rust
if os != env::consts::OS && !(os == "unix" && cfg!(unix)) {
```

docs/config/README.md
```diff
@@ -3010,7 +3011,8 @@ If you have an interesting example not covered there, feel free to share it ther
 | `symbol`      | `""`                            | The symbol used before displaying the command output.                                                                      |
 | `style`       | `"bold green"`                  | The style for the module.                                                                                                  |
 | `format`      | `"[$symbol($output )]($style)"` | The format for the module.                                                                                                 |
-| `disabled`    | `false`                         | Disables this `custom` module.                                                                                             |
+| `disabled`    | `false`                         | Disables this `custom` module.                      |
+| `os`          |                                 | Operating System name on which the module will be shown (linux, macos, windows, ... ) [See possible values](https://doc.rust-lang.org/std/env/consts/constant.OS.html). |
```
andytom, line 3015:
```suggestion
| `os`          |                                 | Operating System name on which the module will be shown (unix, linux, macos, windows, ... ) [See possible values](https://doc.rust-lang.org/std/env/consts/constant.OS.html). |
```

andytom (approved):
Minor suggestion but other than that LGTM

jgoday:
Already changed, thank you!

andytom:
Thank you for your contribution @jgoday

PR #2769, opened 2021-05-28
docs: Added more bracketed segments

zzamboni:
#### Description

A few of the supported segments were missing from the preset, have now
been added.

### DIFF
--- a/docs/presets/README.md
+++ b/docs/presets/README.md
@@ -103,6 +103,9 @@ After:
 ### Configuration
 
 ```toml
+[aws]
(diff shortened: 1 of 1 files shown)


andytom (approved):

davidkna:
Thanks for the contribution!

zzamboni:
Thanks! Any idea when this new preset will be published at https://starship.rs/presets/?

andytom:
> Thanks! Any idea when this new preset will be published at https://starship.rs/presets/?

Hi @zzamboni, we published when we release a new version of starship, this is to avoid confusion about having docs for features that have not been released yet.

PR #2770, opened 2021-05-29
feat(directory): Directory enhancements

### LINKED ISSUE #1837: directory.substitution should allow format strings.
I have a feature request for `[directory.substitution]`!

I would like it if this would work:

```toml
[directory.substitution]
"/" = "[/](dimmed blue)"
```

Instead, it just shows the text verbatim.

### LINKED ISSUE #1840: Replace directory.substitution with a path-based globbing solution

The current `directory.substition` is confusing.

- It doesn't work well with truncation (especially fish-style truncation).
- It only works on literal string replacements.

Instead I propose:

1. Use directory matching instead of string; replace the folders based on actual paths.
1. Allow Ant-style globbing; normal globbing plus `**/blah` for recursion.
1. Require the directory match to always be absolute.
1. Allow controlling the substitution part via a marker of some kind (e.g. `(` and `)`).  These would only work on a whole directory part. e.g., "/et(c)/" would not be allowed.

This would allow the truncation to still happen (it is based on directories, still).

Example config:

```toml
[directory.next_gen_substition]
"/**/(.git)" = ":git:"
"(~/src/github.com)" = ":github:"
"**/(starship)" = "🚀"
"(**/starship/fred)" = "<FRED>"

# "**/star(ship)" = "BAD - invalid ()s
```
 

Example outputs:

```sh
"/home/docwhat/src/github.com/starship/starship" becomes ":github:/🚀/🚀"
"/home/docwhat/src/github.com/starship/starship/.git" becomes ":github:/🚀/🚀/:git:"
"/home/docwhat/src/github.com/starship/starship/fred" becomes "<FRED>"
```

Note that truncation (normal or fish-style) can be done *after* the replacement since the parts are still guaranteed to be path-like.

Not much else that can be done.

### LINKED ISSUE #2452: Intelligent prompt truncation
## Feature Request
As has been discussed a few times, it would be great if Starship could handle its own prompt truncation, instead of offloading that responsibility to the shell. Making use of `$COLUMNS` we could intelligently truncate "less important" information from the prompt, making modules more concise, until they eventually have to be removed.

Currently, on small enough terminal windows, the prompt is either truncated by the shell (#2445), or breaks into the next line.

As the window gets smaller, we can begin by hiding language runtime version numbers, only showing their symbol, we could hide joining words between modules, or we could truncate the PWD directory.

Jason-S-Ross:
#### Description
This pull request provides a number of enhancements to the display of directory strings.

***Full Disclosure: This pull request depends on the [stylish-stringlike](https://crates.io/crates/stylish-stringlike)  crate, of which I am the author.***

##### Directory Fitting
Directories can be shrunk to fit in an available terminal width. For example, the path `some-random-arbitrarily-long-path/with/segments` can be automatically shrunk to `some-ran…ng-path/with/segments`. New configuration variables `fit_width`, `fit_style`, and `fit_symbol` are introduced to control this behavior.

##### Styled Substitution
Style strings in `[directory.substitutions]` are expanded.

##### Regular Expression Substitution
Regular expressions (with back-referencing) are supported through `[directory.substitution_regexes]`, and style strings in the replacement text are expanded. 

##### Path Delimiter Specification
A new configuration variable `delimiter` is provided. This is required in order for the directory fitting to not truncate on path delimiters.

#### Motivation and Context

Closes #1837
Closes #1840
Closes #2452.

Styled substitutions make path delimiters easier to see. Automatic fitting of paths can provide more flexibility if users want as much context as possible without overflowing the terminal, or if they have unfortunately long folder names.

image
image

I have tested this on Linux Mint using the Tilix terminal running zsh. I have also run the test suite on Windows.

I have updated the call signature for the `modules::utils::directory::truncate` function to be generic over objects that are `Splittable` and `Pushable` (which `String` implements), and to require a `delimiter` argument. Corresponding function calls in the `conda` module have been updated for this reason.

#### Checklist:


Jason-S-Ross:
Had the wrong branch checked out when I ran the test suite on windows. Will reopen when Windows works.

lilyball:
Does this work with fish-style truncation? It looks like fish-style truncation is disabled if any substitutions are configured (even if they don't apply to the path). At the very least, I hope fitting paths works with fish-style truncation (e.g. by only fitting the non-truncated portion of the path). Ideally fish-style substitution would work with other substitutions too, though I haven't looked into substitutions (since they're undocumented) so I'm not sure yet as to the normal use-case for them.

lilyball:
Oops substitutions aren't undocumented, they're just hidden behind a disclosure triangle. Still, I don't have any intuition for how they're normally used beyond replacing prefixes (e.g. replacing `/private/etc` with `/etc`). My feeling is substitutions shouldn't have to disable fish-style truncation, though I suppose semantics have to be defined.

For the purposes of this PR, I feel like `[directory.substitution_regexes]` could be used to do things like define custom fitting logic on individual path components, and that really shouldn't break fish-style truncation. More generally, substitutions that occur within the un-truncated portion of a fish-style truncated path shouldn't affect fish-style truncation, and substitutions that replace a single path component with another one shouldn't either.

Jason-S-Ross:
Thanks for the feedback! Looks like I missed a serious bug with how fish-style truncation interacts with regex substitutions. 

For the record, here's what I get when I try to rename a path segment while fish-style truncation is enabled (this is broken):
image
And the same directory with fish-style truncation turned off (looks ok; "sympy" replaced with "ypmys"):
image
And the same directory with fish-style truncation on and regex-replacement off (looks ok):
image

I'll look into this and make any substitutions prior to creating the fish-style truncation string as well.

docwhat (approved):
Lots of good stuff in here.

musjj:
Any updates here? Would love to have this behavior. Long directory names are a pain.

PR #426, opened 2019-09-25
fix: Prevent `rust` module from installing toolchains

qryxip:
#### Description

Prevent `rust` module from installing the toolchain [specified with `$RUSTUP_TOOLCHAIN`, `rustup override set`, or `rustup-toolchain`](https://github.com/rust-lang/rustup.rs/tree/924630ba1162b81c0421e8ce56968d4984c605a4#override-precedence) when it is not installed.

#### Motivation and Context

Closes #417.

#### Types of changes

- Displays ~~error of `rustup run _ rustc --version` (`"toolchain '_' is not installed"`)~~ toolchain names when they are not installed.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Screenshots (if appropriate):
screenshot

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

~~Nothing to update.~~ Added tests but I didn't modify the existing tests.

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #2775, opened 2021-06-01
fix(rust): support for `rust-toolchain.toml`

pineapplehunter:
#### Description

This PR adds support for `rust-toolchain.toml` in the rust module.
A related Issue is #417, and this just adds some lines to PR #426.

#### Motivation and Context

I haven't opened a issue for this, but when I enter a directory with `rust-toolchain.toml` rustup started to install the toolchain. This is the same issue as #417 but with `.toml` file.

#### Checklist:

### DIFF
--- a/src/modules/rust.rs
+++ b/src/modules/rust.rs
@@ -63,7 +63,7 @@ fn get_module_version(context: &Context, config: &RustConfig) -> Option<String> 
     // check
     // 1. `$RUSTUP_TOOLCHAIN`
     // 2. `rustup override list`
-    // 3. `rust-toolchain` in `.` or parent directories
+    // 3. `rust-toolchain` or `rust-toolchain.toml` in `.` or parent directories
     // as `rustup` does.
     // https://github.com/rust-lang/rustup.rs/tree/eb694fcada7becc5d9d160bf7c623abe84f8971d#override-precedence
     //
@@ -163,9 +163,21 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
         }
     }
 
+    if let Ok(true) = context
+        .dir_contents()
+        .map(|dir| dir.has_file("rust-toolchain.toml"))
+    {
+        if let Some(toolchain) = read_channel(Path::new("rust-toolchain.toml")) {
+            return Some(toolchain);
+        }
+    }
+
     let mut dir = &*context.current_dir;
     loop {
         if let Some(toolchain) = read_channel(&dir.join("rust-toolchain")) {
+            return Some(toolchain);
+        }
+        if let Some(toolchain) = read_channel(&dir.join("rust-toolchain.toml")) {
             return Some(toolchain);
         }
         dir = dir.parent()?;
--- a/src/modules/rust.rs
+++ b/src/modules/rust.rs
@@ -157,6 +157,24 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
         .map(|c| c.trim().to_owned())
     }
 
+    if context
+        .dir_contents()
+        .map_or(false, |dir| dir.has_file("rust-toolchain"))
+    {
+        if let Some(toolchain) = read_channel(Path::new("rust-toolchain"), false) {
+            return Some(toolchain);
+        }
+    }
+
+    if context
+        .dir_contents()
+        .map_or(false, |dir| dir.has_file("rust-toolchain.toml"))
+    {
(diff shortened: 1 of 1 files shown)


davidkna:
Thanks for the contribution @pineapplehunter, to get the CI to pass can you either update your commit message to include the prefix `fix(rust):` in the title or just push an empty commit to this branch e.g

```sh
git commit --allow-empty -m 'make Semantic PR bot happy'
git push 
```

pineapplehunter:
@davidkna I've changed the commit message.

davidkna:
This looks good to me, but could you maybe add a test for this new case?

pineapplehunter:
I added some tests!
When making a test a realized that when using the `.toml` extension, you can't use the channel in one line format. So I changed the `read_channel` function so it won't parse in that case.

pineapplehunter:
I had some time, so I added some more tests, and comments.
Found out that the code was checking the current directory twice, so I removed that part.
If the tests pass, I think this PR is done.

src/modules/rust.rs
```diff
@@ -137,12 +140,12 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
...
-    fn read_channel(path: &Path) -> Option<String> {
+    fn read_channel(path: &Path, only_toml: bool) -> Option<String> {
         let contents = fs::read_to_string(path).ok()?;
 
-        match contents.lines().count() {
-            0 => None,
-            1 => Some(contents),
+        match (contents.lines().count(), only_toml) {
+            (0, _) => None,
+            (1, false) => Some(contents),
```
davidkna, line 148: I think something like this would be more clear:
```suggestion
        match contents.lines().count() {
            0 => None,
            1 if !only_toml => Some(contents),
```

pineapplehunter: That looks good!

pineapplehunter:
It seems like clippy is failing on a unrelated to the PR

davidkna:
@pineapplehunter Please rebase on/merge master to make clippy pass again.

pineapplehunter:
I rebased the commits on master.
I think this should work.

pineapplehunter:
The error in the test seems unrelated to this PR, but I don't know.
I don't have a mac to check about this.
Is there something I can do to fix this?

davidkna:
I'm rerunning the tests now. I think this failure was just #886.

src/modules/rust.rs
```diff
@@ -154,18 +157,12 @@ fn find_rust_toolchain_file(context: &Context) -> Option<String> {
...
 
-    if let Ok(true) = context
-        .dir_contents()
-        .map(|dir| dir.has_file("rust-toolchain"))
-    {
-        if let Some(toolchain) = read_channel(Path::new("rust-toolchain")) {
-            return Some(toolchain);
-        }
-    }
-
```
davidkna, line 165: I just looked at the code again and saw that you removed this. Can you add this back again? Unlike the code below this will have the current directory cached in `context`.

pineapplehunter: Got it. I'll fix that part.
But I've got some questions.
1. Should I copy this code for `rust-toolchain.toml` too? I'm gessing if this is for caching so this only needs to be run only once for `rust-toolnain`.
2. Should the loop below looking for parent directories start from a parent? If I add the code back, I think it would try to load the same file in the current directory twice.

davidkna: Please use this for the toml file too. You can also simplify it to something likes this:

```rust
if context
        .dir_contents()
        .map_or(false, |dir| dir.has_file("rust-toolchain"))
```

I think it's fine to have it start from the parent (though that might end up causing issues with case-insensitive file systems).

pineapplehunter: Thanks for the quick response!
I'll use that code.
I found out that starting from the parent in the loop causes a error in the test. It seems that
 
```rust
read_channel(Path::new("rust-toolchain"), false) // returns None, path = "rust-toolchain"
```

and

```rust
let mut dir = &*context.current_dir;
read_channel(&dir.join("rust-toolchain"), false) // returns Some, path = "/tmp/**something**/rust-toolchain"
```

give different outputs. So I'll keep the loop as is.

pineapplehunter:
Updated the code.
What do you think?

davidkna (approved):

davidkna:
Thanks for the contribution @pineapplehunter!

PR #2782, opened 2021-06-04
feat(docker_context): Use DOCKER_HOST and DOCKER_CONTEXT enviroment variables 

### LINKED ISSUE #2022: DOCKER_HOST should also be used for docker_context
#### Current Behavior

Per https://docs.docker.com/engine/context/working-with-contexts/#the-anatomy-of-a-context the Docker Context can also be overriden by the `DOCKER_HOST` environment variable.

#### Expected Behavior

`docker_context` should appear when `DOCKER_HOST` is set to a non-empty value.

The default is actually `Current DOCKER_HOST based configuration`
```none
❯ docker context ls
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                     KUBERNETES ENDPOINT   ORCHESTRATOR
default *           moby                Current DOCKER_HOST based configuration   ssh://trajano@54.224.104.16                               swarm
```

#### Possible Solution

on Windows if `DOCKER_HOST != ''` then show the context but do not use the name instead show the endpoint value.

For now this `custom.docker_context` I use
```toml
[custom.docker_context]
description = "Shows the docker symbol if the current directory has Dockerfile or docker-compose.yml files"
when="[ $DOCKER_HOST ]"
command = "echo -n $DOCKER_HOST"
shell="sh"
symbol = "🐳 "
style = "blue bold"
format= "via [$symbol$output]($style)"
```

#### Environment
- Starship version: 0.47.0
- Shell type: bash
- Shell plugin manager: [e.g. oh-my-fish, oh-my-zsh, fisher, antigen]

```bash
 export  DOCKER_HOST=ssh://trajano@54.224.104.16
```

#### Starship Configuration

```bash
[docker_context]
disabled=false
only_with_files=false
```

yamabiiko:
#### Description
As described in #2022 and the Docker official documentation, the Docker Context can also be overriden by the `DOCKER_HOST` and `DOCKER_CONTEXT` environment variables. This PR detects the enviroment variables and updates the docker_context module accordingly.

#### Motivation and Context
The main motivation is to follow how the docker context works when those enviroments are set, accordingly to expected behavior.
Closes #2022 

We wrote some additional tests for the modules to ensure the correct behavior of the module.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -791,7 +791,8 @@ truncation_symbol = "…/"
 
 The `docker_context` module shows the currently active
 [Docker context](https://docs.docker.com/engine/context/working-with-contexts/) if it's not set to
-`default`.
+`default` or if the `DOCKER_HOST` or `DOCKER_CONTEXT` environment variables are set (as they are meant
+to override the context in use).
 
 ### Options
 
--- a/src/modules/docker_context.rs
+++ b/src/modules/docker_context.rs
@@ -9,10 +9,16 @@ use crate::utils;
 /// Creates a module with the currently active Docker context
 ///
 /// Will display the Docker context if the following criteria are met:
-///     - There is a file named `$HOME/.docker/config.json`
+///     - There is a non-empty enviroment variable named DOCKER_HOST
+///     - Or there is a non-empty enviroment variable named DOCKER_CONTEXT
+///     - Or there is a file named `$HOME/.docker/config.json`
 ///     - Or a file named `$DOCKER_CONFIG/config.json`
 ///     - The file is JSON and contains a field named `currentContext`
 ///     - The value of `currentContext` is not `default`
+///     - If multiple criterias are met, we use the following order to define the docker context:
+///     - DOCKER_HOST, DOCKER_CONTEXT, $HOME/.docker/config.json, $DOCKER_CONFIG/config.json
+///     - (This is the same order docker follows, as DOCKER_HOST and DOCKER_CONTEXT override the
+///     config)
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("docker_context");
     let config: DockerContextConfig = DockerContextConfig::try_load(module.config);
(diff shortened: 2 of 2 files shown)


yamabiiko:
Changed the code according to Rust lint and Cargo fmt tool (squashed and force-pushed)

davidkna:
Can you add the `feat(docker_context): ` prefix to your commit or add another empty commit to make the sematic pull request check pass?

src/modules/docker_context.rs
```diff
@@ -35,50 +42,57 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+            let parsed_json = serde_json::from_str(&json).ok()?;
+
+            match parsed_json {
+                serde_json::Value::Object(root) => {
+                    let current_context = root.get("currentContext")?;
+                    match current_context {
+                        serde_json::Value::String(ctx_json) => OsString::from(ctx_json),
+                        _ => return None,
+                    }
                 }
```
davidkna, line 65: I think putting this into an `OsString` should be avoided.
I think it would be better to just use the non-`_os` version for getting the env var above or just call `.to_string_lossy()` and `.into_owned()` directly above.

And since you're touching this: You can simplify these lines here to `parsed_json.get("currentContext")?.as_str().map(str::to_owned)`

src/modules/docker_context.rs
```diff
@@ -35,50 +42,57 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     )
     .join("config.json");
 
-    if !docker_config.exists() {
-        return None;
-    }
+    let docker_context_env = context
+        .get_env_os("DOCKER_HOST")
+        .or_else(|| context.get_env_os("DOCKER_CONTEXT"));
```
davidkna, line 47: I think something like this would work better here
```rust
std::array::IntoIter::new(["DOCKER_HOST", "DOCKER_CONTEXT"]).find_map(context.get_env_os)
```

davidkna (changes requested):

davidkna (approved):

docs/config/README.md
```diff
@@ -791,7 +791,8 @@ truncation_symbol = "…/"
 
 The `docker_context` module shows the currently active
 [Docker context](https://docs.docker.com/engine/context/working-with-contexts/) if it's not set to
-`default`.
+`default` or if the `DOCKER_HOST` or `DOCKER_CONTEXT` environment variables are set (as they are meant
+to override the context in use according to the docker documentation).
```
davidkna, line 795:
```suggestion
`default` or if the `DOCKER_HOST` or `DOCKER_CONTEXT` environment variables are set (as they are meant
to override the context in use).
```

davidkna (approved):
Sorry for only getting back to this now.
I've left a small suggestion, otherwise LGTM.

yamabiiko:
Updated the documentation, everything should be ok

andytom:
Thank you for your contribution @yamabiiko

PR #2783, opened 2021-06-04
fix(init): Use shorter name for STARSHIP_SHELL PowerShell variable

rashil2000:
#### Description
This renames the environment variable `$Env:STARSHIP_SHELL` from 'powershell' to 'pwsh'.

#### Motivation and Context

The string 'powershell' is just too long and does not provide any extra information. 
Also, Starship doesn't have the ability to right align segments, so left prompt space is a premium. 

This also makes it uniform with respect to the `shell` module - both the `shell` module and STARSHIP_SHELL variable will now say 'pwsh'

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -247,7 +247,7 @@ impl<'a> Context<'a> {
             "bash" => Shell::Bash,
             "fish" => Shell::Fish,
             "ion" => Shell::Ion,
-            "powershell" => Shell::PowerShell,
+            "pwsh" => Shell::PowerShell,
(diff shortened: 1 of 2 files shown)


rashil2000:
Hi, 
Is there any update on this? Am I missing something?

davidkna:
You might have missed a few places that need to be changed to pwsh in the rust code.
Beyond that, because this is something of a breaking change for custom scripts for little benefit, I'd prefer not merging this until either the v1 or v2 bump.

PR #2785, opened 2021-06-05
fix(dotnet): Don't treat folder with .sln file as a .NET project

### LINKED ISSUE #2567: Dotnet module shows in a C/C++ repository
#### Current Behavior
The .NET module shows up when in a repository cloned from https://github.com/Parchive/par2cmdline, which appears to be a pure C/C++ project

#### Expected Behavior
.NET module does not show up when no .NET languages present.

image

#### Possible Solution
.NET will currently trigger on seeing a `.sln` file, which is probably what's triggering it. However, the `.sln` file
in this directory appears to only be used to define C++ builds for Visual Studio.

We could remove the `.sln` file trigger, or we could display the `.NET` module only if a `.sln` file exists AND there are no `.vcxproj` files. I don't understand the VS build systems well enough to comment, but is it possible to have a `.sln` file without a `.<lang>proj` file?

#### Environment
- Shell type: bash
- Shell version: GNU bash, version 4.3.42(1)-release (x86_64-suse-linux-gnu)
- Shell plugin manager: None
- Operating system: Some Linux distro (remote machine, lsb_release missing)

Can reproduce config if needed, but I don't think it's relevant.

#### Starship Configuration

```bash
[git_status]
disabled=true
```

rumia-san:
#### Description
Change the detect_extensions of the .NET module, so that the `.sln` file will not trigger the .NET prompt.
According to the [VS documentation](https://docs.microsoft.com/en-us/visualstudio/ide/solutions-and-projects-in-visual-studio?view=vs-2019),the `.sln` file is 

> simply **a container for one or more related projects**, along with build information, Visual Studio window settings, and any miscellaneous files that **aren't associated with a particular project**.

The project might be C++ project or something else. **Currently, the .NET module will treat all folders with `.sln` file as .NET project, which is confusing.**

#### Motivation and Context

Closes #2567

I found that **the .NET module will be triggered in my C++ project**. 
Here is the screen shot:

图片

After fix, the `.sln` file won't trigger the .NET module
图片

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -839,7 +839,6 @@ the following files are present in the current directory:
 - `Directory.Build.props`
 - `Directory.Build.targets`
 - `Packages.props`
-- `*.sln`
 - `*.csproj`
 - `*.fsproj`
 - `*.xproj`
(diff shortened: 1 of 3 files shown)


andytom (approved):
LGTM

davidkna:
Thanks for the fix!

PR #2789, opened 2021-06-07
Changed the starship file path

Ohssie:
Being that we executed this command `mkdir -p ~/.config && touch ~/.config/starship.toml`, the file path was meant to me `~/.config/starship.toml` and not `~/.starship/config.toml`

#### Motivation and Context

Closes #

#### Checklist:


andytom:
Hi @Ohssie, thank you for your suggestion unfortunately I believe that the current docs are correct as they are supposed to show an example of using a different configuration file so I am going to close this PR.

PR #2791, opened 2021-06-08
feat(username): Detect Admin access in Windows

### LINKED ISSUE #3628: Windows: Display if user has administrative permissions
I cannot see on a glance if the powershell/command line is started with administrative permissions or not.
On linux, the color of the username changes if the shell is used as root.

Use the `style_root` when the powershell/cmd is executed with administrative permissions (e.g. by Rightclick » Run as administrator)

With powershell for example, you can check if the shell has administrative permissions like this:

```powershell
$Identity = [Security.Principal.WindowsIdentity]::GetCurrent()
$Principal = [Security.Principal.WindowsPrincipal] $Identity
$Principal.IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator")
```

rashil2000:
#### Description
This PR adds support for detecting elevated status on Windows, much like root on unix.

#### Motivation and Context
Closes #3628

The root part of the username module was not working on Windows. Hence, the PR.
The elevation detection is done using the `deelevate` crate.

#### Checklist:


rashil2000:
Is the Windows test suite always run as an elevated user? The username tests are failing because of that.

rashil2000:
But why? 

Is there something I can do to rectify it?

davidkna:
> But why?
> 
> Is there something I can do to rectify it?

Likely to make it easier to install additional software. I don't think to can really rectify this, but ideally for testing admin (and root elsewhere) would be mocked.

rashil2000:
So should I modify the tests for this? I'd like some guidance on how to proceed here.

rashil2000:
@davidkna I have added a condition to not detect elevation when tests are running, for the time being. 

Let me know if this is sufficient.

rashil2000:
@davidkna @matchai Since our last discussion was involved around including tests for this PR (as it uses `unsafe`), would it be alright to delegate this functionality to a crate instead? The crate is [deelevate](https://crates.io/crates/deelevate) (and it does have tests in it). It would simplify the implementation a lot.

chipbuster:
@rashil2000 Sorry for the dead time here, let us know if you're still interested in getting this merged!

As far as the question goes, we had a question on Discord in September which asked about using it at runtime, but the deelevate docs described it as working only on Windows 10 (and presumably Windows 11). Do you know if the crate works on Windows 8.1 as well? We nominally support it, but as the Discord discussion revealed, we probably have very, very few users still on 8.1.

rashil2000:
> @rashil2000 Sorry for the dead time here, let us know if you're still interested in getting this merged!

Yes! I still build starship from source at each new release just so I can use this feature 😅

> As far as the question goes, we had a question on Discord in September which asked about using it at runtime, but the deelevate docs described it as working only on Windows 10 (and presumably Windows 11). Do you know if the crate works on Windows 8.1 as well? We nominally support it, but as the Discord discussion revealed, we probably have very, very few users still on 8.1.

I'm not entirely sure. Checking the source, I found a comment in src/psuedocon.rs regarding creation of a pseudoconsole, which requires Windows 10. I don't know if this part is essential in elevation detection, so I opened an issue with the author there - https://github.com/wez/EleDo/issues/3

rashil2000:
The author has responded saying that older systems will run fine as long as we are only querying the privilege level (which is what we need).

Should I switch over to using this crate then?

chipbuster:
Yeah, that sounds like a good idea.

davidkna:
Deelvate seems to pull in a lot of dependencies, some of which only make sense for a binary. For instance, is the `build.rs` going to set the starship icon and file version to something different? Do you know any alternative crates that are more focussed on providing a library?

chipbuster:
It would be nice if this were implemented https://github.com/rust-lang/cargo/issues/1430, since the embed_resource only seems relevant when compiling a bin, but seeing as it's been open with no movement since 2015, I'm guessing it won't be fixed anytime soon.

chipbuster:
I've scoured lib.rs, and the only other thing I've seen that's remotely close is [is_elevated](https://lib.rs/crates/is_elevated). It looks much more like the kind of one-purpose library that we'd want for this kind of thing.

I don't even know that it handles this situation correctly (the [SO post that it's based on](https://stackoverflow.com/a/8196291/2914377) seems to have some controversy in the comments about its correctness) and I'm not sure we can rely on it to get updates, but at a quick 10-minute glance, that seems like our only other option for now.

rashil2000:
> For instance, is the `build.rs` going to set the starship icon and file version to something different? Do you know any alternative crates that are more focussed on providing a library? We just avoided `embed-resource` because it can cause compile failures.

Welp, I think you're right. After merging latest master, I can no longer compile this branch, most likely due deelevate trying to set binary resources and clashing with starship's own:
```
          CVTRES : fatal error CVT1100: duplicate resource.  type:VERSION, name:1, language:0x0409
          LINK : fatal error LNK1123: failure during conversion to COFF: file invalid or corrupt
```

rashil2000:
There's another one - [is_root](https://crates.io/crates/is-root) - but it's implementation is exactly the same as the one initially used in this PR, and no tests...

chipbuster:
Woof.

These are both MIT crates right? We could write our own version of the crate, add tests, and then host that in the starship organization? We already host a maintenance-only version of rust-battery, and I doubt that this would contribute much to our maintenance burden.

I don't have a solid feel for how much work it would be to write our own tests for that kind of crate though.

davidkna:
At that point it could just be pulled into `utils/`, but it might be simpler to test in a separate crate - and there doesn't seem to be a great library in this space so far. There's also another implementation in #1157.

davidkna:
I think this could potentially hurt admin detection, needs to be tested:
https://github.com/starship/starship/blob/a98908f05eab306c3e54820f153de4aa7df41cfe/starship.exe.manifest#L3-L12

rashil2000:
I really should've merged master before making the deelevate commit here 😓

I'll revert the changes, sync with upstream, and try again tomorrow.

Meanwhile I have asked about the deelevate binary issue in https://github.com/wez/EleDo/issues/3 - let's see if there's a way around that.

philippetev:
Guys, any progress here?

rashil2000:
Blocked by https://github.com/wez/EleDo/pull/4

rashil2000:
The upstream PR was merged. @chipbuster @davidkna this is good to go, I think.

rashil2000:
Bump?

src/modules/username.rs
```diff
@@ -16,12 +19,25 @@ const USERNAME_ENV_VAR: &str = "USERNAME";
...
+    #[cfg(any(not(target_os = "windows"), all(target_os = "windows", test)))]
     let username = context.get_env(USERNAME_ENV_VAR)?;
 
     let mut module = context.new_module("username");
     let config: UsernameConfig = UsernameConfig::try_load(module.config);
 
     let is_root = is_root_user();
+
+    #[cfg(all(target_os = "windows", not(test)))]
+    if is_root {
```
davidkna, line 37: If you do `cfg!(target_os = "windows")` here (in tests `is_root` is always `false` on Windows), you can simplify this function and remove all the `#[cfg]` statements

davidkna (approved):
I have a small suggestion, but this looks good to me.

src/modules/username.rs
rashil2000: Sorry, I didn't get you. This part of the statement just changes the username string to "Administrator", no? The `is_root` logic will stay unchanged.

davidkna: `is_root_user()` always returns `false` for tests on Windows, so `not(test)` isn't needed. And, if you used `cfg!` here instead of a cfg-block, `username` could just be `mut` without conditions.

rashil2000:
Done

src/modules/username.rs
```diff
@@ -16,12 +16,18 @@ const USERNAME_ENV_VAR: &str = "USERNAME";
...
-    let username = context.get_env(USERNAME_ENV_VAR)?;
+    let mut username = context.get_env(USERNAME_ENV_VAR)?;
 
     let mut module = context.new_module("username");
     let config: UsernameConfig = UsernameConfig::try_load(module.config);
 
     let is_root = is_root_user();
+
+    cfg!(target_os = "windows");
+    if is_root {
```
davidkna, line 27: The `cfg!`-macro is an expression that returns a boolean value.
```suggestion
    if cfg!(target_os = "windows") && is_root {
```

rashil2000: Apologies for the rookie mistake. Fixed now

src/modules/username.rs
```diff
@@ -67,7 +73,17 @@ fn is_login_user(context: &Context, username: &str) -> bool {
         .map_or(true, |logname| logname == username)
 }
 
-#[cfg(target_os = "windows")]
+#[cfg(all(target_os = "windows", not(test)))]
+fn is_root_user() -> bool {
+    use deelevate::{PrivilegeLevel, Token};
+    let token = Token::with_current_process().unwrap();
+    matches!(
+        token.privilege_level().unwrap(),
```
davidkna, line 80: I also just noticed that you're using `unwrap` here. Can you change this to soft-fail with `log::warn!`?

rashil2000: According to the function definition, `privilege_level` will only return one of the 3 PrivilegeLevel enum values. How about we handle the error in the statement above this one (i.e. `let token = Token::with_current_process().unwrap();`)?

davidkna: It looks like both functions can fail, or they wouldn't be returning a `Result`. Please use something like this for both `unwrap`s:
```rs
let token = match Token::with_current_process() {
    Ok(token) => token,
    Err(e) => {
        log::warn!("Failed to get process token: {e:?}");
        return false;
};
```

rashil2000: Done

davidkna (approved):

davidkna:
No. I'll go ahead and merge it if you fix the conflicts.

davidkna:
Thanks for working on this @rashil2000!

PR #2797, opened 2021-06-13
feat(env_var): allow multiple instances

### LINKED ISSUE #1194: Update the env_var module to allow for multiple instances

At the moment the `env_var` module can only support displaying a single environmental variable. It would be nice to expand this to allow multiple different env vars (similar to the `custom` module). 

I would like to change the way the env var module configuration so that, similar to the `custom`, it is possible to have multiple versions of the env var module displayed at the same time.

Example configuration would be something like the following where each individual instance of the env var module can be configured independently.

```toml
[env_var.bar]
variable = "BAR"
default = "baz"

[env_var.foo]
variable = "FOO"
style = "bold green"
```

This is a breaking change so some kind of migration path would need to be considered.

It is obviously possible with the custom command module, but I think it would be easier/more obvious to configure by keeping a dedicated env var module.

cymruu:
Let me know if it can be merged, then I will update the docs :)

#### Description
Allows to display render multiple instances of `env_var` module.

#### Motivation and Context

Closes #1194

image

#### Checklist:


andytom:
I have only had a quick look at this but it looks like this contains a breaking change for anyone using the env_var module at the moment, is it possible to make this change either in a non breaking way? If not then we will need to provide some extra documentation to provide a migration path for existing users and to make sure we publish this in our release notes for that version.

matchai:
We should be able to introduce this as a non-breaking change by handling `env_var.variable` as though there's one variable, and `env_var.<anything_else>` as multiple. 😊

src/modules/mod.rs
```diff
@@ -94,7 +93,29 @@ pub fn handle<'a>(module: &str, context: &'a Context) -> Option<Module<'a>> {
...
+                            env_var::module(vec!["env_var"], context)
+                        {
+                            env_modules.push(fallback_env_var_module);
+                        }
+                    }
+                    env_var::env_var_displayer(env_modules, context)
+                } else {
+                    None
+                }
+            }
```
matchai, line 118: In an effort to unify the env_var module logic to one place, I think it would be best if this was contained in a function called from `src/modules/env_var.rs`.

matchai (commented):
This is looking good! 👏 
Just left one quick comment, and we'll need to update the docs over in `docs/config/README.md` before we can get this merged.

matchai (approved):
Looks good to me! 👍 
Thank you for your contribution, Filip! 😄 

I've gone ahead and removed the notice indicating that the configuration has changed, so as to avoid making this a breaking change. We'll add a deprecation/update notice when we bump to v1.

glensc:
This is a nice contribution, but coming from search engine here, I don't understand how to use it. the description includes only screenshot, but no usage (or reference to usage)

andytom:
> This is a nice contribution, but coming from search engine here, I don't understand how to use it. the description includes only screenshot, but no usage (or reference to usage)

[Our website has more full documentation](https://starship.rs/config/#environment-variable), also if you check the files of this PR it includes the update to the documentation.

glensc:
ok, so I get it allows to have `[env_var.USER]`, `[env_var.USER2]`, but they will still have only one place in a global `format = "$env_var"`.

in other words, I can't have two instances:
```
format = "$env_var.USER$directory.$env_var_USER2"
```

andytom:
No there is an open issue #3259 for it.

PR #2704, opened 2021-05-09
docs: Added the xonsh shell section

anki-code:
Closes #272

#### Description

The xonsh shell has the pretty tested and full featured support of Starship. I've added the section about how to initilize this in xonsh to the docs.

#### Motivation and Context

The xonsh users like the Starship support in xonsh. The xonsh section will help Starship grow.

#### Screenshots (if appropriate):

image

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [x] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.

PR #1265, opened 2020-05-27
feat: Add xonsh support

vladimyr:
#### Description
Adds `xonsh` init script.

#### Motivation and Context

Closes #272 

#### Screenshots (if appropriate):
image

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #2807, opened 2021-06-19
feat: add support for xonsh

### LINKED ISSUE #272: Support for the xonsh shell
If it is possible to combine starship with the Python-based shell [xonsh](https://xon.sh/tutorial.html), I think the result would be awesome.

jeremyschlatter:
#### Description

Adds `xonsh` init script.

#### Motivation and Context

This is basically #1265 with merge conflicts resolved and a slightly different implementation of job count.

#1265 got stuck on two issues, which I believe are both now resolved:

- `NameError: name 'get_next_job_number' is not defined`
I worked around this by using `len(__xonsh__.all_jobs)` rather than `get_next_job_number`. h/t @anki-code for [that implementation](https://github.com/anki-code/xontrib-prompt-starship/blob/25cf80179874e3dd678854797d596eafcd06dcf3/xontrib/prompt_starship.py#L15).
- https://github.com/xonsh/xonsh/issues/1403
Seems to now be resolved upstream? The upstream issue isn't closed, but some fix PRs have been merged, and I didn't encounter this problem while testing.

Closes #2704
Closes #1265
Closes #272

Screen Shot 2021-06-19 at 10 33 09 AM

#### Checklist:


andytom:
@vladimyr and  @anki-code you have both been working on xonsh support would you be able to take a quick look at this and add your thoughts.

anki-code:
@andytom looks good! 

You can also take a short look into [xontrib-prompt-starship](https://github.com/anki-code/xontrib-prompt-starship) to compare your and my implementation.

jeremyschlatter:
Thanks @anki-code!

The two implementations are very similar. For reference, here they are side-by-side.

[current PR](https://github.com/jeremyschlatter/starship/blob/d673c431ba3f064df250221ac3e45c44147c56c2/src/init/starship.xsh#L1-L6):

```xsh
def starship_prompt():
    last_cmd = __xonsh__.history[-1] if __xonsh__.history else None
    status = last_cmd.rtn if last_cmd else 0
    jobs = len(__xonsh__.all_jobs)
    duration = round((last_cmd.ts[1] - last_cmd.ts[0]) * 1000) if last_cmd else 0
    return $(::STARSHIP:: prompt --status=@(status) --jobs=@(jobs) --cmd-duration=@(duration))
 ```
 
 [xontrib-prompt-starship](https://github.com/anki-code/xontrib-prompt-starship/blob/25cf80179874e3dd678854797d596eafcd06dcf3/xontrib/prompt_starship.py#L9-L16):
 
 ```python
 def _starship_prompt(cfg=None):
    with __xonsh__.env.swap({'STARSHIP_CONFIG': cfg} if cfg else {}):
        return __xonsh__.subproc_captured_stdout([
            'starship', 'prompt',
            '--status', str(int( __xonsh__.history[-1].rtn)) if len(__xonsh__.history) > 0 else '0',
            '--cmd-duration' , str(int((__xonsh__.history[-1].ts[1] - __xonsh__.history[-1].ts[0])*1000)) if len(__xonsh__.history) > 0 else '0',
            '--jobs', str(len(__xonsh__.all_jobs))
        ])
```

anki-code:
Also please take a look into these fixes:

https://github.com/anki-code/xontrib-prompt-starship/blob/25cf80179874e3dd678854797d596eafcd06dcf3/xontrib/prompt_starship.py#L5-L6

```python
__xonsh__.env['STARSHIP_SHELL'] = 'sh'  # Fix https://github.com/anki-code/xontrib-prompt-starship/issues/1
__xonsh__.env['STARSHIP_SESSION_KEY'] = __xonsh__.subproc_captured_stdout(['starship','session']).strip()
```

anki-code:
And I'm using xonsh subprocess call from the library (`__xonsh__.subproc_captured_stdout`) to make the code pure python. It allows to translatable the code to python bytecode.

jeremyschlatter:
Thanks for the pointers.

I added `STARSHIP_SESSION_KEY`.

This PR already sets `$STARSHIP_SHELL = "xonsh"`, which both solves the issue behind anki-code/xontrib-prompt-starship#1 and is better than `$STARSHIP_SHELL = "sh"` in this context.

Re: pure python, I don't understand how starship benefits from having this code be translatable to python bytecode. Is there any immediate benefit to starship? If not, I think it makes sense to leave the init script here in xonsh syntax, since that syntax is cleaner.

anki-code:
> Is there any immediate benefit to starship? If not, I think it makes sense to leave the init script here in xonsh syntax, since that syntax is cleaner.

Nope, you can leave the xonsh syntax.

src/init/starship.xsh
```diff
@@ -0,0 +1,11 @@
...
+    last_cmd = __xonsh__.history[-1] if __xonsh__.history else None
+    status = last_cmd.rtn if last_cmd else 0
+    jobs = len(__xonsh__.all_jobs)
+    duration = round((last_cmd.ts[1] - last_cmd.ts[0]) * 1000) if last_cmd else 0
+    return $(::STARSHIP:: prompt --status=@(status) --jobs=@(jobs) --cmd-duration=@(duration))
+
+
+$PROMPT = starship_prompt
+$STARSHIP_SHELL = "xonsh"
+$STARSHIP_SESSION_KEY = $(::STARSHIP:: session).strip()
```
davidkna, line 11: Is it possible to generate this from `xonsh` instead? Something like `"".join(secrets.choice(string.ascii_letters + string.digits) for _ in range(16))`?

jeremyschlatter: Done. I used `random` rather than `secrets` because xonsh's minimum Python version is 3.5, which did not yet have `secrets`. I assume this key does not need cryptographically secure randomness, but please let me know if I'm mistaken about that.

docs/.vuepress/config.js
```diff
@@ -50,13 +50,13 @@ module.exports = {
   head: [
     ["link", { rel: "icon", href: "/icon.png" }],
     ["meta", { property: "og:title", content: "Starship: Cross-Shell Prompt" }],
-    ["meta", { property: "og:description", content: "Starship is the minimal, blazing fast, and extremely customizable prompt for any shell! Shows the information you need, while staying sleek and minimal. Quick installation available for Bash, Fish, ZSH, Ion, and Powershell."}],
+    ["meta", { property: "og:description", content: "Starship is the minimal, blazing fast, and extremely customizable prompt for any shell! Shows the information you need, while staying sleek and minimal. Quick installation available for Bash, Fish, ZSH, Ion, Xonsh, and Powershell."}],
```
davidkna, line 53: Since you're changing this: Please also list tcsh, elvish and nu (also for the other description items)

src/init/starship.xsh
```diff
@@ -0,0 +1,15 @@
...
+    last_cmd = __xonsh__.history[-1] if __xonsh__.history else None
+    status = last_cmd.rtn if last_cmd else 0
+    jobs = len(__xonsh__.all_jobs)
+    duration = round((last_cmd.ts[1] - last_cmd.ts[0]) * 1000) if last_cmd else 0
+    return $(::STARSHIP:: prompt --status=@(status) --jobs=@(jobs) --cmd-duration=@(duration))
+
+
+$PROMPT = starship_prompt
+$STARSHIP_SHELL = "xonsh"
+$STARSHIP_SESSION_KEY = "".join(random.choice(string.ascii_letters + string.digits) for _ in range(16))
```
davidkna, line 15: I'm getting a `NameError: name 'random' is not defined` here.

src/init/starship.xsh
```diff
@@ -0,0 +1,15 @@
+import random
+import string
+
+
+def starship_prompt():
+    last_cmd = __xonsh__.history[-1] if __xonsh__.history else None
+    status = last_cmd.rtn if last_cmd else 0
+    jobs = len(__xonsh__.all_jobs)
```
davidkna, line 8: After running any command the jobs will sometimes display for me, with only the starship init in my `.xonshrc`.

davidkna (changes requested):

docs/.vuepress/config.js
jeremyschlatter: Done.

src/init/starship.xsh
jeremyschlatter: Reproduced on my end.

I'm pretty confused by xonsh's behavior here. I've reduced some of the surprising behavior to a minimal example and asked about it on the xonsh gitter: https://gitter.im/xonsh/xonsh?at=60e7d763951c58084ecb7108

andytom: As a work around you could use the uuid module from the python std library. I think this would be enough we just need a unique id for each session 
```python
srt(uuid.uuid4())
# or
uuid.uuid4().hex
```

jeremyschlatter: Done. Thanks for that suggestion.

src/init/starship.xsh
jeremyschlatter: [Fixed](https://github.com/starship/starship/pull/2807/commits/57ea974f84017c53f5bf8fc33f27894205180003).

Getting job accounting right in xonsh apparently requires regularly calling [`_clear_dead_jobs`](https://github.com/xonsh/xonsh/blob/c06ed5ef2dcc44ba7a4b1c3522f0864cddffef81/xonsh/jobs.py#L265-L273).

We could accomplish that here by calling [`get_next_job_number`](https://github.com/xonsh/xonsh/blob/c06ed5ef2dcc44ba7a4b1c3522f0864cddffef81/xonsh/jobs.py#L291-L297). In fact, #1265 did [exactly that](https://github.com/starship/starship/blob/f34be39da2230cde361f8c8611e95d95b52b8326/src/init/starship.xsh#L7).

Unfortunately, due to what I [believe](https://gitter.im/xonsh/xonsh?at=60e8832d82dd9050f5e0c96a) is the [same](https://github.com/starship/starship/pull/1265#pullrequestreview-420227964 ) issue as [above](https://github.com/starship/starship/pull/2807#discussion_r666360733), we're pretty limited here: we can't use any imported functions or even any names we define outside of this function. As a workaround, I've duplicated enough of the relevant logic in `_clear_dead_jobs` to ignore dead jobs in our count here.

> After running any command the jobs will sometimes display for me, with only the starship init in my .xonshrc.

Incidentally, I reproduced this on xonsh 0.9.27 but not on xonsh's latest main branch. But I don't understand why they would be different. 🤷🏻‍♂️

src/init/starship.xsh
```diff
@@ -0,0 +1,16 @@
...
+
+
+def starship_prompt():
+    last_cmd = __xonsh__.history[-1] if __xonsh__.history else None
+    status = last_cmd.rtn if last_cmd else 0
+    # I believe this is equivalent to xonsh.jobs.get_next_job_number() for our purposes,
+    # but we can't use that function because of https://gitter.im/xonsh/xonsh?at=60e8832d82dd9050f5e0c96a
+    jobs = sum(1 for job in __xonsh__.all_jobs.values() if job['obj'] and job['obj'].poll() is None)
+    duration = round((last_cmd.ts[1] - last_cmd.ts[0]) * 1000) if last_cmd else 0
+    return $(::STARSHIP:: prompt --status=@(status) --jobs=@(jobs) --cmd-duration=@(duration))
```
davidkna, line 11: Now that `$STARSHIP_SESSION_KEY` works I've found another issue:
Log messages (stderr) from starship don't get displayed (invalid config file). Maybe it needs a workaround like this for [PowerShell](https://github.com/starship/starship/blob/master/src/init/starship.ps1#L56-L62)?

meramsey: seems to work for me when using https://github.com/anki-code/xontrib-prompt-starship latest and 

```
@ xonsh --version
xonsh/0.9.27
@ python -V
Python 3.9.2
@ starship --version
starship 0.55.0
tag:v0.55.0
commit_hash:09b12a52
build_time:2021-06-20 18:31:06
build_env:rustc 1.53.0 (53cb7b09b 2021-06-17)
```

image
image

jeremyschlatter: Good catch, @davidkna.

This was fixed upstream in xonsh just three weeks ago: https://github.com/xonsh/xonsh/pull/4336

But since no released versions of xonsh contain that fix, we should probably work around it here. I've added a workaround: https://github.com/starship/starship/pull/2807/commits/1e0bc8d6a3ccb205b1f4069d29776a8f44285a75

davidkna (approved):
LGTM

PR #2827, opened 2021-06-27
feat: Add the `git_metrics` module

alexfertel:
#### Description

This PR adds a new module named `git_metrics`. It shows the added/deleted lines in the current git repository following the format: `"[+$added_lines]($added_style) [-$deleted_lines]($deleted_style)"`.

I implemented this functionality by parsing the output of `git diff --shortstat`.

#### Motivation and Context

Closes https://github.com/starship/starship/issues/2725

I wrote 4 tests that cover the main functionality. It is a rather simple module, so I don't think there is much to it, but if you think that something's missing let me know.

#### Checklist:


src/modules/git_metrics.rs
```diff
@@ -0,0 +1,288 @@
...
+fn get_deleted_lines(diff: &str) -> Option<String> {
+    let re = Regex::new(r"^\[-.*-\]$").unwrap();
+
+    Some(
+        diff.lines()
+            .filter(|line| re.is_match(line))
+            .count()
+            .to_string(),
+    )
+}
```
matchai, line 90: Instead of iterating through the diff three times, it may be simpler and more declarative to have a `GitDiff` struct that parses output a single time:

```rust
struct GitDiff {
  added: u32,
  modified: u32,
  deleted: u32,
}

impl GitDiff {
  fn parse(diff: &str) -> Self;
}

```

src/modules/git_state.rs
```diff
@@ -337,7 +337,7 @@ mod tests {
 
         // Ensure on the expected branch.
         // If build environment has `init.defaultBranch` global set
-        // it will default to an unknown branch, so neeed to make & change branch
+        // it will default to an unknown branch, so need to make & change branch
```
matchai, line 340: 👍

src/modules/git_metrics.rs
```diff
@@ -0,0 +1,288 @@
...
+
+#[cfg(test)]
+mod tests {
```
matchai, line 93: Tests are looking good 👍

matchai (commented):
Good stuff so far!

src/modules/git_metrics.rs
```diff
@@ -0,0 +1,275 @@
+use regex::Regex;
+
+use crate::{
+    config::RootModuleConfig, configs::git_metrics::GitMetricsConfig, formatter::StringFormatter,
+    module::Module,
+};
+
+use super::Context;
+
+/// Creates a module with the current added/modified/deleted lines in the git repository at the
```
matchai, line 10:
```suggestion
/// Creates a module with the current added/deleted lines in the git repository at the
```

src/modules/mod.rs
```diff
@@ -177,6 +179,7 @@ pub fn description(module: &str) -> &'static str {
         "gcloud" => "The current GCP client configuration",
         "git_branch" => "The active branch of the repo in your current directory",
         "git_commit" => "The active commit (and tag if any) of the repo in your current directory",
+        "git_metrics" => "The currently added/modified/deleted lines in your repo",
```
matchai, line 182:
```suggestion
        "git_metrics" => "The currently added/deleted lines in your repo",
```

matchai (commented):
Dropped a few small suggestions, but otherwise, it looks good to me! 👍

alexfertel:
> Dropped a few small suggestions, but otherwise, it looks good to me!

Sorry, I should have checked all the places where 'modified' was written.

PR #2836, opened 2021-06-30
style: fix name of battery info provider implementation

ahouts:
#### Description
I noticed right after @matchai merged that the implementation of `BatteryInfoProvider` had an old name from when the interface was called `BatteryStatusProvider`, this fixes that mistake.

#### Motivation and Context
The name is confusing.

#### Checklist:


matchai (approved):
Whoops! Good catch! 👏

PR #2843, opened 2021-07-02
perf: Improve custom config OS check

giggio:
#### Description
Moving the OS check before other checks so that there is no performance hit for running the match in the incorrect OS.

#### Motivation and Context
I noticed that the `when` config condition always runs, even in an OS that did not match. This causes a performance hit on Windows, as the command might not exist there, and it will take a while to continue. I noticed 2 to 3 second pauses on Windows. These pauses do not happen on Linux.
Related to #2750 and #2751.

#### Checklist:


andytom (approved):

andytom:
Thank you for your contribution @giggio

PR #2847, opened 2021-07-03
feat: add support for nu shell

### LINKED ISSUE #1930: Starship gets garbled when running the prompt via nushell via zsh
#### Current Behavior
You can see some user reports on the Nushell repo: https://github.com/nushell/nushell/issues/1005. Basically, it seems like if the user runs zsh as their main shell, and then from there invoke Nushell, if Nushell is running starship as its prompt the result is a garbled prompt.

image

#### Expected Behavior
If there's an autodetection happening, I wonder if it should detect Nushell first? I'm just guessing, but it seems like it's seeing the zsh first.

There's a workaround to unset STARSHIP_SHELL before starting Nu, mentioned here: https://github.com/nushell/nushell/issues/1005#issuecomment-716879319

#### Possible Solution
Is this something that Nushell should trying to fix or something that Starship could fix? Definitely happy to chat about possible solutions.

#### Environment
- Shell type: zsh + Nushell

davidkna:
#### Description

This PR adds support for the nu shell.
Init involves piping the output of `starship init nu` to a file, because nushell doesn't support sourcing from stdin. Because of this I added a warning that the init will change in the future. But maybe it would be better to wait until that's supported before merging this.

#### Motivation and Context

Closes #1717
Closes #1930
Closes #2334

#### Checklist:


andytom:
I think this might also close #1716 and #1930.

davidkna:
It doesn't fix #1716 because nu still doesn't expose that yet, but it should close #1930.

andytom (commented):
LGTM, but I am not a nu shell user so probably worth getting a second opinion on this one.

andytom (approved):

Cargo.toml
```diff
@@ -33,6 +33,7 @@ tls-vendored = ["native-tls/vendored"]
 clap = "2.33.3"
 ansi_term = "0.12.1"
 dirs-next = "2.0.0"
+directories-next = "2.0.0"
```
matchai, line 36: It seems to me like `directories-next` is a superset of `dirs-next`.
Since `directories-next` also has a `home_dir()` method, we can use the opportunity to replace `dirs-next` calls altogether and remove the dependency.

davidkna: Done!

PR #2850, opened 2021-07-03
feat(Modules): Use nerd-fonts more consistently

Milo123459:
#### Description

Changes a hand full of modules to use nerd fonts.
#### Motivation and Context

Closes #
Sometimes the regular emojis just don't fit with it.

#### Checklist:


andytom:
Not that this should block this but I have been thinking about proposing the exact opposite change, removing all nerd fonts from the defaults. This would remove dependency on having a nerd font installed which is a source of quite a few issues and questions on Discord and makes starship even easier to install since no extra fonts to install. We still have the preset for nerd fonts which hopefully makes it easy for those who want to use nerd fonts to do so.

Milo123459:
So you are suggesting a way of detecting whether or not they are using a nerd font and if they are apply an icon? For example the git modules, they need the icons. How would that work?

andytom:
No my suggestion would basically be the same as what you have done here but to remove the use of the Nerd Font icons from the 2 or 3 modules that currently use them changing their default symbols to be emojis or some other commonly available glyph. We already have a set of [presets on the website] (https://starship.rs/presets/#nerd-font-symbols) where users can get the configuration from for nerd font symbols, plain text and others so for now it would be a manual change for the users who want to use nerd fonts until we add a config wizard.

matchai:
I'm going to have to side with Tom here. Plenty of users unknowingly use Starship without NerdFonts right now since most of the core modules don't use them. I think this will end up breaking a partly working prompt for many people.

This is certainly something we can revisit when we have a configuration wizard.

PR #2855, opened 2021-07-04
docs(config): change spelling of some words

### LINKED ISSUE #2846: "Golang" module and documentation refers to Go as Golang

The Go language isn't actually called Golang, that's just a useful term for search engines to help find Go-related content. In almost all references to Go on the `starship.rs` website Go is referred to as Golang.

Replace "Golang" with Go in all references on `starship.rs`, perhaps leaving one use of "Golang" somewhere to aid with search engine optimisation.

I considered just not worrying about it but I decided to worry about it.

kidonng:
#### Description

Change spelling of some words, as well as some other small documentation fixes.

#### Motivation and Context

Closes #2846

#### Checklist:


docs/config/README.md
```diff
@@ -2908,9 +2908,9 @@ By default the module will be shown if any of the following conditions are met:
 format = "via [⍱ $version](bold white) "
 ```
 
-## VLang
+## V
```
Milo123459, line 2911: Pretty sure this should still be VLang

Milo123459:
```suggestion
## VLang
```

kidonng: `[citation needed]`

Milo123459: What? Do you need a reason for it? If so, look at Golang, it's Golang, not Go. This should be done for consistency. Or change both I suppose.

kidonng: The initial purpose of this PR is to change "Golang" to "Go". Please read the full diff.

matchai:
It may not be a satisfying answer, but the reason I decided to use "Golang" and"Vlang" is to make our docs indexable and searchable. Search providers will often lower the confidence score of search terms that are very short because they have multiple purposes.

For instance, searching "Starship go" on a search engine will produce results about SpaceX and their starship "going" to mars. Not to say that we can't compete with those indexed result, but Go users have long since gotten used to using the "golang" search term to disambiguate between the "going" verb and the boardgame of "Go".

Similarly, it makes it easy for us to search our own codebase for couplings with the golang module. Searching "go" brings up any reference to "Car**go**".

I'd be open to changing it though. Just sharing the reasoning behind why it is this way today. 😊

kidonng:
As said above, the main reason of this PR is to resolve the concern mentioned in https://github.com/starship/starship/issues/2846.
Personally I don't have any preference, but I feel the changes looks nicer and so I  went to do it.
Note that this is a documentation only change, the module is still names `golang` and `vlang` respectively, and still searchable.
Regarding search engines, actually Google shows `golang` module as the first result, so the changes in this PR doesn't seem to affect that.

PR #2870, opened 2021-07-08
fix: prevent cutting of the time

j-m-hoffmann:
#### Description
Changing the formatstring from double to single quotes with escapes ensures the time is shown correctly

#### Motivation and Context
Looked bad.

Before:
before
After:
after

#### Checklist:

### DIFF
--- a/docs/presets/README.md
+++ b/docs/presets/README.md
@@ -113,7 +113,7 @@ format = '\[[$symbol($profile)(\($region\))(\[$duration\])]($style)\]'
 format = '\[[$symbol($version)]($style)\]'
 
 [cmd_duration]
-format = "[⏱ $duration]($style)"
+format = '\[⏱ $duration\]($style)'
 
 [conda]
 format = '\[[$symbol$environment]($style)\]'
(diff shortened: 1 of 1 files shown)


docs/presets/README.md
```diff
@@ -113,7 +113,7 @@ format = '\[[$symbol($profile)(\($region\))(\[$duration\])]($style)\]'
 format = '\[[$symbol($version)]($style)\]'
 
 [cmd_duration]
-format = "[⏱ $duration]($style)"
+format = '\[⏱ $duration\]($style)'
```
andytom, line 116: I think this should actually be 

```suggestion
format = '\[[⏱ $duration]($style)\]'
```

Otherwise the style will not be applied correctly

j-m-hoffmann: That way it will look like this:

Screenshot_20210803_102628

j-m-hoffmann: Adding a space behind duration works
``` toml
format = '\[[⏱<ICON> $duration<SPACE>]($style)\]' 
```

andytom (approved):

andytom:
Thank you for your contribution @j-m-hoffmann

PR #2873, opened 2021-07-09
fix: PowerShell init - replace Get-Error with $error[0]

potatoqualitee:
#### Description
`Get-Error` does not exist in all versions of PowerShell, and attempting to use it on machines where it does not exist pollutes `$error`. Someone may be tempted to use `-ErrorAction Ignore` but since the command does not exist, it still ends up in `$error`

This is a dual bug fix because it actually gets errors on all machines now and it does not pollute `$error`.
#### Motivation and Context
In looking for unrelated errors, I kept seeing an error about `Get-Error` not existing, and narrowed it down to starship.

I also changed `-expand` to `-ExpandProperty` while I was at it because it is best practice.

image

`$error[0]` exists in all versions of PowerShell and will not cause issues even if there are no errors

image

I didn't test within starship, but I did at the console. First, I opened a fresh console with no errors and you can see that attempting to get `$lastCmdletError` did not cause errors. Next, I forced an error, then I ran the line again and you can see that it did get `$lastCmdletError`

image

#### Checklist:

There are no other instances of `Get-Error`, including tests, so this was likely the only place it was being used

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -82,7 +82,7 @@ function global:prompt {
         # In case we have a False on the Dollar hook, we know there's an error.
         if (-not $origDollarQuestion) {
             # We retrieve the InvocationInfo from the most recent error.
-            $lastCmdletError = try { Get-Error |  Where-Object { $_ -ne $null } | Select-Object -expand InvocationInfo } catch { $null }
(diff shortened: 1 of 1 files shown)


potatoqualitee:
For anyone who wants a workaround right now, you can create the following function in your `$profile` right before you execute starship:

```
    function Get-Error {
        $error[0]
    }
```

matchai (approved):
Looks good to me! 👍 
Thank you for your contribution, Chrissy! 😄

PR #1475, opened 2020-07-13
feat(r-lang): Add module for R programming language

jimleroyer:
#### Description

Added a module for the R programming language. 

#### Motivation and Context

R is number 10 on the latest [Tiobe index](https://www.tiobe.com/tiobe-index/). It made its mark as a statistical language. Putting on support for most popular languages among developers makes starship a bit more attractive with its support arsenal for commonly used tools.

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Screenshots (if appropriate):

image

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [X] I have tested using **Windows**

#### Checklist:

- [X] I have updated the documentation accordingly.
- [X] I have updated the tests accordingly.

PR #2880, opened 2021-07-13
fix(rlang): module not working on MacOs

cymruu:
#### Description
Fix rlang module and refactor other modules to use new method.

#### Motivation and Context
This comment https://github.com/starship/starship/pull/1475#issuecomment-865321898

#### Checklist:


src/utils.rs
```diff
@@ -18,12 +18,21 @@ pub fn read_file<P: AsRef<Path> + Debug>(file_name: P) -> Result<String> {
...
         log::debug!("Error reading file: {:?}", result);
     } else {
-        log::trace!("File read sucessfully");
+        log::trace!("File read successfully");
     };
 
     result
 }
 
+// Reads command output from stderr or stdout depending on to which stream program streamed it's output
```
matchai, line 27: I'd suggest using triple-slashes here so that the comment is interpreted as documentation.
This ensures that Intellisense and IDEs show the description as part of code-completion 😊
 
```suggestion
/// Reads command output from stderr or stdout depending on to which stream program streamed it's output
```

cymruu: Good to know. Thank you 🙂

matchai (approved):
LGTM! Thank you for fixing up the other modules in the process. 😊

PR #2885, opened 2021-07-14
fix(windows): avoid inadvertly running exes from cwd

davidkna:
#### Description

On Windows when running commands with their name instead of the path with `Command::new`, executable with that name from the current working directory will be executed.

This PR replaces all instances of `Command::new` with a new `create_command` function which will first resolve any executable paths and avoid this issue.
I also added clippy configuration options to prevent the use of `Command::new`.

`context.exec_cmd` isn't vulnerable because it already resolves paths.

Apart from the tests only `starship init <shell>` (excluding PowerShell), the rust module (if `$RUSTUP_TOOLCHAIN` is not set) and custom modules are vulnerable.  ~~The `Command::new` statement in the `rust` module can't be reached, because `context.get_env` will never return `None` on Windows.~~

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,6 +68,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 ```
 
+If using `context.exec_cmd` isn't possible, please use `crate::utils::create_command` instead of `std::process::Command::new`.
+
 ## Logging
 
 Debug logging in starship is done with our custom logger implementation.
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,2 @@
+# std::process::Command::new may inadvertly run executables from the current working directory
+disallowed-methods = ["std::process::Command::new"]
--- a/src/bug_report.rs
+++ b/src/bug_report.rs
@@ -266,18 +266,6 @@ mod tests {
     }
 
     #[test]
-    fn test_get_shell_info() {
-        env::remove_var("STARSHIP_SHELL");
-        let unknown_shell = get_shell_info();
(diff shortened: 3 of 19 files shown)


kidonng:
Cool, is this in the same fashion as a bunch of other Rust CLI programs such as [ripgrep](https://github.com/BurntSushi/ripgrep/commit/229d1a8d41b0023420e7815578fa0b39c0d5c2e4), [bat](https://github.com/sharkdp/bat/pull/1724) and [delta](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-36376)? I feel there might or should be a more standard solution to the problem, instead of everyone rolling its own or borrowing from others.

davidkna:
> Cool, is this in the same fashion as a bunch of other Rust CLI programs such as [ripgrep](https://github.com/BurntSushi/ripgrep/commit/229d1a8d41b0023420e7815578fa0b39c0d5c2e4), [bat](https://github.com/sharkdp/bat/pull/1724) and [delta](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-36376)?

Yes.

> I feel there might or should be a more standard solution to the problem, instead of everyone rolling its own or borrowing from others.

I agree. For what it's worth it looks like all the crates you mentioned switched to using [`grep-cli`](https://docs.rs/grep-cli/0.1.6/grep_cli/), but that crate wouldn't be ideal for starship because it looks like it doesn't work with `.cmd` files.

andytom (approved):
LGTM

matchai (approved):

PR #2887, opened 2021-07-14
feat(git_metrics): Git metrics show only nonzero diffs

### LINKED ISSUE #2882: git_metrics: option to suppress if added and deleted are both zero

While the new `git_metrics` feature added in https://github.com/starship/starship/pull/2827 could be handy, I find it intrusive when there are no differences in repository.  The `+0 -0` is a distraction.

An option to only show the metrics when both the added and deleted count are non-zero seems sensible.  Maybe called `only_nonzero_diffs`?  Even though it wouldn't be compatible with the initial behaviour of the feature, I think such an option should default to `true`.

No other options considered.

cymruu:
Add configuration options for git_metrics module to display only if something has changed.

#### Motivation and Context

Closes #2882

#### Checklist:


kidonng:
I wonder if we should just hide it if it's zero.

@alexfertel mind sharing your thoughts on this? I haven't read the original PR carefully so I may have missed why this is not handled.

alexfertel:
It's just a decision I made based on what made sense to me. I thought that maybe it was relevant to know that we were in a state with no changed lines, but that is subjective.

The relevant bit in the code would be:

```rust
    /// Returns the first capture group given a regular expression and a string.
    /// If it fails to get the capture group it will return "0".
    fn get_matched_str(diff: &'a str, re: &Regex) -> &'a str {
        match re.captures(diff) {
            Some(caps) => caps.get(1).unwrap().as_str(),
            _ => "0",
        }
    }
```

`git diff --shortstat` outputs nothing if a variable is `0`, i.e., it doesn't output anything when there are no added lines. The code above puts a `"0"` in the struct if it is the case.

I think not showing anything is ok, in which case I think we might be able to fail so that variable is not rendered.

Milo123459:
Hide if zero is a good idea imo.

kidonng:
> in which case I think we might be able to fail so that variable is not rendered. 

Then it should output an error message. There is no need to distinguish it between empty values by showing the latter all the time.

alexfertel:
Nono, I just meant returning `None`, sorry if it was unclear 😅 I don't think it can fail literally.

chipbuster:
Reopening on author's request (branch was accidentally deleted)

PR #2908, opened 2021-07-22
fix(jobs): Add the symbol and number thresholds respecting the `threshold` option

alexfertel:
#### Description

This PR adds two new config options (`number_threshold` and `symbol_threshold`) to fix the current behavior of `threshold` in the `jobs` module. The main issue is that *greater than* is a bit confusing. We now allow a *greater or equal than* logic that simplifies configuring the behavior of the module.

Namely, we follow:

> The two thresholds would be the minimum value at which they're shown.
> The default functionality is:
> • 0 jobs -> nothing
> • 1 job -> symbol
> • 2 jobs or more -> symbol + number

Which covers all of the use cases by separating the symbol and number thresholds. This PR adds several more tests to ensure the correctness of the feature.

One thing to note is that the `threshold` option will be deprecated and a warning is issued if its value in the config is different from the default of `1`.

#### Motivation and Context

Closes https://github.com/starship/starship/issues/2688

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1549,9 +1549,18 @@ symbol = "🌟 "
 
 The `jobs` module shows the current number of jobs running.
 The module will be shown only if there are background jobs running.
-The module will show the number of jobs running if there is more than 1 job, or
-more than the `threshold` config value, if it exists. If `threshold` is set to 0,
-then the module will also show when there are 0 jobs running.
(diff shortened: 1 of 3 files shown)


andytom (approved):
LGTM

andytom:
Thank you for your contribution @alexfertel

PR #2920, opened 2021-07-27
ci: Use default token for release action

alerque:
See upstream: https://github.com/softprops/action-gh-release/pull/83


alerque:
I realize it's hard to test CI jobs like this, but if you want to compare you can see [this run](https://github.com/alerque/vcsh/runs/3170625401?check_suite_focus=true) on another project where I made exactly the same change and did a test release on a fork to verify that the upstream change had the effect it is advertised to.

matchai (approved):
Perfect! Thank you for the PR, @alerque! 😄

PR #2931, opened 2021-07-28
feat(explain): quote module values

### LINKED ISSUE #2928: `starship explain` should add quotes around each prompt segment
When I run `starship explain` at my home folder, it prints something like

```

 ~  (<1ms)  -  The current working directory
   (<1ms)   -  The currently used shell indicator
 ❯  (<1ms)  -  A character (usually an arrow) beside where the text is entered i
               n your terminal
```

I can't tell where spaces exist here. In particular, if I look at my prompt:

```
~ 
 ❯ 
```

This does not let me figure out where the space before my arrow is.

Each prompt segment in the `starship explain` output should be surrounded by quotes. Alternatively it could use a background color, though this may be confusing for prompts that use background colors already.

#### Tangentially-related notes

The descriptions are being character-wrapped instead of word-wrapped. That's a bit odd.

Also, it would be nice if each entry also quoted the portion of `format` that it came from (e.g. `$directory` for the directory bit). This needs to work correctly for `format = "$all"` too. Or it could just list the module name.

davidkna:
#### Description

This PR changes the `explain` and `timings` commands to add quotes around the module outputs to make spaces inside the modules output more obvious.

#### Motivation and Context

Closes #2928

Before:
```
 ❯  (<1ms)                             -  A character (usually an arrow) beside where the text is entered in your terminal
 starship  (6ms)                       -  The current working directory
```
After:
```
 "❯ " (<1ms)                             -  A character (usually an arrow) beside where the text is entered in your terminal
 "starship " (8ms)                       -  The current working directory
```

#### Checklist:

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -177,10 +177,10 @@ pub fn timings(args: ArgMatches) {
     println!("\n Here are the timings of modules in your prompt (>=1ms or output):");
 
     // for now we do not expect a wrap around at the end... famous last words
-    // Overall a line looks like this: " {module name}  -  {duration}  -  {module value}".
+    // Overall a line looks like this: " {module name}  -  {duration}  -  "{module value}"".
     for timing in &modules {
         println!(
-            " {}{}  -  {}{}  -   {}",
+            " {}{}  -  {}{}  -   \"{}\"",
             timing.name,
             " ".repeat(max_name_width - (timing.name_len)),
             " ".repeat(max_duration_width - (timing.duration_len)),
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM

src/print.rs
```diff
@@ -222,8 +222,8 @@ pub fn explain(args: ArgMatches) {
     let max_module_width = modules.iter().map(|i| i.value_len).max().unwrap_or(0);
 
     // In addition to the module width itself there are also 9 padding characters in each line.
```
cymruu, line 224: I believe it should be updated too (9 ➡️ 11)

PR #2932, opened 2021-07-28
feat(print-config): list expanded `$all` variable

### LINKED ISSUE #2923: `starship print-config` should explain the `$all` variable in `format`
#### Current Behavior
When using the default value for `format`, `starship print-config` just lists

```toml
format = '$all'
```

but does not explain what `$all` is. This is frustrating as it means I can't use this to replace my `format` key in my config with the current effective value.

#### Expected Behavior
When it lists `format = "$all"` it should include a comment showing the effective value, that I can then copy and paste into my config prior to customizing it. This should happen if `format` is unspecified, but it should also happen if I set the `format` key to a value that includes `$all`.

As far as I'm aware, right now the only way to find out the effective value is to go to https://starship.rs/config/#default-prompt-format, though that only tells me about the latest published version, not about whatever is on my machine.

#### Environment
- Starship version: starship 0.55.0
branch:
commit_hash:
build_time:2021-06-30 05:44:17
build_env:rustc 1.53.0,

#### Starship Configuration
The `format` key is not specified in my config, meaning I get the default behavior.

davidkna:
#### Description

In addition to the comment about missing unset values add a comment at the top of the `print-confg` that shows how `$all` is expanded.
This works, but I'm not too happy with the output because it results in a very long line.

#### Motivation and Context

Closes #2923

New output:

```toml
# Warning: This config does not include keys that have an unset value

# $all is shorthand for $username$hostname$shlvl$singularity$kubernetes$directory$vcsh$git_branch$git_commit$git_state$git_metrics$git_status$hg_branch$docker_context$package$cmake$dart$deno$dotnet$elixir$elm$erlang$golang$helm$java$julia$kotlin$lua$nim$nodejs$ocaml$perl$php$purescript$python$rlang$red$ruby$rust$scala$swift$terraform$vlang$vagrant$zig$nix_shell$conda$memory_usage$aws$gcloud$openstack$env_var$crystal$custom$cmd_duration$line_break$jobs$battery$time$status$shell$character
format = '$all'      
scan_timeout = 30
command_timeout = 500
add_newline = true
```

If customer modules are present:
```toml
# Warning: This config does not include keys that have an unset value

# $all is shorthand for $username$hostname$shlvl$singularity$kubernetes$directory$vcsh$git_branch$git_commit$git_state$git_metrics$git_status$hg_branch$docker_context$package$cmake$dart$deno$dotnet$elixir$elm$erlang$golang$helm$java$julia$kotlin$lua$nim$nodejs$ocaml$perl$php$purescript$python$rlang$red$ruby$rust$scala$swift$terraform$vlang$vagrant$zig$nix_shell$conda$memory_usage$aws$gcloud$openstack$env_var$crystal$custom$cmd_duration$line_break$jobs$battery$time$status$shell$character
# $custom (excluding any modules already listed in `format`) is shorthand for ${custom.foo}
format = '$all'      
scan_timeout = 30
command_timeout = 500
add_newline = true
```

#### Checklist:

### DIFF
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -84,7 +84,7 @@ pub fn print_configuration(use_default: bool) {
     let custom_modules = config.get("custom").unwrap().as_table().unwrap();
     if !use_default && !custom_modules.is_empty() {
         println!(
-            "# $custom is shorthand for {}",
+            "# $custom (excluding any modules already listed in `format`) is shorthand for {}",
             custom_modules
                 .keys()
                 .map(|module_name| format!("${{custom.{}}}", module_name))
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -79,6 +79,18 @@ pub fn print_configuration(use_default: bool) {
             .map(|module_name| format!("${}", module_name))
             .collect::<String>()
     );
+
+    // Unwrapping is fine because config is based on FullConfig
+    let custom_modules = config.get("custom").unwrap().as_table().unwrap();
+    if !use_default && !custom_modules.is_empty() {
+        println!(
+            "# $custom is shorthand for {}",
+            custom_modules
+                .keys()
+                .map(|module_name| format!("${{custom.{}}}", module_name))
+                .collect::<String>()
+        );
+    }
     println!("{}", string_config);
 }
 
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -5,6 +5,7 @@ use std::process;
 
 use crate::config::RootModuleConfig;
 use crate::config::StarshipConfig;
+use crate::configs::PROMPT_ORDER;
 use crate::utils;
 use std::fs::File;
 use std::io::Write;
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM

src/configure.rs
```diff
@@ -70,7 +71,14 @@ pub fn print_configuration(use_default: bool) {
 
     let string_config = toml::to_string_pretty(&config).unwrap();
 
-    println!("# Warning: This config does not include keys that have an unset value");
+    println!("# Warning: This config does not include keys that have an unset value\n");
+    println!(
+        "# $all is shorthand for {}",
+        PROMPT_ORDER
+            .iter()
+            .map(|module_name| format!("${}", module_name))
```
jankatins, line 79: For custom modules, you need `${custom.whatever}`.

jankatins:
Just wanted to give my thank you here: I use custom modules which have to go to different places and $all does not work for me. So I really like that this gives me an easy way to compare what the current version on $all is.

src/configure.rs
jankatins: I have it also written as 
```
# empty line on purpose to get a empty line
format = """

${custom.toolbox}\
$username\
$hostname\
...
"""

davidkna: The original code didn't actually have any handling for `$custom`. If changed the code, so that if any custom modules are in the config `$custom` is expanded in addition to `$all`.

Regarding printing the modules line-by-line I wanted to avoid that, because it would make the comment take too much space.

PR #2945, opened 2021-08-01
feat(git_status): added symbol for local repos up-to-date with remote.

### LINKED ISSUE #2927: git_status: Add option for "current with upstream"
`git_status` has symbols for whether the branch is ahead, behind, or diverged with the upstream. But if the branch is current with the upstream, there's no symbol for this.

This is a problem because if the branch has no upstream at all, there's also no symbol for this. So if I'm on a branch that's current, I can't tell if I pushed it to upstream or if it simply has no upstream.

For example, my old prompt shows this: `(lily/my-branch +*)`. This tells me it's got staged and unstaged changes, but crucially, that there's no upstream. Starship shows `[ lily/my-branch][!+]`, which tells me it has staged and unstaged changes, but does not tell me whether it's pushed to upstream. If I look at another branch, my old prompt says `(lily/other-branch=)` which means "status is clean, and it's current with upstream". starship says `[ lily/other-branch]` which again tells me nothing about whether an upstream exists.

Add a new option, maybe `current`, with a default value of an empty string. This option would be used in `$ahead_behind` if the branch has an upstream but is not ahead, behind, or diverged.

Nothing else comes to mind.

billrisher:
#### Description
- `ahead` and `behind` variables for `git_status` have been converted to `Option` types
- new symbol has been added for when a local branch is up-to-date with its remote counterpart (i.e. ahead by 0 commits and behind by 0 commits).
- `ahead_behind` variable mapping now utilizes a match statement.

#### Motivation and Context

It's important for there to exist a distinction between local git repositories that are tracking remotes and those that aren't. By including a symbol to differentiate the two, users are able to better understand the repository they're working with.
Closes #2927 

Clean working branch before:
image

After:
image

- Tested on Arch Linux 5.13.6-arch1-1 with 'alacritty' terminal.
- Runtime is equivalent between this branch and master.
- Certain `git_status` unit tests have had configurations added to hide the `uptodate` symbol. Unit tests where `uptodate` should _not_ appear have not been modified.

#### Checklist:


docs/config/README.md
```diff
@@ -1314,6 +1314,7 @@ current directory.
 | `ahead`      | `"⇡"`                                         | The format of `ahead`               |
 | `behind`     | `"⇣"`                                         | The format of `behind`              |
 | `diverged`   | `"⇕"`                                         | The format of `diverged`            |
+| `uptodate`   | `"✓"`                                         | The format of `uptodate`            |
```
davidkna, line 1317: I think it's better to default to an empty string to avoid changing the existing output.

src/modules/git_status.rs
```diff
@@ -46,23 +47,35 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-                            format_text(config.diverged, "git_status.diverged", |variable| {
-                                match variable {
-                                    "ahead_count" => Some(ahead.to_string()),
-                                    "behind_count" => Some(behind.to_string()),
-                                    _ => None,
+                    "ahead_behind" => {
+                        info.get_ahead_behind()
+                            .and_then(|(ahead, behind)| match (ahead, behind) {
+                                (None, _) | (_, None) => None,
+                                (Some(a), Some(b)) => {
```
davidkna, line 54: I think you can keep the changes smaller and simplify this by using the `?` operator:

```rust
match (ahead?, behind?) {
```

src/modules/git_status.rs
```diff
@@ -271,8 +284,8 @@ impl RepoStatus {
         let re = Regex::new(r"branch\.ab \+([0-9]+) \-([0-9]+)").unwrap();
 
         if let Some(caps) = re.captures(s) {
-            self.ahead = caps.get(1).unwrap().as_str().parse::<usize>().unwrap();
-            self.behind = caps.get(2).unwrap().as_str().parse::<usize>().unwrap();
+            self.ahead = Some(caps.get(1).unwrap().as_str().parse::<usize>().unwrap());
+            self.behind = Some(caps.get(2).unwrap().as_str().parse::<usize>().unwrap());
```
davidkna, line 288: You can make this safer by using the `parse` return value:
```suggestion
            self.ahead = caps.get(1).unwrap().as_str().parse::<usize>().ok();
            self.behind = caps.get(2).unwrap().as_str().parse::<usize>().ok();
```

davidkna (changes requested):

docs/config/README.md
billrisher: In the docstrings in `modules/git_status.rs` (lines 13-26), should I remove the mention of `uptodate` there since it defaults to nothing, or still leave it there with an empty string as its 'default symbol'? Thanks for your review!

davidkna: Please leave it there with an empty string.

src/configs/git_status.rs
```diff
@@ -10,6 +10,7 @@ pub struct GitStatusConfig<'a> {
     pub stashed: &'a str,
     pub ahead: &'a str,
     pub behind: &'a str,
+    pub uptodate: &'a str,
```
davidkna, line 13: I just noticed it might be better to bring this in line with the other names and call this `up_to_date`

src/modules/git_status.rs
```diff
@@ -59,6 +61,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                             format_count(config.ahead, "git_status.ahead", ahead)
                         } else if behind > 0 && ahead == 0 {
                             format_count(config.behind, "git_status.behind", behind)
+                        } else if ahead == 0 && behind == 0 {
+                            format_symbol(config.uptodate, "git_status.uptodate")
                         } else {
                             None
                         }
```
davidkna, line 68: I think else is never reached:
```suggestion
                        } else {
                            format_symbol(config.up_to_date, "git_status.up_to_date")
                        }
```

davidkna (changes requested):
I found two more small issues, but beyond that this looks good to me.

davidkna (approved):

davidkna:
Thanks for the contribution @billrisher!

PR #2959, opened 2021-08-04
feat(package): Add ability to format the version

### LINKED ISSUE #2236: Option to remove 'v' from version number

andytom:
#### Description

Have added the ability to use format the version of the package using
the `version_format` option. While doing this I have also done some
refactoring of the module to remove the if/else if/... block and replace
it with an iterator. This should make fix some edge cases where versions
are not correctly picked up due to other files an example would be a
python project that has a `pyproject.toml` file but using the
`setup.cfg` for the package version. It should also make it easier to
make the order of the list configurable in the future.

#### Motivation and Context

Related to #2236

#### Checklist:

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -1,6 +1,6 @@
 use super::{Context, Module, RootModuleConfig};
 use crate::configs::package::PackageConfig;
-use crate::formatter::StringFormatter;
+use crate::formatter::{StringFormatter, VersionFormatter};
 use crate::utils;
 
 use ini::Ini;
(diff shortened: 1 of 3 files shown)


offbyone (approved):

src/modules/package.rs
```diff
@@ -184,80 +168,88 @@ fn extract_maven_version(file_contents: &str) -> Option<String> {
...
+        get_julia_project_version,
+        get_mix_version,
+        get_helm_package_version,
+        get_maven_version,
+        get_meson_version,
+        get_vmod_version,
+        get_vpkg_version,
+    ];
+
+    package_version_fn.iter().find_map(|f| f(context, config))
```
davidkna, line 242: Is there a reason `format_version` is called in each function instead of here? Beyond this LGTM.

andytom: We do some matching on the parsed string when getting the version from the npm package.json and I couldn't come up with a way to make it work nicely.

davidkna (approved):

andytom:
Thank you for reviewing @offbyone and @davidkna

PR #2974, opened 2021-08-10
Enable reproducible builds

alerque:
I recently filed [this issue](https://github.com/baoyachi/shadow-rs/issues/61) with shadow-rs because it seemed to be the one blocked keeping starship [builds from being reproducibly](https://reproducible-builds.org/). My subsequent PR was merged a new release cut.

This PR bumps the starship dependency to the latest library version. Merging this should make future starship releases reproducible (provided there isn't something else overlooked, testing this is relatively hard until it gets to the packaging stage).

Closes #2975

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1660,9 +1660,9 @@ dependencies = [
 
 [[package]]
 name = "shadow-rs"
-version = "0.6.4"
+version = "0.6.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7781b59cf09391a3657672076913dbb5d8488a0165d6ffe4fb2909499a215d66"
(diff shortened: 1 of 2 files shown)


chipbuster:
@alerque The linked issue on shadow-rs seems to have [a comment added which points out a flaw with how `TZ` is handled.](https://github.com/baoyachi/shadow-rs/issues/61#issuecomment-896202622) Is that going to be an issue? 

If not, we can merge now. If so, I'd prefer to wait until that gets sorted.

davidkna:
@alerque Please also bump shadow-rs in the `build-dependencies` section.

alerque:
Stand by for now, that looks like it might actually be a problem that needs fixing for this to work.

alerque:
Upstream has a patch release out to fix the issue. I've bumped this and added the build-dependencies section.

PR #2987, opened 2021-08-15
feat(red,vlang): Add version formatting

### LINKED ISSUE #2236: Option to remove 'v' from version number

andytom:
#### Description

Have added version formatting to the red and vlang modules. Note the
docs for red already mentioned the `version_format` string but it had
not actually been added.

#### Motivation and Context

Closes #2236

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3044,21 +3044,22 @@ By default the module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Option              | Default                                      | Description                                  |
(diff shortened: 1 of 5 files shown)


cappyzawa (approved):

andytom:
Thanks for the review @cappyzawa

PR #2993, opened 2021-08-17
ci: disable dependabot cron on starship forks

offbyone:
#### Description

Check for the primary fork of starship and only run the merge-dependabot job on
that fork.

#### Motivation and Context

The CI workflow that merges dependabot PRs will run even on forks of starship,
which probably don't have valid GH tokens. Certainly mine doesn't!

#### Checklist:


matchai (approved):
Thanks for the fix, @offbyone!

PR #2994, opened 2021-08-18
feat(cobol): added COBOL module

### LINKED ISSUE #2377: GNUCOBOL support
Hello,

I'd like to see GNUCOBOL support get added into Starship, it would function like the other language compilers in Starship.

The version would be gotten by: 
`echo -n v && cobc -version | head -n 1 | awk '{ print $3 }'`
And it would be triggered by the detection of a .cbl/CBL or a .cob/COB file in the current folder.

I know that probably isn't the fastest way to get the version, but since I don't really know rust or do bash scripting, I'm leaving this part to the experts.

I've been getting into COBOL recently and I would like to see it added to Starship. I know that it may not be the most popular feature, but even then, if someone out there can provide a patch or similar, I'd be forever grateful.

Many thanks,
Zenithium

billrisher:
#### Description
Adds support for the COBOL / GNUCOBOL language, displaying the currently installed version in the prompt.

#### Motivation and Context

Closes #2377

COBOL Prompt

Tested on Arch Linux using the `alacritty` terminal, wrote unit tests to cover each of the different file extensions that the module detects.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -521,7 +521,7 @@ By default, the module will be shown if any of the following conditions are met:
 | `detect_extensions` | `["cbl", "cob", "CBL", "COB"]`       | Which extensions should trigger this module.                             |
 | `detect_files`      | `[]`                                 | Which filenames should trigger this module.                              |
 | `detect_folders`    | `[]`                                 | Which folders should trigger this module.                                |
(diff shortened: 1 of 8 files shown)


chipbuster:
@billrisher Thanks for your contributions!

PR #2996, opened 2021-08-21
perf(node): on windows try getting version without cmd execution

davidkna:
#### Description

If the node.js executable is located at `C:\Program Files\nodejs\node.EXE` (and the executable is likely not a shim) attempt to get the file version from the executable metadata with the Windows file version APIs instead of executing `node --version`.

This improves `nodejs` modules performance from 30-100ms to about 10ms on my system.

I used `windows` over `winapi` because `winapi` was missing some bindings, and I was having issues with the `winapi`-version of `VerQueryValueW`. It might make sense to switch to `windows` completely in the future.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -33,6 +33,30 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     }
 
     let nodejs_version = Lazy::new(|| {
+        #[cfg(all(windows, not(test)))]
+        {
+            use crate::modules::utils::version_win;
+            let resolved_node_path = match which::which("node") {
+                Ok(p) => p,
+                Err(e) => {
+                    log::error!("Failed to resolve path for node: {:?}", e);
+                    return None;
+                }
+            };
+
+            // If executable is located in known-good location, it's unlikely to be a wrapper executable
+            // thus it's possible to get the version from executable metadata
+            if Path::new(r"C:\Program Files\nodejs\node.EXE") == resolved_node_path {
+                return match version_win::from_metadata(&resolved_node_path) {
+                    Ok(v) => Some(v),
+                    Err(e) => {
+                        log::error!("Failed to get node version from executable: {:?}", e);
+                        None
+                    }
+                };
+            }
+        }
+
         context
             .exec_cmd("node", &["--version"])
             .map(|cmd| cmd.stdout)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,6 +277,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "const-sha1"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fb58b6451e8c2a812ad979ed1d83378caa5e927eef2622017a45f251457c2c9d"
+
+[[package]]
 name = "constant_time_eq"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1738,6 +1744,7 @@ dependencies = [
  "versions",
  "which",
  "winapi",
+ "windows",
  "yaml-rust",
 ]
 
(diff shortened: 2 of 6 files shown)


kidonng:
Is there any node version manager/tools that generate `.exe` shims? If not this can be expanded to all `node.exe` instances, not just the one installed by the official installer.

src/modules/nodejs.rs
```diff
@@ -33,6 +33,30 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     }
 
     let nodejs_version = Lazy::new(|| {
+        #[cfg(all(windows, not(test)))]
+        {
+            use crate::modules::utils::version_win;
+            let resolved_node_path = match which::which("node") {
+                Ok(p) => p,
+                Err(e) => {
+                    log::error!("Failed to resolve path for node: {:?}", e);
```
andytom, line 42: Maybe this shouldn't be logged as an error since this will occur if the user does not have node installed.

src/modules/nodejs.rs
```diff
@@ -33,6 +33,30 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+            };
+
+            // If executable is located in known-good location, it's unlikely to be a wrapper executable
+            // thus it's possible to get the version from executable metadata
+            if Path::new(r"C:\Program Files\nodejs\node.EXE") == resolved_node_path {
+                return match version_win::from_metadata(&resolved_node_path) {
+                    Ok(v) => Some(v),
+                    Err(e) => {
+                        log::error!("Failed to get node version from executable: {:?}", e);
+                        None
```
andytom, line 54: If we can't get the version by inspecting the exe would it be better to fall back to the normal method of by calling `node - - version` and parsing the output rather than returning none?

andytom (commented):
Have left some suggestions but this is a really cool idea. Can we do this for other programs?

davidkna:
> Is there any node version manager/tools that generate `.exe` shims? If not this can be expanded to all `node.exe` instances, not just the one installed by the official installer.

I think [volta](https://volta.sh) is using shims. In general for windows the [scoop](https://scoop.sh) package manager also uses shims. I also haven't checked if cygwin exes are compatible yet either.

> Can we do this for other programs? This could be a great performance win.

Yes, but some programs (i.e. starship) don't set this version in the metadata and the version extracted with this method might be less complete (no suffixes like `-beta`), but there's also a version string version of the version API.

vladimyr:
While I appreciate the motivation and acknowledge it is a really clever and well-executed hack this gets no :-1: from me. IMO performance gains are not big enough to justify bringing in another dependency and increasing maintenance burden. Also, treating metadata as a source of truth is a risky game. There are no guarantees about metadata availability and/or correctness.

vladimyr:
Also to further strengthen my point this relies on File Version ([`dwFileVersion*`](https://docs.microsoft.com/en-us/windows/win32/api/VerRsrc/ns-verrsrc-vs_fixedfileinfo)) but there is also a Product Version ([`dwProductVersion*`](https://docs.microsoft.com/en-us/windows/win32/api/VerRsrc/ns-verrsrc-vs_fixedfileinfo)) attribute: https://www.bojankomazec.com/2011/07/file-version-vs-product-version.html Which one should we favor/use?

:warning: I checked both using [`pev`](https://github.com/merces/pev) and here is what I got:
image

vladimyr:
> > Can we do this for other programs? This could be a great performance win.
> 
> Yes, but some programs (i.e. starship) don't set this version in the metadata and the version extracted with this method might be less complete (no suffixes like `-beta`), but there's also a version string version of the version API.

Tried rust and metadata on PE (`rustc`) isn't set:
image

vladimyr:
Python product version is really funky:
image

davidkna:
> While I appreciate the motivation and acknowledge it is a really clever and well-executed hack this gets no 👎 from me. IMO performance gains are not big enough to justify bringing in another dependency and increasing maintenance burden. Also, treating metadata as a source of truth is a risky game. There are no guarantees about metadata availability and/or correctness.

At least as far as the extra dependency goes, I think down the line I could get rid of `winapi` and replace it with `windows`, but I could also make it work with `winapi` and some manual bindings if that's an issue.

> Also to further strengthen my point this relies on File Version ([`dwFileVersion*`](https://docs.microsoft.com/en-us/windows/win32/api/VerRsrc/ns-verrsrc-vs_fixedfileinfo)) but there is also a Product Version ([`dwProductVersion*`](https://docs.microsoft.com/en-us/windows/win32/api/VerRsrc/ns-verrsrc-vs_fixedfileinfo)) attribute: https://www.bojankomazec.com/2011/07/file-version-vs-product-version.html
> Which one should we favor/use?

It looks like product version would be the better choice here. But it's interesting that the output in starship and the test are not broken even though the use the file version. Maybe `pev` is using the string API?

> Python product version is really funky:

It might be possible to create a mapping from the product version to the real version for python. The `winget` repo has the product version combined with the download url (which includes the real version) for quite a few python versions.

Also, for a future caching mechanism presence of version metadata also seems to be a good indicator of whether an executable is a shim and it could also be helpful to detect version changes of executables.

Milo123459:
Would support for Scoop be added? It installs to the home directory and is

davidkna:
> Would support for Scoop be added?

Not as part of this PR, but it's something that will be addressed in the future. Implementation-wise, getting the true executable path with scoop is fairly simple.

Milo123459:
> > Would support for Scoop be added?
> 
> Not as part of this PR, but it's something that will be addressed in the future. Implementation-wise, getting the true executable path with scoop is fairly simple.

Simple enough. I might make a PR for it.

PR #2997, opened 2021-08-21
refactor: allow passing OsStr as-is to `exec_cmd`

davidkna:
#### Description

This PR changes `utils::exec_cmd` and `context.exec_cmd` to accept `AsRef<OsStr>` (this includes `&str`) instead of `&str` like `Command` from `std`.
This allows passing the git folder in the `-C` argument without `.to_string_lossy()` (which wouldn't work in the lossy case, but this fix still doesn't seem to fix the issue).

#### Motivation and Context

Closes #

#### Checklist:


andytom (approved):
LGTM

PR #3014, opened 2021-08-26
feat: conditional style based on enviroment variables

### LINKED ISSUE #607: Allow environment variables in config
## Feature Request

The use-case I have in mind for this is admittedly pretty niche, but I'm guessing there are many other possibilities as well. At my work, we're regularly SSH'ing into various servers, each named after a color. Sometimes it's hard to quickly remember which server you're connected to, so it'd be awesome to be able to change the color of the hostname in the prompt depending on the system. These servers all share a network drive, so I can't easily configure one `~/.config/starship.toml` differently from another.

It might be nice to be able to use environment variables in the `starship.toml` with something like the following:

```toml
[hostname]
ssh_only = false
prefix = "⟪"
suffix = "⟫"
style = "bold dimmed ${COLOR}"
```

Where `${COLOR}` is replaced by the `COLOR` environment variable.

I'd definitely be up for implementing this after it's discussed a bit! Is there a better solution? Better format?

### LINKED ISSUE #2784: Add style_root/style_user for Directory module
Seeing the cwd in a different colour depending on user/root would make it much easier to see who you were at a glance.

Add the `style_user` and `style_root` options to the `Directory` module, just like for the `Username` module.

Could make two convoluted custom commands based on `$PWD` and `$USER` or similar
and use them instead of `Directory` module, for example a very simple way might be -

```
[custom.usercwd]
when = '[ "$USER" != "root" ]'
command = '[ "$PWD" == "$HOME" ] && echo -n "~" || echo -n "$PWD"'
style = "bright-cyan"
format = " [$output]($style)"
```

```
[custom.rootcwd]
when = '[ "$USER" == "root" ]'
command = '[ "$PWD" == "$HOME" ] && echo -n "~" || echo -n "$PWD"'
style = "bright-red"
format = " [$output]($style)"
```

The problem with this is that it is convoluted, there is code duplication and it is not as fast as a native implemenation. And this is just a simple example. To replicate the entire `Directory` module would be irksome and unwieldy.

cymruu:
Proposition of implementation of conditional style based on environmental variables. I'll work on docs after it's confirmed it can be merged 🙂 
#### Description
Allows to configure starship to use different styles based on environment variables. 

#### Motivation and Context
Closes #607
Closes #2784

"directory" module can have different style, depending on some environmental variable 
image

#### Checklist:


.github/config-schema.json
```diff
@@ -338,7 +338,14 @@
...
         "repo_root_style": null,
-        "style": "cyan bold",
+        "style": [
+          {
+            "env": null,
+            "expected_value": null,
+            "operator": null,
+            "style": "cyan bold"
+          }
+        ],
```
rogeruiz, line 348: Interesting approach here. Is there a better way for this to be written though? I don't know how I feel about `style.style` for the keys here. But I also haven't taken a look at everything in this PR yet. 

I did find this PR because I was thinking of contributing this but I was considering using a `when =` or something that feels more Starship. I do think this looks great though. I haven't thought my solution through yet.

rogeruiz: Maybe it'd be good to have both styles? Like if it's a string it gets parsed the old way, and if it's a "list of objects?" then it could use all those. But again, mostly just thinking out loud here.

cymruu: @rogeruiz thank you for your comments! I agree the names I've chosen aren't best. I'm happy to change them if anyone has any suggestions. I just wanted it to get it reviewed quickly to verify if my implementation can be merged 😄 

>Maybe it'd be good to have both styles? Like if it's a string it gets parsed the old way, and if it's a "list of objects?" then it could use all those. But again, mostly just thinking out loud here.

It's already working that way. You can use a plain string expressions as a `style` value. Here is a test for it https://github.com/cymruu/starship/blob/eb28f7f728fcd8e594cabd4ab0ebc93a190d2c2b/src/conditional_style.rs#L181

src/conditional_style.rs
```diff
@@ -0,0 +1,258 @@
...
+            }))
+        );
+    }
+
+    #[test]
+    fn should_deserialize_from_table_value() {
+        let config = toml::toml! {
+            env = "HOSTNAME"
+            operator = "equal"
+            expected_value = "home"
```
rogeruiz, line 203: What if rather than all this, it just sticks to the API convention of using `when = ''' <script> '''` rather than these three values?

I think it would be a much better approach and keep the mechanics of other ways things are displayed on a condition. 

I expect this would mean that users would need to write their own small scripts that exit cleanly or naw, but that would be okay as those small scripts would be more useful outside of Starship.

.github/config-schema.json
```diff
@@ -338,7 +338,14 @@
         "read_only_style": "red",
         "repo_root_format": "[$before_root_path]($style)[$repo_root]($repo_root_style)[$path]($style)[$read_only]($read_only_style) ",
         "repo_root_style": null,
-        "style": "cyan bold",
+        "style": [
+          {
+            "env": null,
+            "expected_value": null,
+            "operator": null,
```
rogeruiz, line 345:
```suggestion
            "when": null,
```

rogeruiz: I think it would be better to do this. See my other other comment above.

rogeruiz (commented):
@cymruu Thanks for reply to my feedback. I like this. I'm on mobile so I'm leaving this as a comment feedback. But what do you think of this refactor?

Also I think documentation would be needed here to get this through to the finish line.

.github/config-schema.json
cymruu: @rogeruiz If I understand correctly you are suggesting to evaluate shell command and apply style accordingly basing on the command output? (similarly to what custom module does) 

The solution you're proposing is in fact simpler, however it is not platform independent. Starship runs on many different shells and for each of them a script can have different syntax. Thus starship's configurations using conditional-style, won't be applicable for all setups.  For example a simple `when` script to compare a value of some env variable with some string is implemented differently on for `bash` and `powershell`

rogeruiz: Hey @cymruu good call on that. I think that could be solved with the way custom commands are implemented. Because this seems like a custom style based on some external thing being true or not. 

[=> custom command docs](https://starship.rs/config/#custom-commands)

From my understanding then maybe it should be up to the user to define all these caveats? Perhaps the non-string version of style should take a map of the `os`, `when`, and the `format`?

I know that seems complex, but I think the level of customization you're looking looking to achieve here should require that many granular controls. 

Perhaps something like this? It feels like a lot of repetition and I agree that `when` commands aren't portable.

```suggestion
    {
        os = "linux",
        when = """ true """,
        custom_style = "bold dimmed"
    },
    {
        os = "windows",
        when = """ true """,
        custom_style = "bold dimmed"
    },
```

rogeruiz: Also sorry that patch won't resolve properly. I was trying to get my point across quickly. I will need to take a look at the branch locally to see if what I'm saying makes sense.

PR #3018, opened 2021-08-28
feat(module): added iterm2_mark module

expnn:
#### Description

Adds support to display the iTerm2's marker, so that the users can navigate marks with Cmd-Shift-Up and Down-arrow keys.

#### Motivation and Context

This feature makes it easier for iTerm2 users on macOS to jump between prompts.
For non-iTerm2 users, this module displays nothing. 
The module automatically detects whether the user uses iTerm2 according to the environment variables ITERM_PROFILE, TERM_PROGRAM, and then decides whether to display the marker. 

image

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1660,6 +1660,7 @@ dependencies = [
  "shell-words",
  "starship-battery",
  "starship_module_config_derive",
+ "strip-ansi-escapes",
  "strsim",
  "sys-info",
  "tempfile",
@@ -1707,6 +1708,15 @@ name = "static_assertions"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
+
+[[package]]
+name = "strip-ansi-escapes"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "011cbb39cf7c1f62871aea3cc46e5817b0937b49e9447370c93cacbe93a766d8"
+dependencies = [
+ "vte",
+]
 
 [[package]]
 name = "strsim"
@@ -1961,6 +1971,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7cf7d77f457ef8dfa11e4cd5933c5ddb5dc52a94664071951219a97710f0a32b"
 
 [[package]]
+name = "utf8parse"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "936e4b492acfd135421d8dca4b1aa80a7bfc26e702ef3af710e0752684df5372"
+
+[[package]]
 name = "vcpkg"
 version = "0.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


chipbuster:
A few questions about this module overall, since I'm not terribly familiar with the iTerm2 mark: 

- Does this replicate the functionality provided by the current iTerm2 hooks?
- Does this add any functionality not currently possible with the default iTerm2 hooks?
- Would this require any special intervention from current iTerm2 users? What about current users of iTerm2 who are trying out starship?
- Would this require any special intervention from the maintainers of the existing iTerm2 integration script?

gnachman:
Hi! iTerm2 maintainer here. I found this because in [this issue](https://gitlab.com/gnachman/iterm2/-/issues/10537#note_1069174232) a user tried to hack this and failed. This PR is pretty old and I don't know Rust but I'd love to help move this forward.

> Does this replicate the functionality provided by the current iTerm2 hooks?

iTerm2 provides a bash/zsh/tcsh/fish script you can load that implements five distinct features (start of prompt, end of prompt, start of output, command status at termination, and reporting current hostname and directory). This module affects the "start of prompt" feature only. It's kinda useful alone but really needs the other features to be complete.

That said, it solves an otherwise unsolvable problem which is to correctly move the prompt mark to the second line of the multi-line prompt.

> Does this add any functionality not currently possible with the default iTerm2 hooks?

Yeah, see the aforelinked issue. Moving the mark to the second line seems to be impossible because `format` can't take nonprinting characters. Changing `format` might be a simpler way to solve this problem, though.

> Would this require any special intervention from current iTerm2 users? What about current users of iTerm2 who are trying out starship?

It shouldn't. Loading the module will move the mark to the second line is all.

> Would this require any special intervention from the maintainers of the existing iTerm2 integration script?

I don't expect so but I'm happy to make changes to support starship. We have many users in common.

gnachman:
Friendly ping

andrewgazelka:
@gnachman I am going to try to get this through in #5381

andrewgazelka:
> ❯

Hmm this still has issues for me (see extra line)

image


image

crabique:
For the trigger to work, you need to `export ITERM2_SQUELCH_MARK=1` before sourcing the integration script. Also do not inject the prompt mark anywhere, iTerm will do it for you based on the trigger. 

In order to reduce false prompt detections, I suggest adding the trigger as `^❯[[:blank:]]+`.

andrewgazelka:
> ^❯[[:blank:]]+

Interesting. I still get weird behavior when presing backspace I get an extra blank line.

crabique:
> Interesting. I still get weird behavior when presing backspace I get an extra blank line.

Does ⇧⌘↑ highlight those blank lines? If yes, you still inject the prompt somewhere (.zshrc or your starship config).

andrewgazelka:
> > Interesting. I still get weird behavior when presing backspace I get an extra blank line.
> 
> Does ⇧⌘↑ highlight those blank lines? If yes, you still inject the prompt somewhere (.zshrc or your starship config).

It does not highlight blank lines.

image

Also the new line is added seemingly "randomly" while I am typing.

andrewgazelka:
It appears pressing opt-backspace after I typed a very long command always adds the newline 🤔

crabique:
Doesn't happen to me, could be something under `Profiles > Terminal > Shell Integration`?
image

andrewgazelka:
> Doesn't happen to me, could be something under `Profiles > Terminal > Shell Integration`? image

hmmmmm I reset my iTerm preferences too and still get it... so weird.

andrewgazelka:
> Doesn't happen to me, could be something under `Profiles > Terminal > Shell Integration`? image

Do you mind sending/linking your iTerm preferences?

wday-cs:
I simply have this in the trigger: ` ❯` that is `[space]❯`. The trigger needs to Instant as well otherwise it'll require the new line to be detected. @andrewgazelka 

I have not set the `ITERM2_SQUELCH_MARK` variable @crabique

andrewgazelka:
> ❯

Confused as to why your pattern is ` ❯` given there is no space (at least before my prompt with default settings)

image

andrewgazelka:
mark this as closed since no progress on it?

PR #3019, opened 2021-08-29
docs: Correct the "VLang" configuration example

machupicchubeta:
#### Description

Correct the "VLang" configuration example.
This commit is a document correction only; to avoid the configuration error.
The documents to be corrected are "docs/[lang]/README.md".

#### Motivation and Context

With a config key of `[v]` on macOS, it has the following error.
```
[WARN] - (starship::configs::starship_root): Unknown config key 'v'
```
I think it might be the error in other OS as well.

In the other document "docs/[lang]/presets/README.md", the config key is `[vlang]`.

#### Checklist:

### DIFF
--- a/docs/ar-SA/config/README.md
+++ b/docs/ar-SA/config/README.md
@@ -2925,7 +2925,7 @@ The `vlang` module shows you your currently installed version of V. By default t
 
 ```toml
 # ~/.config/starship.toml
-[vlang]
+[v]
 format = "via [V $version](blue bold) "
 ```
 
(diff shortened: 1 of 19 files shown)


vladimyr:
Thanks but please revert changes to translated pages, they are done automatically. You should only ever touch English docs and it gets propagated to translated versions.

machupicchubeta:
@vladimyr 
Thank you for the advice. I added a commit to revert changes made to non-English pages.
I would appreciate it if you could tell me if I should combine the commits into one.

vladimyr:
Thanks for addressing it. It's fine, it'll get squashed on merge.

vladimyr (approved):
LGTM

davidkna:
Thanks for the contribution @machupicchubeta and thanks for reviewing @vladimyr!

machupicchubeta:
@vladimyr 
Thank you for your review and merge!

PR #3026, opened 2021-08-31
feat: Add support for RPROMPT (right prompt)

mjeffryes:
#### Description
Many shells support a right prompt that appears on the far right hand side of the input line. This change makes it easy to configure a right prompt with starship.  

This change adds a `right_format` to the config for specifying the format of the right prompt. Modules that are included explicitly in `format` or `right_format` are removed from `$all` to avoid redundant output. A new clap flag `--right` tells starship to output the right prompt instead of the default (left) prompt. 

I've updated the init scripts for fish, zsh and elvish, but only tested with zsh so far. 

#### Motivation and Context
While it is possible to set the right prompt using starship today, it is cumbersome. The user must create an alternate config file and manually configure their shell to use it. What's more, the user would likely have to modify their main config format in order to move modules to the right prompt.

This feature seems to be a recurring request:
https://github.com/starship/starship/issues/2925
https://github.com/starship/starship/issues/387

There's even a pull request that shipped that canonizes the workaround for xonsh:
https://github.com/anki-code/xontrib-prompt-starship/pull/3
This change would allow users of that workaround to simplify their configuration.

Example config:
Screen Shot 2021-08-31 at 3 36 45 PM
Result:
Screen Shot 2021-08-31 at 3 36 26 PM

Tested in zsh, but can easily be added to the init script for any shell that supports right prompts.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -151,12 +151,14 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Option            | Default                        | Description                                                  |
-| ----------------- | ------------------------------ | ------------------------------------------------------------ |
-| `format`          | [link](#default-prompt-format) | Configure the format of the prompt.                          |
-| `scan_timeout`    | `30`                           | Timeout for starship to scan files (in milliseconds).        |
-| `command_timeout` | `500`                          | Timeout for commands executed by starship (in milliseconds). |
-| `add_newline`     | `true`                         | Inserts blank line between shell prompts.                    |
+| Option            | Default                         | Description                                                      |
+| ----------------- | ------------------------------- | ---------------------------------------------------------------- |
+| `format`          | [link](#default-prompt-format)  | Configure the format of the prompt.                              |
+| `right_format`    | `""`                            | See [Enable Right Prompt](/advanced-config/#enable-right-prompt) |
+| `scan_timeout`    | `30`                            | Timeout for starship to scan files (in milliseconds).            |
+| `command_timeout` | `500`                           | Timeout for commands executed by starship (in milliseconds).     |
+| `add_newline`     | `true`                          | Inserts blank line between shell prompts.                        |
+
 
 ### Example
 
@@ -245,6 +247,14 @@ $time\
 $status\
 $shell\
 $character"""
+```
+
+If you just want to extend the default format, you can use `$all`;
+modules you explicitly add to the format will not be duplicated. Eg.
+
+```toml
+# Move the directory to the second line
+format="$all$directory$character"
 ```
 
 ## AWS
--- a/src/init/starship.elv
+++ b/src/init/starship.elv
@@ -11,6 +11,7 @@ fn starship-after-readline-hook [line]{
 
 fn starship-before-readline-hook {
     cmd-end-time = (::STARSHIP:: time)
+    cmd-duration=(- $cmd-end-time $cmd-start-time)
 }
 
 # Install Hooks
@@ -25,9 +26,17 @@ edit:prompt = {
     if (== $cmd-start-time 0) {
         ::STARSHIP:: prompt --jobs=$num-bg-jobs
     } else {
-        ::STARSHIP:: prompt --jobs=$num-bg-jobs --cmd-duration=(- $cmd-end-time $cmd-start-time)
+        ::STARSHIP:: prompt --jobs=$num-bg-jobs --cmd-duration=$cmd-duration
     }
 }
 
-# Get rid of default rprompt
-edit:rprompt = { }
+edit:rprompt = {
+    # Note:
+    # Elvish does not appear to support exit status codes (--status)
+
+    if (== $cmd-start-time 0) {
+        ::STARSHIP:: prompt --right --jobs=$num-bg-jobs
+    } else {
+        ::STARSHIP:: prompt --right --jobs=$num-bg-jobs --cmd-duration=$cmd-duration
+    }
+}
--- a/src/init/starship.elv
+++ b/src/init/starship.elv
@@ -11,7 +11,7 @@ fn starship-after-readline-hook [line]{
 
 fn starship-before-readline-hook {
     cmd-end-time = (::STARSHIP:: time)
-    cmd-duration=(- $cmd-end-time $cmd-start-time)
+    cmd-duration = (- $cmd-end-time $cmd-start-time)
 }
 
 # Install Hooks
--- a/src/init/starship.elv
+++ b/src/init/starship.elv
@@ -4,6 +4,7 @@ set-env STARSHIP_SESSION_KEY (::STARSHIP:: session)
 # Define Hooks
 local:cmd-start-time = 0
 local:cmd-end-time = 0
+local:cmd-duration = 0
 
 fn starship-after-readline-hook [line]{
     cmd-start-time = (::STARSHIP:: time)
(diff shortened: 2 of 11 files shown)


matchai:
Hey @mjeffryes. Thank you for the PR! 😊 
I really appreciate you taking the time to update the various shells to add right-prompt support. Unfortunately, reaching for each shell's built-in right-prompt support does limit the number of shells that starship would support `right_format` for.
I think that, ideally, Starship would handle right-prompt support across all shells using the "fill" solution you wrote up in #3029, allowing us to have the same consistent behavior across all shells, whether or not they natively support right-prompt.

Can you think of any way that a shell's native right-prompt support would be preferable to a left-prompt using padding to act as a right-prompt too?
The main one that comes to mind is terminal resizing being handled elegantly, re-rendering the right-prompt when the terminal width is changed. Is that reason enough to defer right-prompt handling to the shell? 🤔 

The best solution may be to use the shell's built-in right-prompt support when available and also implementing a generic right-prompt handler for shells that don't support right-prompts.

mjeffryes:
That's a fair question. 

One obvious difference is that it just allows for more compact layouts (eg. [typewritten](https://github.com/reobin/typewritten) keeps the whole prompt on one while keeping the input aligned on the left). There are also some cases where people are using the rprompt in addition to right aligned sections in the main prompt, though that seems less common. (I scanned though the oh-my-zsh [themes](https://github.com/ohmyzsh/ohmyzsh/wiki/Themes) and close to half seem to use the right prompt, most for a single line prompt. I only saw two, bureau and jonathan using right aligned main prompt sections + rprompt.)

I think it mostly comes down to a judgement call based on the project's goals. If we want to allow users to fully customize their prompt to look like any of the popular themes, we need a way to set the right prompt on shells that support it. If the goal is just to provide a nice default with some styling options, we might exclude it. 

My personal inclination would be to support right prompts, since it's not a lot of overhead and completely optional if users want to enable it. (Plus I'm partial to one line prompt styles.)

(I do like the idea of supplying a generic right-prompt handler for shells that don't support it natively, eg. this seems to work for bash: https://mina86.com/2015/bash-right-prompt/)

matchai:
> One obvious difference is that it just allows for more compact layouts (eg. [typewritten](https://github.com/reobin/typewritten) keeps the whole prompt on one while keeping the input aligned on the left).

Ah right! I was just thinking about two-line right-prompt, but single-line right-prompts are certainly well worth supporting.
Nice find with the bash right-prompt! If it works well, we can definitely include it in our bash init script. 👍 
Unless we manage to add support for right-prompts on most of the supported shells, I think that this may best be documented under "Advanced Configuration".

I can foresee this causing some confusion if `fill` ends up being included as a module. Both serve as ways to have right-aligned elements in the prompt – how do we best users to reaching for the right option?

docs/config/README.md
```diff
@@ -247,6 +254,14 @@ $shell\
 $character"""
 ```
 
+If you just want to extend the default format, you can use `$all`;
```
matchai, line 257: Love this inclusion! 😍

src/init/starship.elv
```diff
@@ -29,5 +29,13 @@ edit:prompt = {
...
-# Get rid of default rprompt
-edit:rprompt = { }
+edit:rprompt = {
+    # Note:
+    # Elvish does not appear to support exit status codes (--status)
+
+    if (== $cmd-start-time 0) {
+        ::STARSHIP:: prompt --right --jobs=$num-bg-jobs
+    } else {
+        ::STARSHIP:: prompt --right --jobs=$num-bg-jobs --cmd-duration=(- $cmd-end-time $cmd-start-time)
```
matchai, line 39: **nitpick:** Instead of evaluating `cmd-duration` once in each prompt, it may be best to evaluate it, store it as a variable, and pass it to both `starship prompt` commands.

src/init/starship.fish
```diff
@@ -11,6 +11,19 @@ function fish_prompt
...
+    switch "$fish_key_bindings"
+        case fish_hybrid_key_bindings fish_vi_key_bindings
+            set STARSHIP_KEYMAP "$fish_bind_mode"
+        case '*'
+            set STARSHIP_KEYMAP insert
+    end
+    set STARSHIP_CMD_STATUS $status
+    # Account for changes in variable name between v2.7 and v3.0
+    set STARSHIP_DURATION "$CMD_DURATION$cmd_duration"
+    ::STARSHIP:: prompt --right --status=$STARSHIP_CMD_STATUS --pipestatus=$pipestatus --keymap=$STARSHIP_KEYMAP --cmd-duration=$STARSHIP_DURATION --jobs=(count (jobs -p))
```
matchai, line 24: Similarly here, `$fish_key_bindings` shouldn't have changed, and we can avoid spawning `count (jobs -p)` a second time.

src/print.rs
```diff
@@ -403,3 +395,66 @@ pub fn format_duration(duration: &Duration) -> String {
         format!("{:?}ms", &milis)
     }
 }
+
+// Return the modules from $all that are not already in the list
```
matchai, line 399: **nitpick:** triple-slashes to allow editors to show the description as part of code completion.

```suggestion
/// Return the modules from $all that are not already in the list
```

src/print.rs
```diff
@@ -403,3 +395,66 @@ pub fn format_duration(duration: &Duration) -> String {
...
+    for module in PROMPT_ORDER.iter() {
+        if !module_list.contains(*module) {
+            prompt_order.push(String::from(*module))
+        }
+    }
+
+    prompt_order
+}
+
+fn load_formatter_and_modules<'a>(context: &'a Context) -> (StringFormatter<'a>, BTreeSet<String>) {
```
matchai, line 411: Love that this is broken out into a separate function! 👍 
Could you please add a descriptive comment for this function?

chipbuster:
I feel like this discussion has already been hashed out above, but apparently I'm slow at reading, so I wanted to add my own summary of the key difference I see.

Fill modules (like proposed on #3029) allow you to right-align things in two-line prompts, but don't allow you to put things after the cursor. If you imagine manually typing out the prompt with a typewriter that lets you put down color codes + emoji, you're allowed to press any button you want *except* for Backspace. All of this text shows up *before* the default cursor location.

Shell-native RPROMPT support allows us to put text *after* the cursor's default location. That is, it's like if you typed part of the prompt, hit space a bunch, typed out the RPROMPT, and then hit Backspace until you got back to the starting cursor location. If you type a very long command, the command text can overrun the RPROMPT text.

We cannot emulate native RPROMPT support in starship without issuing terminal control sequences that affect the cursor location (which is a tricky move given how long it took us to be mostly-free of cursor positioning bugs).

matchai:
I that's a solid roundup, @chipbuster. I think we may need both features in our docs to link to one another.
I can imagine users saying "I want the time to appear on the right of my prompt", searching for "right align" in our docs, and ending up at the `right_format` rather than `fill`.

mjeffryes:
I've rebased off master and addressed the review comments from @matchai.  

I agree that it's probably best to treat right prompt as advanced config and moved the documentation there. (With a link to the hopefully-soon-to-be-merged fill module docs in case someone looking to right align a section of their left prompt winds up there.)

src/init/starship.fish
mjeffryes: Done, although it does rely on the left prompt being executed first. (Which seems to be true in my testing)

matchai (approved):
Looks good to me. Thank you for tackling this PR @mjeffryes! 😊 
I tested it locally with fish, zsh, and elvish, and everything seems to work perfectly 👍

PR #3032, opened 2021-09-03
fix(config): make terminal editors work again when launched by `starship config`

alexmaco:
Due to the introduction of `utils::create_command`, commands now have stdin set to null, and stdout and stderr set to piped.
This prevented console editors from working when invoked via `starship config`

#### Description
This just changes the io streams of the editor command back to inherit from piped/null.

#### Motivation and Context
In starship 0.56 `starship config` worked fine with vim
In starship 0.57 `starship config` does not display the editor UI in the terminal.

#### Checklist:

### DIFF
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -2,6 +2,7 @@ use std::env;
 use std::ffi::OsString;
 use std::io::ErrorKind;
 use std::process;
+use std::process::Stdio;
 
 use crate::config::RootModuleConfig;
 use crate::config::StarshipConfig;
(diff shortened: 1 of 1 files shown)


vladimyr (approved):
LGTM

mattn:
This change seems breaking `starship configure`. Probably, stdout should not be inherit. I got fix my issue with previous version. So this change break something.

mattn:
Ah, sorry. I was wrong. This change fixes my issue.

PR #3033, opened 2021-09-04
feat(juju): Add Juju module

knkski:
#### Description

Adds support for https://juju.is/

Displays version information as well as active controller/model.

#### Motivation and Context

The motivation is that it is useful to see what Juju version and/or active controller you have

image

Tested by adding unit tests, as well as extensive dogfooding.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1614,6 +1614,40 @@ number_threshold = 4
 symbol_threshold = 0
 ```
 
+## Juju
+
+The `juju` module shows the currently installed version of [Juju](https://juju.is/),
+along with the active controller and model, if set.
+
+### Options
+
+| Option     | Default                                  | Description                                      |
+| ---------- | ---------------------------------------- | ------------------------------------------------ |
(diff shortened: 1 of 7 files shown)


docs/config/README.md
```diff
@@ -1614,6 +1614,40 @@ number_threshold = 4
...
+
+\*: This variable can only be used as a part of a style string
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[juju]
+disabled = false
```
Milo123459, line 1648:
```suggestion
symbol = "🐼"
```

In my honest opinion, setting it to disabled in the example is a bad idea. Changing the symbol is something commonly used in other modules.

knkski: Sure, updated :+1:

chipbuster:
Looks like this is failing a test (specifically, it looks like `src/configs/mod.rs` hasn't been completely updated for the new config).

knkski:
@chipbuster whoops, thanks for pointing that out, fixed

PR #3037, opened 2021-09-06
perf: increase optimization level on release builds

thedrow:
#### Description

This PR increases the optimization level to 3 at the expense of build time (although with caching this is greatly mitigated).

#### Motivation and Context

Since we're using starship in the terminal, it can be useful to have a faster binary that can allow us to work more quickly.

#### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,6 +99,7 @@ mockall = "0.10"
 [profile.release]
 codegen-units = 1
 lto = true
+opt-level = 3
 
 [[bin]]
 name = "starship"


davidkna:
Thanks for the PR, but I believe `opt-level=3` is already set in the [default `release` profile](https://doc.rust-lang.org/cargo/reference/profiles.html#release).

Because of this means setting it to 3 won't change the optimisation level, I'll close this PR. Please re-open if you think I was mistaken.

thedrow:
Thanks!
I wasn't aware this is already the case.

PR #3042, opened 2021-09-07
feat(explain): allow specifying prompt arguments in explain and timings subcommands

### LINKED ISSUE #3040: `starship explain` does not take the flags that `starship prompt` does
#### Current Behavior
`starship prompt` takes a set of flags to gather information from the shell (such as status, pipestatus, keymap, etc). `starship explain` does not take these flags, meaning it's impossible to ask starship to explain the particular formatting of any modules which require it.

#### Expected Behavior
`starship explain` should take the same flags, allowing me to replicate my prompt.

Additionally (and optionally), it would be neat if the prompt function could record the flags last given to `starship prompt` in an env var such that a subsequent call to `starship explain` could read those flags. This would allow a bare `starship explain` to explain the most recent prompt. This would actually be really useful as it would avoid the user needing to have to figure out what the right flags are to reproduce their most recent prompt. In this scenario any flags explicitly given to `starship explain` would take precedence over those from the env var.

#### Environment
- Starship version: 0.57.0
- Shell type: fish
- Shell version: fish, version 3.3.1
- Shell plugin manager: none
- Terminal emulator: macOS Terminal.app
- Operating system: macOS 11.5.2

davidkna:
#### Description

This PR allows adding the `prompt` arguments to the `explain` and `timings` subcommands. The `explain` and `timings` implementations are already handling the additional arguments correctly without any changes.

The reason for the big diff is `cargo fmt`.

#### Motivation and Context

Closes #3040

```
starship.exe-explain 
Explains the currently showing modules

USAGE:
    starship.exe explain [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -d, --cmd-duration <CMD_DURATION>    The execution duration of the last command, in milliseconds
    -j, --jobs <JOBS>                    The number of currently running jobs
    -k, --keymap <KEYMAP>                The keymap of fish/zsh
    -P, --logical-path <LOGICAL_PATH>    The logical path that the prompt should render for. This path should be a
                                         virtual/logical representation of the PATH argument.
    -p, --path <PATH>                    The path that the prompt should render for.
        --pipestatus <PIPESTATUS>...     Status codes from a command pipeline
    -s, --status <STATUS_CODE>           The status code of the previously run command
```

#### Checklist:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -81,126 +81,143 @@ fn main() {
         .help("Print the main initialization script (as opposed to the init stub)");
 
     let long_version = crate::shadow::clap_version();
-    let mut app = App::new("starship")
-        .about("The cross-shell prompt for astronauts. ☄🌌️")
-        // pull the version number from Cargo.toml
-        .version(shadow::PKG_VERSION)
-        .long_version(long_version.as_str())
-        // pull the authors from Cargo.toml
-        .author(crate_authors!())
-        .after_help("https://github.com/starship/starship")
-        .setting(AppSettings::SubcommandRequiredElseHelp)
-        .subcommand(
-            SubCommand::with_name("init")
-                .about("Prints the shell function used to execute starship")
-                .arg(&shell_arg)
-                .arg(&init_scripts_arg),
-        )
-        .subcommand(
-            SubCommand::with_name("prompt")
-                .about("Prints the full starship prompt")
-                .arg(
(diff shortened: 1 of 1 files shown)


chipbuster (approved):
LGTM :+1:

Shame about cargo fmt.

PR #3049, opened 2021-09-09
fix(zsh init script): avoid adding preexec commands twice

mjeffryes:
#### Description
A typo in the zsh init script was breaking the predicate to prevent adding duplicate hooks.

#### Motivation and Context
Minor nuisance, but it was gumming up my shell while testing 

#### Checklist:


chipbuster (approved):
D'oh.

Thanks for the catch!

PR #2890, opened 2021-07-15
feat(package): Support package version from setup.cfg (python).

FollowTheProcess:
#### Description

Add an additional package version extraction function to parse
the 'version' attribute under the 'metadata' section in a python
package 'setup.cfg' file.

The implementation was largely based on the very similar
`extract_poetry_version` function.

Also add similar tests from the poetry extraction function to test
the desired behaviour.

#### Motivation and Context

This feature would enable starship to display the package version from a python package
`setup.cfg` file (the defacto standard).

Closes #2886 

#### How Has This Been Tested?

I followed the contributing guidelines and ran

`cargo clippy --all-targets --all-features`

and ran `cargo test`

I added two additional tests to cover the new function I added which are based heavily on the tests for
the poetry equivalent function, which are both passing locally.

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly. 
- [x] I have updated the tests accordingly.

I think the documentation will need updating to include a mention to `setup.cfg` in the packages module which I am more than happy to do, just want to submit this and get someone who knows way more about rust than I do to check it over (I am very new to rust).

Thanks!

PR #3054, opened 2021-09-11
fix(package): Ignore setup.cfg attr: and file:

twm:
#### Description

The setup.cfg version key allows [`attr:` and `file:` directives][1] to load the actual value from elsewhere. Treat these as if no version was found, since the text of the directive isn't the version number we're after.

[1]: https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#specifying-values

#### Motivation and Context

Parsing of `setup.cfg` was introduced in #2890.

I added tests covering the `attr:` and `file:` directives. I ran `cargo test` to verify that they failed without implementation changes.
I then changed the implementation and ran Clippy and `cargo test`.

#### Checklist:

I'm not sure if a documentation update is required.

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -84,7 +84,11 @@ fn get_setup_cfg_version(context: &Context, config: &PackageConfig) -> Option<St
     let ini = Ini::load_from_str(&file_contents).ok()?;
     let raw_version = ini.get_from(Some("metadata"), "version")?;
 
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM

chipbuster:
@twm  Thanks for the fix!

twm:
@chipbuster Thanks for Starship! :rocket:

PR #3055, opened 2021-09-12
feat(pulumi) Add pulumi module

iwahbe:
#### Description
Adds a module for [Pulumi](https://github.com/pulumi/pulumi). The module can display:
- The current stack
- The Pulumi CLI version. 

#### Motivation and Context

It is useful to see what stack your commands are operating on (Can delete test, don't delete prod...). When hacking on or debugging pulumi, it is useful to see what version you are using. 

In addition to unit and integration tests, I have been dogfooding this feature for about a week. 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -218,6 +218,7 @@ $nodejs\
 $ocaml\
 $perl\
 $php\
+$pulumi\
 $purescript\
 $python\
 $red\
@@ -2313,6 +2314,64 @@ By default the module will be shown if any of the following conditions are met:
 
 [php]
 format = "via [🔹 $version](147 bold) "
+```
+
+## Pulumi
+
+The `pulumi` module shows the currently selected [Pulumi Stack](https://www.pulumi.com/docs/intro/concepts/stack/) and version.
+
+::: tip
+
+By default the Pulumi version is not shown, since it takes an order of magnitude longer to load then most plugins (~70ms).
+If you still want to enable it, [follow the example shown below](#with-pulumi-version).
+
+:::
+
+By default the module will be shown if any of the following conditions are met:
+
+- The current directory contains either `Pulumi.yaml` or `Pulumi.yml`
+- A parent directory contains either `Pulumi.yaml` or `Pulumi.yml`
+
+### Options
+
+| Option              | Default                              | Description                                                               |
+| ------------------- | ------------------------------------ | ------------------------------------------------------------------------- |
+| `format`            | `"via [$symbol$stack]($style) "`     | The format string for the module.                                         |
+| `version_format`    | `"v${raw}"`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
+| `symbol`            | `"🚀 "`                                | A format string shown before the Pulumi stack.                          |
+| `style`             | `"bold 5"`                           | The style for the module.                                                 |
+| `disabled`          | `false`                              | Disables the `pulumi` module.                                             |
+
+### Variables
+
+| Variable | Example    | Description                          |
+| -------- | ---------- | ------------------------------------ |
+| version  | `v0.12.24` | The version of `pulumi`              |
+| stack    | `dev`      | The current Pulumi stack             |
+| symbol   |            | Mirrors the value of option `symbol` |
+| style\*  |            | Mirrors the value of option `style`  |
+
+\*: This variable can only be used as a part of a style string
+
+### Example
+
+#### With Pulumi Version
+
+```toml
+# ~/.config/starship.toml
+
+[pulumi]
+format = "[🛥 ($version )$stack]($style) "
+```
+
+#### Without Pulumi version
+
+```toml
+# ~/.config/starship.toml
+[pulumi]
+symbol = "🛥 "
+format = "[$symbol$stack]($style) "
+
 ```
 
 ## PureScript
--- a/src/modules/pulumi.rs
+++ b/src/modules/pulumi.rs
@@ -1,0 +1,275 @@
+#![warn(missing_docs)]
+use sha1::{Digest, Sha1};
+use std::ffi::OsStr;
+use std::fs::File;
+use std::io::Read;
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+use yaml_rust::{Yaml, YamlLoader};
+
+use super::{Context, Module, RootModuleConfig};
+use crate::configs::pulumi::PulumiConfig;
+use crate::formatter::{StringFormatter, VersionFormatter};
+
+static PULUMI_HOME: &str = "PULUMI_HOME";
+
+/// Creates a module with the current Pulumi version and stack name.
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("pulumi");
+    let config = PulumiConfig::try_load(module.config);
+
+    let project_file = find_package_file(&context.logical_dir)?;
+
+    let parsed = StringFormatter::new(config.format).and_then(|formatter| {
+        formatter
+            .map_meta(|variable, _| match variable {
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "version" => {
+                    let stdout = context.exec_cmd("pulumi", &["version"])?.stdout;
+                    VersionFormatter::format_module_version(
+                        module.get_name(),
+                        parse_version(&stdout),
+                        config.version_format,
+                    )
+                }
+                .map(Ok),
+                "stack" => stack_name(&project_file, context).map(Ok),
+                _ => None,
+            })
+            .parse(None)
+    });
+
+    match parsed {
+        Ok(x) => {
+            module.set_segments(x);
+            Some(module)
+        }
+        Err(e) => {
+            log::warn!("Error in module `pulumi`:\n{}", e);
+            None
+        }
+    }
+}
+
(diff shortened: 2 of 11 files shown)


chipbuster:
@iwahbe Thanks for the PR! It looks like our clippy check isn't happy about a `&mut` in the code: you can either look at what's going on in the CI window above or [install clippy locally](https://github.com/rust-lang/rust-clippy#step-2-install-clippy) to check it out.

Looking forward to getting this merged!

iwahbe:
> @iwahbe Thanks for the PR! It looks like our clippy check isn't happy about a `&mut` in the code: you can either look at what's going on in the CI window above or [install clippy locally](https://github.com/rust-lang/rust-clippy#step-2-install-clippy) to check it out.

Sorry about that. Fixed now.

src/modules/pulumi.rs
```diff
@@ -0,0 +1,295 @@
+#![warn(missing_docs)]
+use sha1::{Digest, Sha1};
```
chipbuster, line 2: Just want to say in advance: I'm not at all familiar with pulumi, so forgive me if some of these questions are naive.

src/modules/pulumi.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+}
+
+/// Parse the output of `pulumi version` into just the version string.
+///
+/// Normally, this just means returning it. When Pulumi is being developed, it
+/// can return results like `3.12.0-alpha.1630554544+f89e9a29.dirty`, which we
+/// don't want to see. Instead we display that as `3.12.0-alpha`.
+fn parse_version(version: &str) -> &str {
+    &version[0..version
```
chipbuster, line 67: While elegant in a functional sense, this takes a while to unpack and understand. Is there a reason we can't just increment a counter until we see three `'.'` characters (or hit the end of string) and then slice up to that index?

src/modules/pulumi.rs
```diff
@@ -0,0 +1,295 @@
...
+        .trim()
+        .split('.')
+        .take(3)
+        .map(&str::len)
+        .sum::<usize>()
+        + 2]
+}
+
+/// Find a file describing a Pulumi package in the current directory (or any parrent directory).
+fn find_package_file(path: &Path) -> Option<PathBuf> {
```
chipbuster, line 77: Thanks for adding traces here.

src/modules/pulumi.rs
```diff
@@ -0,0 +1,295 @@
...
+    path.push(unique_file_name);
+    Some(path)
+}
+
+const HEXTABLE: &[char] = &[
+    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f',
+];
+
+/// Encode a u8 slice into a hexadecimal string.
+fn encode_to_hex(slice: &[u8]) -> String {
```
chipbuster, line 157: Not necessarily a blocker (we can add another PR afterwards if it would slow this down too much), but this feels general enough to be in `utils.rs` instead of hanging out here.

src/modules/pulumi.rs
```diff
@@ -0,0 +1,295 @@
...
+}
+
+/// Calculates the path of the workspace settings file for a given pulumi stack.
+fn get_pulumi_workspace(context: &Context, name: &str, project_file: &Path) -> Option<PathBuf> {
+    let project_file = if cfg!(test) {
+        // Because this depends on the absolute path of the file, it changes in
+        // each test run. We thus mock it.
+        "test".to_string()
+    } else {
+        let mut hasher = Sha1::new();
```
chipbuster, line 141: So pulumi uses a hash of the absolute path as bytes to create a custom workspace dir. What happens if the OS does not encode paths with UTF-8? Would the pulumi tool have to decode, reinterpret as UTF-8, and then re-encode?

(And perhaps the real question: is this anything we're ever going to have to deal with?)

src/modules/pulumi.rs
```diff
@@ -0,0 +1,295 @@
...
+/// the location is simple. We get it ourselves.
+fn stack_name(project_file: &Path, context: &Context) -> Option<String> {
+    let mut file = File::open(&project_file).ok()?;
+
+    let mut contents = String::new();
+    file.read_to_string(&mut contents).ok()?;
+    let name =
+        YamlLoader::load_from_str(&contents)
+            .ok()
+            .map(|yaml| -> Option<Option<String>> {
```
chipbuster, line 105:
```suggestion
    .and_then(|yaml| -> Option<String> {
            log::trace!("Parsed {:?} into yaml", project_file);
            let yaml = yaml.into_iter().next()?;
            yaml.into_hash()
                .map(|mut hash| -> Option<String> {
                    hash.remove(&Yaml::String("name".to_string()))?
                        .into_string()
                })
                .flatten()
        })?;
```

I believe this section can be rewritten like this to avoid some of the issues with triple-nesting options.

(Not convinced it's significantly better, but I do find it amusing that there's a `???` at the end of this block).

chipbuster: This is more of a cosmetic change, feel free to reject it!

src/modules/pulumi.rs
```diff
@@ -0,0 +1,295 @@
...
+                [pulumi]
+                format = "in [$symbol($stack)]($style) "
+            })
+            .env("HOME", root.to_str().unwrap())
+            .collect();
+        let expected = format!("in {} ", Color::Fixed(5).bold().paint("🚀 launch"));
+        assert_eq!(expected, rendered.expect("a result"));
+        dir.close()?;
+        Ok(())
+    }
```
chipbuster, line 294: I noticed you pushed an update that allowed you to deal with empty Pulumi configs. Is there any way to write a test for that? (E.g. by creating an empty file masquerading as a config and then running that). Or are there other problems that I haven't forseen?

chipbuster (changes requested):
Not gonna lie: I had no idea what pulumi did going into this, and after reading this PR and skimming the website, I'm still not 100% sure.

However, this looks generally pretty good. It looks like you missed adding a description for this in the `description` function (not really your fault--there's way too much to keep track of when adding a new module) and there's a few small changes/additions I'd like to see in the module itself.

Looking forward to getting this merged!

docs/config/README.md
```diff
@@ -2315,6 +2316,64 @@ By default the module will be shown if any of the following conditions are met:
...
+- The current directory contains either `Pulumi.yaml` or `Pulumi.yml`
+- A parent directory contains either `Pulumi.yaml` or `Pulumi.yml`
+
+### Options
+
+| Option              | Default                              | Description                                                               |
+| ------------------- | ------------------------------------ | ------------------------------------------------------------------------- |
+| `format`            | `"via [$symbol$stack]($style) "`     | The format string for the module.                                         |
+| `version_format`    | `"v${raw}"`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
+| `symbol`            | `"🚀 "`                                | A format string shown before the Pulumi stack.                          |
```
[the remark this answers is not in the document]
chipbuster: Is there any connection between Pulumi and a rocket? Looking on their site, the default symbol looks pretty close to a cube. I wonder if one of the nerd-font cube symbols might be a better fit.

src/modules/pulumi.rs
iwahbe: I'm happy to move it

src/modules/pulumi.rs
iwahbe: We can (and I will) write such a test. It will be included with the next commit.

src/modules/pulumi.rs
iwahbe: I find the functional style easier to read. I will change it to be explicitly imperative on my next commit.

src/modules/pulumi.rs
iwahbe: It's a good point. Pulumi doesn't deal with any encoding, that is handled by Go's [os.Getwd](https://pkg.go.dev/os#Getwd). As far as I can tell, this translates whatever it gets to utf8, but it's platform specific (In Go, a `string` is just `&[u8]`). I don't think we need to care about invalid utf8. 

If it happens, starship will fail to find the stack name, but will otherwise continue. I don't think it's work caring.

docs/config/README.md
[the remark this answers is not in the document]
iwahbe: I don't have strong opinions. I changed it to `` (which is a cube, when rendered)

src/modules/pulumi.rs
chipbuster: For functional style, I personally would have gone with a split-take-join method, but I understand not wanting to cause additional allocations. Perhaps the thing to do to make this easier to understand would have been to name the last index with its own variable?

chipbuster (approved):
Looks good to me! :+1:

chipbuster:
The hazards of pulling in additional crates.

Looks like we have a build failure in zvariant, which is a dependency of our notification feature for long-running jobs. I suspect that this is a regression on nightly, since the rules on variable size shouldn't have changed recently.

I think the simplest thing to do will be to wait a few days for `nightly` to shake itself out, then see if the issue is gone. If this hasn't been rebuilt by mid-week to see if the issue is gone, feel free to ping here to request a rerun of the CI.

chipbuster:
https://github.com/rust-lang/rust/pull/89125 got merged 8hrs ago. Hopefully the 2021-09-21 nightly builds correctly and then we'll be off to the races.

iwahbe:
Doesn't seem like. Hopefully it will be fixed it in the near future.

iwahbe:
It seems like all checks are passing now.

chipbuster:
@iwahbe Thank you for taking the time to write a PR!

PR #3057, opened 2021-09-12
ci: use an alternative caching method

Milo123459:
#### Description

This implements a system for caching dependencies.
#### Motivation and Context

Closes #

#### Checklist:

This makes the CI run faster

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -43,15 +43,8 @@ jobs:
       - name: Setup | Checkout
         uses: actions/checkout@v2.3.4
 
-      # Cache files between builds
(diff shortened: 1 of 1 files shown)


matchai:
@Milo123459 What does this caching solution do that we wouldn't get with `actions/cache`?

Milo123459:
Caches more and it's smarter

davidkna:
It's also supposed to  fix the issue that caused the macOS cache to break builds, but it looks like that has already been handled in `actions/cache`.  [This post](https://matklad.github.io/2021/09/04/fast-rust-builds.html) goes into some detail regarding how `Swatinem/rust-cache` is smarter.

PR #3066, opened 2021-09-15
feat: Pure prompt preset

timkelty:
- Add Pure prompt preset
- Add screenshot

#### Motivation and Context

Closes #

#### Checklist:


docs/presets/README.md
```diff
@@ -466,3 +466,60 @@ format = 'via [$symbol]($style)'
...
+
+[git_status]
+format = "[[$conflicted$deleted$renamed$modified$untracked](218) $ahead_behind$stashed]($style)"
+style = "cyan"
+diverged = "*"
+untracked = "*"
+modified = "*"
+staged = "*"
+renamed = "*"
+deleted = "*"
```
davidkna, line 511: This will print multiple `*`-characters if multiple of the conditions for these characters are met. I think you want to do something like `(*$conflicted$deleted$renamed$modified$untracked)` in the above format string and set all these variable to zero-width spaces (empty strings would cause them to be ignored).

matchai:
For the sake of consistency with the other preset, would you be open to using this screenshot?

CleanShot 2021-09-30 at 21 42 34@2x

timkelty:
@matchai of course, thanks for making it! PR updated.

docs/presets/README.md
```diff
@@ -466,3 +466,60 @@ format = 'via [$symbol]($style)'
...
+style = "blue"
+
+[character]
+success_symbol = "[❯](5)"
+error_symbol = "[❯](red)"
+vicmd_symbol = "[❮](green)"
+
+[git_branch]
+format = "[$branch]($style)"
+style = "242"
```
matchai, line 501: It seems `242` is fairly similar to what most themes assign to `bright-black` (which is `8`), as shown here:
image

Would you be open to having it changed for the values set to `242`?
It would allow for the preset to better match themes with different color schemes.

```suggestion
style = "bright-black"
```

docs/presets/README.md
```diff
@@ -466,3 +466,60 @@ format = 'via [$symbol]($style)'
...
+$cmd_duration\
+$line_break\
+$python\
+$character"""
+
+[directory]
+style = "blue"
+
+[character]
+success_symbol = "[❯](5)"
```
matchai, line 495: `5` is the same as `purple` here.
For consistency, I'd suggest we change it:
```suggestion
success_symbol = "[❯](purple)"
```

timkelty:
@matchai yep, good call. Tested them out and couldn't even tell the difference.

matchai:
Thank you for contributing to Starship's presets, @timkelty! 😄

ChangeHow:
does it will sync to Linux? 
image

PR #3067, opened 2021-09-15
fix(windows): make powershell init work on windows 8 again

### LINKED ISSUE #3064: Powershell Init does not work on Windows 8
#### Current Behavior

When running `Invoke-Expression (&starship init powershell)` on Windows 8.1 SP1, the prompt becomes

`
PS>
`

instead of the starship prompt.

#### Expected Behavior

Starship should become the prompt when running this command.

This behavior was initially discovered by user cybernoid23 on Discord. I have bisected the issue to commit 20d845f9b3ad3b0c82e43f2af9e7d2d037690976, which is linked to #2104.

#### Possible Solution

I can see two possible ways to deal with this: one is to fix the powershell init so that it works with Windows 8.

Another is to simply label this as WONTFIX: the fact that we haven't gotten any prior bug reports on this issue (which has existed since Feb of this year) suggests that we either have zero Windows 8 users, or that they have no reason to update.

#### Environment
- Starship version: N/A
- Shell type: powershell
- Shell version: 4.0
- Shell plugin manager: N/A
- Terminal emulator: Windows Console
- Operating system: Windows 8

None

#### Starship Configuration
None

davidkna:
#### Description

`[System.Diagnostics.ProcessStartInfo]::new` which was introduced #2104 doesn't exist on Windows 8, but  `New-Object` works instead. Using the implementation from on older suggestion of mine https://github.com/starship/starship/pull/2104#discussion_r558873960.

#### Motivation and Context

Closes #3064

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -24,12 +24,13 @@ function global:prompt {
 
     function Invoke-Native {
         param($Executable, $Arguments)
-        $startInfo = [System.Diagnostics.ProcessStartInfo]::new($Executable);
-        $startInfo.StandardOutputEncoding = [System.Text.Encoding]::UTF8;
-        $startInfo.RedirectStandardOutput = $true;
-        $startInfo.RedirectStandardError = $true;
-        $startInfo.CreateNoWindow = $true;
-        $startInfo.UseShellExecute = $false;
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM

PR #3088, opened 2021-09-22
fix: pipestatus quoting on Zsh/Bash

### LINKED ISSUE #3046: Starship print error message about 'pipestatus' when open terminal
#### Current Behavior
Starship print some error message when i open the terminal
```
error: The argument '--pipestatus <PIPESTATUS>...' requires a value but none was supplied

USAGE:
    starship prompt --cmd-duration <CMD_DURATION> --jobs <JOBS> --keymap <KEYMAP> --pipestatus <PIPESTATUS>... --status <STATUS_CODE>

For more information try --help
```

#### Expected Behavior
Starship runs whithout error message

image

#### Environment
- Starship version: 0.57.0
- zsh version: zsh 5.8 (x86_64-pc-linux-gnu)
- Operating system: Arch Linux Unknown
- Terminal emulator: alacritty 0.9.0
- Git Commit Hash: f03c0825
- Branch/Tag: v0.57.0
- Rust Version: rustc 1.54.0
- Rust channel:  release
- Build Time: 2021-08-27 08:31:01 +00:00

```bash
source ~/.zinit/bin/zinit.zsh
eval "$(starship init zsh)"
```

#### Starship Configuration

```toml
add_newline = false

[character]
success_symbol = "[=>](bold)"
error_symbol = "[=>](bold red)"

[directory]
truncation_length = 5
read_only = " !w"
read_only_style = "red bold"

[package]
disabled = true

[cmd_duration]
disabled = true
#min_time = 500
#format = "[ | ](fg:208 bold)[$duration]($style)"

[git_branch]
format = "[| ](fg:208 bold)[$symbol$branch]($style) "

[git_commit]
commit_hash_length = 6

[git_state]
disabled = true

[git_status]
disabled = true
staged_count.enabled = true
staged_count.style = "yellow"

[line_break]
disable = false

```

chipbuster:
#### Description

Modifies pipestatus parsing for the starship binary and changes the initscripts for bash/zsh to quote pipestatus. This allows starship to start without errors in the case where pipestatus is empty (usually due to interactions with other shell frameworks).

#### Motivation and Context

The previous implementation of pipestatus assumed that the arguments to pipestatus would be word-split (bash, zsh) or repeated by the shell (fish). Unfortunately, this causes a shell parse error if pipestatus is not set on first invocation (which can apparently happen if other shell frameworks are present).

There are two ways to fix this:
  1. Modify the init scripts to not use `--pipestatus` at all when pipestatus is detected to be empty.
  2. Quote the arguments to `--pipestatus`, which allows clap to detect an empty argument, and then change the parsing code in starship to split the arguments ourselves.

I have implemented the second here, since the first requires significantly complicating the init scripts for bash/zsh.

~I'll mark this as ready once I can confirm that this fixes the linked bug in the provided docker.~

Confirmed that this does solve the issue mentioned above. There is a separate issue which appears, where on shell startup, the exit status of the previous command is not defined, so starship prints a warning about not being able to parse the exit status. However, it's not as clear to me how to fix that issue, since we do want the warning to appear when a status cannot be parsed.

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -115,10 +115,10 @@ impl<'a> Context<'a> {
 
         let pipestatus = arguments
             .values_of("pipestatus")
-            .map(|args| Context::get_and_flatten_pipestatus(args))
+            .map(Context::get_and_flatten_pipestatus)
             .flatten();
 
-        log::info!("Pipestatus is {:?}", pipestatus);
+        log::trace!("Received completed pipestatus of {:?}", pipestatus);
 
         // Canonicalize the current path to resolve symlinks, etc.
         // NOTE: On Windows this converts the path to extended-path syntax.
--- a/src/context.rs
+++ b/src/context.rs
@@ -113,15 +113,11 @@ impl<'a> Context<'a> {
             .map(|(a, b)| (*a, b.vals.first().cloned().unwrap().into_string().unwrap()))
             .collect();
 
-        // Due to shell differences, we can potentially receive individual or space
-        // separated inputs, e.g. "0", "1", "2", "0" is the same as "0 1 2 0" and 
-        // "0 1", "2 0". We need to be able to accept for all these formats
-        let pipestatus = arguments.values_of("pipestatus").map(|args| {
-            args.into_iter().map(|x| x.split_ascii_whitespace())
-                .flatten()
-                .map(|x| x.to_string())
-                .collect()
-        });
+        let pipestatus = arguments
+            .values_of("pipestatus")
+            .map(|args| Context::get_and_flatten_pipestatus(args))
+            .flatten();
+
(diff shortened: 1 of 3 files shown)


andytom (approved):

pbar1:
I built this commit on macOS and while it seems to resolve the original issue, I started instead seeing:
```
(starship::modules::status): Error parsing exit_code string to int
```

chipbuster:
Dammit.

I was afraid of something like that, though I'm surprised it's an error (I noted a similar issue in the PR, but for me it would only come out as a warning).

Is it possible for you to share your configs so I can try to repro?

pbar1:
I must not have copied the whole string - it was definitely just a warning, not an error! 😄 

Apologies for the confusion, shoulda read the original post more closely!

PR #3058, opened 2021-09-12
fix(zsh): always escape percent character

davidkna:
#### Description

This PR makes sure `%`-characters are always escaped, by moving the handling from the battery module to `wrap_colorseq_for_shell` where interpretable characters are already being handled for bash.

#### Motivation and Context

Closes #2891

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #3107, opened 2021-09-30
fix(escaping): move escaping to individual variables

### LINKED ISSUE #3072: ZSH: % escaping breaks prompt escapes
#### Current Behavior
As of commit 5ac7ad741fdcb199671c63ae215a06f216fa78b8 custom modules using ZSH's prompt expansion variables (`%h` for example) are no longer interpolated as they are unconditionally escaped.

For example, the following configuration used to work prior to commit 5ac7ad741fdcb199671c63ae215a06f216fa78b8:

```
[custom.hist]
command="echo {%h}"
style = "fg:129"
when = '[ "$STARSHIP_SHELL" == "zsh" ]'
shell = ["bash", "--noprofile", "--norc"]
format = '[$output]($style)'
disabled=false
```
This would correctly interpolate %h to the history event number.

#### Expected Behavior
Any legitimate `%X` value is correctly interpolated.

#### Possible Solution
In speaking with chipbuster on Disord, we concluded the easiest approach is to disable the escaping for custom modules.  Although not as configurable, it's a nice approach between fixing this, and seeing how this might affect other users, until they let us know.

mcfedr:
#### Description

Moves shell prompt escaping to the level of individual format variables

#### Motivation and Context

Several previous attempts to fix this have been made, with different levels of success.

This version means that the user can have special prompt chars in expected places, e.g. `format` setting and also in the output of custom commands. But at the same time protects against injections from outside sources, e.g. git branch names or special chars in folder names.

As all the module expose outside data though variables this is a good place to do the escaping, whilst still allowing purposeful special chars in the format string.

Closes #3072
Also see context #3058 and #658

Unfortunately its quite large because of the need to pass the `context.shell` var into the string formatter.

Fixed special chars (% in this case for zsh) not interpreteted:

Screenshot 2021-09-30 at 12 41 59

Previously (the green is a custom command that outputs control sequences: 

Screenshot 2021-09-30 at 16 37 56

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3316,6 +3316,19 @@ If you have an interesting example not covered there, feel free to share it ther
 
 :::
 
+::: warning Command output is printed unescaped to the prompt
+
+Whatever output the command generates is printed unmodified in the prompt. This means if the output
+contains special sequences that are interpreted by your shell they will be expanded when displayed.
+These special sequences are shell specific, e.g. you can write a command module that writes bash sequences,
+e.g. `\h`, but this module will not work in a fish or zsh shell.
+
+Format strings can also contain shell specific prompt sequences, e.g.
+[Bash](https://www.gnu.org/software/bash/manual/html_node/Controlling-the-Prompt.html),
+[Zsh](https://zsh.sourceforge.io/Doc/Release/Prompt-Expansion.html).
+
+:::
+
 ### Options
 
 | Option        | Default                         | Description                                                                                                                |
--- a/src/formatter/string_formatter.rs
+++ b/src/formatter/string_formatter.rs
@@ -7,6 +7,7 @@ use std::error::Error;
 use std::fmt;
 
 use crate::config::parse_style_string;
+use crate::context::{Context, Shell};
 use crate::segment::Segment;
 
 use super::model::*;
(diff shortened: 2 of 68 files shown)


mcfedr:
I think I agree, this model is a bit complicated at the moment

I think for me the key thing is that I'm letting the code that generates the output apply escaping to the places where its needed. The idea is similar to how a lot of html templates work - the template, in our case the `format` string is assumed to be valid output, and the variables we put into it need to be escaped appropriately, or not escaped because of `map_safe`

> One of our goals with the next major starship change is to remove the coupling of frontend to backend. 

This is clearly helpful - but that work should acknowledge there are places where the input knows about the output - in my case a custom input, knows its outputting to zsh and i want to put zsh commands in that output without the formatter interfering - I'd argue this PR helps here - thats what `map_safe` gives us - a marker to the formater that this output shouldnt be touched.

chipbuster:
> and maybe now introducing `SafePlain` actually bring some value when you get around to the new version.

Yes, on further reflection, this is correct, because it points out where escape-inhibited strings need to be generated (as of right now, just in custom, and perhaps in the hypothetical `passthrough` module that I'll get around to writing someday)

mcfedr:
@chipbuster rebased on master to catch a new module, so pipeline needs approval again.

andytom:
Thank you for your contribution @mcfedr and thank you for the review @chipbuster.

PR #3120, opened 2021-10-04
fix: Correctly handle durations of 0ms

### LINKED ISSUE #3103: Correctly handle durations of zero milliseconds in module cmd_duration
## Feature Request

I am using `fish`. If one uses the following configuration

```toml
[cmd_duration]
min_time = 0
show_milliseconds = true
format = "x [$duration]($style) x "
```
the module displays `x  x ` when there is no previous command or when the previous command took zero milliseconds.

When a command takes zero milliseconds, it should display `x 0ms x `. When there is no previous command (starting a new shell for instance) it should not display.

Setting `min_time` to `1` fixes the second problem but not the first one.

ananas-dev:
fix: Correctly handle durations of 0ms (#3103) 

#### Description

Display when a command took 0ms to complete.

#### Motivation and Context

Closes #3103 

Tested using cargo test and trying by hand using the config mentioned in the issue.

#### Checklist:

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -479,19 +479,21 @@ pub fn render_time(raw_millis: u128, show_millis: bool) -> String {
     let mut rendered_components: Vec<String> = components
         .iter()
         .zip(&suffixes)
-        .map(render_time_component)
+        .map(|time_data| render_time_component(time_data, true))
         .collect();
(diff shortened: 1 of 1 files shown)


ananas-dev:
My bad, it doesn't work in a certain case, I'll open another pr later when I'm sure it works correctly

PR #3124, opened 2021-10-05
fix: allow unset KEYMAP and STARSHIP_DURATION in zsh

harrysarson:
which prevents errors if a user has `set -u` on in there terminal.

#### Description

In my terminal I have `set -u` on which I find really helpful to catch typos I make when using env variables. However, because the prompt is run in the context of my shell the `set -u` applies to it too.

In a couple of places, the starship zsh prompt uses possibly unset variables relying on their value falling back to an empy string if they are unset. This PR uses bash parameter expansion to do this explicitly.

current behaviour:

image

 I have tested on linux by `cargo install`ing from this PR branch.

#### Checklist:


harrysarson:
Thanks!

PR #3131, opened 2021-10-07
fix(init): Use stderr for init error message

ajeetdsouza:
#### Description

Uses `stderr` for initialization errors instead of generating a `printf` statement that needs to be evaluated.

#### Motivation and Context

When `starship init foo` is called, starship currently outputs a line of shell code. When run with `eval`, this shell code would output a nicely formatted error message.

However, if this code is not passed through `eval`, deciphering the `printf` statement can be a challenge. This PR uses `stderr` for printing the error message. That way, it does not get captured, and is printed correctly whether or not you use `eval`.

#### Checklist:


chipbuster:
@ajeetdsouza The semantic PR bot is upset because we can't squash (and therefore make the commit message conventional) with a single commit.

Could you add a dummy commit with `git commit --allow-empty` and a commit message something like `Add a dummy commit to appease the Semantic PR bot` and then push? Thanks!

chipbuster:
Nevermind, it turns out that it wasn't letting me add commits because I was accidentally on master instead of this branch.

andytom:
Thank you for your contribution @ajeetdsouza.

PR #3135, opened 2021-10-08
feat(module): Add `sudo` module

### LINKED ISSUE #816: Hint at sudo credential caching
## Feature Request

To indicate whether user credentials are currently cached for `sudo` usage.

I want to know when user credentials are cached for running any command as I want `sudo` to prompt for credentials when used by untrusted programs (just to know that the program is trying to do superuser stuff, in which case I kill it).

A new module that adds this information to the prompt.

  - (I use fish) Customize `fish_prompt`. Currently not possible (see for example #625)?
  - Manually run `sudo -K` to remove user cached credentials or spawn a new shell before running any nontrusted command. Not really a solution.
  - Deactivate credential cache? Not really a solution.

Related to #664.

gabrielvictorcf:
#### Description
Add a sudo credential caching hint module. It's disabled by default to avoid cluttering the prompt too much, however users that want this feature only need to change `disabled = "false"`.

#### Motivation and Context

Closes #816 

image

I was only able to create one test.

#### Checklist:

### DIFF
--- a/src/modules/sudo.rs
+++ b/src/modules/sudo.rs
@@ -99,7 +99,7 @@ mod tests {
                     stdout: "".to_owned(),
                     stderr: "".to_owned(),
                 }),
-             )
+            )
             .config(toml::toml! {
                 [sudo]
                 disabled = false
--- a/src/modules/sudo.rs
+++ b/src/modules/sudo.rs
@@ -61,6 +61,7 @@ mod tests {
             .config(toml::toml! {
                 [sudo]
                 disabled = false
+                allow_windows = true
             })
             .collect();
         let expected = None;
@@ -82,6 +83,7 @@ mod tests {
             .config(toml::toml! {
                 [sudo]
                 disabled = false
+                allow_windows = true
             })
             .collect();
         let expected = Some(format!("{}", Color::Blue.bold().paint("as 🧙 ")));
--- a/src/modules/sudo.rs
+++ b/src/modules/sudo.rs
@@ -79,7 +79,7 @@ mod tests {
                 disabled = false
             })
             .collect();
-        let expected = Some(format!("{}", Color::Blue.bold().paint("as 🧙‍ ")));
+        let expected = Some(format!("{}", Color::Blue.bold().paint("as 🧙 ")));
 
         assert_eq!(expected, actual);
     }
--- a/src/modules/sudo.rs
+++ b/src/modules/sudo.rs
@@ -104,7 +104,7 @@ mod tests {
     #[cfg(not(windows))]
     fn test_doas_cached() {
         let actual = ModuleRenderer::new("sudo")
-            .cmd("doas true", None)
+            .cmd("doas -n true", None)
             .config(toml::toml! {
                 [sudo]
                 disabled = false
--- a/src/modules/sudo.rs
+++ b/src/modules/sudo.rs
@@ -1,0 +1,68 @@
+use super::{Context, Module, RootModuleConfig};
+
+use crate::configs::sudo::SudoConfig;
+use crate::formatter::StringFormatter;
+
+/// Creates a module with sudo credential cache status
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("sudo");
+    let config = SudoConfig::try_load(module.config);
+
+    if config.disabled {
+        return None;
+    }
+
+    let is_sudo_cached = context.exec_cmd("sudo", &["-n", "true"]).is_some();
+
+    if !is_sudo_cached {
+        return None;
+    }
+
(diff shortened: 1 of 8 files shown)


gabrielvictorcf:
Hi, sorry for the delay -.- . I inserted the `sudo` and the `doas` examples in the `docs/config/README.md` as a _subtle guide_ to users that the `binary` option (to the extent of my knowledge, at least) doesn't make a lot of sense with other sudo replacements - mostly because we're not very transparent that we run with `<binary> -n true`. 

Should i make it more clear, somewhere in the README, that `binary = sudo || doas` or is what i commited enough?

gabrielvictorcf:
Hmm, the `doas` tests seem to fail because the binary is missing. What should i do in this case?

src/modules/sudo.rs
```diff
@@ -0,0 +1,107 @@
...
+                disabled = false
+            })
+            .collect();
+        let expected = None;
+
+        assert_eq!(expected, actual);
+    }
+
+    #[test]
+    #[cfg(not(windows))]
```
davidkna, line 72: Can you add `allow_windows = true` to all the tests currently blocking windows and allow them to run there, to ensure `allow_windows = true` works?

src/modules/sudo.rs
```diff
@@ -0,0 +1,107 @@
...
+        let expected = Some(format!("{}", Color::Blue.bold().paint("as 🧙 ")));
+
+        assert_eq!(expected, actual);
+    }
+
+    #[test]
+    #[cfg(windows)]
+    fn test_allow_windows_disabled_blocks_windows() {
+        let actual = ModuleRenderer::new("sudo")
+            .cmd("sudo -n true", None)
```
davidkna, line 96: For this test to work as intended, `sudo` needs to succeed to check if it's being blocked properly.

```suggestion
            .cmd(
                "sudo -n true",
                Some(CommandOutput {
                    stdout: "".to_owned(),
                    stderr: "".to_owned(),
                }),
             )
```

gabrielvictorcf:
Done ^^

src/modules/sudo.rs
```diff
@@ -0,0 +1,115 @@
...
+        formatter
+            .map_meta(|variable, _| match variable {
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .parse(None)
```
davidkna, line 37: To fix compilation:
```suggestion
            .parse(None, Some(context))
```

src/modules/sudo.rs
```diff
@@ -0,0 +1,115 @@
...
+    Some(module)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{test::ModuleRenderer, utils::CommandOutput};
+    use ansi_term::Color;
+
+    #[test]
+    #[cfg(not(windows))]
```
davidkna, line 57: This test can run on Windows with the `allow_windows = true` change.

```suggestion
```

src/modules/sudo.rs
davidkna: Same here:
```suggestion
```

gabrielvictorcf:
Wow, sorry for the trouble -.- . The tests passed fine here, i didn't have the compilation error.

gabrielvictorcf:
The `.parse(None, Some(context))` appears to still be an issue, i'm looking into it

davidkna (approved):

andytom (approved):

andytom:
Thank you for your contribution @gabrielvictorcf and thanks for reviewing @davidkna.

gabrielvictorcf:
This was my first contribution ever, so thank you both for all the patience ^^

zanna-37:
@gabrielvictorcf I think that the docs should state that calling `sudo -n true` has the side effect of resetting the credentials timeout.

>     -v, --validate
>                 Update the user's cached credentials, authenticating
>                 the user if necessary.  For the sudoers plugin, this
>                 extends the sudo timeout for another 5 minutes by
>                 default, but does not run a command.  Not all security
>                 policies support cached credentials.

Refs:
- [man page](https://man7.org/linux/man-pages/man8/sudo.8.html)
- [server fault question](https://serverfault.com/questions/364334/test-whether-a-user-has-sudo-privileges-without-requiring-user-input)

andytom:
@zanna-37 if this is a problem please open a new ticket rather than commenting on an old PR, but from what I can see there might be some mix up between the `-n` and `-v` arguments here.

zanna-37:
My bad, I had this problem in the past and in the attempt of quickly providing evidence I messed up. I'll edit it out.

The problem however is real and the timeout is reset probably because sudo is being used (even if it is to check itself).
Oh, and I commented here to have a discussion before creating a new issue.

PR #3143, opened 2021-10-09
docs(3070): add some precision on how to hide non installed plugins

### LINKED ISSUE #3070: Check for existence of relevant binaries in $PATH
Certain modules may enable themselves, even when the relevant programs are not installed. For example I noticed that Vagrant and Rust modules are both enabled when there are Vagrantfiles/Rust source files present, despite not having Vagrant/rustc/cargo etc installed.

My proposal would be to check for the presence of relevant binaries in the user path, before enabling modules.

Not everyone may want this behaviour, so it would probably make sense to gate it behind a configuration option.

sylvainmetayer:
#### Description

Update configuration documentation to indicate how to hide plugin which are not installed. 

#### Motivation and Context

Closes #3070

I'm not sure if this is the best way to address this issue, I tried to find an "update all configuration" method with a regex matching most of the plugin but there is maybe a better way.

Only documentation have been changed. 

#### Checklist:

- [ ] I have updated the tests accordingly

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -112,6 +112,18 @@ For example:
 - When `$all` is a shortcut for `\[$a$b\] `, `($all)` will show nothing only if `$a` and `$b` are both `None`.
   This works the same as `(\[$a$b\] )`.
 
+::: tip
+
+By default, icons tools will still be displayed, even if the tool is not installed on your system. If you do not want this behaviour, you can edit the configuration file to replace `via [$symbol($version )]($style)` with `(via [$symbol$version]($style))` so that the icon will only be displayed if you have it installed on your system (meaning that `$version` is not `None`).
+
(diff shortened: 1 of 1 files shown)


sylvainmetayer:
As seen with @chipbuster and @matchai , a shell command is not the best solution and can cause more harm than good. A better solution may be to add a root option to enable/disable this behaviour globally.

chipbuster:
@sylvainmetayer Thanks for the attempt! Unfortunately, it's hard to see in advance what sort of compatibility issues can arise from tricks like these (especially with the ridiculous number of platforms we support).

I still think your idea is worth being recorded. If you'd like, you can [open a discussion](https://github.com/starship/starship/discussions) and put your solution there (with a note that it's only going to work for GNU sed and default config location) and then link it on 3070.

PR #3144, opened 2021-10-09
feat(python): Show value of PYENV_VERSION when present

### LINKED ISSUE #2872: Starship recognizes PYENV_VERSION in environment
I use `direnv` to setup my environment in a given working directory. When I put `export PYENV_VERSION=foo` in a `.envrc` file, pyenv recognizes that I want to activate `foo` and does so, but my prompt does not change (in fact, the Python portion of the prompt is missing).

I'd like Starship to show the virtualenv as it would (and currently does) if I were to have a `.python-version` file in the directory.

Using a `.python-version` file works, but I don't like cluttering my working directory with files simply for the purpose of setting a single environment variable.

cbolgiano:
#### Description
if config.pyenv_version_name is true and if there is a PYENV_VERSION environment present when executing starship module python the output uses the value of PYENV_VERSION otherwise it will show the version in the output of pyenv version-name.

#### Motivation and Context

Closes #2872

Executed these commands:
1. pyenv local 3.8.0
2. ./starship module python
3. export PYENV_VERSION="3.7.0"
4. ./starship module python
5. export PYENV_VERSION=""
6. ./starship module python

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -76,12 +76,17 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn get_pyenv_version(context: &Context) -> Option<String> {
-    let version_name = context
+    let mut version_name = context.get_env("PYENV_VERSION");
+
+    if version_name.is_none() {
+      version_name = Some(context
         .exec_cmd("pyenv", &["version-name"])?
         .stdout
         .trim()
-        .to_string();
-    Some(version_name)
+        .to_string())
+    }
+    
+    version_name
 }
 
 fn get_python_version(context: &Context, config: &PythonConfig) -> Option<String> {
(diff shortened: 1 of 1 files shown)


cbolgiano:
@chipbuster Good evening! Do you by chance have time to provide a code review? Thank you!

src/modules/python.rs
```diff
@@ -76,12 +76,19 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+        version_name = Some(
+            context
+                .exec_cmd("pyenv", &["version-name"])?
+                .stdout
+                .trim()
+                .to_string(),
+        )
+    }
+
+    version_name
```
andytom, line 91: Not a blocker but I think this can be simplified to
```suggestion
    match context.get_env("PYENV_VERSION") {                                                                                
        Some(version) => Some(version),                                                                                     
        _ => Some(                                                                                                          
            context                                                                                                         
                .exec_cmd("pyenv", &["version-name"])?                                                                      
                .stdout                                                                                                     
                .trim()                                                                                                     
                .to_string(),                                                                                                                                                                                     
        ),                                                                                                                  
    }
```

andytom (approved):
Have left a suggestion but other than that LGTM.

andytom:
Thank you for your contribution @cbolgiano

PR #3146, opened 2021-10-10
fix: Modules not working in fish RPROMPT

### LINKED ISSUE #3137: Variables not working in right_format
#### Current Behavior
When the previous command fails `$character` should be the `error_symbol` but it only works in `format` and not `right_format`.
image
As you can see there is a red `X` on the left after the incorrect command but stays a green check on the right.
(`$character` on the left was just added to debug, also doesn't work if only used on the right)

EDIT: `$cmd_duration` doesn't work either.

#### Expected Behavior
`$character` & `$cmd_duration` should correctly render in `right_format`

Same without `$character` in `format`
image

#### Possible Solution
N/A

#### Environment
- Starship version: 0.58.0
- fish version: fish, version 3.3.1
- Operating system: Ubuntu 20.04 (WSL2 WIN11)
- Terminal emulator: Windows Terminal 1.10.2714.0
- Git Commit Hash: 0a60d1e9
- Branch/Tag: v0.58.0
- Rust Version: rustc 1.55.0 (c8dfcfe04 2021-09-06)
- Rust channel:  release
- Build Time: 2021-09-21 13:43:49 +00:00

```bash
if status is-interactive
end

# Start starship
starship init fish | source
```

#### Starship Configuration

```toml
# Use custom format
format = "[  $directory](bg:#3465a4) $character"

right_format = "$character$duration"

# Wait 10 milliseconds for starship to check files under the current directory.
scan_timeout = 10

# Disable the blank line at the start of the prompt
add_newline = false


[character]                            
success_symbol = "[](bold #4e9a06)"
error_symbol = "[](bold #a92020)"

[directory]
style = "bg:#3465a4 fg:white"
```

andytom:
#### Description

This should fix some modules not working correctly in the fish
`RPROMPT`. I have done this the very naive way by simply duplicating the
code that is currently in the `fish_prompt` function to avoid having to
change the scope of any of the variables (currently all local) but happy to
change this if someone has a better suggestion.

#### Motivation and Context

Closes #3137

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -13,6 +13,16 @@ function fish_prompt
 end
 
 function fish_right_prompt
+    switch "$fish_key_bindings"
+        case fish_hybrid_key_bindings fish_vi_key_bindings
+            set STARSHIP_KEYMAP "$fish_bind_mode"
+        case '*'
+            set STARSHIP_KEYMAP insert
+    end
+    set STARSHIP_CMD_STATUS $status
+    # Account for changes in variable name between v2.7 and v3.0
+    set STARSHIP_DURATION "$CMD_DURATION$cmd_duration"
+    set STARSHIP_JOBS (count (jobs -p))
     ::STARSHIP:: prompt --right --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus=$pipestatus --keymap=$STARSHIP_KEYMAP --cmd-duration=$STARSHIP_DURATION --jobs=$STARSHIP_JOBS
 end
 


davidkna:
Is it possible to just change the visibility of the variables in the first `fish_prompt` function instead?

andytom:
> Is it possible to just change the visibility of the variables in the first `fish_prompt` function instead?

I thought of that but wasn't sure if `fiah_prompt` is always executed before `fish_right_prompt` or there is some case where either `fish_right_prompt` is executed before `fish_prompt` or it is executed first.

PR #3157, opened 2021-10-16
feat(directory): Windows path formatting via path_slash::PathBufExt

### LINKED ISSUE #1106: Windows Directory Formatting
#### Current Behavior
Hi all,

I was pleased to see that starship supports Powershell in addition to macOS and Linux. However when using Powershell in Windows 10 (i.e. not WSL), the directory formatting is off. The prompt shows `C:\Windows` as `/c/Windows`

Is this intended? This makes sense for WSL or Powershell under macOS/Linux but not for Powershell under Windows.

Thanks for all the work you do.

#### Expected Behavior
Prompt should display `C:\Windows` as `C:\Windows`

Annotation 2020-04-14 133141

#### Environment
- Starship version: 0.40.1
- powershell version: 5.1.18362.628
- Terminal emulator: Windows Terminal 0.10.781.0

`powershell`

#### Starship Configuration

`default configuration`

segevfiner:
#### Description
Format the path in the directory module using `path_slash::PathBufExt::from_slash` to convert it into Windows style.

I do hope this is the right approach...

#### Motivation and Context

Closes #1106

~This is just an experiment in order to garner feedback if this is the right approach, and probably breaks tests. For me or someone else to pick up on if this is right.~

#### Checklist:


segevfiner:
Tried fixing CI failures, still waiting on workflow approval...

segevfiner:
Fixed CI failures... (Must remember to configure editor to run fmt and clippy...)

docs/config/README.md
```diff
@@ -807,12 +807,12 @@ it would have been `nixpkgs/pkgs`.
...
 | `home_symbol`       | `"~"`                                              | The symbol indicating home directory.                                           |
+| `convert_slash`     | `true`                                             | Convert slash on output on Windows                                               |
 
 <details>
 <summary>This module has a few advanced configuration options that control how the directory is displayed.</summary>
 
-| Advanced Option             | Default | Description                                                                                                                                                            |
-| --------------------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Advanced Opt-------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
chipbuster, line 815: It looks like you might have accidentally deleted the rest of this line, and then the editor took over?

docs/config/README.md
```diff
@@ -807,12 +807,12 @@ it would have been `nixpkgs/pkgs`.
 | `truncation_symbol` | `""`                                               | The symbol to prefix to truncated paths. eg: "…/"                                |
 | `repo_root_style`   | `None`                                             | The style for the root of the git repo when `truncate_to_repo` option is set to false.|
 | `home_symbol`       | `"~"`                                              | The symbol indicating home directory.                                           |
+| `convert_slash`     | `true`                                             | Convert slash on output on Windows                                               |
```
chipbuster, line 810: Can we get a more detailed description here? Just looking at this, I have no idea what it means to convert slash output.

What's the intention behind this option? My impression is that, when enabled, it will display paths on windows using backslashes instead of forward slashes? In that case, I think it might be better to name it for what it does (e.g. `win_display_backslash`, though I'd like a shorter name) instead of how it accomplishes this.

src/modules/directory.rs
```diff
@@ -117,6 +117,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         _ => ["".to_string(), "".to_string(), prefix + &dir_string],
     };
 
+    if config.convert_slash {
```
chipbuster, line 120: We don't actually need to make `path_vec` mutable for this right? We can shadow the old variable, e.g.

```suggestion
    let path_vec = if config.convert_slash { path_vec.iter().for_each(|i| *i = from_slash(i)).collect() } else { path_vec };
```

though this also might just be my dislike for mutable state getting the better of me.

chipbuster: Definitely don't have to implement this if you feel the current approach is better--I'm sort of ambivalent about this.

src/modules/directory.rs
```diff
@@ -321,6 +325,10 @@ fn to_fish_style(pwd_dir_length: usize, dir_string: String, truncated_dir_string
         .join("/")
 }
 
+fn from_slash(path: &str) -> String {
```
chipbuster, line 328: Similar complaint here: I don't know what `from_slash` is supposed to do from its name. If we can't get a more descriptive name that isn't a mouthful (because the only names I can think of are way too long, like "convert_to_platform_path_separators"), I'd like to see a doc comment that describes what it does.

src/modules/directory.rs
```diff
@@ -573,7 +581,7 @@ mod tests {
                 home_symbol = "🚀"
             })
             .collect();
-        let expected = Some(format!("{} ", Color::Cyan.bold().paint("🚀")));
+        let expected = Some(from_slash(&format!("{} ", Color::Cyan.bold().paint("🚀"))));
```
chipbuster, line 584: I hate to ask this, but could you move the calls to `from_slash` into the tightest scope possible? In this case, it would be something like 

```suggestion
  let expected = Some(&format!("{} ", Color::Cyan.bold().paint(from_slash("🚀"))));
```

Where the call currently is, it's not obvious from context whether the from_slash call requires a formatted color string, or can work on bare paths. We can add a test that tests this in the current scope if you'd like (after coloring is added), but I think the current tests don't accurately reflect how the function is meant to be used.

chipbuster (changes requested):
This review is mostly me being picky about naming and tests-as-documentation, but I think this is a fine PR overall.

segevfiner:
@chipbuster Made requested changes.

src/modules/directory.rs
chipbuster: Ahh, you're right, I didn't think about that.

Well, let's keep it for now, and then once the benchmarks are online, you can come back and say "I told you so" :D

chipbuster (approved):
:+1:

Thanks for the quick turnaround on changes @segevfiner!

chipbuster:
Well, looks like one of the AWS tests is failing with a frankly hilarious error. I'll rerun the CI to see if that just happens to fix it, and if not I think it's safe to merge through that: unless we have some strange race condition which causes rounding errors, this PR is almost certainly not responsible for the failed test.

andytom:
Thank you for your contribution @segevfiner and thanks for reviewing @chipbuster.

PR #3160, opened 2021-10-16
fix: save pipestatus in fish init functions

### LINKED ISSUE #3159: Status module's pipestatus feature not working with fish
#### Current Behavior
When using starship with fish, the status module's pipestatus feature doesn't work. All i get is the exit code for the final command if it errored out.

#### Expected Behavior
I expect that when I execute a pipeline in fish with the status module not-disabled and the pipeline status feature enabled that I will see an indicator of the exit status for all the commands in the pipeline.

Screenshot:
image

The default fish prompt includes a pipestatus indicator if the last command in the pipeline failed but prefer starship's as it will show me pipestatus on all pipelined commands including ones where the final command exists successfully.

With my starship configuration `bash` works fine.

#### Possible Solution
~Both `bash` and `zsh` set `$PIPESTATUS` while fish sets `$pipestatus` - the format is the same but the case of the name is different. Possibly an issue?~ EDIT: This was not an issue - the inits look like they use appropriate variables.

#### Environment
- Shell type: fish
- Shell version: fish, version 3.3.1
- Shell plugin manager: none
- Terminal emulator: iTerm
- Operating system: ubuntu via multipass on mac os

```fish
if status is-interactive
	set fish_cursor_default block
	set fish_cursor_insert line
	set fish_cursor_replace_one underscore
	set fish_cursor_visual block
	fish_vi_key_bindings insert
	starship init fish | source
end
```

#### Starship Configuration

```toml
[status]
disabled = false
format = "[$symbol$common_meaning$signal_name$maybe_int]($style)"
pipestatus_format = "\\[$pipestatus\\]=>[$symbol$common_meaning$signal_name$maybe_int]($style)"
pipestatus = true
map_symbol = true
style = "bold #0000ff"
symbol = "[](bold #ff0000)"
recognize_signal_code = true
success_symbol = "[✓](#00ff00)"
signal_symbol = "[](#ffff00)"
```

underrun:
#### Description

`pipestatus` variable needs to be treated the same way as normal `status` variable, so we save it out in order to use it in the prompt.

#### Motivation and Context

Closes #3159 

image

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -20,6 +20,7 @@ function fish_right_prompt
         case '*'
             set STARSHIP_KEYMAP insert
     end
+    set STARSHIP_CMD_PIPESTATUS $pipestatus
     set STARSHIP_CMD_STATUS $status
     # Account for changes in variable name between v2.7 and v3.0
     set STARSHIP_DURATION "$CMD_DURATION$cmd_duration"
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -5,11 +5,12 @@ function fish_prompt
         case '*'
             set STARSHIP_KEYMAP insert
     end
+    set STARSHIP_CMD_PIPESTATUS $pipestatus
     set STARSHIP_CMD_STATUS $status
     # Account for changes in variable name between v2.7 and v3.0
     set STARSHIP_DURATION "$CMD_DURATION$cmd_duration"
     set STARSHIP_JOBS (count (jobs -p))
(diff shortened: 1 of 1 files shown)


src/init/starship.fish
```diff
@@ -23,7 +24,7 @@ function fish_right_prompt
     # Account for changes in variable name between v2.7 and v3.0
```
andytom, line 24: You will also need to set the `STARSHIP_CMD_PIPESTATUS` env var in the `fist_right_prompt` function.

```suggestion
    set STARSHIP_CMD_PIPESTATUS $pipestatus
    # Account for changes in variable name between v2.7 and v3.0
```

PR #3179, opened 2021-10-25
feat: print-config subset of config

### LINKED ISSUE #2934: `starship print-config` should accept optional paths to print
`starship print-config` prints a lot of info, which can be annoying to look through if I just want to see one module.

`starship print-config` should optionally take one or more paths to print. This could be modules, but it would be more generally useful if it was paths, e.g. I could write `starship print-config format` to see my format, or `starship print-config git_status git_branch.format` to see the whole `git_status` module and just the `format` key in `git_branch`.

For example:

```toml
# starship print-config format
format = '$all'

# starship print-config git_status git_branch.format
[git_branch]
format = '\[[$symbol$branch]($style)\]'

[git_status]
format = '([\[$all_status$ahead_behind\]]($style))'
style = 'bold yellow'
stashed = '\$'
ahead = '⇡'
behind = '⇣'
diverged = '⇕'
conflicted = '='
deleted = '✘'
renamed = '»'
modified = '!'
staged = '+'
untracked = '?'
disabled = false
```

mdonoughe:
#### Description

`starship print-config` now takes an optional number of paths to filter the configuration by. It doesn't look like the `print-config` command is documented so I didn't add documentation for this new behavior.

#### Motivation and Context

Closes #2934 

```
$ starship print-config git_status git_branch.format
# Warning: This config does not include keys that have an unset value

# $all is shorthand for $username$hostname$shlvl$singularity$kubernetes$directory$vcsh$git_branch$git_commit$git_state$git_metrics$git_status$hg_branch$docker_context$package$cmake$cobol$dart$deno$dotnet$elixir$elm$erlang$golang$helm$java$julia$kotlin$lua$nim$nodejs$ocaml$perl$php$pulumi$purescript$python$rlang$red$ruby$rust$scala$swift$terraform$vlang$vagrant$zig$nix_shell$conda$memory_usage$aws$gcloud$openstack$env_var$crystal$custom$cmd_duration$line_break$jobs$battery$time$status$shell$character
[git_status]
format = '([\[$all_status$ahead_behind\]]($style) )'
style = 'red bold'
stashed = '\$'
ahead = '⇡'
behind = '⇣'
up_to_date = ''
diverged = '⇕'
conflicted = '='
deleted = '✘'
renamed = '»'
modified = '!'
staged = '+'
untracked = '?'
disabled = false

[git_branch]
format = 'on [$symbol$branch]($style)(:[$remote]($style)) '

```

#### Checklist:

### DIFF
--- a/src/configure.rs
+++ b/src/configure.rs
@@ -65,7 +65,7 @@ fn handle_update_configuration(doc: &mut Document, name: &str, value: &str) -> R
     Ok(())
 }
 
-pub fn print_configuration(use_default: bool) {
+pub fn print_configuration(use_default: bool, paths: &[&str]) {
     let config = if use_default {
         // Get default config
         let default_config = crate::configs::FullConfig::default();
@@ -79,8 +79,6 @@ pub fn print_configuration(use_default: bool) {
         // Convert back to Value because toml can't serialize FullConfig directly
         toml::value::Value::try_from(user_config).unwrap()
     };
-
-    let string_config = toml::to_string_pretty(&config).unwrap();
 
     println!("# Warning: This config does not include keys that have an unset value\n");
     println!(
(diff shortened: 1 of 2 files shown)


src/configure.rs
```diff
@@ -102,9 +100,72 @@ pub fn print_configuration(use_default: bool) {
                 .collect::<String>()
         );
     }
```
davidkna, line 102: Can you ensure the comments related to `$all` only get printed if `paths` is empty or if it contains `format` or `right_format`?

src/configure.rs
```diff
@@ -102,9 +100,72 @@ pub fn print_configuration(use_default: bool) {
...
+    } else {
+        extract_toml_paths(config, paths)
+    };
+
+    let string_config = toml::to_string_pretty(&print_config).unwrap();
+
     println!("{}", string_config);
 }
 
+fn extract_toml_paths(mut config: toml::Value, paths: &[&str]) -> toml::Value {
```
davidkna, line 115: I think it might be simpler to remove any values from the config that are not listed in `paths` instead of copying them over, but this also looks good to me.

src/main.rs
```diff
@@ -166,6 +166,12 @@ fn main() {
                             .long("default")
                             .help("Print the default instead of the computed config")
                             .takes_value(false),
+                    )
+                    .arg(
+                        Arg::with_name("path")
+                            .help("Configuration section or value to print")
```
davidkna, line 172: I think this should be plural, and I don't think value quite fits. Maybe subsets would be a better term?

davidkna (changes requested):

src/configure.rs
mdonoughe: To do it that way, the code would need to recursively compare the keys that are present to the provided paths. I guess if the paths were parsed into a recursive structure first it wouldn't be too bad, but I think it would be significantly more complicated.

src/main.rs
mdonoughe: I changed the argument name to "name" and its help to "keys" to match the `config` command.

davidkna (approved):

andytom:
Thank you for your contribution @mdonoughe and thanks for reviewing @davidkna.

PR #3206, opened 2021-11-05
feat(ruby): Add environment variable checks to ruby module

aaronkollasch:
#### Description

This PR adds a check in the ruby module for the environment variables `RUBY_VERSION` and `RBENV_VERSION`, which indicate that ruby is currently activated with RVM or rbenv (respectively). 
The variables to check are configurable with the `detect_variables` option.

#### Motivation and Context
Starship does not detect ruby environments activated with environment variables.
See my [comment](https://github.com/starship/starship/issues/698#issuecomment-961592678) in #698

I've tested this with rbenv and RVM.
##### rbenv (1.2.0)
```
brew install rbenv ruby-build
rbenv install 3.0.2
```
In a new terminal, `rbenv shell 3.0.2` adds the correct ruby version to the prompt (3.0.2).

##### RVM (rvm 1.29.12-next)
```
\curl -sSL https://get.rvm.io | bash -s -- --ignore-dotfiles
source $HOME/.rvm/scripts/rvm
rvm install 3.0.1
```
In a new terminal, `source $HOME/.rvm/scripts/rvm` adds the correct ruby version to the prompt (3.0.1).

One issue with RVM is that by default, it is activated permanently through the user's `.profile` files. If someone has a ruby env activated this way, they may not want to see the ruby version all the time, so I added the detect_variables option to make this check configurable.

#### Checklist:

- [x] I have updated the documentation accordingly
(but only `docs/config/README.md`, not the internationalizations).
though the tests don't actually use a ruby environment, so only the presence of the relevant variable matters. I suppose that matches how the code works at the moment.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2644,6 +2644,8 @@ The module will be shown if any of the following conditions are met:
 - The current directory contains a `.rb` file
 - The environment variables `RUBY_VERSION` or `RBENV_VERSION` are set
 
+Starship gets the current Ruby version by running `ruby -v`.
+
 ### Options
 
 | Option              | Default                              | Description                                                               |
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2642,6 +2642,7 @@ The module will be shown if any of the following conditions are met:
 - The current directory contains a `Gemfile` file
 - The current directory contains a `.ruby-version` file
 - The current directory contains a `.rb` file
+- The environment variables `RUBY_VERSION` or `RBENV_VERSION` are set
 
 ### Options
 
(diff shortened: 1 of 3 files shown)


docs/config/README.md
```diff
@@ -2642,6 +2642,7 @@ The module will be shown if any of the following conditions are met:
 - The current directory contains a `Gemfile` file
 - The current directory contains a `.ruby-version` file
 - The current directory contains a `.rb` file
+- The environment variables `RUBY_VERSION` or `RBENV_VERSION` are set
```
andytom, line 2645: Should we mention that the value of these env vars will not be used? If the `RUBY_VERSION` is set to x but `ruby - - version` outputs y this might cause some confusion.

andytom (approved):
Have added a comment/question about the docs but other than that this LGTM.

docs/config/README.md
aaronkollasch: Perhaps we could add a message below the bullet points like this:
"Starship gets the current Ruby version by checking `ruby -v`."
The same thing applies to the current directory files as well.

andytom:
Thank you for your contribution @aaronkollasch.

PR #3224, opened 2021-11-11
feat(hostname): Add hostname style change support.

PatrickChoDev:
#### Description

Add `ssh_style` to HostnameConfig to support different color for same hostname in SSH connection.

#### Motivation and Context

Improve #3214 

#### Checklist:


PatrickChoDev:
I don't sure why document changed so much. I'm actually open VS Code, touch nothing except Hostname config zone. Sorry if it the problems. T-T

chipbuster:
> I don't sure why document changed so much. I'm actually open VS Code, touch nothing except Hostname config zone. Sorry if it the problems. T-T

You probably have VSCode set to auto-format documents on save. The VSCode formatter is known to do weird things like this to markdown tables. Don't worry too much about it!

src/modules/hostname.rs
```diff
@@ -40,11 +40,16 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+        ("ssh_style", config.ssh_style)
+    } else {
+        ("style", config.style)
+    };
 
     let parsed = StringFormatter::new(config.format).and_then(|formatter| {
         formatter
             .map_style(|variable| match variable {
-                "style" => Some(Ok(config.style)),
+                _choose_style => Some(Ok(choose_config)),
```
chipbuster, line 52: This likely doesn't work the way you think it does. This is always going to match, shadowing whatever value `_choose_style` had previously. The `_ => None` arm after this will never be matched.

Here is one way to rewrite it, though I don't think it's optimal:

```
match variable {
  "ssh_style" if ssh_connection.is_some() => Some(Ok(config.ssh_style)),
  "ssh_style" if ssh_connection.is_none() => Some(Ok(config.style)),
  "style" => Some(Ok(config.style)),
  _ => None,
}
```

src/modules/hostname.rs
```diff
@@ -123,7 +128,7 @@ mod tests {
             })
             .env("SSH_CONNECTION", "something")
             .collect();
-        let expected = Some(format!("{} in ", style().paint(hostname)));
+        let expected = Some(format!("{} in ", ssh_style().paint(hostname)));
```
chipbuster, line 131: If we're going to do this, perhaps we should use a non-standard ssh_style in the config so that we can be sure that that codepath is actually being triggered?

chipbuster (changes requested):
The warnings generated by Clippy in this case are serious: the match statement in the current implementation is broken. See my comment on that match arm for more details.

The fact that the tests didn't catch this also suggests that the current test suite is inadequate, though that's not really your problem to fix (just a note for future me).

chipbuster:
@PatrickChoDev Looks like there are failures in the automated test suite.

docs/config/README.md
```diff
@@ -1591,13 +1590,14 @@ The `hostname` module shows the system hostname.
...
+| Option      | Default                     | Description                                                                                                                          |
+| ----------- | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| `ssh_only`  | `true`                      | Only show hostname when connected to an SSH session.                                                                                 |
+| `trim_at`   | `"."`                       | String that the hostname is cut off at, after the first match. `"."` will stop after the first dot. `""` will disable any truncation |
+| `format`    | `"[$hostname]($style) in "` | The format for the module.                                                                                                           |
+| `ssh_style` | `"bold yellow"`             | The style for remote hostname in SSH connection.                                                                                     |
```
chipbuster, line 1598: If I understand correctly, this also needs to be documented in the variables section. It also might be worth documenting that this is only used if an SSH connection is detected, otherwise strings styled with this variable will fall back to using `style`.

PatrickChoDev:
How to do a doc format?

chipbuster:
You need to install dprint. Check out this part of CONTRIBUTING.md for details: https://github.com/starship/starship/blob/master/CONTRIBUTING.md#formatting

docs/config/README.md
```diff
@@ -1199,7 +1199,7 @@ This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gc
 | Option           | Default                                                  | Description                                                     |
 | ---------------- | -------------------------------------------------------- | --------------------------------------------------------------- |
 | `format`         | `'on [$symbol$account(@$domain)(\($region\))]($style) '` | The format for the module.                                      |
-| `symbol`         | `"☁️  "`                                                 | The symbol used before displaying the current GCP profile.      |
+| `symbol`         | `"☁️ "`                                                  | The symbol used before displaying the current GCP profile.      |
```
chipbuster, line 1202: This change needs to be reverted (as do the other emoji changes below).

docs/config/README.md
```diff
@@ -2858,7 +2861,7 @@ set to a number and meets or exceeds the specified threshold.
 | ----------- | ---------------------------- | ------------------------------------------------------------- |
 | `threshold` | `2`                          | Display threshold.                                            |
 | `format`    | `"[$symbol$shlvl]($style) "` | The format for the module.                                    |
-| `symbol`    | `"↕️  "`                     | The symbol used to represent the `SHLVL`.                     |
+| `symbol`    | `"↕️ "`                      | The symbol used to represent the `SHLVL`.                     |
```
chipbuster, line 2864: This needs to be reverted as well.

chipbuster (commented):
Please undo the doc changes with emoji spacing, and I think I'll be done with changes I want.

I'd also like to request someone else from @starship/astronauts review this before we merge, since I've been looking at it for too long at this point.

PatrickChoDev:
I'm not sure if some space was created by `dprint` or not.. because I didn't touch those things around.

PatrickChoDev:
Maybe finished?

docs/config/README.md
```diff
@@ -2933,7 +2934,7 @@ set to a number and meets or exceeds the specified threshold.
 | ----------- | ---------------------------- | ------------------------------------------------------------- |
 | `threshold` | `2`                          | Display threshold.                                            |
 | `format`    | `"[$symbol$shlvl]($style) "` | The format for the module.                                    |
-| `symbol`    | `"↕️  "`                     | The symbol used to represent the `SHLVL`.                     |
+| `symbol`    | `"↕️"`                       | The symbol used to represent the `SHLVL`.                     |
```
chipbuster, line 2937: Not sure what happened here, but this seems to have gone in the wrong direction since I reviewed. The symbol needs to have two spaces after it before the close quote (how it was in the original documentation).

src/configs/hostname.rs
```diff
@@ -18,6 +20,8 @@ impl<'a> Default for HostnameConfig<'a> {
             ssh_only: true,
             trim_at: ".",
             format: "[$hostname]($style) in ",
+            ssh_format: "[$hostname]($ssh_style) in ",
+            ssh_style: "yellow bold",
```
chipbuster, line 24: @matchai Thoughts on whether this would be considered breaking?

Right now, the defaults is to not display on localhost and to display in bold dimmed green when SSH'ed.

With current proposed changes, we will not display on localhost and display in bold yellow when SSH'ed.

Possible chance would be to switch ssh_style and style.

chipbuster (commented):
After discussions on #3544, I need to check if this is breaking or not.

PatrickChoDev:
I'd like to remake the module (new idea and implementations, so I'll close this PR.)

PR #3227, opened 2021-11-11
ci(chglog): Update chglog and clean up config

andytom:
#### Description

Have updated the version of chglog we use in the deployment pipeline to
the most recent version and updated the config order the config
correctly.

#### Motivation and Context

Related to #3205

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -126,7 +126,7 @@ jobs:
       - name: Setup | Go
         uses: actions/setup-go@v2
         with:
-          go-version: "^1.15.7"
+          go-version: "1.16"
 
       - name: Setup | Artifacts
         uses: actions/download-artifact@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -136,7 +136,7 @@ jobs:
 
       - name: Setup | Release notes
         run: |
-          GO111MODULE=on go get github.com/git-chglog/git-chglog/cmd/git-chglog@0.9.1
+          GO111MODULE=on go get github.com/git-chglog/git-chglog/cmd/git-chglog@0.15.0
```
davidkna, line 139: If you're updating the chglog version, you might also need to update the go version.

davidkna: In addition, the tag needs to be changed and it looks like `go get` has been deprecated in favour of `go install`:
```suggestion
          go install github.com/git-chglog/git-chglog/cmd/git-chglog@v0.15.0
```

andytom: I've bumped the version of go to the same version that `git-chglog` use in their [CI workflow](https://github.com/git-chglog/git-chglog/blob/master/.github/workflows/ci.yml#L15).

Thanks I didn't know `go get` was deprecated, I've been out of the Go ecosystem for longer than I thought.

davidkna (approved):

.github/workflows/deploy.yml
davidkna: I think the warning was added with go 1.17, so this seems to a fairly recent development.

PR #3244, opened 2021-11-17
feat: Install script check for last installed version checksum before installing

### LINKED ISSUE #3243: Make install script checksum before installing

I have a script which update starship along with some other manually installed packages, currently starship just reinstall each time which require root permission.

Make the install script checksum before elevate root.

haoadoreorange:
#### Description

Install script check for last installed version checksum before installing

#### Motivation and Context

Closes #3243

- [X] I have tested using **Linux** (Not a test though I quickly ran it on my machine)

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -149,24 +149,29 @@ elevate_priv() {
 install() {
   ext="$1"
 
-  if test_writeable "${BIN_DIR}"; then
-    sudo=""
-    msg="Installing Starship, please wait…"
-  else
-    warn "Escalated permissions are required to install to ${BIN_DIR}"
-    elevate_priv
-    sudo="sudo"
-    msg="Installing Starship as root, please wait…"
-  fi
-  info "$msg"
-
   archive=$(get_tmpfile "$ext")
 
   # download to the temp file
   download "${archive}" "${URL}"
-
-  # unpack the temp file to the bin dir, using sudo if required
-  unpack "${archive}" "${BIN_DIR}" "${sudo}"
+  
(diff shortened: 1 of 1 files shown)


matchai:
Please update your PR to use the provided PR template.

haoadoresorange:
updated !

haoadoresorange:
@matchai can you review it please ?

chipbuster (changes requested):
For some reason, GitHub will not let me make individual review comments on this PR (literally just this PR, other PRs work fine), so you get one giant blob comment instead of any sane sort of breakdown.

Line numbers refer to lines in the proposed changes unless otherwise stated.

### Using `has`

On line 158, you use both `sha256sum` and `awk` to generate the checksum. I'd like to see these wrapped in a check to `has` (see e.g. line 46 where we use it to check if mktemp exists).

I honestly don't expect that it'll ever actually trigger, but we do support some relatively rare platforms, and I'd like to make sure we get a graceful message if we don't have those tools.

### Double brackets

This install script is intended to be run as a POSIX script, so double brackets (`[[` `]]`) are not supported.

### Failure behaviour

Right now, if the checksums don't match, the install script will print "Starship installed" (line 498) without actually installing anything. This isn't great. (This will also become an issue when you add `has` checks from the first section: what will you do if those checks fail?)

Ideally I'd like to see an interactive option to continue, paired with a command line option (e.g. `--no-checksum`) which allows a user to skip checksum validation pre-emptively. However, I would also accept just printing an error message and exiting the script immediately---something that isn't saying "Install successful" without actually installing anything.

PR #3260, opened 2021-11-22
feat(directory): Allow to disable contractions

### LINKED ISSUE #3249: Make path contraction optional

I'd like to have the option to disable path contraction, i.e., if I'm in `$HOME/foo`, $path is `/home/user/foo` and not `~/foo`.

fgsch:
#### Description

Add `path_contraction` and `repo_contraction` to control whether contractions are enabled for path and repos, respectively. Enabled by default (no behaviour change).

#### Motivation and Context

Closes #3249.

TBC

#### Checklist:

### DIFF
--- a/src/configs/directory.rs
+++ b/src/configs/directory.rs
@@ -20,6 +20,8 @@ pub struct DirectoryConfig<'a> {
     pub read_only_style: &'a str,
     pub truncation_symbol: &'a str,
     pub home_symbol: &'a str,
+    pub path_contraction: bool,
+    pub repo_contraction: bool,
 }
 
 impl<'a> Default for DirectoryConfig<'a> {
@@ -39,6 +41,8 @@ impl<'a> Default for DirectoryConfig<'a> {
             read_only_style: "red",
             truncation_symbol: "",
             home_symbol: "~",
+            path_contraction: true,
+            repo_contraction: true,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


fgsch:
Hi. This still requires documentation and other updates but wanted to check first if the direction of this PR is acceptable, or you will prefer a different approach (assuming this is a feature you want to have). Thanks.

chipbuster:
Hey @fgsch, I'm really sorry about the delayed response to this!

I think this is a fine feature to add, but I'm a little concerned about the API. The directory module already has quite a few options, now we're adding two more options. What if someone wants to add another contraction option in the future? Do we have three booleans just to control that?

One idea I have (though I'm not sure if it's sound) is to have an array of strings which controls what gets contracted to, e.g. `contractions = ["repo", "home"]`. Your desired config would be expressed as `contractions = []`. I have a few concerns with this idea though:

1. Should the order of keys matter?
2. How do we deal with unrecognized keys in the config?
3. Will this potentially make it harder to generate/override defaults, since we have a non-empty default option?
4. Does this make a potential type error possible when parsing the config? E.g. if I put `contractions = [1, "aString", 2022-01-01]`, what happens when Rust tries to parse this? Do we just break and never successfully parse the config? If so, that might not be acceptable.

Anyways, just some thoughts with trying to avoid adding too many keys into the directory module. Perhaps it'll turn out that your idea is the better one---I haven't thought through things very carefully here and would appreciate your thoughts on this style.

davidkna:
Isn't `repo_contraction` equivalent to the existing `truncate_to_repo` option? Given that `truncate_to_repo` it might be difficult to introduce a new option for this without a breaking change. The name for `path_contraction` should also be aligned to `truncate_home`.

fgsch:
@chipbuster @davidkna It's been a while since I opened this so let me refresh my memory and I will get back to you. Thanks!

PR #2553, opened 2021-04-03
docs: update install instructions

vladimyr:
#### Description

Use POSIX `sh`, mitigate possible timing attacks.

#### Motivation and Context

Install script became (almost fully) POSIX compliant with #2228 landing in. This alters advertised installation snippets as discussed here: https://github.com/starship/starship/pull/2228#issuecomment-794882078

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #3268, opened 2021-11-23
Chage the curl command sintax

victorhck:
This way can be running in fish shell without errors messages

#### Description

When you run the command in fish shell will prompt an error. With this workaround is fixed and can be executed the curl command without errors. Tested in fish (my current shell), bash and zsh.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ shown below. Can't see yours? Have a look at the [extra platform instructions](h
    ##### From prebuilt binary, with Shell:
 
    ```sh
-   sh -c "$(curl -fsSL https://starship.rs/install.sh)"
+   curl -fsSL https://starship.rs/install.sh | sh
    ```
 
    To update the Starship itself, rerun the above script. It will replace the current version without touching Starship's configuration.
@@ -173,7 +173,7 @@ shown below. Can't see yours? Have a look at the [extra platform instructions](h
    **Note** - The defaults of the install script can be overridden see the built-in help.
 
    ```sh
-   sh -c "$(curl -fsSL https://starship.rs/install.sh)" -- --help
+   curl -fsSL https://starship.rs/install.sh | sh -s -- --help
    ```
 
    #### Install via Package Manager


README.md
```diff
@@ -165,15 +165,15 @@ shown below. Can't see yours? Have a look at the [extra platform instructions](h
    ##### From prebuilt binary, with Shell:
 
    ```sh
-   sh -c "$(curl -fsSL https://starship.rs/install.sh)"
+   curl -fsSL https://starship.rs/install.sh | sh
```
andytom, line 168: It looks like we changed this in #2553 to avoid potential attacks, is there some way to avoid timing attacks but still be compatible with fish?

davidkna: `sh -c 'sh -c "$(curl -fsSL https://starship.rs/install.sh)"'` would work.

victorhck: > sh -c 'sh -c "$(curl -fsSL https://starship.rs/install.sh)"'

Yes, it works in fish! 👍

matchai: If I were to see that script, I'd probably think the repetition was a typo, if I'm honest.
 
I don't know that we should set the precedent making our init script support fish. It would probably lead to us making similar changes to scripts in our docs.
Most users of fish are aware that they need to run bash scripts with either [bass](https://github.com/edc/bass) or a bourne-compatible shell.

chipbuster: My vote is to leave this as is: we also support xonsh, nu, ion, and powershell--it's highly unlikely that we'll find syntax that works for all of them.

I think the correct way to handle this is to create a setup page as has been discussed on the v2 page, with multiple collapsible sections per-shell.

victorhck: Ok. Then you can close this PR.

chipbuster: @victorhck Sorry! :grimacing: There's a lot of different concerns that go into our main documentation page due to the sheer number of shells we support--it's why we'd like to eventually move to a per-shell instruction setup.

Hope to see other contributions from you in the future!

PR #3274, opened 2021-11-26
feat: Add package version detection for sbt projects

pejax:
#### Description
This patch enables version detection for projects configured by the interactive build tool SBT.

#### Motivation and Context

The implementation uses a regex which is applied line by line to the configuration file 'build.sbt'.
Closes #

#### Checklist:


andytom (commented):
Hi @pejax, thank you for your PR, can you revert the changes to the non English versions of the documentation as this causes problems for the translation too we use.

andytom (approved):

PR #3276, opened 2021-11-26
feat(git_status): Add/change configuration variables in `git_status` module

ajschmidt8:
#### Description

Prior to this change, there was no distinction between deleted, modified, or added files in the working directory versus the staging area for the `git_status` module. This PR addresses that by updating the `git_status` module to include the following changes:

- Removes the top-level `deleted`, `modified` and `renamed` options
- Adds a top-level `work_dir` option
- Makes the following variables available in the `staged` configuration option string:
  - `added_count`
  - `deleted_count`
  - `modified_count`
  - `renamed_count`
  - `count`
- Makes the following variables available in the `work_dir` configuration option string:
  - `added_count`
  - `deleted_count`
  - `modified_count`
  - `count`

**Note**: This should be considered a breaking change since it removes/reworks some existing options in the `git_status` module.

#### Motivation and Context

I'm coming from the [powerlevel10k ](https://github.com/romkatv/powerlevel10k/) `zsh` theme where there is a distinction between working directory changes and staging area changes. This distinction seems more logical to me than `starship`'s current behavior of grouping together change counts between both `git` trees. Hopefully you all agree :slightly_smiling_face:.

**Disclaimer**: This is my first Rust PR, so I'm open to any change suggestions no matter how large or small.

#### Checklist:


ajschmidt8:
@matchai, any thoughts on these changes?

matchai:
Hey AJ 👋 
I really appreciate the time you took to extend the functionality of `git_status`.
It looks like this change makes `git_status` configuration a fair bit more complex. As much as I value the flexibility being proposed, I'm afraid this may reduce approachability even further for an already complicated module to configure.

Perhaps we can add a toggle with advanced configuration like we have for the `directory` module. 🤔 
Ideally, I think this level of granular configuration would best be added as part of #2502.

ajschmidt8:
> Hey AJ I really appreciate the time you took to extend the functionality of `git_status`. It looks like this change makes `git_status` configuration a fair bit more complex. As much as I value the flexibility being proposed, I'm afraid this may reduce approachability even further for an already complicated module to configure.
> 
> Perhaps we can add a toggle with advanced configuration like we have for the `directory` module. thinking Ideally, I think this level of granular configuration would best be added as part of #2502.

@matchai, I understand your concerns about the complexity. I'll close this PR for now. If you reconsider, leave a comment and I'll re-open this and fix the unit tests. I'm a bit bummed though, because it seems that both [powerlevel10k](https://github.com/romkatv/powerlevel10k/) and [oh-my-posh](https://github.com/jandedobbeleer/oh-my-posh) handle their Git status module in the way I've described in this PR. `starship` seems to be the oddball here. The current Git module setup for `starship` doesn't seem very practical in describing the current `git` tree status.

PR #3277, opened 2021-11-26
feat: add support for cmd

### LINKED ISSUE #1766: Cmder (clink) prompt support
## Feature Request
Add Support for Cmder (cmd with clink) prompt

Can the starship prompt be integrated with cmder clink lua init files?

I am using clink lua prompt files currently.

rashil2000:
#### Description
Add support for Cmd shell on Windows, utilizing the [Clink](https://chrisant996.github.io/clink/clink.html) framework.

#### Motivation and Context

Closes #1766

cmd2

#### Checklist:


rashil2000:
Along with RightPrompt (added in this PR), Cmd also supports TransientPrompt and TransientRightPrompt. I haven't added those here yet because, if my knowledge is correct, Starship doesn't support transience yet, right?

chipbuster:
That is correct--it does not support transience (at least, not in the way I'm used to using the term for ZSH. I'm not sure if it means something different in Cmd).

rashil2000:
Yes, I'm referring to the same thing as Zsh.

rashil2000:
After some more digging, I figured transience has more to do with how the shell handles prompts, rather than Starship. 

As a result, I have added TransientPrompt and TransientRightPrompt support for Cmd. You'll notice in the commit that there are very little relevant changes, that too only in the init script - no Rust code was involved. Also updated the docs accordingly.

cmd2

rashil2000:
Clink also supports [Asynchronous Prompts](https://chrisant996.github.io/clink/clink.html#asynchronous-prompt-filtering), but I left it out for now because it's quite involved.

docs/advanced-config/README.md
```diff
@@ -10,6 +10,78 @@ The configurations in this section are subject to change in future releases of S
 
 :::
 
+## TransientPrompt and TransientRightPrompt in Cmd
```
matchai, line 13: I'm afraid single-shell functionality like this is strictly out of scope for Starship.

rashil2000: As I said, it doesn't really have anything to do with Starship. Starship just provides a pretty formatted text to use, even if I add a functionality in Starship, called `--transient` (similar to `--right`, which at the moment is only supported by 3 out of 9 shells), it will be upto the shell to employ it.

Zsh already supports it (but someone needs to enable it in the init script), and it was added to Fish recently too - https://github.com/fish-shell/fish-shell/issues/7602 (not sure if it's released or not).

Besides, doesn't the Advanced Config section of Readme already have features that are present in only 1-2 shells? I can only see Bash, Zsh and PowerShell there.

rashil2000:
Is there anything wrong with this PR? It's been sitting for a while now with no feedback.

I have tried to address @matchai's question, is there anything else I should do?

chipbuster:
@rashil2000 Sorry, we're a little overwhelmed by PRs right now. I'm trying to grind through them over my break, but it's going slower than I'd like (especially since a lot of PRs are new modules, and those have expansive review requirements).

rashil2000:
No worries @chipbuster, thanks for the update.

docs/.vuepress/config.js
```diff
@@ -55,13 +55,13 @@ module.exports = {
   head: [
     ["link", { rel: "icon", href: "/icon.png" }],
     ["meta", { property: "og:title", content: "Starship: Cross-Shell Prompt" }],
-    ["meta", { property: "og:description", content: "Starship is the minimal, blazing fast, and extremely customizable prompt for any shell! Shows the information you need, while staying sleek and minimal. Quick installation available for Bash, Fish, ZSH, Ion, Tcsh, Elvish, Nu, Xonsh, and Powershell."}],
+    ["meta", { property: "og:description", content: "Starship is the minimal, blazing fast, and extremely customizable prompt for any shell! Shows the information you need, while staying sleek and minimal. Quick installation available for Bash, Fish, ZSH, Ion, Tcsh, Elvish, Nu, Xonsh, Cmd and Powershell."}],
```
chipbuster, line 58: I'm gonna be that guy...can we Oxford comma this to maintain consistency with the previous slug? And on line 64 as well.

```suggestion
    ["meta", { property: "og:description", content: "Starship is the minimal, blazing fast, and extremely customizable prompt for any shell! Shows the information you need, while staying sleek and minimal. Quick installation available for Bash, Fish, ZSH, Ion, Tcsh, Elvish, Nu, Xonsh, Cmd, and Powershell."}],
```

docs/README.md
```diff
@@ -16,7 +16,7 @@ footer: ISC Licensed | Copyright © 2019-present Starship Contributors
 
 # Used for the description meta tag, for SEO
 metaTitle: "Starship: Cross-Shell Prompt"
-description: Starship is the minimal, blazing fast, and extremely customizable prompt for any shell! Shows the information you need, while staying sleek and minimal. Quick installation available for Bash, Fish, ZSH, Ion, Tcsh, Elvish, Nu, Xonsh, and PowerShell.
+description: Starship is the minimal, blazing fast, and extremely customizable prompt for any shell! Shows the information you need, while staying sleek and minimal. Quick installation available for Bash, Fish, ZSH, Ion, Tcsh, Elvish, Nu, Xonsh, Cmd and PowerShell.
```
chipbuster, line 19: And here as well.

```suggestion
description: Starship is the minimal, blazing fast, and extremely customizable prompt for any shell! Shows the information you need, while staying sleek and minimal. Quick installation available for Bash, Fish, ZSH, Ion, Tcsh, Elvish, Nu, Xonsh, Cmd, and PowerShell.
```

docs/advanced-config/README.md
chipbuster: I'm of a mind with @matchai on this, at least as far as this file goes. Most other sections here are 20-40 line vignettes with a fair bit of whitespace, for functionality that is at least present in one or two other shells. This is 70 lines of fairly-dense text (dense = "not a lot of whitespace", not "hard to read"), which inflates the file line count by almost 50%, for details of how to do a single feature in a single shell.

I think this is important documentation to have, I just don't think this file is the right place for it in this PR. There are a few places we could put this:

  - Abuse the per-platform nature of the `Advanced Installation` guide to add this there (though I don't think that's a great idea either)
  - Separate it into a GitHub discussion thread
  - If the other shells are going to have similarly-complex configuration for transient mode, we should probably collect the documentation for all of them in a common place.

@matchai At the rate we're going, we're going to need some way to formalize the support differences between shells. Do you think now is a good time to set that up?

chipbuster: @rashil2000 For the moment, is there a way you can preserve this text without introducing it into the PR? Heck, you can open a GitHub discussion/gist and just paste it there for now (and add a link here) if you're okay with that. I want this information to be preserved, I just don't think this file is the right place to be displaying it.

src/main.rs
```diff
@@ -27,7 +27,7 @@ fn main() {
         .long("pipestatus")
         .value_name("PIPESTATUS")
         .help("Status codes from a command pipeline")
-        .long_help("Bash and Zsh supports returning codes for each process in a pipeline.")
+        .long_help("Bash, Fish and Zsh support returning codes for each process in a pipeline.")
```
chipbuster, line 30: Thanks for the catch!

src/init/mod.rs
```diff
@@ -172,6 +176,7 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
             r#"execx($({} init xonsh --print-full-init))"#,
             starship.sprint_posix()?
         ),
+        "cmd" => print_script(CMD_INIT, &StarshipPath::init()?.sprint_cmd()?),
```
chipbuster, line 179: Similarly, call it `CMDEXE_INIT` (or something similar)

src/init/mod.rs
```diff
@@ -46,6 +46,10 @@ impl StarshipPath {
             .map(|s| s.replace("'", "''"))
             .map(|s| format!("'{}'", s))
     }
+    /// Cmd specific path escaping
+    fn sprint_cmd(&self) -> io::Result<String> {
```
chipbuster, line 50: Can we call the shell `cmdexe` or something similar when it's not obvious that it's a shell from context? (e.g. the addition of `Shell::Cmd` is fine because it's obviously a shell in that context). I'm afraid in two years, I'm going to be reading this code and get confused by `cmd` as an abbreviation for "Command" and `cmd` meaning the shell.

src/init/mod.rs
```diff
@@ -279,4 +287,25 @@ mod tests {
         assert_eq!(starship_path.sprint_pwsh()?, r"'C:\''starship.exe'");
         Ok(())
     }
+
+    #[test]
+    fn escape_cmd() -> io::Result<()> {
```
chipbuster, line 292: Similar name changes here (`escape_cmd` sounds like we're testing escaping inputs to an external command).

src/modules/custom.rs
```diff
@@ -247,6 +248,15 @@ fn handle_powershell(command: &mut Command, shell: &str, shell_args: &[&str]) {
     }
 }
 
+#[cfg(windows)]
+fn handle_cmd(command: &mut Command, shell: &str, shell_args: &[&str]) {
```
chipbuster, line 252: And another cmd -> cmdexe sub here.

chipbuster (commented):
@rashil2000 Sorry for the delay! I've made some notes on code (mostly just about me getting confused in a few years as to whether `cmd` means "command" or "cmd.exe"), but the code quality mostly seems fine.

I think we need to decide what to do with that Transient Prompt section. I agree with @matchai that it shouldn't go into the advanced config doc as it is now, but I don't know if the answer is to create per-shell config guides, or to create a big Transient Prompt guide, or...something else. Would you be okay with holding onto the text for now? We could do a gist, a GitHub discussion, I'd even be okay with you removing it from this PR and opening a new PR with the text in there. Something while we work out the best structure for this.

Finally, I want to actually test this out myself before giving final approval (hopefully relatively soon now that I've burned through some of my review backlog, but not quite tonight).

docs/advanced-config/README.md
rashil2000: I have reverted the commit

rashil2000:
Review comments addressed and master merged.

rashil2000:
The AWS test `expiration_date_set_from_file` is failing, but I haven't touched that part of the code.

Is this common - https://github.com/starship/starship/pull/3157#issuecomment-1002860756?

chipbuster (approved):

chipbuster:
I'm not gonna lie: I still can't quite believe this works. But it works beautifully. Thanks @rashil2000 for the code contributions!

rashil2000:
Neither can I! 

Clink is one of those rare magical tools that will never cease to amaze me 😅

Thank you for the review :)

src/modules/custom.rs
```diff
@@ -247,6 +248,15 @@ fn handle_powershell(command: &mut Command, shell: &str, shell_args: &[&str]) {
...
 
+#[cfg(windows)]
+fn handle_cmdexe(command: &mut Command, shell: &str, shell_args: &[&str]) {
+    let is_cmd = shell.ends_with("cmd.exe") || shell.ends_with("cmd");
+
+    if is_cmd && shell_args.is_empty() {
+        command.arg("/C");
+    }
+}
+
```
davidkna, line 259: I just tried this, and it doesn't seem to work with a simple custom module of just `echo test`. I think the issue is that the command gets passed via stdin. See also the discussion in #2624 and #2629.

rashil2000: Does it work if you don't call this function at all (in `fn shell_command()`)? I will remove it then.

davidkna: I believe the output will include the prompt in that case. It might be best to just `log::error!` and return if the shell is detected as `cmd`.

rashil2000: I think the reason the prompt gets shown is that cmd shell gets invoked without any params, i.e. `cmd echo test`. 

I have made the requested change.

src/modules/custom.rs
```diff
@@ -160,6 +160,11 @@ fn shell_command(cmd: &str, shell_args: &[&str]) -> Option<Output> {
             .stdout(Stdio::piped())
             .stderr(Stdio::piped());
 
+        if forced_shell.ends_with("cmd.exe") || forced_shell.ends_with("cmd") {
+            log::error!("Custom command output with Command shell is not supported.");
```
davidkna, line 164: I would prefer calling it `cmd` and I think this should include a troubleshooting suggestion.

```suggestion
            log::error!("Executing custom commands with the cmd shell is not currently supported. Please set a different shell with the \"shell\" option.");
```

src/bug_report.rs
```diff
@@ -213,6 +213,7 @@ fn get_config_path(shell: &str) -> Option<PathBuf> {
             "elvish" => Some(".elvish/rc.elv"),
             "tcsh" => Some(".tcshrc"),
             "xonsh" => Some(".xonshrc"),
+            "cmd" => Some("starship.lua"),
```
davidkna, line 216: I think this is the usual location:
```suggestion
            "cmd" => Some("AppData/Local/clink/starship.lua"),
```
But maybe this should be set to `.inputrc` or `AppData/Local/clink/clink_settings`?

rashil2000: > I think this is the usual location:

I left out the `AppData/Local/clink/` part because this location is configurable in multiple ways (through `clink.path` setting, or through `clink installscripts` command)

> But maybe this should be set to `.inputrc` or `AppData/Local/clink/clink_settings`?

I don't this is relevant for Starship. FWIW, .inputrc is also there for bash, but we don't use it. Similarly clink_settings file mostly stores stuff other than prompt (except transience, which we removed in this PR). `AppData/Local/clink/starship.lua` is a better fit.

src/modules/custom.rs
rashil2000: Although I have made the requested change here, there's one important thing to note: the `when` condition of custom command works correctly with cmd shell. It's just that the `command` field (whose output gets shown actually) is problematic. This is why I mentioned _Custom command output_ in my original error message.

rashil2000:
Hi!

Looks like there was a pretty big refactor some days ago. I've precariously resolved the merge conflicts 😅

Let me know if this is good to go.

src/modules/custom.rs
davidkna: The log message is going to annoy them either way, so I'm not sure how helpful rewording this will be. Unless this can be restricted not to show up for `when` maybe this should just become an `info` log. Does anyone else have any ideas?

src/bug_report.rs
davidkna: > I left out the `AppData/Local/clink/` part because this location is configurable in multiple ways (through `clink.path` setting, or through `clink installscripts` command)

I know, but I suggested this because the other option was even more likely to be wrong.

> I don't think this is relevant for Starship. FWIW, .inputrc is also there for bash, but we don't use it. Similarly clink_settings file mostly stores stuff other than prompt (except transience, which we removed in this PR). `AppData/Local/clink/starship.lua` is a better fit.

Thanks for this info. I wasn't very sure what each of these files was responsible for.

src/modules/custom.rs
rashil2000: This brings me to a question: when we specify shell explicitly through the `shell` option, is that explicit shell used both for `when` and `command`, or just for `command`? 

If it's the former, then I guess we don't really have a choice and we can keep the current message. 

However, if it's the latter (where `when` will always be executed through cmd.exe, and `command` will get executed through the explicitly specified `shell`), then we will need to be more verbose here as to what is actually not allowed/supported.

davidkna: They both use `shell`

rashil2000: Then I think this is good to go

davidkna: In that case, can you move the message into [the `"output" =>` part of `map_no_escaping`](https://github.com/starship/starship/blob/808ddfca7a067b6625a4fe2869a36fcbb78ad9e7/src/modules/custom.rs#L60-L69) and gate it with `context.shell == Shell::Cmd && config.shell == None`?

rashil2000: Done

davidkna (approved):

chipbuster:
Thanks @rashil2000 for sticking with the changeset and @davidkna for the reviews!

Can't wait to see this in production :)

saumyajyoti:
Thanks @rashil2000  for the enhancement.

PR #3287, opened 2021-12-01
perf(git_status): tweak exec flags to omit unnecessary info

arcnmx:
`git status` can be prohibitively slow on some repos, so allow the config to influence what flags are passed to git. For instance, if there is no configured symbol for untracked files, tell git to omit them from its output.

#### Description

Note that since this requires the user to opt-in to omitting certain details from the module (and does not change the defaults), this is not really a "fix" for #2625.

A follow-up to this might be to read `git config` data to customize this behaviour per-repo. There is precedent for this in [oh-my-zsh git prompts](https://github.com/ohmyzsh/ohmyzsh/blob/e86c6f5e7fc9f024a427e2870ab70644b5454725/lib/git.zsh#L40), so looking for the config entries that people are already using in the wild might be nice.

#### Motivation and Context

With the repos I mentioned in [my comment](https://github.com/starship/starship/issues/2625#issuecomment-982206526) as a test case, here are some random before and after comparisons:

- small repo with 5 submodules of varying sizes: **163.2ms** on 1.0.0 vs **5.3ms** with this patch
- small repo with no submodules: **5.2ms** vs **5.1ms**
- medium repo with no submodules: **10.2ms** vs **3.9ms**
- large repo with no submodules: **164.8ms** vs **18.7ms**

(testing done with ahead/behind/up_to_date/diverged/untracked cleared and ignore_submodules=true)

#### Checklist:

### DIFF
--- a/src/configs/git_status.rs
+++ b/src/configs/git_status.rs
@@ -18,6 +18,7 @@ pub struct GitStatusConfig<'a> {
     pub modified: &'a str,
     pub staged: &'a str,
     pub untracked: &'a str,
+    pub ignore_submodules: bool,
     pub disabled: bool,
 }
 
@@ -37,6 +38,7 @@ impl<'a> Default for GitStatusConfig<'a> {
             modified: "!",
             staged: "+",
             untracked: "?",
+            ignore_submodules: false,
             disabled: false,
         }
     }
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -27,10 +27,10 @@ const ALL_STATUS_FORMAT: &str = "$conflicted$stashed$deleted$renamed$modified$st
 ///   - `»` — A renamed file has been added to the staging area
 ///   - `✘` — A file's deletion has been added to the staging area
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
-    let info = Arc::new(GitStatusInfo::load(context));
-
     let mut module = context.new_module("git_status");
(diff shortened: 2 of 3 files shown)


src/configs/git_status.rs
```diff
@@ -18,6 +18,7 @@ pub struct GitStatusConfig<'a> {
     pub modified: &'a str,
     pub staged: &'a str,
     pub untracked: &'a str,
+    pub ignore_submodules: bool,
```
matchai, line 21: Since this PR introduces a new `ignore_submodules` option, we would need to have documentation and tests added for this to merge.

src/modules/git_status.rs
```diff
@@ -181,21 +183,37 @@ impl<'a> GitStatusInfo<'a> {
...
-    )?;
+    let mut args = vec![
+        OsStr::new("-C"),
+        context.current_dir.as_os_str(),
+        OsStr::new("--no-optional-locks"),
+        OsStr::new("status"),
+        OsStr::new("--porcelain=2"),
+    ];
+
+    // for performance reasons, only pass flags if necessary...
```
matchai, line 198: Looks great to me! 👍 
Seems so obvious in hindsight! Thank you for tackling this 😄

matchai (commented):
Looks good to me so far! All that's left is adding tests and docs for the new `ignore_submodules` option. 😊

andytom (approved):
LGTM

PR #3289, opened 2021-12-01
feat(localip): add module to print the current ipv4 address

quickstar:
#### Description
This PR adds support for printing your current ip to the prompt. By default it behaves like the hostname module and is therefore only rendered if an SSH session is present.

The way it obtains the localip is by opening a UDP socket to `8.8.8.8:80` and connecting to it. This is a very fast cross platform way to determine your primary interface's IP address.

#### Motivation and Context
Since I'm using SSH a lot, it is very handy to have the IP of the remote machine always shown. 

image

- [X] I have tested using **MacOS** - _~I don't have a Mac, maybe someone can help out? The Unittests are working though.~_

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -192,7 +192,7 @@ format = "$all"
 format = """
 $username\
 $hostname\
-$localipv4\
+$localip\
 $shlvl\
 $singularity\
 $kubernetes\
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -192,7 +192,7 @@ format = "$all"
 format = """
 $username\
 $hostname\
-$localip\
+$localipv4\
 $shlvl\
 $singularity\
 $kubernetes\
@@ -1811,7 +1811,7 @@ The `localip` module shows the IPv4 address of the primary network interface.
 | Option     | Default                     | Description                                                                                                                          |
 | ---------- | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
 | `ssh_only` | `true`                      | Only show IP address when connected to an SSH session.                                                                               |
-| `format`   | `"[$hostname]($style)"`     | The format for the module.                                                                                                           |
+| `format`   | `"[$localipv4]($style) "`     | The format for the module.                                                                                                           |
 | `style`    | `"bold yellow"`              | The style for the module.                                                                                                            |
 | `disabled` | `false`                     | Disables the `localip` module.                                                                                                      |
 
@@ -1819,7 +1819,6 @@ The `localip` module shows the IPv4 address of the primary network interface.
 
 | Variable | Example | Description                          |
 | -------- | ------- | ------------------------------------ |
-| symbol   |         | Mirrors the value of option `symbol` |
 | style\*  |         | Mirrors the value of option `style`  |
 
 \*: This variable can only be used as a part of a style string
@@ -1831,7 +1830,7 @@ The `localip` module shows the IPv4 address of the primary network interface.
 
 [localip]
 ssh_only = false
-format =  "@[$localip](bold red) "
+format =  "@[$localipv4](bold red) "
 disabled = false
 ```
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1819,7 +1819,7 @@ The `localip` module shows the IPv4 address of the primary network interface.
 
 | Variable    | Example      | Description                          |
 | --------    | -------      | ------------------------------------ |
-| localipv4\* | 192.168.1.13 | Contains the primary IPv4 address    |
+| localipv4 | 192.168.1.13 | Contains the primary IPv4 address    |
 | style\*     |              | Mirrors the value of option `style`  |
 
 \*: This variable can only be used as a part of a style string
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -192,7 +192,7 @@ format = "$all"
 format = """
 $username\
 $hostname\
-$localip\
+$localipv4\
 $shlvl\
 $singularity\
 $kubernetes\
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -192,7 +192,7 @@ format = "$all"
 format = """
 $username\
 $hostname\
-$localipv4\
+$localip\
 $shlvl\
 $singularity\
 $kubernetes\
@@ -1817,9 +1817,10 @@ The `localip` module shows the IPv4 address of the primary network interface.
 
 ### Variables
 
-| Variable | Example | Description                          |
-| -------- | ------- | ------------------------------------ |
-| style\*  |         | Mirrors the value of option `style`  |
+| Variable    | Example      | Description                          |
+| --------    | -------      | ------------------------------------ |
+| localipv4\* | 192.168.1.13 | Contains the primary IPv4 address    |
+| style\*     |              | Mirrors the value of option `style`  |
 
 \*: This variable can only be used as a part of a style string
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3005,12 +3005,22 @@ This module is not supported on nu shell.
 
 ### Options
 
-# | Option | Default | Description || ----------------------- | ----------------------------------------------------------------------------- | ------------------------------------------------------- | --------------------------------------------------- || `format` | `"[$symbol$status]($style) "` | The format of the module || `symbol` | `"✖"` | The symbol displayed on program error || `success_symbol` | `"✔️"` | The symbol displayed on program success || `not_executable_symbol` | `"🚫"` | The symbol displayed when file isn't executable || `not_found_symbol` | `"🔍"` | The symbol displayed when the command can't be found || `sigint_symbol` | `"🧱"` | The symbol displayed on SIGINT (Ctrl + c) || `signal_symbol` | `"⚡"` | The symbol displayed on any signal || `style` | `"bold red"` | The style for the module. || `recognize_signal_code` | `true` | Enable signal mapping from exit code || `map_symbol` | `false` | Enable symbols mapping from exit code || `pipestatus` | `false` | Enable pipestatus reporting |<<<<<<< HEAD| `pipestatus_separator` | `|` |
-
-| `pipestatus_separator` | `|` | The symbol that separate in pipe program exit codes |
->>>>>>> 7c1ac0c8 (feat(localip): add module to print the current ipv4 address)
-| `pipestatus_format` | `\\[$pipestatus\\] => [$symbol$common_meaning$signal_name$maybe_int]($style)` | The format of the module when the command is a pipeline |
-| `disabled` | `true` | Disables the `status` module. |
+| Option                  | Default                                                                       | Description                                             |
+| ----------------------- | ----------------------------------------------------------------------------- | ------------------------------------------------------- |
+| `format`                | `"[$symbol$status]($style) "`                                                 | The format of the module                                |
+| `symbol`                | `"✖"`                                                                         | The symbol displayed on program error                   |
+| `success_symbol`        | `"✔️"`                                                                        | The symbol displayed on program success                 |
+| `not_executable_symbol` | `"🚫"`                                                                         | The symbol displayed when file isn't executable         |
+| `not_found_symbol`      | `"🔍"`                                                                         | The symbol displayed when the command can't be found    |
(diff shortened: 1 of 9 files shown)


chipbuster:
I like the idea of this module, but I'm a little bit wary of hitting `8.8.8.8`. I understand that Google is used to handling DDOS-level traffic, but it still feels a little wasteful to spin up a UDP socket just for a prompt draw.

Is there another crate, or a purely-local method that we could use?

quickstar:
Hi  @chipbuster

Thanks for your thoughts. Initially I was also searching for a "local" way to achieve this. Unfortunately it is very tricky to get the primary NIC, or in other words you need to dig deep in the respective kernels.

Since sockets are very crucial these days, i assume the creation of a socket is blazingly fast on any OS. I also did some simple benchmarks comparing the socket vs. the built in utils like `hostname -I`, `ipconfgi`, `iwconfig`, `ip addr` and I event tried it the powershell way on windows (I don't remember the cmdlet but is was horribly slow). Interstingly enough it seems that most of the tools internally also use the socket method to determine the IP of the primary NIC, they just do a lot more in addition, which would explain the slower performance.

docs/config/README.md
```diff
@@ -1753,6 +1754,39 @@ By default the module will be shown if any of the following conditions are met:
...
+## localip
+
+The `localip` module shows the IPv4 address of the primary network interface.
+
+### Options
+
+| Option     | Default                     | Description                                                                                                                          |
+| ---------- | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| `ssh_only` | `true`                      | Only show IP address when connected to an SSH session.                                                                               |
+| `format`   | `"[$hostname]($style)"`     | The format for the module.                                                                                                           |
```
chipbuster, line 1766: I think this should be `$localip` and not `$hostname`.

src/configs/localip.rs
```diff
@@ -0,0 +1,23 @@
...
+    pub format: &'a str,
+    pub style: &'a str,
+    pub disabled: bool,
+}
+
+impl<'a> Default for LocalipConfig<'a> {
+    fn default() -> Self {
+        LocalipConfig {
+            ssh_only: true,
+            format: "@[$localip]($style)",
```
chipbuster, line 18: Is there a reason to have the `@` in here, or did that get copied from the example configuration?

chipbuster: Also, please insert a space at the end of the format string--otherwise subsequent modules get screwed up

```suggestion
            format: "[$localip]($style) ",
```

src/modules/localip.rs
```diff
@@ -0,0 +1,128 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("localip");
+    let config: LocalipConfig = LocalipConfig::try_load(module.config);
+
+    let ssh_connection = context.get_env("SSH_CONNECTION");
+    if config.ssh_only && ssh_connection.is_none() {
+        return None;
+    }
+
+    let localip = local_ipaddress::get().unwrap();
```
chipbuster, line 25: This panics for me as soon as I disconnect my internet.

```
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', src/modules/localip.rs:25:42
```

src/modules/localip.rs
```diff
@@ -0,0 +1,128 @@
...
+    if localip.is_empty() {
+        log::warn!("unable to determine local ipv4 address");
```
chipbuster, line 27: I don't 100% remember the semantics we have for this kind of stuff, but I think `log::warn!` will warn once per prompt draw, so anyone working with their internet disconnected will see this warning every time starship does a draw. It might be better to demote this to an `info!` log so that it doesn't spam when working offline.

docs/config/README.md
```diff
@@ -1753,6 +1754,39 @@ By default the module will be shown if any of the following conditions are met:
...
+| `style`    | `"bold yellow"`              | The style for the module.                                                                                                            |
+| `disabled` | `false`                     | Disables the `localip` module.                                                                                                      |
+
+### Variables
+
+| Variable | Example | Description                          |
+| -------- | ------- | ------------------------------------ |
+| symbol   |         | Mirrors the value of option `symbol` |
```
chipbuster, line 1774: This also seems wrong, since there is no `symbol` option in this module.

In your defense, it seems you copied this from the hostname module, and I think that module is just wrong as well....gonna go open a PR on it once I finish here.

chipbuster (changes requested):
Hey @quickstar, thanks for the PR! It looks like you hit every point on the new module checklist (I still struggle with that myself, the list is just so long...),so thanks for that.

The biggest blocker for me right now is that this module panics all of starship if I just disable all my network interfaces (in that case, I suspect the underlying OS calls fail, since there's no way to actually connect to anything). I think we should fix that and make sure that the warnings aren't too verbose when working offline (should only be a few lines of changes).

Quick question, since I'm not really a networks person: is there any possibility that someone would ever want to extend this to display IPv6 addresses? If so, would it be better to name the current variable `localipv4` and leave extension for ipv6, or is the convention to use no-suffix to mean v4?

chipbuster:
Apparently I suck at checklists, because there actually does appear to be one thing missing: we need preset options for the various presets. I think you can just copy-paste in a default config, since there's not a symbol or anything for this module.

src/modules/localip.rs
chipbuster: It seems I was mistaken about this.

chipbuster:
Sorry, I don't think I explained myself well! I didn't mean to rename the entire module, just the variable used inside of it. I think it makes more sense to keep the IP code together in the same module, we can just have two variables (`localipv4` and `localipv6`) and the users can choose to use either (or both) at some point in the future.

src/modules/localip.rs
quickstar: Would unwrap_or_default() do the trick?

quickstar:
Since the last rebase onto master i get the following error:

```
error[E0658]: use of unstable library feature 'command_access'
   --> src/utils.rs:392:56
    |
392 |             log::info!("Unable to run {:?}, {:?}", cmd.get_program(), error);
    |                                                        ^^^^^^^^^^^
    |
    = note: see issue #44434 <https://github.com/rust-lang/rust/issues/44434> for more information
```

Any ideas?

src/modules/localip.rs
chipbuster: Yep.

chipbuster:
If you haven't already solved it, I suspect the problem is that you're not on the latest version of rust. If you installed with rustup you can do a `rustup update` to get latest stable.

quickstar:
> If you haven't already solved it, I suspect the problem is that you're not on the latest version of rust. If you installed with rustup you can do a `rustup update` to get latest stable.

Tnx, worked like a charm

docs/config/README.md
```diff
@@ -181,6 +181,7 @@ format = "$all"
 format = """
 $username\
 $hostname\
+$localipv4\
```
chipbuster, line 184: This needs to be reverted, I think. Otherwise the module doesn't show up for me at all.
```suggestion
$localip\
```

docs/config/README.md
```diff
@@ -1755,6 +1756,38 @@ By default the module will be shown if any of the following conditions are met:
...
+| ---------- | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
+| `ssh_only` | `true`                      | Only show IP address when connected to an SSH session.                                                                               |
+| `format`   | `"[$localipv4]($style) "`     | The format for the module.                                                                                                           |
+| `style`    | `"bold yellow"`              | The style for the module.                                                                                                            |
+| `disabled` | `false`                     | Disables the `localip` module.                                                                                                      |
+
+### Variables
+
+| Variable | Example | Description                          |
+| -------- | ------- | ------------------------------------ |
```
chipbuster, line 1775: `localipv4` variable should be documented here.

src/configs/starship_root.rs
```diff
@@ -19,6 +19,7 @@ pub struct StarshipRootConfig {
 pub const PROMPT_ORDER: &[&str] = &[
     "username",
     "hostname",
+    "localipv4",
```
chipbuster, line 22: Unclear to me whether this should be `localip` or not (but I think it should?)

chipbuster (changes requested):
A few minor fixes (mostly from me not communicating what I wanted clearly with the rename)

docs/config/README.md
```diff
@@ -1755,6 +1756,39 @@ By default the module will be shown if any of the following conditions are met:
...
+| `ssh_only` | `true`                      | Only show IP address when connected to an SSH session.                                                                               |
+| `format`   | `"[$localipv4]($style) "`     | The format for the module.                                                                                                           |
+| `style`    | `"bold yellow"`              | The style for the module.                                                                                                            |
+| `disabled` | `false`                     | Disables the `localip` module.                                                                                                      |
+
+### Variables
+
+| Variable    | Example      | Description                          |
+| --------    | -------      | ------------------------------------ |
+| localipv4\* | 192.168.1.13 | Contains the primary IPv4 address    |
```
chipbuster, line 1776:
```suggestion
| localipv4 | 192.168.1.13 | Contains the primary IPv4 address    |
```

quickstar: Here we go, tnx for the patience and your time 🙏

chipbuster:
Note to self: pull-merge is a bad strategy and should not be set as default >.>

chipbuster (approved):
Well, now that I accidentally completely butchered the PR, I think we should approve it in its current state without any more changes (the current state is the state I would have approved it in, with an accidental auto-merge from starship master to include all PRs that have been merged between the base of this PR and now).

chipbuster:
Rebased onto master to resolve a merge conflict and keep things pretty.

docs/config/README.md
```diff
@@ -1199,7 +1200,7 @@ This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gc
 | Option           | Default                                                  | Description                                                     |
 | ---------------- | -------------------------------------------------------- | --------------------------------------------------------------- |
 | `format`         | `'on [$symbol$account(@$domain)(\($region\))]($style) '` | The format for the module.                                      |
-| `symbol`         | `"☁️  "`                                                 | The symbol used before displaying the current GCP profile.      |
+| `symbol`         | `"☁️ "`                                                  | The symbol used before displaying the current GCP profile.      |
```
matchai, line 1203: Looks like there may've been an inadvertent change here.
This is intentionally double-spaced.

docs/config/README.md
```diff
@@ -2858,7 +2893,7 @@ set to a number and meets or exceeds the specified threshold.
 | ----------- | ---------------------------- | ------------------------------------------------------------- |
 | `threshold` | `2`                          | Display threshold.                                            |
 | `format`    | `"[$symbol$shlvl]($style) "` | The format for the module.                                    |
-| `symbol`    | `"↕️  "`                     | The symbol used to represent the `SHLVL`.                     |
+| `symbol`    | `"↕️ "`                      | The symbol used to represent the `SHLVL`.                     |
```
matchai, line 2896: Similarly here.

docs/config/README.md
```diff
@@ -1767,6 +1769,39 @@ By default the module will be shown if any of the following conditions are met:
...
+The `localip` module shows the IPv4 address of the primary network interface.
+
+### Options
+
+| Option     | Default                   | Description                                            |
+| ---------- | ------------------------- | ------------------------------------------------------ |
+| `ssh_only` | `true`                    | Only show IP address when connected to an SSH session. |
+| `format`   | `"[$localipv4]($style) "` | The format for the module.                             |
+| `style`    | `"bold yellow"`           | The style for the module.                              |
+| `disabled` | `false`                   | Disables the `localip` module.                         |
```
matchai, line 1783: I'm a little concerned about adding this module as enabled by default.
The prompt, when sshing, already has some default options that lengthen the prompt a fair bit: username and hostname. I worry that we may be making the prompt when sshing a little too long with the inclusion of local IP.
How would you feel about having this be disabled by default?

quickstar: Hi
I don't mind if this is disabled! I run my own config anyways :+1:

matchai (approved):
LGTM! 👍

PR #3290, opened 2021-12-01
fix(git_branch): more robust handling of .git

arcnmx:
Modules that use `context.get_repo()` re-open the repo based on the reported workdir, which can be inaccurate and was a source of bugs. Instead, store the gitdir's path for later use.

#### Description

Each module has its own unique way of handling git repos and what parts of `context::Repo` they use. This can result in inconsistencies across modules where one (git_status) might work while another (git_branch) might inexplicably fail.

#### Motivation and Context

Might be related to #2462. I had many issues with submodules that used relative `gitdir:` paths in their `.git` file.

This attempts to fix things somewhat, though a more consistent approach might be to just store the `git2::Repository` inside the `Repo` directly - is there any specific reason why this isn't currently done?

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -274,7 +274,8 @@ impl<'a> Context<'a> {
             }?;
             Ok(Repo {
                 branch: get_current_branch(&repository),
-                root: repository.workdir().map(Path::to_path_buf),
+                workdir: repository.workdir().map(Path::to_path_buf),
+                path: Path::to_path_buf(repository.path()),
                 state: repository.state(),
                 remote: get_remote_repository_info(&repository),
             })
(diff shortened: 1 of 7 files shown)


matchai:
> This attempts to fix things somewhat, though a more consistent approach might be to just store the `git2::Repository` inside the `Repo` directly - is there any specific reason why this isn't currently done?

Looking back in history, it doesn't look like there is.
We originally hand-picked a couple of values from `Repository` that we needed, not doing anything else with it, but dependence on it gradually grew with time without it having been refactored.

matchai (approved):
Looks good to me! 👍
Thank you for the fix. This should certainly help avoid git bugs of the sort from cropping up 😄

PR #3299, opened 2021-12-07
fix: grammar and layout of CN

miranquil:
#### Description

There's a grammar error in original text. In Chinese we express in different sequence.

Also, I believe the linter made Markdown syntax wrong which caused wrong highlighted words.

#### Motivation and Context

Just help maintain README in my language.

### DIFF
--- a/docs/zh-CN/guide/README.md
+++ b/docs/zh-CN/guide/README.md
@@ -298,7 +298,7 @@
 
 ## 🤝 贡献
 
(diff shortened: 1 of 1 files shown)


matchai:
Thank you for your contribution, @miranquil! 😊
We currently manage translations with Crowdin at http://translate.starship.rs/
Submitting a translation on Crowdin will auto-generate a PR with the updates that are made.

I'll be closing this PR for now, but I encourage you to suggest those same changes over on Crowdin.

PR #3304, opened 2021-12-08
feat: add a container indicator

### LINKED ISSUE #3104: Add toolbox container indicator

haraldh:
#### Description

Adds a container type indicator, if inside a container, detected via the presence of some marker files.

E.g. inside a podman container entered with `toolbox enter`:

```
starship on  container_rebased [$!] is 📦 v1.0.0 via 🦀 v1.56.1
❯ toolbox enter

starship on  container_rebased [$!] is 📦 v1.0.0 via 🦀 v1.56.1
⬢ [fedora-toolbox:35] ❯
```

Additionally a test method for absolute file names is added.

#### Motivation and Context

Closes #2590
Closes #3104

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -208,6 +208,7 @@ $docker_context\
 $package\
 $cmake\
 $cobol\
+$container\
 $dart\
 $deno\
 $dotnet\
@@ -672,6 +673,40 @@ This does not suppress conda's own prompt modifier, you may want to run `conda c
 
 [conda]
 format = "[$symbol$environment](dimmed green) "
+```
+
+## Container
+
+The `container` module displays a symbol and container name, if inside a container.
+
+
+### Options
+
+| Option     | Default                           | Description                               |
+|------------|-----------------------------------|-------------------------------------------|
+| `symbol`   | `"⬢"`                             | The symbol shown, when inside a container |
+| `style`    | `"bold red dimmed"`               | The style for the module.                 |
+| `format`   | "[$symbol \\[$name\\]]($style) "  | The format for the module.                |
+| `disabled` | `false`                           | Disables the `container` module.          |
+
+
+### Variables
+
+| Variable | Example             | Description                          |
+|----------|---------------------|--------------------------------------|
+| name     | `fedora-toolbox:35` | The name of the container            |
+| symbol   |                     | Mirrors the value of option `symbol` |
+| style\*  |                     | Mirrors the value of option `style`  |
+
+\*: This variable can only be used as a part of a style string
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[container]
+format = "[$symbol \\[$name\\]]($style) "
 ```
 
 ## Crystal
--- a/src/modules/container.rs
+++ b/src/modules/container.rs
@@ -1,0 +1,196 @@
+use super::{Context, Module};
+
+#[cfg(not(target_os = "linux"))]
+pub fn module<'a>(_context: &'a Context) -> Option<Module<'a>> {
+    None
+}
+
+#[cfg(target_os = "linux")]
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    use super::RootModuleConfig;
+    use crate::configs::container::ContainerConfig;
+    use crate::formatter::StringFormatter;
+    use crate::utils::read_file;
+
+    pub fn container_name(context: &Context) -> Option<String> {
+        use crate::utils::context_path;
+
+        if context_path(context, "/proc/vz").exists() && !context_path(context, "/proc/bc").exists()
+        {
+            // OpenVZ
+            return Some("OpenVZ".into());
+        }
+
+        if context_path(context, "/run/host/container-manager").exists() {
+            // OCI
+            return Some("OCI".into());
+        }
+
+        if context_path(context, "/run/systemd/container").exists() {
+            // systemd
+            return Some("Systemd".into());
+        }
+
+        let container_env_path = context_path(context, "/run/.containerenv");
+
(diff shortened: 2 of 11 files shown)


jankatins:
Duplications potential with https://github.com/starship/starship/pull/2590

haraldh:
@jankatins ready for review

haraldh:
Removed the external dependency on `systemd-detect-virt` and check some marker files.
Also in the case of `podman`, display the container name taken from `/run/.containerenv`, which basically implements https://github.com/starship/starship/pull/2590.

haraldh:
Additionally a test method for absolute file names is added.

haraldh:
The nightly macOS rust compiler decided to have an ICE.

haraldh:
Issue of the nightly failures: https://github.com/rust-lang/rust/issues/91663

haraldh:
> Issue of the nightly failures: [rust-lang/rust#91663](https://github.com/rust-lang/rust/issues/91663)

CI passed after re-run

jkonecny12:
This looks great. Thanks Harald for writing this!

haraldh:
> This looks great. Thanks Harald for writing this!

@jkonecny12 thanks for the flowers :)

src/modules/container.rs
```diff
@@ -0,0 +1,181 @@
...
+            // systemd
+            return Some("Systemd".into());
+        }
+
+        let container_env_path = context_path(context, "/run/.containerenv");
+
+        if container_env_path.exists() {
+            // podman and others
+
+            let image_res = std::fs::File::open(container_env_path)
```
andytom, line 39: Not a blocker but it is recommended to use  [`utils::read_file`](https://github.com/starship/starship/blob/master/src/utils.rs#L13) for getting the contents of a file unless there is a very specific reason not to.

andytom (commented):
Have left a few suggestions but this mostly looks ok to me. Would like to get some input from other @starship/astronauts regarding the new `context_path` function as there might be something I'm not thinking of at the moment.

src/modules/container.rs
haraldh: fixed

docs/config/README.md
```diff
@@ -670,6 +671,39 @@ This does not suppress conda's own prompt modifier, you may want to run `conda c
...
+## Container
+
+The `container` module displays a symbol and container name, if inside a container.
+
+
+### Options
+
+| Option          | Default             | Description                               |
+|-----------------|---------------------|-------------------------------------------|
+| `symbol`        | `"⬢"`               | The symbol shown, when inside a container |
```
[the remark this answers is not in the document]
haraldh: fixed

docs/config/README.md
```diff
@@ -674,6 +675,40 @@ This does not suppress conda's own prompt modifier, you may want to run `conda c
 format = "[$symbol$environment](dimmed green) "
 ```
 
+## Container
+
+The `container` module displays a symbol and container name, if inside a container.
```
jankatins, line 680: Do I read the code right, that only podman containers (e.g. ones with `/run/.containerenv` which contain a `image=...`line) get a non-generic name? If so it would be nice to describe that somehow.

(currently trying to figure out how this interacts with https://github.com/89luca89/distrobox)

jankatins:
Really looking forward to this one (no detailed review done, just tried to understand the new plugin and what it does and liking it)

andytom (approved):

docs/config/README.md
haraldh:
```
~ via 🐹 v1.16.13 via ☕ v11.0.13 via  via 🐪 v5.34.0 via 🐍 v3.10.1 via 🦀 v1.58.0 
❯ toolbox enter 

~ via 🐹 via ☕ via  via 🐪 v5.34.0 via 🐍 v3.10.0 via 🦀 v1.58.0 
[fedora-toolbox:35] ❯ cat /run/.containerenv
engine="podman-3.4.4"
name="fedora-toolbox-35"
id="1e61eb104cd571b8a8b535542cedcdef225a9f53c6ff24e905b3b2ccc98a98a3"
image="registry.fedoraproject.org/fedora-toolbox:35"
imageid="ab8bc106d4a710a7a27c538762864610467b3559f80b413d30e0a1bfcfe272a5"
rootless=1
graphRootMounted=1
```

haraldh: @jankatins ^

chipbuster:
There are only really two issues I see with `context_path` at the moment: one is that it could interfere with the upcoming™ bench suite, but I think I know how to work around that.

The other one is that we potentially have two pathways for working with paths now: an unmocked one that can be used for most things and a mocked one which needs to be used for absolute paths.

However, I think as long as this remains API-internal, we don't have to worry about it too much. We don't currently design our library to be consumed by third-parties, right? (Stability is only guaranteed in the binary behavior, not at the API level).

chipbuster:
Confirmed that we only treat stability at the binary level, so this is good to go. Thank you @haraldh for writing this and @andytom and @jankatins for reviewing/commenting!

PR #3312, opened 2021-12-12
feat(status): Add hex_status

mika-fischer:
#### Description
Add hex_status variable to status module containing the exit code in hex format.

#### Motivation and Context
This is useful on windows because many NTSTATUS codes have high bits set leading to large negative numbers (see below).
Refs: https://github.com/starship/starship/pull/2025

Before:
💥 -1073741510❯
After:
💥 0xC000013A❯

#### Checklist:

### DIFF
--- a/src/modules/status.rs
+++ b/src/modules/status.rs
@@ -106,6 +106,11 @@ fn format_exit_code<'a>(
         }
     };
 
+    let hex_status = exit_code
+        .parse::<i32>()
+        .ok()
+        .map(|code| format!("0x{:X}", code));
+
     let common_meaning = status_common_meaning(exit_code_int);
 
     let raw_signal_number = match config.recognize_signal_code {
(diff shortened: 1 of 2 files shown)


src/modules/status.rs
```diff
@@ -279,6 +285,28 @@ mod tests {
         }
     }
 
+    #[test]
+    fn failure_hex_status() {
+        let exit_values = [1, 2, 130, -2147467260];
+
+        for status in exit_values.iter() {
```
chipbuster, line 292: Can we create a second array that matches `exit_values` (maybe called `string_values`) that contains the actual string values of the outputs we expect to see?

As someone who doesn't remember what the `:X` formatter does off the top of my head (and likely will not in the foreseeable future), I'm currently having to scramble through the docs for std::format trying to find out what that format string does. This seems like a great opportunity to add some examples in the code itself.

chipbuster (changes requested):
Almost-perfect! One tiny request based on the fact that I can *never* remember what format strings do.

src/modules/status.rs
mika-fischer: I changed it accordingly, thanks for the review!

PR #3320, opened 2021-12-13
feat(status): Add os_abort_name

mika-fischer:
#### Description
os_abort_name is non-empty if the process was aborted via signal (POSIX) or was aborted with an NTSTATUS exit code (Windows).

#### Motivation and Context
Refs: https://github.com/starship/starship/pull/2025

Note that this is based on https://github.com/starship/starship/pull/3312

Before:
💥 0xC000013A❯
After:
💥 0xC000013A (CONTROL_C_EXIT)❯

#### Checklist:

### DIFF
--- a/src/modules/status.rs
+++ b/src/modules/status.rs
@@ -132,6 +132,8 @@ fn format_exit_code<'a>(
         false => None,
     };
 
+    let os_abort_name = status_os_abort_name(exit_code_int);
+
     StringFormatter::new(format).and_then(|formatter| {
         formatter
             .map_meta(|var, _| match var {
@@ -164,6 +166,7 @@ fn format_exit_code<'a>(
                 "common_meaning" => Ok(common_meaning).transpose(),
                 "signal_number" => Ok(signal_number.as_deref()).transpose(),
                 "signal_name" => Ok(signal_name).transpose(),
+                "os_abort_name" => Ok(os_abort_name).transpose(),
                 "pipestatus" => {
                     let pipestatus = pipestatus.unwrap_or_else(|| {
                         // We might enter this case if pipestatus hasn't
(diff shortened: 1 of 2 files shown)


chipbuster (dismissed):
I don't actually need changes right now, but I want to prevent an accidental merge of this before #3312.

mika-fischer:
Great, I rebased on master, thanks!

src/modules/status.rs
```diff
@@ -226,6 +229,55 @@ fn status_signal_name(signal: SignalNumber) -> Option<&'static str> {
     }
 }
 
+#[cfg(not(windows))]
+fn status_ntstatus_name(exit_code: &str) -> Option<&str> {
```
chipbuster, line 233:
```suggestion
fn status_ntstatus_name(_exit_code: &str) -> Option<&str> {
```

Clippy is upset by the unused variable (I'm actually surprised it didn't fail the CI entirely, it's usually a lot more strict). Prepend `_` to the argument to suppress the warning.

chipbuster (changes requested):
One minor change to appease Clippy and keep our warning volume low.

Quick question: is there any easy way to integrate this into the `common_meaning` function? We seem to be developing two sets of independent variables here (one for UNIX, one for Windows) and it would be nice if we could unify the two.

If there's no easy way, no big deal--we can come back around and revisit how we want to architect this module for v2.0.

src/modules/status.rs
mika-fischer: Ah, good catch, thanks. Fixed it.

mika-fischer:
> Quick question: is there any easy way to integrate this into the `common_meaning` function? We seem to be developing two sets of independent variables here (one for UNIX, one for Windows) and it would be nice if we could unify the two.

As discussed in [#2025 (comment)](https://github.com/starship/starship/pull/2025#issuecomment-991890796), I'm not at all a fan of `common_meaning` and don't use it at all in my config because it is so often wrong. In short, `common_meaning` is purely convention, `signal_name` is OS-enforced and `ntstatus_name` is as well. Therefore there is a large overlap between `signal_name` and `ntstatus_name` since they have similar functions in that they communicate an OS-forced exit due to a specific reason.

However, I felt that just reusing `signal_name` on Windows would also be somewhat confusing, since they are just not signals on Windows...

I would also be open to have a shared variable that uses has the value of `ntstatus_name` / `signal_name` when either of them is set (they can never be both set at the same time). I just don't know a good name for it. The original PR used `text_status`. Maybe `os_status_name` would be better? Maybe you have a better idea?
But I would argue against also including `common_meaning` into the mix here, because it is something very different and has a lot more false positives and I would not want to be forced to use it.

> If there's no easy way, no big deal--we can come back around and revisit how we want to architect this module for v2.0.

A migration path for v2.0 might be to introduce the new shared variable now and deprecate `signal_name` sometime in v1.x and then remove it in v2.0.

Of course, we could also just re-use `signal_name` on Windows if you think that's fine. That would certainly be the simplest solution.

chipbuster:
My bad, I didn't realize there was a previous PR for this.

> In short, common_meaning is purely convention, signal_name is OS-enforced

Technically, `signal_name` is also done by convention, albeit an incredibly widespread one. UNIX shells generally do not make the entire exit code available to users of the shell. Instead, they provide the lower 8 bits and use exit statuses above 128 to indicate that the process was signaled. Unlike `common_meaning` though, this convention is widespread enough that it's at least somewhat-reliable (crucially, I don't think any shells have conventions that would create false positives on signals).

> However, I felt that just reusing signal_name on Windows would also be somewhat confusing, since they are just not signals on Windows...

I concur that reusing `signal_name` on Windows would be confusing. 

> Maybe `os_status_name` would be better?

It seems that in both cases, we're not looking at an ordinary status, but some sort of "abnormal OS-induced termination" report, where instead of the process reporting its own exit status, the OS terminates the process and then gives us a status in the dead process's place. `os_status_name` seems a slightly misleading in that regard since it could also refer to an ordinary status, but I can't seem to come up with a good name that captures this idea at the moment.

> A migration path for v2.0 might be to introduce the new shared variable now and deprecate signal_name sometime in v1.x and then remove it in v2.0.

This sounds great. Just need to come up with a name (or if we can't find a better one in a day, `os_status_name` seems acceptable, if not optimal)

mika-fischer:
Hm, interesting. After some research into this I think what happens on Linux & bash is this:
- Linux sets the proper exit status or signal number on process handle after process termination
- This exit status can be fetched via [waitpid](https://linux.die.net/man/2/wait)
  -  At this point it's still possible to reliably distinguish between a normal exit and a forced termination via signal by the kernel (see `WIFEXITED` and `WIFSIGNALED` in the link above)
  - The normal exit code is [8 bits](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/bits/waitstatus.h#L28), while the signal number is [7 bits](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/bits/waitstatus.h#L31)
- Then the shell comes and [mangles this all into a single 8 bit integer](https://github.com/bminor/bash/blob/15409324f1974d41c183904ad575da7188058c1c/jobs.c#L2809-L2818)
  - It uses the normal 8 bit exit code as is and adds 128 to the signal number
  - This causes the problem that a process exiting normaly with an exit code > 128 is indistinguishable from a process being killed via a signal, which is unfortunate...

Demonstration:
```
❮ bash -c 'kill -HUP $$'
zsh: hangup     bash -c 'kill -HUP $$'

mfischer in arthur in ~
💥 0x81 [SIGHUP] ❮
```
```
❮ bash -c 'exit 129'

mfischer in arthur in ~
💥 0x81 [SIGHUP] ❯
```

But there's not really much we can do about this and the situation is exactly the same on Windows, where it's also possible to exit a process with an exit code that corresponds to an NTSTATUS error code and that is also indistiguishable from the process really being killed by the OS.

But I think these things should be rare enough that we can consider signal_name and nstatstus_name as reliable, even though it's really not 100%...

I'll think a bit more about the name and update this sometime during the holidays.

And happy holidays 🎄

chipbuster:
> But I think these things should be rare enough that we can consider signal_name and nstatstus_name as reliable, even though it's really not 100%...

Yes, I think we're helped by the fact that most programs try to use lower exit code before higher ones, and that on UNIX, there are only really about 32 signals, of only which 16 will kill the process (signals like "window size changed" or "your child process completed" will very rarely ever terminate the process that receives it). That means that the only range we're worried about on UNIX is 128-144. Based on your PR, it seems like you'd also have to be targeting those Windows exit codes specifically, unless your program just happens to have 2^32 exit codes.

So I think it's safe to call this "reliable", even if it's really not.

> I'll think a bit more about the name and update this sometime during the holidays.

Don't worry too much about it.

> And happy holidays

To you too! 🍾

chipbuster:
A thought I had on this: the "abnormal OS-induced termination" can be conceptually distinguished from an "exit" code: in one case, the process exits on its own and reports its own code, in the other, the code is supplied by OS code.

It seems that Tannenbaum refers to this concept as a "termination status" instead of an "exit status" (see e.g. [this SO question](https://stackoverflow.com/questions/61980810/what-is-the-difference-between-termination-status-and-end-status-as-returned)), so if we could just somehow get a short form of "OS termination status" we'd probably have a good candidate.

mika-fischer:
Getting back to this. I can only come up with "abort". I would also use "name" as the suffix for the textual error message, since "status" ist already the name of the exit code and therefore it would be confusing.

So we have:
- os_termination_name
- os_term_name
- os_abort_name

I think I like "os_abort_name" best, since it's resonably short without abbreviating the middle term. But let me know what you think @chipbuster

chipbuster:
Oh yes, I like "abort" as well. Let's go with that.

mika-fischer:
Hi @chipbuster et al. Sorry for forgetting about this. I've updated the pull request with the agreed-upon `os_abort_name`. Let me know what you think and if there's anything else I should change.

PR #3322, opened 2021-12-14
feat: set a continuation prompt for supporting shells

### LINKED ISSUE #3134: Set a continuation prompt for supporting shells
The PowerShell continuation prompt is left at its default.

image

The same in other shells, such as Bash with `PS2`, Zsh with `PROMPT2`, etc.

It would be nice to set it to something more consistent with the prompt by default. e.g. A yellow right angle bracket. Though if the user customizes the prompt, he would also need to change this.

This can be set in PowerShell via `Set-PSReadLineOption -ContinuationPrompt` as far as I can tell. Other shells have different mechanisms, such as `PS2` in Bash, `PROMPT2` in Zsh, etc.

Set it by myself, or ignore it.

raccog:
#### Description
Adds options for customizing the continuation prompt for shells that support it (currently bash, zsh, and PowerShell).

The `continuation_format` option is added to customize a format for continuation prompts that is separate from the normal prompt format. The default for `continuation_format` is `"$character"`.

In the character module, the `continuation_symbol` option is added to customize the character used during a continuation prompt. A bold, yellow `❯` character is used as the default for `continuation_symbol`.

To construct the continuation prompt, the `--continuation` flag is added to `starship prompt`. When called with this flag, starship pipes the continuation prompt to standard output instead of the normal prompt. Thus, an extra call to starship is added to the bash, zsh, and PowerShell templates for setting the continuation prompts.

#### Motivation and Context
Currently, starship does not set the continuation prompt on any shells that support it. It would be nice to be able to customize these prompts using starship.

Closes #3134

Screenshots were taken using iTerm2 on MacOS with zsh as the shell.

Continuation prompt before:
Screen Shot 2021-12-13 at 4 56 23 PM

Continuation prompt after:
Screen Shot 2021-12-13 at 4 55 59 PM

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -144,7 +144,7 @@ Produces a prompt like the following:
 ## Continuation Prompt
 
 Some shells support a continuation prompt along with the normal prompt. This prompt is rendered instead of the normal prompt when the
-user has entered an uncomplete statement (such as a single left parenthesis).
+user has entered an incomplete statement (such as a single left parenthesis).
 
 Starship can set the continuation prompt format using the `continuation_format` option. The
 default format is `"$character"`.
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -141,6 +141,34 @@ Produces a prompt like the following:
 ▶                                   starship on  rprompt [!] is 📦 v0.57.0 via 🦀 v1.54.0 took 17s
 ```
 
+## Continuation Prompt
+
+Some shells support a continuation prompt which is rendered instead of the normal prompt when the
+user has entered an uncomplete statement that needs to be continued.
+
+Starship can set the continuation prompt format by setting the `continuation_format` field. The
+default format is `"$character"`.
+
+Note: Continuation prompts are only available in the following shells:
+
+  - `bash`
+  - `zsh`
+  - `PowerShell`
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+# A continuation prompt that displays two characters
+continuation_format = """$character$character"""
+```
+
+Produces a continuation prompt like the following:
+
+```
+▶▶
+```
 
 ## Style Strings
 
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -143,11 +143,9 @@ Produces a prompt like the following:
 
 ## Continuation Prompt
 
-Some shells support a continuation prompt along with the normal prompt. This prompt is rendered instead of the normal prompt when the
-user has entered an incomplete statement (such as a single left parenthesis).
+Some shells support a continuation prompt along with the normal prompt. This prompt is rendered instead of the normal prompt when the user has entered an incomplete statement (such as a single left parenthesis or quote).
 
-Starship can set the continuation prompt format using the `continuation_format` option. The
-default format is `"$character"`.
+Starship can set the continuation prompt using the `continuation_prompt` option. The default prompt is `"[❯](bold yellow)"`.
 
 Note: Continuation prompts are only available in the following shells:
 
@@ -155,19 +153,15 @@ Note: Continuation prompts are only available in the following shells:
   - `zsh`
   - `PowerShell`
 
+Note: Unlike `format` and `right_format`, the `continuation_prompt` is only set a single time; when `starship init` is called for one of the available shells. This may break modules used in `continuation_prompt` that depend on being called after every command.
+
 ### Example
 
 ```toml
 # ~/.config/starship.toml
 
-# A continuation prompt that displays two characters
-continuation_format = """$character$character"""
-```
-
-Produces a continuation prompt like the following:
-
-```
-▶▶
+# A continuation prompt that displays two filled in arrows
+continuation_prompt = "▶▶"
 ```
 
 ## Style Strings
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -143,10 +143,10 @@ Produces a prompt like the following:
 
 ## Continuation Prompt
 
-Some shells support a continuation prompt which is rendered instead of the normal prompt when the
-user has entered an uncomplete statement that needs to be continued.
+Some shells support a continuation prompt along with the normal prompt. This prompt is rendered instead of the normal prompt when the
+user has entered an uncomplete statement (such as a single left parenthesis).
 
-Starship can set the continuation prompt format by setting the `continuation_format` field. The
+Starship can set the continuation prompt format using the `continuation_format` option. The
 default format is `"$character"`.
 
 Note: Continuation prompts are only available in the following shells:
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -147,13 +147,13 @@ Some shells support a continuation prompt along with the normal prompt. This pro
 
 Starship can set the continuation prompt using the `continuation_prompt` option. The default prompt is `"[❯](bold yellow)"`.
 
+Note: `continuation_prompt` should be set to a literal string without any variables.
+
 Note: Continuation prompts are only available in the following shells:
 
   - `bash`
   - `zsh`
   - `PowerShell`
-
-Note: Unlike `format` and `right_format`, the `continuation_prompt` is only set a single time; when `starship init` is called for one of the available shells. This may break modules used in `continuation_prompt` that depend on being called after every command.
 
 ### Example
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -143,14 +143,13 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Option                | Default                         | Description                                                      |
-| --------------------- | ------------------------------- | ---------------------------------------------------------------- |
-| `format`              | [link](#default-prompt-format)  | Configure the format of the prompt.                              |
-| `right_format`        | `""`                            | See [Enable Right Prompt](/advanced-config/#enable-right-prompt) |
-| `continuation_format` | `"$character"`                  | See [Continuation Prompt](/advanced-config/#continuation-prompt) |
-| `scan_timeout`        | `30`                            | Timeout for starship to scan files (in milliseconds).            |
-| `command_timeout`     | `500`                           | Timeout for commands executed by starship (in milliseconds).     |
-| `add_newline`         | `true`                          | Inserts blank line between shell prompts.                        |
+| Option            | Default                         | Description                                                      |
+| ----------------- | ------------------------------- | ---------------------------------------------------------------- |
+| `format`          | [link](#default-prompt-format)  | Configure the format of the prompt.                              |
+| `right_format`    | `""`                            | See [Enable Right Prompt](/advanced-config/#enable-right-prompt) |
+| `scan_timeout`    | `30`                            | Timeout for starship to scan files (in milliseconds).            |
+| `command_timeout` | `500`                           | Timeout for commands executed by starship (in milliseconds).     |
+| `add_newline`     | `true`                          | Inserts blank line between shell prompts.                        |
 
 
 ### Example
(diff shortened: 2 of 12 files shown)


docs/advanced-config/README.md
```diff
@@ -141,6 +141,34 @@ Produces a prompt like the following:
 ▶                                   starship on  rprompt [!] is 📦 v0.57.0 via 🦀 v1.54.0 took 17s
 ```
 
+## Continuation Prompt
+
+Some shells support a continuation prompt along with the normal prompt. This prompt is rendered instead of the normal prompt when the
+user has entered an uncomplete statement (such as a single left parenthesis).
```
segevfiner, line 147:
```suggestion
user has entered an incomplete statement (such as a single left parenthesis).
```

chipbuster:
Before I get into the performance concerns with starship, I just want to say that this is generally a very well-written PR. I wish they would all come like this.

Starship pays the cost of a process launch every time we invoke the binary. This is relatively small, but does actually add up when it comes to command latency. As a simple example, on my (relatively beefy) development machine in `HOME`, the individual segments take less than 3ms to render, and there are fewer segments than I have cores, yet it takes something like 12-15ms to actually complete a draw cycle. Part of this is the additional overhead that we pay per starship launch, and even at 15 ms, it's noticeable if you just hold down ENTER to try to get as many prompt draws as you can.

I was already a little unhappy with having to pay a second process launch cost to get RPROMPT support, but paying an *additional* process launch in order to set a continuation character (which will almost never change over the lifetime of a shell session) seems somewhat excessive to me.

Is there a use case we envision where the user will regularly want to include information in the continuation prompt that isn't a static string? If not, can we set the continuation string once at launch and avoid having to run starship three times every time the user presses ENTER?

raccog:
@chipbuster Thank you very much for the kind feedback! This is my first pull request, so I'm glad that I'm going in the right direction!

I would also be happy to fix these performance issues. So far, I can think of the following alternative methods of returning every prompt on a single run:
- Set environment variables to the right and continuation prompts, while piping the normal prompt to stdout. However, I do not have enough experience using environment variables on non-Unix systems to know if this would be cross compatible.
- Pipe all prompts to stdout with delimiters. However, this requires an external way of separating this string, so it could require an extra binary call.

The main use case that I can think of for a dynamic continuation prompt is the default way that zsh uses it by describing what part of the command needs to be continued (can be `if`, `while`, `bquote`, etc.). However, after looking through the zsh documentation, this is implemented by expanding special characters in the prompt each time. Thus, I agree that this can be implemented as a static prompt.

I'm going to test out some fixes for the performance issues related to the right prompt, so when I'm done I will write a new pull request. As for the continuation prompt, should I add a new commit to this pull request that changes it to a static string? Or should I create a new pull request that completes this change?

chipbuster:
@raccog Let's not get too far into optimizing right now, for several reasons:\

1. The RPROMPT optimization will have some pretty deep-reaching implications for starship's architecture, whatever path we choose.
2. We don't have the appropriate tools to measure and compare performance at the moment.
3. **Even if you improve the performance, we don't have a way of measuring performance regressions, so there's a decent probability the gains in your work would just be wiped out later on by a different PR**.

That last point is probably the biggest reason I'd dissaude you from tackling perf optimizations at the moment. I'm currently [building out a set of tools](https://github.com/chipbuster/criterion-compare) that will (hopefully) let us measure performance changes in PRs. Once that's in place, we should have a much better time tracking perf across the codebase.

For continuation, you can do whatever you think is best. It would probably be easiest to force-push into this PR with changes that convert this to a static string, but if git starts giving you trouble like it does to me, you can feel free to close this PR and open a new one.

raccog:
@chipbuster Thanks again for the advice! In this case I will wait for the results of your performance toolset before attempting to fix any performance issues.

It turns out that only the bash script was setting the continuation prompt multiple times. Making the continuation prompt static only required me to change this script. Please let me know if there is anything else that I can do to improve this pull request.

docs/config/README.md
```diff
@@ -143,13 +143,14 @@ This is the list of prompt-wide configuration options.
...
-| `right_format`    | `""`                            | See [Enable Right Prompt](/advanced-config/#enable-right-prompt) |
-| `scan_timeout`    | `30`                            | Timeout for starship to scan files (in milliseconds).            |
-| `command_timeout` | `500`                           | Timeout for commands executed by starship (in milliseconds).     |
-| `add_newline`     | `true`                          | Inserts blank line between shell prompts.                        |
+| Option                | Default                         | Description                                                      |
+| --------------------- | ------------------------------- | ---------------------------------------------------------------- |
+| `format`              | [link](#default-prompt-format)  | Configure the format of the prompt.                              |
+| `right_format`        | `""`                            | See [Enable Right Prompt](/advanced-config/#enable-right-prompt) |
+| `continuation_format` | `"$character"`                  | See [Continuation Prompt](/advanced-config/#continuation-prompt) |
```
chipbuster, line 150: I'm a little concerned about the presence of this here, since it seems to imply that we can set the continuation prompt to something that will be dynamically updated with every prompt draw (like the right prompt is). Any ideas on how to clarify this? Perhaps a warning in advanced-config?

docs/config/README.md
```diff
@@ -457,21 +458,28 @@ look at [this example](#with-custom-error-shape).
...
+| `disabled`            | `false`              | Disables the `character` module.                                                            |
 
 ### Variables
 
-| Variable | Example | Description                                                           |
-| -------- | ------- | --------------------------------------------------------------------- |
-| symbol   |         | A mirror of either `success_symbol`, `error_symbol` or `vicmd_symbol` |
+| Variable | Example | Description                                                                                   |
+| -------- | ------- | --------------------------------------------------------------------------------------------- |
+| symbol   |         | A mirror of either `success_symbol`, `error_symbol`, `continuation_symbol`, or `vicmd_symbol` |
```
chipbuster, line 482: Why would this mirror `continuation_symbol`?

src/init/starship.zsh
```diff
@@ -94,4 +94,5 @@ setopt promptsubst
 
 PROMPT='$(::STARSHIP:: prompt --terminal-width="$COLUMNS" --keymap="${KEYMAP:-}" --status="$STARSHIP_CMD_STATUS" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --cmd-duration="${STARSHIP_DURATION:-}" --jobs="$STARSHIP_JOBS_COUNT")'
 RPROMPT='$(::STARSHIP:: prompt --right --terminal-width="$COLUMNS" --keymap="${KEYMAP:-}" --status="$STARSHIP_CMD_STATUS" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --cmd-duration="${STARSHIP_DURATION:-}" --jobs="$STARSHIP_JOBS_COUNT")'
+PROMPT2='$(::STARSHIP:: prompt --continuation --terminal-width="$COLUMNS" --keymap="${KEYMAP:-}" --status="$STARSHIP_CMD_STATUS" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --cmd-duration="${STARSHIP_DURATION:-}" --jobs="$STARSHIP_JOBS_COUNT")'
```
chipbuster, line 97: Unfortunately, this still causes an additional starship execution per prompt draw.

chipbuster: Also, we *really* shouldn't need these other arguments right? This is just supposed to set a PS2 exactly once, when the shell launches?

chipbuster (changes requested):
Unfortunately, the ZSH init still causes a prompt draw per execution on my system. You can check this by starting a continuation prompt, then changing the config. If the continuation prompt changes immediately without needing to restart the shell, starship is re-evaluating the character every time:

image


A general question about the design/positioning of this: my impression of this is that this is a convenience function for starship users, i.e. it wouldn't be hard to set this yourself, but we're providing a set of shell wrappers to set simple continuation prompts for them.

Given this, does it make sense for this to be in the character module, which usually deals with displaying dynamic information (e.g. command success, vi-mode, etc)? I wonder if it's a little misleading, because it implies that the module responds to stuff that the character module usually responds to.

src/init/starship.zsh
segevfiner: I think just removing the single quotes should be enough to get it to run once. And yeah since the other shells seem to use only `--continuation` without the rest of the flags.

```suggestion
PROMPT2=$(::STARSHIP:: prompt --continuation)
```

chipbuster: It's been a long time since I messed around with the prompt variable quoting directly, but I seem to vaguely recall that double-quoting will work as well (which will be needed if e.g. someone tries to set `> >` as their continuation for some reason).

I don't have the slightest clue why that happens though, and someone should *definitely* double check me on that.

segevfiner: Yes it should work as well. The behavior of shells is a bit weird in that area, and I'm not sure what POSIX mandates, but both Bash and Zsh won't do word splitting on variable assignment, and Zsh in particular doesn't do word splitting on any parameter/variable expansion.

segevfiner: https://unix.stackexchange.com/questions/178294/is-field-splitting-and-filename-expansion-turn-off-in-rhs-of-variable-assignment

Though there are outlier shell in this regard due to how confusing this is.

src/main.rs
```diff
@@ -112,6 +112,11 @@ fn main() {
                             .long("right")
                             .help("Print the right prompt (instead of the standard left prompt)"),
                     )
+                    .arg(
+                        Arg::with_name("continuation")
+                            .long("continuation")
+                            .help("Print the continuation prompt (instead of the standard prompt)"),
```
davidkna, line 118: I think the language should match `--right` and the new argument should be marked as incompatible with `--right`.

```suggestion
                            .help("Print the continuation prompt (instead of the standard left prompt)")
                            .conflicts_with("right"),
```

docs/config/README.md
raccog: I'm actually wondering if the static continuation prompt should utilize modules at all. Most modules seem to depend on the fact that they get rerun after every command, so they would be mostly useless here. Please let me know your opinion on this.

Do you think it would be a good idea to call it `continuation_prompt` instead of `continuation_format`; implying that it does not get formatted like the default and right prompts? If we choose to not utilize modules, I can also add this fact as a warning in advanced-config.

docs/config/README.md
raccog: I thought that the value of `symbol` would be set to the value of either `success_symbol`, `error_symbol`, `continuation_symbol`, or `vicmd_symbol`, depending on the last entered command.

As the continuation prompt is going to be static and will be moved out of the character module anyway, I'll fix this by the next commit.

src/init/starship.zsh
raccog: @chipbuster @segevfiner Thanks for testing this, I did not know that the prompt was still being updated after every command.

raccog:
>A general question about the design/positioning of this: my impression of this is that this is a convenience function for starship users, i.e. it wouldn't be hard to set this yourself, but we're providing a set of shell wrappers to set simple continuation prompts for them.

@chipbuster Yes, my purpose for this feature is simply to be able to set the continuation prompt once in the starship config file and have starship set the same continuation prompt whether I am using bash, zsh, or powershell. As such, I don't necessarily need any of the dynamic features in starship.

I am going to create a new commit under this merge request that separates the continuation prompt from these dynamic features. As you mention that the character module is dynamic and the continuation prompt does not fit here, I will move it away from this module.

raccog:
@chipbuster In my new commit, I have fixed the continuation prompt to be set a single time and moved it out of the character module.

However, I left the modules to still be usable in the continuation prompt while adding a note that they will not be updated when used in that prompt. Let me know if you think modules should not be usable in the continuation prompt and I can add another commit to fix this.

chipbuster (commented):
One comment about doc, then I need to pull and test this and we should be good to go.

docs/advanced-config/README.md
```diff
@@ -141,6 +141,28 @@ Produces a prompt like the following:
...
+
+Starship can set the continuation prompt using the `continuation_prompt` option. The default prompt is `"[❯](bold yellow)"`.
+
+Note: Continuation prompts are only available in the following shells:
+
+  - `bash`
+  - `zsh`
+  - `PowerShell`
+
+Note: Unlike `format` and `right_format`, the `continuation_prompt` is only set a single time; when `starship init` is called for one of the available shells. This may break modules used in `continuation_prompt` that depend on being called after every command.
```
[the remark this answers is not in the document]
raccog: Thanks for the advice, I have just changed this.

chipbuster (approved):
:+1: LGTM!

Thanks @raccog for the changes!

davidkna:
Thanks for implementing this @raccog, and thanks for reviewing @chipbuster!

raccog:
Thank you for the review @chipbuster!

PR #3359, opened 2021-12-28
fix(rust): fix overrides on windows and set cwd

### LINKED ISSUE #3358: Rust version is incorrect when creating a rustup override
#### Current Behavior

I have created an override with rustup, `rustup override set nightly` in one of my folders.

#### Expected Behavior

Starship thinks that the directory is using the stable toolchain when it isn't.

image

#### Environment
- Starship version: 1.1.1
- Operating system: Windows 10.0.22000
- Git Commit Hash: 9a4b2bb5
- Branch/Tag: v1.1.1
- Rust Version: rustc 1.57.0 (f1edd0429 2021-11-29)
- Build Time: 2021-12-21 20:14:52 +00:00

```bash
<unknown config>
```

#### Starship Configuration

```toml
[time]
disabled = false
format = '[\[ $time \]]($style) '
time_format = "%T"

[nodejs]
disabled = true

[git_status]
disabled = true

[cmd_duration]
min_time = 500
format = "took [$duration](bold yellow)"

[battery]
disabled = true

[git_metrics]
disabled = false

[lua]
disabled = true
```

davidkna:
#### Description

This PR changes override parsing to be more reliable (spaces in filenames, most specific matches) and [makes the prefix detection work properly on Windows](https://github.com/starship/starship/issues/3358#issuecomment-1002265978).
Because #3358 doesn't occur when running from terminal, I assumed this was related to unset cwd like #3309, but that didn't end up fixing it, but I am including the changes anyway because it increases timeout coverage.

#### Motivation and Context

Closes #3358

#### Checklist:

### DIFF
--- a/src/modules/rust.rs
+++ b/src/modules/rust.rs
@@ -79,7 +79,7 @@ fn get_module_version(context: &Context, config: &RustConfig) -> Option<String> 
         .or_else(|| find_rust_toolchain_file(context))
         .or_else(|| execute_rustup_default(context))
     {
-        match execute_rustup_run_rustc_version(&toolchain) {
+        match execute_rustup_run_rustc_version(context, &toolchain) {
             RustupRunRustcVersionOutcome::RustcVersion(rustc_version) => {
                 format_rustc_version(&rustc_version, config.version_format)
             }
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM

chipbuster:
@davidkna The comment in this PR notes that "rustup strips \\?\ prefix". Do you know if this stripping is done when the path is stored in `settings.toml`, or is it done when `rustup` runs and displays things?

#3414 no longer parses the output of `rustup`---instead, it uses the `rustup_home()` function (which should be identical to what `rustup` itself uses) to find the override file, then parses that directly. If `rustup` strips these prefixes before storing them, then we might not need to worry about that part anymore.

Unfortunately, I don't know how to check this myself.

PR #3370, opened 2021-12-31
refactor: update clap to v3

davidkna:
#### Description

Updates clap to v3. I switched the implementation to the new `derive`/`structopt` style because there were so many breaking changes.
Completions are now handled by `clap-completions`. ~~I changed some arguments to allow setting them from environment variables.~~ Shared `module`/`prompt`/`explain` arguments are now handled by `context::Properties` and `--right`/`--continuation` are handled separately for `prompt` and passed on as a `Target` (the other subcommands ignore target and set it to `Main`). 

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -15,6 +15,7 @@ use std::env;
 use std::ffi::{OsStr, OsString};
 use std::fmt::Debug;
 use std::fs;
+use std::num::ParseIntError;
 use std::path::{Path, PathBuf};
 use std::string::String;
 use std::time::{Duration, Instant};
@@ -593,7 +594,7 @@ pub struct Properties {
     #[clap(short = 'k', long, default_value = "viins")]
     pub keymap: String,
     /// The number of currently running jobs
-    #[clap(short, long, default_value_t)]
+    #[clap(short, long, default_value_t, parse(try_from_str=parse_jobs))]
     pub jobs: i64,
 }
 
@@ -610,6 +611,10 @@ impl Default for Properties {
             jobs: 0,
         }
     }
+}
+
+fn parse_jobs(jobs: &str) -> Result<i64, ParseIntError> {
+    jobs.trim().parse::<i64>()
 }
 
 #[cfg(test)]
--- a/src/context.rs
+++ b/src/context.rs
@@ -296,6 +296,14 @@ impl<'a> Context<'a> {
         }
     }
 
+    // TODO: This should be used directly by clap parse
+    pub fn get_cmd_duration(&self) -> Option<u128> {
+        self.properties
+            .cmd_duration
+            .as_deref()
+            .and_then(|cd| cd.parse::<u128>().ok())
+    }
+
     /// Execute a command and return the output on stdout and stderr if successful
     #[inline]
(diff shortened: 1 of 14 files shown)


chipbuster:
This will likely need to be rebased on #3322 (which just got approved)

davidkna:
Removed the feature of giving arguments with environment variables, because they overwrite explicit arguments.

src/context.rs
```diff
@@ -48,11 +51,8 @@ pub struct Context<'a> {
     /// The shell the user is assumed to be running
     pub shell: Shell,
 
-    /// Construct the right prompt instead of the left prompt
-    pub right: bool,
-
-    /// Construct the continuation prompt instead of the normal prompt
-    pub continuation: bool,
+    /// Which prompt to print (main, right, ...)
+    pub target: Target,
```
matchai, line 55: Nice 👍

src/context.rs
```diff
@@ -574,6 +556,67 @@ pub enum Shell {
...
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Target {
+    Main,
+    Right,
+    Continuation,
+}
+
+/// Properties as passed on from the shell as arguments
+#[derive(Parser, Debug)]
+
```
matchai, line 572: nitpick: Looks like a lingering newline was left here between the attribute and the definition 😄 
```suggestion
```

src/context.rs
```diff
@@ -574,6 +556,67 @@ pub enum Shell {
     Unknown,
 }
 
+fn default_width() -> usize {
+    terminal_size().map_or(80, |(w, _)| w.0 as usize)
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Target {
```
matchai, line 564: Could we add a doc comment here explaining what `Target` is?

src/main.rs
```diff
@@ -4,297 +4,165 @@ use clap::crate_authors;
...
-
-    let keymap_arg = Arg::with_name("keymap")
-        .short("k")
-        .long("keymap")
-        .value_name("KEYMAP")
-        // fish/zsh only
-        .help("The keymap of fish/zsh")
-        .takes_value(true);
+    let args = Cli::parse();
+    log::trace!("Parsed arguments: {:#?}", args);
```
matchai, line 112: 👍

matchai (approved):
LGTM. Just a couple small comments 👍

src/main.rs
```diff
@@ -4,297 +4,165 @@ use clap::crate_authors;
...
+        #[clap(requires = "value")]
+        name: Option<String>,
+        /// Value to place into that key
+        value: Option<String>,
+    },
+    /// Explains the currently showing modules
+    Explain(Properties),
+    ///  Prints the shell function used to execute starship
+    Init {
+        shell: String,
```
tranzystorek-io, line 54: We could instead create an enum with all shell types and derive `ArgEnum` for it

davidkna: I didn't do that because passing the shell path is also supported for this argument. Also, how to handle unknown shell names will require some thought.

PR #3374, opened 2022-01-01
fix: update continuation prompt to be more minimal

matchai:
#### Description

This PR updates the continuation prompt to be a more minimal alternative to the current yellow chevron.

I think the yellow chevron may be a little bold and visually loud for something that I feel is more of an accent than an actionable piece of information. Interested to hear your thoughts @raccog, @chipbuster, @davidkna.

**Changes**
- Replace the chevron with a gray (bright black) bullet
- Add a space after the prompt to better align with the default prompt

#### Motivation and Context

Closes #

**Current continuation prompt defaults**
CleanShot 2022-01-01 at 16 56 03@2x

**Proposed update**
CleanShot 2022-01-01 at 16 57 23@2x

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -145,7 +145,7 @@ Produces a prompt like the following:
 
 Some shells support a continuation prompt along with the normal prompt. This prompt is rendered instead of the normal prompt when the user has entered an incomplete statement (such as a single left parenthesis or quote).
 
-Starship can set the continuation prompt using the `continuation_prompt` option. The default prompt is `"[•](bright-black)"`.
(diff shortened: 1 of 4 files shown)


davidkna:
I would also prefer this as the default.

raccog:
Looks good to me! I like that the circle differentiates it from a normal prompt.

I also noticed that I forgot to include a space after the arrow, so thank you for fixing that!

matchai:
Alternatively, we could go even smaller with the "dot product" symbol, which I think I prefer.

<table>
<tr>
	<td>Bullet
	<td>Dot product
<tr>
	<td>CleanShot 2022-01-01 at 16 57 23@2x
	<td>CleanShot 2022-01-01 at 17 29 22@2x

</table>

chipbuster:
I like the look.

PR #3375, opened 2022-01-01
feat(env_var): Add support for env_var.VAR in format

### LINKED ISSUE #3259: Customize env_var display order in format like what can be done for custom
Can't reorder multiple `env_var` sections individually using `format`.

I want to be able to reorder `env_var` the same as you can reorder custom sections by including `${custom.section}` in the `format`.

e.g.
```toml
format = "${env_var.WSL_DISTRO_NAME}$all"

[env_var.WSL_DISTRO_NAME]
format = "[$env_value]($style) in "
style = "bold dimmed green"
```

Using `custom` instead of `env_var`.

segevfiner:
#### Description
Allow using ${env_var.VAR} in format like you can do for custom.

:warning: This required a few hacks  which I will comment on after submitting the PR.

#### Motivation and Context

Closes #3259

image

Using:
```toml
format = "${env_var.USERPROFILE}$all"

[env_var.USERPROFILE]
format = "profile at [$env_value]($style) "
```
(Obviously not very useful, just for showing it working)

#### Checklist:

- [ ] I have updated the tests accordingly. &ndash; Might want to add a more explicit test that it works with the global `format` key.


src/config.rs
```diff
@@ -343,13 +343,40 @@ impl StarshipConfig {
...
     pub fn get_custom_modules(&self) -> Option<&toml::value::Table> {
         self.get_config(&["custom"])?.as_table()
     }
     /// Get the table of all the registered env_var modules, if any
-    pub fn get_env_var_modules(&self) -> Option<&toml::value::Table> {
-        self.get_config(&["env_var"])?.as_table()
+    pub fn get_env_var_modules(&self) -> Option<toml::value::Table> {
+        let config_table = self.get_config(&["env_var"])?.as_table()?;
+
+        // Old configuration is present in starship configuration
```
segevfiner, line 372: This basically transforms:
```
{"env_var": ...}
```
To:
```
{"": {"env_var"}}
```

Not sure if this is the best way to handle this.

src/configs/env_var.rs
```diff
@@ -23,6 +24,7 @@ impl<'a> Default for EnvVarConfig<'a> {
             variable: None,
             default: None,
             format: "with [$env_value]($style) ",
+            description: "<env_var module>",
```
segevfiner, line 27: Is this good? There is still also the previously used static description. Technically we could try generating a better one if the description is unset since unlike a custom module, we do know the variable name unlike an arbitrary shell command, for example to print something more descriptive.

src/print.rs
```diff
@@ -142,7 +142,13 @@ pub fn module(module_name: &str, args: ArgMatches) {
 }
 
 pub fn get_module(module_name: &str, context: Context) -> Option<String> {
-    modules::handle(module_name, &context).map(|m| m.to_string())
+    Some(
```
segevfiner, line 145: image

`get_module` used to call `modules::handle` but the handling for `custom`/`env_var` is only in `handle_prompt`, which caused tests using `ModuleRenderer` to fail. So I made a hack to have it go through `handle_prompt` but there might be a cleaner way.

src/print.rs
```diff
@@ -142,7 +142,13 @@ pub fn module(module_name: &str, args: ArgMatches) {
...
+        handle_module(module_name, &context, &BTreeSet::new())
+            .iter()
+            .map(|m| m.to_string())
+            .collect::<Vec<String>>()
```
segevfiner, line 149: This allocated a Vec that we immediately drop after joining, it doesn't seem like there is a way in std to join an `Iterator` of `String`.

src/print.rs
```diff
@@ -389,6 +433,28 @@ fn should_add_implicit_custom_module(
         .unwrap_or(false)
 }
 
+fn should_add_implicit_env_var_module(
```
segevfiner, line 436: As you might have noticed, there is a bit of code duplication from custom, should another module with this behavior be added, we might want to consider refactoring, or maybe even refactor some/all of it now, though it was complicated enough getting it to work as it is.

src/modules/env_var.rs
```diff
@@ -253,6 +223,21 @@ mod test {
         assert_eq!(expected, actual);
     }
 
+    #[test]
+    fn display_single_with_multiple_defined() {
```
segevfiner, line 227: We probably also want a test that tests this using the top-level `format` key. Is there any similar kind of test somewhere?

src/print.rs
davidkna: Shouldn't it be possible to make this particular function work for both by using `let explicit_module_name = format!("{parent_module}.{child_module}");`?

davidkna (changes requested):

PR #3386, opened 2022-01-03
fix(pwsh): avoid polluting the global function namespace

davidkna:
#### Description

#3322 changed the PowerShell init in order to pass the current starship arguments to `starship prompt --continuation`. ~~This PR reverts these changes and just uses plain `&starship prompt --continuation` (`Invoke-Native` isn't really needed there, and the zsh init does not pass along the arguments either) to avoid polluting the global function namespace with `Invoke-Native` and `Get-Arguments`.~~ I also moved the continuation prompt code to ensure `$STARSHIP_SHELL` and `$STARSHIP_SESSION_KEY` are set beforehand.

To avoid putting functions into the user scope with common names, this PR prefixes all of these functions with `__Starship-`. `Get-Arguments` is removed because it isn't needed for the continuation prompt.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -1,6 +1,7 @@
 #!/usr/bin/env pwsh
 
-function Get-Arguments {
+function global:prompt {
+
     function Get-Cwd {
         $cwd = Get-Location
         $provider_prefix = "$($cwd.Provider.ModuleName)\$($cwd.Provider.Name)::"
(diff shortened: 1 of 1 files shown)


segevfiner:
Not using `Invoke-Native` might break on Windows PowerShell (And maybe even on PowerShell Core on Windows) by not interpreting the output of starship as UTF-8 correctly.You might be able to define the functions local to the prompt function or the script, see https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_scopes?view=powershell-7.2

davidkna:
Even when wrapping with a function/ScriptBlock `prompt` will not use the functions from the local scope, so they need to be global when defined outside `prompt`. As a workaround I've prefixed the function names with `__Starship-`.

segevfiner:
Hmmm... Might require using `GetNewClosure` or `New-Module` the rules might be a bit iffy for "reasons".

PR #3394, opened 2022-01-05
Changed 'which' to 'command -v' which is more general, and available …

sandalbanditten:
…on more systems in docs/faq/README.md

#### Description

This changes the shell command for removing the starship binary on UNIX-like systems.
It changes the command `which` to `command -v`.
These two commands do the exact same thing.

#### Motivation and Context

This will make sure that starship works on more systems, since the command `which` is not universal, while `command` is.
`command -v` is a part of POSIX and is also a drop-in replacement for `which`.

#### Checklist:

### DIFF
--- a/docs/faq/README.md
+++ b/docs/faq/README.md
@@ -143,5 +143,5 @@ If Starship was installed using the install script, the following command will d
 
 ```sh
 # Locate and delete the starship binary
-sh -c 'rm "$(which starship)"'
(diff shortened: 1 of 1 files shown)


chipbuster (approved):

andytom:
Thank you for your contribution @sandalbanditten and thanks for reviewing @chipbuster.

PR #3405, opened 2022-01-08
docs: Added installation section with choco package manager

logotip4ik:
#### Description
Added to installation section another windows package manager `choco`.

#### Motivation and Context
I think that there are a lot of users that are using `choco` rather than scoop. This change was made just for user convention, there are no need to search for this awesome package in `choco` registry.

#### Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -56,6 +56,12 @@ description: Starship is the minimal, blazing fast, and extremely customizable p
    ```powershell
    scoop install starship
    ```
+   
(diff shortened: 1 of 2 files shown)


davidkna:
Thanks for the contribution, but there area already installing instructions for chocolatey in the advanced installation section on the website (`docs/installing`). The `Readme.md` file is deliberately kept short, but there might be a rework coming up soon which could include `chocolatey` instructions.

PR #3416, opened 2022-01-11
fix(elvish): Reset time for cmd_duration after each prompt

moko256:
This PR will make the starship reset the time for cmd_duration after each prompt on elvish.

#### Description

In bash, cmd_duration module shows the duration from the last prompt, but in elvish, it shows the duration from the last command run.
When inserting newline, the behavior in elvish differ from bash.
This PR will fix this.

#### Motivation and Context

Closes #

| In bash | In elvish before fix | In elvish after fix |
| -- | -- | -- |
|behavior in bash | behavior in elvish before fix | behavior in elvish after fix |

#### Checklist:

### DIFF
--- a/src/init/starship.elv
+++ b/src/init/starship.elv
@@ -3,6 +3,7 @@ set-env STARSHIP_SESSION_KEY (::STARSHIP:: session)
 
 # Define Hooks
 var cmd-status-code = 0
+var cmd-duration = 0
 
 fn starship-after-command-hook {|m|
     var error = $m[error]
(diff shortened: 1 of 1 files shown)


matchai:
For context, fish also persists command duration if no command is executed:
CleanShot 2022-01-11 at 18 39 13@2x

This is the default behavior of the shell, not something that Starship does on the shells that behave differently.
Though I appreciate the consistency, it may be an unexpected change for users who make use of cmd duration persisting on their shell.

I'd be interested in hearing what others think about the change.

davidkna:
See also the discussion #2482, there is some support for both versions. I think there might also be a similar issue concerning the status module. By the way, this should solve #3291 for elvish.

moko256:
Thanks for giving feedbacks and related informations, I close this for now and try thinking the design.

PR #3424, opened 2022-01-13
fix(pwsh): Avoid polluting the global function namespace

segevfiner:
#### Description
This is an alternative to #3386 using `New-Module` to create a new dynamic script module and `Export-ModuleMemeber` to make it not export anything, other than the global prompt function, instead of prefixing the functions.

> #3322 changed the PowerShell init in order to pass the current starship arguments to starship prompt --continuation. This PR reverts these changes and just uses plain &starship prompt --continuation (Invoke-Native isn't really needed there, and the zsh init does not pass along the arguments either) to avoid polluting the global function namespace with Invoke-Native and Get-Arguments. I also moved the continuation prompt code to ensure $STARSHIP_SHELL and $STARSHIP_SESSION_KEY are set beforehand.
> 
> To avoid putting functions into the user scope with common names, this PR prefixes all of these functions with __Starship-. Get-Arguments is removed because it isn't needed for the continuation prompt.

cc @davidkna 

#### Motivation and Context

See above.
Closes #3386

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -132,14 +132,6 @@ $null = New-Module starship {
 
     }
 
-    # Invoke Starship and set continuation prompt
-    Set-PSReadLineOption -ContinuationPrompt (
-        Invoke-Native -Executable ::STARSHIP:: -Arguments @(
-            "prompt",
-            "--continuation"
-        )
-    )
-
     # Disable virtualenv prompt, it breaks starship
     $ENV:VIRTUAL_ENV_DISABLE_PROMPT=1
 
@@ -147,6 +139,14 @@ $null = New-Module starship {
 
     # Set up the session key that will be used to store logs
     $ENV:STARSHIP_SESSION_KEY = -join ((48..57) + (65..90) + (97..122) | Get-Random -Count 16 | ForEach-Object { [char]$_ })
+
+    # Invoke Starship and set continuation prompt
+    Set-PSReadLineOption -ContinuationPrompt (
+        Invoke-Native -Executable ::STARSHIP:: -Arguments @(
+            "prompt",
+            "--continuation"
+        )
+    )
 
     Export-ModuleMember
 }
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -1,58 +1,26 @@
 #!/usr/bin/env pwsh
 
+# Create a new dynamic module so we don't pollute the global namespace with our functions and
+# variables
 $null = New-Module starship {
-    function Get-Arguments {
-        function Get-Cwd {
-            $cwd = Get-Location
-            $provider_prefix = "$($cwd.Provider.ModuleName)\$($cwd.Provider.Name)::"
-            return @{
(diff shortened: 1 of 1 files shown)


segevfiner:
Curse you flaky AWS test! 🙃 similar to #3377.

src/init/starship.ps1
```diff
@@ -1,153 +1,157 @@
...
-            # NOTE: Attempt to trim any "provider prefix" from the path string.
-            # E.g. `Microsoft.PowerShell.Core\FileSystem::Dev:\` -> `Dev:\`
-            LogicalPath =
-                if ($cwd.Path.StartsWith($provider_prefix)) {
-                    $cwd.Path.Substring($provider_prefix.Length)
-                } else {
-                    $cwd.Path
-                };
+$null = New-Module starship {
+    function Get-Arguments {
```
davidkna, line 4: Please inline `Get-Arguments` again (`--continuation` doesn't need it after all, and it makes the whole thing a bit more readable).

src/init/starship.ps1
```diff
@@ -1,153 +1,157 @@
...
-            # Resolve the provider-logical path 
-            # NOTE: Attempt to trim any "provider prefix" from the path string.
-            # E.g. `Microsoft.PowerShell.Core\FileSystem::Dev:\` -> `Dev:\`
-            LogicalPath =
-                if ($cwd.Path.StartsWith($provider_prefix)) {
-                    $cwd.Path.Substring($provider_prefix.Length)
-                } else {
-                    $cwd.Path
-                };
+$null = New-Module starship {
```
davidkna, line 3: Please add a comment explaining this line.

src/init/starship.ps1
```diff
@@ -1,153 +1,157 @@
...
+    function Get-Arguments {
+        function Get-Cwd {
```
davidkna, line 5: With the module architecture, I think this should move to the outer scope.

davidkna (commented):
I made a few suggestions, but this seems to work well overall.

segevfiner:
@davidkna Changes applied

src/init/starship.ps1
```diff
@@ -21,133 +23,130 @@ function Get-Arguments {
...
-# Get arguments for starship continuation prompt
-$arguments = Get-Arguments
-$arguments += "--continuation"
+    # Invoke Starship and set continuation prompt
+    Set-PSReadLineOption -ContinuationPrompt (
+        Invoke-Native -Executable ::STARSHIP:: -Arguments @(
+            "prompt",
+            "--continuation"
+        )
+    )
```
davidkna, line 141: Please move this after `STARSHIP_SHELL` and `STARSHIP_SESSION_KEY`, or the continuation prompt could get escaped for a different shell.

davidkna (approved):
LGTM

segevfiner:
Yay... a release was made without this. 🙃

davidkna:
Thanks for figuring out PowerShell modules @segevfiner!

PR #3426, opened 2022-01-13
ci: Use `dprint` to format documentation + TOML files

chipbuster:
#### Description
Adds Rust-based standardized markdown formatting and a CI step to catch unformatted files.

#### Motivation and Context
Occasionally, a contributor will have a different markdown formatter than the typical one, resulting in lots of changes throughout a document. This has been exacerbated by the fact that many markdown formatters [don't support partial formatting](https://github.com/prettier/prettier-vscode/issues/2330), meaning you have to format the entire document to format changes.

Don't know if this link will remain valid in the future, but [here](https://github.com/starship/starship/pull/3224/commits/4e8d30b80f08c3f28da13d24c95bcbfda713cc5d) is an example of where one such accidental change happened in a PR.

I have gone with a rust-based formatting tool (even though it's less popular than well-established tools like prettier and vscode-markdown) because it's editor agnostic and doesn't require installing a new language runtime, since you already need `cargo` to develop. (Thanks matchai for the suggestion!)

#### Checklist:

### DIFF
--- a/.dprint.json
+++ b/.dprint.json
@@ -7,8 +7,8 @@
   },
   "toml": {},
   "includes": [
-    "docs/**/*.{ts,js,cjs,mjs,json}",
-    "**/*.{md,toml}"
+    "docs/**/*.{ts,js,cjs,mjs}",
+    "**/*.{md,toml,json}"
   ],
   "excludes": [
     "**/node_modules",
@@ -31,7 +31,8 @@
     "docs/tr-TR/**",
     "docs/vi-VN/**",
     "docs/zh-CN/**",
-    "docs/zh-TW/**"
+    "docs/zh-TW/**",
+    "target/"
   ],
   "plugins": [
     "https://plugins.dprint.dev/typescript-0.62.0.wasm",
--- a/.dprint.json
+++ b/.dprint.json
@@ -1,0 +1,44 @@
+{
+  "incremental": true,
+  "typescript": {},
+  "json": {},
+  "markdown": {
+    "lineWidth": 100
+  },
+  "toml": {},
+  "includes": [
+    "docs/**/*.{ts,js,cjs,mjs,json}",
+    "**/*.{md,toml}"
+  ],
+  "excludes": [
+    "**/node_modules",
+    "**/*-lock.json",
+    ".github/*",
+    "docs/ar-SA/**",
+    "docs/-SA/**",
+    "docs/ar-SA/**",
+    "docs/ckb-IR/**",
+    "docs/de-DE/**",
+    "docs/es-ES/**",
+    "docs/fr-FR/**",
+    "docs/id-ID/**",
+    "docs/it-IT/**",
+    "docs/ja-JP/**",
+    "docs/ko-KR/**",
+    "docs/nl-NL/**",
+    "docs/pl-PL/**",
+    "docs/pt-BR/**",
+    "docs/pt-PT/**",
+    "docs/ru-RU/**",
+    "docs/tr-TR/**",
+    "docs/vi-VN/**",
+    "docs/zh-CN/**",
+    "docs/zh-TW/**"
+  ],
+  "plugins": [
+    "https://plugins.dprint.dev/typescript-0.62.0.wasm",
+    "https://plugins.dprint.dev/json-0.14.0.wasm",
+    "https://plugins.dprint.dev/markdown-0.12.0.wasm",
+    "https://plugins.dprint.dev/toml-0.5.3.wasm"
+  ]
+}
--- a/.dprint.json
+++ b/.dprint.json
@@ -15,8 +15,6 @@
     "**/*-lock.json",
     ".github/*",
     "docs/ar-SA/**",
-    "docs/-SA/**",
-    "docs/ar-SA/**",
     "docs/ckb-IR/**",
     "docs/de-DE/**",
     "docs/es-ES/**",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,21 +1,21 @@
 [package]
 name = "starship"
 version = "1.2.1"
-edition = "2021"
 authors = ["Starship Contributors"]
-homepage = "https://starship.rs"
-documentation = "https://starship.rs/guide/"
-repository = "https://github.com/starship/starship"
-readme = "README.md"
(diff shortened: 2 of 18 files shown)


.dprint.json
```diff
@@ -0,0 +1,25 @@
...
+  },
+  "includes": ["docs/**/*.{ts,tsx,js,jsx,cjs,mjs,json,md,toml}", "**/*.md"],
+  "excludes": [
+    "**/node_modules",
+    "**/*-lock.json",
+    "docs/.vuepress/",
+    ".github/*"
+  ],
+  "plugins": [
+    "https://plugins.dprint.dev/typescript-0.62.0.wasm",
```
chipbuster, line 20: My thought here was to use the TS/JS/JSON formatters to hit anything in the docs directory that was directly written by us, but it seems that there are no such files (or they all exist in .vuepress).

Do we want to get rid of the typescript and JSON formatters in that case? (The TOML one should stay because it's used to format TOML in Markdown code blocks, which we have a ton of).

chipbuster: I don't think speed is an issue at all (maybe if you have the starship directory on a network drive). I just wasn't sure if .vuepress was something we should format. Will add the directory.

.dprint.json
```diff
@@ -0,0 +1,25 @@
...
+  "markdown": {
+    "lineWidth": 100
+  },
+  "toml": {
+  },
+  "includes": ["docs/**/*.{ts,tsx,js,jsx,cjs,mjs,json,md,toml}", "**/*.md"],
+  "excludes": [
+    "**/node_modules",
+    "**/*-lock.json",
+    "docs/.vuepress/",
```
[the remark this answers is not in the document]
chipbuster: Okay, I wasn't sure if those were handwritten of not.

.dprint.json
```diff
@@ -0,0 +1,25 @@
...
+    ".github/*"
+  ],
+  "plugins": [
```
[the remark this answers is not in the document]
chipbuster: I assume so. The perks of pre-1.0 software, I guess.

The `dprint` README says that the CLI tool will let us know when this needs to happen, and the release notes will provide an upgrade path, so we at least shouldn't be in a situation where upgrades break everything and we have no clue what's going on.

.dprint.json
```diff
@@ -0,0 +1,25 @@
...
+  "typescript": {
+  },
+  "json": {
+  },
+  "markdown": {
+    "lineWidth": 100
+  },
+  "toml": {
+  },
+  "includes": ["docs/**/*.{ts,tsx,js,jsx,cjs,mjs,json,md,toml}", "**/*.md"],
```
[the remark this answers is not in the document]
chipbuster: I had originally envisioned this as a docs-only change (which is why there's no Rust plugin and why only files within the docs folder are touched).

I think it should be very much possible to modify this to become a general-purpose format-as-much-as-we-can tool, but I should probably restructure the PR a little bit (particularly, which parts of CI are handled by the "code" workflow and which are handled by the "formatting" workflow) to achieve that.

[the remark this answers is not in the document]
chipbuster: Okay, workflows are now split conceptually into "compilation/testing" and "formatting/linting".

docs/ar-SA/README.md
```diff
@@ -36,7 +36,6 @@ description: Starship is the minimal, blazing fast, and extremely customizable p
```
matchai, line 36: I think it'd be best if we avoid formatting translated files this time around and have Crowdin translate the files with updated formatting.

andytom: I think this is a good idea. I just caused a lot of merge conflicts for this PR (sorry @chipbuster) by merging the most latest Crowdin PR.

chipbuster: @andytom No worries!

.dprint.json
```diff
@@ -0,0 +1,44 @@
...
+    "docs/**/*.{ts,js,cjs,mjs,json}",
+    "**/*.{md,toml}"
+  ],
+  "excludes": [
+    "**/node_modules",
+    "**/*-lock.json",
+    ".github/*",
+    "docs/ar-SA/**",
+    "docs/-SA/**",
+    "docs/ar-SA/**",
```
matchai, line 19: Looks like a few duplicate lines snuck in:
```suggestion
    "docs/ar-SA/**",
```

.dprint.json
```diff
@@ -0,0 +1,44 @@
...
+  "incremental": true,
+  "typescript": {},
+  "json": {},
+  "markdown": {
+    "lineWidth": 100
+  },
+  "toml": {},
+  "includes": [
+    "docs/**/*.{ts,js,cjs,mjs,json}",
+    "**/*.{md,toml}"
```
matchai, line 11: Do you think there'd be value in adding `**/*.json` to catch `.dprint.json` file?

chipbuster: Hmm, good idea. I think I had it in the previous version, but some stuff appears to have gotten lost in the rebase.

.dprint.json
chipbuster: d'oh

Cargo.toml
```diff
@@ -1,21 +1,21 @@
 [package]
 name = "starship"
 version = "1.2.1"
-edition = "2021"
 authors = ["Starship Contributors"]
-homepage = "https://starship.rs"
+build = "build.rs"
```
chipbuster, line 5: It seems that the plugin wants to reorg our TOML files to be alphabetized within each table. Are we okay with that?

.dprint.json
chipbuster: @matchai Fixed, and I think I've also nuked some workflow regressions which snuck in.

Cargo.toml
matchai: Looks like it's a deliberate Cargo.toml sorting order rather than alphabetical:
https://github.com/dprint/dprint-plugin-toml/blob/main/src/cargo.rs#L17

I personally like the sound of that 👍

chipbuster:
Okay, I think I've caught all the major whoopsies from the rebase onto latest master.

matchai (approved):
LGTM! 👍

PR #3427, opened 2022-01-13
ci: Block changes to crowdin-generated files

chipbuster:
#### Description

Fails a PR if changes to Crowdin-generated files are made, unless the PR comes from the `i18n_master` branch.

Depends on #3426 

#### Motivation and Context
This is less useful than it used to be since we need to manually approve workflow runs for first-time users, but creates an easy (automated) way for us to detect when Crowdin-generated files have been changed.

One note: as of right now, all the languages we have translations for are of the form `(2-3 lowercase letters)-(two uppercase letters)`. However, there are some [Crowdin language codes](https://support.crowdin.com/api/language-codes/) which do not follow this format. If we ever add one of those languages, we'll be in an interesting spot because `faq` looks like a language code, even though it's not. Since files are detected by regex, not by some sort of tree structure, it'll be a real interesting time trying to work around that.

#### Checklist:

### DIFF
--- a/.github/workflows/format-workflow.yml
+++ b/.github/workflows/format-workflow.yml
@@ -15,3 +15,12 @@ jobs:
         uses: actions/checkout@v2.4.0
       - name: Docs | Format
         uses: dprint/check@v2.0
+
+  # If this is not a Crowdin PR, block changes to translated documentation
+  block-crowdin:
+    name: Block Translated Changes
+    uses: xalvarez/prevent-file-change-action@v1.2.0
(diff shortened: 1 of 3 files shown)


chipbuster:
Amusingly, the Docs CI workflow does not run on the PR in its current state because we haven't changed any documentation or TOML files.

Commit bfe6e7a had changes which exercised this code path and confirmed that changing a Crowdin file would error the workflow (see https://github.com/starship/starship/actions/runs/1725059186)

andytom (approved):

PR #3428, opened 2022-01-13
feat: add username to Pulumi module

AaronFriel:
Adds the "username" of the current Pulumi account to the prompt.

#### Description

Parses ~/.pulumi/credentials.json to retrieve the username on the currently active account.

#### Motivation and Context

Enables users to know what their current user account is when logged in to Pulumi.

image

Added one and modified one unit test, and validated against credentials.json on Linux (Ubuntu 20.04).

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2440,13 +2440,13 @@ By default the module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Option              | Default                                          | Description                                                               |
-| ------------------- | ------------------------------------------------ | ------------------------------------------------------------------------- |
-| `format`            | `"via [$symbol($username@)$stack]($style) "`     | The format string for the module.                                         |
-| `version_format`    | `"v${raw}"`                                      | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
-| `symbol`            | `" "`                                            | A format string shown before the Pulumi stack.                          |
-| `style`             | `"bold 5"`                                       | The style for the module.                                                 |
-| `disabled`          | `false`                                          | Disables the `pulumi` module.                                             |
+| Option           | Default                                      | Description                                                               |
+| ---------------- | -------------------------------------------- | ------------------------------------------------------------------------- |
+| `format`         | `"via [$symbol($username@)$stack]($style) "` | The format string for the module.                                         |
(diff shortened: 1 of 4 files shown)


docs/config/README.md
```diff
@@ -2407,20 +2407,21 @@ By default the module will be shown if any of the following conditions are met:
...
+| `style`             | `"bold 5"`                                       | The style for the module.                                                 |
+| `disabled`          | `false`                                          | Disables the `pulumi` module.                                             |
 
 ### Variables
 
 | Variable | Example    | Description                          |
 | -------- | ---------- | ------------------------------------ |
 | version  | `v0.12.24` | The version of `pulumi`              |
 | stack    | `dev`      | The current Pulumi stack             |
+| username | `foo`      | The current Pulumi username          |
```
chipbuster, line 2424: Can we use a slightly more descriptive example here?

chipbuster (commented):
1 minor change that's just me being picky about docs.

I don't know much about Pulumi, so it'd be nice if we could get a review from someone who does, but the Rust side of this looks solid enough.

docs/config/README.md
AaronFriel: Sure, this would be any (I'd have to check) url safe string though, for example `friel`, or `chipbuster`.

AaronFriel:
@chipbuster I'm currently at Pulumi, we had a quick chat about parsing the username though I don't think we make any guarantees about future proofing `credentials.json`.

chipbuster:
Well, I think that counts then :)

docs/config/README.md
chipbuster: Sure, that sounds good. Or we could come up with a dummy username like `admin` or `user`, just something that carries some connotations with it.

AaronFriel: /shrug Changed to `alice`

AaronFriel:
🤞

chipbuster (approved):
:+1: LGTM

AaronFriel:
@chipbuster I think this needs you or someone else to click the go button :)

chipbuster:
Indeed it does.

chipbuster:
@AaronFriel Thanks for the contribution!

PR #3435, opened 2022-01-14
fix: allow compilation without battery feature

davidkna:
#### Description

Recent changes made the lifetime `'a` become unused if the `battery` feature was disabled. This PR adds a `PhantomData` field to `context` that hold a phantom reference to the lifetime.

#### Motivation and Context

Closes #3434

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [x] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #3440, opened 2022-01-15
ci: Add jobs for `cargo check` with all features and no features

mandx:
#### Description
Augmenting #3435: These jobs will check that compilation still succeeds when no features are selected and with all features selected.

#### Motivation and Context
I was also hit by #3434, and quickly #3435 was added, so I figured this would be useful. The idea would be that, when in the future new features are added, these quick checks can prevent miss-compilation for users that like to tweak with feature sets.

While these CI operations can be run under the same `cargo_check` job, I decided to add two more independent jobs so they run in parallel. I'll be happy to roll them up into the `cargo_check` job if y'all decide it's not worth it.

```
# This always compiles
$ cargo check --workspace --locked --all-features
# This would not compile before #3435
$ cargo check --workspace --locked --no-default-features
```

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -88,6 +88,7 @@ jobs:
   # Ensure that the project could be successfully compiled with no features enabled
   cargo_check_no_features:
     name: Compile with no features enabled
+    needs: cargo_check
     runs-on: ubuntu-latest
     steps:
       - name: Setup | Checkout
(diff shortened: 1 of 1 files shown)


davidkna:
Please add `needs: cargo_check` to the new tests, because if the regular `check` fails, this will likely fail too.

mandx:
@davidkna done in [f9a7177](https://github.com/starship/starship/pull/3440/commits/f9a717789de4cb4c9f43e29fc6b3508d06af9e08)

andytom (approved):

davidkna:
Thanks for ensuring this doesn't break in the future @mandx and thanks for reviewing @andytom!

PR #3445, opened 2022-01-15
feat(pwsh): Set WorkingDirectory to CWD

### LINKED ISSUE #2965: Working directory is not correct in custom modules when starship is invoked with `--path`
#### Current Behavior
When I have a custom command, the working directory of the custom command is not set to the path argument from the commandline. When I invoke `starship prompt` under a svn directory, then my custom command which runs svn info works correctly. When I invoke `starship prompt -p <my svn dir>` from a different directory, my svn info is not included in the prompt. It seems that the powershell hook script doesn't set the current directory before it calls starship, rather it passes the `--path` and `--logical-path` arguments.

#### Expected Behavior
I expect `starship prompt` run in a directory X to behave identically to `starship prompt -p X` run from any directory.

#### Possible Solution
Either change the powershell prompt script to fix the working directory, or have all commands invoked by custom module (and others?) set the directory before they execute.

Alternatively, document this behaviour and have some environment variable which is the directory we're trying to make a prompt for, so I can use that in my custom module configs.

#### Environment
- Starship version: 0.56.0
- Git Commit Hash: f64cfa6a
- Branch/Tag: v0.56.0
- Rust Version: rustc 1.53.0 (53cb7b09b 2021-06-17)
- Build Time: 2021-07-13 22:05:09

```bash
<unknown config>
```

#### Starship Configuration

```toml
[status]
disabled = false

[shell]
disabled = false
fish_indicator = ''
powershell_indicator = '_'
bash_indicator = ''

[custom.svn]
command = """ svn info | grep Rela | cut -d'/' -f2-3 | sed "s/releases/R/; s/trunk/T/; s/branches\\///; s/\\//:/" """
description = "svn info"
format = '(SVN: $output)'
shell = ["C:/msys64/usr/bin/sh.exe"]
when = "true"

[custom.svn2]
command = "svn info"
description = "svn info"
format = '(SVN info: $output)'
when = "true"
```

ealap:
#### Motivation and Context

Closes #2965

#### Checklist:


davidkna:
I don't think this is the best approach. I tested the invocation in my shell, and it looks like this will prevent starship from running when the working directory is located in non-fs provider paths, like the registry.

Instead, the custom module code should be change to explicitly set the cwd.

thorstenkampe:
I ran into the issue myself and it's hard to find the root cause. The issue doesn't just affect the `custom` module but also the `env_var` module (for instance).
```toml
[env_var]
format   = '[$env_value](bright-white) '
variable = 'PWD'
```
(PowerShell on Windows has no environment variable `PWD` but on Linux it has)

thorstenkampe:
@ealap @davidkna the `custom` module is already completely broken when the provider is not the file system. Why not simply specify that we want a file system location?
```
> Get-Location

Path
----
C:\

> Set-Location -Path HKLM:\Software\Microsoft

> Get-Location

Path
----
HKLM:\Software\Microsoft

> Get-Location -PSProvider FileSystem

Path
----
C:\
```
Quod erat demonstrandum.

davidkna:
I don't think any of the proposed solution will set `PWD`.

As I said, I would prefer a solution that fixes this in the rust code (by populating `cmd.current_dir` with `context.current_dir` in `src/modules/custom.rs` for every `Command`).

Nevertheless, setting the cwd to `Get-Location -PSProvider FileSystem` sounds acceptable if it works with long paths (this might require #3590). Also, it might be better to fall back to `nul` or `/dev/null` if the provider is not `FileSystem` if it doesn't break anything.

PR #3459, opened 2022-01-18
fix(elvish): fix too large status code on Windows

moko256:
This PR will fix too large status code on elvish shell on Windows.

#### Description
The status module parse received string status code as `i32`.
The range of the status code is `i32` in powershell, but it is `u32` in elvish.
On *nix that the status code is from 0 to 255, it does not cause the crash, but on Windows that the NTSTATUS is >= 0xC0000000, it causes crash.
This PR will add rounding status code to `i32`, fix crash, and make the behavior in elvish the same in powershell.

#### Motivation and Context

Closes #

| In powershell | In elvish before fix | In elvish after fix |
|-|-|-|
A screenshot in powershell | A screenshot in elvish before fix | A screenshot in elvish after fix

```toml
[status]
disabled = false
format = "[$symbol$status $hex_status]($style) "
```

#### Checklist:

### DIFF
--- a/src/init/starship.elv
+++ b/src/init/starship.elv
@@ -11,6 +11,7 @@ fn starship-after-command-hook {|m|
     } else {
         try {
             set cmd-status-code = $error[reason][exit-status]
+            set cmd-status-code = (- (% (+ $cmd-status-code 0x80000000) 0x100000000) 0x80000000)
         } except {
             # The error is from the built-in commands and they have no status code.
             set cmd-status-code = 1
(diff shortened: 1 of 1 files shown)


davidkna:
I believe this has mostly been already fixed in v1.2.1 via #3399. Can you confirm this?
There's still some parts of the `status` that make limited assumption about the status code range, but I would rather see them fixed there instead of the init scripts.

moko256:
Sorry, I missed #3399. I'll close this.

PR #3462, opened 2022-01-19
fix(status): Enable to convert from i64 to hex_status by casting instead of parsing status.

moko256:
This PR will enable to convert from i64 to hex_status by casting instead of parsing status.

#### Description
In elvish shell on Windows, the status code is passed as u32, and it can't parse as i32, so `hex_status` is same as `status`.
This PR will replace parsing status to i32 by casting from i64 to i32 and accept this.
The status code out of i32 will be rounded by underflow/overflow.

#### Motivation and Context

Closes #
Related #3459

| In powershell | In elvish before fix | In elvish after fix |
|-|-|-|
| A screenshot in powershell | A screenshot in elvish before fix | A screenshot in elvish after fix |

```toml
[status]
disabled = false
format = "[$symbol$status $hex_status]($style) "
```

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -569,7 +569,7 @@ pub enum Target {
 /// Properties as passed on from the shell as arguments
 #[derive(Parser, Debug)]
 pub struct Properties {
-    /// The status code of the previously run command, in unsigned or signed 32bit integer
+    /// The status code of the previously run command as an unsigned or signed 32bit integer
     #[clap(short = 's', long = "status")]
     pub status_code: Option<String>,
     /// Bash, Fish and Zsh support returning codes for each process in a pipeline.
--- a/src/context.rs
+++ b/src/context.rs
@@ -569,7 +569,7 @@ pub enum Target {
 /// Properties as passed on from the shell as arguments
 #[derive(Parser, Debug)]
 pub struct Properties {
-    /// The status code of the previously run command
+    /// The status code of the previously run command, in unsigned or signed 32bit integer
     #[clap(short = 's', long = "status")]
     pub status_code: Option<String>,
     /// Bash, Fish and Zsh support returning codes for each process in a pipeline.
--- a/src/modules/status.rs
+++ b/src/modules/status.rs
@@ -6,7 +6,7 @@ use crate::configs::status::StatusConfig;
 use crate::formatter::{string_formatter::StringFormatterError, StringFormatter};
 use crate::segment::Segment;
 
-type ExitCode = i64;
+type ExitCode = i32;
 type SignalNumber = u32;
 #[derive(PartialEq)]
 enum PipeStatusStatus<'a> {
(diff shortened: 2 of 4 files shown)


src/modules/status.rs
```diff
@@ -109,10 +109,7 @@ fn format_exit_code<'a>(
         }
     };
 
-    let hex_status = exit_code
-        .parse::<i32>()
-        .ok()
-        .map(|code| format!("0x{:X}", code));
+    let hex_status = format!("0x{:X}", exit_code_int as u32);
```
davidkna, line 112: I think this will just work and avoid truncation:

```suggestion
    let hex_status = format!("0x{:X}", exit_code_int);
```

moko256: Avoiding truncation makes hex_status longer when status code is negative number like this:
```
✖-1 0xFFFFFFFFFFFFFFFF ❯
```
On Windows, status code is in range of u32 or i32 (e.g, MSVC `exit()` receive `int`), so I think this cast is necessary.

davidkna: In that case, should this also apply to `int` too? Furthermore, maybe this should be restricted to Windows (with another `type ExitCodeTarget`?) and I feel like this should be documented in the `status` docs.

moko256: Whether i32 or u32 the status fits in is consistent in one shell, and in supported platform the status code fits in 32bit.
Until the platform that have larger status code than 32bit will be supported, I think that assuming that the status code fits in 32bit is no problem and hex_status doesn't have to be restricted only for Windows. Similally, I think it doesn't have to cast like hex_status.
But I agree that the size of status code that the module receives should be documented.

moko256: Instead of that, it is possible to make ExitCode i32.
```diff
diff --git a/src/modules/status.rs b/src/modules/status.rs
index 5f643a95..31d26c2c 100644
--- a/src/modules/status.rs
+++ b/src/modules/status.rs
@@ -6,7 +6,7 @@ use crate::configs::status::StatusConfig;
 use crate::formatter::{string_formatter::StringFormatterError, StringFormatter};
 use crate::segment::Segment;
 
-type ExitCode = i64;
+type ExitCode = i32;
 type SignalNumber = u32;
 #[derive(PartialEq)]
 enum PipeStatusStatus<'a> {
@@ -101,15 +101,16 @@ fn format_exit_code<'a>(
     config: &'a StatusConfig,
     context: &'a Context,
 ) -> Result<Vec<Segment>, StringFormatterError> {
-    let exit_code_int: ExitCode = match exit_code.parse() {
-        Ok(i) => i,
+    // First, parse as i64 to accept both i32 or u32, then normalize to u32.
+    let exit_code_int: ExitCode = match exit_code.parse::<i64>() {
+        Ok(i) => i as ExitCode,
         Err(_) => {
             log::warn!("Error parsing exit_code string to int");
             return Ok(Vec::new());
         }
     };
 
-    let hex_status = format!("0x{:X}", exit_code_int as u32);
+    let hex_status = format!("0x{:X}", exit_code_int);
 
     let common_meaning = status_common_meaning(exit_code_int);
 

```

davidkna: In that case, I think it's better to leave `int` as-is for now.

moko256: Done, I pushed the diff to this PR. Thanks for reviewing!

davidkna: Please also add a note in the `status` module description in `docs/config/README.md` about this and that `hex_status` is unsigned.

src/context.rs
```diff
@@ -569,7 +569,7 @@ pub enum Target {
 /// Properties as passed on from the shell as arguments
 #[derive(Parser, Debug)]
 pub struct Properties {
-    /// The status code of the previously run command
+    /// The status code of the previously run command, in unsigned or signed 32bit integer
```
davidkna, line 572:
```suggestion
    /// The status code of the previously run command as an unsigned or signed 32bit integer
```

src/modules/status.rs
moko256: Done.

davidkna (approved):

davidkna:
Thanks for handling this @moko256!

PR #3467, opened 2022-01-20
feat(config): Adds support for --profile <custom profile name>

### LINKED ISSUE #3053: accept argument for alternate config file
It would be really nice if starship accepted an argument that allowed it to use an alternate config file.  I was surprised it seems not to already be able to do this, but I looked through `starship --help` commands and didn't see it.

The use case I have in mind is the ability to toggle different prompt forms.  In particular, I'd really like the ability to toggle between a single-line and multi-line font.  This would be really easy to set up in `fish`, but without the ability to set the config location, one would have to actually swap the file, which seems excessive.

A (perhaps preferable) alternative would be to allow redundant values in the single config file (especially for the main `format` string) which could be chosen when `starship prompt` is called.

### LINKED ISSUE #3111: short prompt on enter
## Feature Request

Hi! I'm migrating out of the zSH world and there's one quirk that I love, is to have a 'short' prompt when hitting only the enter character, it's implemented in  bash, through a zpresto plugin, using the preexec / precmd hooks. 
If I enter a full command, then the full prompt is rendered.

image

I'm trying to replicate it, either in `zsh` or `nushell` but this is a bit whacky.

with nushell, I managed to bind on the command duration, but I need to pass on a full other STARSHIP_CONFIG

```
def starship_prompt [] {
  let dur = $nu.env.CMD_DURATION_MS;
  if $dur == "0" {
    let-env STARSHIP_CONFIG = (build-string $nu.env.HOME "/.config/starship-short.toml")
    starship prompt --cmd-duration $nu.env.CMD_DURATION_MS
  } {
    starship prompt --cmd-duration $nu.env.CMD_DURATION_MS
  }
}
```

For ZSH,  it's done with some additions to the default `/usr/local/bin/starship init zsh --print-full-init`

Ideally,I think the solution would look like a new short_prompt "template" (not sure if it's the right term), similar to the right_prompt.
 
Currently passing anoter config work, but unfortunately, that's problematic as it's not super clean and limits the amount of sharing we can do between the configs, for example, my short prompt is the end of the long prompt.

### LINKED ISSUE #3389: Privacy mode: disable some modules
## Feature Request: Privacy mode

It would be great to be able to toggle a privacy mode that disables some modules (would be defined in `starship.toml`).

I would like to quickly turn off the time, username, host, etc when I am screen-sharing for example.

flovilmart:
#### Description

PR to add support for short prompts or alternate prompts.
Allows to render a temporary prompt while reusing the configurations from the main prompt.

Adds a `--profile=short` CLI param 

this will lookup the `profiles` section of the config, and render the format with the name:

```
[profiles]
short = "$time$character"
danger="!!!! $character"
```

#### Motivation and Context

Closes #3111
Closes #3053
Closes #3389

#### Checklist:


chipbuster:
Just a thought: perhaps it'd be better to gate this behind a flag like `--format=short` in case we want to add more styles in the future?

I don't know if `format` is a good name for that flag though, since the word is pretty overloaded.

flovilmart:
@chipbuster great idea! We can have a section

```
[prompts]
short = "$character"
cloud = "$aws$kubernetes$character"
sudo = "SUDO #"
...
```

And the the user could provide as many prompts they see fit and toggle the one they want on current env vars in their zsh / fish config

flovilmart:
@chipbuster I went ahead, I'm no rust expert but I [refactored](https://github.com/starship/starship/pull/3467/files#diff-72ef898a77aa664fcc63de579c878c87e101688fc902cb1922fc884d7b2bea65R415) the `load_formatter_and_modules` method to better accomodate the new flow, it will lead to less allocations. One caveat is that some errors on side prompts won't be printed anymore.

I had to remove `derive(Copy)` from the `enum Target` so it encapsulate the string passed by the CLI. There are some loose ends on the docs etc... waiting for feedback before committing too much on this!

chipbuster:
Will have a gander at reviewing this tomorrow.

jankatins:
I guess this also needs changes in the shell integrations to automatically be used? If so: can this be finally be used to get rid of a leading line break in a new shell session?

flovilmart:
@jankatins what do you have in mind; not sure I follow.

matchai:
@jankatins The issue of the leading line break is being tracked at #560. If you'd like to see it addressed, you're welcome to propose suggestions there or create a PR.

matchai:
May I recommend using "profile" over "prompt" or "format"?
I'm not sure how clear it is as a name, but it's not a term we've previously used in Starship configuration.

flovilmart:
@matchai so `starship prompt --profile=short` and
```
[profiles]
short="test$character
```

for the config?

chipbuster:
I had originally envisioned that if you used `--format=short`, the prompt would look for a `short_format` and then render that. Perhaps the dynamic lookup involved with this is too much.

A separate section in the config is also viable, but I'm a little skittish about adding a table called `profiles`, because we want to eventually merge a presets/profiles change, and I don't know if we want to keep that word reserved until then.

flovilmart:
> Perhaps the dynamic lookup involved with this is too much.

the configuration is very structured currently, adding a dynamic lookup a the top level would work, but is probably very involved; Not sure if it's worth the hassle 😅 and the maintenance cost + the precedent it causes that we have some random keys that may be used for random stuff. Your call after all, just my 2cents.

> because we want to eventually merge a presets/profiles change

What's the idea behind this?

chipbuster:
> the precedent it causes that we have some random keys that may be used for random stuff.

Yeah, this was my concern as well. Probably better to keep it structured like this.

> What's the idea behind this?

We currently keep [presets configs](https://starship.rs/presets/) on the website, for things like pure NerdFonts, or ASCII-only. The goal would be to eventually be able to do something like `STARSHIP_PRESET=lots_of_red` or put something like

```
[scala]
preset = "emoji"
```

in `starship.toml` to pull default options from some selected set.

The issue with this is that we haven't really architected this yet (I've prioritized merging PRs and getting the benchmark system online ahead of it) so I don't really know what that's going to look like. Maybe we'll just need presets, maybe we'll need both presets and profiles, I really have no clue.

flovilmart:
I see! those "presets" would be alternate configs.
perhaps those presets could be declared as such:

```
[preset.<name>]
format = ""

[preset.name.aws]
symbol = "A"
```

This could help maintain the consistency, and the type would be `preset: IndexMap<String, StarshipConfig>`;
easily representable with toml and rust's type system

src/context.rs
```diff
@@ -559,11 +559,12 @@ pub enum Shell {
 }
 
 /// Which kind of prompt target to print (main prompt, rprompt, ...)
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum Target {
     Main,
     Right,
     Continuation,
+    Profile(String),
```
chipbuster, line 567: `String` isn't `Copy`, so it makes sense that `Copy` is no longer derivable on `Target`.

Removing `Copy` from things can sometimes result in huge ergonomic pains, but based on the fact that your code compiles and I haven't seen a `.clone()` explosion yet, I'm going to guess that it wasn't *that* bad in this case.

src/print.rs
```diff
@@ -114,13 +115,13 @@ pub fn get_prompt(context: Context) -> String {
 
     let module_strings = root_module.ansi_strings_for_shell(context.shell, Some(context.width));
     if config.add_newline && context.target != Target::Continuation {
-        // continuation prompts normally do not include newlines, but they can
+        // continuation profiles normally do not include newlines, but they can
```
chipbuster, line 118: I think in this case, it's better to call it the "continuation prompt". That's not naming from our code, it's terminology that's established from the shell. (See e.g. [this answer](https://unix.stackexchange.com/a/193660/379272))

src/print.rs
```diff
@@ -114,13 +115,13 @@ pub fn get_prompt(context: Context) -> String {
...
         writeln!(buf).unwrap();
     }
     write!(buf, "{}", ANSIStrings(&module_strings)).unwrap();
 
     if context.target == Target::Right {
-        // right prompts generally do not allow newlines
+        // right profiles generally do not allow newlines
```
chipbuster, line 124: Same here: in almost all shells, if you look at the init scripts, the variable/function that controls this is named some variant of `RIGHT_PROMPT`.

src/print.rs
```diff
@@ -414,31 +415,36 @@ fn all_modules_uniq(module_list: &BTreeSet<String>) -> Vec<String> {
...
-    let cformatter = StringFormatter::new(&config.continuation_prompt);
-    if lformatter.is_err() {
-        log::error!("Error parsing `format`")
-    }
-    if rformatter.is_err() {
-        log::error!("Error parsing `right_format`")
-    }
-    if cformatter.is_err() {
-        log::error!("Error parsing `continuation_prompt`")
+    let (formatter, prompt_name) = match &context.target {
```
chipbuster, line 418: Should this be `profile_name`? Or is `prompt_name` correct in this context?

chipbuster (commented):
Nice! All the comments here are me being nitpicky about naming/commenting (because I guess that's my personal branding at this point), but I think this is overall headed in a healthy direction.

src/configs/mod.rs
```diff
@@ -84,6 +84,7 @@ pub struct FullConfig<'a> {
     pub scan_timeout: u64,
     pub command_timeout: u64,
     pub add_newline: bool,
+    profiles: IndexMap<String, String>,
```
chipbuster, line 87: Is there a particular reason we need an IndexMap instead of a HashMap? Is this because of something we plan to do with iteration order later, or is this a requirement placed on us by Serde?

src/context.rs
flovilmart: That was my conclusion as well. I was afraid things would start to badly fail.

src/print.rs
flovilmart: Bad replace; sorry

src/print.rs
flovilmart: Bar replace; will fix

jankatins:
@flovilmart @matchai 

My understanding of this change was that it also needs changes in the shell code to detect when to use which prompt (which are not part if this pr?). 

If such shell code can already recognize "no command was run", then it shouldn't be hard to recognize "nothing was run at all yet"? Anyway, will wait for any shell code to show up and then see if i can come up with anything to address #560.

flovilmart:
@jankatins not sure how those relate to each other. This PR's only goal is to render alternate prompts. For example, one user may want to use a different prompt layout depending on some env vars for example.

chipbuster:
@jankatins This PR indirectly enables that, but it would still be up to someone to write the shell adapter code to make that actually run. I wouldn't want that directly included in this PR either, if only for concerns of logical separation (digging through PR history for init made me appreciate how annoying it can be when multiple different changes get lumped together).

Also, note that "no command has been run yet" would only be a partial fix to #560, as you can still see that on e.g. screen clear via Ctrl+L.

src/configs/mod.rs
flovilmart: Nope, just because I saw a similar requirement below implemented with an IndexMap, rust n00b here 😅. If you feel HashMap is better, will use it!

src/print.rs
```diff
@@ -414,31 +415,36 @@ fn all_modules_uniq(module_list: &BTreeSet<String>) -> Vec<String> {
...
     }
 
-    match (lformatter, rformatter, cformatter) {
-        (Ok(lf), Ok(rf), Ok(cf)) => {
+    match formatter {
+        Ok(lf) => {
             let mut modules: BTreeSet<String> = BTreeSet::new();
             if context.target != Target::Continuation {
                 modules.extend(lf.get_variables());
-                modules.extend(rf.get_variables());
```
davidkna, line 435: I think both `lf` and `rf` were added to `modules` here to ensure that modules listed in `right_format` are automatically excluded from `$all` in `$format`. `context.target == Target::Main` might be enough here, but I'm not sure if this is used elsewhere.

flovilmart: Ok, now I get why we rendered all formatters 🥲 will see how I can factor better

flovilmart: @davidkna I've updated the code, anything else you had in mind?

src/print.rs
flovilmart: fixed

src/configs/mod.rs
flovilmart: fixed

src/print.rs
```diff
@@ -413,10 +414,19 @@ fn all_modules_uniq(module_list: &BTreeSet<String>) -> Vec<String> {
...
+    let (pformatter, profile_name) = match &context.target {
+        Target::Profile(name) => (
+            match config.profiles.get(name) {
+                Some(format) => StringFormatter::new(format),
+                _ => Err(StringFormatterError::Custom("Invalid Profile".to_string())),
+            },
+            Some(name),
+        ),
+        _ => (StringFormatter::new(""), None),
+    };
```
davidkna, line 429: If this isn't too difficult to refactor, I would prefer If only the required targets get initialized like in the previous version of your code.

src/print.rs
```diff
@@ -426,9 +436,13 @@ fn load_formatter_and_modules<'a>(context: &'a Context) -> (StringFormatter<'a>,
...
+    if pformatter.is_err() {
+        if let Some(profile) = profile_name {
+            log::error!("Error parsing profile `{}`", profile)
+        }
+    }
+    match (lformatter, rformatter, cformatter, pformatter) {
+        (Ok(lf), Ok(rf), Ok(cf), Ok(pf)) => {
             let mut modules: BTreeSet<String> = BTreeSet::new();
             if context.target != Target::Continuation {
                 modules.extend(lf.get_variables());
```
davidkna, line 448: This should either be limited to `matches!(context.target, Target::Main | Target::Right)` instead. Alternatively, this should maybe also handle `Profile(_)` and add either `lf` or `pf` depending on which is active to `modules` allow both to ignore `rf` modules from `$all`.

davidkna (commented):
If this is added, `explain` and `timings` will also need to be improved to better handle multiple format strings. Right now, they just look at `modules` for `Target::Main` (which means `lf` and `rf` without any logical separation).
Since this PR mostly lays groundwork for eventual proper adaption, I think this doesn't necessarily need to be included in this PR.

src/print.rs
flovilmart: If I understand correctly, we initialize either the Main OR Profile(_) + Right and pass on modules both. Is that right?

src/print.rs
flovilmart: sure that make total sense!

src/print.rs
davidkna: Yes.

flovilmart: @davidkna just updated the code, let me know what you think.

chipbuster:
@flovilmart Unfortunately, due to the recent release of Rust 1.59.0, you'll need to rebase on master to get the tests to run.

chipbuster:
Sorry for the delay! The code here looks good to me, and testing locally shows that it does what we want.

One final question: how should we document this? Does this go in its own section in the docs? Or does it act like a pseudo-module within the documentation section?

(Also, unless anyone objects, I think I might want to roll e.g. `right` and `continuation` into `profiles` by default the next time we make a breaking change. It might be worth noting this in the docs as well).

Once we have the docs in place, I think we should be ready to merge.

chipbuster:
Since I lost this multiple times while looking at other stuff, here's a comment from David:

> If this is added, explain and timings will also need to be improved to better handle multiple format strings. Right now, they just look at modules for Target::Main (which means lf and rf without any logical separation).

I agree that we don't really need this done for this PR, but we should probably open an issue once this is merged.

flovilmart:
Awesome @chipbuster ! I'll check for the docs. 

Thanks for the added commit - this is helpful 😅

flovilmart:
@chipbuster any suggestions/guide to contribute to the guide?

chipbuster:
I think this should be an advanced configuration for a very simple reason: users will need to write their own init/integration scripts to be able to take advantage of this feature. Without doing that, the added profiles are unreachable. So perhaps we add a section in advanced configuration about this, note that it uses the same syntax as the other prompt format strings, and give an example of how it could be used?

davidkna:
With the clap [`env` feature](https://docs.rs/clap/latest/clap/struct.Arg.html#method.env), it would be easy to allow overwriting the main prompt (which does not use any flags) with an env var.

flovilmart:
agreed @chipbuster  this is advanced!

chipbuster:
@davidkna Would that be appropriate for this PR? I'd imagine that with something like that, we might want to make a more general "runtime override" PR, while we can start using this right away.

There might be backwards compatibility issues I'm not thinking of at the moment though.

davidkna:
> @davidkna Would that be appropriate for this PR? I'd imagine that with something like that, we might want to make a more general "runtime override" PR, while we can start using this right away.
> 
> There might be backwards compatibility issues I'm not thinking of at the moment though.

I feel like it would make the profile switch much easier to take advantage of, but I'm not opposed to handling this later.

protiumx:
hey guys, any news about this?
My two cents: 
I think `profile` in this PRs could be better understood as `format`. Having multiple formats per file sounds awesome.
On the other hand, `profile` could be the ability of having multiple config files. Switching could be done by `starship use <profile/config>`. So at the end `starship` only needs to provide the `use` command to hot swap the config.

At the moment this can be achieved by replacing the env variable and reloading starship. Which is fairly fine.
But I'd like to automate the swapping, for instance I want to have a script that will call `starship use power` when my laptop is disconnected from the power.

For this to work we need starship to load folder with config files and use the one called `default` if no other specified.
Any thoughts?

davidkna (approved):
It's been a while, but I think this looks fine.
@chipbuster do you have any outstanding concerns about this?

chipbuster:
@davidkna None about the existing code. I still don't see docs, but at this point I think I'm fine merging this undocumented and the putting up new issues about it. I think we'll need to create a few issues on this anyways, to address the other comments brought up during review.

PR #3481, opened 2022-01-21
ci: Only run translation block work flow on PRs 

andytom:
#### Description

Have updated the workflow to only run the "Block Translated Changes" job
on PRs since it will fail when run on pushes to a branch. Have fixed the
formatting while I was here.

#### Checklist:

### DIFF
--- a/.github/workflows/format-workflow.yml
+++ b/.github/workflows/format-workflow.yml
@@ -1,9 +1,10 @@
+---
 name: Format + Docs Workflow
 on:
   push:
-    paths: ["docs/**","**.md","**.toml","**.js","**.json","**.ts"]
+    paths: ["docs/**", "**.md", "**.toml", "**.js", "**.json", "**.ts"]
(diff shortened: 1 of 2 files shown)


.github/workflows/format-workflow.yml
```diff
@@ -1,9 +1,10 @@
+---
 name: Format + Docs Workflow
 on:
   push:
-    paths: ["docs/**","**.md","**.toml","**.js","**.json","**.ts"]
+    paths: ["docs/**", "**.md", "**.toml", "**.js", "**.json", "**.ts"]
```
chipbuster, line 5: Could we equivalently fix this just by deleting `on push`, or does that result in other issues?

chipbuster: No, that applies to the entire workflow. Ignore me.

chipbuster (approved):
:+1:

PR #3489, opened 2022-01-22
feat: add container_only config option to hostname.

misobarisic:
#### Description

I have added `container_only` option to the `hostname` module. It works by checking for `/run/.containerenv` or `/.dockerenv`.

#### Motivation and Context

I'm sharing my home folder between my host machine and containers and I wanted to easily see the hostname inside my containers without seeing it on the host. Supports podman/docker.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1627,7 +1627,7 @@ The `hostname` module shows the system hostname.
 | Option           | Default                     | Description                                                                                                                          |
 | ---------------- | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------ |
 | `ssh_only`       | `true`                      | Only show hostname when connected to an SSH session.                                                                                 |
-| `container_only` | `false`                     | Only show hostname in a container.
+| `container_only` | `false`                     | Only show hostname in a container.                                                                                                   |
 | `trim_at`        | `"."`                       | String that the hostname is cut off at, after the first match. `"."` will stop after the first dot. `""` will disable any truncation |
 | `format`         | `"[$hostname]($style) in "` | The format for the module.                                                                                                           |
 | `style`          | `"bold dimmed green"`       | The style for the module.                                                                                                            |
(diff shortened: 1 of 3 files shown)


chipbuster:
Ignoring what the code actually does at the moment, what do we think should happen when both `ssh_only` and `container_only` are both `true`?

If we go with boolean logic, the implication should be that we should only display hostname when we're SSHed and in a container (SSH'ed into a container?). But it seems to me that it'd be much more practical to have the module display if we're in an SSH environment *or* in a container.

Looking at your PR, it seems you implement the more practical option. Is there a way we can make this more obvious in the way the options are written?

misobarisic:
> If we go with boolean logic, the implication should be that we should only display hostname when we're SSHed and in a container (SSH'ed into a container?). But it seems to me that it'd be much more practical to have the module display if we're in an SSH environment *or* in a container.

Yup you're correct. It would imply the chaining of those 2 conditions and should technically be applied in such a way.

I agree with you. It would be much more practical having it act as "or". The docs need to be updated appropriately.

chipbuster:
Question: can you think of other situations in which it might be nice to display the hostname that are not currently covered by our modules?

If this is the only additional change to hostname forseeable, I might be okay with two booleans, but I'm wondering if it might be worth the work now to implement something like

```toml
[hostname]
display_when: ["ssh", "container", "always"]
```

It would be more general, but it also introduces a whole new host of problems that we have to solve.

misobarisic:
> Question: can you think of other situations in which it might be nice to display the hostname that are not currently covered by our modules?

I don't think so.

> If this is the only additional change to hostname forseeable, I might be okay with two booleans, but I'm wondering if it might be worth the work now to implement something like
> 
> ```toml
> [hostname]
> display_when = ["ssh", "container", "always"]
> ```
> 
> It would be more general, but it also introduces a whole new host of problems that we have to solve.

I like this. Making it more general would mean easier future feature implementations if needed.

PR #2666, opened 2021-04-29
fix(gcloud): add double space after cloud emoji symbol

vladimyr:
#### Description

Symbols using double-width emojis typically use double trailing space as listed here: https://github.com/starship/starship/issues/2616#issuecomment-829338508

#### Motivation and Context

Ensure GCloud follows established conventions.

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #3503, opened 2022-01-24
fix: Remove duplicate space in default symbol for the Gcloud plugin

mdryden:
The default symbol value has an extra space, making it look different than other default configurations.

#### Description

The default symbol value has an extra space, making it look different than other default configurations.

#### Motivation and Context

The default looks mis-aligned and doesn't fit with the other defaults as far as I can tell.  It also doesn't match what the documents claim that it is (cloud icon with a single space).

image

I have not manually built starship, so I have not tested this change directly.  I have tested the corrected symbol by overriding in my toml file, which works as expected.

#### Checklist:

### DIFF
--- a/src/configs/gcloud.rs
+++ b/src/configs/gcloud.rs
@@ -16,7 +16,7 @@ impl<'a> Default for GcloudConfig<'a> {
     fn default() -> Self {
         GcloudConfig {
             format: "on [$symbol$account(@$domain)(\\($region\\))]($style) ",
-            symbol: "☁️  ",
+            symbol: "☁️ ",
             style: "bold blue",
             disabled: false,
             region_aliases: HashMap::new(),


mdryden:
I see that there were some unit tests here which must have been coded to use the extra space.  I can take a shot at fixing those and re-submitting.

chipbuster:
This was part of a fix made in #2666. See [this link](https://github.com/starship/starship/issues/2616#issuecomment-829338508) for more details on why it was necessary.

The short version is that without this, certain terminals cannot correctly display double-width emoji, and we consider it better to have uneven spacing by default than to break rendering by default.

mdryden:
I see, thank-you for the explanation, @chipbuster.

I will close the pull request, as I can't dedicate the time to try and change the implementation.  I guess the toml override will have to do :)

mdryden:
Just an update for anyone who googles this up...

I noticed the double space on Konsole, which I was using as my primary terminal, but then when I enabled starship in the VSCode terminal, I saw it with only a single space (as described in the related issue).

I tested a few more terminals, and decided to ditch Konsole rather than editing the toml file, because it seemed easier than trying to change the behavior on VSCode.

Some test results:
Konsole - two spaces
Gnome Terminal - one space
VSCode Terminal - one space
XFCE Terminal - one space

PR #2461, opened 2021-03-15
feat(aws): hide AWS prompt if no profile section is associated to current profile

clvanbrandt:
#### Description
The goal of this PR is to add some conditions to disable the AWS prompt by checking the existence of a profile section in either the  credentials or the configuration file (if any profile environment variable is set).

I added a new configuration option (`hide_profiles_without_section`) to the module that toggles this new functionality (it should probably be removed in the future but it impacted too many tests for them to be changed before receiving some feedback).

I am looking for feedback regarding the implementation and potential oversights/issues.

#### Motivation and Context

Hides the AWS prompt when it is usually not needed (e.g. when no profiles have been set). Goes a bit further than the linked issue since it also disables the prompt for profiles other than `default` if the set profile has no matching sections in either the credentials or config file (**Note**: this might be an undesirable behavior but it was intentionally implemented this way).

Closes #3457
Closes #2184
Closes #2871
Closes #3504

#### How Has This Been Tested?
I added some tests to the AWS module to reflect the new functionality and I ran `cargo test modules::aws` to ensure no breaking changes were introduced.
- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:
- [ ] I have updated the documentation accordingly. (Not done since I am unsure about the future of the new option)
- [x] I have updated the tests accordingly.

PR #3504, opened 2022-01-24
fix(aws): Only display AWS if there are credentials configured

### LINKED ISSUE #2184: Respect AWS settings and show when appropriate
The default configuration for the AWS module will show up on the prompt even if there is no valid credentials in the ~/.aws/credentials file for the current environment.   

```
~ on ☁️  (us-west-2) took 2s
❯ cat C:\Users\me\.aws\credentials
[personal]
aws_access_key_id = KEY
aws_secret_access_key = SECRET

~ on ☁️  (us-west-2)
❯ cat C:\Users\me\.aws\config
[default]
region = us-west-2
output = json

~ on ☁️  (us-west-2)
❯ gci env:A*

Name                           Value
----                           -----
APPDATA                        C:\Users\me\AppData\Roaming
ALLUSERSPROFILE                C:\ProgramData

~ on ☁️  (us-east-1)
❯ aws sts get-caller-identity

Unable to locate credentials. You can configure credentials by running "aws configure".
```

## Describe the solution you'd like
Following the expected pattern of how AWS CLI/SDK functions I would expect the prompt to do the follow:

When there is no `[default]` section in the ~/.aws/credentials file the prompt should not trigger if no environment variables are set.   

When there is a `AWS_PROFILE` environment variable, then look against the ~/.aws/config & ~/.aws/credentials file for Region/Profile info onto the prompt.   When checking against ~/.aws/config look for the [default] config and then override any [default] settings with [$env:AWS_PROFILE] settings.   

```
~
❯ gci env:A*

Name                           Value
----                           -----
APPDATA                        C:\Users\me\AppData\Roaming
ALLUSERSPROFILE                C:\ProgramData


~
❯ cat C:\Users\ian\.aws\credentials
[personal]
aws_access_key_id = KEY
aws_secret_access_key = SECRET

~
❯ cat C:\Users\ian\.aws\config
[default]
region = us-west-2
output = json

~ 
❯ aws sts get-caller-identity

Unable to locate credentials. You can configure credentials by running "aws configure".

~
❯ $env:AWS_PROFILE="personal"

~ on ☁️  personal(us-west-2)
❯ $env:AWS_REGION="us-east-1"

~ on ☁️  octopus(us-east-1) took 5s
❯ aws sts get-caller-identity
{
    "UserId": "KEY",
    "Account": "ACCOUNT",
    "Arn": "arn:aws:iam::111111111:user/me"
}
```

I don't see a means to use the existing config of AWS module to accomplish this.  I have disabled it at this time.

### LINKED ISSUE #2871: AWS Profile Does not switch
#### Current Behavior
AWS Profile does not get displayed. When using `export AWS_PROFILE=profile_name` it shows the profile but the region disappears. When not using the env variable, only the region is displayed.

#### Expected Behavior
Display the profile name and region

With `AWS_PROFILE` set, region disappears. 
image

Removing the ENV variable, shows the symbol and region but not the current profile [default]
image

#### Environment

- AWS credentials
```
File: /Users/leo/.aws/credentials
[default]
aws_access_key_id = 
aws_secret_access_key = 

[leosada]
aws_access_key_id = 
aws_secret_access_key = 

[leos]
aws_access_key_id = 
aws_secret_access_key = 
```

- Starship version: 
```
starship 0.55.0
branch:
commit_hash:
build_time:2021-06-20 19:48:59
build_env:rustc 1.52.1,
```
- Shell type: `fish`
- Shell version: `fish, version 3.3.0`
- Shell plugin manager: `fisher, omf`
- Terminal emulator: `iTerm 3.4.8`
- Operating system: `macOS 11.4`

```bash
# Your configuration here
```

#### Starship Configuration
```
# Disable the newline at the start of the prompt
add_newline = false

    # $gcloud\
# Overwrite a default_prompt_order and  use custom prompt_order
format = """\
    $hostname\
    $aws\
    $username\
    $python\
    $package\
    $kubernetes\
    $directory\
    $git_branch\
    $git_commit\
    $git_state\
    $git_status\
    $dotnet\
    $golang\
    $java\
    $nodejs\
    $ruby\
    $rust\
    $nix_shell\
    $conda\
    $memory_usage\
    $env_var\
    $cmd_duration\
    $line_break\
    $jobs\
    $battery\
    $time\
    $character\
"""
[line_break]
disabled = false

[aws]
# format = '[$symbol($profile)(\($region\))]($style) '
format = '[$symbol($profile )(\($region\) )(\[$duration\])]($style) '
style = "bold blue"
symbol = "🅰 " # symbol = "AWS"

#add to format in the top:
#    $gcloud\

[gcloud]
format = '[$symbol$account(\($project | $region\))]($style) '
#symbol = "� "
style = "green"

[hostname]
ssh_only = false
format =  "on [$hostname](bold green) "
trim_at = "."
disabled = false

[directory]
truncation_length = 14

# [character]
# symbol = "➜"
# error_symbol = "✗"
# use_symbol_for_status = true

[python]
python_binary = "python3"

[kubernetes]
symbol = "\u2388" 
style = "yellow"
format = '[$symbol $context \($namespace\)]($style) '
disabled = false
[kubernetes.context_aliases]

[git_branch]
truncation_length = 30
truncation_symbol = "..."

[git_status]
ahead = "⇡${count}"
diverged = "⇕⇡${ahead_count}⇣${behind_count}"
staged = "stg: ${count} " 
untracked = "new: ${count} "
behind = "⇣${count} "
deleted = "del: ${count} "
modified = "mod: ${count} "

[terraform]
format = "[$version$workspace]($style) "

[env_var]
variable = "HTTP_PROXY"
format = "[$env_value]($style)"
default = ""

[time]
disabled = true

```

```bash
# Your configuration here
```

### LINKED ISSUE #3457: Only display AWS if there are credentials configured
At work we have custom tooling to manage federating into different AWS accounts that will dynamically add profiles AWS config file and activate them. The tool will default to the region defined in default profile when creating new profiles and to this end my default profile is defined simply as:

```
[default]
region = eu-west-1
```

Specifically this profile has no credentials defined and can't be used directly (i.e. it's just a placeholder). With the AWS plugin however my shell always displays "☁️  (eu-west-1)" (which is understandable)

I would like to only display the AWS data if the profile has valid credentials/`credential_process`  defined.

Make displaying for default profile optional.

alago1:
#### Description

Adds guard, and relevant tests to only display AWS data if either a credential_process is defined in `.aws/config` or there are valid credentials for the same profile in `.aws/credentials`.

#### Motivation and Context

Closes #3457
Closes #2184
Closes #2871
Closes #2461

I've tested these changes on WSL2 (Ubuntu 18.04 and 20.04)

#### Checklist:

I'm not sure if this change requires the documentation to be updated.
If so, please let me know where I should change it.

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -126,7 +126,7 @@ fn get_credential_process(context: &Context, aws_profile: Option<&Profile>) -> O
     let contents = read_file(get_config_file_path(context)?).ok()?;
 
     let profile_line = if let Some(aws_profile) = aws_profile {
-        format!("[{}]", aws_profile)
+        format!("[profile {}]", aws_profile)
     } else {
         "[default]".to_string()
     };
@@ -170,7 +170,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    // only display if profile has valid credentials or credential_process is defined
+    // only display if credential_process is defined or has valid credentials
     if get_credential_process(context, aws_profile.as_ref()).is_none()
         && get_defined_credentials(context, aws_profile.as_ref()).is_none()
     {
@@ -243,6 +243,7 @@ mod tests {
     fn region_set() {
         let actual = ModuleRenderer::new("aws")
             .env("AWS_REGION", "ap-northeast-2")
+            .env("credential_process", "/opt/bin/awscreds-retriever")
             .collect();
         let expected = Some(format!(
             "on {}",
@@ -256,6 +257,7 @@ mod tests {
     fn region_set_with_alias() {
         let actual = ModuleRenderer::new("aws")
             .env("AWS_REGION", "ap-southeast-2")
+            .env("credential_process", "/opt/bin/awscreds-retriever")
             .config(toml::toml! {
                 [aws.region_aliases]
                 ap-southeast-2 = "au"
@@ -271,6 +273,7 @@ mod tests {
         let actual = ModuleRenderer::new("aws")
             .env("AWS_REGION", "ap-northeast-2")
             .env("AWS_DEFAULT_REGION", "ap-northeast-1")
+            .env("credential_process", "/opt/bin/awscreds-retriever")
             .collect();
         let expected = Some(format!(
             "on {}",
@@ -284,7 +287,7 @@ mod tests {
     fn profile_set() {
         let actual = ModuleRenderer::new("aws")
             .env("AWS_PROFILE", "astronauts")
-            .env("credential_process", "/opt/bin/awscreds-custom")
+            .env("credential_process", "/opt/bin/awscreds-retriever")
             .collect();
         let expected = Some(format!(
             "on {}",
@@ -299,7 +302,7 @@ mod tests {
         let actual = ModuleRenderer::new("aws")
             .env("AWS_VAULT", "astronauts-vault")
             .env("AWS_PROFILE", "astronauts-profile")
-            .env("credential_process", "/opt/bin/awscreds-custom")
+            .env("credential_process", "/opt/bin/awscreds-retriever")
             .collect();
         let expected = Some(format!(
             "on {}",
@@ -314,7 +317,7 @@ mod tests {
         let actual = ModuleRenderer::new("aws")
             .env("AWSU_PROFILE", "astronauts-awsu")
             .env("AWS_PROFILE", "astronauts-profile")
-            .env("credential_process", "/opt/bin/awscreds-custom")
+            .env("credential_process", "/opt/bin/awscreds-retriever")
             .collect();
         let expected = Some(format!(
             "on {}",
@@ -329,7 +332,7 @@ mod tests {
         let actual = ModuleRenderer::new("aws")
             .env("AWSUME_PROFILE", "astronauts-awsume")
             .env("AWS_PROFILE", "astronauts-profile")
-            .env("credential_process", "/opt/bin/awscreds-custom")
+            .env("credential_process", "/opt/bin/awscreds-retriever")
             .collect();
         let expected = Some(format!(
             "on {}",
@@ -344,7 +347,7 @@ mod tests {
         let actual = ModuleRenderer::new("aws")
             .env("AWS_PROFILE", "astronauts")
             .env("AWS_REGION", "ap-northeast-2")
-            .env("credential_process", "/opt/bin/awscreds-custom")
+            .env("credential_process", "/opt/bin/awscreds-retriever")
             .collect();
         let expected = Some(format!(
             "on {}",
@@ -367,7 +370,6 @@ mod tests {
                 "AWS_CREDENTIALS_FILE",
                 config_path.to_string_lossy().as_ref(),
             )
-            .env("credential_process", "/opt/bin/awscreds-custom")
             .collect()
             .is_none());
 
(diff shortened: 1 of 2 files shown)


chipbuster:
> I'm not sure if this change requires the documentation to be updated.
> If so, please let me know where I should change it.

I believe we'll need to update the rules on when the module is displayed in the config README.md. You can see the section [here](https://github.com/starship/starship/blob/master/docs/config/README.md?plain=1#L270).

src/modules/aws.rs
```diff
@@ -118,6 +118,49 @@ fn alias_region(region: String, aliases: &HashMap<String, &str>) -> String {
     }
 }
 
+fn get_credential_process(context: &Context, aws_profile: Option<&Profile>) -> Option<String> {
+    if let Some(cred_proc) = context.get_env("credential_process") {
```
andytom, line 122: From the [AWS docs](https://docs.aws.amazon.com/sdkref/latest/guide/setting-global-credential_process.html) it doesn't look like you can set this can be read from an env var.

src/modules/aws.rs
```diff
@@ -127,6 +170,13 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
+    // only display if credential_process is defined or has valid credentials
```
andytom, line 173: Should we also support the [environment variable](https://docs.aws.amazon.com/sdkref/latest/guide/setting-global-credential_process.html) e.g `AWS_SECRET_ACCESS_KEY` and `AWS_SESSION_TOKEN`?

src/modules/aws.rs
alago1: I added this option because almost all tests in the module would have to be rewritten in order to create the temporary config file. I thought that option would be quite clunky as opposed to just checking for an env var. I also couldn't find an example of another program who uses a `credential_process` env var that could conflict with this.
With that being said, I'd appreciate some advice in this case. Should I stick with the env var even if it is not technically supported for the sake making the tests more readable and concise or should I rewrite all of them? Is there another way around this?
Thanks in advance.

alago1:
Hey, sorry but didn't have a chance to get back onto this since the latest review.
I added the support for `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` as requested and left a comment for the other note. I'd appreciate any feedback on it!

src/modules/aws.rs
andytom: Since you have also added support for the `AWS_...` env vars you could use that as an alternative.

andytom (approved):

andytom:
Thank you for your contribution @alago1.

quaintops:
I had a similar concern whereby I only expect aws context to be presented on my prompt **while my aws session is authenticated and active**, so I created a lightweight service for this purpose.
Check it out here: https://github.com/quaintops/starship-aws-monitor

The built-in Starship `[aws]` module shows AWS info even when credentials are expired. This service solves that by:

- Running as a background systemd service (near-zero prompt latency)
- Checking authentication via STS `GetCallerIdentity` every 60 seconds
- Writing status to `~/.aws/auth-status` for Starship to read
- Only displaying AWS info when actually authenticated

Currently supports systemd only; macOS launchd plist support can be added easily. So far tested mainly with AWS SSO and bash on Debian & Ubuntu.

PR #3510, opened 2022-01-25
docs(i18n): add a little Chinese translation

AtomAlpaca:
#### Description
add a little Chinese translation to docs/zh-CN/README.md

#### Motivation and Context
I noticed that a few of the English instructions in the doc above were not translated into Chinese.

### DIFF
--- a/docs/zh-CN/README.md
+++ b/docs/zh-CN/README.md
@@ -30,7 +30,7 @@ description: Starship is the minimal, blazing fast, and extremely customizable p
 
 ### 前置要求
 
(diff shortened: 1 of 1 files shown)


chipbuster:
Hi @AtomAlpaca, thanks for the contributions!

We don't directly merge translations via GitHub. Instead, translations are managed on [Crowdin](https://crowdin.com/project/starship-prompt) and then merged in so that we don't run into conflicts.

I'd love to have your contributions over on Crowdin. Would you mind making an account and adding these translations there?

PR #3515, opened 2022-01-26
feat(cmd_duration): Make notification timeout configurable

### LINKED ISSUE #2194: Increase cmd_duration notification duration

Lyndeno:
The default hard-coded notification timeout is very brief. This PR allows it to be configured.

#### Description

This PR adds the notification_timeout option to cmd_duration. This allows for the notification timeout to be configured by the user if starship is compiled with notify-rust support. The default value is equal to the previous hardcoded value.

The notification_timeout is coded as a u64, but the notify-rust TImeout::Milliseconds is coded to take u32, so the u64 is cast down. If there is a better way to handle this I am open to suggestions.

#### Motivation and Context
The hardcoded notification timeout was brief and lacked customization. Making this option configurable will allow the notification feature to be more useful.

Closes #2194
Closes #2268

Make sure starship is compiled with ```--features "notify-rust"```
Run ```STARSHIP_LOG=debug /path/to/starship module cmd_duration -d 60000``` to simulate a long-running program. Test it with and without the option specified in the configuration file. Observe the amount of time it takes for the notification to dismiss automatically.

This was tested on NixOS 21.11 using the mako notification daemon in the Sway window manager. Tested with bash and zsh.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -597,16 +597,16 @@ running `eval $(starship init $0)`, and then proceed as normal.
 
 ### Options
 
-| Option                 | Default                       | Description                                                                                          |
-| ---------------------- | ----------------------------- | ---------------------------------------------------------------------------------------------------- |
-| `min_time`             | `2_000`                       | Shortest duration to show time for (in milliseconds).                                                |
-| `show_milliseconds`    | `false`                       | Show milliseconds in addition to seconds for the duration.                                           |
-| `format`               | `"took [$duration]($style) "` | The format for the module.                                                                           |
-| `style`                | `"bold yellow"`               | The style for the module.                                                                            |
-| `disabled`             | `false`                       | Disables the `cmd_duration` module.                                                                  |
-| `show_notifications`   | `false`                       | Show desktop notifications when command completes.                                                   |
-| `min_time_to_notify`   | `45_000`                      | Shortest duration for notification (in milliseconds).                                                |
-| `notification_timeout` | `750`                         | Duration to show notification for (in milliseconds). Not all notification daemons honor this option. |
+| Option                 | Default                       | Description                                                                                                                                                       |
+| ---------------------- | ----------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `min_time`             | `2_000`                       | Shortest duration to show time for (in milliseconds).                                                                                                             |
+| `show_milliseconds`    | `false`                       | Show milliseconds in addition to seconds for the duration.                                                                                                        |
+| `format`               | `"took [$duration]($style) "` | The format for the module.                                                                                                                                        |
+| `style`                | `"bold yellow"`               | The style for the module.                                                                                                                                         |
+| `disabled`             | `false`                       | Disables the `cmd_duration` module.                                                                                                                               |
+| `show_notifications`   | `false`                       | Show desktop notifications when command completes.                                                                                                                |
+| `min_time_to_notify`   | `45_000`                      | Shortest duration for notification (in milliseconds).                                                                                                             |
+| `notification_timeout` | Unset                         | Duration to show notification for (in milliseconds). If unset, notification timeout will be determined by daemon. Not all notification daemons honor this option. |
 
 ::: tip
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -606,7 +606,7 @@ running `eval $(starship init $0)`, and then proceed as normal.
 | `disabled`             | `false`                       | Disables the `cmd_duration` module.                                                                                                                               |
 | `show_notifications`   | `false`                       | Show desktop notifications when command completes.                                                                                                                |
 | `min_time_to_notify`   | `45_000`                      | Shortest duration for notification (in milliseconds).                                                                                                             |
-| `notification_timeout` | Unset                         | Duration to show notification for (in milliseconds). If unset, notification timeout will be determined by daemon. Not all notification daemons honor this option. |
+| `notification_timeout` |                               | Duration to show notification for (in milliseconds). If unset, notification timeout will be determined by daemon. Not all notification daemons honor this option. |
 
 ::: tip
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -597,16 +597,16 @@ running `eval $(starship init $0)`, and then proceed as normal.
 
 ### Options
 
-| Option                 | Default                       | Description                                                |
-| ---------------------- | ----------------------------- | ---------------------------------------------------------- |
-| `min_time`             | `2_000`                       | Shortest duration to show time for (in milliseconds).      |
-| `show_milliseconds`    | `false`                       | Show milliseconds in addition to seconds for the duration. |
-| `format`               | `"took [$duration]($style) "` | The format for the module.                                 |
-| `style`                | `"bold yellow"`               | The style for the module.                                  |
-| `disabled`             | `false`                       | Disables the `cmd_duration` module.                        |
-| `show_notifications`   | `false`                       | Show desktop notifications when command completes.         |
-| `min_time_to_notify`   | `45_000`                      | Shortest duration for notification (in milliseconds).      |
-| `notification_timeout` | `750`                         | Duration to show notification for (in milliseconds).       |
+| Option                 | Default                       | Description                                                                                          |
+| ---------------------- | ----------------------------- | ---------------------------------------------------------------------------------------------------- |
+| `min_time`             | `2_000`                       | Shortest duration to show time for (in milliseconds).                                                |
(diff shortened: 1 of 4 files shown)


src/configs/cmd_duration.rs
```diff
@@ -12,6 +12,7 @@ pub struct CmdDurationConfig<'a> {
     pub disabled: bool,
     pub show_notifications: bool,
     pub min_time_to_notify: i64,
+    pub notification_timeout: u64,
```
andytom, line 15: Maybe I am missing something but can this not just be a `u32`? Then we don't need to cast it?

Lyndeno: I tried building as u32 but then got this error:
```
   Compiling starship v1.2.1 (/home/lsanche/Projects/starship)
error[E0599]: no method named `load_config` found for type `u32` in the current scope
  --> src/configs/cmd_duration.rs:6:17
   |
6  | #[derive(Clone, ModuleConfig, Serialize)]
   |                 ^^^^^^^^^^^^ method not found in `u32`
   |
   = help: items from traits can only be used if the trait is implemented and in scope
note: `ModuleConfig` defines an item `load_config`, perhaps you need to implement it
  --> src/config.rs:41:1
   |
41 | / pub trait ModuleConfig<'a>
42 | | where
43 | |     Self: Sized + Clone,
44 | | {
...  |
55 | |     }
56 | | }
   | |_^
   = note: this error originates in the derive macro `ModuleConfig` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0599`.
error: could not compile `starship` due to previous error
```

So I came to the assumption that module options weren't designed with u32 in mind. But I did not look very hard to see if this is indeed the case.

Lyndeno:
I have added checks to make sure configured timeout is in bounds for u32. The default timeout will be cast to u32 if configured value if not valid. Inspired by the conversation in #2268.

Also added a test to make sure the coded default is small enough to be cast to u32.

src/configs/cmd_duration.rs
chipbuster: Without looking further into the code, my guess is that this is because TOML requires that implementations can "losslessly" handle signed 64-bit integers, so the deserialization code will not not allow us to deserialize integers into u32s due to possible overflow. Dunno why it allows us to do u64s though...

src/modules/cmd_duration.rs
```diff
@@ -79,12 +79,24 @@ fn undistract_me<'a, 'b>(
             unstyle(&ANSIStrings(&module.ansi_strings()))
         );
 
+        let timeout: u32 = match u32::try_from(config.notification_timeout) {
```
chipbuster, line 82: I haven't tested the code, but I believe this can be written as something like the following:

```
let timeout = u32::try_from(config.notification_timeout).unwrap_or_else(|| {
  log::trace!("Configured notification timeout not within bounds. Defaulting to {} ms", v);
  CmdDurationConfig::default().notification_timeout as u32
})
```

which would be stylistically more similar to the rest of the codebase.

chipbuster (commented):
I'm not sure what the best way to inform the user of this is, but we should note somewhere that this timeout is only advisory and doesn't have to be respected by the notification server. The `notify-rust` docs [say as much themselves](https://docs.rs/notify-rust/latest/notify_rust/struct.Notification.html#structfield.timeout), with this being the docstring for the timeout field:

> Lifetime of the Notification in ms. Often not respected by server, sorry.

Where do you think the best place to put that information would be?

src/modules/cmd_duration.rs
Lyndeno: Thanks, will give this a try after work.

src/configs/cmd_duration.rs
andytom: I think we would need to implement the ModuleConfig trait for u32 similar to [u64 here](https://github.com/starship/starship/blob/master/src/config.rs#L83).

Lyndeno: > I think we would need to implement the ModuleConfig trait for u32 similar to [u64 here](https://github.com/starship/starship/blob/master/src/config.rs#L89). 

I can work on this later. Should I open a separate PR for this?

Lyndeno:
> Where do you think the best place to put that information would be?

I think it would be good to put it in the config README. Right beside the option description.

src/configs/cmd_duration.rs
Lyndeno: Made a commit that implements ModuleConfig for u32, removes the need to check value inside of cmd_duration.rs.

src/modules/cmd_duration.rs
Lyndeno: Latest commits add ModuleConfig for u32, which handles the checking and assigning of the default value instead of logic inside of cmd_duration.rs

Lyndeno:
I have implemented ModuleConfig for u32. This removes the need to check the given value inside of cmd_duration.rs. ModuleConfig will assign the default value in the even an invalid value is given in the config.

Lyndeno:
If this looks good I do not have anything to add and am ready for merge.

src/configs/cmd_duration.rs
```diff
@@ -24,6 +25,7 @@ impl<'a> Default for CmdDurationConfig<'a> {
             disabled: false,
             show_notifications: false,
             min_time_to_notify: 45_000,
+            notification_timeout: 750,
```
chipbuster, line 28: As long as we're here, we don't have to keep the default notification length.

You noted in the original PR that 750ms is too short--what's the minimum value that you think is long enough to be usable?

chipbuster (commented):
I would like to grump about the fact that conversion from TOML can silently fail, resulting in fallback to a default value, without any information appearing in the logs.

However, this design flaw is present in our other `ModuleConfig` impls as well, so it's not on @Lyndeno to fix. And this should only be an issue if someone wants their notifications to last negative time or over a month, so it's not likely to be hit in practice.

One question about the default: is there a value that you think is a better default? IIRC we were a little nervous about adding this feature originally so we kept the duration short to keep things out-of-the-way. Now that it's been proven that this is a desirable feature, maybe we should configure a more pleasant default.

src/configs/cmd_duration.rs
chipbuster: And if you're happy with this or don't have a suggestion, let me know and I'll approve.

davidkna: Another option might be to change this to `Option<u32>` (if so, don't forget adding ` #[serde(skip_serializing_if = "Option::is_none")]` to the struct item) and map `None` to `Timeout::Default`.

Lyndeno: A default of 2 seconds seems reasonable.

Using an Option to make the default Timeout:: Default is something I haven't looked at. I like the idea and it is worth considering. I'll see what I can implement.

Lyndeno:
The latest commit doesn't have a default value, but instead defaults to letting the notification daemon decide if a value is not provided, or invalid.

docs/config/README.md
```diff
@@ -597,15 +597,16 @@ running `eval $(starship init $0)`, and then proceed as normal.
...
+| `style`                | `"bold yellow"`               | The style for the module.                                                                                                                                         |
+| `disabled`             | `false`                       | Disables the `cmd_duration` module.                                                                                                                               |
+| `show_notifications`   | `false`                       | Show desktop notifications when command completes.                                                                                                                |
+| `min_time_to_notify`   | `45_000`                      | Shortest duration for notification (in milliseconds).                                                                                                             |
+| `notification_timeout` | Unset                         | Duration to show notification for (in milliseconds). If unset, notification timeout will be determined by daemon. Not all notification daemons honor this option. |
```
chipbuster, line 609:
```suggestion
| `notification_timeout` |                          | Duration to show notification for (in milliseconds). If unset (default), duration will be determined by daemon. Not all notification daemons honor this option. |
```

chipbuster (commented):
One minor doc change and I think we'll be good.

docs/config/README.md
chipbuster: I don't particularly like the use of Unset here because it's not clear if that should be the bareword, or if we forgot to quote it, or what. If we start doing this in lots of other places we might be able to add it, but for now I think it's safer if we just leave that spot in the table blank and rely on the description to explain what's going on.

Lyndeno: Done

chipbuster (approved):
:+1: LGTM.

andytom:
Thank you for your contribution @Lyndeno and thanks for reviewing @chipbuster.

Lyndeno:
Thank you very much everyone for taking the time to review!

PR #3529, opened 2022-01-27
feat: starship bug-report sets syntax highlighting for config file

CLIDragon:
#### Description
Set the syntax highlighting for the config file in starship bug-report, using the triple backtick syntax.

#### Motivation and Context
It seemed unintuitive that the syntax highlighting for the config file was set to bash even when it was known that it was in another language, such as powershell or lua.

I have confirmed that this works for powershell, and all tests that pass on master pass on this branch. 

#### Checklist:

### DIFF
--- a/src/bug_report.rs
+++ b/src/bug_report.rs
@@ -53,7 +53,7 @@ fn get_pkg_branch_tag() -> &'static str {
 
 fn make_github_issue_link(environment: Environment) -> String {
     let shell_syntax = match environment.shell_info.name.as_ref() {
-        "powershell" => "pwsh",
+        "powershell" | "pwsh" => "pwsh",
         "fish" => "fish",
         "cmd" => "lua",
         // GitHub does not seem to support elvish syntax highlighting.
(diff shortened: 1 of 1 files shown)


src/bug_report.rs
```diff
@@ -52,6 +52,15 @@ fn get_pkg_branch_tag() -> &'static str {
 }
 
 fn make_github_issue_link(environment: Environment) -> String {
+    let shell_syntax = match environment.shell_info.name.as_ref() {
+        "powershell" => "pwsh",
```
chipbuster, line 56: This does not work for powershell core (or...whatever it's called? I'm not very familiar with the renamings around Core) because the binary name is "pwsh"
```suggestion
        "powershell" | "pwsh" => "pwsh",
```

chipbuster: One minor change to account for a recent PR which added a new name for pwoershell core.

CLIDragon: Done.

matchai (approved):
LGTM 👍
Thank you for the fix-up, Joseph! 😄

PR #3533, opened 2022-01-28
fix(nodejs): use e718 as the default of symbol in node configuration

### LINKED ISSUE #2546: New nodejs character positioned incorrectly
#### Current Behavior
The prompt for nodejs directories uses a new nerd-font character that does not render correctly. It’s positioned incorrectly and results in a collision with neighboring characters.

#### Expected Behavior
The nodejs character should not collide with neighboring characters.

Screen Shot 2021-04-02 at 15 14 33

#### Possible Solution
Revert #1649.

#### Environment
- Starship version: starship 0.51.0
- Shell type: zsh
- Shell version: zsh 5.8 (x86_64-apple-darwin20.0)
- Shell plugin manager: _none_
- Terminal emulator: Terminal
- Operating system: macOS 11.2.3

```zsh
# Bash style word boundaries (including / and -)
autoload -U select-word-style
select-word-style bash

# Git completions
autoload -Uz compinit && compinit

# rbenv (Ruby version manager)
# [Disabled for performance reasons]
# eval "$(rbenv init -)"

# nvm (node version manager)
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"  # This loads nvm
[ -s "$NVM_DIR/bash_completion" ] && . "$NVM_DIR/bash_completion"  # This loads nvm bash_completion

export PATH="/usr/local/sbin:$PATH"

eval "$(starship init zsh)"
```

#### Starship Configuration

```bash
add_newline = false

# Replace the "❯" symbol in the prompt
[character]
success_symbol = "[❯]()"

# Disable the package module, hiding it from the prompt completely
# [package]
# disabled = true
```

TrickyPi:
#### Description

Use e718 as the default of symbol in node configuration
#### Motivation and Context

Closes #2546 

img

Visually checked on Iterm2, hyper and mac terminal.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2168,7 +2168,7 @@ By default the module will be shown if any of the following conditions are met:
 | ------------------- | ------------------------------------ | ----------------------------------------------------------------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"` | The format for the module.                                                                            |
 | `version_format`    | `"v${raw}"`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch`                             |
-| `symbol`            | `" "`                               | A format string representing the symbol of Node.js.                                                   |
+| `symbol`            | `" "`                               | A format string representing the symbol of Node.js.                                                   |
 | `detect_extensions` | `["js", "mjs", "cjs", "ts"]`         | Which extensions should trigger this module.                                                          |
 | `detect_files`      | `["package.json", ".node-version"]`  | Which filenames should trigger this module.                                                           |
 | `detect_folders`    | `["node_modules"]`                   | Which folders should trigger this module.                                                             |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,9 +8,9 @@ If you have any questions that aren't addressed in this document, please don't h
 
 ## Glossary
 
-- **Module**: A component in the prompt giving information based on contextual information from your OS. For example, the `nodejs` module shows the version of NodeJS that is currently installed on your computer, if your current directory is a NodeJS project.
+- **Module**: A component in the prompt giving information based on contextual information from your OS. For example, the `rust` module shows the version of Rust that is currently installed on your computer, if your current directory is a Rust project.
 
-- **Segment**: Smaller sub-components that compose a module. For example, the `symbol` segment in the `nodejs` module contains the character that is shown before the version number (`⬢` by default).
+- **Segment**: Smaller sub-components that compose a module. For example, the `symbol` segment in the `rust` module contains the character that is shown before the version number (`🦀` by default).
 
 ## Philosophy
 
--- a/docs/presets/README.md
+++ b/docs/presets/README.md
@@ -62,6 +62,9 @@ symbol = " "
 [nix_shell]
 symbol = " "
 
+[nodejs]
+symbol = " "
+
 [package]
 symbol = " "
 
--- a/src/configs/nodejs.rs
+++ b/src/configs/nodejs.rs
@@ -21,7 +21,7 @@ impl<'a> Default for NodejsConfig<'a> {
         NodejsConfig {
             format: "via [$symbol($version )]($style)",
             version_format: "v${raw}",
-            symbol: " ",
+            symbol: " ",
             style: "bold green",
             disabled: false,
             not_capable_style: "bold red",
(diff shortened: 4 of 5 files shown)


chipbuster:
Tested on MacOS under the following terminals:
  - Terminal
  - iTerm2
  - Hyper
  - Alacritty
  - Kitty

Works on all of them.

chipbuster:
I would appreciate it if someone could test this on Windows, since my current setups are based on virtualized systems which do not support OpenGL. At the least, I'd like to see Windows Terminal, one of the OpenGL terminals (Alacritty, WezTerm, Kitty), and, if possible, the standard conhost.

TrickyPi:
Tested on windows under Alacritty terminal.
before changing:
before
after changing:
after

davidkna:
Windows Terminal, Alacritty and WezTerm seem fine. conhost didn't list my nerd fonts as a font option and was thus broken.

chipbuster:
Okay, in that case, all we need to do is make the appropriate edit in `docs/config` and `docs/presets` and we should be good to merge.

docs/config/README.md
```diff
@@ -2168,7 +2168,7 @@ By default the module will be shown if any of the following conditions are met:
 | ------------------- | ------------------------------------ | ----------------------------------------------------------------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"` | The format for the module.                                                                            |
 | `version_format`    | `"v${raw}"`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch`                             |
-| `symbol`            | `" "`                               | A format string representing the symbol of Node.js.                                                   |
+| `symbol`            | `" "`                               | A format string representing the symbol of Node.js.                                                   |
```
chipbuster, line 2171: Oh. Oh boy.

image

It seems that the new symbol is part of the NerdFonts CJK overlap issue. See https://github.com/ryanoasis/nerd-fonts/issues/365 for details. In short, several NerdFont symbols bled out of the PUA into the BMP and can override/be overridden by symbols from other human languages.

This is strictly speaking a NerdFonts issue, but I think we should be prepared for some issues about this symbol not appearing correctly on systems with CJK support.

chipbuster (commented):
Unlucky for us, the new symbol is part of the NerdFonts codepoint leakage reported in 2019. I don't have a feel for how close they are to fixing that issue, and I don't think it should be a blocker, but we should make a mental note that this is a potential problem for anyone who uses a language that uses CJK codepoints.

docs/config/README.md
TrickyPi: I wonder if there are other symbols like this new symbol in the default configuration. Maybe we can list them in the docs.

chipbuster: Quite a few, as it turns out. From this system, in the NerdFont preset, it looks like AWS, golang, julia, perl, ruby, rust, and scala are affected. There may be more in other parts of the default config.

TrickyPi: Ok～

I have a confusion, the new nodejs symbol is e718, it belong the PUA area (U+E000..U+F8FF).

TrickyPi: it should be displayed correctly.

chipbuster: Oh.

Hm, I wonder if this system is actually misconfigured then? I'm using Pop_OS's default fontconfigs, which I assumed should be correct, but maybe those are broken...

chipbuster (approved):
Some confusion over whether my browser or Nerd Fonts is broken at the moment, but either way I don't think it's a big enough reason to hold this PR back. 

If it turns out that this is, in fact, a codepoint leakage, we'll probably need to add an FAQ explaining what's going on.

andytom:
Thank you for your contribution @TrickyPi and thank you for the indepth testing @chipbuster and @davidkna.

docs/config/README.md
davidkna: It doesn't render for me on Windows in the browser, and I have some CJK fonts installed.

TrickyPi: Which browser are you using? It seems to me that it should render correctly because the browser loads the nerd-font.woff2 font.

TrickyPi: > It doesn't render for me on Windows in the browser, and I have some CJK fonts installed.

Could you check the nerd-font.woff2 whether has been loaded?
image
I ran into some unexpected issues when the nerd-font.woff2 has not been loaded.
image
image

PR #3534, opened 2022-01-28
fix(elvish): use `$pwd` for `logical-path`

### LINKED ISSUE #3512: Directory is not correctly displayed when starting WezTerm with `--cwd` flag and `elvish` shell
#### Current Behavior

When running `wezterm start --cwd <path>`, the terminal is correctly started at `<path>`, but the directory module will display the PWD of the launching terminal no matter how many commands are entered, until a `cd` command is issued.

This behavior only occurs with the trifecta of starship, elvish, and wezterm, leading me to believe that the behavior is linked to something in the that we're not accounting for in the init script, but I don't have time to explore further at the moment.

#### Expected Behavior
Terminal launches with correct PWD.

image

Doing any of the following resolves this bug:

- Using `zsh` instead of `elvish` as the `default_prog` for `wezterm`
- Using the default prompt instead of starship in elvish
- Using `konsole` instead of `wezterm` as the terminal emulator

#### Possible Solution
....yeah, this is a weird one.

#### Environment
- Starship version: built from 09656678
- Shell type: 
  + elvish 0.18.0-dev.fda62502ad7f33222674a97602822114c4fd1037
  + zsh zsh 5.8 (x86_64-pc-linux-gnu)
- Shell plugin manager: None
- Terminal emulator:
  + wezterm 20220101-133340-7edc5b5a
  + konsole 21.12.1
- Operating system: Arch

```
# Your configuration here
```

#### Starship Configuration

```toml
scan_timeout = 100
add_newline = false

continuation_prompt = "▶▶"
# right_format="$time"

[git_status]
disabled = true

[time]
disabled = false
format = '[$time](purple) '

[directory]
use_logical_path = true
fish_style_pwd_dir_length = 1

[python]
disabled = false
python_binary = "python3"

[shell]
disabled = false
elvish_indicator = "𐌴"
fish_indicator = "𐍆"
bash_indicator = "𐌱"
zsh_indicator = "𐌶"
xonsh_indicator = "𐍇"
powershell_indicator = "𐍀"
ion_indicator = "𐌹"
nu_indicator = "𐌽"
```

davidkna:
#### Description

The `$PWD` environment variable used by starship is not updated on startup by `elvish`. `elvish` also provides a separate `$pwd` variable that does get updated and seems to be be a better indicator for the current logical path.

#### Motivation and Context

Closes #3512

#### Checklist:

### DIFF
--- a/src/init/starship.elv
+++ b/src/init/starship.elv
@@ -24,10 +24,10 @@ set edit:after-command = [ $@edit:after-command $starship-after-command-hook~ ]
 # Install starship
 set edit:prompt = {
     var cmd-duration = (printf "%.0f" (* $edit:command-duration 1000))
-    ::STARSHIP:: prompt --jobs=$num-bg-jobs --cmd-duration=$cmd-duration --status $cmd-status-code
+    ::STARSHIP:: prompt --jobs=$num-bg-jobs --cmd-duration=$cmd-duration --status=$cmd-status-code --logical-path=$pwd
 }
 
 set edit:rprompt = {
     var cmd-duration = (printf "%.0f" (* $edit:command-duration 1000))
-    ::STARSHIP:: prompt --right --jobs=$num-bg-jobs --cmd-duration=$cmd-duration --status $cmd-status-code
+    ::STARSHIP:: prompt --right --jobs=$num-bg-jobs --cmd-duration=$cmd-duration --status=$cmd-status-code --logical-path=$pwd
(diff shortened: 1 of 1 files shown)


chipbuster (approved):
:+1: Thanks for digging this up!

andytom:
Thank you for the fix @davidkna and thanks for reviewing @chipbuster.

PR #3543, opened 2022-01-31
fix: Correctly detect older versions of powershell in bug-report

### LINKED ISSUE #3531: Starship bug-report doesn't detect powershell version correctly
#### Current Behavior
Version is set to \<unknown version>.

This appears to be caused because starship attempts to call `STARSHIP_SHELL --version` to determine the current version. However,  Powershell before version 7.0 doesn't support this - you must use `Get-Host | Select-Object Version` instead. ~While Powershell >7.0 does support `--version`, it uses a different executable name to be compatible with earlier versions on the same system - `pwsh`.~

~As `STARSHIP_SHELL` evaluates to `powershell` when using `pwsh` (at least on my system), the current shell detection may have to be modified to set `STARSHIP_SHELL` to `pwsh` when appropriate.~ This has already been implemented in #3443.

#### Expected Behavior
Version is set to the correct version - in my case, `7.2.1`.

#### Possible Solution
Special case the version detection for when `STARSHIP_SHELL` is `powershell`, and pass `Get-Host | Select-Object Version` instead of `--version`. ~As dicussed above, this will return incorrect results when the user is actually using `pwsh`, so this solution will only work if detection for whether the shell is `pwsh` or `powershell` is improved.~

#### Environment
- Starship version: 1.2.1
- Git Commit Hash: 00dce0ca
- Branch/Tag: pwsh-format
- Rust channel: stable-x86_64-pc-windows-msvc debug
- Build Time: 2022-01-27 20:52:18 +10:00

```pwsh
```

#### Starship Configuration

```toml
```

chipbuster:
#### Description
Will lead to correct detection of older versions of powershell using starship bug-report. Special-cases the version code on `powershell` (can be extended to any shell, but all the others support `--version` for now).

Has been tested in a Windows VM against powershell 5.1 and the most recent version of pwsh available on the store.

#### Motivation and Context

Closes #3531

#### Checklist:

### DIFF
--- a/src/bug_report.rs
+++ b/src/bug_report.rs
@@ -142,10 +142,7 @@ fn get_shell_info() -> ShellInfo {
 
     let shell = shell.unwrap();
 
-    let version = exec_cmd(&shell, &["--version"], Duration::from_millis(500)).map_or_else(
-        || UNKNOWN_VERSION.to_string(),
-        |output| output.stdout.trim().to_string(),
-    );
+    let version = get_shell_version(&shell);
 
     let config = get_config_path(&shell)
         .and_then(|config_path| fs::read_to_string(config_path).ok())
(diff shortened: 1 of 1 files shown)


chipbuster:
An interesting point I discovered while testing this: `bug-report` currently assumes that you've already run `starship init` in order to get its shell info. Given that we're not in a performance-constrained path, it might be an interesting project to use some sort of process information (e.g. chase the implicit parent-process-linked-list) to try to find shell information.

davidkna (approved):

andytom:
Thank you for your contribution @chipbuster and thank you for the review @davidkna .

PR #3544, opened 2022-01-31
feat!: Use pure symbols, emoji and Powerline glyphs only by default

### LINKED ISSUE #2563: Make Nerd Fonts icons the default
Since Starship already has Nerd Fonts listed as a prerequisite and uses some Nerd Font glyphs by default, why are some modules using a less descriptive emoji as the default setting (eg. Python using snake instead of Python logo)? I suppose there might be a reason for this, since there is even a [community preset](https://starship.rs/presets/#nerd-font-symbols) fixing this, but I don't understand why this is not the default since it seems rather sane (given that Starship already depends on Nerd Fonts).

segevfiner:
#### Description

Use pure symbols, emoji and Powerline glyphs only by default. This is due to various issues with the Nerd Fonts and to be consistent as we have a Nerd Fonts preset but somehow still use a few Nerd Fonts symbols by default. See #2563 for more details.

Existing Nerd Fonts symbols used by default are moved to the Nerd Fonts preset.

After this starship should only require a powerline font, which are readily available. And even included by default in some coding fonts or some coding fonts having official variants with them. This doesn't include website updates to reflect this. And there might be places in the documentation that should also be updated to reflect this.

The exact symbol choices are, of course, subjective and up for debate. And we do need to make sure we aren't using a symbol that isn't commonly available in fonts.

#### Motivation and Context

Closes #2563

#### Checklist:

- [x] I have updated the documentation accordingly. (Is there any installation docs that need updating to state a Nerd Font isn't required? If so, what should the text say or recommended to install?)

@chipbuster

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -359,7 +359,7 @@ The `azure` module shows the current Azure Subscription. This is based on showin
 | Variable   | Default                                  | Description                                |
 | ---------- | ---------------------------------------- | ------------------------------------------ |
 | `format`   | `"on [$symbol($subscription)]($style) "` | The format for the Azure module to render. |
-| `symbol`   | `"ﴃ "`                                   | The symbol used in the format.             |
+| `symbol`   | `"🅰 "`                                   | The symbol used in the format.             |
 | `style`    | `"blue bold"`                            | The style used in the format.              |
 | `disabled` | `true`                                   | Disables the `azure` module.               |
 
@@ -384,11 +384,11 @@ The module is only visible when the device's battery is below 10%.
 
 | Option               | Default                           | Description                                         |
 | -------------------- | --------------------------------- | --------------------------------------------------- |
-| `full_symbol`        | `" "`                            | The symbol shown when the battery is full.          |
-| `charging_symbol`    | `" "`                            | The symbol shown when the battery is charging.      |
-| `discharging_symbol` | `" "`                            | The symbol shown when the battery is discharging.   |
-| `unknown_symbol`     | `" "`                            | The symbol shown when the battery state is unknown. |
-| `empty_symbol`       | `" "`                            | The symbol shown when the battery state is empty.   |
+| `full_symbol`        | `"• "`                            | The symbol shown when the battery is full.          |
+| `charging_symbol`    | `"⇡ "`                            | The symbol shown when the battery is charging.      |
+| `discharging_symbol` | `"⇣ "`                            | The symbol shown when the battery is discharging.   |
+| `unknown_symbol`     | `"❓ "`                            | The symbol shown when the battery state is unknown. |
+| `empty_symbol`       | `"❗ "`                            | The symbol shown when the battery state is empty.   |
 | `format`             | `"[$symbol$percentage]($style) "` | The format for the module.                          |
 | `display`            | [link](#battery-display)          | Display threshold and style for the module.         |
 | `disabled`           | `false`                           | Disables the `battery` module.                      |
@@ -1167,7 +1167,7 @@ By default the module will be shown if any of the following conditions are met:
 | ------------------- | ------------------------------------ | ------------------------------------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"` | The format for the module.                                                |
 | `version_format`    | `"v${raw}"`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
-| `symbol`            | `" "`                               | The symbol used before displaying the version of erlang.                  |
+| `symbol`            | `"e "`                               | The symbol used before displaying the version of erlang.                  |
 | `style`             | `"bold red"`                         | The style for the module.                                                 |
 | `detect_extensions` | `[]`                                 | Which extensions should trigger this module.                              |
 | `detect_files`      | `["rebar.config", "elang.mk"]`       | Which filenames should trigger this module.                               |
@@ -2204,7 +2204,7 @@ By default the module will be shown if any of the following conditions are met:
 | ------------------- | ------------------------------------ | ----------------------------------------------------------------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"` | The format for the module.                                                                            |
 | `version_format`    | `"v${raw}"`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch`                             |
-| `symbol`            | `" "`                               | A format string representing the symbol of Node.js.                                                   |
+| `symbol`            | `"⬢ "`                               | A format string representing the symbol of Node.js.                                                   |
 | `detect_extensions` | `["js", "mjs", "cjs", "ts"]`         | Which extensions should trigger this module.                                                          |
 | `detect_files`      | `["package.json", ".node-version"]`  | Which filenames should trigger this module.                                                           |
 | `detect_folders`    | `["node_modules"]`                   | Which folders should trigger this module.                                                             |
@@ -2480,7 +2480,7 @@ By default the module will be shown if any of the following conditions are met:
 | ---------------- | -------------------------------------------- | ------------------------------------------------------------------------- |
 | `format`         | `"via [$symbol($username@)$stack]($style) "` | The format string for the module.                                         |
 | `version_format` | `"v${raw}"`                                  | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
-| `symbol`         | `" "`                                       | A format string shown before the Pulumi stack.                            |
+| `symbol`         | `"⠿ "`                                       | A format string shown before the Pulumi stack.                            |
 | `style`          | `"bold 5"`                                   | The style for the module.                                                 |
 | `disabled`       | `false`                                      | Disables the `pulumi` module.                                             |
 
--- a/README.md
+++ b/README.md
@@ -167,7 +167,13 @@
 
 ### Prerequisites
 
-- A [Nerd Font](https://www.nerdfonts.com/) installed and enabled in your terminal (for example, try the [Fira Code Nerd Font](https://www.nerdfonts.com/font-downloads)).
+- A [Powerline Font](https://github.com/powerline/fonts) installed and enabled in your terminal.
+  Many coding fonts already include Powerline symbols or have official variants with them included.
+  (For example [Fira Code](https://github.com/tonsky/FiraCode) or [Cascadia Code/Mono PL](https://github.com/microsoft/cascadia-code/releases))
+
+  You can also use a [Nerd Font](https://www.nerdfonts.com/). (for example, try the [Fira Code Nerd Font](https://www.nerdfonts.com/font-downloads)).
+  Checkout [Presets](https://starship.rs/presets/#nerd-font-symbols) for a configuration that uses
+  Nerd Font symbols.
 
 ### Step 1. Install Starship
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,7 +28,13 @@ description: Starship is the minimal, blazing fast, and extremely customizable p
 
 ### Prerequisites
 
-- A [Nerd Font](https://www.nerdfonts.com/) installed and enabled in your terminal.
+- A [Powerline Font](https://github.com/powerline/fonts) installed and enabled in your terminal.
+  Many coding fonts already include Powerline symbols or have official variants with them included.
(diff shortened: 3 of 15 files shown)


docs/config/README.md
```diff
@@ -2204,7 +2204,7 @@ By default the module will be shown if any of the following conditions are met:
 | ------------------- | ------------------------------------ | ----------------------------------------------------------------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"` | The format for the module.                                                                            |
 | `version_format`    | `"v${raw}"`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch`                             |
-| `symbol`            | `" "`                               | A format string representing the symbol of Node.js.                                                   |
+| `symbol`            | `"⬢ "`                               | A format string representing the symbol of Node.js.                                                   |
```
jankatins, line 2207: This symbol is already taken by the container module: https://github.com/starship/starship/blame/98afd217c6002feb007b3c21c16ea22c22f7374d/docs/config/README.md#L686

But maybe that one could be replaced with 🔴  :-)

segevfiner: Others, please respond with your opinions as well before I make any change (To save me from shuffling things back and forth 😝).

chipbuster: We do already have duplicate symbols in starship (e.g. .NET's TFM and Dart use the same symbol, while perl and ocaml render identically in some emoji sets) so this isn't a killer. I'd like to get reviews from other maintainers signing off on this idea before proceeding though, since this has traditionally been a contentious issue.

chipbuster: Also, note that container is default `bold red dimmed` while nodejs is `bold green` so I don't think there is a risk these will be mistaken except by our colorblind users. It also depends on the colorscheme being used. I'll take a look with MacOS's colorblind filters at some point to see if this will be an issue under popular colorschemes.

chipbuster:
I would like the opinions of @starship/astronauts and @starship/mission-control given the amount of discussion that this has generated.

Here is my assessment of the issue. Note that I am weakly biased in favor of an emoji-based default, given the ease of configuration and usage:

### Status Quo

##### For

- We have an expanded symbol library to choose from (both emoji and NF symbols) for new modules or symbol changes to existing modules.
- Least amount of work

##### Against

- NerdFont configurational issues (see below)
- Emoji consistency issues (see below)

### All-Emoji Preset

##### For

- Simplest to set up: windows and MacOS already have emoji support on the terminal, and most Linux users just need to install one package and fallback will be automatically configured. No special configuration is needed since almost all IDEs/editors support emoji now.

##### Against

- Emoji availability is inconsistent: we will have to be careful about which emoji we choose and make sure they're available on most systems.
- Emoji display is inconsistent: what appears on one user's system will not necessarily appear on another's. For example, in some emoji-sets, one-humped and two-humped camel share a symbol, meaning that there is no way to distinguish perl and ocaml on the command line. This makes getting a 100% consistent experience more challenging.
- Emoji range is not as large: Emojis were not created to express programming/computer concepts, so some things are not easily expressible. For example, we can see in this PR that the symbols for battery are less easily-interpretable than their NF counterparts.

### All-Nerdfont Preset

##### For

- NerdFonts provide the most direct mapping of symbols to programming concepts: almost every major language and system has a symbol specifically for it, instead of having to reinterpret an existing symbol as a language (e.g crab --> Rust).
- NerdFonts are consistent: what you see on one system is (pretty much) what you'll see on another system.

##### Against
- Configuration burdens: configuring a NerdFont is actually not all that easy. Installing the font is usually not all that bad (unless you're on Linux without the font in a repo, in which case good luck to you), but actually configuring the application to use the font is a pain. Do you use the patched name (e.g. "FuraCode") or not (this depends on how you got the font and whether the maintainer buys in to the font-renaming procedure)? Do you modify your italic fonts or not? Do you use the "Mono" variant or not? And finally, if it doesn't work, how do you even begin to tell what went wrong? Did the font not install properly? Are you using the wrong name? Is the terminal you're using bugged?
- Related to above: selection of NerdFonts has to be done on a per-application basis, making the initial setup cost quite high.

chipbuster:
Storytime: A friend of mine recently told me that he wanted to install starship, but was scared off by the NerdFont requirement. 

He's a grad student in a top 10 CS department in the US, and spent just under a year as a software engineer for a startup before going to grad school. Grad students are not famed as good programmers in general, but I think it's fair to say that he's fairly smart and relatively capable with computers.

Based on that, I think it's fair to say that the NerdFont requirement presents a *very real* impediment to people who are newer to this whole commandline business, or just want to try out starship casually without really jumping all the way in.

davidkna:
I also think defaulting to emoji might be best. If this PR covers everything, NF symbols aren't used all that often in the default config, so it would be the smaller change. NF symbols rendering has also not been completely consistent, especially with the non-PUA characters in RTL-ranges.

docs/config/README.md
```diff
@@ -384,11 +384,11 @@ The module is only visible when the device's battery is below 10%.
...
-| `full_symbol`        | `" "`                            | The symbol shown when the battery is full.          |
-| `charging_symbol`    | `" "`                            | The symbol shown when the battery is charging.      |
-| `discharging_symbol` | `" "`                            | The symbol shown when the battery is discharging.   |
-| `unknown_symbol`     | `" "`                            | The symbol shown when the battery state is unknown. |
-| `empty_symbol`       | `" "`                            | The symbol shown when the battery state is empty.   |
+| `full_symbol`        | `"• "`                            | The symbol shown when the battery is full.          |
+| `charging_symbol`    | `"⇡ "`                            | The symbol shown when the battery is charging.      |
+| `discharging_symbol` | `"⇣ "`                            | The symbol shown when the battery is discharging.   |
+| `unknown_symbol`     | `"❓ "`                            | The symbol shown when the battery state is unknown. |
+| `empty_symbol`       | `"❗ "`                            | The symbol shown when the battery state is empty.   |
```
davidkna, line 391: I think all of these should use some sort of battery emoji or a battery emoji combined with a status emoji, in the past the meaning of these symbols wasn't clear enough without a battery context.

segevfiner: Might be. I just copied this over from the similar spaceship prompt.

Do you have any specific emojis in mind for this?

davidkna: 🔋 and ⚡️ from the configuration example  could work. There's also a low battery emoji in Unicode 14.0, but that's still too new.

andytom:
Firat thank you for the excellent summary @chipbuster. For me the key pro for this PR is that it lowers the barrier for entry to new users, as you have pointed out nerd fonts are not easy to get working correctly and have a lot of edge cases that casue lots of additional questions. I don't think it is too much of an extra burden for users who already have nerd fonts set up for their terminal of choice to download the preset to enable them for starship. Long term I don't think this will be completely solved until we have some kind of setup wizard and a way to use presets without having to copy/paste them from the website so this PR seems like the best middle ground for now.

matchai:
I'm also in favor of this change, though I imagine this should be a breaking change, with users' prompts changing from beneath them.
Are we prepared for a breaking change or should we hold off until we have a few other changes lined up for a major version bump? (perhaps a conversation to be had elsewhere)

chipbuster:
I think we need to have a conversation about what qualifies as a breaking change. IMO this would be less severe a change than #3414 and I believe the last time we looked at that, we decided that #3414 did not qualify as breaking.

chipbuster:
@segevfiner While we're hashing out the details, I think the relevant docs are the "Prerequisites" section in the main README, as well as [this FAQ entry](https://github.com/starship/starship/tree/master/docs/faq#why-dont-i-see-a-glyph-symbol-in-my-prompt).

A search for "nerd" or "font" in the `docs/` directory might also be a good idea, in case I missed something.

chipbuster:
Note: this is what it looks like in my system under red-green colorblindness simulation.

image

While it's dependent on colorscheme, I tried all default colorschemes in iTerm2 and they are easily distinguishable side-by-side (though under one or two schemes, they might be difficult to distinguish separately).

docs/config/README.md
chipbuster: Note: this is what it looks like in my system under red-green colorblindness simulation.

image

While it's dependent on colorscheme, I tried all default colorschemes in iTerm2 and they are easily distinguishable side-by-side (though under one or two schemes, they might be difficult to distinguish separately).

segevfiner:
How about we offer this as a preset until the time comes where you want to make a major version? This will allow to iterate on the icon choices as well.

chipbuster:
@segevfiner Excellent idea!

Finii:
Just noticed (through https://github.com/ryanoasis/nerd-fonts/issues/1110) that Starship uses the `heavy right-pointing angle quotation mark` (unicode `276F`) per default as prompt like ... image taken from above:

image

This glyph is neither part of Nerd Fonts nor of Powerline (to my best knowledge).
A patched font has it only if the unpatched font already had it, which not all fonts do.

I had users raise Issues with `Delugiua` and also with `Nerd Fonts`. The former has the symbol now, and Nerd Fonts will get that in the next release. Because there are people that want to use fonts that do not have that originally and some terminal clients do not support font fallback and/or have it set up wrongly.

I am no Starship user, so my knowledge is marginal. But it looks like it's hard coded

https://github.com/starship/starship/blob/1f3353e5c64f5f8d56ce6cf30283839e9a9d107a/src/configs/character.rs#L26

(well can be overridden but that is 'advanced').

I'm not really sure where you expect users to get that glyph from.
Just wanted to mention this in this seemingly related PR, please ignore if this is not appropriate.

GreyXor:
Hello, there's ETA for this becoming "the default preset in a future release of starship." ?

docs/config/README.md
kryzet: Has anything changed with the UI update in macOS 26?

davidkna (commented):
Please fix the merge conflicts, so this can be merged.

injust:
For migration, I think we should have (at the bare minimum) a list of modules with changed defaults.

zmberber:
I have looked at the current starship symbols and searched them for problematic symbols.  I am not sure if I found all of them, but this is what I could find.

These are enabled by default and broken without Nerd Fonts (they are PUA glyphs:
nonstandard codepoints that no normal font contains).  I am in favor of changing all of these.
- ``: git_branch. Powerline PUA (U+E0A0). Every git repo shows this.
  - The merge request says "emoji and powerline glyphs by default", but since this is the only powerline symbol in use, and powerline is PUA, I would suggest changing this one, which the PR doesn't do.
- ``: nodejs. Devicons PUA (U+E718).
- ``: erlang. Devicons PUA (U+E7B1).
- ``: pulumi. Font Awesome PUA (U+F1B2).
- `󰁹`/`󰂄`/`󰂃`/`󰂑`/`󰂎`: battery icons. Nerd Fonts SupPUA. All five states.

The following PUA symbols are disabled by default, but for the same reason as above, I am also in favor of changing these.
- ``: fossil_branch, hg_branch, pijul_channel. Same Powerline glyph as git_branch.
- `󰠅`: azure. Nerd Fonts SupPUA.
- ``: os module (Void Linux). PUA.
- ``/``/``/``/``/``/``: os module future entries. Font Awesome PUA.

Adding to what @Finii said, the following are enabled by default, and are sometimes broken.  These are real Unicode (non PUA) but in obscure blocks.  I would maybe handle these in a separate PR:

- `❯` / `❮`: the prompt character and vimcmd variant. Dingbats block (U+276F/U+276E). Many terminal fonts skip Dingbats entirely.
- `ஃ`: julia. Tamil visarga (U+0B83).
- `⍱`: vagrant. APL symbol (U+2371). Terminal fonts almost never have APL.
- `⎈`: helm. Misc Technical (U+2388), spotty coverage.
- `⌘`: haxe. Also Misc Technical (U+2318).
- `∙`: continuation prompt. Bullet operator (U+2219), minimal fonts often omit it.
- `⬢`: container. Black hexagon (U+2B22).
- `⬢`: meson. Same.

And the following are also real Unicode (non PUA) but in obscure blocks, but diabled by default.  I would also like to handle these in a separate PR.

- `☸`: kubernetes. Wheel of Dharma (U+2638), not very common.
- `⏲`: os module (InstantOS). Timer clock (U+23F2), Misc Technical.
- `Ⓐ`: os module (ALTLinux). Circled Latin A (U+24B6), Enclosed Alphanumerics.

PR #3547, opened 2022-01-31
feat: Enable Undistract Me for Linux

schrieveslaach:
#### Description

In #1019 the undistract feature has been added to starship without
enabling it by default due to the system dependency on libdbus. With
recent versions of notify-rust the dependency is no longer required and
the feature can be enabled by default.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -26,9 +26,6 @@ jobs:
           profile: minimal
           components: rustfmt
 
-      - name: Setup | libdbus (ubuntu)
-        run: sudo apt-get install libdbus-1-dev
-
       - name: Build | Format
         run: cargo fmt --all -- --check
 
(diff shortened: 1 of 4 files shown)


andytom (approved):

chipbuster:
Sweet! @schrieveslaach thanks for the update, and thanks for reviewing @andytom.

schrieveslaach:
@chipbuster, thanks for merging!

trajano:
Is this available on Windows?

chipbuster:
Yes, I believe this should now build on Windows by default. The problem was that `notify-rust` depended on a piece of software on Linux that wasn't installed by default.

Windows never relied on this, so it should be available natively if you enable it. See [here](https://starship.rs/config/#command-duration) for the settings to enable it (specifically, anything with `notification` in the name)

happysalada:
hey, sorry to bother, on nix and darwin (macos) we are experiencing compilation problems with this depedency.
If I made a PR to put notify-rust behind a feature flag and enable it by default, would you be ok with this ?
(adding the feature, would be just to disable it for darwin on nix)

There are various discussions as to where exactly the problem is coming from, but until we find a solution, it would be much easier to have this disable-able.

davidkna:
@happysalada Sounds good.

PR #3560, opened 2022-02-03
feat(cli): Print arguments if argument parsing fails

### LINKED ISSUE #3554: Print arguments to starship if argument parsing fails

A majority of starship invocations are not made directly by the user, but are instead made by the shell adapter. In the event that the shell adapter screws up, it's not always clear from the error message what went wrong.

c.f. #3553, where having this information would make it much easier to debug what went wrong in the shell adapter.

If Clap reports an error, print the command line `argv` that starship was called with, so that we can see what arguments were used.

This has the potential to be annoying for commands that *are* used interactively (e.g. bug-report) so we could special-case this to e.g. if the command is `prompt` or if it's unknown.

Ideally ideally, this could be done by Clap itself, but we can also just grab os.args() if we need to.

Haven't thought about it much.

CyberGsus:
#### Description

I've changed `main` in `src/main.rs` to call `Cli::try_parse` instead
of `Cli::parse` to manage the error case manually and print the arguments passed
to starship when argv parsing error, just like the issue explains. Other code has not been touched;
thoughts on those changes are in the commit message.

I'm not sure how to test this properly, but I certainly want to discuss:
  - whether all arguments are printed correctly:
    - what does 'correctly' mean?
      - do we need any escaping?
      - do we need any measure of 'easy to read'?
  - exiting with code 2 when having an error (code 2 is what [`clap::Error`](https://docs.rs/clap/latest/clap/struct.Error.html#method.exit) uses)
  - exit 0 on `--help`/`-h`
  - whether to exit with code 2/print arguments if `ErrorKind` is `DisplayHelp*`
    - certainly, exiting with code 2 when help was wanted could be misleading and it currently exits w/ code 0
    - will printing arguments be useful here?

#### Motivation and Context

Closes #3554

image

- [x] I have tested using **Linux**\*

#### Checklist:

\*: As I've commented above, tests are incomplete. Though current tests (`cargo test`) have run OK.
EDIT: Added a screenshot of the error + note preview

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -108,7 +108,23 @@ fn main() {
     let _ = ansi_term::enable_ansi_support();
     logger::init();
 
-    let args = Cli::parse();
+    let args = match Cli::try_parse() {
+        Ok(args) => args,
+        Err(e) => {
+            // print the error
+            e.print().expect("error while writing error");
+            // print the arguments
+            eprintln!(
+                "\nNOTE:\n    passed arguments: {:?}",
+                // collect into a vec to format args as a slice
+                std::env::args().skip(1).collect::<Vec<_>>()
+            );
+
+            // clap exits with status 2 on error: 
+            //  https://docs.rs/clap/latest/clap/struct.Error.html#method.exit
+            std::process::exit(2);
+        }
+    };
     log::trace!("Parsed arguments: {:#?}", args);
 
     match args.command {
--- a/src/main.rs
+++ b/src/main.rs
@@ -111,16 +111,14 @@ fn main() {
     let args = match Cli::try_parse() {
         Ok(args) => args,
         Err(e) => {
-            let is_info_only = matches!(
-                e.kind,
-                clap::ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand
-                    | clap::ErrorKind::DisplayHelp
-                    | clap::ErrorKind::DisplayVersion
-            );
-            // print the error and avoid panicking in case of stdout/stderr closing unexpectedly
+            // if the error is not printed to stderr, this means it was not really
+            // an error but rather some information is going to be listed, therefore
+            // we won't print the arguments passed
+            let is_info_only = !e.use_stderr();
+            // print the error and void panicking in case of stdout/stderr closing unexpectedly
             let _ = e.print();
             // if there was no mistake by the user and we're only going to display information,
-            // we won't put arguments or exit
+            // we won't put arguments or exit with non-zero code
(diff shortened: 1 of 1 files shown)


CyberGsus:
PS: I've seen that the zsh script might call starship with many arguments, so probably an env var that sets this optionally (e.g `export VAR_NAME=1` before the binary is run again through the prompt shell functions) to avoid filling the screen with the argument list

src/main.rs
```diff
@@ -108,7 +108,36 @@ fn main() {
...
+        Ok(args) => args,
+        Err(e) => {
+            let is_info_only = matches!(
+                e.kind,
+                clap::ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand
+                    | clap::ErrorKind::DisplayHelp
+                    | clap::ErrorKind::DisplayVersion
+            );
+            // print the error
+            e.print().expect("error while writing error");
```
davidkna, line 121: It's better to avoid panicking here (#3280):
Upstream `parse` uses this:
```suggestion
            // avoid panic if stdout/stderr close unexpectedly
            let _ = e.print();
```

src/main.rs
```diff
@@ -108,7 +108,36 @@ fn main() {
...
+            // if there was no mistake by the user and we're only going to display information,
+            // we won't put arguments or exit
+            let exit_code = if is_info_only {
+                0
+            } else {
+                // print the arguments
+                eprintln!(
```
davidkna, line 128: Same thing:
```suggestion
                let mut stderr = io::stderr();
                let _ = writeln!(
                    stderr,
```

src/main.rs
CyberGsus: thank you, will push asap

src/main.rs
```diff
@@ -108,7 +108,40 @@ fn main() {
...
-    let args = Cli::parse();
+    let args = match Cli::try_parse() {
+        Ok(args) => args,
+        Err(e) => {
+            let is_info_only = matches!(
+                e.kind,
+                clap::ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand
+                    | clap::ErrorKind::DisplayHelp
+                    | clap::ErrorKind::DisplayVersion
+            );
```
davidkna, line 119: clap uses this check internally:
```suggestion
            let is_info_only = !e.use_stderr();
```

Should be slightly more reliable. Beyond this LGTM.

davidkna (approved):

andytom:
Thank you for your contribution @CyberGsus and thanks for reviewing @davidkna .

PR #3563, opened 2022-02-04
fix: escape text segments in meta variables

### LINKED ISSUE #3222: character module incorrectly renders "}" after "%" (zsh)
(`starship bug-report` tried including my whole .zshrc in the body of the report and was truncated, so creating this manually.)

#### Current Behavior
The character module has an extra `}` after the configured `%` content. This appeared after upgrading to 1.0.0 via homebrew. It was not occurring on 0.58.0 with the same configuration.

#### Expected Behavior
The output of the module should just be `%`.

Screenshot 2021-11-10 at 17 26 37

On 0.58.0, there is an extra `%` between the two `%{%}`s at in the 2nd group. Looks like it's not being escaped properly now?

At a quick guess, c1f2d345aac0b0241ea1b6d99977fea20fa3f5bb seems a likely commit to have changed the behaviour.

#### Environment
- Starship version:
```
starship 1.0.0
branch:
commit_hash:
build_time:2021-11-09 21:30:28 +00:00
build_env:rustc 1.56.1,
```

- Shell type: zsh
- Shell version: zsh 5.8 (x86_64-apple-darwin19.6.0)
- Shell plugin manager: none, hand-crafted zsh config :)
- Terminal emulator: iTerm2
- Operating system: macOS 11.6.1

Source is at:

https://github.com/jamesoff/zsh/blob/master/local-plugins/starship/init.zsh (written originally to allow me to mix and match a self-built starship from local source with the brew-installed one)
https://github.com/jamesoff/zsh/blob/master/.zshrc#L325

#### Starship Configuration

```bash
add_newline = false
right_format = "$all"

[aws]
format = '[($symbol($profile)(/$region))]($style)'
symbol = " "
style = "#FF9900"
disabled = true

[custom.aws]
command = "echo -n $AWS_PROFILE; if [[ -n $AWS_DEFAULT_REGION ]]; then echo -n \"/$AWS_DEFAULT_REGION\"; fi; echo"
when = "[[ -n $AWS_PROFILE ]]"
description = "Show AWS profile and region if set"
format = "[$symbol$output]($style) "
symbol = " "
style = "#FF9900"

[character]
success_symbol = "[%](blue)"
error_symbol = "[%](red)"

[directory]
style = "white"

[line_break]
disabled=false

[custom.shlvltmux]
# have to substract two as the shell being used for evaluating output adds 1, as does tmux
command = "echo $(( $SHLVL - 2 ))"
when = "[[ -n $TMUX && $SHLVL -gt 3 ]]"
description = "Show the current shell nesting level (in tmux)"
symbol = " "
style = "blue"
format = "[$symbol$output]($style) "

[custom.shlvl]
# have to substract one as the shell being used for evaluating output adds 1
command = "echo $(( $SHLVL - 1 ))"
when = "[[ -z $TMUX && $SHLVL -gt 2 ]]"
description = "Show the current shell nesting level"
symbol = " "
style = "blue"
format = "[$symbol$output]($style) "

[custom.awscreds]
command = "echo static credentials"
when = "[[ -n $AWS_ACCESS_KEY_ID && -z $AWS_SESSION_TOKEN ]]"
description = "Warn if AWS credentials are configured in the environment"
symbol = "⚠️  "
style = "red"
format = "with [$symbol$output]($style) "

[custom.awscredstemp]
command = "echo session credentials"
when = "[[ -n $AWS_ACCESS_KEY_ID && -n $AWS_SESSION_TOKEN ]]"
description = "Warn if temporary AWS credentials are configured in the environment"
symbol = "⚠️  "
style = "yellow"
format = "with [$symbol$output]($style) "

[package]
symbol = " "

[perl]
symbol = " "

[php]
symbol = " "

[python]
symbol = " "
detect_folders=[".venv"]
style="#366E9D"
format='[${symbol}${pyenv_prefix}(${version} )(\($virtualenv\) )]($style)'

[ruby]
symbol = " "

[rust]
symbol = " "

[git_branch]
symbol = " "
style="#F45D27"

[git_status]
style="#F44D27"


[status]
disabled=false
format="[($status)]($style)"

```

davidkna:
#### Description

In addition to `map` also handle escape for `Text` segments  in meta variables in `parse_format` for `map_meta` akin to the handling for regular `map`. I'm not very familiar with this area of the code, but it seems to work.

#### Motivation and Context

Closes #3222

#### Checklist:

### DIFF
--- a/src/formatter/string_formatter.rs
+++ b/src/formatter/string_formatter.rs
@@ -287,7 +287,16 @@ impl<'a> StringFormatter<'a> {
                 .into_iter()
                 .map(|el| {
                     match el {
-                        FormatElement::Text(text) => Ok(Segment::from_text(style, text)),
+                        FormatElement::Text(text) => Ok(Segment::from_text(
+                            style,
+                            shell_prompt_escape(
+                                text,
+                                match context {
+                                    None => Shell::Unknown,
+                                    Some(c) => c.shell,
+                                },
+                            ),
+                        )),
                         FormatElement::TextGroup(textgroup) => {
                             let textgroup = TextGroup {
                                 format: textgroup.format,


matchai (approved):
LGTM 👍

chipbuster:
> This seems to break the #3546 workaround.

We can temporarily fix this by adding a custom module (since custom module outputs are unescaped) which prints the appropriate zsh escape sequence. We can then set the continuation prompt to only print the custom module.

This will cause the custom module to always print, but we can then arrest this by adding a `when` check for an environment variable `STARSHIP_CONTINUATION` and then modifying the starship init script to set this environment variable only on continuation prompt draw.

The final modifications to files looks like this:

```
# In starship.toml

# continuation_prompt = "[%_∙](bright-black) "
continuation_prompt = "${custom.zshcontinuation}"

[custom.zshcontinuation]
command = "printf '%s' '%_∙'" # shows output of command
shell = "zsh"
when =""" test "$STARSHIP_CONTINUATION" = "true" """
format = "[$output ]($style)"
style = "bright-black"
```

```zsh
# In zsh init file (usually .zshrc)
# First line already exists, add new line right below it
eval $(starship init zsh)
PROMPT2="$(STARSHIP_CONTINUATION=true $(which starship) prompt --continuation)"
```

Note that this example uses `starship` and `which starship`, but if you are using an absolute path to starship, you should use it in both locations, e.g.  


```zsh
eval $(/programs/starship init zsh)
PROMPT2="$(STARSHIP_CONTINUATION=true /programs/starship prompt --continuation)"
```

Under this regime, we'll have a result that correctly emulates the current behavior:

image

bitti:
This also broke some of my prompts, since I use escaped `\[ \]` sequences to compensate for overwidth characters. I know you're supposed to use a special single width unicode font with starfish, but this was a nice way to avoid that.

E.g. I've
```
not_executable_symbol = '🚫\\\[ \\\]'
```
Where bash is supposed to end up seeing `🚫\[ \]` where the escaped brakets compensate for the double width of 🚫 as documented under `LESS='-p ^PROMPTING' man bash`:

    \[     begin a sequence of non-printing characters, which could be used to embed a terminal control sequence into the prompt
    \]     end a sequence of non-printing characters

Instead I get a literal `🚫\[ \]`in the prompt now. Is there a workaround for that? When will 
https://github.com/starship/starship/pull/3563#issuecomment-1139849022 be implemented?

PR #3567, opened 2022-02-06
Update README.md

Shan-Zhou:
Update the doc for Nushell.

#### Description

In Nushell the prompt should be "starship prompt" instead of "starship_prompt". 

#### Motivation and Context

The later one leads to an error (prompt "starship_prompt" not found) and could be fixed by replacing the underscore with a space.

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,7 +333,7 @@ startup = [
   "starship init nu | save ~/.cache/starship/init.nu",
   "source ~/.cache/starship/init.nu",
 ]
-prompt = "starship_prompt"
+prompt = "starship prompt"
 ```
 
 </details>


davidkna:
Did you add the `startup` section to your nu config? It defines a `starship_prompt` function that will pass additional arguments to `starship prompt` to integrate it with the shell.

Shan-Zhou:
> Did you add the `startup` section to your nu config? It defines a `starship_prompt` function that will pass additional arguments to `starship prompt` to integrate it with the shell.

Yes I added the startup section. It seems that starship_prompt works in most cases, except for setting nushell configs by commands like "config set path $nu.path$ (after running "config set ...", I get an error message "command starship_prompt not found" and starship stops to work.

davidkna:
This looks more like an upstream issue regarding the handling of `startup`. I don't think falling back to `starship prompt` is a good option. The config format is going to change significantly in the next nu version, please create an issue if you still experience your issue once the init scipt is adapted to that.

PR #3571, opened 2022-02-07
ci: Notarize MacOS Binaries and Add Flat Package Installers

### LINKED ISSUE #1774: Notarize Starship
With MacOS evolving most applications need to be code signed and Notarized before they can be installed on MacOS 

Please notarize the app as described here https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution

chipbuster:
#### Description
This PR adds notarization, codesigning, and flat packages (.pkg) to our deploy workflow.

It also adds internal documentation for the procedure, since it was so stupidly complicated.

#### Motivation and Context

Closes #1774

image

(Unfortunately, it looks like the logos don't show up in dark mode, for any flat installers)

This was tested by using a dummy job (consisting of part of the `github_build` job from the deploy workflow) to build simulated build artifacts, then placing the new job after that as a dependent workflow.

There is no guarantee that this will work in the real deploy workflow, but I think this is the best that I can do.

To see the GitHub workflow that I (believe) verified that this was basically-working, see https://github.com/starship/starship/blob/c3932b327a2bec9e1e919f17cf4934e96cf9f1a8/.github/workflows/apple_deploy.yml.

To see the artifacts generated by that workflow, see https://github.com/starship/starship/actions/runs/1843775397 (will expire approximately at the end of Feb).

#### Checklist:


chipbuster:
I apologize for the somewhat horrific mess of commits involved here. There are interesting implications to only being able to test code by pushing.

davidkna:
I'm getting a "starship cannot run because it is from an unidentified developer" prompt when running `starship` via `starship-aarch64-apple-darwin.tar.gz` from the linked workflow artifacts. Is that supposed to happen?

I think it might be to test the current version of the workflow with a [workflow_dispatch trigger](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/).

davidkna:
> From my understanding, we can just keep the trigger permanently embedded this into the deploy workflow?

I think so.

chipbuster:
Rebased to get rid of some nasty shenanigans where my fork's master branch somehow picked up a bunch of unrelated PRs.

(Also to get rid of the 50+ commits which were starting to hurt when I looked at them).

.github/workflows/deploy.yml
```diff
@@ -105,6 +105,100 @@ jobs:
...
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Notarize | Set up secrets
+      env:
+        APP_CERTIFICATE_BASE64: ${{ secrets.APPLEDEV_APPSIGNKEY_BASE64 }}
+        INSTALL_CERTIFICATE_BASE64: ${{ secrets.APPLEDEV_INSTALLERSIGNKEY_BASE64 }}
+        P12_PASSWORD: ${{ secrets.APPLEDEV_SIGNKEY_PASS }}
+        KEYCHAIN_PASSWORD: ${{ secrets.APPLEDEV_SIGNKEY_PASS }}
+        APPLEID_USERNAME: ${{ secrets.APPLEDEV_ID_NAME }}
+        APPLEID_TEAMID: ${{ secrets.APPLEDEV_TEAM_ID }}
```
matchai, line 139: Out of curiosity, will this be signed as Starship or as you?

install/macos_packages/build_and_notarize.sh
```diff
@@ -0,0 +1,93 @@
...
+    usage
+    exit 1
+fi
+
+starship_binary="$1"
+starship_docs_dir="$2"
+arch="$3"
+pkgname="${4:-}"
+
+if [[ ! -d "$starship_docs_dir/.vuepress/dist" ]]; then
```
matchai, line 50: Nice 👍

install/macos_packages/build_component_package.sh
```diff
@@ -0,0 +1,90 @@
...
+
+if [ ! -d "$starship_documentation_dir" ]; then
+    error "Could not find starship documentation at $starship_documentation_dir"
+fi
+
+pkgdir="$(mktemp -d)"
+mkdir -p "$pkgdir/usr/local/bin"
+cp "$starship_program_file" "$pkgdir/usr/local/bin/starship"
+
+# Now we get to make documentation! Vuepress was not designed to build locally
```
matchai, line 57: Oof. If it must be done! 🙃

install/macos_packages/readme.md
```diff
@@ -0,0 +1,262 @@
+# MacOS Codesigning Scripts
```
matchai, line 1: Absolutely *love* this documentation! 😍 
Hopefully we won't need to refer to it too often 🤞

matchai (approved):
Looks great (as a person with zero experience with signing and notarization)!
Thank you for taking on this _huge_ task!!

.github/workflows/deploy.yml
chipbuster: This is going to be signed by me. We can get signing certificates for at least one other mission control member as well, in case I get kidnapped by cephalopods or hit by a bus, but Apple has some [rather high requirements](https://developer.apple.com/programs/enroll/) for enrolling an organization: in particular, the organization needs to have legal entity status (be capable of sued) and a [DUNS number](https://www.dnb.com/duns-number.html).

Also, I believe the signing certificate I exported was due to expire in 2027, so if it happens in the next couple years, you should have quite a bit of time to work out a replacement :)

install/macos_packages/build_and_notarize.sh
```diff
@@ -0,0 +1,93 @@
...
+if [[ -z ${KEYCHAIN_ENTRY+x} ]]; then
+  error "Environmental variable KEYCHAIN_ENTRY must be set."
+fi
+
+if [[ -z ${RUNNER_TEMP+x} ]]; then
+  error "Environmental variable RUNNER_TEMP must be set."
+fi
+
+if [[ -z ${KEYCHAIN_FILENAME+x} ]]; then
+  error "Environmental variable KEYCHAIN_ENTRY must be set."
```
chipbuster, line 32: Oops.

install/macos_packages/build_component_package.sh
chipbuster: Yeah, I was a little surprised that `npm serve` didn't actually yield an HTML hierarchy that could be scraped by wget. It seems to do the thing where you have a single line of JS which can be evaluated to generate the DOM for the real deal, but wget can't evaluate the JS, so it just pulls a single script.

chipbuster:
Results from a straight copy-paste of the current subsection of the deploy workflow can be found at https://github.com/starship/starship/actions/runs/1906861805 (with irrelevant artifacts deleted).

If someone could confirm that the binaries in the `.tar.gz` are notarized and the `.pkg` files are notarized, I think we'd be ready to merge.

Command to check notarization: `spctl -a -vvv -t install <file>`

williamwebb:
> If someone could confirm that the binaries in the `.tar.gz` are notarized and the `.pkg` files are notarized, I think we'd be ready to merge.
> 
> Command to check notarization: `spctl -a -vvv -t install <file>`

Both LGTM

image

davidkna:
The binaries in the `tar.gz` file are reported as signed and notarized. The `pkg` are only reported as validly signed, but there's no warning when attempting to run it and the installed binary is reported as signed and notarized.

davidkna:
Also, Finder still doesn't seem to like the binary files from the tar.gz files.

chipbuster:
> Also, Finder still doesn't seem to like the binary files from the tar.gz files.

What does it do? On my system it seems to display a terminal executable and that's about it

image

williamwebb:
When attempting to run `starship` directly from `finder` you get,

image

Which I believe is because of this,

image

chipbuster:
Oh I see. I'm not sure why, but I'm not getting that even when using the versions downloaded from the GH Action. Possibly because I codesigned something with an identical hash locally, so Gatekeeper is letting it through?

Did previous releases of starship have this issue? Not sure if there's much we can do about that at the moment since binaries still can't be stapled.

davidkna:
> Did previous releases of starship have this issue?

I don't think any past release got past gatekeeper.

Also, despite the missing stapling, `Terminal.app` will execute it without complaints. So it's probably fine.

chipbuster:
Alright. I'd like to do one last set of tests on a clean slate Mac (my university has a few public access terminals that should work nicely) and if those all come up clean, I'll go ahead and merged this.

.github/workflows/deploy.yml
matchai: I'm not at my computer right now, but we may qualify as part of our Open Collective fiscal sponsorship, which qualifies us as a non-profit in the US, which may give us that legal status.

chipbuster:
Unfortunately, I am unable to replicate these results on any Macs I tested, including school macs, which should (in principle) be running on the strictest policy possible. I don't know what's causing the difference, but all systems I tried reported that the new binaries and pkgs are notarized and David reports that the binaries run in the Terminal without problems, so I'm going to go ahead and declare that this is working and accept new issues to the contrary.

chipbuster:
Thank you @davidkna, @matchai, @williamwebb, and @JaySoyer for ideas, verification, debugging, and review! This was a hell of a PR and I couldn't have gotten any of this done without your help.

.github/workflows/deploy.yml
chipbuster: Oh interesting. If that's the case, I believe you would have to be the one to do the registration, since another requirement is

> As the person enrolling your organization in the Apple Developer Program, you must have the legal authority to bind your organization to legal agreements. You must be the organization’s owner/founder, executive team member, senior project lead, or an employee with legal authority granted to you by a senior employee.

Once that's done though, I believe individuals should be able to acquire signing keys that work on behalf of the organization.

I don't believe that signing keys from an organization would fundamentally be any different than those from an individual (the organization ID + signing key IDs would have to change in the GH secrets, but that's about it) so switching over if we decide to go for an Apple Dev organization shouldn't be too much of an issue.

PR #3587, opened 2022-02-09
feat(haskell): Add Haskell module

### LINKED ISSUE #1458: Support for Haskell

ksqsf:
#### Description
Add Haskell module to detect the version of GHC and/or the selected Stack resolver.

#### Motivation and Context
Support one more popular industrial-strength programming language.

Fixes #1458
Fixes #2601

GHC

Stack resolver

Tested by running it on my machine.

- I have GHC 9.0.2 installed (via GHCup) as a default GHC compiler.
- Stack is also popular among Haskellers: it manages GHC version itself, so [a resolver version (e.g. `lts-18.12`) determines a specific GHC version](https://www.stackage.org), and Stack will install it when necessary.

This is a standalone module and it does not affect other code.

(I don't have Linux and Windows environments at the moment.)

#### Checklist:

### DIFF
--- a/src/configs/haskell.rs
+++ b/src/configs/haskell.rs
@@ -21,7 +21,7 @@ impl<'a> Default for HaskellConfig<'a> {
             format: "via [$symbol($version )]($style)",
             version_format: "v${raw}",
             symbol: " ",
-            style: "purple",
+            style: "bold purple",
             disabled: false,
             detect_extensions: vec!["hs", "cabal", "hs-boot"],
             detect_files: vec!["stack.yaml", "cabal.project"],
--- a/src/configs/haskell.rs
+++ b/src/configs/haskell.rs
@@ -1,0 +1,31 @@
+use crate::config::ModuleConfig;
+
+use serde::Serialize;
+use starship_module_config_derive::ModuleConfig;
+
+#[derive(Clone, ModuleConfig, Serialize)]
+pub struct HaskellConfig<'a> {
+    pub format: &'a str,
+    pub version_format: &'a str,
+    pub symbol: &'a str,
+    pub style: &'a str,
+    pub disabled: bool,
+    pub detect_extensions: Vec<&'a str>,
+    pub detect_files: Vec<&'a str>,
+    pub detect_folders: Vec<&'a str>,
+}
+
+impl<'a> Default for HaskellConfig<'a> {
+    fn default() -> Self {
+        HaskellConfig {
+            format: "via [$symbol($ghc_version )]($style)",
+            version_format: "v${raw}",
+            symbol: " ",
+            style: "purple",
+            disabled: false,
+            detect_extensions: vec!["hs", "cabal"],
+            detect_files: vec!["stack.yaml"],
+            detect_folders: vec![],
+        }
+    }
+}
--- a/src/configs/haskell.rs
+++ b/src/configs/haskell.rs
@@ -20,7 +20,7 @@ impl<'a> Default for HaskellConfig<'a> {
         HaskellConfig {
             format: "via [$symbol($version )]($style)",
             version_format: "v${raw}",
-            symbol: " ",
+            symbol: "λ ",
             style: "bold purple",
             disabled: false,
             detect_extensions: vec!["hs", "cabal", "hs-boot"],
--- a/src/configs/haskell.rs
+++ b/src/configs/haskell.rs
@@ -18,13 +18,13 @@ pub struct HaskellConfig<'a> {
 impl<'a> Default for HaskellConfig<'a> {
     fn default() -> Self {
         HaskellConfig {
-            format: "via [$symbol($ghc_version )]($style)",
+            format: "via [$symbol($version )]($style)",
             version_format: "v${raw}",
             symbol: " ",
             style: "purple",
             disabled: false,
-            detect_extensions: vec!["hs", "cabal"],
-            detect_files: vec!["stack.yaml"],
+            detect_extensions: vec!["hs", "cabal", "hs-boot"],
+            detect_files: vec!["stack.yaml", "cabal.project"],
             detect_folders: vec![],
         }
     }
--- a/src/modules/haskell.rs
+++ b/src/modules/haskell.rs
@@ -61,25 +61,21 @@ fn get_ghc_version(context: &Context) -> Option<String> {
 }
 
 fn get_stack_resolver_version(context: &Context) -> Option<String> {
+    if !is_stack_project(context) {
+        return None;
+    }
     let file_contents = utils::read_file(context.current_dir.join("stack.yaml")).ok()?;
     let yaml = yaml_rust::YamlLoader::load_from_str(&file_contents).ok()?;
-    let version = yaml.first()?["resolver"]
-        .as_str()
-        .or(yaml.first()?["snapshot"].as_str())?;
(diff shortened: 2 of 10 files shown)


ksqsf:
I didn't realize that there was at least two previous attempts (#2601)!

Please close my PR if there is any conflict.

chipbuster:
@ksqsf No conflict, but given that performance is going to be the primary blocker to merging (and what more-or-less sank the previous attempts) it'd be good to get some benchmark numbers to see what kind of performance we can expect from this PR.

If those numbers look good, I'll run some more tests on my end to make sure they agree, then we can talk about features + configuration.

ksqsf:
@chipbuster Hi, sure!

On my machine,

| Command | Avg. Time |
|-----------|-----------|
| ghc --numeric-version |  76ms |
| python --version | 46ms |
| python3 --version | 15ms |
| rustc --version | 127ms |
| rustup run <...> rustc --version | 33ms |

I don't know what is the exact criterion and if there is a standard benchmark practice, though. I've been using my fork for some time, and I consider this acceptable personally.

The `stack.yaml` support uses starship's yaml parser like the Kubernetes module, so it should be as fast.

chipbuster:
Well, this within acceptable ranges. I'll do some testing on my lower powered systems and see what I find.

chipbuster:
Interesting. My VPS has absolutely no problem with this (less than half the times you reported). I'll also want to test on my old x64 MBP, since this is a system that has historically been challenging in terms of latency for version commands.

chipbuster:
Okay, after comparing against other commands, I think `ghc --numeric-version` is still too slow to be default: on my 2015 MBP, it takes almost 160ms to complete (which is in-line with the benchmarks we had last time we decided to pull it).

That being said, this module doesn't fully rely on `ghc --numeric-version` to get its version info, the same way the Rust module currently treats `rustc --version` as a fallback. So the real question is how often we expect to see Haskell projects with vs without Stack, and whether we're willing to pay the extra runtime cost if there isn't.

Does anyone have thoughts on this? The time to run `ghc --numeric-version` is (with warm caches) roughly the same as `java --version`. This was the initial implementation of the Java module which I blocked on performance concerns (it now uses `-Xinternalversion` which is much faster), but the default behavior is to attempt to resolve via a stack file first, which (as I understand it) is currently the gold standard for dependency management in Haskell.

ksqsf:
Maybe we can remove `ghc --numeric-version` and only check `stack.yaml`?

andytom:
I am not a Haskell user so this is kind of an outsider looking in.

Since getting the version from the stack file is fast and seems to be the common case (again as an outsider) I don't see using `ghc - - numeric-version` as a fall back to be a problem. We might have a few cases where the prompt is slow but these "should" be exceptions rather than the rule.

chipbuster:
Okay, I generally concur. The last time I used Haskell seriously, the community still seemed to be evenly split between stack and cabal. It seems that stack has taken over since then, and if anyone has performance issues, they can disable the module since there doesn't seem to be anything else we can do without implementing async rendering.

docs/presets/README.md
```diff
@@ -44,6 +44,9 @@ symbol = " "
 [golang]
 symbol = " "
 
+[haskell]
```
chipbuster, line 47: Unfortunately, the preset system has been reworked somewhat in 1.4.0. The new location for preset files is `starship/docs/.vuepress/public/presets/toml`. Could you revert this to the version on master and then update the preset files in there?

src/configs/haskell.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub detect_files: Vec<&'a str>,
+    pub detect_folders: Vec<&'a str>,
+}
+
+impl<'a> Default for HaskellConfig<'a> {
+    fn default() -> Self {
+        HaskellConfig {
+            format: "via [$symbol($version )]($style)",
+            version_format: "v${raw}",
+            symbol: " ",
```
chipbuster, line 23: I can't actually see this symbol in the web browser due to font issues. Is this supposed to be a nerdfont symbol? If so, we'll need to change it over to unicode.

NerdFont symbols can go in the NerdFont preset if you'd like to preserve it.

chipbuster: (Same applies to documentation for defaults).

src/modules/haskell.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn folder_without_hs_files() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        let actual = ModuleRenderer::new("haskell").path(dir.path()).collect();
+        let expected = None;
+        assert_eq!(expected, actual);
+        dir.close()
+    }
+
+    #[test]
+    fn folder_stack() -> io::Result<()> {
```
chipbuster, line 106: Can we also add a test case here with a more polluted YAML content? Something that we might see in real life, e.g. with packages and flags. Could be a toy project with a few dependencies, or just something pulled from a random repo.

I'm not so concerned about it for this specific PR, but it would be a small amount of protection against future breakage.

chipbuster (dismissed):
Overall, a lovely PR. I suppose I should expect nothing less from a Haskeller.

You'll need to rebase on master because the preset system change has unfortunately introduced some conflicts (I left a more specific comment at the site).

Also, the choice of default symbol is unfortunately not going to work here. Due to #3544, we're slowly phasing out NerdFont symbols as default symbols in all modules. Would a bold `λ` work in this case? You can insert the NerdFont symbol into the NerdFont preset, if you'd like.

chipbuster:
@ksqsf I hit the wrong button, meant to request changes, not approve. Sorry :|

src/configs/haskell.rs
ksqsf: Yes, it is a NerdFont symbol. 
image

Moved to the NerdFont preset.

src/modules/haskell.rs
ksqsf: Copy-pasted a stack file from Reanimate. https://github.com/reanimate/reanimate/blob/master/stack-lts-14.yaml

ksqsf:
Rebased and updated accordingly. Please review. :3

chipbuster (approved):
:+1: LGTM. Thanks @ksqsf for such a quick response on the review!

andytom:
Thank you for your contribution @ksqsf and thanks for reviewing @chipbuster.

PR #3590, opened 2022-02-10
feat: Add a Windows application manifest

### LINKED ISSUE #3589: Add a Windows application manifest
starship doesn't include a Windows application manifest, which triggers all sorts compatibility behavior by Windows which can possibly slow down process startup (And which, regrettably, don't seem to be well documented, at least I wasn't able to find a clear documentation of them).

Just add a manifest with the wanted attributes. Sadly, this isn't yet handled by Rust automatically. (And even MSVC still doesn't have UI to add the `supportedOS` stuff, it only adds the UAC attribute).

It can also be included as a separate file next to the executable, but an embedded manifest is still cleaner IMHO.

In addition, we can consider if we need [activeCodePage](https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#activecodepage) and/or [longPathAware](https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests#longPathAware). Though I think Rust stdio already handles console output encoding so `activeCodePage` might not have any effect, and `longPathAware` kinda depends on if the shell uses it or not for it to have any meaning. (It could be passing/using literal `\\?\` paths)

This can also be used to add `VERSIONINFO` but there might be an easier way to generate that. Or an icon, which isn't really important for a CLI app.

https://docs.microsoft.com/en-us/windows/win32/sbscs/application-manifests

segevfiner:
#### Description
Add a Windows application manifest using [`embed-resource`](https://crates.io/crates/embed-resource). This should disable any Windows compatibility shims/switchback which could possibly slow down process startup, and also UAC virtualization for 32-bit, if anyone is still on 32-bit Windows that is.

#### Motivation and Context

Closes #3589

#### Checklist:

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,5 @@
 fn main() -> shadow_rs::SdResult<()> {
-    shadow_rs::new()
+    shadow_rs::new()?;
+    embed_resource::compile("starship-res.rc");
+    Ok(())
 }
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,13 @@
-fn main() -> shadow_rs::SdResult<()> {
-    shadow_rs::new()?;
-    embed_resource::compile("starship-res.rc");
+use std::error::Error;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    shadow_rs::new().map_err(|err| err.to_string())?;
+
+    if cfg!(target_os = "windows") {
+        let mut res = winres::WindowsResource::new();
+        res.set_manifest_file("starship.exe.manifest");
+        res.compile()?;
+    }
+
     Ok(())
 }
--- a/build.rs
+++ b/build.rs
@@ -3,7 +3,8 @@ use std::error::Error;
 fn main() -> Result<(), Box<dyn Error>> {
     shadow_rs::new().map_err(|err| err.to_string())?;
 
-    if cfg!(target_os = "windows") {
+    #[cfg(windows)]
+    {
         let mut res = winres::WindowsResource::new();
         res.set_manifest_file("starship.exe.manifest");
         res.compile()?;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -507,17 +507,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e78d4f1cc4ae33bbfc157ed5d5a5ef3bc29227303d595861deb238fcec4e9457"
 
 [[package]]
-name = "embed-resource"
-version = "1.6.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


davidkna (commented):
Please also add `longPathAware`. I don't think `activeCodePage` should be added at this time.

segevfiner:
@davidkna Added

segevfiner:
P.S. Doesn't look like `pwsh` actually enabled `longPathAware`. It might support and work with long paths regardless via the PSProvider path handling though, and it might enable it in the future, see https://github.com/PowerShell/PowerShell/blob/master/assets/pwsh.manifest.

build.rs
```diff
@@ -1,3 +1,5 @@
 fn main() -> shadow_rs::SdResult<()> {
-    shadow_rs::new()
+    shadow_rs::new()?;
+    embed_resource::compile("starship-res.rc");
```
davidkna, line 3: The issue in the `embed-resource` crate regarding spurious build failures doesn't sound very good.

Something like this snippet should suffice without `embed-resource`:
```suggestion
    #[cfg(target_env = "msvc")]
    {
        let linker_flags = ["/MANIFEST:EMBED", "/MANIFESTINPUT:starship.exe.manifest"];
        for flag in linker_flags {
            println!("cargo:rustc-link-arg={}", flag);
        }
    }
```

segevfiner: Only if you are not going to ever build with GCC or cross compile for Windows. Which might be fine here. We can also use `/MANIFESTUAC` instead of specifying it in the input manifest in this case.

davidkna: I think some people were using the GNU toolchain, so it would be better to support `windows-gnu`. Supporting it will require scripting `windres` which shouldn't be too hard, most of the complexity in other tools seems to be in locating MSVC `rc.exe`.

Instead of custom code, [windres](https://crates.io/crates/windres) also looks like a good alternative to embed-resource, but it's much less popular. [embed-manifest](https://crates.io/crates/embed-manifest) takes an approach that matches my snippet for MSVC, but the code for `windows-gnu` is scary.

segevfiner: Tried to see if I can just find the issue in `embed-resource`, I might have found it, or I might not have: https://github.com/nabijaczleweli/rust-embed-resource/issues/20#issuecomment-1040254856

To truly be sure I would need to actually reproduce the error...

davidkna: I think I would still prefer the `windres` crate. It would be nice if rust shipped `llvm-rc` to avoid all this.

davidkna: Also,  it would be nice to `cfg(windows)` the build dependency.

davidkna:
Thanks for the contribution!

PR #3591, opened 2022-02-10
fix: Respect $GIT_CEILING_DIRECTORIES

coolreader18:
#### Description

Switches `Context::get_repo()` from using `git2::Repository::discover()` to `::open_ext()`

#### Motivation and Context

Starship in a specific environment (my university's computers, which have network-mounted user dirs) was taking around 180ms to produce a prompt, all of which was in the directory module. When I cloned the starship git repo to investigate, I noticed that it was taking much less time inside the repo then in my home directory, and after a bit discovered the [`$GIT_CEILING_DIRECTORIES` env var](https://git-scm.com/docs/git#Documentation/git.txt-codeGITCEILINGDIRECTORIEScode), which was exactly what I was looking for, though `git_repository_discover` doesn't pay attention to the variable. So, switching to `open_ext` fixes it. AFAICT they're basically the same function, except `open_ext` obviously respects more environment variables.

#### Checklist:


davidkna:
Does this new function also handle the `Repository::open_from_env()` case?

coolreader18:
No, providing the path overrides GIT_DIR, but I keep the check for GIT_DIR so it's still the same behavior there.

PR #3480, opened 2022-01-21
refactor: Improve compatibility with Zsh prompt theme system

marlonrichert:
Zsh's `promptinit` expects a theme's hook functions to be named `prompt_<theme>_<hook>`. See https://github.com/zsh-users/zsh/blob/2876c25a28b8052d6683027998cc118fc9b50157/Functions/Prompts/promptinit#L155

#### Description

* Rename `starship_precmd` to `prompt_starship_precmd`.
* Rename `starship_preexec` to `prompt_starship_preexec`.

#### Motivation and Context

Adhering to this de facto standard makes it easier for `promptinit` and other Zsh software following these conventions (such as [Znap's `prompt` command](https://github.com/marlonrichert/zsh-snap/issues/163)) to interact correctly with Starship.

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

I started Zsh with the new output of `starship init zsh --print-full-init` and it works the same as the old one.

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

This does not cause any changes in documentation or tests.

PR #3593, opened 2022-02-10
fix: ignore empty `--jobs` argument

davidkna:
#### Description

In some instances, shell managers might prevent shell hooks from running before the first prompt draw, causing the starship CLI args like `--jobs=` to be empty and fail at argument parsing. To improve robustness, treat empty `--jobs` arguments as 0. This PR handles this with a new `parse_trim` function used for both `--jobs` and the prior `--cmd-duration` parsing.

#### Motivation and Context

Related #3480

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -18,6 +18,7 @@ use std::fs;
 use std::marker::PhantomData;
 use std::num::ParseIntError;
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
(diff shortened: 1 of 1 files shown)


andytom (approved):

moltar:
Does this fix:

> error: Invalid value for '--jobs <JOBS>': cannot parse integer from empty string

?

PR #3604, opened 2022-02-12
feat(zsh): add right continuation prompt (via RPROMPT2)

aaschmid:
Thanks for the this great prompt :-)

#### Description

Besides left / normal continuation prompt, `zsh` also provides right continuation prompt. This PR makes it configurable via starship.

#### Motivation and Context

This makes it no longer necessary to configure continuation prompt settings on different places, e.g. left on in starship.toml and right one via `RPROMPT2` environment variable (at least for me).

* Added `right_continuation_prompt = " [%_](fg:bright-blue)[─](fg:bright-black)"` to my `starship.toml`
  * (afterwards my installed `starship` correctly logged `[WARN] - (starship::configs::starship_root): Unknown config key 'right_continuation_prompt'`)
* I built the forked `starship` containing the new commit on my Mac and manually set `export RPROMPT2="$(./target/x86_64-apple-darwin/debug/starship prompt --right --continuation)"`.
* Afterwards I entered `"` on the terminal in order to show the continuation prompt on my `zsh`
* The correctly set right continuation prompt was shown

#### Checklist:

Please don't hesitate to give me any feedback on this. This is my first PR on this repository and even though I am an experienced software developer, I have not a lot of experience with Rust yet.

#### P.S.:
There are even tests failing locally however I ignored them as they also fail without having changes (= using `master`). Concrete tests failed:
* modules::git_status::tests::shows_diverged
* modules::git_status::tests::shows_diverged_with_count


chipbuster:
@aaschmid Two questions:

1. Are there any other shells that support this feature?
2. If someone wants to revert to the default variable, how would they do that?

aaschmid:
1. I don't know and as I wrote above, I just familiar with `zsh`
2. No, but one can use `%_` or `%^` which are the defaults. See also https://github.com/starship/starship/issues/3546, where the same question arises for `PROMPT2`

aaschmid:
@chipbuster any more questions or feedback here?

matchai:
I'm afraid this feature may be too niche to justify adding to Starship since it only works for one shell.

aaschmid:
Ok, thanks for the quick answer. I am fine with this if this is your decision. Maybe it makes sense to document this principle somewhere. 

I will close this PR!

aaschmid:
Workaround:       
```
      # in addition to starships `continuation_prompt` (however coloring does not work)
      RPROMPT2 = " %{$fg_bold[black]%}[%{$fg_no_bold[blue]%}%^%{$fg_bold[black]%}]─%{$reset_color%}";
```

aaschmid:
@matchai Re-open the ticket because I see that you add support for things working only on one operating system like `winget` for `PowerShell` but not want to add this addition which (currently) works only for `zsh`. Have you changed your strategy or what did I miss in my understanding?

PR #3612, opened 2022-02-14
docs: Rename Heading localip -> Local IP

pe:
I misread the lowercase l as uppercase I because all other headings are uppercase.


andytom (approved):

andytom:
Thank you for the quick fix @pe.

PR #3613, opened 2022-02-14
docs: "Local IP" chapter in alphabetical order

pe:
The chapter is currently sorted in the wrong place.


chipbuster (approved):
Thanks for the catch.

davidkna:
Thanks, @pe and @chipbuster!

PR #3629, opened 2022-02-19
ci: deny warnings, useful env vars, do not cache deploy

davidkna:
#### Description

I noticed #3623 was green despite deprecation warnings from the rust compiler. This PR sets the `RUSTFLAGS` env var to `-Dwarnings` to make this fail in CI.
I also added some extra useful env vars used by some other rust projects to enable backtraces, increase cargo/rustup network retries and disable incremental compilation (less useful in a CI context, the cache action also sets this).

I did not add `-Dwarnings` to the deployment workflow to prevent it from breaking it. Furthermore, I also removed the cache action from `deploy.yml` because it is less useful in release builds, and it was still using the old cache action.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,11 @@ on:
   push:
     tags:
       - "v*"
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_NET_RETRY: 10
+  RUST_BACKTRACE: short
+  RUSTUP_MAX_RETRIES: 10
 
 jobs:
   # Build sources for every OS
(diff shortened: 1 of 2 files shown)


andytom (commented):
This seems like the right thing to do but the docker context module seems to have some issues.

davidkna:
@andytom That warning is nightly-only and I want to avoid including potentially less stable warnings from nightly. Because there doesn't seem to be a nice cross-platform way to reset `RUSTFLAGS` I've disabled `-D warnings` completely for the `test` matrix. I think the `clippy` task should still catch any `test`-target warnings.

andytom (approved):

davidkna:
Thanks for the review @andytom

PR #3631, opened 2022-02-20
feat: Add a module for C projects

DrHyde:
#### Description

Adds a module to detect and handle C language projects

#### Motivation and Context

There's no existing module for C, and surely _everyone_ has C projects checked out, even if just to read the source!

Screenshot 2022-02-19 at 23 49 41

- [X] I have tested using **FreeBSD**

I have not yet tested with Windows, as I don't have a Windows box.

#### Checklist:


src/context.rs
```diff
@@ -335,6 +335,23 @@ impl<'a> Context<'a> {
             Duration::from_millis(self.root_config.command_timeout),
         )
     }
+
+    /// Attempt to execute several commands with exec_cmd, return the results of the first that works
+    pub fn exec_cmds_return_first (&self, commands: &[&[&str]]) -> Option<CommandOutput> {
+        let mut attempts_iterator = commands.iter();
```
davidkna, line 341: Please try [`find_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map)

src/context.rs
```diff
@@ -335,6 +335,23 @@ impl<'a> Context<'a> {
...
+        loop {
+            let attempt = attempts_iterator.next();
+            if attempt.is_none() {
+                break None;
+            }
+            let attempt = attempt.unwrap();
+            // this should be generalised to pass more args if they exist
+            let c_compiler_info = self.exec_cmd(attempt[0], &[attempt[1]]);
```
davidkna, line 349: This might work
```suggestion
            let c_compiler_info = self.exec_cmd(attempt[0], attempt[1..]);
```

chipbuster:
@DrHyde It looks like you'll also need to grab the formatting tools. You can find the instructions hidden away (seriously, we do need to make those more prominent) over at https://github.com/starship/starship/blob/master/CONTRIBUTING.md#formatting

src/context.rs
DrHyde: `&attempt[1..]` - close enough. Thanks!

src/context.rs
DrHyde: Ah-ha, I knew it had to be _somewhere_ in the standard libraries.

src/context.rs
```diff
@@ -335,6 +335,18 @@ impl<'a> Context<'a> {
...
+
+    /// Attempt to execute several commands with exec_cmd, return the results of the first that works
+    pub fn exec_cmds_return_first(&self, commands: &[&[&str]]) -> Option<CommandOutput> {
+        commands.iter().find_map(|attempt| {
+            let c_compiler_info = self.exec_cmd(attempt[0], &attempt[1..]);
+            if c_compiler_info.is_some() {
+                c_compiler_info
+            } else {
+                None
+            }
```
davidkna, line 347: Even shorter.
```suggestion
            self.exec_cmd(attempt[0], &attempt[1..])
```

src/modules/c.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    #[test]
+    fn folder_with_c_file() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        File::create(dir.path().join("any.c"))?.sync_all()?;
+
+        // FIXME voodoo here to set up a mock for 'cc --version' claiming to be gcc
+        // NB the test for when it claims to be clang is in the ..._with_h_file test
+        let actual = ModuleRenderer::new("c").path(dir.path()).collect();
```
davidkna, line 131: `ModuleRenderer` has `cmd` for dynamic mocking.

davidkna: Usage: https://github.com/starship/starship/blob/master/src/modules/ocaml.rs#L314-L332

DrHyde: Ah that looks useful, thanks!

DrHyde:
Not sure why clippy is grumbling about things I haven't changed

chipbuster:
Rust 1.59.0 was released earlier today, and our CI is set to use the latest stable toolchains, so that might be what you're seeing.

David has a PR at #3656 which should fix those issues. If you'd like, you can rebase on top of that branch.

DrHyde:
Oh, that makes sense

src/modules/c.rs
```diff
@@ -0,0 +1,222 @@
...
+                // the compiler is usually cc, and --version works on gcc and clang
+                &["cc", "--version"],
+                // but on some platforms gcc is installed as *gcc*, not cc
+                &["gcc", "--version"],
+                // for completeness, although I've never seen a clang that wasn't cc
+                &["clang", "--version"],
+            ])
+        } else {
+            None
+        };
```
davidkna, line 36: Please use `Lazy::new` for this, like in the nodejs module.

src/modules/c.rs
```diff
@@ -0,0 +1,222 @@
...
+                    } else if c_compiler_info.contains("Free Software Foundation") {
+                        "gcc"
+                    } else {
+                        "Unknown compiler"
+                    };
+                    Some(c_compiler).map(Ok)
+                }
+                _ => None,
+            })
+            .map(|variable| match variable {
```
davidkna, line 62: These should share `map`:
```suggestion
```

DrHyde: I tried that, but ...

```
error[E0308]: mismatched types
  --> src/modules/c.rs:79:25
   |
79 | /                         VersionFormatter::format_module_version(
80 | |                             module.get_name(),
81 | |                             c_compiler_info.split_whitespace().nth(3)?,
82 | |                             config.version_format,
83 | |                         )
84 | |                         .map(Ok)
   | |________________________________^ expected `&str`, found struct `std::string::String`
   |
   = note: expected enum `std::option::Option<Result<&str, string_formatter::StringFormatterError>>`
              found enum `std::option::Option<Result<std::string::String, _>>`
```

and I couldn't figure out how to resolve it.

davidkna: They both need to have the same type (`as_str` or `as_deref` on the `String` might work). If that does not work, you could try using [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html).

DrHyde: With this code ...

```
+                        let formatted: Cow<String> = Cow::Owned(
+                            VersionFormatter::format_module_version(
+                                module.get_name(),
+                                c_compiler_info.split_whitespace().nth(3)?,
+                                config.version_format,
+                            )?
+                        );
+                        Some(formatted.into_owned().as_str()).map(Ok)
```

The type problem seems to be fixed, but now the compiler complains about ...

```
error[E0515]: cannot return value referencing temporary value
  --> src/modules/c.rs:84:25
   |
84 |                         Some(formatted.into_owned().as_str()).map(Ok)
   |                         ^^^^^----------------------^^^^^^^^^^^^^^^^^^
   |                         |    |
   |                         |    temporary value created here
   |                         returns a value referencing data owned by the current function
```

Unless I've misunderstood the doco for Cow, that's exactly what `into_owned` should be doing. What am I missing?

davidkna: Cow allows you to return both Borrowed and Owned Values from the same closure. The borrowed valued you create here don't have sufficient lifetimes. At the very least, I would expect `formatted.into_owned()` to require a variable binding.

But the fix here is much simpler. Just add `.map(Cow::Borrowed)` after `Some(compiler_name)`  and `.map(Cow::Owned)` after the `VersionFormatter` call.

DrHyde: Thanks!

src/modules/c.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    let parsed = StringFormatter::new(config.format).and_then(|formatter| {
+        let c_compiler_info = Lazy::new(|| {
+            context.exec_cmds_return_first(&[
+                // the compiler is usually cc, and --version works on gcc and clang
+                &["cc", "--version"],
+                // but on some platforms gcc is installed as *gcc*, not cc
+                &["gcc", "--version"],
+                // for completeness, although I've never seen a clang that wasn't cc
+                &["clang", "--version"],
```
davidkna, line 34: I think it might be nice to make this configurable. Some people might use compiler variants like `zig cc` and `arm-none-eabi-gcc`.

src/modules/c.rs
```diff
@@ -0,0 +1,220 @@
...
+            .map(|variable| match variable {
+                "compiler_name" => {
+                    let c_compiler_info = &c_compiler_info.deref().as_ref()?.stdout;
+
+                    let c_compiler = if c_compiler_info.contains("clang") {
+                        "clang"
+                    } else if c_compiler_info.contains("Free Software Foundation") {
+                        "gcc"
+                    } else {
+                        "Unknown compiler"
```
davidkna, line 56: I would prefer `None` in the case of an unknown compiler.

src/modules/c.rs
```diff
@@ -0,0 +1,220 @@
...
+                    };
+                    Some(c_compiler).map(Cow::Borrowed).map(Ok)
+                }
+                "compiler_version" => {
+                    let c_compiler_info = &c_compiler_info.deref().as_ref()?.stdout;
+                    if config.format.contains("$compiler_version")
```
davidkna, line 62: Please remove this line. For instance, `${compiler_version}` is also valid, but would be skipped by this code. If `$compiler_version` is matched in the `match`, it is in the format string.

src/modules/c.rs
```diff
@@ -0,0 +1,220 @@
...
+                        && (c_compiler_info.contains("clang")
+                            || c_compiler_info.contains("Free Software Foundation"))
```
davidkna, line 64: If basically all -nix C compilers follow this convention, it might be better not to gate it like this.

davidkna (changes requested):
Per the module checklist in `CONTRIBUTIN.md`, please also update the presets and add the module to the "Default Prompt Format" section,

src/modules/c.rs
DrHyde: It would be nice, but I'm not sure how practical it is to make it entirely configurable. We'd need to have a way of specifying what command to run, with what arguments, what to look for in the output, and possibly how to parse the version out of that.

src/modules/c.rs
DrHyde: I was undecided about that myself, so happy to go with that.

src/modules/c.rs
DrHyde: Ah, that's a leftover from my previous code that was looking for `$compiler_name` as well.

src/modules/c.rs
DrHyde: I wouldn't want to make that assumption, I think it's more a happy accident that clang and gcc match.

DrHyde: This prompted me to look at a really old version of gcc. It doesn't fit the pattern. Bother, I'm going to need to fix that:

```
$ cc --version
cc (GCC) 3.3.5 (Debian 1:3.3.5-13)
Copyright (C) 2003 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

DrHyde: Fixed by now looking for the first whitespace-seperated word that looks like valid semver. Which *should* also work for a lot more compilers, so now I've also removed that condition.

davidkna (changes requested):
I have some more suggestions, but overall this is in a good shape now.

There's also been a rework of the preset system, please rebase on/merge master.

src/modules/c.rs
```diff
@@ -0,0 +1,216 @@
...
+                        // so again in position 3.
+                        // But this is CRUFTY AS HELL
+                        VersionFormatter::format_module_version(
+                            module.get_name(),
+                            c_compiler_info.split_whitespace().nth(3)?,
+                            config.version_format,
+                        )
+                        .map(Cow::Owned)
+                        .map(Ok)
+                    } else {
```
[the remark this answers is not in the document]
DrHyde: Nah. I'm happy with just not doing anything if we don't recognise the compiler. Any user who cares will notice pretty darned quickly that their compiler isn't recognised and the solution, whether they notice because it's not in the prompt or because they notice an error message, is the same - raise a PR (or an issue). I'll put a note in the docs about this, and I'd be happy to deal with such issues.

davidkna (changes requested):

docs/config/README.md
```diff
@@ -450,6 +451,69 @@ discharging_symbol = "💦"
...
+
+The `c` module shows some information about your C compiler. By default
+the module will be shown if the current directory contains a `.c` or `.h`
+file.
+
+### Options
+
+| Option              | Default                            | Description                                                               |
+| ------------------- | ---------------------------------- | ------------------------------------------------------------------------- |
+| `format`            | `"via [$symbol ($name )]($style)"` | The format string for the module.                                         |
```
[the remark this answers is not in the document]
DrHyde: If we're going to show one of the compiler name or version, the name is much more useful. If it says "C v1.2.3" what is that version 1.2.3 of?

docs/config/README.md
```diff
@@ -450,6 +451,69 @@ discharging_symbol = "💦"
...
+| `version_format`    | `"v${raw}"`                        | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
+| `symbol`            | `""`                              | The symbol used before displaying the compiler details                    |
+| `detect_extensions` | `["c", "h"]`                       | Which extensions should trigger this module.                              |
+| `detect_files`      | `[]`                               | Which filenames should trigger this module.                               |
+| `detect_folders`    | `[]`                               | Which folders should trigger this module.                                 |
+| `detectors`         | see details below                  | How to detect what the compiler is                                        |
```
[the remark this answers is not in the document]
DrHyde: I thought `detectors` worked best with the existing multitude of `detect*` options

DrHyde:
The test failure appears to be unrelated

docs/config/README.md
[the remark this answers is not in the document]
davidkna: I think you can usually tell from the version number. As is, `$symbol` and `$name` feel repetitive, and all the other versions do print versions. To make it feel less repetitive, maybe something like the rust module `via [$symbol $version(-$name) ]($style)`?

docs/config/README.md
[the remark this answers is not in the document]
davidkna: I think that doesn't quite match the actual usage of the option.

docs/config/README.md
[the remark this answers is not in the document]
DrHyde: Hmmm, I suppose we're likely to have a version by virtue of looking for something semver-ish, and _normal_ people only have one compiler installed and therefore know what it means :-)

docs/config/README.md
```diff
@@ -495,6 +496,54 @@ The `buf` module shows the currently installed version of [Buf](https://buf.buil
...
+
+The `c` module shows some information about your C compiler. By default
+the module will be shown if the current directory contains a `.c` or `.h`
+file.
+
+### Options
+
+| Option              | Default                                                                     | Description                                                               |
+| ------------------- | --------------------------------------------------------------------------- | ------------------------------------------------------------------------- |
+| `format`            | `"via [$symbol $version(-$name )]($style)"`                                 | The format string for the module.                                         |
```
davidkna, line 509: I just noticed the symbol is supposed to contain the space in all modules, not the main`format`. If you fix all the `symbol`/`format` occurrences, this PR looks good to me.

docs/config/README.md
```diff
@@ -495,6 +496,54 @@ The `buf` module shows the currently installed version of [Buf](https://buf.buil
...
+
+The `c` module shows some information about your C compiler. By default
+the module will be shown if the current directory contains a `.c` or `.h`
+file.
+
+### Options
+
+| Option              | Default                                                                     | Description                                                               |
+| ------------------- | --------------------------------------------------------------------------- | ------------------------------------------------------------------------- |
+| `format`            | `"via [$symbol$version(-$name )]($style)"`                                  | The format string for the module.                                         |
```
davidkna, line 509: I'm sorry to bother you with this again, but I noticed another issue with the default format string. If `$name` is missing, there won't be a space after the module. Please change it to `"via [$symbol($version(-$name) )]($style)"`.

docs/.vuepress/public/presets/toml/bracketed-segments.toml
```diff
@@ -1,6 +1,9 @@
 [aws]
 format = '\[[$symbol($profile)(\($region\))(\[$duration\])]($style)\]'
 
+[c]
+format = '\[[$symbol $version(-$name)]($style)\]'
```
davidkna, line 5: Please also carry over the changes into the bracketed segments presets.

davidkna (approved):

matchai:
Thank you for taking the time to implement C support, @DrHyde, and thank you @davidkna for the review! 😄

PR #3654, opened 2022-02-24
feat(directory): Allows to highlight the last component of the pwd

### LINKED ISSUE #3653: Allows highlighting of the last component of the pwd.
## Feature Request

I used to be a user of p10k, and I'm glad I got to know starship, I love it and decided to use it instead of p10k, but after some study I found that the directory module's path display doesn't support highlighting the last component. I think this feature could be added, because highlighting the last component would make it more visible where you are currently roaming in the directory tree.

BTW: Of course, p10k is also excellent, except that it does not support fish.

Add a configuration item to allow adding a unique style setting for the last component of the directory.

Or you could set it to the same style as the repository root for simplicity purposes.

flw-cn:
sell also #3653

#### Description

* `src/modules/directory.rs`
* `docs/config/README.md`

added two variables, `$last_component` and `$path_base` in module `directory`.

#### Motivation and Context

This PR allows to highlight the last component of the pwd.
It can close the gap between starship and p10k at this point.

Closes #3653

#### Checklist:

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -99,7 +99,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         String::from("")
     };
 
-    let path_vec = match &repo.and_then(|r| r.workdir.as_ref()) {
+    const BEFORE_ROOT_PATH: usize = 0;
+    const REPO_ROOT: usize = 1;
+    const PATH: usize = 2;
+    const PATH_BASE: usize = 3;
+    const LAST_COMPONENT: usize = 4;
+
+    let mut path_vec = [0; 5].map(|_| String::new());
+
+    match &repo.and_then(|r| r.workdir.as_ref()) {
         Some(repo_root) if config.repo_root_style.is_some() => {
             let contracted_path = contract_repo_path(display_dir, repo_root)?;
             let repo_path_vec: Vec<&str> = contracted_path.split('/').collect();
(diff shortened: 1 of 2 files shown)


flw-cn:
Can anyone tell me what I should do to get this PR merged?

src/modules/directory.rs
```diff
@@ -99,7 +99,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         String::from("")
     };
 
-    let path_vec = match &repo.and_then(|r| r.workdir.as_ref()) {
+    const BEFORE_ROOT_PATH: usize = 0;
```
chipbuster, line 102: After looking at the code, I think I'd prefer it if this got factored into a struct with appropriate `impl`s on it. While this works, it gets very hard to read (e.g. I still can't figure out what the conditional on lines 142-146 is supposed to be doing).

src/modules/directory.rs
```diff
@@ -109,22 +117,46 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                path_vec[PATH] = prefix + &dir_string;
             }
         }
-        _ => ["".to_string(), "".to_string(), prefix + &dir_string],
+        _ => {
+            path_vec[PATH] = prefix + &dir_string;
+        }
     };
 
+    (path_vec[PATH_BASE], path_vec[LAST_COMPONENT]) =
```
chipbuster, line 132: I think this would better be done with std methods like `file_name` and `parent`.

src/modules/directory.rs
```diff
@@ -109,22 +117,46 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+        path_vec[PATH]
+            .rfind('/')
+            .map_or(("".to_string(), path_vec[PATH][..].to_string()), |pos| {
+                (
+                    path_vec[PATH][..pos].to_string(),
+                    path_vec[PATH][pos + 1..].to_string(),
+                )
+            });
+
+    if path_vec[LAST_COMPONENT] != home_symbol
```
chipbuster, line 142: Would be nice to have this factored out into a named function (or an `impl` on a struct which contains this information) because I can't work out what this is supposed to do.

chipbuster (commented):
Hi @flw-cn, sorry for the delay on the review.

Right now, the biggest blocker to merging is that when I check out your commit in git locally, the parent information is invalid. Git thinks that this commit is its own parent, which makes it impossible to do operations like merge/rebase or diffs. This will definitely need to be fixed before we can merge this, because right now, I can't tell if the automated checks are working correctly due to this commit having no parent and no shared history with the rest of starship.

TBH, I'm not entirely sure how to fix this--you might be able to cherry-pick these commits onto the latest copy of master, or you might have to make a patch and apply it over the top, since git is doing something weird. Once that's fixed, I can take another look at this.

(I've also left some review comments down below, but the biggest blocker right now is this git history issue).

src/modules/directory.rs
chipbuster: I forgot the review doesn't apply to blocks: when I say "this", I'm referring to the usage of `const` indices + a vec to store things in to form a simple dictionary. I would prefer if we could just do something like `components.before_root_path`.

flw-cn:
Thank you for the serious review. I think I have understood what you are saying. I will refactor my code as soon as possible and rebase it to the latest master.

PR #3660, opened 2022-02-25
docs: Mention set -o functrace in the text

joliss:
This tripped me up because I only read the text and didn't read the code sample.

This PR is a follow-on to #1526. The same text occurs in several of the translations, but I wasn't sure if I needed to update those.


chipbuster:
Given that #3749 has revealed that `set -o functrace` breaks the init script with certain debug functions, we might need to put this on pause while we think about how to add the hooks properly.

PR #3667, opened 2022-02-27
perf(rayon): restrict thread count

davidkna:
#### Description

Currently, [starship will spawn a thread for each logical cpu core](https://docs.rs/rayon/latest/rayon/struct.ThreadPoolBuilder.html#method.num_threads). On systems with high core counts, this can be expensive.

Thus, limit the maximum core count to 8. This is still cheap if there is nothing to do, but it doesn't seem to performance when a high number of modules are enabled.

Right now, different thread counts can also be set with the `RAYON_NUM_THREADS` variable.

#### Motivation and Context

Related #3180

#### Checklist:

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,6 +2,7 @@
 
 use clap::crate_authors;
 use std::io;
+use std::thread::available_parallelism;
 use std::time::SystemTime;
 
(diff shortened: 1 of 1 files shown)


matchai (approved):
Looks good to me! 👍

PR #1829, opened 2020-10-24
docs: Add a new content "Gallery" to the documentation

nukopy:
Hi, thanks for excellent tool, Starship:rocket:!!
I implemented the feature to show examples of Starship prompt settings, "Gallery".

#### Description

Add a new content "Gallery" to the documentation.
By this feature, current/new Starship users can easily see, reference, and mimic other users prompt setting like below:

image

Users can:

- See other users prompts as images
- See source code of Starship settings file `starship.toml` on GitHub
- Share own settings to other users(**see `CONTRIBUTING.md`**)
- Enjoy other users' interesting prompts:sunglasses:

#### Motivation and Context

Starship's documentation is great, BUT this is short of examples.

#### Implement

To auto-generate "Gallery", use Vue component.
My `starship.toml`s & screenshots is put in `docs/.vuepress/public/gallery` as mock data, so PR is big but this should be smaller when merge.

I'm newer to Vue.js, so plz tell me anything if you have any ideas about implement.

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #1862, opened 2020-11-09
Translated requirements.

emorozov:
Translated requirements in the Russian docs.

PR #3683, opened 2022-03-03
docs: Update Preset system with several additional features

### LINKED ISSUE #3580: Extract presets from markdown into separate TOML files
I want easy access to starship presets without having to manually copy them from the website. Right now the [starship presets docs](https://github.com/starship/starship/blob/e6a921b5b2a54ffd81368f9b3a573e49b5af1c9a/docs/presets/README.md) just include the preset TOML configs inline, which means I can't just grab them out of the repo either. It also means the presets can have syntax errors introduced that won't be automatically caught (this has happened before).

The presets should be extracted into individual TOML files and then the static site generator should inline them back into the page when rendering. I don't know anything about Vuepress but I imagine it's capable of doing this.

By having the files be separate, I can then just copy them out of the source repo myself and use local tooling to merge my custom config on top of the presets I want.

Ideally these preset TOML files would be listed in the `Cargo.toml` `include` field as well, so that way I get them when I download the published package in case I want to do anything with that (for example, right now Nix just fetches starship from GitHub, but I'm not sure if there's anything stopping it from switching over to downloading the crates.io package instead).

#3039 asks for the ability to build a layered config file via imports. The direct request is orthogonal to presets, but in https://github.com/starship/starship/issues/3039#issuecomment-913846919 I talk about using this to let me import presets as well.

Also in https://github.com/starship/starship/issues/3039#issuecomment-913846919 I suggest a hypothetical `starship print-config --preset` command that lets me list the presets and print each one by name. This would be a good way of letting me build my config without having to grab preset files from the source directly. And it can be done without doing the rest of #3039. With a `starship print-config --preset` command I can use my local configuration (Nix) to synthesize my config file easily.

chipbuster:
#### Description
The original motivation of this PR was to close #3580. However, in order to do so, several sweeping changes had to be made to the documentation organization, and it ultimately became easier to do a mini-rework.

A full changelist (with screenshots, where appropriate) can be found at the end of this PR.

#### Motivation and Context

Closes #3580

Tested locally with `npm run dev`

#### Checklist:

#### Changelist

- Factored all preset TOMLs out into their own files to allow for the request on #3580
- Removed everything but a short description and screenshot from the main Presets page: this makes the feel much more like that of #1829. IMO this is much more pleasant to browse.

image

- Replaced all screenshots with a semi-standardized set of screenshots. Note: I'm not 100% sure about this change: currently the screenshots all look very similar to each other, but I don't know how to fix this. Maybe it's because the terminal background is too similar to the default Vuepress grey?

- Presets have been factored out into their own pages, that have a screenshot, the full config inlined via a [code snippet import](https://vuepress.vuejs.org/guide/markdown.html#import-code-snippets), prerequisites (if any), and a download link for the TOML.

image

- Modified Cargo.toml to no longer include translated READMEs (in accordance with the comment above the `include` line) and include all preset TOML files.

- Added the Python symbol to the NerdFont preset, which was previously missing (though now that I look at the screenshots again, I can see that the Rust symbol is also missing...dammit).


chipbuster:
I really liked the aesthetic of #1862 (it inspired the screenshot-only setup used in this PR), but I'm not convinced about the current JS implementation: not only do you have to be able to hover the image (impossible on mobile), but you can't obtain the links from the browser due to the implementation. I also have concerns about accessibility for that design: it works wonderfully for mouse-and-keyboard, but I'm concerned about how it works if you use a screen reader, or don't have a mouse. I did not have the time to alter the implementation.

Additionally, because that PR was designed to point to external preset links hosted on GitHub instead of to stuff hosted within the starship repo, it would have been difficult to make the TOML files accessible from the Cargo package without compile-time web-scraping.

However, since I did basically yoink all my design ideas from that PR, I would be happy to make @nukopy a co-author of this PR, if they would like to do that.

davidkna:
> The design works wonderfully for mouse-and-keyboard, but I'm concerned about how it works if you use a screen reader, or don't have a mouse.

You could put a link on the headings, `title=`  could help and a button next to the headings might look good.

> Replaced all screenshots with a semi-standardized set of screenshots. Note: I'm not 100% sure about this change: currently the screenshots all look very similar to each other, but I don't know how to fix this. Maybe it's because the terminal background is too similar to the default Vuepress grey?

For standardizing the screenshots, I looked at how OhMyPosh is handling the screenshots for their [high number of themes](https://ohmyposh.dev/docs/themes), and they use a fork of [termshot](https://github.com/homeport/termshot). It could also be useful for testing different screenshot settings.

chipbuster:
Yeah, I think we should probably also go in that direction and start trimming down what our main page themes look like. We can add additional screenshots on the per-theme page if there are features that the original screenshot doesn't show off, but I think right now it's hard to browse.

chipbuster:
Vanilla termshot appears to have difficulties with non-powerline unicode (these should be package, nodejs, python, and crab, respectively):

image

Probably why they opted to fork it. For now, I think I'll stick with what we have here.

davidkna (approved):
LGTM

andytom (approved):
LGTM

matchai (commented):
The images appear to be pretty big, amounting to 3.2MB for the five of them, ~2k pixels in width. Could we resize them to 50% the size and pass them through a compression tool like ImageOptim?

Cargo.toml
```diff
@@ -8,7 +8,7 @@ documentation = "https://starship.rs/guide/"
 edition = "2021"
 homepage = "https://starship.rs"
 # Keep `/` in front of `README.md` to exclude localized readmes
-include = ["src/**/*", "build.rs", "Cross.toml", "LICENSE", "README.md"]
+include = ["src/**/*", "build.rs", "Cross.toml", "LICENSE", "/README.md", "docs/.vuepress/public/presets/toml/"]
```
matchai, line 11: What do we include the preset tomls for? I don't think people installing them with Cargo will know where to find them.

chipbuster: This was the original request of #3580, the motivating issue for this PR. The idea is that users can grab the TOML presets out of the cargo package and merge them in over their current configurations, providing one way for them to stay up to date.

I don't think it's strictly better than a proper preset system, but given that nobody's stepped up to work on it (and I'm likely not going to be able to dedicate any time to it until the end of this year at the earliest) I think this is a reasonable mid-term solution.

chipbuster:
@matchai Resized to ~1200 x ~800 and passed through optipng, resulting size for 5 images is about 1MB.

I'm not sure how this will scale to more presets (esp. given that we currently have 2 additional presets waiting in the wings).

I see three options, though there may be more that I haven't thought of:
  - Shrink the images even further, at the risk of appearing blurry on hi-res screens.
  - Change the image size by e.g. using only a single line in the preview like OMP.
  - Switch to an SVG format.

I don't have a good feel for if the third option will save much more space, and the only tool I can find for doing this (https://github.com/nbedos/termtosvg) seems to be unmaintained as of mid-2020.

matchai:
Nice, sounds like a good improvement for now 👍 

Another fourth option would be to create the terminal previews in HTML/CSS. We already load a NerdFont font on the website to be able to preview the glyphs in the config file. With a little bit of work, we could create a component that converts ANSI output to a styled component.

PR #3690, opened 2022-03-05
refactor(windows): replace winapi with windows

davidkna:
#### Description

#2791 reminded me I wanted to replace `winapi` with `windows` at some point. `windows` has a higher API coverage (this PR removes a manual binding, a number of arguments are now marked as constant), better type safety, built-in error handling facilities, and it is officially backed by Microsoft.

The type safety isn't complete (constant are typed, but most function arguments are just `u32`), so some breaking changes could still happen, but I feel like the crate has already matured somewhat.

The only functional changes are that the exit code of the first `GetFileSecurityW` is verified for `ERROR_INSUFFICIENT_BUFFER` and `STANDARD_RIGHTS_READ` is replaced with `TOKEN_READ_CONTROL` which is supposed to do the same thing but has the expected type (both are accepted arguments). In addition, error messages now include the error value throughout.

#### Motivation and Context

Closes #

I have tested that the following targets still build:
- [x] aarch64-pc-windows-msvc
- [x] i686-pc-windows-msvc
- [x] x86_64-pc-windows-gnu
- [x] x86_64-pc-windows-msvc

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1941,7 +1941,7 @@ dependencies = [
  "urlencoding",
  "versions",
  "which",
- "winapi",
+ "windows 0.35.0",
  "winres",
  "yaml-rust",
 ]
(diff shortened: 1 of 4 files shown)


matchai (approved):
LGTM! 👍

davidkna:
Thanks for the review @matchai.

PR #3692, opened 2022-03-06
feat(gcloud): created a function to validate config name in GCloud module

### LINKED ISSUE #3658: Feature Request: Allow gcloud to be hidden when the active config is NONE

Grubba27:
#### Description

related to this [issue](https://github.com/starship/starship/issues/3658)
_Still draft of PR_ I'm stuck with the last part of the issue, how could I make sure that nothing is sent to the console when in google cloud module config is set to "NONE"

#### Motivation and Context

As said in Issue, when the config is none it should not show up in the console, in this PR I have done the first part that is creating s function to validate when this happens.

Closes #3658 
Closes  #4092

All tests in Google Cloud Module still passes.

- [ x] I have tested using **MacOS**

#### Checklist:

Maybe create tests as the ones that name projects as ABC for testing this new issue ?

### DIFF
--- a/src/modules/gcloud.rs
+++ b/src/modules/gcloud.rs
@@ -106,6 +106,16 @@ fn get_active_config(context: &Context, config_dir: &Path) -> Option<String> {
     })
 }
 
+fn get_active_config_name(config_name: String) -> String {
+    let name = if config_name == "NONE" {
+        log::error!("Error in module `gcloud`: No config ");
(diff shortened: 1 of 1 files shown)


chipbuster:
Usually we disable the module by returning `None` from the `module` function. Maybe you could match on the config name and then return `None` if the config is that value?

PR #3697, opened 2022-03-08
fix(fish): allow generating session keys in older versions of fish

### LINKED ISSUE #3696: Generating session keys for fish fails in Linux
#### Current Behavior
When executing `starship init fish | source` with older fish shell, it produces the following error:
`random: Expected zero or one argument, got 2`

#### Expected Behavior
No errors should be output.

`random` in newer fish shell (version 3.2.1) can generate random integers within the specified range:
```bash
$ random 1 5
4
```

But in older fish shell (version 2.3.1), `random` cannot generate:

```bash
$ random 1 5
random: Expected zero or one argument, got 2
```

#### Possible Solution
For the compatibility, it might be better not to use ranged `random` generation.
https://github.com/starship/starship/blob/master/src/init/starship.fish#L40

#### Environment
- Starship version: 1.3.0
- fish version: fish, version 2.3.1
- Operating system: CentOS 7.9.2009
- Git Commit Hash: e6a921b5
- Branch/Tag: v1.3.0
- Rust channel:  release
- Build Time: 2022-02-07 15:03:09 +00:00

```fish
starship init fish | source
```

#### Starship Configuration

```toml
<unknown config>
```

tttanikawa:
To prevent initialization errors with older fish shell, I modified the way to generate random session keys.

#### Description
`random` in older fish shell cannot generate random integers within the specified range, so executing `starship init fish | source` with fish shell will output some errors.
I modified the init script to generate random integers without range specification.
Session keys will be generated in the manner same as `src/init/starship.bash`.

#### Motivation and Context
Closes #3696 

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -37,6 +37,4 @@ builtin functions -e fish_mode_prompt
 set -gx STARSHIP_SHELL "fish"
 
 # Set up the session key that will be used to store logs
-set STARSHIP_SESSION_KEY (random)(random)(random)(random)(random) # Random generates a number b/w 0 - 32767
-set STARSHIP_SESSION_KEY "$STARSHIP_SESSION_KEY"0000000000000000 # Pad it to 16+ chars.
-set -gx STARSHIP_SESSION_KEY (echo "$STARSHIP_SESSION_KEY" | cut -c 1-16) # Trim to 16-digits if excess.
+set -gx STARSHIP_SESSION_KEY (string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -37,4 +37,6 @@ builtin functions -e fish_mode_prompt
 set -gx STARSHIP_SHELL "fish"
 
 # Set up the session key that will be used to store logs
-set -gx STARSHIP_SESSION_KEY (random 10000000000000 9999999999999999)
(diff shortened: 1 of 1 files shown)


davidkna:
I think the issue is rather that your older version of fish does not support ranged `random`. Does `(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)` from #1904 work in your version?

tttanikawa:
Sorry, it was my mistaken. I think this is a compatibility issue that is present only in older versions of fish.
I tried `(string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)` in my version (2.3.1), and it works well.
Thank you for the suggestion!

davidkna (approved):

src/init/starship.fish
```diff
@@ -37,4 +37,4 @@ builtin functions -e fish_mode_prompt
 set -gx STARSHIP_SHELL "fish"
 
 # Set up the session key that will be used to store logs
-set -gx STARSHIP_SESSION_KEY (random 10000000000000 9999999999999999)
+set -gx STARSHIP_SESSION_KEY (string sub -s1 -l16 (random)(random)(random)(random)(random)0000000000000000)
```
matchai, line 40: Any chance we can add a comment explaining the reason we don't use `random [min] [max]`?

tttanikawa: I added a comment explaining the reason.

andytom:
Thank you for your contribution @tttanikawa.

PR #3699, opened 2022-03-08
feat(aws): Add profile aliases

aes:
#### Description
Add an aws.profile_aliases feature just like aws.region_aliases

#### Motivation and Context
AWS regions might have slightly irritating names, but it's nothing compared to the enterprise naming schemes sometimes applied to team profile names.

Added two simple unit tests, one with and one without similar alias used for region.

#### Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -376,6 +376,46 @@ mod tests {
     }
 
     #[test]
+    fn profile_set_with_alias() {
+        let actual = ModuleRenderer::new("aws")
+            .env("AWS_PROFILE", "CORPORATION-CORP_astronauts_ACCESS_GROUP")
+            .env("AWS_REGION", "ap-northeast-2")
+            .env("AWS_ACCESS_KEY_ID", "dummy")
+            .config(toml::toml! {
+                [aws.profile_aliases]
+                CORPORATION-CORP_astronauts_ACCESS_GROUP = "astro"
+            })
+            .collect();
+        let expected = Some(format!(
+            "on {}",
+            Color::Yellow.bold().paint("☁️  astro (ap-northeast-2) ")
+        ));
+
+        assert_eq!(expected, actual);
+    }
+
+    #[test]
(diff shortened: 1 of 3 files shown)


tkey91:
add profile aliases = @

src/modules/aws.rs
```diff
@@ -203,6 +197,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         })
     };
 
+    let mapped_region = if let Some(aws_region) = aws_region {
```
andytom, line 200: Then this could just be

```rust
let mapped_region = alias_name(aws_region, &config.region_aliases);
```

src/modules/aws.rs
```diff
@@ -203,6 +197,18 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+        Some(alias_name(aws_region, &config.region_aliases))
+    } else {
+        None
+    };
+
+    let mapped_profile = if let Some(aws_profile) = aws_profile {
```
andytom, line 206: and this

```rust
let mapped_profile = alias_name(aws_profile, &config.profile_aliases);
```

andytom (commented):
Can you update the config docs for this module too (you only need to update the English version the others will be updated automatically). Side note it is better if you use the template for the PR description.

aes:
@andytom I've tried to keep the edits down to a minimum, but sure, I can do some of that.

aes:
Completely missed that there's a PR template, I just did `gh pr create -f`.

tkey91:
Aws is Amazon's server right??

matchai:
> Aws is Amazon's server right??

Your comment is not meaningfully contributing to this PR.
Continuing to make non-constructive or offtopic comments will lead to you being blocked from the Starship org's repos.

andytom (approved):
LGTM

davidkna:
Thanks for the contribution @aes, and thanks for reviewing @andytom.

PR #3719, opened 2022-03-09
docs: Specify minimum supported rust version

ahmadjubair33:
Updated the rust version in `Cargo.toml` from `1.4.0` to `1.59` .please review this pr and merge it. Thank you
#3716

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "starship"
-version = "1.4.0"
+version = "1.59"
 authors = ["Starship Contributors"]
 build = "build.rs"
 categories = ["command-line-utilities"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "starship"
-version = "1.59"
+version = "1.4.0"
 authors = ["Starship Contributors"]
 build = "build.rs"
 categories = ["command-line-utilities"]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "starship"
-version = "1.4.0"
+version = "1.59"
```
davidkna, line 3: This sets the version of starship and not the supported rust version. Please also add a comment that the MSRV is intended as a hint and that only the latest version is tested, and that the specified minimum rust version is not guaranteed to work.

ahmadjubair33: Hi @davidkna Can you please the exact location of the rust version where it resides so I can change it.

davidkna: There is no minimum rust version specified in the file. You will need to add it.

ahmadjubair33: In which file?

davidkna: https://github.com/starship/starship/issues/3716 includes a link to the relevant documentation.

ahmadjubair33: Hi @davidkna I have added the rust version can you please review it.

davidkna: Please ensure that `Cargo.toml` is still valid TOML, ensure the formatting (dprint, see the contributing guidelines) passes and add the comment I requested.

PR #3724, opened 2022-03-09
ci: Correct dependencies for deploy workflow

### LINKED ISSUE #3714: Deploy notarized artifacts
## Bug Report

For some reason, the [Deploy action for v1.4.0](https://github.com/starship/starship/actions/runs/1954934580) successfully built and uploaded the notarized Macos binary and flat package installers, then proceeded to not make them part of the release.

I'm not sure why that's going on, but I'm pretty sure those files aren't of much use unless they wind up in the release.

As a stopgap, I've manually uploaded the .pkg files from the Action to the release, so that people who absolutely need notarized binaries can get started (if they need the binaries, they can run `pkgutil --expand starship.pkg an_output_dir` to extract just the binaries). There are no SHA hashes for those, but the entire package is signed with a developer certificate and notarized, so it should be possible to validate that way.

chipbuster:
#### Description
Changes the github release dependencies to include the notarization step so that notarized binaries get properly deployed

#### Motivation and Context

Closes #3714

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -202,7 +202,7 @@ jobs:
   # Create GitHub release with Rust build targets and release notes
   github_release:
     name: Create GitHub Release
-    needs: github_build
+    needs: [github_build, notarize_and_pkgbuild]
     runs-on: ubuntu-latest
     steps:
       - name: Setup | Checkout


andytom (approved):
LGTM

PR #3733, opened 2022-03-11
fix(aws): Make AWS_REGION orverrides AWS_DEFAULT_REGION (#3619)

### LINKED ISSUE #3619: wrong order for awscli region value ?
#### Current Behavior

It seems current starship make env var `AWS_DEFAULT_REGION` overrides `AWS_REGION`, https://github.com/starship/starship/blob/master/src/modules/aws.rs#L70

#### Expected Behavior

but according to the [aws official doc](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)

```markdown
[AWS_DEFAULT_REGION](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-region)
Specifies the AWS Region to send the request to.

If defined, this environment variable overrides the value for the profile setting region and . You can override this environment variable by using the --region command line parameter and the AWS SDK compatible AWS_REGION environment variable.
```

the correct behavior should be the opposite. 

#### Environment
- Starship version: 1.3.0
- fish version: fish, version 3.3.1
- Operating system: Mac OS 12.1.0
- Terminal emulator: vscode 1.64.2
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.58.1
- Rust channel:  release
- Build Time: 2022-02-07 14:56:29 +00:00

```fish
if status is-interactive
end
fish_add_path /usr/local/sbin
```

#### Starship Configuration

```toml
<unknown config>
```

kuno:
#### Description
Making environmental values `AWS_REGION` overrides `AWS_DEFAULT_REGION`, if both values is existing.

#### Motivation and Context

Closes #3619

#### Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -68,8 +68,8 @@ fn get_aws_profile_and_region(context: &Context) -> (Option<Profile>, Option<Reg
         .iter()
         .find_map(|env_var| context.get_env(env_var));
     let region = context
-        .get_env("AWS_DEFAULT_REGION")
-        .or_else(|| context.get_env("AWS_REGION"));
+        .get_env("AWS_REGION")
+        .or_else(|| context.get_env("AWS_DEFAULT_REGION"));
     match (profile, region) {
         (Some(p), Some(r)) => (Some(p), Some(r)),
         (None, Some(r)) => (None, Some(r)),
@@ -224,7 +224,7 @@ mod tests {
             .collect();
         let expected = Some(format!(
             "on {}",
-            Color::Yellow.bold().paint("☁️  (ap-northeast-1) ")
+            Color::Yellow.bold().paint("☁️  (ap-northeast-2) ")
         ));
 
         assert_eq!(expected, actual);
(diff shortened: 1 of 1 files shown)


src/modules/aws.rs
```diff
@@ -68,8 +68,8 @@ fn get_aws_profile_and_region(context: &Context) -> (Option<Profile>, Option<Reg
         .iter()
         .find_map(|env_var| context.get_env(env_var));
     let region = context
```
andytom, line 70: Not a blocker but maybe it makes sense to refactor this so it is the same as how we are finding the profile, something like:

```rust
 ​    ​let​ refion_env_vars ​=​ ​vec!​[​​"AWS_REGION", AWS_DEFAULT_REGION"]; 
 ​    ​let​ region ​=​ region_env_vars 
 ​        .​iter​() 
 ​        .​find_map​(​|​env_var​|​ context.​get_env​(env_var));
```

andytom (approved):
Have left a suggestion but not a blocker.

andytom:
Thank you for your contribution @kuno.

PR #3734, opened 2022-03-11
feat(nodejs): check for `.mts` and `.cts` files

kidonng:
#### Description

These are [TypeScript counterparts](https://www.typescriptlang.org/docs/handbook/esm-node.html#new-file-extensions
) of Node.js `.mjs` and `.cjs` file extensions.

Tests were added but removed later since they are unnecessary and are probably going to be dropped in the future, see conversations below.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2248,7 +2248,7 @@ By default the module will be shown if any of the following conditions are met:
 
 ### Options
 
-| Option              | Default                              | Description                                                                                           |
+| Option              | Default                                    | Description                                                                                           |
 | ------------------- | ------------------------------------ | ----------------------------------------------------------------------------------------------------- |
 | `format`            | `"via [$symbol($version )]($style)"`       | The format for the module.                                                                            |
 | `version_format`    | `"v${raw}"`                                | The version format. Available vars are `raw`, `major`, `minor`, & `patch`                             |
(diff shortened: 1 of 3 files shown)


kidonng:
Leaving some random thoughts here:

- Ideally we should generate some metainfo file from `src/configs` and use that as data source in the docs instead. Single source of truth is better and avoid manual sync (and mistakes).
- Tests for `detect_extensions`, `detect_files` and `detect_folders` are *way* more than necessary. We should only have _one_ test for each of their functionality.

davidkna:
> Ideally we should generate some metainfo file from `src/configs` and use that as data source in the docs instead. Single source of truth is better and avoid manual sync (and mistakes).

Case in point: you need to update `detect_extesions` in the table. Would you mind creating an issue about this? It might be possible to follow whatever rustdoc is doing or maybe a [JSON-schema](https://docs.rs/schemars/latest/schemars/index.html) could be used as a data source (#1951).

> Tests for `detect_extensions`, `detect_files` and `detect_folders` are way more than necessary. We should only have one test for each of their functionality.

True, there's not much value in that. I feel like a macro could handle most of those tests anyway.

kidonng:
> Would you mind creating an issue about this?

If there's an issue open for that then it's good, we need to have the schema _first_ anyway.

davidkna (approved):

andytom:
Thank you for your contribution @kidonng and thanks for the review @davidkna.

PR #3738, opened 2022-03-11
fix(docs): fix and cleanup VuePress config

kidonng:
#### Description

- Use relative links: so they can be picked up by VuePress
- Remove [v0.45.0](https://github.com/starship/starship/releases/tag/v0.45.0) migration link from the sidebar: it's been almost 1.5 years
- Fix wrong zh-TW and zh-CN config position: previously they don't have any effect
- Refactor sidebar: languages besides the en-US now have "Advanced Installation" in the sidebar, also makes the sidebar more maintainable
- Enable VuePress evergreen: make the site smaller and faster

Regarding the sidebar refactor, this is the biggest part of this PR. Currently overrides are configured so existing sidebar items stay the same, but eventually the each translated pages should have the right title and then remove the override. [Refer to comments in the code](https://github.com/starship/starship/pull/3738/files#diff-52eb86351480925d61947a58caca1ea76d10950ec26533f16f37cb12a3519dffR3-R5).

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -121,7 +121,6 @@ module.exports = {
                     ["/advanced-config/", "Advanced Configuration"],
                     ["/faq/", "Frequently Asked Questions"],
                     ["/presets/", "Presets"],
-                    ["/migrating-to-0.45.0/", "Migrating to v0.45.0"],
                 ],
             },
             "/de-DE/": {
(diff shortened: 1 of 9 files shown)


docs/.vuepress/config.js
```diff
@@ -134,14 +153,12 @@ module.exports = {
                 // Custom navbar values
                 nav: [{ text: "Konfiguration", link: "/config/" }],
```
davidkna, line 154: Just noticed a bug, while testing:
```suggestion
                nav: [{ text: "Konfiguration", link: "/de-DE/config/" }],
```

davidkna (commented):
@matchai Can you check GA what the impact of enabling evergreen would amount to?

matchai:
@davidkna Great idea referring to GA!
If Analytics is anything to go by, we've had 25 IE sessions this year, with a 100% bounce rate.
If VuePress isn't shooing them off, something else might be. In any case, I think enabling `evergreen` is a worthwhile improvement.

PR #3741, opened 2022-03-12
fix(status): pipestatus and fill don't play well together

### LINKED ISSUE #3162: status module pipestatus segment and fill don't play well together on bash
#### Current Behavior
When a pipeline is executed with pipestatus segement of status enabled and the fill module is used, the fill character is only printed one time instead of enough times to fill.

#### Expected Behavior
fill works with status and pipestatus to fill the line

image

#### Environment
- Starship version: starship 0.58.0
- Shell type: bash
- Shell version: GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
- Shell plugin manager: none
- Terminal emulator: iTerm
- Operating system: ubuntu via multipass on macos

flw-cn:
pipestatus is a combination of multiple exit codes. In the past, each rendered exit code was passed to the final pipestatus as a string that actually contained the ansi code formatter, which causes problems with the length calculation of the pipestatus component. This is the reason why the length of the fill component is not correct. Just like issue 3162 says.

The main idea of this patch is to replace `map` with `map_variables_to_segments` to extend the variable `pipestatus` to `Vec<Segment>` instead of string. 

To do this I need to do a some prep work first to satisfy the data type check. All the code is here, and I've tested it in several of my environments, including macOS and Linux, and it all works fine.

#### Description

* `src/modules/status.rs`

#### Motivation and Context

Closes #3162

#### Checklist:

### DIFF
--- a/src/modules/status.rs
+++ b/src/modules/status.rs
@@ -55,28 +55,29 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     // Create pipestatus string
     let pipestatus = match pipestatus_status {
-        PipeStatusStatus::Pipe(pipestatus) => pipestatus
-            .iter()
-            .map(
-                |ec| match format_exit_code(ec.as_str(), config.format, None, &config, context) {
-                    Ok(segments) => segments
-                        .into_iter()
-                        .map(|s| s.to_string())
-                        .collect::<Vec<String>>()
-                        .join(""),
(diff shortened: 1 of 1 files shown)


davidkna (changes requested):
Sorry for only reviewing this now. I've found a few things that need to be changed before this can be merged. Please close this PR if you no longer wish to work on this.

PR #3753, opened 2022-03-15
feat(git_branch): add 'ignore_branches' option

### LINKED ISSUE #3689: Only show git branch when not on master/main
I like to keep the prompt as simple as possible and for it to only show information when it is relevant and useful. I use a single-line prompt and it usually looks nearly identical to the standard Debian bash prompt (coloured) when there aren't any active starship modules.

While I would like the git branch symbol to appear when it a git directory, I only want the branch name to appear when not in main/master. Not only would it be a lot neater, it would be much better at reminding me when I am not in main, which I am often doing even when the branch name is always displaying.

Does anybody know of any workarounds that can make this possible?

IsaacHorvath:
#### Description
This PR adds an ignore_branches option to the git_branch module in starship.toml. The option is a toml array of git branch names that the module will not display, e.g.

```
[git_branch]
ignore_branches = ["master", "main"]
```

#### Motivation and Context

Closes #3689 

The intention here is to allow starship to hide displaying your default branch when you're on it, for cleanliness. Since the git initialization system creates `master` by default, but it's not required to be a branch, I chose to allow the user to ignore whichever names they wish.

#### Checklist:

### DIFF
--- a/src/modules/git_branch.rs
+++ b/src/modules/git_branch.rs
@@ -385,7 +385,7 @@ mod tests {
         let actual = ModuleRenderer::new("git_branch")
             .config(toml::toml! {
                 [git_branch]
-                    ignore_branches = "dummy,test_branch"
+                    ignore_branches = ["dummy", "test_branch"]
             })
             .path(&repo_dir.path())
             .collect();
--- a/src/modules/git_branch.rs
+++ b/src/modules/git_branch.rs
@@ -37,7 +37,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let branch_name = repo.branch.as_ref()?;
     let mut graphemes: Vec<&str> = branch_name.graphemes(true).collect();
     
-    for ignore_branch in config.ignore_branch.split(',') {
+    for ignore_branch in config.ignore_branches.split(',') {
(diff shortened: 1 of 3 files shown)


src/modules/git_branch.rs
```diff
@@ -37,6 +37,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let branch_name = repo.branch.as_ref()?;
     let mut graphemes: Vec<&str> = branch_name.graphemes(true).collect();
 
+    for ignore_branch in config.ignore_branches {
```
andytom, line 40: Not a blocker, but maybe it would be better to use an iterator and the `branch_name` here, something like

```rust
let is_ignored_branch = config.ignored_branches.iter()
    .any(|ignore_branch| branch_name.eq(ignore_branch));

if is_ignored_branch {
    return None;
} 
```

(I am on my phone so the above probably has something wrong with it but hopefully you get the idea).

andytom:
Thank you for your contribution @IsaacHorvath.

eggbean:
Never thought this would actually be implemented when I suggested it. Thanks a lot. Looking forward to the next release.

IsaacHorvath:
I'm happy I could help! I'm very new to Rust so it feels good to be able to contribute something useful.

PR #3761, opened 2022-03-17
docs(preset): add Pastel Powerline preset

Huhni:
#### Description
A preset similar to the M365Princess one from Oh My Posh.
It also includes an example on how to substitute paths, because I couldn't find it in the docs.
(Updated to work with 1.40 - I hope I did it right)

#### Motivation and Context
There was no "powerline-like" preset for starship before and I hope it helps people make the switch to starship or to create more powerline presets.

See [c523727](https://github.com/starship/starship/commit/c5237272c94bcdf7188f6fb55686d0c870b6de77)

#### Checklist:


docs/.vuepress/public/presets/toml/pastel-powerline.toml
```diff
@@ -0,0 +1,69 @@
...
+# "Important  " = "  "
+
+[git_branch]
+symbol = ""
+style = "bg:#FCA17D"
+format = '[[ $symbol $branch ](bg:#FCA17D)]($style)'
+
+[git_status]
+style = "bg:#FCA17D"
+format = '[[ $all_status$ahead_behind ](bg:#FCA17D)]($style)'
```
matchai, line 53: In testing locally, I noticed there was some extra padding around the git status, which is especially visible when there's no status indicator.

CleanShot 2022-03-17 at 17 47 51@2x

CleanShot 2022-03-17 at 17 49 27@2x

Parentheses will only show their contents when variables within them have values, so this change will remove any excess space when there's no status 😄 

```suggestion
format = '[[($all_status$ahead_behind )](bg:#FCA17D)]($style)'
```

matchai (approved):
Looks good to me! 👍 
Thank you, @Huhni, for the excellent looking preset 😄

PR #3769, opened 2022-03-19
fix(init): Change Elvish init to `catch` for elvish 0.18

chipbuster:
#### Description
Change `except` to `catch` in the elvish init script and updates the appropriate documentation, in anticipation of the upcoming [elvish v0.18 release](https://github.com/elves/elvish/blob/master/website/blog/0.18.0-release-notes.md)

#### Motivation and Context
Elvish v0.18 (which is now in rc2) has deprecated the use of `except` in favor of `catch`. Anyone using the current init system with v0.18 will see the following message when using the current init script:

```
deprecation: "except" is deprecated; use "catch" instead
[eval 13], line 14:         } except {
```

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ Add the following to the end of `~/.elvish/rc.elv`:
 eval (starship init elvish)
 ```
 
-Note: Only Elvish v0.17+ is supported
+Note: Only Elvish v0.18+ is supported
 
 </details>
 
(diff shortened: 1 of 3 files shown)


andytom (approved):

chipbuster:
Elvish v0.18 is released.

davidkna (approved):

davidkna:
Thanks for the fix @chipbuster, and thanks for reviewing @andytom!

PR #3773, opened 2022-03-20
fix(init): Change Nushell init for nu 0.60

### LINKED ISSUE #1716: Nu shell doesn't support showing status of last command
#### Current Behavior
The `error_symbol` setting is ignored (`success_symbol` is always shown).

#### Expected Behavior
Whatever is set as `error_symbol` would be displayed on an error.

error_prompt_issues

#### Environment
- Starship version: 0.45.2
- nushell version: 0.20.0
- Operating system: Mac OS 10.15.7
- Terminal emulator: iTerm.app 3.3.12

```toml
path = ["/usr/local/opt/tomcat@8/bin", "/usr/local/opt/php@7.3/sbin", "/usr/local/opt/php@7.3/bin", "/usr/local/opt/node@12/bin", "/Users/rmcquen/bin", "/Users/rmcquen/.cargo/bin", "/usr/local/bin", "/usr/bin", "/bin", "/usr/sbin", "/sbin", "/Library/Apple/usr/bin"]
prompt = "echo $(starship prompt)"
startup = ["alias  jo [file] { open $file | to json --pretty 4 }"]

[line_editor]
auto_add_history = true
completion_type = "circular"
history_duplicates = "ignoreconsecutive"
history_ignore_space = false
max_history_size = 500000
tab_stop = 4

[textview]
tab_width = 4
```

#### Starship Configuration

```toml
add_newline = true

[character]
error_symbol = "[λ](bold red)"
success_symbol = "[λ](bold green)"
symbol = "[λ](bold blue)"

```

### LINKED ISSUE #3526: Starship fails when `%USERPROFILE%` is changed
#### Current Behavior
Start your preferred shell however you want (using starship), then set `%USERPROFILE%` to any path that is not a valid home directory. Curiously, when `%USERPROFILE%` is set to null, or to a valid home directory that is not yours, `starship` works fine. However, the following error is raised when `%USERPROFILE%` is not a valid home directory.

```
thread 'main' panicked at 'Unable to find home directory', C:\Users\CLIDragon\.cargo\registry\src\github.com-1ecc6299db9ec823\starship-1.2.1\src\logger.rs:27:22
```

#### Expected Behavior
To continue to work as normal.

This error is really strange. In `src/log.rs:27`, `crate::utils::home_dir()` is called, which calls into `directories_next`, which ultimately calls into `dirs_sys_next`. The strange part is that `dirs_next` also calls `dir_sys_next`, yet provides the correct response when `%USERPROFILE%` is modified. I've cross-posted this issue over on https://github.com/xdg-rs/dirs/issues/57, to hopefully get some input on why this occurs.

```rs
fn main() {
    // Returns None.
    let home_dir = directories_next::BaseDirs::new().map(|base_dirs| base_dirs.home_dir().to_owned());
    println!("Home directory (next): {:#?}", home_dir);   

    // Correctly returns Some("C:\Users\CLIDragon")
    let home_dir = dirs_next::home_dir();
    println!("Home dir (next): {:#?}", home_dir);
}
``` 

#### Possible Solution
Switch to using `dirs_next` instead of `directories_next`.

#### Environment
- Starship version: 1.2.1
- powershell version: 7.2.1
- Terminal emulator: Windows Terminal 1.11.3471.0
- Git Commit Hash: 
- Branch/Tag: 
- Build Time: 2022-01-27 17:04:43 +10:00

```pwsh
```

#### Starship Configuration

```toml

```

davidkna:
#### Description

This PR updates the nu init for the upcoming v0.60 release. `directories-next` is also replaced by `dirs-next` again to match nushell behavior.

#### Motivation and Context

Closes #1716
Closes #3526

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,16 +327,15 @@ eval $(starship init ion)
 <details>
 <summary>Nushell</summary>
 
-Add the following to the end of your Nushell configuration (find it by running `config path`):
+Add the following to the end of your Nushell configuration (find it by running `$nu.config-path`):
 
-```toml
-startup = [
-  "mkdir ~/.cache/starship",
-  "starship init nu | save ~/.cache/starship/init.nu",
-  "source ~/.cache/starship/init.nu",
-]
-prompt = "starship_prompt"
+```sh
+mkdir ~/.cache/starship
+starship init nu | save ~/.cache/starship/init.nu
+source ~/.cache/starship/init.nu
 ```
+
+Note: Only Nushell v0.60+ is supported
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -327,10 +327,16 @@ eval $(starship init ion)
 <details>
 <summary>Nushell</summary>
 
-Add the following to the end of your Nushell configuration (find it by running `$nu.config-path`):
+Run the following:
 
 ```sh
 mkdir ~/.cache/starship
+starship init nu | save ~/.cache/starship/init.nu
+```
+
+And add the following to the end of your Nushell configuration (find it by running `$nu.config-path`):
+
+```sh
 starship init nu | save ~/.cache/starship/init.nu
 source ~/.cache/starship/init.nu
 ```
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -49,7 +49,7 @@ assignees: ''
     Fish: ~/.config/fish/config.fish
     Xonsh: ~/.config/xonsh/rc.xsh
     Elvish: ~/.config/elvish/rc.elv
-    Nushell: ~/.config/nushell/config.nu
+    Nushell: $nu.config-path
     Ion: ~/.config/ion/initrc
 -->
 
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -41,7 +41,7 @@ assignees: ''
 - Operating system: [e.g. macOS 10.13.4, Windows 10]
 
 #### Relevant Shell Configuration
-<!-- 
+<!--
   Based on the shell you use, please paste the appropriate configuration.
   The default location for your shell is:
     Bash: ~/.bashrc
@@ -49,7 +49,7 @@ assignees: ''
     Fish: ~/.config/fish/config.fish
     Xonsh: ~/.config/xonsh/rc.xsh
     Elvish: ~/.config/elvish/rc.elv
-    Nushell: ~/.config/nu/config.toml
+    Nushell: ~/.config/nushell/config.nu
     Ion: ~/.config/ion/initrc
 -->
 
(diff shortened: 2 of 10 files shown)


matchai (approved):

jfchevrette:
I just tried these new instructions on macOS but they do not work as the config file is read from `~/Library/Application Support/nushell/config.nu` and nushell will not use `~/.config/nushell/config.nu`

jfchevrette:
It might be worth checking issue https://github.com/nushell/nushell/issues/4773 which was opened for zoxide but I believe will apply to starship and the updated instructions from this PR

The TLDR is that the `save` would happen at "eval time" and the `source` would happen at "sourcing" time and that's a nu language design decision

davidkna:
@jfchevrette Thanks for the suggestions.

chen244:
> 

```nu
def starship_prompt [] {
    # jobs are not supported
    # status is not supported
    starship prompt --cmd-duration $env.CMD_DURATION_MS
}
let-env PROMPT_COMMAND = { (starship_prompt) }
```
Maybe you can try this one?

davidkna:
@chen244  Given the high likelihood of future changes to the shell, I would rather avoid that.

README.md
```diff
@@ -327,17 +327,22 @@ eval $(starship init ion)
...
+```sh
+mkdir ~/.cache/starship
+starship init nu | save ~/.cache/starship/init.nu
 ```
 
+And add the following to the end of your Nushell configuration (find it by running `$nu.config-path`):
+
+```sh
+starship init nu | save ~/.cache/starship/init.nu
+source ~/.cache/starship/init.nu
```
jntrnr, line 341: You'll need to run the `starship init nu | save ~/.cache/starship/init.nu` as a separate step. Running them back to back like this in the startup, and the init.nu won't be found (the parser is looking for "init.nu" but the evaluator won't have run the previous line.

jntrnr:
@matchai - thanks for working on this fix. I made a comment about Nushell's parser/evaluator separation. For that init step, since it won't run before the `source` on the next line you'll need to run it as a separate step first before starting up the config.

README.md
davidkna: This is for the init file, where I believe there is no way to add additional steps. Users are instructed to run `mkdir ~/.cache/starship/` and `source ~/.cache/starship/init.nu` to partially work around this.

fdncred:
`$env.LAST_EXIT_CODE` is also supported in nushell now.

jntrnr:
@davidkna - oops, I should have tagged you for ^

README.md
jntrnr: For line 340 - I think this will confuse users as it looks like if they update starship and then startup nushell things will be fine, but if the newer starship has a newer init.nu, the config won't see it until it's restarted.

Maybe best to drop line 340 and just have users do the `source ~/.cache/starship/init.nu` in the config, letting them know they need to manually update after updating starship?

davidkna: It's a compromise. I think not having users update the file manually will be more reliable.

PR #3783, opened 2022-03-23
docs: Fix entry for container config

Hofer-Julian:
The entry describing the container style config was missing backticks.
This results in a broken render as can be seen [here](https://starship.rs/config/#container).


davidkna (approved):

davidkna:
Thanks for the fix!

PR #3786, opened 2022-03-24
refactor: replace module_config_derive with serde

### LINKED ISSUE #553: Discussion: Refactor config to use serde deserialize
Requested by @matchai, I have done some experiments on re-implementing the config with `serde`. I've solved some potential bottleneck of implementing `trait Deserialize` for some data types. You can see the changes in [branch refactor-use-serde](https://github.com/starship/starship/tree/refactor-use-serde).

### Pros
- Using `serde` is more generic and can also be used to parse file types other than `toml`.
- `serde` is better maintained than `starship_module_config_derive`.
- Currently, we load config by parsing the config file with `toml` and then load the config to `impl ModuleConfig`, which may not be as fast as loading it straightly with `toml::from_str::<Config>()`.
- We will no longer need a wrapper of `struct StarshipConfig` to store the global config. All we need is a `struct Config` which we can put straightly in `struct Context`.

### Cons
- We may have an awful developing experience to have a custom deserialization process implemented. (92+ lines vs 34 lines for `SegmentConfig`, and 26 lines vs 4 lines for `Style`, also we may have to tweaking lifetime and ownership in the implementation)
- All configs have to be loaded in the context, so there could be unseen difficulties to get the required subset of the config related to the module.
- We have to mess up the code base again with another big breaking refactoring.
- No `toml::Value` will be generated, which means we have to implement `Display` ourselves or use `derive(Debug)`, which may not be as efficient as `Display`.

Any thoughts? @starship/astronauts

davidkna:
#### Description

Changes include:
* Removing `starship_module_config_derive` and replacing it with `serde::Deserialize`
* Removing `RootModuleConfig::load_config`. While potentially useful, it was only used in tests. And it would require something like `serde::DeserializeSeed` which is not derived by serde.
* Merging `RootModuleConfig` into `ModuleConfig`
* Implementing a `ValueDeserializer` that holds a reference to a `toml::Value` in `serde_utils.rs`
* Deserialization errors (invalid type) are now logged and include the current key and the struct names
* Unknown keys are now considered an error. "Did you mean?"-messages are still possible

#### Motivation and Context

Closes #553

#### Checklist:


src/serde_utils.rs
```diff
@@ -0,0 +1,385 @@
...
+        ValueDeserializer {
+            value,
+            info,
+            current_key: Some(current_key),
+        }
+    }
+}
+
+impl ValueDeserializer<'_> {
+    /// Pretify an error message by adding the current key and struct name to it.
```
matchai, line 44:
```suggestion
    /// Prettify an error message by adding the current key and struct name to it.
```

matchai (approved):
Wow, such a huge cleanup! Looks great 😄 
LGTM! ✅

andytom:
Awesome work, thanks @davidkna.

PR #3804, opened 2022-03-28
fix(docker_context): ignore the "default" context

### LINKED ISSUE #3803: docker_context does not check for 'default' context
#### Current Behavior

When including the `docker_context` module, my prompt says (in a folder where a `Dockerfile` exists):

```
some-project via 🐳 default
❯
```

#### Expected Behavior

Since I am using the `default` context, I expect the module to be hidden:
```
some-project
❯
```

The [documentation](https://starship.rs/config/#docker-context) states that
> The docker_context module shows the currently active [Docker context (opens new window)](https://docs.docker.com/engine/context/working-with-contexts/) if it's not set to default [...]

My understanding is that it should not be shown if the context set to `default` (which is usually the case if one did not explicitly create a custom context).

#### Possible Solution

The same behavior is mentioned in the source code

https://github.com/starship/starship/blob/master/src/modules/docker_context.rs#L17

but there does not seem to be any kind of check in the module.

#### Environment
- Starship version: 1.5.4
- zsh version: zsh 5.8 (x86_64-apple-darwin21.0)
- Operating system: Mac OS 12.3.0
- Terminal emulator: iTerm.app 3.5.0beta5
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.59.0
- Rust channel:  release
- Build Time: 2022-03-24 19:12:59 +00:00

```bash
# Set window title
function set_window_title() {
    echo -ne "\033]0; $(basename "$PWD") \007"
}
precmd_functions+=(set_window_title)

# Locales (needed for mosh)
export LANG=en_US.UTF-8
export LC_CTYPE=en_US.UTF-8

# Add Homebrew sbin to path
export PATH="/usr/local/sbin:$PATH"

# Add Cargo to path

# Load NVM
# Note: still need to call `nvm init` before using Node
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && source "$NVM_DIR/nvm.sh" --no-use

# Load Ruby
which rbenv > /dev/null && eval "$(rbenv init -)"
# Use Homebrew's (updated) OpenSSL with Ruby
export RUBY_CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)"

# Setup starship prompt
eval "$(starship init zsh)"

# Enable autosuggestions
source /usr/local/share/zsh-autosuggestions/zsh-autosuggestions.zsh

# Enable fuzzy finder
FZF_DEFAULT_OPTS="--layout=reverse --border=rounded --info=inline"
```

#### Starship Configuration

```toml
add_newline = true

[nodejs]
disabled = true

[package]
disabled = true

[rust]
disabled = true

format = "$all"

```

tkintscher:
The [documentation](https://starship.rs/config/#docker-context) states that

> The docker_context module shows the currently active [Docker context (opens new window)](https://docs.docker.com/engine/context/working-with-contexts/) if it's not set to default [...]

This PR checks for the "default" docker context and only shows the module, if the context is not "default", in agreement with the documentation.

#### Description
The PR checks the docker context, which was discovered earlier, for the word "default".

#### Motivation and Context
Closes #3803

I have tested the before/after behavior by running `test_docker_context_default`, and also in a directory with a Dockerfile where I ran into the issue before.

#### Checklist:
- [ ] I have updated the documentation accordingly. _(does not apply, I have fixed the code to match the docs)_

### DIFF
--- a/src/modules/docker_context.rs
+++ b/src/modules/docker_context.rs
@@ -57,6 +57,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         }
     };
 
+    if ctx == "default" {
+        return None;
+    }
+
     let parsed = StringFormatter::new(config.format).and_then(|formatter| {
         formatter
             .map_meta(|variable, _| match variable {
@@ -301,6 +305,24 @@ mod tests {
             })
             .collect();
         let expected = Some(format!("via {} ", Color::Blue.bold().paint("🐳 starship")));
+
+        assert_eq!(expected, actual);
+
+        cfg_dir.close()
+    }
+
+    #[test]
+    fn test_docker_context_default() -> io::Result<()> {
+        let cfg_dir = tempfile::tempdir()?;
+
+        let actual = ModuleRenderer::new("docker_context")
+            .env("DOCKER_CONTEXT", "default")
+            .config(toml::toml! {
+                [docker_context]
+                only_with_files = false
+            })
+            .collect();
+        let expected = None;
 
         assert_eq!(expected, actual);
 


andytom (approved):
LGTM

cappyzawa (approved):

andytom:
Thank you for your contribution @tkintscher and thanks for the review @cappyzawa.

PR #3810, opened 2022-03-29
docs(dotnet): fix link to Microsoft’s documentation for Dotnet

ariasuni:
#### Description
Anchor in link was broken, I fixed it.

Edit: change «when there is a csproj file» to «when there is a `.csproj` file», and add backticks forgotten around a `true`.


matchai (approved):
LGTM! 👍 
Thank you for the fix-up, @ariasuni!

PR #3825, opened 2022-04-01
fix(custom): set cwd for external commands

### LINKED ISSUE #2965: Working directory is not correct in custom modules when starship is invoked with `--path`

davidkna:
#### Description

This PR makes sure `current_dir` is set for external command that are run as part of `command` and `when` in custom modules.

This PR also adds `ModuleRenderer` support for custom modules.

#### Motivation and Context

Closes #2965
Closes #3445

#### Checklist:

### DIFF
--- a/src/modules/custom.rs
+++ b/src/modules/custom.rs
@@ -1,5 +1,6 @@
 use std::env;
 use std::io::Write;
+use std::path::Path;
 use std::process::{Command, Output, Stdio};
 use std::time::Instant;
 
@@ -36,7 +37,7 @@ pub fn module<'a>(name: &str, context: &'a Context) -> Option<Module<'a>> {
 
     if !is_match {
         if let Some(when) = config.when {
-            is_match = exec_when(when, &config.shell.0);
+            is_match = exec_when(when, &config.shell.0, &context.current_dir);
         }
 
         if !is_match {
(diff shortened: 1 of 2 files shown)


davidkna:
I'll combine this with a larger rework that also solves https://github.com/starship/starship/issues/3801.

PR #3828, opened 2022-04-02
Check if process is running in docker or not

younes-io:
Related to https://github.com/starship/starship/issues/3821

1. Need confirmation that solution is conceptually OK
2. Adjust test units

### DIFF
--- a/src/modules/container.rs
+++ b/src/modules/container.rs
@@ -12,6 +12,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     use crate::formatter::StringFormatter;
     use crate::utils::read_file;
 
+    
+    fn check_if_docker_in_cgroup(context: &Context) -> Option<String> {        
+        Some(context
+            .exec_cmd("grep", &["-cim1", "/docker", "/proc/1/cgroup"])?
+            .stdout
+            .trim()
+            .to_string())        
+    }
+
     pub fn container_name(context: &Context) -> Option<String> {
         use crate::utils::context_path;
 
@@ -24,6 +33,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         if context_path(context, "/run/host/container-manager").exists() {
             // OCI
             return Some("OCI".into());
+        }
+
+        if check_if_docker_in_cgroup(context).unwrap() == "1" {
+            // Docker
+            return Some("Docker".into());
         }
 
         if context_path(context, "/run/systemd/container").exists() {
@@ -51,11 +65,6 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 .unwrap_or_else(|_| "podman".into());
 
             return Some(image_res);
-        }
-
-        if context_path(context, "/.dockerenv").exists() {
-            // docker
-            return Some("Docker".into());
         }
 
         None


src/modules/container.rs
```diff
@@ -12,6 +12,16 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     use crate::formatter::StringFormatter;
     use crate::utils::read_file;
 
+    fn check_if_docker_in_cgroup(context: &Context) -> Option<String> {
+        Some(
+            context
+                .exec_cmd("grep", &["-cim1", "/docker", "/proc/1/cgroup"])?
+                .stdout
+                .trim()
+                .to_string(),
```
davidkna, line 21: The method seems fine, but please handle this with native rust code (`File::open`, `BufReader`, `.lines()` and perhaps `regex`).

younes-io: Done

cblessing24:
For some reason `/proc/1/cgroup` returns `0::/` on all the images (ubuntu:18.04/20.04) I have tried.

src/modules/container.rs
```diff
@@ -7,11 +7,22 @@ pub fn module<'a>(_context: &'a Context) -> Option<Module<'a>> {
...
+
     use super::ModuleConfig;
     use crate::configs::container::ContainerConfig;
     use crate::formatter::StringFormatter;
     use crate::utils::read_file;
 
+    fn check_if_docker_in_cgroup() -> Option<bool> {
+        let mut cgroup_file = File::open("/proc/1/cgroup").ok()?;
+        let mut cgroup_content = String::new();
+        cgroup_file.read_to_string(&mut cgroup_content).ok()?;
```
davidkna, line 21: You could use the helper `crate::utils::read_file` to shorten this a bit.
Beyond this, please use `context_path` for the absolute `/proc/1/cgroup` path (documented in `Contributing.md` and also used mainly in this file) and please use it to write a test for this.

src/modules/container.rs
```diff
@@ -26,6 +37,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             return Some("OCI".into());
         }
 
+        if check_if_docker_in_cgroup().unwrap() {
```
davidkna, line 40: Have you tested this? This looks like it will crash if `/proc/1/cgroup` does not exist. Maybe just ignore the error and use the following:
```suggestion
        if check_if_docker_in_cgroup().unwrap_or(false) {
```

davidkna (changes requested):

src/modules/container.rs
```diff
@@ -26,6 +26,14 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
         }
 
+        if context_path(context, "/proc/1/cgroup").exists() {
+            let cgroup_content = crate::utils::read_file("/proc/1/cgroup").ok()?;
+            if cgroup_content.contains("/docker") {
+                // Docker
+                return Some("Docker".into());
+            }
+        }
+
```
davidkna, line 36: Please use `context_path` throughout to ensure a test con properly mock the file. Perhaps just combine `read_file` and `exists`?
```rust
if let Ok(cgroup_content) = crate::utils::read_file("/proc/1/cgroup") {
    if cgroup_content.contains("/docker") {
        // Docker
        return Some("Docker".into());
    }
}
```
And in case you have trouble writing a test, I would be glad to help.

younes-io: Thank you @davidkna 
I want to write a test case for this; would you suggest I follow the approach of `containerenv()` ?

davidkna: Sounds good. As the tests in this file are somewhat special, maybe also look at the other files how `ModuleRenderer` tests usually look.

chipbuster:
A heads-up: the workflow failures in this PR appear to have been spurious (aside from fmt), but I can't appear to re-run them at the moment. I don't know if this is because GitHub is bugged at the moment or if this is designed.

Could you run `cargo fmt` to make the rustfmt checks pass, and then push a new commit? Hopefully this will also trigger a new workflow that doesn't have the codecov failure.

younes-io:
is sth still required here ?

PR #3833, opened 2022-04-03
fix(nu): Use `=` instead of space to pass command line parameters

Hofer-Julian:
#### Description

Clap (?) doesn't like to parse negative integers like this: `--status -1`.
Changing it to `--status=-1` fixes that.
With this change the nu script also follows the fish and powershell script more closely. 

#### Motivation and Context

This fixes https://github.com/nushell/nushell/issues/5074

### DIFF
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -9,7 +9,7 @@ let-env PROMPT_INDICATOR = ""
 let-env PROMPT_COMMAND = {
     # jobs are not supported
     let width = (term size -c | get columns | into string)
-    ^::STARSHIP:: prompt --cmd-duration $env.CMD_DURATION_MS --status $env.LAST_EXIT_CODE --terminal-width $width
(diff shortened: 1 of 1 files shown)


davidkna:
Nushell is very particular about expansion of values in strings. You will need to use [`$""`-strings](https://www.nushell.sh/book/working_with_strings.html#string-interpolation).

Hofer-Julian:
> Nushell is very particular about expansion of values in strings. You will need to use [`$""`-strings](https://www.nushell.sh/book/working_with_strings.html#string-interpolation).

Indeed, I've missed that.

I've pushed the updated version (and actually tested if `--cmd-duration` and `--status` work as expected)

davidkna (approved):

davidkna:
Thanks for uncovering and fixing this @Hofer-Julian!

Hofer-Julian:
Thanks for the review!

PR #3839, opened 2022-04-05
fix(fish): add proper vi mode detection for fish shell

### LINKED ISSUE #625: More informative VI mode indicator

Recognise this might be a Fish-specific problem. Delving onwards...

Migrating from SpaceFish, where the  vi-mode indicator gave feed back on whether you were in replace, visual, insert or normal mode, as described [here](https://github.com/matchai/spacefish/blob/master/docs/Options.md). I must admit, I'm a big fan of the character switching between insert and normal mode! But it gives me confusing information on visual or replace mode. 

If you are in normal mode and then switch into either visual or replace mode, the prompt character switches back to insert mode. This is incredibly confusing, as there is now no clear indication of whether you are in insert, replace, or visual. When you then exit visual/replace mode, it switches back to the normal mode character, as expected.

Solution A: Allow custom configuration of characters for each mode. Basically, the currently implementation expanded to have a character for visual and replace modes.

Solution B: Allow inclusion of the default `fish_mode_prompt`, though I suspect this would be more complicated.

I have manually configured `fish_mode_prompt`, but this is blocked/disabled by Starship.

mrjones2014:
#### Description

Adds separate symbols for the rest of the missing vi keymap modes for Fish shell. I don't know how to do it for zsh because it's much more complicated for zsh: https://github.com/starship/starship/issues/625#issuecomment-732454148

#### Motivation and Context

Currently it shows the insert mode icon for any vi mode other than normal which is incredibly confusing and not correct.

Closes #625 

I don't have a Linux or Windows machine to test on.

#### Checklist:

I couldn't find the right place to update the documentation. Please inform me where and I'll update it.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -576,13 +576,16 @@ look at [this example](#with-custom-error-shape).
 
 ### Options
 
-| Option           | Default             | Description                                                                      |
-| ---------------- | ------------------- | -------------------------------------------------------------------------------- |
-| `format`         | `"$symbol "`        | The format string used before the text input.                                    |
-| `success_symbol` | `"[❯](bold green)"` | The format string used before the text input if the previous command succeeded.  |
-| `error_symbol`   | `"[❯](bold red)"`   | The format string used before the text input if the previous command failed.     |
-| `vicmd_symbol`   | `"[❮](bold green)"` | The format string used before the text input if the shell is in vim normal mode. |
-| `disabled`       | `false`             | Disables the `character` module.                                                 |
+| Option                     | Default              | Description                                                                             |
+| -------------------------- | -------------------- | --------------------------------------------------------------------------------------- |
+| `format`                   | `"$symbol "`         | The format string used before the text input.                                           |
+| `success_symbol`           | `"[❯](bold green)"`  | The format string used before the text input if the previous command succeeded.         |
+| `error_symbol`             | `"[❯](bold red)"`    | The format string used before the text input if the previous command failed.            |
+| `vicmd_symbol`             | `"[❮](bold green)"`  | The format string used before the text input if the shell is in vim normal mode.        |
+| `vicmd_replace_one_symbol` | `"[❮](bold purple)"` | The format string used before the text input if the shell is in vim `replace_one` mode. |
+| `vimcmd_replace_symbol`    | `"[❮](bold purple)"` | The format string used before the text input if the shell is in vim replace mode.       |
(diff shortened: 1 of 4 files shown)


mrjones2014:
I'm not seeing the string `vimcmd` anywhere in `config-schema.json`? Not sure what I'm missing here. Please advise.

davidkna:
> If your changes cause changes to the configuration, you will need to update the configuration schema in `.github/config-schema.json` with `cargo run --features config-schema -- config-schema > .github/config-schema.json`.

The job also shows the expected diff.

mrjones2014:
Ah, I missed that, I've updated it.

chipbuster (changes requested):
The code here looks good! Please also update the documentation (specifically, add the new symbols into the options in `docs/config/README.md` after line 585) and we should be good to merge.

mrjones2014:
@chipbuster updated!

docs/config/README.md
```diff
@@ -576,13 +576,16 @@ look at [this example](#with-custom-error-shape).
...
+| `format`                   | `"$symbol "`         | The format string used before the text input.                                           |
+| `success_symbol`           | `"[❯](bold green)"`  | The format string used before the text input if the previous command succeeded.         |
+| `error_symbol`             | `"[❯](bold red)"`    | The format string used before the text input if the previous command failed.            |
+| `vicmd_symbol`             | `"[❮](bold green)"`  | The format string used before the text input if the shell is in vim normal mode.        |
+| `vicmd_replace_one_symbol` | `"[❮](bold purple)"` | The format string used before the text input if the shell is in vim `replace_one` mode. |
+| `vimcmd_replace_symbol`    | `"[❮](bold purple)"` | The format string used before the text input if the shell is in vim replace mode.       |
+| `vimcmd_visual_symbol`     | `"[❮](bold yellow)"` | The format string used before the text input if the shell is in vim replace mode.       |
+| `disabled`                 | `false`              | Disables the `character` module.                                                        |
```
chipbuster, line 588: I should have thought of this on my last request, but I'm now worried that ZSH users will try to use these symbols and get confused when they don't work. Could we add a [warning](https://v1.vuepress.vuejs.org/guide/markdown.html#custom-containers) here that notes that the new additions only work with Fish shell due to upstream issues with ZSH? We could even link the relevant issue, if you think it would be informative.

mrjones2014: Done!

chipbuster (approved):
LGTM.

andytom:
Thank you for your contribution @mrjones2014 and thanks as always for reviewing @chipbuster.

PR #3847, opened 2022-04-07
ci: add release-please to automate release creation

### LINKED ISSUE #3013: Add Please Release to simplify releases
At the moment when we want to make a release we need to make a few manual steps and push directly to the master branch. While these steps are not complicated like all manual processes it is prone to human error. Additionally this process is not very transparent to our contributors and end users. It would be nice if we could automate the release process in a way that both eliminates the possibility of manual error and makes the process a bit more transparent.

[Release Please](https://github.com/googleapis/release-please) recently added support for rust. This would simplify our release process to simply merging a PR when we want to do a release which removes the manual steps and also makes it a bit more obvious what our next version will be and what it will contain. It also has the added bonus of making it possible to release from the Github app or via the CLI.

We could just leave the process as it is, we have not had any issues so far with the manual process and as the saying goes if it isn't broke don't fix it.

matchai:
#### Description

Add a [release-please](https://github.com/googleapis/release-please) action to automate the creation of release PRs.
When the PR is merged, release-please will:
- Update the repo's CHANGELOG.md
- Update the Cargo.toml's version number
- Create a tag with the new version number

That will then go on to trigger our standard release flow.

#### Motivation and Context

Closes #3013

Now that we no longer manage a monorepo, Release Please's defaults should work perfectly for us.

#### Checklist:


davidkna:
Does this mean the chglog action needs to be removed?

matchai:
Ah! This also creates a GitHub release. I thought it'd just create the tag.
Hmm... This'll need a little more work before it's ready.

matchai:
I've gone ahead and updated it to now be the first step in our release process, instead of treating it as a trigger for the release process.

davidkna:
The changes look fine, but you maybe you should also remove the chglog config files.

davidkna:
The changes look fine, but you maybe you should also remove the chglog config files.

davidkna (approved):

PR #3849, opened 2022-04-08
feat(battery): Display only when discharging

### LINKED ISSUE #3845: Option to only show battery status when not charging
## Feature Request
Right now, I can set a threshold for when to start showing the battery percentage, and have set it to `10%`, so I know when the battery is too low. Would be really cool if you could add an option to remove it when adaptor has been plugged in and the device is charging. Starship already can check if the device is charging or not, so I would assume it's easy to add such an option

You could create your own custom config to check the battery status and only print to prompt when not charging using the `when` clause or conditional strings

KentaKariya:
#### Description
Adds option `only_on_discharge` to limit the display of the `battery` module only when the battery is discharging (not plugged in), given that the existing `threshold` option evalutes to `true`.
As this is my first contribution to this project, I'll keep this as a draft for now and would appreciate some feedback. Thanks!

#### Motivation and Context

Closes #3845 

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -446,8 +446,9 @@ The default value for the `charging_symbol` and `discharging_symbol` option is r
 The `display` option is an array of the following table.
 
 | Option               | Default    | Description                                                                                               |
-| -------------------- | ---------- | --------------------------------------------------------------------------------------------------------- |
+|----------------------|------------|-----------------------------------------------------------------------------------------------------------|
 | `threshold`          | `10`       | The upper bound for the display option.                                                                   |
+| `only_on_discharge`  | `false`    | Display only when discharging.                                                                            |
 | `style`              | `bold red` | The style used if the display option is in use.                                                           |
 | `charging_symbol`    | `-`        | Optional symbol displayed if display option is in use, defaults to battery's `charging_symbol` option.    |
 | `discharging_symbol` | `-`        | Optional symbol displayed if display option is in use, defaults to battery's `discharging_symbol` option. |
@@ -459,8 +460,9 @@ The `display` option is an array of the following table.
 threshold = 10
 style = "bold red"
 
-[[battery.display]] # "bold yellow" style and 💦 symbol when capacity is between 10% and 30%
+[[battery.display]] # "bold yellow" style and 💦 symbol when capacity is between 10% and 30% and battery is discharging
 threshold = 30
+only_on_discharge = true
 style = "bold yellow"
 discharging_symbol = "💦"
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -446,7 +446,7 @@ The default value for the `charging_symbol` and `discharging_symbol` option is r
 The `display` option is an array of the following table.
 
 | Option               | Default    | Description                                                                                               |
-|----------------------|------------|-----------------------------------------------------------------------------------------------------------|
+| -------------------- | ---------- | --------------------------------------------------------------------------------------------------------- |
 | `threshold`          | `10`       | The upper bound for the display option.                                                                   |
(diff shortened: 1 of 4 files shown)


andytom:
Hi @KentaKariya in general this looks fine but you will need to run the following commands to fix the CI
```bash
# update the config schema
cargo run --features config-schema -- config-schema > .github/config-schema.json
# format the docs
dprint fmt
```

KentaKariya:
Thanks, I missed a step in the contribution guide.

davidkna:
I've sometimes had unknown as the state when discharging. It might be worth making this more generic.

docs/config/README.md
```diff
@@ -448,6 +448,7 @@ The `display` option is an array of the following table.
 | Option               | Default    | Description                                                                                               |
 | -------------------- | ---------- | --------------------------------------------------------------------------------------------------------- |
 | `threshold`          | `10`       | The upper bound for the display option.                                                                   |
+| `only_on_discharge`  | `false`    | Display only when discharging.                                                                            |
```
matchai, line 451: I wonder if we can make this more generic, instead providing a `when` that accepts various values: `charging`, `discharging`, `all`.
This way, folks can change the style depending on status (e.g. green when charging, red when discharging).

KentaKariya: That would definitely be useful. I'll add some commits to this PR for that, but I'm still unsure about the following points:

- Instead of specifying just a single battery state for the `when` clause, what about implementing it as a list so the same style can be applied to multiple battery states?
- Besides `charging` and `discharging`, the `State` enum also has types like `empty` or `full`. Maybe we could use this enum for the config field? It could be implemented as an `Option` so that a missing field (`None`) would apply the display config to all charging states.

chipbuster: I think both ideas sound good. We're slowly moving in this direction in other modules as well (using a list of states instead of 5+ separate binary toggles).

KentaKariya: Update: I tried to implement the `when` clause as an `Option<Vec<State>>`. I added the `Serialize` and `Deserialize` traits to `State` and although the config schema can be generated, Starship fails to load the config with the error message that `when` is given as a string when it expects an enum.

What I also attempted is to use the `serde_with` crate to serialize/deserialize `State` with the existing `Display` and `Serialize` implementations respectively using `DisplayFromStr`. However, this conflicts with the config schema generation where it cannot serialize `State` as `DisplayFromStr` does not implement `JsonSchema`...

davidkna: Does `#[cfg_attr(feature = "config-schema", derive(schemars::JsonSchema))]` not work?

KentaKariya: I'll probably have to implement `JsonSchema` for `State`, that's true. Using `serde_as`, I couldn't get the default derive for `JsonSchema` to work as it conflicts with `DisplayFromStr` for not implementing `JsonSchema`.

davidkna: Do `#[serde(deserialize_with = "path")]` and `#[serde(deserialize_with = "path")]` or `#[serde(remote = "...")]` work? If not, I think just adding a `State` `enum` to starship would be best.

KentaKariya: After trying around with `#[serde(remote = "...")]` and manually implementing the serialization logic for `JsonSchema`, I think the best way to implement the feature is to just add the traits for `serde` and `JsonSchema` directly to `starship_battery::State`. `serde_with` and `JsonSchema` don't seem to play along nicely so as far as I can tell you have to write the serialization logic twice to use them with _remote_ enums.

I'll go ahead and open a MR in the other repo to add the `serde` and `JsonSchema` traits if that's okay.

davidkna: Sounds good.

PR #3855, opened 2022-04-10
feat(cmd_duration): make notify feature optional (compat with nix darwin)

happysalada:
#### Description

the notify-rust has some very cryptic compilation error on nix and darwin. (reference issue https://github.com/NixOS/nixpkgs/issues/160876)

In the meanwhile, version has been moved forward in nix to not penalize linux users.

Hopefully we can find the source of the error, but in the meanwhile, making this dependency optional is very helpful!

Thank you again for your terminal! (going a month without it, feels like going back to the stone age #addicted)

#### Motivation and Context

Closes #

I ran `cargo test`, `cargo test --no-default-features`, `cargo clippy --all-features --all-targets` and couldn't find problems with it, let me know however if you need changes.

#### Checklist:

### DIFF
--- a/src/modules/cmd_duration.rs
+++ b/src/modules/cmd_duration.rs
@@ -51,6 +51,16 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     Some(undistract_me(module, &config, elapsed))
 }
 
+#[cfg(not(feature = "notify"))]
+fn undistract_me<'a, 'b>(
+    module: Module<'a>,
+    _config: &'b CmdDurationConfig,
+    _elapsed: u128,
+) -> Module<'a> {
+    module
+}
+
+#[cfg(feature = "notify")]
 fn undistract_me<'a, 'b>(
     module: Module<'a>,
     config: &'b CmdDurationConfig,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,10 @@ The minimal, blazing-fast, and infinitely customizable prompt for any shell! ☄
 """
 
 [features]
-default = ["battery"]
+default = ["battery", "notify"]
 battery = ["starship-battery"]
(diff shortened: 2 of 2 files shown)


happysalada:
Semantic pull request is unhappy, is this a feat or a fix ?
I couldn't decide.

davidkna:
I would consider this a feat. Please don't disable the whole `cmd_duration` module if the feature is disabled. Maybe try building a partial revert for #3547? Though I would prefer if you keep using `notify` instead of `notify-rust` as the feature name.

happysalada:
I've only made the calls to the notification optional.
I could have made the call to `undistract_me` optional, but this would require a cfg_if! so I thought this was simpler.
Let me know if anything.

happysalada:
alright, I think I've got it this time.
Compilation with no warnings.
I've also changed the commit message to reflect the title of this PR
let me know if anything.

src/modules/cmd_duration.rs
```diff
@@ -51,14 +51,26 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
 fn undistract_me<'a, 'b>(
     module: Module<'a>,
     config: &'b CmdDurationConfig,
     elapsed: u128,
 ) -> Module<'a> {
     use ansi_term::{unstyle, ANSIStrings};
+    #[cfg(feature = "notify")]
     use notify_rust::{Notification, Timeout};
 
+    #[cfg(feature = "notify")]
```
davidkna, line 73: LGTM, but please remove the inner `cfg`s:

```suggestion
    use notify_rust::{Notification, Timeout};

```

happysalada:
my bad for forgetting those, fixed and pushed.

davidkna (approved):

davidkna:
Thanks for the contribution @happysalada!

PR #3865, opened 2022-04-12
ci(release-please): use manifest command

davidkna:
#### Description

This PR changes the release-please workflow to use the manifest command and adds the required manifest files. This should also lead to older versions being added to the initial changelog file.
In case building the changelog leads to exceeding GitHub rate limits, a cut-off bootstrap commit will need to be specified.

[I've decided not to adjust the changelog configuration](https://github.com/starship/starship/pull/3861#issuecomment-1096519206), because it looks like the most important sections are not hidden by default.

Also restrict release-please to only run the main repo.

#### Motivation and Context

Related #3861

#### Checklist:


matchai:
Outside of the changelog, I don't think this'll change a whole lot about the release process.
I think that we should be able to generate the changelog once and commit it.

As for changelog configuration, it should be able to be adjusted as an action property:
https://github.com/starship/starship/commit/ebb8636790f02d7b6f8c0f9ba1344d8dee982f42

PR #3868, opened 2022-04-12
ci: set changelog types for release-please

matchai:
#### Description

Set changelog types to add "perf" and "refactor" conventional commits to our changelog.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,6 +23,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           release-type: rust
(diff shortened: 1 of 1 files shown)


davidkna:
By the way, I checked and perf is already included by default in the changelog.

matchai:
Would you be opposed to leaving it as the defaults then?
We could search for refactors in the commit history rather than the changelog.

davidkna:
@matchai  I'm not opposed to leaving the defaults.

matchai:
In that case, I think it would be simplest and easiest to leave it as-is then.
I felt that our current changelogs list a few too many things that aren't actionable or valuable for our users. I think the defaults make sense. 👍

PR #3880, opened 2022-04-16
feat: Add a module for C++ projects

lucasaf04:
#### Description

Adds a module to detect and handle C++ language projects

#### Motivation and Context

There's no existing module for C++, and one for C was merged recently (#3631).

Screenshot 2022-04-16 at 17 47 59

#### Checklist:

### DIFF
--- a/docs/.vuepress/public/config-schema.json
+++ b/docs/.vuepress/public/config-schema.json
@@ -216,6 +216,44 @@
         }
       ]
     },
+    "cpp": {
+      "default": {
+        "commands": [
+          [
+            "c++",
+            "--version"
+          ],
+          [
+            "cpp",
+            "--version"
+          ],
+          [
+            "g++",
+            "--version"
+          ],
+          [
+            "clang++",
+            "--version"
+          ]
+        ],
+        "detect_extensions": [
+          "cpp",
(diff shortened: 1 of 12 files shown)


davidkna:
What's your use-case for this? I don't think there's any popular C compiler that's not also a C++ compiler at the same time. At the very least, it would be nice if this would share most of its code with the C module.

GMbrianlaw:
Some people may want different attributes for `C` and `C++`. For example, they may want a different color for each of them.

lucasaf04:
@davidkna Sorry, I’m new to rust and I don’t know how to share code with the C module. Help would be appreciated.

Should I close this PR instead?

davidkna:
@lucasaf04 Ok. But if you want, you could make another PR that adds C++ file extensions to the C module or adds conditional styling options if C++ files are detected.

RayZ0rr:
> What's your use-case for this? I don't think there's any popular C compiler that's not also a C++ compiler at the same time. At the very least, it would be nice if this would share most of its code with the C module.

I agree with this. In my config I am using
```toml
[c]
detect_extensions = ["c", "h", "cc", "cpp", "hh", "hpp"]
```

PR #3482, opened 2022-01-21
fix: Allows the use of git worktrees together with sparse-checkout.

knutwalker:
Note: This PR requires https://github.com/rust-lang/git2-rs/pull/791
Will update this PR once that one has been merged and released.

#### Description

Calling [git sparse-checkout init](https://git-scm.com/docs/git-sparse-checkout#Documentation/git-sparse-checkout.txt-eminitem) on a [git worktree](https://git-scm.com/docs/git-worktree#_name) will reconfigure the repository to use worktree specific configs.
This is done by enabling the [`worktreeConfig`](https://git-scm.com/docs/git-worktree#_configuration_file) extension.
The default behavior for git2 is to refuse to open any repository with an extension, which in turn disables all git-related features in starship.

#### Motivation and Context

Closes #3245

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #2888, opened 2021-07-14
refactor(git_commit): Stop using libgit2

andytom:
#### Description

This switches the git_commit module from using libgit2 to shelling out
to git. This removes the dependency on libgit2 and should solve some
edge cases where libgit2 doesn't respond the same way that git does.

#### Motivation and Context

Closes #

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #3883, opened 2022-04-17
feat(git): replace `git2` with `git-repository`

### LINKED ISSUE #4022: v1.7.1 Homebrew build failure
#### Current Behavior

Failing Homebrew formula update PR here: https://github.com/Homebrew/homebrew-core/pull/102194

Failed to build for homebrew with the following error message:
```
    Installing starship v1.7.0 (/private/tmp/starship-20220524-38219-1pth72q/starship-1.7.0)
      Updating crates.io index
  error: failed to select a version for the requirement `libgit2-sys = "^0.13.4"`
  candidate versions found which didn't match: 0.13.2+1.4.2, 0.13.1+1.4.2, 0.13.0+1.4.1, ...
```
https://github.com/Homebrew/homebrew-core/runs/6582861876?check_suite_focus=true#step:6:113

### LINKED ISSUE #4131: Can't upgrade after 1.6.3 in Ubuntu jammy
#### Current Behavior

When trying to cargo upgrade, on Ubuntu 22.04 (Jammy), one can only upgrade up to 1.6.3, from 1.7.0 onwards (switch to libgit2-sys I guess) we get : 

```
Updating starship
    Updating crates.io index
  Installing starship v1.9.1
error: failed to compile `starship v1.9.1`, intermediate artifacts can be found at `/tmp/cargo-installbCceWR`

Caused by:
  failed to select a version for the requirement `libgit2-sys = "^0.13.4"`
  candidate versions found which didn't match: 0.13.2+1.4.2, 0.13.1+1.4.2, 0.13.0+1.4.1, ...
  location searched: crates.io index
  required by package `git2 v0.14.4`
      ... which satisfies dependency `git2 = "^0.14.4"` of package `starship v1.9.1`
```

#### Expected Behavior

Being able to upgrade, or have instructions on how to upgrade for this specific platform. 

https://packages.ubuntu.com/search?keywords=libgit2-sys-dev seems to indicate that vers 0.13.4 has not yet been packaged for Ubuntu, but maybe I'm looking at the wrong thing. 

#### Environment
- Starship version: 1.6.3
- fish version: fish, version 3.5.0
- Operating system: Ubuntu 22.04
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.59.0
- Rust channel:  release
- Build Time: 2022-06-29 14:50:28 +02:00

davidkna:
#### Description

This PR replaces the `libgit2`-based `git2` crate with the `gitoxide` `git-repository` crate.

~~In addition, add some ownership-based trust handling to the `git_status` and `git_metrics` modules with `git-sec`.~~

In addition, improve the `git_branch` output by preferring the reference name over plain `HEAD`.

#### Motivation and Context

Closes #4022
Closes #3482
Closes #4131

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -7,7 +7,8 @@ use crate::modules;
 use crate::utils::{self, home_dir};
 use clap::Parser;
 use git_repository::{
-    commit::describe::SelectRef::AllRefs, state as git_state, Repository, ThreadSafeRepository,
+    self as git, commit::describe::SelectRef::AllRefs, open::Options as GitOpenOptions,
+    sec as git_sec, state as git_state, Repository, ThreadSafeRepository,
 };
 use once_cell::sync::OnceCell;
 #[cfg(test)]
@@ -258,11 +259,36 @@ impl<'a> Context<'a> {
     }
 
     /// Will lazily get repo root and branch when a module requests it.
-    pub fn get_repo(&self) -> Result<&Repo, git_repository::discover::Error> {
+    pub fn get_repo(&self) -> Result<&Repo, git::discover::Error> {
         self.repo
-            .get_or_try_init(|| -> Result<Repo, git_repository::discover::Error> {
-                let shared_repo =
-                    ThreadSafeRepository::discover_with_environment_overrides(&self.current_dir)?;
+            .get_or_try_init(|| -> Result<Repo, git::discover::Error> {
+                // custom open options
+                let git_open_opts = GitOpenOptions::default()
+                    // trust the git repo and avoid additional checks
+                    .with(git_sec::Trust::Full)
+                    // don't use the global git config
+                    .permissions(git::Permissions {
+                        config: git::permissions::Config {
+                            system: false,
+                            git: false,
+                            user: false,
+                            env: true,
+                            includes: true,
+                        },
+                        ..git::Permissions::all()
+                    });
+
+                // convert the option options into a trust-based mapping for `discover`
+                let open_opts_mapping = git_sec::trust::Mapping {
+                    full: git_open_opts.clone(),
+                    reduced: git_open_opts,
+                };
+
+                let shared_repo = ThreadSafeRepository::discover_with_environment_overrides_opts(
+                    &self.current_dir,
+                    Default::default(),
+                    open_opts_mapping,
+                )?;
 
                 let repository = shared_repo.to_thread_local();
                 let branch = get_current_branch(&repository);
--- a/src/context.rs
+++ b/src/context.rs
@@ -8,7 +8,6 @@ use crate::utils::{self, home_dir};
 use clap::Parser;
 use git_repository::{
     self as git,
-    commit::describe::SelectRef::AllRefs,
     sec::{self as git_sec, trust::DefaultForLevel},
     state as git_state, Repository, ThreadSafeRepository,
 };
@@ -22,7 +21,6 @@ use std::fmt::Debug;
 use std::fs;
 use std::marker::PhantomData;
 use std::num::ParseIntError;
-use std::ops::Deref;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::string::String;
@@ -295,14 +293,7 @@ impl<'a> Context<'a> {
 
                 let repository = shared_repo.to_thread_local();
                 let branch = get_current_branch(&repository);
-                // Only valid if we have a real branch
-                let remote = match branch {
-                    Some(Branch::HeadName(ref b)) => {
-                        get_remote_repository_info(&repository, b.as_str())
-                    }
-                    _ => None,
-                };
-
+                let remote = get_remote_repository_info(&repository, branch.as_deref());
                 let path = repository.path().to_path_buf();
                 Ok(Repo {
                     repo: shared_repo,
@@ -540,7 +531,7 @@ pub struct Repo {
 
     /// If `current_dir` is a git repository or is contained within one,
     /// this is the current branch name of that repo.
-    pub branch: Option<Branch>,
+    pub branch: Option<String>,
 
     /// If `current_dir` is a git repository or is contained within one,
     /// this is the path to the root of that repo.
@@ -567,24 +558,6 @@ impl Repo {
 pub struct Remote {
     pub branch: Option<String>,
     pub name: Option<String>,
-}
-
-#[derive(Clone)]
-pub enum Branch {
-    HeadName(String),
-    Ref(String),
-    Head,
-}
-
-impl Deref for Branch {
-    type Target = str;
-    fn deref(&self) -> &str {
-        match self {
-            Branch::HeadName(name) => name.as_str(),
-            Branch::Ref(name) => name.as_str(),
-            Branch::Head => "HEAD",
-        }
-    }
 }
 
 // A struct of Criteria which will be used to verify current PathBuf is
@@ -631,31 +604,18 @@ impl<'a> ScanDir<'a> {
     }
 }
 
-fn get_current_branch(repository: &Repository) -> Option<Branch> {
-    get_current_head_name(repository)
-        .or_else(|| describe_head_ref(repository))
-        .or_else(|| repository.head().is_ok().then(|| Branch::Head))
-}
-
-// Get the conanical name of the current HEAD
-fn get_current_head_name(repository: &Repository) -> Option<Branch> {
+fn get_current_branch(repository: &Repository) -> Option<String> {
     let name = repository.head_name().ok()??;
     let shorthand = name.shorten();
 
-    Some(shorthand.to_string()).map(Branch::HeadName)
+    Some(shorthand.to_string())
 }
 
-// Try to get the reference name that the head commit is pointing to. (`origin/main`)
-fn describe_head_ref(repository: &Repository) -> Option<Branch> {
-    let head_commit = repository.head_commit().ok()?;
-
-    let describe_platform = head_commit.describe().names(AllRefs);
-    let formatter = describe_platform.try_format().ok()??;
-
-    Some(formatter.name?.to_string()).map(Branch::Ref)
-}
-
-fn get_remote_repository_info(repository: &Repository, branch_name: &str) -> Option<Remote> {
+fn get_remote_repository_info(
+    repository: &Repository,
+    branch_name: Option<&str>,
+) -> Option<Remote> {
+    let branch_name = branch_name?;
     let branch = repository
         .remote_ref(branch_name)
         .and_then(|r| r.ok())
--- a/src/context.rs
+++ b/src/context.rs
@@ -6,7 +6,7 @@ use crate::utils::{create_command, exec_timeout, read_file, CommandOutput};
 use crate::modules;
 use crate::utils::{self, home_dir};
 use clap::Parser;
-use git2::{ErrorCode::UnbornBranch, Repository, RepositoryState};
+use git_repository::{state as git_state, Repository, ThreadSafeRepository};
 use once_cell::sync::OnceCell;
 #[cfg(test)]
 use std::collections::HashMap;
@@ -255,22 +255,25 @@ impl<'a> Context<'a> {
     }
 
     /// Will lazily get repo root and branch when a module requests it.
-    pub fn get_repo(&self) -> Result<&Repo, git2::Error> {
-        self.repo.get_or_try_init(|| -> Result<Repo, git2::Error> {
-            let repository = if env::var("GIT_DIR").is_ok() {
-                Repository::open_from_env()
-            } else {
-                let dirs: [PathBuf; 0] = [];
-                Repository::open_ext(&self.current_dir, git2::RepositoryOpenFlags::FROM_ENV, dirs)
-            }?;
-            Ok(Repo {
-                branch: get_current_branch(&repository),
-                workdir: repository.workdir().map(Path::to_path_buf),
-                path: Path::to_path_buf(repository.path()),
-                state: repository.state(),
-                remote: get_remote_repository_info(&repository),
+    pub fn get_repo(&self) -> Result<&Repo, git_repository::discover::Error> {
+        self.repo
+            .get_or_try_init(|| -> Result<Repo, git_repository::discover::Error> {
+                let shared_repo =
+                    ThreadSafeRepository::discover_with_environment_overrides(&self.current_dir)?;
+
+                let repository = shared_repo.to_thread_local();
+                let branch = get_current_branch(&repository);
+                let remote = get_remote_repository_info(&repository, branch.as_deref());
+                let path = repository.path().to_path_buf();
+                Ok(Repo {
+                    repo: shared_repo,
+                    branch,
+                    workdir: repository.work_dir().map(PathBuf::from),
+                    path,
+                    state: repository.state(),
+                    remote,
+                })
             })
-        })
     }
 
     pub fn dir_contents(&self) -> Result<&DirContents, std::io::Error> {
@@ -494,6 +497,8 @@ impl DirContents {
 }
 
 pub struct Repo {
+    pub repo: ThreadSafeRepository,
+
     /// If `current_dir` is a git repository or is contained within one,
     /// this is the current branch name of that repo.
     pub branch: Option<String>,
@@ -506,7 +511,7 @@ pub struct Repo {
     pub path: PathBuf,
 
     /// State
-    pub state: RepositoryState,
+    pub state: Option<git_state::InProgress>,
 
     /// Remote repository
     pub remote: Option<Remote>,
@@ -514,8 +519,8 @@ pub struct Repo {
 
 impl Repo {
     /// Opens the associated git repository.
-    pub fn open(&self) -> Result<Repository, git2::Error> {
-        Repository::open(&self.path)
+    pub fn open(&self) -> Repository {
+        self.repo.to_thread_local()
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/context.rs
```diff
@@ -255,22 +255,29 @@ impl<'a> Context<'a> {
...
+        self.repo
+            .get_or_try_init(|| -> Result<Repo, git_repository::discover::Error> {
+                let shared_repo =
+                    ThreadSafeRepository::discover_with_environment_overrides(&self.current_dir)?;
+
+                let repository = shared_repo.to_thread_local();
+                let branch = get_current_branch(&repository);
+                let remote = get_remote_repository_info(&repository, branch.as_deref());
+                let path = repository.path().to_path_buf();
+                let trust_level = git_sec::Trust::from_path_ownership(&path)
```
Byron, line 268: Ideally `git-repository` users won't ever have the need to pull in a plumbing crate directly. Thus `git_sec` should be `git::sec` here unless it's a matter of preference.

The reason for that is that they will only have to maintain one crate version to use anything `git` related, which is `git-repository`. Note that as per [gitoxide's stability policy](https://github.com/Byron/gitoxide/blob/498072ea42dca7b9d00bedba42829bdac92195b9/STABILITY.md#L1-L0) breaking changes in exposed dependencies of `git-repository` will automatically lead to a breaking semver change in `git-repository` transitively, so doing so should be preferred and is save to do as well.

Byron (commented):
I just took a very first glance at the transition and come up with a suggestion. Please also note that independently of that I am working on a new release which will hopefully be done very soon (`cargo smart-release` has problems, again).

src/context.rs
```diff
@@ -255,22 +255,29 @@ impl<'a> Context<'a> {
...
+                    .unwrap_or(git_sec::Trust::Reduced);
+                Ok(Repo {
+                    repo: shared_repo,
+                    branch,
+                    // gitoxide might relativize the path, so we need to expand it again
```
Byron, line 273: Is this not fixed in `gitoxide` by now? I thought it won't over-eagerly make paths relative anymore.

src/modules/git_commit.rs
```diff
@@ -80,23 +50,27 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-fn format_tag(symbol: &str, tag_name: &str) -> Option<Result<String, StringFormatterError>> {
-    if tag_name.is_empty() {
-        None
-    } else {
-        Some(Ok(format!("{}{}", &symbol, &tag_name)))
-    }
+fn git_tag(repo: &Repo) -> Option<String> {
+    let git_repo = repo.open();
+    let mut git_head = git_repo.head().ok()?;
+    let head_commit = git_head.peel_to_commit_in_place().ok()?;
```
Byron, line 56: There is a [`head_commit()`](https://github.com/Byron/gitoxide/blob/b362fb594546400e6c42688103df438954df7eeb/git-repository/src/repository/reference.rs#L213) method now.

Generally, `gitoxide` wants to make common tasks simple, and I am happy to add what's seemingly missing, too.

src/modules/git_commit.rs
```diff
@@ -80,23 +50,27 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-        .map(|b| format!("{:02x}", b))
-        .collect::<String>()
-        .chars()
-        .take(len)
-        .collect()
+fn git_hash(repo: &Repo, config: &GitCommitConfig) -> Option<String> {
+    let git_repo = repo.open();
+    let mut git_head = git_repo.head().ok()?;
+    let head_commit = git_head.peel_to_commit_in_place().ok()?;
+    let head_id = head_commit.id();
```
Byron, line 68: There is a [`head_id()`](https://github.com/Byron/gitoxide/blob/b362fb594546400e6c42688103df438954df7eeb/git-repository/src/repository/reference.rs#L194) method as well.

Byron (commented):
`git-repository` has been published and is now at version 0.19. Furthermore I made some suggestions which will remove some boilerplate.

Overall, I am happy to see how the `gitoxide` version overall seems to require less lines of code without loosing clarity. It certainly also comes at the cost of having to deal with `ThreadSaveRepository` which might be a little confusing at first (but I see no way around it either).

src/context.rs
davidkna: Thanks, I didn't notice that this was included in `git-repository`.

src/modules/git_commit.rs
davidkna: Thanks for the heads-up!

src/context.rs
davidkna: Thanks, I hadn't updated that yet.

Byron (approved):

src/modules/git_metrics.rs
```diff
@@ -21,6 +22,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     };
 
     let repo = context.get_repo().ok()?;
+
+    if repo.trust_level < git_sec::Trust::Full {
+        log::debug!("skipping, not a trusted repo {:?}", repo.trust_level);
+        return None;
+    }
```
davidkna, line 29: I thought I would be a good idea to handle repo trust with some basic checks, though it might be better to wait until `secure.directories` can be handled to improve the experience. At this time, (our) starship builds with git2 don't handle this at all.
I've used `debug` over `warn` here to make this less noisy.

Byron: I see, it's probably preferable to fail gracefully than have `git` bail out, even though with the current implementation I bet `starship` would be over-eagerly bailing even though `git` might be configured to accept the path.

I don't know if using `git config` to obtain the `safe.directories` value is feasible in the meantime (or does that bail out too :D?)

davidkna: This is mostly intended as a fallback in case git hasn't yet been updated to use a security model.

Using `git config` might slow down the prompt too much on Windows (#2888).

andytom (approved):
I've been using this for a wee bit on my Linux machine and haven't had any problems but would be good to get some additional data.

matchai:
I'll use it as my daily driver for the day 👍

gdzx:
@davidkna I see that you also fixed the issue about detached HEADs. I made a slightly different patch for that feature (sadly it relies on `git2` from `master`). It is based on the reflog, which is how `git status` finds the refname when in a detached HEAD state ([`git/wt-status:1640`](https://github.com/git/git/blob/e4a4b31577c7419497ac30cebe30d755b97752c5/wt-status.c#L1640)). Here's the [diff](https://github.com/gdzx/starship/commit/17cb0835de597c249120fe7856310a38c8c7ac45#diff-6008532e472abe869f74b6354c4142405356928ba87b05d4cba0bb9a4f55cfffR132) for the Rust implementation.

Because the target refname ends up in the reflog message, it has the advantage that it returns what the user expects after a checkout of a specific tag or branch, even when multiple refs are pointing to the same object. It may also be more efficient than git-describe (which I guess iterates on all the refs to find a matching one, whereas the checkout likely is among the last few log entries).

Note that it doesn't work with relative refs (because you just get a relative refname like `@~1`). In this case, `git status` falls back to the commit ID, but this is where your implementation using describe will be able to provide a better output.

Byron:
I have also been using an earlier version of this branch on Windows 11 (VM) and on MacOS 12 without problems thus far (but I am definitely not quite the power user running it off-the-shelf mostly).

Regarding @gdzx's suggestion, as reflog access is fully implemented in `gitoxide`, a similar implementation will be possible here as well. I don't know what's most desirable, but it seems like both options could work together, fetching a name from the reflog and if that fails, trying to do what's currently implemented here. Parsing reflogs of course seems quite brittle, but then again, git itself does it and the messages seem stable and [are not localized](https://github.com/git/git/blob/e4a4b31577c7419497ac30cebe30d755b97752c5/builtin/checkout.c#L931:L936) either.

matchai (approved):
It worked perfectly for me throughout the day 👍

Byron:
I think I may have found an issue with the current implementation, here is how my prompt looks like:

```
cargo-rfc-3176-ng ( fix-#10431:"refs/heads/fix-#10431") [$] via 🐍
```

I am in a worktree checkout with `HEAD` being `refs/heads/fix-#10431` and the remote configuration for it is:

```git-config
[branch "fix-#10431"]
	remote = fork
	merge = "refs/heads/fix-#10431"
```

When listing the configuration with git, it seems git is removing the quotes: `branch.fix-#10431.merge=refs/heads/fix-#10431`.

My feeling is that it should rather look like this…

```
cargo-rfc-3176-ng ( fix-#10431) [$] via 🐍
```

…but cannot because it sees the remote ref with quotes (which it shouldn't), and subsequently fails to determine that the tracking branch is the same as the local one.

I have fixed the issue in `git-config` and validated it against this PR compiled with the latest `main` (`cargo-rfc-3176-ng ( fix-#10431) [$] via 🐍`), and the fix will be available with the release of `git-repository` v0.20. I plan to release it when [rev-parsing is implemented](https://github.com/Byron/gitoxide/pull/441), but could also do it earlier to not block `starship`.

davidkna:
@Byron Per the usual starship release schedule, unless it will take more than a month, the fix shouldn't block us.

Byron:
> @Byron Per the usual starship release schedule, unless it will take more than a month, the fix shouldn't block us.

Good to know, thanks. Is there a way for me to get pinged a day or two before the release is happening? I'd make sure that a new release exists and this PR is updated then.

davidkna:
@matchai As you usually push new releases, please notify @Byron ahead of the next release.

Byron:
Maybe learning when this PR should be merged is really the call to action for me, as before then we should upgrade to the latest version of `gitoxide`. After merging, releases can happen at any time as far as I am concerned.

matchai:
@Byron With the v0.6.0 release of gitoxide, should this PR now be ready to merge?
I was planning to let you know that we should release Starship in the coming days, but it looks like you beat us to the punch :blush:

Byron:
@matchai 😁, I think you beat _me_ to the punch! I arrived here to upgrade to the latest version and deal with the breaking changes, but @davidkna was faster! I hope it wasn't too much of a hassle. Please not that the major feature for this release is a full implementation of complete `git-config`, so the git repository will pick up all the git config files that git would.
However, this also means that technically it's doing more work than it probably has to since `starship` seems to only be interested in repository-local configuration. If that's indeed the case, you can adjust the `git::open::Options` as [shown here](https://github.com/Byron/gitoxide/blob/47724c0edb382c036a3fc99884becfd2b0740d4b/git-repository/src/open.rs#L104) to pick exactly what configuration to load. I imagine this makes a measurable difference on Windows.

If it was not for the above, I see no obstacle for merging this 🎉.

PS: My gut feeling here is indeed that `starship` would want to control exactly what work is done when the `Repository` is coming up. This probably should also involve [setting the trust](https://github.com/Byron/gitoxide/blob/47724c0edb382c036a3fc99884becfd2b0740d4b/git-repository/src/open.rs#L157) to [`Full`](https://github.com/Byron/gitoxide/blob/238581cc46c7288691eed37dc7de5069e3d86721/git-sec/src/lib.rs#L22-L23) to avoid yet another disk permission check. That's all that comes to mind, everything else is lazy and I'd hope there is no waste.

davidkna:
> I arrived here to upgrade to the latest version and deal with the breaking changes, but @davidkna was faster! I hope it wasn't too much of a hassle. 

Thanks for all the work! The update didn't actually break anything. 😀

> If that's indeed the case, you can adjust the `git::open::Options` as [shown here](https://github.com/Byron/gitoxide/blob/47724c0edb382c036a3fc99884becfd2b0740d4b/git-repository/src/open.rs#L104) to pick exactly what configuration to load. I imagine this makes a measurable difference on Windows.

I've done that and set the trust to full. Did I use the API correctly? I'm not sure if I handled the permissions API the right way.

Either way, starship will likely end up needing to read the global config for `secure.directories` and do a permissions check to decide whether to run git executables against a repository. But it looks like the API will need some changes first.

src/context.rs
```diff
@@ -258,11 +259,36 @@ impl<'a> Context<'a> {
...
+    pub fn get_repo(&self) -> Result<&Repo, git::discover::Error> {
         self.repo
-            .get_or_try_init(|| -> Result<Repo, git_repository::discover::Error> {
-                let shared_repo =
-                    ThreadSafeRepository::discover_with_environment_overrides(&self.current_dir)?;
+            .get_or_try_init(|| -> Result<Repo, git::discover::Error> {
+                // custom open options
+                let git_open_opts = GitOpenOptions::default()
+                    // trust the git repo and avoid additional checks
+                    .with(git_sec::Trust::Full)
```
Byron, line 268: Actually I was wrong about this, and the way it's currently implemented this value is overridden by the trust level that was determined anyway rendering the `with(…)` adjustment ineffective.

For a moment I was about to neuter the implementation to not be so strict and all in the name of saving a permission check which right now isn't proven to save any perceivable time. All this resulted in [this commit](https://github.com/Byron/gitoxide/commit/270242c707bd086b7746ee45b55791587f1484b1) to add more information on the security implications.

For starship this means it will definitely not accidentally cause the execution of a binary controlled by an attacker, now or in the future, which I think is desirable.

That said, once actual performance measurements of the cost of the permission check (probably mostly on windows) are available, it might be worth reviewing this stance. For now I think it's good to start on the safe side.

Byron:
> I've done that and set the trust to full. Did I use the API correctly? I'm not sure if I handled the permissions API the right way.

It seems having called settings for git-config permissions was a bit of overkill, but it's coming from knowing that a simple attack vector is gaining control over git-config to cause the execution of exploits.

> Either way, starship will likely end up needing to read the global config for secure.directories and do a permissions check to decide whether to run git executables against a repository. But it looks like the API will need some changes first.

It would be very interesting for me to understand the source of this need - when starting up `safe.directory` [is respected](https://github.com/Byron/gitoxide/blob/270242c707bd086b7746ee45b55791587f1484b1/git-repository/src/open.rs#L410) by `gitoxide` for one, and secondly `gitoxide` doesn't ever refuse to operate on a repo entirely. It merely avoids to use attacker controlled configuration which, in this stage, is not an issue yet. For example, in order to get paths from the configuration, one has to use [`trusted_path()`](https://github.com/Byron/gitoxide/blob/47724c0edb382c036a3fc99884becfd2b0740d4b/git-repository/src/config/snapshot.rs#L69) which only uses values from trusted configuration files.

In any case, getting access to `safe.directory` would be possible by either not restricting access to global configuration or by loading the non-local part of the configuration by hand with [`git_config::File::from_globals()`](https://github.com/Byron/gitoxide/blob/47724c0edb382c036a3fc99884becfd2b0740d4b/git-config/src/file/init/comfort.rs#L27).

I hope that helps, any suggestions on making the API more suitable are welcome. `gitoxide` wants to cover all use cases and ideally do so while making accidental misconfiguration hard, even more so if it impacts security. To me `starship` is the 'paragon of safety' as it gets executed wherever the user may go.

davidkna:
> It would be very interesting for me to understand the source of this need - when starting up `safe.directory` [is respected](https://github.com/Byron/gitoxide/blob/270242c707bd086b7746ee45b55791587f1484b1/git-repository/src/open.rs#L410) by `gitoxide` for one, and secondly `gitoxide` doesn't ever refuse to operate on a repo entirely. It merely avoids to use attacker controlled configuration which, in this stage, is not an issue yet. For example, in order to get paths from the configuration, one has to use [`trusted_path()`](https://github.com/Byron/gitoxide/blob/47724c0edb382c036a3fc99884becfd2b0740d4b/git-repository/src/config/snapshot.rs#L69) which only uses values from trusted configuration files.
> 
> In any case, getting access to `safe.directory` would be possible by either not restricting access to global configuration or by loading the non-local part of the configuration by hand with [`git_config::File::from_globals()`](https://github.com/Byron/gitoxide/blob/47724c0edb382c036a3fc99884becfd2b0740d4b/git-config/src/file/init/comfort.rs#L27).
> 
> I hope that helps, any suggestions on making the API more suitable are welcome. `gitoxide` wants to cover all use cases and ideally do so while making accidental misconfiguration hard, even more so if it impacts security. To me `starship` is the 'paragon of safety' as it gets executed wherever the user may go.

As far as I can tell, `check_safe_directories` will be called via `open_from_paths` and in case the permissions check fails, it will completely bail out on the repository. It would be nice to be able to tell whether git would consider a repository trusted, and if it would be considered safe to run git in it (and if it would just bail out). I assumed preventing applications from running git in untrusted repositories is the primary reason libgit2 and gitoxide have permissions checking? Even further, for more security, I can imagine an option in starship to require explicit (`secure.directories`) instead of only implicit (ownership) trust.

src/context.rs
davidkna: Ok, I've removed the line. Is `git::Permissions::all()` likely to become an issue? I noticed it includes some DoS protections, which might even be desirable in trusted repos.

Byron: > Is git::Permissions::all() likely to become an issue?

You are right! It's better to go with the secure defaults for the trust map and only override the pieces that need changing. Maybe like so:

```Rust
let mut options = git_sec::trust::Mapping::default();
options.full.permissions.config = …;
options.reduced.permissions.config = …;
```

It's super valuable to do this as it seems to be easy, even for me, to suggest potentially insecure settings. I am definitely still learning.

Byron:
> As far as I can tell, check_safe_directories will be called via open_from_paths and in case the permissions check fails, it will completely bail out on the repository. It would be nice to be able to tell whether git would consider a repository trusted, and if it would be considered safe to run git in it (and if it would just bail out). 

The complete bailout is just implemented so API users who need `git` behavior can have it. `gitoxide` however is configured to not enter that path I believe that the system is able to simply not use untrusted configuration. Thus both [secure and trusted permissions](https://github.com/Byron/gitoxide/blob/1b765ec6ae70d1f4cc5a885b3c68d6f3335ba827/git-repository/src/repository/permissions.rs#L106-L116) both use full read-write permissions for the repository.

I can totally imagine this to turn into another configuration flag that says: `bail_on_untrusted_repo`, as opening a repository in read-only mode is totally orthogonal to that and might be useful in its own right. In other words, at this point the read-write setting is kind of overkill and actually not even checked for in methods that write into the repository. Definitely something to fix and I lean towards removing it entirely in favor of a `bail_on_untrusted_repo` flag or the likes.

davidkna:
@matchai The `git-repository` update should be handled now.

Byron (approved):
Looks good to me now! Thanks for your diligence and critical re-review as it ultimately helped to find suitable settings.

davidkna:
I've reverted e347e2dfcb3e9ca83dd6acfc9377990b00d9f8f3 (to be handled at a later date) because there is some undesirable overlap with the git tag in the git commit module.

PR #3884, opened 2022-04-17
refactor: Elixir dynamic variable capability

DonizeteVida:
#### Description
Add to the elixir module the capability to resolve
multiple versions' variables allowing user to
write custom toml describing which variables
to format and how to get them.

The core purpose is to make a flexible code 
which will not break every time we need another 
way to get elixir version (e.g. elixir --version, elixir --short-version),
allowing user to write a custom regex to get the variable from stdout 
which comes from 'version_command' and adjust them on format 
dynamically.

#### Motivation and Context
Closes https://github.com/starship/starship/issues/3867

PS: I have to test it using toml file. I've tested using tests 
and building the program to see how it's behaves on a real 
elixir project later my changes

#### Checklist:


atavistock:
Hmm, I'm not sure I understand how this addresses https://github.com/starship/starship/issues/3867

Stating the issue in a different way...

The default for the elixir plugin is 
```
> elixir --version
Erlang/OTP 24 [erts-12.3.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Elixir 1.13.4 (compiled with Erlang/OTP 24)
```  Executed in  368.72 millis 

It should support

```
> elixir --short-version
1.13.4
``` Executed in    9.24 millis

davidkna:
@atavistock By the way, does `--short-version` fail fast in older versions of elixir?

DonizeteVida:
> Hmm, I'm not sure I understand how this addresses https://github.com/starship/starship/issues/3867
> 
> Stating the issue in a different way...
> 
> --- 
> 
> The default for the elixir plugin is 
> ```
> > elixir --version
> Erlang/OTP 24 [erts-12.3.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]
> 
> Elixir 1.13.4 (compiled with Erlang/OTP 24)
> ```  
> Executed in  368.72 millis 
> 
> --- 
> 
> It should support
> 
> ```
> > elixir --short-version
> 1.13.4
> ``` 
> Executed in    9.24 millis 

Hi,

It addresses by letting you write your custom elixir.toml to use the "--short-version", instead of the default "--version" command. But, it returns a different response, so I've made a fell modifications to allow the user to write a custom regex to recover this variable from output, on your case "([\\d\\.\\w]+)" and adjust it properly.

My intention is to avoid every time we made a new "if" to support a new way to get this version variable.

Tell me if it is "usable" on real world projects or even if you have multiple compiler versions on your machine, because it will fail if you write your custom toml and try to use on an old compiler.

Thanks!!!

matchai:
I think the ideal approach wouldn't involve user configuration, which is where I think David's question is leading.
For instance, Rust implementation has 4 approaches to retrieving the version number, checking which would produce the quickest and most accurate version number first.

If we're able to quickly try retreiving the elixir version the quick way _and_ the `$otp_version` isn't being used, we should do just that. Otherwise, we can resort to the standard approach.

atavistock:
> @atavistock By the way, does `--short-version` fail fast in older versions of elixir?

Unfortunately on older versions it has to start the VM to pass the `--short-version` option before discovering the option is not valid.  The entire problem is that the VM has no fast start mode as most use cases are for long running daemons, so its setting up process pools and lots of other things.  I believe the `--short-version` flag is caught in a bash script before the VM even starts.

atavistock:
There is a hacky way that would fail fast, but might not be future proof.  Even though its not future proof it could fall back to the slow version.

Since the `elixir` command is really just a bash script that does a bunch of stuff before starting the VM, its would be easy to grep the version out of the script.    In a version that supports `--short-version` in the script will assign `ELIXIR_VERSION=1.13.4`.  So running `cat $(which elixir) | grep "ELIXIR_VERSION=" | cut -f 2 -d"=" ` will either give you a value or it won't.  If theres no value or if the user expects the OTP version it would need to fall back to `eleixir --version`

DonizeteVida:
Hi,

I'll rework on it. 
Sorry for no response. 
I have only weekends to work on it.

Tomorrow (today), I'll refactor it.

Thanks for the feedback!!!

atavistock:
So I'm only a little familiar with rust so I'm not the best judge, but this solution seems like a level of abstraction and dynamism that is making it more complex and difficult than it needs to be.  

I think a very simple solution would be to add a config option of `prefer_short_version: (boolean)`.
- Any non-truthy value OR if the format includes `$otp_version` then the extension simply continues to use the original flow.  
- Any true value AND if format does **not** include `$otp_version` then the extension attempts to run `elixir --short-version`, if that takes longer than 100ms then it falls back to the original flow.

DonizeteVida:
> So I'm only a little familiar with rust so I'm not the best judge, but this solution seems like a level of abstraction and dynamism that is making it more complex and difficult than it needs to be.  
> 
> I think a very simple solution would be to add a config option of `prefer_short_version: (boolean)` which defaults to `false` if not set.
> - On a false value the extension simply continues to use the original flow.  
> - On a true value AND the format includes `$otp_version`, should spit out a warning and continues to use the original flow. 
> - On a true value it attempts to run `elixir --short-version`, if that takes longer than 100ms then it falls back to the original flow.
> 

Hi, 
I stopped to work on it.

But seems no one resolved it.
I'll use your approach and use a fixed flag to activate the new *elixir version* way.

Tonight I'll update another PR.

Thanks!!!

PR #3891, opened 2022-04-18
ci: automate merging of Crowdin PR before publishing

matchai:
#### Description

Locate and merge the existing Crowdin PR as part of the deployment process.

#### Motivation and Context

Merging existing Crowdin PRs is currently a manual process that is often forgotten before a release.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -279,17 +279,10 @@ jobs:
     if: ${{ needs.release_please.outputs.release_created == 'true' }}
     continue-on-error: true
     steps:
-      - name: Setup | Find Crowdin PR
-        uses: juliangruber/find-pull-request-action@v1
-        id: find-crowdin-pr
-        with:
-          branch: i18n_master
       - name: Merge | Merge Crowdin PR
-        uses: juliangruber/merge-pull-request-action@v1
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


davidkna:
I think just running `gh pr merge --squash i18n_master` would also work.

matchai:
Much better! 😄

.github/workflows/deploy.yml
```diff
@@ -272,11 +272,22 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  publish_docs:
+    name: Publish docs to Netlify
+    runs-on: ubuntu-latest
+    needs: merge_crowdin_pr
     steps:
       - name: Setup | Checkout
         uses: actions/checkout@v3
```
davidkna, line 293: I think this needs `with: ref: master` to include the just-merged commit.

matchai: Yep! Excellent point! 👍

davidkna: And I just noticed `notarize_and_pkgbuild` will also build the docs. Maybe all deploy workflows should use it?

matchai: I've gone ahead and added the ref and dependency on the `merge_crowdin_pr` step. I think that should be fine for now 👍

davidkna (approved):

PR #3904, opened 2022-04-21
perf(package): only try to read files that exist

andytom:
#### Description

Have refactored the package module to improve performance. Before this
change the module would try to open every single file that could contain
some package information until it found a valid version. This resulted
in a lot of unneeded disk IO. Have added a new fn, `read_file_from_pwd`
that uses the current context to check if that file already exists and
fast failing if it doesn't. 

#### Motivation and Context

From my local testing this speeds up the package module from taking ~1ms to ~50µs in an empty directory.

#### Checklist:

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -48,6 +48,15 @@ pub fn read_file<P: AsRef<Path> + Debug>(file_name: P) -> Result<String> {
     result
 }
 
+/// Returns the string contents of a file from the current working directory
+pub fn read_file_from_pwd(context: &Context, file_name: &str) -> Option<String> {
+    if !context.try_begin_scan()?.set_files(&[file_name]).is_match() {
+        return None;
+    }
+
+    read_file(context.current_dir.join(file_name)).ok()
+}
+
 /// Reads command output from stderr or stdout depending on to which stream program streamed it's output
 pub fn get_command_string_output(command: CommandOutput) -> String {
     if command.stdout.is_empty() {
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -48,15 +48,6 @@ pub fn read_file<P: AsRef<Path> + Debug>(file_name: P) -> Result<String> {
     result
 }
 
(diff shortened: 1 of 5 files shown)


src/utils.rs
```diff
@@ -48,6 +48,15 @@ pub fn read_file<P: AsRef<Path> + Debug>(file_name: P) -> Result<String> {
...
 }
 
+/// Returns the string contents of a file from the current working directory
+pub fn read_file_from_pwd(context: &Context, file_name: &str) -> Option<String> {
+    if !context.try_begin_scan()?.set_files(&[file_name]).is_match() {
+        return None;
+    }
+
+    read_file(context.current_dir.join(file_name)).ok()
+}
```
davidkna, line 58: Might be a bit nice to put this into `Context` and have a log message when skipping a file.

```suggestion
impl Context {
    pub fn read_file_from_pwd(&self, file_name: &str) -> Option<String> {
        if !self.try_begin_scan()?.set_files(&[file_name]).is_match() {
            log::debug!("Not attempting to read {file_name} because, it was not included in scan.");
            return None;
        }
    
        read_file(context.current_dir.join(file_name)).ok()
    }
}
```

davidkna (commented):
The changes themselves look good.

A few other files seem to have similar `utils::read_file(context.current_dir.join("..."))` calls, so this should perhaps also cover those.

src/utils.rs
andytom: I didn't think about that but now you have mentioned it, it seems so obvious.

andytom:
> The changes themselves look good.
> 
> A few other files seem to have similar `utils::read_file(context.current_dir.join("..."))` calls, so this should perhaps also cover those.

The only one I found from a quick check was the haskell module, which I have updated, but maybe I missed something.

davidkna:
@andytom There's also the nodejs module has (`base_dir` is `current_dir`) and the terraform module (though it would be harder to replace there since it might read from a different location`).

PR #3908, opened 2022-04-24
fix(nu): use the most recent starship init

davidkna:
#### Description

The old nushell init approach caused the init only to update if the shell was opened a second time due to how nu handles sourcing (https://github.com/starship/starship/pull/3773#issuecomment-1076625914).

Thanks to @ajeetdsouza (thanks for notifying me!), there's a different approach that doesn't have this issue, by utilizing the nushell env file: https://github.com/nushell/nushell/issues/4773#issuecomment-1106114104

The default env file also contains some prompt-related configuration, so using it for this seems reasonable. The supported nushell version needed to be raised to 0.61, because that version introduced the env file feature.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ eval $(starship init ion)
 <details>
 <summary>Nushell</summary>
 
-Run the following:
+Add the following to the end of your Nushell env file (find it by running `$nu.env-path` in Nushell):
 
 ```sh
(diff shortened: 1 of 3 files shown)


andytom (approved):
LGTM

PR #3930, opened 2022-04-26
ci: explicitly provide repo owner/name when merging PR

matchai:
#### Description

Explicitly provide a repo owner/name so that the Crowdin PR could be merged by its branch name.

#### Motivation and Context

Address failure in CI where `gh` was unable to determine the base repo:
https://github.com/starship/starship/runs/6181720956?check_suite_focus=true#step:2:10

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -284,7 +284,7 @@ jobs:
     continue-on-error: true
     steps:
       - name: Merge | Merge Crowdin PR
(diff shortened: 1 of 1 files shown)


andytom (approved):

andytom:
The other option would be to checkout the repo but I think this is better since it avoids any ambiguity.

davidkna:
Just confirmed this works by merging this.

PR #3937, opened 2022-04-28
feat: add npm module

### LINKED ISSUE #557: NPM Version next to Node Version

sambehrens:
#### Description

This adds the npm module so that when in a directory with a
package-lock.json or an .npmrc, you will see the npm version displayed.
Usually this will be after the nodejs module because by default the
nodejs module will display if there is a package.json and where there is
a package-lock.json, there is almost always a package.json.

The npm module is disabled by default because it can add extra unwanted
noise to your prompt. However it can also be very useful to have in your
prompt because npm is known for having different behaviors between major
versions that will subtly break your project.

This also mimics the nodejs module when it comes to checking the
engines field in package.json. The default color is a dark red, and if
the current npm version is incompatible with the engines field in the
local package.json, it will display as the default red, to match the
color of the nodejs module when it is incompatible. Yes, npm's color is
red and when there is an error it is just a lighter red, but I didn't
know what color to make npm by default and still have an error color
that made sense 😅 .

#### Motivation and Context

As a developer who works in the node ecosystem often, I find it to be very
useful to know what version of npm that I'm on at all times. It is not so much of a problem with
other package managers like yarn and pnpm etc. but npm often breaks things
between major versions and different projects require different versions so
having the npm version available in my prompt would be very useful.

Closes #557

This module is very very similar to the nodejs module so I essentially modified the
tests that that module has to work with the npm module.
I also set up my local zsh to use the debug build so that as I was creating the module,
I could test it visually to see how it looked and make sure it was working correctly.

#### Checklist:

### DIFF
--- a/src/configs/npm.rs
+++ b/src/configs/npm.rs
@@ -1,0 +1,28 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Deserialize, Serialize)]
+#[cfg_attr(feature = "config-schema", derive(schemars::JsonSchema))]
+#[serde(default)]
+pub struct NpmConfig<'a> {
+    pub format: &'a str,
+    pub version_format: &'a str,
+    pub symbol: &'a str,
+    pub style: &'a str,
+    pub disabled: bool,
+    pub not_capable_style: &'a str,
+    pub detect_files: Vec<&'a str>,
+}
+
+impl<'a> Default for NpmConfig<'a> {
+    fn default() -> Self {
+        NpmConfig {
+            format: "via [$symbol($version )]($style)",
+            version_format: "v${raw}",
+            symbol: " ",
+            style: "bold 88",
+            disabled: true,
+            not_capable_style: "bold red",
+            detect_files: vec!["package-lock.json", ".npmrc"],
+        }
+    }
+}
--- a/src/modules/npm.rs
+++ b/src/modules/npm.rs
@@ -1,0 +1,226 @@
+use super::{Context, Module, ModuleConfig};
+
+use crate::configs::npm::NpmConfig;
+use crate::formatter::{StringFormatter, VersionFormatter};
+
+use once_cell::sync::Lazy;
+use regex::Regex;
+use semver::Version;
+use semver::VersionReq;
+use serde_json as json;
+use std::ops::Deref;
+
+/// Creates a module with the current Node.js version
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("npm");
+    let config = NpmConfig::try_load(module.config);
+    let is_npm_project = context
+        .try_begin_scan()?
+        .set_files(&config.detect_files)
+        .is_match();
+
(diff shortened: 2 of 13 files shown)


src/configs/npm.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub not_capable_style: &'a str,
+    pub detect_files: Vec<&'a str>,
+}
+
+impl<'a> Default for NpmConfig<'a> {
+    fn default() -> Self {
+        NpmConfig {
+            format: "via [$symbol($version )]($style)",
+            version_format: "v${raw}",
+            symbol: " ",
```
davidkna, line 21: Please use an emoji or a plain-text symbol here.

davidkna, line 21: Please use an emoji or a plain-text symbol here.

src/modules/npm.rs
```diff
@@ -0,0 +1,226 @@
...
+
+/// Creates a module with the current Node.js version
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("npm");
+    let config = NpmConfig::try_load(module.config);
+    let is_npm_project = context
+        .try_begin_scan()?
+        .set_files(&config.detect_files)
+        .is_match();
+
```
davidkna, line 21: Modules that are disabled by default require an explicit check for `config.disabled`.

davidkna, line 21: Modules that are disabled by default require an explicit check for `config.disabled`.

davidkna (changes requested):

davidkna:
Sorry for the late review, please implement my suggestions to get this ready to merge.

davidkna:
See discussion in https://github.com/starship/starship/issues/557

sambehrens:
Hey, sorry this totally fell off my radar but I'm willing to update the mr as I'm still interested in getting this feature.

PR #3943, opened 2022-04-28
feat(git_extensions): show what git extensions are configured on a repo

DrHyde:
#### Description
Adds a new module `git_extensions` that shows info on whether a git repository uses any git extensions. Currently it only knows about [git lfs](https://git-lfs.github.com/) and [git svn](https://git-scm.com/docs/git-svn).

#### Motivation and Context
Programmers have a little more to think about when git lfs is in use, so it's useful to show it in the prompt.

You can see behaviour without this patch in my prompt, and the `starship explain` output in the same repository with the patch applied: 
Screenshot 2022-06-07 at 18 13 00

#### Checklist:

#### Other stuff

I have refactored a bunch of repeated code from src/modules/git_{metrics,state}.rs. That's in a single commit 648f267 which might be better done in a separate PR for greater readability. Let me know if you'd prefer that.

### DIFF
--- a/src/modules/git_extensions.rs
+++ b/src/modules/git_extensions.rs
@@ -7,7 +7,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("git_extensions");
     let config: GitExtensionsConfig = GitExtensionsConfig::try_load(module.config);
 
-    let git_config = context.get_repo().ok()?.open().ok()?.config().ok()?;
+    let git_config_binding = context.get_repo().ok()?.open();
+    let git_config = git_config_binding.config_snapshot();
 
     let exts = config.extensions;
     let mut active_exts: Vec<&'a str> = vec![];
@@ -18,11 +19,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         match *ext_configured {
             "lfs" => {
                 // we don't care what the value is, merely that it exists.
-                git_config.get_entry("lfs.repositoryformatversion").ok()?;
+                git_config.string("lfs.repositoryformatversion")?;
                 active_exts.push("lfs")
             }
             "svn" => {
-                git_config.get_entry("svn-remote.svn.url").ok()?;
+                git_config.string("svn-remote.svn.url")?;
                 active_exts.push("svn")
             }
             _ => {
@@ -67,7 +68,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
 #[cfg(test)]
 mod tests {
-    use ansi_term::Color;
+    use nu_ansi_term::Color;
     use std::io;
 
     use crate::test::{create_repo, run_git_cmd, ModuleRenderer};
--- a/src/modules/git_extensions.rs
+++ b/src/modules/git_extensions.rs
@@ -1,0 +1,238 @@
+use super::{Context, Module, ModuleConfig};
+use crate::configs::git_extensions::GitExtensionsConfig;
+use crate::context::Repo;
+use crate::formatter::StringFormatter;
+
+/// Creates a module with any git extensions active in the repo at the current directory
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("git_extensions");
+    let config: GitExtensionsConfig = GitExtensionsConfig::try_load(module.config);
+
+    let repo = context.get_repo().ok()?;
+
+    let exts = config.extensions;
+    let mut active_exts = "active exts detected: ";
+
+    // iterate over exts, running functions and appending (with trailing ', ') to active_exts
+
+    // strip trailing ', '
+
+
+    let parsed = StringFormatter::new(config.format).and_then(|formatter| {
+        formatter
+            .map_meta(|variable, _| match variable {
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "active_exts" => Some(Ok(active_exts)),
+                _ => None,
+            })
+            .parse(None, Some(context))
+    });
+
+    module.set_segments(match parsed {
+        Ok(segments) => segments,
+        Err(error) => {
+            log::warn!("Error in module `git_extensions`:\n{}", error);
+            return None;
+        }
+    });
+
+    Some(module)
+}
+
+#[cfg(test)]
+mod tests {
+    use ansi_term::Color;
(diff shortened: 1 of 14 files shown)


src/modules/git_status.rs
```diff
@@ -52,6 +54,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "style" => Some(Ok(config.style)),
                 _ => None,
             })
+            .map(|variable| match variable {
+                "lfs" => context
+                    .exec_cmd("git", &["config", "--get", "lfs.repositoryformatversion"])
+                    .map(|_| Ok(config.lfs)),
```
davidkna, line 60: Please try handling this with `git2` or `git-config` for better performance. Any git operations should also pass along the repo paths, as in some instances the current directory might not match the starship working directory.

davidkna (commented):
I'd also prefer disabling this by default.

src/modules/git_status.rs
DrHyde: given that starship runs at the time a prompt is generated when would the directories not match?

chipbuster: It's true that the real-runtime conditions for this are not that common. It's possible for there to be a mismatch between cwd and the passed directory due to shenanigans with logical paths and symlinks, or because someone is parsing the output of starship as a way to get information about directories (yes, we have seen at least one real instance of this).

However, in my personal opinion, the bigger issue here is just modularity and testability of the code. Historically, many pieces of starship were tied to the environment, and once we started hitting bugs, we had to extract them into their own pieces so that we could mock those tests. The most significant example of this is `get_env()` in our context manager, but there are lots of other small examples throughout the codebase.

DrHyde: @chipbuster do you think it would be better broken out into a separate module for git extensions?

chipbuster: I don't have a great feel for what the best architecture for this is.

On one hand, it would be nice to have more modularity so that we can add other information (e.g. #3644) without having to stuff a million options into each git module. On the other hand, modules cannot communicate with each other (outside of caching information for each other in the OnceCells), so there's risk of duplicate work, as well as the problem of just having 10 different git modules that each do different things.

Can you think of other git extensions that might be added at some point?

davidkna: Re potential cwd issues: PowerShell will not set the cwd location in the prompt function. Though considering it again, `context.exec_cmd` will properly handle this. But anyway, the module has all the git directory locations at hand, so it should use that information.

DrHyde: The most obvious one would be a heuristic (based on looking at branch names) to remind the user in the prompt that `git flow` is in use in a repo.

It is unfortunately hard to search for git extensions, because there's a Windows thing called "git extensions" which appears to not be a git extension. Argh!

DrHyde:
I'm going to re-write this from scratch, as a new module `git_extensions`. I propose configuring it thus:

```toml
[git_extensions]
disabled = true
format = "($symbol $extensions)[$style]"
symbol = "git exts:"
style = "shocking pink"
git_extension_lfs_symbol = "lfs"
git_extension_lfs_disabled = false
git_extension_flow_symbol = "flow"
git_extension_flow_disabled = false
```

where `$extensions` is Magical. It will resolve to a comma separated list (eg `flow, lfs`) of the symbols of all the extensions which are enabled and detected, in ASCIIbetical order of the extension name. I'm only proposing to actually implement `lfs` support to start with, but if we want to add more in the future the rather wordy `git_extension_${extension_name}_symbol` and `git_extension_${extension_name}_disabled` variable names show the pattern that would be followed.

src/test/mod.rs
```diff
@@ -44,6 +47,100 @@ pub fn default_context() -> Context<'static> {
...
+
+    // let write_file = |text: &str| {
+    //     let mut file = OpenOptions::new()
+    //         .write(true)
+    //         .create(true)
+    //         .truncate(true)
+    //         .open(&file_in_repo)?;
+    //     writeln!(file, "{}", text)
+    // };
+
```
davidkna, line 63:
```suggestion
```

src/test/mod.rs
```diff
@@ -44,6 +47,100 @@ pub fn default_context() -> Context<'static> {
     context
 }
 
+pub fn create_repo() -> io::Result<tempfile::TempDir> {
```
davidkna, line 50: I think I might prefer if `create_repo_with_commit` stays around as a thin repo around `create_repo`.

davidkna (changes requested):

src/modules/git_extensions.rs
```diff
@@ -0,0 +1,198 @@
+use super::{Context, Module, ModuleConfig};
+use crate::configs::git_extensions::GitExtensionsConfig;
+use crate::formatter::StringFormatter;
+
+/// Creates a module with any git extensions active in the repo at the current directory
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("git_extensions");
+    let config: GitExtensionsConfig = GitExtensionsConfig::try_load(module.config);
+
+    let git_config = context.get_repo().ok()?.open().ok()?.config().ok()?;
```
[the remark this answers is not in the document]
DrHyde: There's no such method `config_snapshot()` in `git2::Repository`. Not sure what you're getting at here.

src/modules/git_extensions.rs
```diff
@@ -0,0 +1,198 @@
...
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "active_exts" => {
+                    if active_exts.is_empty() {
+                        None
```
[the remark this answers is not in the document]
DrHyde: That's in the test `show_nothing_on_repo_without_extensions`.

src/modules/git_extensions.rs
```diff
@@ -0,0 +1,198 @@
...
+            "svn" => {
+                git_config.get_entry("svn-remote.svn.url").ok()?;
+                active_exts.push("svn")
+            }
+            _ => {
+                panic!("I don't know about git extension '{}'", ext_configured)
+            }
+        };
+    }
+    active_exts.sort_unstable();
```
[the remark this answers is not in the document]
DrHyde: Done in d833412

src/modules/git_extensions.rs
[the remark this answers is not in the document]
davidkna: As you may have already noticed, this is a method provided by `git_repository::Repository`, which is used instead of `git2` now.

davidkna:
@DrHyde You will need to rebase this or merge the latest master branch.

DrHyde:
I'm going to start again and do this a bit cleaner

PR #3950, opened 2022-04-30
feat(package): Extract package version from PEP 621 compliant pyproject.toml

### LINKED ISSUE #3949: Show a package version for python projects following PEP 621

When developing in a python project that follows [PEP 621](https://peps.python.org/pep-0621/) (storing project metadata in `pyproject.toml`) the package version module is not shown.

This is because the format differs from the already implemented parser for a poetry configured python project:

**Poetry:**

```toml
[tool.poetry]
name = "project_name"
version = "0.1.0"
```

**PEP 621:**

```toml
[project]
name = "project_name"
version = "0.1.0"
```

It's worth noting that these can have directives similar to `setup.cfg` e.g. something like this in a `setup.cfg`:

```ini
version = attr: project.__version__
```

Would translate to this in a PEP 621 `pyproject.toml`:

```toml
version = {attr = "project.__version__"}
```

Which starship would want to ignore, similar to how it does already for `setup.cfg`

A parser for PEP 621 `pyproject.toml` package versions, it should be a reasonably simple change so I'll have a
go 🙂 

Live with it not showing the version, but that's no fun!

FollowTheProcess:
#### Description

This PR adds a function to the package module to extract the package version from a [PEP 621](https://peps.python.org/pep-0621/) compliant `pyproject.toml`

#### Motivation and Context

Closes #3949 

Ran `cargo test`, `cargo clippy --all-targets --all-features` and `cargo fmt`

#### Checklist:

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -66,10 +66,8 @@ fn get_node_package_version(context: &Context, config: &PackageConfig) -> Option
     Some(formatted_version)
 }
 
-fn get_poetry_version(context: &Context, config: &PackageConfig) -> Option<String> {
-    let file_contents = context.read_file_from_pwd("pyproject.toml")?;
-    let poetry_toml: toml::Value = toml::from_str(&file_contents).ok()?;
-    let raw_version = poetry_toml
+fn get_poetry_version(pyproject: &toml::Value, config: &PackageConfig) -> Option<String> {
+    let raw_version = pyproject
         .get("tool")?
         .get("poetry")?
         .get("version")?
@@ -78,10 +76,8 @@ fn get_poetry_version(context: &Context, config: &PackageConfig) -> Option<Strin
     format_version(raw_version, config.version_format)
 }
 
-fn get_pep621_version(context: &Context, config: &PackageConfig) -> Option<String> {
-    let file_contents = context.read_file_from_pwd("pyproject.toml")?;
-    let pep621_toml: toml::Value = toml::from_str(&file_contents).ok()?;
-    let version = pep621_toml.get("project")?.get("version")?;
+fn get_pep621_version(pyproject: &toml::Value, config: &PackageConfig) -> Option<String> {
+    let version = pyproject.get("project")?.get("version")?;
 
     if version.is_table() {
         None
@@ -89,6 +85,14 @@ fn get_pep621_version(context: &Context, config: &PackageConfig) -> Option<Strin
         let raw_version = version.as_str()?;
         format_version(raw_version, config.version_format)
     }
+}
+
+fn get_pyproject_version(context: &Context, config: &PackageConfig) -> Option<String> {
+    let file_contents = context.read_file_from_pwd("pyproject.toml")?;
+    let pyproject_toml: toml::Value = toml::from_str(&file_contents).ok()?;
+
+    get_pep621_version(&pyproject_toml, config)
+        .or_else(|| get_poetry_version(&pyproject_toml, config))
 }
 
 fn get_setup_cfg_version(context: &Context, config: &PackageConfig) -> Option<String> {
@@ -265,8 +269,7 @@ fn get_version(context: &Context, config: &PackageConfig) -> Option<String> {
         get_cargo_version,
         get_nimble_version,
         get_node_package_version,
-        get_poetry_version,
-        get_pep621_version,
+        get_pyproject_version,
         get_setup_cfg_version,
         get_composer_version,
         get_gradle_version,
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -78,6 +78,19 @@ fn get_poetry_version(context: &Context, config: &PackageConfig) -> Option<Strin
     format_version(raw_version, config.version_format)
 }
 
+fn get_pep621_version(context: &Context, config: &PackageConfig) -> Option<String> {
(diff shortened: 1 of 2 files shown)


src/modules/package.rs
```diff
@@ -78,6 +78,19 @@ fn get_poetry_version(context: &Context, config: &PackageConfig) -> Option<Strin
...
+    let version = pep621_toml.get("project")?.get("version")?;
+
+    if version.is_table() {
+        None
+    } else {
+        let raw_version = version.as_str()?;
+        format_version(raw_version, config.version_format)
+    }
+}
+
```
davidkna, line 93: Can you combine this with the poetry version function to avoid reading and parsing the file twice? By the way, if both versions are present, which is expected to take precedence?

FollowTheProcess: Yeah sure that makes perfect sense. AFAIK if both a `[tool.poetry]` and a `[project]` table are present then it's an invalid `pyproject.toml` so I think in that case it's fine for starship not to show either?

FollowTheProcess:
I'm afraid my noob level rust is showing, I've tried a few different variations on the following but can't seem to get it to work:

```rust
fn get_pyproject_version(context: &Context, config: &PackageConfig) -> Option<String> {
    let file_contents = context.read_file_from_pwd("pyproject.toml")?;
    let pyproject_toml: toml::Value = toml::from_str(&file_contents).ok()?;

    if let Some(poetry_version) = pyproject_toml
        .get("tool")?
        .get("poetry")?
        .get("version")?
        .as_str()
    {
        format_version(poetry_version, config.version_format);
    }

    if let Some(pep621_version) = pyproject_toml.get("project")?.get("version")?.as_str() {
        format_version(pep621_version, config.version_format);
    }

    None
}
```

I get the following test failures:

image

So both a poetry pyproject.toml and a pep621 pyproject.toml are returning `None` and I can't see why. In my head the code above should check if there's a:

```toml
[tool.poetry]
version = "0.1.0"
```

If so, grabbing the version and formatting it.

If not, check for a:

```toml
[project]
version = "0.1.0"
```

If so, grabbing the version and formatting it. And if version is a table here, the `.as_str()` should return `None` according to the docs.

And finally if we get here and haven't found anything, return a `None` anyway.

Obviously my understanding is wrong otherwise I wouldn't be here! Can anyone help me out? 🙂

davidkna:
When encountering `None` `?` will return `None` from the function and stop proceeding. You can either write two additional functions for both approaches to ensure the function `?` does not stop execution (`get_poetry_version(&pyproject_toml).or_else(|| get_pep612_version(&pyproject_toml))`) or you could handle the `None`-values without returning, for instance with `and_then`.
```rust
pyproject_toml
        .get("tool")
		.and_then(|tool| tool.get("poetry"))
        .and_then(|poetry| poetry.get("version"))
        .as_deref()
```

FollowTheProcess:
@davidkna Worked perfectly! Thanks for the help

FollowTheProcess:
The way I've written this atm is it will prioritise a PEP621 pyproject.toml so in the weird case where someone has both a `[project]` and a `[tool.poetry]`, it will show the version from the `[project]` table and ignore the `[tool.poetry]` which I think is sensible behaviour as PEP 621 is the formal standard and is currently supported in a number of python build backends.

Poetry is the odd one out here and if/when it eventually adopts PEP 621, poetry pyproject.toml files will replace `[tool.poetry]` with `[project]` anyway.

src/modules/package.rs
```diff
@@ -78,6 +76,19 @@ fn get_poetry_version(context: &Context, config: &PackageConfig) -> Option<Strin
...
+    format_version(raw_version, config.version_format)
+}
+
+fn get_pyproject_version(context: &Context, config: &PackageConfig) -> Option<String> {
+    let file_contents = context.read_file_from_pwd("pyproject.toml")?;
+    let pyproject_toml: toml::Value = toml::from_str(&file_contents).ok()?;
+
+    get_pep621_version(&pyproject_toml, config)
+        .or_else(|| get_poetry_version(&pyproject_toml, config))
+}
```
davidkna, line 90: This looks good, but I would prefer handling `format_version` in `get_pyproject_version`:
```rust
fn get_poetry_version<'a>(pyproject: &'a toml::Value) -> Option<&'a str> {
    pyproject
        .get("tool")?
        .get("poetry")?
        .get("version")?
        .as_str()
}

fn get_pep621_version<'a>(pyproject: &'a toml::Value) -> Option<&'a str> {
    pyproject.get("project")?.get("version")?.as_str()
}

fn get_pyproject_version(context: &Context, config: &PackageConfig) -> Option<String> {
    let file_contents = context.read_file_from_pwd("pyproject.toml")?;
    let pyproject_toml: toml::Value = toml::from_str(&file_contents).ok()?;

    get_pep621_version(&pyproject_toml)
        .or_else(|| get_poetry_version(&pyproject_toml))
        .and_then(|raw_version| format_version(raw_version, config.version_format))
}
```

davidkna (approved):

src/modules/package.rs
FollowTheProcess: Done 👍🏻 When I implemented your snippet with the explicit lifetime annotations I got a clippy warning saying that could be elided so I just did it without them

andytom:
Thank you for your contribution @FollowTheProcess and thanks for reviewing @davidkna.

PR #3970, opened 2022-05-06
feat(hg_branch): Add support for mercurial topics and locating .hg in parent directories

### LINKED ISSUE #870: hg_branch module doesn't show up when I am in subdirectory
Simple problem: hg_branch module doesn't show up in subdirectories
```
/tmp ❯ hg init test

/tmp ❯ cd test

/tmp/test on  default ❯ mkdir subdir

/tmp/test on  default ❯ cd subdir

/tmp/test/subdir ❯ 
```
The latter prompt should also show Mercurial branch, we are inside repo.

I fixed this in my fork: https://github.com/Mekk/starship/commit/d405b5197b6586260f5ceedb95f0d5f9b6a85030 but I also have some doubts:

- I feel it would make sense to make some „utility” module shared between hg_branches, hg_commit and hg_status (starting from the aforementioned function which checks whether we are inside mercurial repo and where is this repo toplevel) – but I am not sure what is the current policy (so far I cross import from hg_commit to hg_status and partially copy&paste to hg_branch, but it isn't pretty)

- this check is currently re-done by every of my modules, it would be nice to cache it somehow so they reuse the same calculation. I don't know what is the approach towards it (there is some repo attribute in context but it seems tailored towards git…)

### LINKED ISSUE #1384: [mercurial] support for topic (lightweight branches)

tomchuk:
#### Description

*Repo Detection*
The current `hg_branch` module only displayed the branch/bookmark information when in the `hg root`. Both the detection of whether the module should run and the retrieval of the branch/bookmark information were tied to `context.current_dir` being the root of the repo.

I've added `find_dothg_path` and used that to feed into the existing functions branch/bookmark functions so that they continue to work anywhere under the root of the repository.

*Topics Support*
I've added topic detection to the existing branch/bookmark detection.

#### Motivation and Context
To me the repo detection part is obvious - the git module does this, why not the hg one? Concerning [topics](https://www.mercurial-scm.org/wiki/TopicPlan) support, topics/evolve are the current best practice for short-lived, git-style "branches" in mercurial. My team has been using them for a few years now.

This duplicates some topics work in #1461 which I only realized exists as I write this - but it appears to have been abandoned for some time now.
Closes #1384
Closes #870

CleanShot 2022-05-06 at 16 50 45

Been running these changes locally on MacOS & Linux for a week or so. Added a test for topics, but I'll probably need to update the fixture bundle to be able to add more tests for the repo detection. I'm very new to the Rust ecosystem, and I've searched for docs about how to go about modifying/replacing the bundle but have come up blank. Any guidance would be appreciated.

This will also require the installation of the topics extension to mercurial in the test environment, and I'm also unsure about the best way to do that.

#### Checklist:

I've never written a line of rust before today, so I'm very open to criticism and feedback. Also some guidance on testing as described above would be appreciated.

### DIFF
--- a/src/modules/hg_branch.rs
+++ b/src/modules/hg_branch.rs
@@ -91,18 +91,18 @@ fn find_dothg_path(cwd: &PathBuf) -> Option<PathBuf> {
     }
 }
 
-fn get_hg_branch_name(dothg_path: &PathBuf) -> String {
+fn get_hg_branch_name(dothg_path: &Path) -> String {
     std::fs::read_to_string(dothg_path.join("branch"))
         .map_or_else(|_| "default".to_string(), |s| s.trim().into())
 }
 
-fn get_hg_current_bookmark(dothg_path: &PathBuf) -> Option<String> {
+fn get_hg_current_bookmark(dothg_path: &Path) -> Option<String> {
     std::fs::read_to_string(dothg_path.join("bookmarks.current"))
         .map(|s| s.trim().into())
         .ok()
 }
 
-fn get_hg_current_topic(dothg_path: &PathBuf) -> Option<String> {
+fn get_hg_current_topic(dothg_path: &Path) -> Option<String> {
     std::fs::read_to_string(dothg_path.join("topic"))
         .map(|s| s.trim().into())
         .ok()
--- a/src/modules/hg_branch.rs
+++ b/src/modules/hg_branch.rs
@@ -211,6 +211,20 @@ mod tests {
 
     #[test]
     #[ignore]
+    fn test_hg_topic() -> io::Result<()> {
+        let tempdir = fixture_repo(FixtureProvider::Hg)?;
+        let repo_dir = tempdir.path();
+        run_hg(&["topics", "feature-topic"], repo_dir)?;
+        expect_hg_branch_with_config(
+            repo_dir,
+            None,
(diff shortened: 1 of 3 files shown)


src/modules/hg_branch.rs
```diff
@@ -74,13 +75,34 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     Some(module)
 }
 
-fn get_hg_branch_name(ctx: &Context) -> String {
-    std::fs::read_to_string(ctx.current_dir.join(".hg").join("branch"))
+fn find_dothg_path(cwd: &PathBuf) -> Option<PathBuf> {
```
davidkna, line 78: Does mercurial support something like `$GIT_CEILING_DIRECTORIES` (#3591)? It would be nice to avoid crossing devices at least.

tomchuk: Unfortunately, it does not. This is the mercurial python code that locates the `.hg` directory.

```
def findrepo(p):
    while not os.path.isdir(os.path.join(p, b".hg")):
        oldp, p = p, os.path.dirname(p)
        if p == oldp:
            return None
    return p
```

davidkna: In that case, please change the implementation to stay on the same device if possible, and perhaps split that off into a new PR. 

It also looks like your hg topic changes broke some of the existing tests. Please fix that.

PR #3973, opened 2022-05-07
ci: Set MacOS deployment target at 10.7 in CI

### LINKED ISSUE #3945: Error due to older MacOS version
## Bug Report

I am on macOS Mojave `10.14.6`. I installed starship and set it up for my bash environment.

#### Current Behavior

When opening my bash terminal I receive the following error:
```bash
dyld: Symbol not found: _OBJC_CLASS_$_NSConstantDictionary
  Referenced from: /usr/local/bin/starship (which was built for Mac OS X 11.0)
  Expected in: /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
 in /usr/local/bin/starship
```

**I also receive this error if I try any `starship` commands such as `--version` or `bug-report`**

#### Expected Behavior

Should open the default starship command prompt with no errors (i've tried it on a ubuntu vm and it works fine there)

#### Environment
- Starship version: (unable to get version due to the `starship` command not working however starship was downloaded only a day or two ago)
- Shell type: Bash
- Shell version: `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)`
- Shell plugin manager: n/a
- Operating system: MacOS 10.14.6

```bash
# Fig pre block. Keep at the top of this file.
. "$HOME/.fig/shell/bashrc.pre.bash"

eval "$(starship init bash)"

# Fig post block. Keep at the bottom of this file.
. "$HOME/.fig/shell/bashrc.post.bash"
```

#### Starship Configuration
Default configuration is used so I don't have a `starship.toml` file.

chipbuster:
#### Description
Sets MACOSX_DEPLOYMENT_TARGET to 10.7 to force backwards compatibility for x64 MacOS builds.

I have tested this on #3945 and it does resolve the error. I've also tested it for aarch64 builds and it doesn't screw up cargo at all (a binary is generated which correctly targets MacOS 11.0 even though the deployment target is set lower). I'm assuming that non-MacOS builds won't look at this variable, so it shouldn't affect them.

In the absence of this variable, [rust is supposed to target 10.7 as a minimum version on x64 anyways](https://github.com/rust-lang/rust/blob/6af09d2505f38e4f1df291df56d497fb2ad935ed/compiler/rustc_target/src/spec/apple_base.rs#L68), and it's not doing so right now. ~I was unable to find anything in [the documentation on the MacOS runners](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md) as to why this might not be working. In short, I don't know if this is masking a deeper issue.~

EDIT: Discussion on the Rust Discourse [suggests that -sys crates may cause this issue when they invoke a native compiler](https://users.rust-lang.org/t/compile-rust-binary-for-older-versions-of-mac-osx/38695). We have a few of these crates in our dependency tree (notably, git2 and notifications) so that's probably the cause.

#### Motivation and Context

Closes #3945

#### Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,6 +8,7 @@ env:
   CARGO_NET_RETRY: 10
   RUST_BACKTRACE: short
   RUSTUP_MAX_RETRIES: 10
+  MACOSX_DEPLOYMENT_TARGET: 10.7
 
 jobs:
   # Update release PR
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -15,6 +15,7 @@ env:
   RUST_BACKTRACE: short
   RUSTFLAGS: "-D warnings"
   RUSTUP_MAX_RETRIES: 10
+  MACOSX_DEPLOYMENT_TARGET: 10.7
 
 jobs:
   # Run the `rustfmt` code formatter


.github/workflows/deploy.yml
```diff
@@ -8,6 +8,7 @@ env:
   CARGO_NET_RETRY: 10
   RUST_BACKTRACE: short
   RUSTUP_MAX_RETRIES: 10
+  MACOSX_DEPLOYMENT_TARGET: 10.7
```
davidkna, line 11: Just in case, please also add this to `workflow.yml`.

chipbuster: Done.

davidkna (approved):

davidkna:
Thanks for the fix and researching the cause!

PR #3983, opened 2022-05-12
ci: Add daily Crowdin Pretranslation trigger

### LINKED ISSUE #3790: [Crowdin] Starship translations are regularly invalidated
According to Crowdin, there are 4,359 strings and 20 488 words right now.

E-mails I received this month (omitting the smaller ones):
- 2022-03-07: 257 new strings for translation (3,202 words).
- 2022-03-12: 2,396 new strings for translation (11,622 words).
- 2022-03-18: 1,547 new strings for translation (7,117 words).
- 2022-03-25: 2,486 new strings for translation (11,042 words).

Needless to say, it’s impossible to follow, even with Crowdin «Translation Memory».

---

Some examples of invalidated strings that I can confirm have been translated before include:
- Lines from the README that haven’t been changed since 3 years, like `Copyright © 2019-present, [Starship Contributors](https://github.com/starship/starship/graphs/contributors).<br>` or the alt text of `Starship rocket icon`
- Strings like ```The `cmake` module shows the currently installed version of [CMake](https://cmake.org/).``` (last modified 11 months ago) or `## Conda` (last modified 3 years ago) and dozens of dozens of others like this. It’s just impossible to translate the Configuration page because every once in a while half of it got reset.

I should have finished the translation by now (modulo the new modules and legit changes), but instead French is at 55%. I won’t contribute to the translations anymore at least until this is resolved, because it’s pretty much useless as of right now.

chipbuster:
#### Description
Adds a Github CI action which fires every 24 hours, running "Pre-Translate with Translation Memory" on the CrowdIn project.

#### Motivation and Context

Closes #3790

The action has been run against a test repository which triggered on push, so I've created a lot of pre-translate requests, all of which appear to have gone through. It seems that if there have been no translations suggested on Crowdin since the last pre-translate, the translation is a no-op, so there's not a great way for me to guarantee it's working with this repo until we get some more suggestions on Crowdin.

#### Checklist:

### DIFF
--- a/.github/workflows/crowdin-pretranslate.yml
+++ b/.github/workflows/crowdin-pretranslate.yml
@@ -9,7 +9,7 @@ jobs:
     name: Crowdin Translation Memory Trigger
     runs-on: ubuntu-latest
     steps:
-      - uses: starship/crowdin-pretranslate-action@main
+      - uses: starship/crowdin-pretranslate-action@v0.1.0
         with:
           project_id: 372655
           api_key: ${{ secrets.CROWDIN_API_KEY }}
--- a/.github/workflows/crowdin-pretranslate.yml
+++ b/.github/workflows/crowdin-pretranslate.yml
@@ -1,0 +1,15 @@
+# Run pre-translate with translation memory, all files/langs, at midnight daily.
+name: Crowdin Updates
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  trigger_crowdin_tm:
+    name: Crowdin Translation Memory Trigger
+    runs-on: ubuntu-latest
+    steps:
+      - uses: starship/crowdin-pretranslate-action@main
+        with:
+          project_id: 372655
+          api_key: ${{ secrets.CROWDIN_API_KEY }}


.github/workflows/crowdin-pretranslate.yml
```diff
@@ -0,0 +1,15 @@
...
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  trigger_crowdin_tm:
+    name: Crowdin Translation Memory Trigger
+    runs-on: ubuntu-latest
+    steps:
+      - uses: starship/crowdin-pretranslate-action@main
```
matchai, line 12: I think it would be preferred to publish the action so that we can confidently version it with semver 😄

matchai (approved):
LGTM! 👍

.github/workflows/crowdin-pretranslate.yml
chipbuster: Sure thing! I think I can just push tags and it'll create a version (we don't have to publish on the marketplace).

I'm thinking `v0.1.0` for now--a full-featured action should allow you to ignore certain files/languages as well as trigger pre-translation via machine translation (with the arguments that entails) but I don't really want to write/maintain that if nobody is going to use it.

PR #3987, opened 2022-05-14
feat: Add Kerberos module

edamato:
#### Description

Add a module that displays whether user has a valid kerberos ticket based on the output of the standard kerberos tool on linux `klist`. This tool has the `-s` option, which is designed to be used in shell scripts etc to check for whether a valid ticket is present.

#### Motivation and Context

This is a feature request. In AD/krb auth environments users often have failures on commands, etc and have to debug why, only to find out that their kerberos ticket had expired. This feature allows users to see very quickly that they have a valid krb5 ticket.

Screenshot 2022-05-14 at 17 02 04

- [] I have tested using **MacOS**
- [] I have tested using **Windows**

#### Checklist:

#### Caveats:

This has only been tested in a a setup with a single kerberos domain and no trust involved. This should not affect anything unless users have very customised setups.


andytom (changes requested):
Can you update the [module configuration docs](https://github.com/starship/starship/tree/master/docs/config) to include the new module. You only need to update the English version translations are handled by a separate system.

PR #3996, opened 2022-05-18
docs: Add MPR installation instructions to README.md

hwittenborn:
#### Description

This PR adds installation instructions for the [makedeb Package Repository](https://mpr.makedeb.org), an AUR-like platform for Debian and Ubuntu based distributions.

#### Motivation and Context

This brings native .deb packaging for Starship. It won't bring Starship into the Debian/Ubuntu repositories themselves, but it provides a means of distribution for .deb packages at the least.

N/A

#### Checklist:


davidkna:
I would rather not add this at this time, as the package manager isn't very popular and somebody is already working to package starship officially in debian. Ubuntu users also have the option to use the snap package.

PR #4028, opened 2022-05-26
refactor(memory_usage): replace `sys-info` with `systemstat`

davidkna:
#### Description

This PR replaces `sys-info` with `systemstat`.

#### Motivation and Context

systemstat is still at 0.1 (which is why I specified the version with `=`), but I feel much better about the codebase (it's much less likely to get us another rustsec warning) and the support for different operating systems is comparable to sys-info.

While working on this, I also tried improving the test coverage, but it still doesn't verify the complete module outputs.

Related?: #3951

#### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ notify = ["notify-rust"]
 
 [dependencies]
 ansi_term = "0.12.1"
-byte-unit = "4.0.14"
 chrono = "0.4.19"
 clap = { version = "3.1.18", features = ["derive", "cargo", "unicode"] }
 clap_complete = "3.1.4"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -70,7 +69,7 @@ shadow-rs = "0.11.0"
 # see: https://github.com/svartalf/rust-battery/issues/33
 starship-battery = { version = "0.7.9", optional = true }
 strsim = "0.10.0"
-sys-info = "0.9.1"
+systemstat = "=0.1.11"
```
matchai, line 72: What does the equal mean within the quotes?
I can't seem to find it in the Cargo.toml reference.

matchai (approved):
LGTM 👍

Cargo.toml
davidkna: The `=` ensures only the specified version is allowed, and newer versions won't be used even without `--locked`.

matchai: Is there a reason that we're specifically locking this dependency's version?

davidkna: Because it's still at 0.1, there might be API-breaking patch releases.

PR #4031, opened 2022-05-26
feat(release): add windows msi installers

davidkna:
#### Description

This PR adds the required configurations files and changes to the release workflow to publish MSI installers. Windows aarch64 builds don't get installers yet, because they're not yet supported by wix.
As there's no easy way to include MSVC-redist in the installer, this PR also changes the Windows builds to use static linking. This is the approach recommended by cargo-wix.

#### Motivation and Context

Required for #1295
Related #690

#### Checklist:


.github/workflows/release.yml
```diff
@@ -93,13 +98,28 @@ jobs:
           profile: minimal
           target: ${{ matrix.target }}
 
+      - name: Setup | Install cargo-wix [Windows]
```
matchai, line 101: A passing thought — not something we have to address in this PR:
Our release-time build process is getting pretty complex with building installers and notarizing, which leads to a lot of unexpected failures at the time of release.

For this, in particular, I think there'd be value in adding this to CI. Maybe not on every commit, like the rest of our suite, but perhaps on every push to `master`.

.github/workflows/release.yml
```diff
@@ -93,13 +98,28 @@ jobs:
...
+        # aarch64 is only supported in wix 4.0 development builds
+        if: matrix.os == 'windows-latest' && matrix.target != 'aarch64-pc-windows-msvc'
+        run: cargo install --version 0.3.2 cargo-wix
```
matchai, line 104: It's unfortunate that this, being a hard-coded version, won't be updated by Dependabot.
I wish we could have cargo subcommands be retrieved from our dev-dependencies. That's not possible though, right?

matchai (approved):
It looks to me like `icon.ico` is translucent. Is that intentional?

Left a couple of comments, but nothing blocking. LGTM! 👍

davidkna:
> It looks to me like `icon.ico` is transparent. Is that intentional?

`rsvg-convert` and `inkscape` keep giving me transparent images for some reason. I'll need to investigate a bit more.

matchai:
@davidkna It looks like the icon svg in the repo is transparent. That's probably my fault 😅 
Let me fix that up.

matchai:
@davidkna Done! Feel free to try using the updated `icon.svg`

.github/workflows/release.yml
davidkna: Done.

.github/workflows/release.yml
davidkna: I don't think that's possible.

matchai (approved):

PR #4043, opened 2022-06-05
feat: Add the ability to have some file extensions *prevent* a module from triggering

### LINKED ISSUE #4033: Starship detects MPEG transport stream files as Javascript
#### Current Behavior
Starship detects MPEG transport stream .ts files in the Node module and if present puts a Node symbol in the prompt.

#### Expected Behavior
The node module should ignore MPEG transport stream files as they are not source code:

```
$ file *.ts
some-random-stream-capture.video.ts: MPEG transport stream data
```

#### Possible Solution
Refer to the Unix file type recognition database. I assume that there's a Rust implementation:

```
$ grep MPEG\ transport\ stream /usr/share/file/magic/*
/usr/share/file/magic/animation:>188	byte			0x47		MPEG transport stream data
```

Alternatively, as I have only ever seen these files with extensions `.video.ts` and `.audio.ts` make it possible to exclude those from the `.ts` default.

#### Environment
- Starship version: 1.6.3
- bash version: GNU bash, version 5.1.16(1)-release (x86_64-apple-darwin20.6.0)
Copyright (C) 2020 Free Software Foundation, Inc.

- Operating system: Mac OS 11.6.5
- Terminal emulator: iTerm.app 3.4.14
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.59.0
- Rust channel:  release
- Build Time: 2022-04-26 18:54:07 +00:00

```bash
eval "$(starship init bash)"
```

#### Starship Configuration

```toml
add_newline = false

[c]
detect_extensions = ["c", "h", "xs"]
format = "using [$symbol]($style)"

[character]
error_symbol = "[\\$](bold red)"
success_symbol = "[\\$](bold green)"

[cmd_duration]
disabled = true

[custom.git_lfs]
command = "grep -q ^.lfs .git/config && echo git lfs"
format = "[($output )](bold purple)"
when = "[ -e .git/config ]"

[directory]
truncate_to_repo = true
truncation_length = 0
truncation_symbol = " …/"

[git_branch]
format = "[on $branch ]($style)"

[hostname]
format = "[$hostname]($style):"

[line_break]
disabled = true

[perl]
detect_extensions = ["pl", "pm"]
format = "using [$symbol ($version )]($style)"
symbol = "perl"

[username]
format = "[$user]($style)@"


```

DrHyde:
This adds "negative extensions" to `detect_extensions`. A negative extension, signified by starting with a "!" character, will prevent a module from triggering.

#### Description
You can now add "negative extensions" to any `detect_extensions` variable. If any file is found that matches any negative extension then the module will not be triggered.

I also added support for multi-part extensions like "tar.gz" and "video.ts" because for my particular use case I need to prevent the Node module from triggering on "video.ts" files.

#### Motivation and Context

Closes #4033

Aside from adding unit tests, I have also tested this for real with `detect_extensions = ["ts", "!video.ts"]` in my config and verified that after adding `"!video.ts"` the directory was no longer showing up as a Node project in my prompt.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -152,6 +152,27 @@ format = '''
 \$'''
 ```
 
+### File Extensions
+
+Many modules have a `detect_extensions` variable, which takes a list of file extensions to match or not match.
+"Negative" extensions, those which should not be matched, are indicated with a leading "!" character.
+
+Extensions are matched against both the characters after the last dot in a filename, and the characters after
+the first dot in a filename. For example, consider the filename `foo.bar.tar.gz`. To see if that matches anything
+mentioned in `detect_extensions` we will consider both the characters after the last dot, "gz", and those after
+the first dot, "bar.tar.gz".
+
+We first look for any files matching a negative extension. If there are any, the module is not triggered.
+
+If there were none, we then look for any files matching a positive extension.
+
+To see how this works in practice, and why you might want to use negative extensions, consider how a Node
+programmer who also works with video might configure the Nodejs module. By default its `detect_extensions`
+is `["js", "mjs", "cjs", "ts", "mts", "cts"]`. However, in their video work they also deal with `.video.ts`
+and `.audio.ts` files containing MPEG Transport Stream encoded data. They don't want a directory full of
+video data to trigger the Nodejs module! So they add two negative extensions to the list:
+`["js", "mjs", "cjs", "ts", "mts", "cts", "!audio.ts", "!video.ts"]`.
+
 ## Prompt
 
 This is the list of prompt-wide configuration options.
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -152,26 +152,23 @@ format = '''
 \$'''
 ```
 
-### File Extensions
+### Negative matching
 
-Many modules have a `detect_extensions` variable, which takes a list of file extensions to match or not match.
-"Negative" extensions, those which should not be matched, are indicated with a leading "!" character.
+Many modules have `detect_extensions`, `detect_files`, and `detect_folders` variables. These take
+lists of strings to match or not match. "Negative" options, those which should not be matched, are
+indicated with a leading "!" character. The presence of _any_ negative indicator in the directory
+will result in the module not being matched.
 
-Extensions are matched against both the characters after the last dot in a filename, and the characters after
-the first dot in a filename. For example, consider the filename `foo.bar.tar.gz`. To see if that matches anything
-mentioned in `detect_extensions` we will consider both the characters after the last dot, "gz", and those after
-the first dot, "bar.tar.gz".
+Extensions are matched against both the characters after the last dot in a filename, and the
+characters after the first dot in a filename. For example, `foo.bar.tar.gz` will be matched
+against `bar.tar.gz` and `gz` in the `detect_extensions` variable. Files whose name begins with a
+dot are not considered to have extensions at all.
 
-We first look for any files matching a negative extension. If there are any, the module is not triggered.
+To see how this works in practice, you could match TypeScript but not MPEG Transport Stream files thus:
 
-If there were none, we then look for any files matching a positive extension.
-
-To see how this works in practice, and why you might want to use negative extensions, consider how a Node
-programmer who also works with video might configure the Nodejs module. By default its `detect_extensions`
-is `["js", "mjs", "cjs", "ts", "mts", "cts"]`. However, in their video work they also deal with `.video.ts`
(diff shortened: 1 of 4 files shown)


src/context.rs
```diff
@@ -397,8 +397,23 @@ impl DirContents {
...
                             .map(|ext| extensions.insert(ext.to_string_lossy().to_string()));
+
+                        // find the full extension on a file. ie, the tar.gz in foo.tar.gz
+                        path.file_name().map(|file_name| {
+                            file_name
+                                .to_string_lossy()
+                                .to_string()
+                                .split_once('.')
+                                .map(|(_, after)| extensions.insert(after.to_string()))
+                        });
```
davidkna, line 416: We will need to evaluate if this will have a performance impact when there are a large number of files.

docs/config/README.md
```diff
@@ -152,6 +152,27 @@ format = '''
...
+We first look for any files matching a negative extension. If there are any, the module is not triggered.
+
+If there were none, we then look for any files matching a positive extension.
+
+To see how this works in practice, and why you might want to use negative extensions, consider how a Node
+programmer who also works with video might configure the Nodejs module. By default its `detect_extensions`
+is `["js", "mjs", "cjs", "ts", "mts", "cts"]`. However, in their video work they also deal with `.video.ts`
+and `.audio.ts` files containing MPEG Transport Stream encoded data. They don't want a directory full of
+video data to trigger the Nodejs module! So they add two negative extensions to the list:
+`["js", "mjs", "cjs", "ts", "mts", "cts", "!audio.ts", "!video.ts"]`.
```
davidkna, line 174: Please shorten this text a bit.

davidkna (changes requested):
Perhaps also update the Deno and Typescript modules to make them mutually exclusive (#2627).

src/context.rs
```diff
@@ -397,8 +397,23 @@ impl DirContents {
...
+
+                        // find the minimal extension on a file. ie, the gz in foo.tar.gz
                         path.extension()
                             .map(|ext| extensions.insert(ext.to_string_lossy().to_string()));
+
+                        // find the full extension on a file. ie, the tar.gz in foo.tar.gz
+                        path.file_name().map(|file_name| {
+                            file_name
+                                .to_string_lossy()
+                                .to_string()
```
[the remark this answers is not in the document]
DrHyde: That pattern appears a few times in the project, including elsewhere in context.rs. I'll investigate them all and see if there's any others that can be tidied up.

src/context.rs
DrHyde: `hyperfine -w 10 -r 200 "target/debug/starship prompt" "target/debug/starship-without-multi-exts prompt"` says that there's about a 2% +/- 10% difference in runtime in a directory with 500 files in on this 'ere machine, although from one run to the next it can't tell whether it's faster with or without that `map`. IOW, it's too fast to measure and is lost in the noise.

src/context.rs
```diff
@@ -448,8 +463,15 @@ impl DirContents {
...
     }
 
-    pub fn has_any_extension(&self, exts: &[&str]) -> bool {
-        exts.iter().any(|ext| self.has_extension(ext))
+    pub fn has_any_positive_extension(&self, exts: &[&str]) -> bool {
+        exts.iter()
+            .any(|ext| !ext.starts_with('!') && self.has_extension(ext))
+    }
+
+    pub fn has_no_negative_extension(&self, exts: &[&str]) -> bool {
```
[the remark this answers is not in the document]
DrHyde: Good idea

DrHyde:
> Perhaps also update the Deno and Typescript modules to make them mutually exclusive (#2627).

c1394fd7 does that. I tried to come up with a way of reading the Deno config into the Nodejs config. I could append its default `detect_files` vector to Nodejs's default `detect_files`, but couldn't figure out a way of appending the files each wit ha leading `!`, as Rust's variable lifetime checking kept getting in the way.

davidkna:
Sorry for only getting back to this now.

I'd indeed prefer if they don't exclude each other in this manner. How about just manually adding the extensions filenames prefixed with `!` to each module instead? Not sure if it's worth it, but a test could ensure that both configs are complete.

The other changes look fine and mergeable, if you would rather not handle the deno/nodejs problem, you could also exclude the changes.

DrHyde:
I've reverted the mutual incompatibility for now. I _really_ don't like the idea of copying config from one module to another, so I'll try to figure out a way of doing it without that, and if I can I'll raise another PR for that.

davidkna (approved):

davidkna:
Thanks for the PR @DrHyde.

PR #4046, opened 2022-06-05
fix(install): ignore tarfile ownership values when installing as root

### LINKED ISSUE #4039: ownership of starship binary created by install.sh
#### Current Behavior

After running `curl -sS https://starship.rs/install.sh | sh` the installed starship binary keeps its ownership from the archive:
```
ls -lan /usr/local/bin/starship
-rwxr-xr-x 1 1001 121 7355184 May 25 00:51 /usr/local/bin/starship
```

#### Expected Behavior

The starship binary should probably be owned by root or by myself, not whatever local user happens to match the uid/gid in the archive.

Latest starship, bash shell on Ubuntu 22.04.

DrHyde:
If the install script used `sudo` to install Starship from a tarball, it left the file with what the userid/groupid was in the tarball. That user/group might or might not actually exist. This change fixes that.

#### Description
We now use the `-o` flag when untarring, to force tar to use the current user's userid/groupid instead of whatever is in the tarball. That is on by default for non-root users, but not for root.

#### Motivation and Context
Closes #4039

- [X] I have tested the install script on **MacOS**
- [X] I have tested the install script on **Linux**

The script claims to support FreeBSD but doesn't - it can't find a binary to download.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -113,6 +113,30 @@ unpack() {
     *.tar.gz)
       flags=$(test -n "${VERBOSE-}" && echo "-xzvf" || echo "-xzf")
       ${sudo} tar "${flags}" "${archive}" -C "${bin_dir}"
+      if [ "$sudo" != "" ]; then
+        # set the owner to root instead of whatever was in the tarball
+        "${sudo}" chown root "${bin_dir}/starship"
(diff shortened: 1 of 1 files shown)


davidkna:
I think you can just use the `tar` `-o` flag.

andytom (approved):

PR #4059, opened 2022-06-12
build: upgrade open for non-blocking 'open::that()' on linux

Byron:
#### Description
A dependency upgrade, no breaking changes.


davidkna:
Thanks for the contribution, but please also update the lockfile.

Byron:
What a great catch! I forced-pushed the commit this time with `Cargo.lock` included.

davidkna (approved):

PR #4092, opened 2022-06-19
feat(gcloud): disable if active config is NONE

### LINKED ISSUE #3658: Feature Request: Allow gcloud to be hidden when the active config is NONE

cnpryer:
Adds additional match condition to return None for active config if set to "NONE".

#### Motivation and Context
Closes #3658
Closes #3692

(See [comment](https://github.com/starship/starship/pull/4092#issuecomment-1159818107))

#- [ ] I have tested using **MacOS**
#- [ ] I have tested using **Linux**
#- [ ] I have tested using **Windows**

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1380,7 +1380,7 @@ AA -------------------------------------------- BB -----------------------------
 ## Google Cloud (`gcloud`)
 
 The `gcloud` module shows the current configuration for [`gcloud`](https://cloud.google.com/sdk/gcloud) CLI.
-This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gcloud/configurations/config_{CONFIG NAME}` file and the `CLOUDSDK_CONFIG` env var.
+This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gcloud/configurations/config_{CONFIG NAME}` file and the `CLOUDSDK_CONFIG` env var. Setting `active` to `"NONE"` will disable the module.
 
 ### Options
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1380,7 +1380,7 @@ AA -------------------------------------------- BB -----------------------------
 ## Google Cloud (`gcloud`)
 
 The `gcloud` module shows the current configuration for [`gcloud`](https://cloud.google.com/sdk/gcloud) CLI.
-This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gcloud/configurations/config_{CONFIG NAME}` file and the `CLOUDSDK_CONFIG` env var. Setting `active` to `"NONE"` will disable the module.
+This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gcloud/configurations/config_{CONFIG NAME}` file and the `CLOUDSDK_CONFIG` env var. Setting active name to `"NONE"` will disable the module.
 
 ### Options
 
--- a/src/modules/gcloud.rs
+++ b/src/modules/gcloud.rs
@@ -88,6 +88,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let gcloud_context = GcloudContext::new(&config_name, &config_path);
     let account: Lazy<Option<Account<'_>>, _> = Lazy::new(|| gcloud_context.get_account());
 
+    // Disable the Module if gcloud's active config is set to "NONE".
+    // This is a valid gcloud configuration.
+    if gcloud_context.config_name == "NONE" {
+        return None;
+    }
+
     let parsed = StringFormatter::new(config.format).and_then(|formatter| {
         formatter
             .map_meta(|variable, _| match variable {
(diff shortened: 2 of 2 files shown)


cnpryer:
Hey @davidkna lmk if weekend pings are no good :)

Any pointers on a test for this one? Seems tricky but probably doable and I'm too raw here. I'd imagine we can still use a tempfile to write the config
```toml
[core]
active = NONE
```
and use `ModuleRender` similarly to the other tests, but I'm not sure yet.

davidkna:
Something like that. Try starting with one of the existing tests (there should be one for the regular existing case), but instead expecting None.

cnpryer:
But isn't this supposed to disable the module? I'm imagining the test here would be test that Model is None. Maybe I'm misunderstanding something.

cnpryer:
I gave the test a shot. The last few remaining areas for me are 

1. Is the docs change here what we'd expect?
2. Are we fully capturing the feature request from @nathanscain?
3. Is this sufficient testing? I'm not a gcloud user, I'm on mac, and the only method tested here would be the actual tests. From the PR template it seems there's a kind of platform testing?

Thanks for the help!

src/modules/gcloud.rs
```diff
@@ -124,7 +124,9 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                             .map_or(project, Cow::Borrowed)
                     })
                     .map(Ok),
-                "active" => Some(Ok(Cow::Borrowed(&gcloud_context.config_name))),
+                "active" if gcloud_context.config_name != "NONE" => {
+                    Some(Ok(Cow::Borrowed(&gcloud_context.config_name)))
+                }
```
davidkna, line 129: I think I misread the original request. I think they wanted the whole module to return `None` if `gcloud_context.config_name` is `NONE`, which would be handled via a regular `if`-condition ahead of the `let parsed` part.:
```
if gcloud_context.config_name == "NONE" {
    return None;
}
```

The documentation for this would need to be in the main description after `# Google Cloud`, like the conditions listed in the AWS module.

cnpryer: I'll convert this to a draft and make the adjustment. Thanks! I was thinking maybe there was a little confusion but wasn't sure enough of the ask.

I'm thinking I'd need a more clever test, no? Idk if this tests that the module is *disabled*. Which I'd think is a Module that's None somehow? I can play with it but any advice here would be super appreciated. :)

davidkna: The `ModuleRenderer` will return `None` if the module has no output. I think the test is fine, assuming it fails if `active` is not set to `NONE`.

cnpryer:
Note that I did not set "NONE" as a constant here. Lmk if that's preferred.

src/modules/gcloud.rs
cnpryer: Ready to mark this as resolved if you are. I can give it a double take tomorrow. I think this achieves what we'd like. Would love a gcloud user to validate this PR though. 😇

docs/config/README.md
```diff
@@ -1380,7 +1380,7 @@ AA -------------------------------------------- BB -----------------------------
 ## Google Cloud (`gcloud`)
 
 The `gcloud` module shows the current configuration for [`gcloud`](https://cloud.google.com/sdk/gcloud) CLI.
-This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gcloud/configurations/config_{CONFIG NAME}` file and the `CLOUDSDK_CONFIG` env var.
+This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gcloud/configurations/config_{CONFIG NAME}` file and the `CLOUDSDK_CONFIG` env var. Setting `active` to `"NONE"` will disable the module.
```
cnpryer, line 1383: Trying to read up on the "NONE" config mentioned. I'm not sure if it's `active = "NONE"` or if it's the active *name* is `"NONE".

So maybe:

```suggestion
This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gcloud/configurations/config_{CONFIG NAME}` file and the `CLOUDSDK_CONFIG` env var. Setting active name to `"NONE"` will disable the module.
```

davidkna:
```suggestion
This is based on the `~/.config/gcloud/active_config` file and the `~/.config/gcloud/configurations/config_{CONFIG NAME}` file and the `CLOUDSDK_CONFIG` env var. If the active config name in `~/.config/gcloud/active_config` is set to `"NONE"`, the module will be disabled.
```

src/modules/gcloud.rs
```diff
@@ -88,6 +88,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let gcloud_context = GcloudContext::new(&config_name, &config_path);
     let account: Lazy<Option<Account<'_>>, _> = Lazy::new(|| gcloud_context.get_account());
 
+    // Disable the Module if gcloud's active config is set to "NONE".
+    // This is a valid gcloud configuration.
+    if gcloud_context.config_name == "NONE" {
+        return None;
```
davidkna, line 94: This line isn't covered by the tests, which likely means the test doesn't work as expected.

cnpryer: So I'm looking at this now. Originally I used `region_not_set_with_display_region` as a reference.
https://github.com/starship/starship/blob/b9746100e2ee3746ea418e26ebc50494710a9d03/src/modules/gcloud.rs#L396-L410

Looking into `ModuleRenderer`, isn't `collect()` supposed to return `None` for tests like these? 
https://github.com/starship/starship/blob/b9746100e2ee3746ea418e26ebc50494710a9d03/src/test/mod.rs#L156-L159

Maybe I'm missing something. I can look at how `module(context)` is utilized, but I want to make sure I'm on the right track first.

davidkna: You'll need to base the test on something that would usually return `Some(...)`, but returns `None` instead because the active config is `NONE`.

PR #4109, opened 2022-06-23
feat: Add operating system module

### LINKED ISSUE #4292: Module for retrieving the Operating System / Kernel
## Feature Request

I'm currently using Windows Terminal with several WSL instances and starship. Sometimes I'm switching between Tabs or returning to the Window and I don't know in which environment I am at the moment. Besides the Hostname I think it would be neat to have a Module that can show the operating system and/or kernel.

exincore:
#### Description

The `os` module shows the current operating system.
OS information is detected via the [os_info](https://lib.rs/crates/os_info) crate.

Disabled by default due to `os_info` inaccuracies for the time being.

NOTE: [os_info v3.5.0](https://github.com/stanislav-tkach/os_info/releases/tag/v3.5.0) should have much improved detection, but this remains to be confirmed 

#### Motivation and Context

Inspired by `os_icon` 'segment' of [romkatv/powerlevel10k](https://github.com/romkatv/powerlevel10k)

Fixes #4292

#### Screenshots:
image
image
image

#### Checklist:


andytom:
I have only had a quick look but think we need to think about the default configuration here we are moving away from using nerd fonts (see #3544). However happy to see this module happening we seem to get a lot of questions/requests for this feature.

exincore:
Ah, well I have exactly what you request (probably)! I spent a good few hours coming up with a reasonable list of emojis that represent the operating systems instead of nerd font. I have yet to finalize my edits to the Presets themselves, but you can see the emoji list is implemented as one of the `modules/os.rs` tests

docs/config/README.md
```diff
@@ -2544,6 +2545,111 @@ style = "bold yellow"
...
+This module has an advanced configuration option that controls how the operating system symbol is displayed.
+</summary>
+
+| Advanced Option | Default | Description                                            |
+| --------------- | ------- | ------------------------------------------------------ |
+| `symbols`       |         | A table that maps each operating system to its symbol. |
+
+`symbols` allows you to define arbitrary symbols to display for each operating system type.
+Operating system types not defined by your configuration use the default symbols table below.
+See [os_info::Type](https://docs.rs/os_info/latest/os_info/enum.Type.html) for the operating system types currently recognized by this module.
```
davidkna, line 2579: I would rather list the supported OS-names directly here, and suggest users create feature requests in starship for unsupported operating systems.

src/configs/os.rs
```diff
@@ -0,0 +1,71 @@
+use indexmap::{indexmap, IndexMap};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Deserialize, Serialize)]
+#[cfg_attr(feature = "config-schema", derive(schemars::JsonSchema))]
+#[serde(default)]
+pub struct OSConfig<'a> {
+    pub format: &'a str,
+    pub style: &'a str,
+    pub symbols: IndexMap<String, &'a str>,
```
davidkna, line 10: This doesn't need to be owned.
```suggestion
    pub symbols: IndexMap<&'a str, &'a str>,
```

davidkna (changes requested):

src/modules/os.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+fn get_symbol<'a>(config: &'a OSConfig, os: &os_info::Info) -> Option<&'a str> {
+    // String from os_info::Type
+    let key = &format!("{:?}", os.os_type());
+    config
+        .symbols
+        .get(key)
+        .cloned()
+        .or_else(|| OSConfig::default().symbols.get(key).cloned())
```
davidkna, line 60: Please add a comment explaining this line.

src/configs/os.rs
exincore: I followed the solution that the `directory` module uses for its own `IndexMap` as the deserializer parses the config into an owned string and panics.
```
[WARN] - (starship::config): Failed to load config value: Error in 'OS' at 'symbols': invalid type: string "Alpine", expected a borrowed string
```
`HashMap` also has the same error. Though, switching to `HashMap` is likely better regardless of this issue since the index and order does not matter for the module.

davidkna: Interesting, the `ValueDeserializer` is supposed to forward a borrowed string. `IndexMap` is useful as it keeps the `print-config` ordering predictable, though the std `BTreeMap` type should also handle this.

docs/config/README.md
```diff
@@ -2544,6 +2545,111 @@ style = "bold yellow"
 symbol = "☁️ "
 ```
 
+## OS
+
+The `os` module shows the current operating system.
+OS information is detected via the [os_info](https://lib.rs/crates/os_info) crate.
```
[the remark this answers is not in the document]
exincore: I'm unaware about a 'snaps' issue; is this Canonical's Snap you are referring to?

exincore:
Docs may be typeset weirdly; I don't think this is terribly important though.
image

docs/config/README.md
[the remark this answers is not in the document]
exincore: I assume @davidkna is referring to #3933 , whose bug report is bugged as OracleLinux

exincore:
[os_info 3.5.0](https://github.com/stanislav-tkach/os_info/releases/tag/v3.5.0) has much improved os detection reliability :partying_face:

src/configs/os.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub disabled: bool,
+}
+
+impl<'a> OSConfig<'a> {
+    pub fn get_symbol(&self, key: &str) -> Option<&'a str> {
+        self.symbols.get(key).cloned()
```
davidkna, line 16: Looking at it this again, I think I would prefer if this table would work case-insensitively. But we can also change this later in case the case-sensitive version ends up confusing users.

src/configs/os.rs
```diff
@@ -0,0 +1,77 @@
+use indexmap::{indexmap, IndexMap};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Deserialize, Serialize)]
+#[cfg_attr(feature = "config-schema", derive(schemars::JsonSchema))]
```
davidkna, line 5: Please rebase/merge master and add `schemars(deny_unknown_keys)` here, like in the other config modules.

davidkna (changes requested):
Sorry, I completely forgot this wasn't merged yet! Please change the documentation a bit and change the OsConfig to align with recent changes to similar modules.
I also think the symbols table lookups should be case-insensitive, but you don't need to change that, for me to approve this PR.

src/configs/os.rs
exincore: Working on it!

davidkna (approved):
LGTM

exincore:
I'm looking into adding a test with clippy lint that will fail if new `Type` cases are added

exincore:
Here's a potential refactor:

By making the config use `IndexMap<os_info::Type, &'a str>` if there are invalid keys within the config, it errors on deserialization and therefore tells the user that they have typed the key incorrectly. While this reintroduces case sensitivity, it prints the list of correct values, without the user needing to consult the docs, and it can alert the user to spelling errors.

```
[WARN] - (starship::config): Failed to load config value: Error in 'OS' at 'symbols': unknown variant
`amazon`, expected one of `Alpine`, `Amazon`, `Android`, `Arch`, `CentOS`, `Debian`, `DragonFly`,
`Emscripten`, `EndeavourOS`, `Fedora`, `FreeBSD`, `Garuda`, `Gentoo`, `HardenedBSD`, `Illumos`,
`Linux`, `Macos`, `Manjaro`, `Mariner`, `MidnightBSD`, `Mint`, `NetBSD`, `NixOS`, `OpenBSD`,
`openSUSE`, `OracleLinux`, `Pop`, `Raspbian`, `Redhat`, `RedHatEnterprise`, `Redox`, `Solus`, `SUSE`,
`Ubuntu`, `Unknown`, `Windows`
```

davidkna:
@exincore That sounds good. Does this require #3941?

exincore:
@davidkna No; `IndexMap`'s deserializer seem to handle the `enum`.

davidkna:
@exincore I think it works because table keys are handled via the serde string deserializer instead of the starship value-type. Please update the PR to use the enum, it should make it easier to spot mistakes.

exincore:
@davidkna I think the CI tests screwed up? Could you run them again?

davidkna (approved):

andytom (approved):
LGTM

andytom:
Thank you for your contribution @exincore and thank you for reviewing @davidkna.

fadeevab:
@andytom @exincore It's inspiring that Nerd fonts have actual logos for OSes:
image

There are Windows and Apple logs too.

**What about to utilize them instead of standard emojis?** You already use Nerd-specific symbols for Rust and NodeJS and other.

Here is how I trick the _shell_ module:
image

Although, I find pretty convenient to use specific brand icons for `cmd` and `powershell` while using Ubuntu symbol for `bash`.

image

From this point of view, **it would be cool to introduce an `$os` variable to the `shell` module to use OS as an icon** for some shells.

exincore:
@fadeevab The list of Nerd Font operating system symbols is provided within the preset "Nerd Font Symbols"; one may copy the `[os.symbols]` section into one's own `starship.toml`. Starship is moving towards emoji as default, so that Nerd Font is not a default requirement; see #3544 . 

As for implementing `os` in the `shell` module, that will likely not happen. I actually use both the `os` and `shell` module side by side to get a quick glance at what commands I can expect to use on a particular system. 

image

In this case, Fedora with [Fish](https://fishshell.com/).

fadeevab:
@exincore Gotcha. If you don't mind, I have a question: now, the `shell` module default symbols are actual words like `bsh`, `fsh`, etc. Is there a chance to make emoji/nerd icons by default in the `shell` module, so it would look similarly to your screenshot? It would be more consistent to overall ideology and design (we have a Nerd icon for Rust, not `rst` work, don't we).

exincore:
@fadeevab I apologize for the delay. I would suggest bringing these topics up in the Starship discord server. I am merely a lowly contributor and I do not manage design decisions.

PR #4112, opened 2022-06-25
feat: Add starship preset command

### LINKED ISSUE #4006: Install presets with starship CLI

Currently, there is no code-free way of installing themes or presets. I noticed that official presets are shared on this [page](https://starship.rs/presets/). However, it requires manually copying and pasting codes. 

It would be a much better experience if presets can be managed through CLI just like how oh-my-fish manages themes.

Flygrounder:
#### Description

Added `starship preset` subcommand:
- `starship preset -l` lists all available presets.
- `starship preset <name>` prints a preset with a specified name.

#### Motivation and Context

Closes  #4006

изображение

#### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,6 +110,7 @@ nix = { version = "0.24.1", default-features = false, features = ["feature", "fs
 
 [build-dependencies]
 shadow-rs = "0.12.0"
+dunce = "1.0.2"
 
 [target.'cfg(windows)'.build-dependencies]
 winres = "0.1.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,6 @@ quick-xml = "0.23.0"
 rand = "0.8.5"
 rayon = "1.5.3"
 regex = "1.6.0"
-rust-embed = "6.4.0"
 rust-ini = "0.18.0"
 semver = "1.0.12"
 serde = { version = "1.0.138", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,7 @@ quick-xml = "0.23.0"
 rand = "0.8.5"
 rayon = "1.5.3"
 regex = "1.5.6"
+rust-embed = "6.4.0"
 rust-ini = "0.18.0"
 semver = "1.0.10"
 serde = { version = "1.0.137", features = ["derive"] }
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,4 @@
-use std::fmt::Debug;
+use dunce;
 use std::fs::{self, File};
 use std::io::Write;
 
@@ -27,7 +27,7 @@ fn gen_presets_hook(mut file: &File) -> SdResult<()> {
     for path in paths {
         let unwrapped = path?;
         let file_name = unwrapped.file_name();
-        let full_path = fs::canonicalize(unwrapped.path())?;
+        let full_path = dunce::canonicalize(unwrapped.path())?;
         let full_path = full_path.to_str().expect("failed to convert to string");
         let name = file_name
             .to_str()
--- a/build.rs
+++ b/build.rs
@@ -1,7 +1,11 @@
-use std::error::Error;
+use std::fs::{self, File};
+use std::io::Write;
 
-fn main() -> Result<(), Box<dyn Error>> {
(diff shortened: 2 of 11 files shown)


Flygrounder:
I've opened it as a draft because I am not sure whether I should write tests for this kind of functionality. All tests I can think of will be either flaky or will test nothing useful.

Flygrounder:
I cannot hit the required patch coverage because diff is very small and large part of it is in `main.rs`, which is not covered by tests. What am I supposed to do with that?

Cargo.toml
```diff
@@ -60,6 +60,7 @@ quick-xml = "0.23.0"
 rand = "0.8.5"
 rayon = "1.5.3"
 regex = "1.5.6"
+rust-embed = "6.4.0"
```
davidkna, line 63: I think at this time I would prefer `include_bytes!`, but this is acceptable.

davidkna (changes requested):

src/print.rs
```diff
@@ -450,6 +452,57 @@ pub fn print_schema() {
...
 
+#[derive(clap::ValueEnum, Clone, Debug)]
+pub enum Preset {
+    BracketedSegments,
+    NerdFontSymbols,
+    NoRuntimeVersions,
+    PastelPowerline,
+    PlainTextSymbols,
+    PurePreset,
+}
```
davidkna, line 463: I actually meant handling the validation via a type that manually implement the `clap::ValueEnum` trait, by looking at the values in `Presets`, instead of the manual validation you had before.
If the implementation does require presets to be added to the code manually, please don't use `rust-embed`.

Flygrounder: I am quite new to Rust so there might be another way around. The only way I can think of implementing `fn value_variants<'a>() -> &'a [Self]` method without explicitly listing all the options would be to create a procedural macro. As far as I understand, proc macros require a separate crate to maintain, which would be an unreasonable thing to do for this small feature.

I liked the idea of not listing presets manually, but if my understanding is correct, I will rewrite it using `include_bytes!`.

davidkna: I don't think it needs to be an enum, `struct Preset(String)` should work without macros.

Flygrounder: I've tried implementing it this way, the problem that I have is a lifetime of a slice. If I create it's value in runtime inside `value_variants`, then it's lifetime is limited to this function and thus I can't return it. So it has to be a compile time constant

davidkna: Could `struct Preset(&'static str)` work?

Flygrounder: That solves part of the problem. The other part is from where `&[Preset]` will be borrowed from. The only thing I can think of is using something like `static mut PRESETS: Vec<Preset> = Vec::new() `, changing it in the method and returning it's slice, but that requires `unsafe` block and overall feels ugly

Flygrounder:
I've managed to implement `ValueEnum` by generating definitions in `build.rs`. Also `rust-embed` is not needed anymore because I just generate code with `include_bytes!`. @davidkna thanks for suggestion

davidkna (changes requested):

Flygrounder:
I've addressed issues you've found. But now diff outside of `main.rs` is too small to hit patch test coverage :sweat_smile:

build.rs
```diff
@@ -13,3 +17,55 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+            .to_str()
+            .and_then(|v| v.strip_suffix(".toml"))
+            .expect("Failed to process filename");
+        presets.push_str(format!("print::Preset(\"{}\"),\n", name).as_str());
+        match_arms.push_str(
+            format!(
+                r#"
+"{name}" => {{
+        let mut stdout = io::stdout().lock();
+        stdout.write_all(include_bytes!(r"{full_path}")).expect("Failed to print preset");
```
davidkna, line 42: `expect` makes for poor error messages (the other instances in this PR should never be hit, so they should be fine). I think doing nothing is preferable here, as printing an error message when something like `head` is used is unexpected. (#885)

```suggestion
        let _ = stdout.write_all(include_bytes!(r"{full_path}"));
```

davidkna (approved):

davidkna:
Thanks for the PR @Flygrounder.

PR #4114, opened 2022-06-26
Double application of style fix in pastel-powerline

WhiteBlackGoose:
#### Description
```
[blabla]($style)
```
already applies `style` to `blabla` (assuming `$style` exists), so there's no need to add the same style again.

Just fixed the config, and nothing changed, so it seems to be working/

#### Checklist:


davidkna (approved):

davidkna:
Thanks for the fix @WhiteBlackGoose!

PR #4115, opened 2022-06-26
ci(release): fix msi upload

davidkna:
#### Description

The upload of MSI files failed in the last release. One issue that both of the MSI had the same name. For some reason, the git checkout was also causing some issues with the upload (I did some testing, but with `gh` instead of `softprops/action-gh-release@v1` because it was easier). `softprops/action-gh-release@v1` is supposed to be able to recognize the repo without any checkout via the GitHub env.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -120,7 +120,7 @@ jobs:
         uses: actions-rs/cargo@v1.0.3
         with:
           command: wix
-          args: -v --nocapture -I install/windows/main.wxs --target ${{ matrix.target }} --output target/wix/starship.msi
+          args: -v --no-build --nocapture -I install/windows/main.wxs --target ${{ matrix.target }} --output target/wix/starship-${{ matrix.target }}.msi
 
       - name: Post Build | Prepare artifacts [Windows]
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -120,7 +120,7 @@ jobs:
         uses: actions-rs/cargo@v1.0.3
         with:
           command: wix
-          args: -v --nocapture -I install/windows/main.wxs --target ${{ matrix.target }} --output target/wix/starship.msi
+          args: -v --no-build --nocapture -I install/windows/main.wxs --target ${{ matrix.target }} --output target/wix/starship-${{ matrix.target }}.msi
```
matchai, line 123: Is the addition of `--no-build` intentional here?

davidkna: Yes, it's just for a tiny speed-up – If you prefer, I could remove it again.

matchai: No problem! Just making sure it was intentional.

matchai (approved):

PR #4121, opened 2022-06-27
fix(timings): count time spent on custom on 'when' command failure

### LINKED ISSUE #4093: Custom module 'when' execution time not counted if it returns an error
#### Current Behavior
If the `when` entry of a custom command returns a non-zero exit code then the execution time of the command is not reported in the `starship timings` output.

#### Expected Behavior
The execution time should be reported whether or not the `when` command failed. After all it does have an effect on how long it takes to draw the prompt.

If you have a custom command like this
```toml
[custom.test]
when = 'sleep 0.1; true'
shell = ["bash", "--noprofile", "--norc"]
format = "TEST "
```
it gets reported correctly in the timings report:
```
~ TEST
$ starship timings

 test        -  104ms  -   "TEST "
 battery     -    6ms  -   ""
 directory   -   <1ms  -   "~ "
 line_break  -   <1ms  -   "\n"
 character   -   <1ms  -   "$ "
```

However if you change the `true` to `false` it is no longer reported:
```toml
[custom.test]
when = 'sleep 0.1; false'
shell = ["bash", "--noprofile", "--norc"]
format = "TEST "
```
```
~
$ starship timings

 battery     -   6ms  -   ""
 directory   -  <1ms  -   "~ "
 line_break  -  <1ms  -   "\n"
 character   -  <1ms  -   "$ "
```

I used https://github.com/romkatv/zsh-prompt-benchmark to verify that the module does indeed still take ~100ms in this case.

#### Environment
- Starship version: 1.8.0
- zsh version: zsh 5.9 (x86_64-debian-linux-gnu)
- Operating system: Debian unstable
- Terminal emulator: tmux 3.2a
- Git Commit Hash: c9f74f7f
- Branch/Tag: master
- Rust Version: rustc 1.61.0 (fe5b13d68 2022-05-18)
- Rust channel:  release
- Build Time: 2022-06-16 12:53:48 +00:00

majutsushi:
#### Description
Return a `Module` instance with timing information for custom modules even if the `when` execution returns `false`.

#### Motivation and Context

Closes #4093.
Previously the custom module evaluation function would return early if the `when` execution returned `false`, but the timing information was only added when returning a module instance at the end of the function. This meant that slow `when` commands would invisibly slow down the prompt. This change causes the function to not return early in this case and instead return a proper module instance with timing information whether or not the `when` command indicates that the module should be shown.

I tested this manually with the custom command I described in #4093. I wasn't sure whether I should add a test since it would probably need to have small sleep in it, which isn't ideal for unit tests. I did run the existing tests though.

#### Checklist:

### DIFF
--- a/src/modules/custom.rs
+++ b/src/modules/custom.rs
@@ -34,6 +34,8 @@ pub fn module<'a>(name: &str, context: &'a Context) -> Option<Module<'a>> {
         }
     }
 
+    let mut module = Module::new(name, config.description, Some(toml_config));
+
     let mut is_match = context
         .try_begin_scan()?
         .set_extensions(&config.detect_extensions)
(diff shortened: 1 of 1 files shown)


majutsushi:
I don't know why the diff looks so weird, I just wrapped the rest of the function in an `if` which only changed the indentation.

majutsushi:
I had a quick look at testing since I realized that it wouldn't actually be necessary to sleep in a test, but currently the tests are purely based on checking the output, so to be able to verify that a module is returned that contains a duration but no output would require some test refactoring.

davidkna:
Thanks for handling this @majutsushi.

PR #4125, opened 2022-06-27
ci: fix release hiccups caused by draft release

matchai:
#### Description

Fix issues encountered during the release of v1.9.1:
- Draft release couldn't be found by `mislav/bump-homebrew-formula-action`
  Fixed by providing a GitHub token with permissions to view draft release
- Draft release was not updated by `softprops/action-gh-release`
  Fixed by publishing the draft release before adding artifacts

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -265,12 +265,16 @@ jobs:
       - name: Setup | Checksums
         run: for file in starship-*/starship-*; do openssl dgst -sha256 -r "$file" | awk '{print $1}' > "${file}.sha256"; done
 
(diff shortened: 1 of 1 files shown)


davidkna:
The winget action also requires a homebrew-pr style token to create a PR.

matchai:
The winget action happens after the release is published, so I think it should be fine.

davidkna:
@matchai I think it also had insufficient permissions (but not because of the drafting), it likely needs a personal access token instead of `GITHUB_TOKEN` like the homebrew action https://github.com/starship/starship/runs/7079837899

PR #4138, opened 2022-07-01
feat: Add support for blink, hidden, and strikethrough styles.

dtanabe:
#### Description

Add support for `blink`, `hidden`, and `strikethrough` format strings. The underlying `ansi_term` library supports them on supported terminals; this PR merely exposes those options.

#### Motivation and Context

I find it useful to occasionally use blinking text in my prompts to _really_ get me to pay attention, so it would be nice to be able to style things using blinking text. I noticed strikethrough and hidden could be added trivially, so I threw them into the mix too.

  * I noticed that iTerm supports blinking and strikethrough, but does _not_ support hidden text (https://gitlab.com/gnachman/iterm2/-/issues/4564).
  * Terminal.app _does_ support hidden text, but does not appear to support strikethrough text.
  * Tested with Gnome Terminal; supports all three of `blink`, `hidden`, and `strikethrough`
  * Tested both in the traditional terminal (Windows Console Host) and the new one (Windows Terminal); both support all three of `blink`, `hidden`, and `strikethrough`

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -278,6 +278,9 @@ Style strings are a list of words, separated by whitespace. The words are not ca
 - `underline`
 - `dimmed`
 - `inverted`
+- `blink`
+- `hidden`
+- `strikethrough`
 - `bg:<color>`
 - `fg:<color>`
 - `<color>`
@@ -297,3 +300,9 @@ A color specifier can be one of the following:
 - A number between 0-255. This specifies an [8-bit ANSI Color Code](https://i.stack.imgur.com/KTSQa.png).
 
 If multiple colors are specified for foreground/background, the last one in the string will take priority.
+
+Not every style string will be displayed correctly by every terminal. In particular, the following known quirks exist:
+
+- Many terminals disable support for `blink` by default
+- `hidden` is not supported on iTerm (https://gitlab.com/gnachman/iterm2/-/issues/4564).
+- `strikethrough` is not supported by the default macOS Terminal.app
--- a/src/config.rs
+++ b/src/config.rs
@@ -268,6 +268,7 @@ where
  - 'bold'
  - 'italic'
  - 'inverted'
+ - 'blink'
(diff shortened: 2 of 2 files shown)


dtanabe:
> `blink` doesn't seem to work for me in `iTerm`. Is this feature disabled by default? Please change the docs to mention that these options may not be supported.

Done; additionally tested on Linux and Windows too and the three new settings seem to be well-supported on those platforms.

docs/advanced-config/README.md
```diff
@@ -236,3 +239,12 @@ A color specifier can be one of the following:
...
 If multiple colors are specified for foreground/background, the last one in the string will take priority.
+
+Most terminals allow you to customize whether blinking text is supported:
+
+- On iTerm, blinking is turned off by default.
+- In Terminal.app, iTerm, and most Linux terminals, go to "Profiles > Text > Allow blinking text" under the preferences for your terminal application if blinking text is not working for you.
+
+`hidden` is not supported on iTerm (https://gitlab.com/gnachman/iterm2/-/issues/4564).
+
+`strikethrough` is not supported on Mac OS Terminal.app.
```
davidkna, line 250:
```suggestion

Not every style string will be displayed correctly by every terminal, in particular the following quirks exist:
- Many terminals disable support for `blink` by default
- `hidden` is not [supported by iTerm](https://gitlab.com/gnachman/iterm2/-/issues/4564).
- `strikethrough` is not supported by the default macOS Terminal.app
```

davidkna (commented):
Just a nitpick regarding the docs, otherwise this looks good to me.

davidkna (approved):

davidkna:
Thanks for the PR and thoroughly testing this @dtanabe.

dtanabe:
> Thanks for the PR and thoroughly testing this @dtanabe.

And thanks for `starship` too! It has really improved my terminal experience too, especially as I switch between prompts/ecosystems all the time.

PR #4142, opened 2022-07-02
feat(username): Adds substitution for usernames

colings86:
#### Description
Adds `username.substitutions` to the configuration which can be used to substitute usernames in a similar way to the directory module.

Example configuration:
```toml
[username]
disabled = false
show_always = true
format = '[$user]($style)'

[username.substitutions]
"colings86" = "🤷"
"root" = "🛡"

```

#### Motivation and Context
For common usernames I find it useful to still show an indicator of the user but to have an icon or something short that doesn't take up lots of space.

Screenshot 2022-07-02 at 09 42 04

The PR adds tests for the new functionality and I tested on my local shell prompt.

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1403,7 +1403,8 @@
         "format": "[$user]($style) in ",
         "show_always": false,
         "style_root": "red bold",
-        "style_user": "yellow bold"
+        "style_user": "yellow bold",
+        "substitutions": {}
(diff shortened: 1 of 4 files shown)


colings86:
Just saw the failure on the schema check. I hadn’t seen that there is a schema file. I’ll update that later today or tomorrow.

PR #3442, opened 2022-01-15
feat(pwsh): Set PromptText (syntax_error_prompt)

segevfiner:
#### Description
Set `PromptText` using a new `prompt --syntax-error` and related config, similar to the continuation prompt. This enables a different prompt when there is a syntax error.

There is one problem though. This adds a call to `module character` to every prompt invocation (As it varies depending on the status), which might slow things down, is there any better way to do this?

Also the exact choice of what the format/color of the syntax error should be is left open for debate. It can't be the normal red as that is used for the previous command failing.

#### Motivation and Context

Closes #2069

#### Screenshots (if appropriate):
starship-syntax-error-prompt

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [x] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #4143, opened 2022-07-02
feat: Enable transience for Cmd and PowerShell

rashil2000:
#### Description

Enables both left and right transience for Cmd.
Enables only left transience for PowerShell.

#### Motivation and Context

Related to #888

#### Asciinema Recording:
[asciicast](https://asciinema.org/a/Vl7k1E3IpRFBvpV33Y1h1OyUv?autoplay=1)

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -64,6 +64,35 @@ $null = New-Module starship {
         $process.StandardOutput.ReadToEnd();
     }
 
+    function Enable-TransientPrompt {
+        Set-PSReadLineKeyHandler -Key Enter -ScriptBlock {
+            $previousOutputEncoding = [Console]::OutputEncoding
+            try {
+                $parseErrors = $null
+                [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$null, [ref]$null, [ref]$parseErrors, [ref]$null)
+                if ($parseErrors.Count -eq 0) {
+                    $script:TransientPrompt = $true
+                    [Console]::OutputEncoding = [Text.Encoding]::UTF8
+                    [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
+                }
+            } finally {
+                if ($script:DoesUseLists) {
+                    # If PSReadline is set to display suggestion list, this workaround is needed to clear the buffer below
+                    # before accepting the current commandline. The max amount of items in the list is 10, so 12 lines
+                    # are cleared (10 + 1 more for the prompt + 1 more for current commandline).
+                    [Microsoft.PowerShell.PSConsoleReadLine]::Insert("`n" * [math]::Min($Host.UI.RawUI.WindowSize.Height - $Host.UI.RawUI.CursorPosition.Y - 1, 12))
+                    [Microsoft.PowerShell.PSConsoleReadLine]::Undo()
+                }
+                [Microsoft.PowerShell.PSConsoleReadLine]::AcceptLine()
+                [Console]::OutputEncoding = $previousOutputEncoding
+            }
+        }
+    }
+
+    function Disable-TransientPrompt {
+        Set-PSReadLineKeyHandler -Key Enter -Function AcceptLine
+    }
+
     function global:prompt {
         $origDollarQuestion = $global:?
         $origLastExitCode = $global:LASTEXITCODE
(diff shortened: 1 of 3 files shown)


rashil2000:
From the information in this thread https://github.com/fish-shell/fish-shell/pull/8142, I was able to get transience working for fish too. The problem was that the current commandline seemed to overwrite the promptline instead of clearing it first, when I pressed enter. So I have not included the fish changes.

rashil2000:
@chipbuster bump?

chipbuster (commented):
The documentation and lua files look good to me. The powershell scripts also look okay, but I would be more comfortable with a once-over from @andytom or @davidkna for those, since I'm not as familiar with pwsh as I'd like to be.

src/init/starship.ps1
```diff
@@ -64,6 +64,35 @@ $null = New-Module starship {
...
+                    [Microsoft.PowerShell.PSConsoleReadLine]::Undo()
+                }
+                [Microsoft.PowerShell.PSConsoleReadLine]::AcceptLine()
+                [Console]::OutputEncoding = $previousOutputEncoding
+            }
+        }
+    }
+
+    function Disable-TransientPrompt {
+        Set-PSReadLineKeyHandler -Key Enter -Function AcceptLine
```
davidkna, line 93: To be safe:
```suggestion
        Set-PSReadLineKeyHandler -Key Enter -Function AcceptLine
        $script:TransientPrompt = $false
```

src/init/starship.ps1
```diff
@@ -106,7 +135,16 @@ $null = New-Module starship {
         $arguments += "--status=$($lastExitCodeForPrompt)"
 
         # Invoke Starship
-        $promptText = Invoke-Native -Executable ::STARSHIP:: -Arguments $arguments
+        $promptText = if ($script:TransientPrompt) {
+            $script:TransientPrompt = $false
+            if (Test-Path function:Invoke-Starship-TransientFunction) {
+                Invoke-Starship-TransientFunction
+            } else {
+                "> "
```
davidkna, line 143: I think I'd prefer going with the default style and symbol of the character module here, but on the other hand, that could mislead people into thinking this is the character module.

davidkna (commented):
In general, the PowerShell script looks good to me.

src/init/starship.ps1
```diff
@@ -141,6 +179,9 @@ $null = New-Module starship {
     # Disable virtualenv prompt, it breaks starship
     $ENV:VIRTUAL_ENV_DISABLE_PROMPT=1
 
+    $script:TransientPrompt = $false
+    $script:DoesUseLists = (Get-PSReadLineOption).PredictionViewStyle -eq 'ListView'
```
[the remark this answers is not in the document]
rashil2000: It can be changed during runtime, but for permanent effect, people mostly set it in the $PROFILE and it becomes static then

src/init/starship.ps1
rashil2000: Hmm, yeah I'll add this

src/init/starship.ps1
rashil2000: I would have gone with the character module too, but unless https://github.com/starship/starship/pull/3442#issuecomment-1024270805 is resolved, I'd really prefer avoiding another starship invocation here.

davidkna: I meant something like this: ``"`e[1;32m❯`e[0m"`` instead of `starship module character`

src/init/starship.ps1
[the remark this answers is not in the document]
davidkna: Now that I think about it, it might also be set in `$PROFILE` after starship has initialized.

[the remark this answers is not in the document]
rashil2000: That shouldn't be a problem, because in the docs we advice people to initialize starship at the very end.

src/init/starship.ps1
rashil2000: Oh, yeah that I can add

src/init/starship.ps1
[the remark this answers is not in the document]
davidkna: Is this approach really that much faster? But this features is beta anyway, and we can always change it later if it ends up becoming an issue.

davidkna (approved):

chipbuster:
Thanks @rashil2000 for making the changes, and thanks @davidkna for the review!

PR #4159, opened 2022-07-07
perf(pulumi): allow disabling upwards discovery

davidkna:
#### Description

I noticed the module was slowing down the prompt a bit on a system with networked drives. This PR disables this upwards discovery by default to avoid this and adds an `search_upwards` option to revert to the old behavior. If this is considered a breaking change, I can also change the default value to `true`.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1026,6 +1026,7 @@
       "default": {
         "disabled": false,
         "format": "via [$symbol($username@)$stack]($style) ",
+        "search_upwards": false,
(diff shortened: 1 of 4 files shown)


andytom (commented):
This looks good to me but I think we should consider this a breaking change as the module will stop showing in some directories where it use to. Let's set the default for `search_upwards` to `true`.

andytom (approved):

PR #4160, opened 2022-07-07
feat(pwsh): Support right_prompt

### LINKED ISSUE #3177: Shell-agnostic right-prompt for shells that don't implement it themselves
After switching from oh my posh, i realised that starship does not allow for right prompt on powershell, while oh my posh does.

integrate right prompt on pwsh >=7

switch back to oh my posh and make a starship clone using omp

rashil2000:
#### Description
Support printing text on the right-hand-side of input line, commonly known as rprompt.

Note that this is not officially available in PowerShell yet (see https://github.com/PowerShell/PowerShell/issues/17641) but this workaround works well enough.

#### Motivation and Context

Closes #3177

image

#### Checklist:


rashil2000:
@davidkna So after the discussion in #4208, this one is good to go? 

It's been sitting for a while now with no feedback.

src/init/starship.ps1
```diff
@@ -136,22 +136,37 @@ $null = New-Module starship {
...
+
         # Set the number of extra lines in the prompt for PSReadLine prompt redraw.
-        Set-PSReadLineOption -ExtraPromptLineCount ($promptText.Split("`n").Length - 1)
+        Set-PSReadLineOption -ExtraPromptLineCount ($promptLines.Length - 1)
 
         # Return the prompt
-        $promptText
+        # After printing the left prompt, the cursor position is saved,
+        # right prompt is printed, and then cursor position is restored.
+        "$promptText$([char]0x1B)[s{0,$rpromptOffset}$([char]0x1B)[u" -f $rpromptText
```
davidkna, line 169: Do you know if there is a notable difference between this and the ANSI escape I used?

Please also avoid printing the rprompt if it wouldn't fit like my PR did.

src/init/starship.ps1
```diff
@@ -177,6 +192,10 @@ $null = New-Module starship {
 
     }
 
+    # The following regex is used to strip out ANSI sequences from a string,
+    # giving us just plain text. This is useful for string length calculation.
+    $ansiRegex = '([\u001B\u009B][[\]()#;?]*(?:(?:(?:[a-zA-Z\d]*(?:;[-a-zA-Z\d\/#&.:=?%@~_]*)*)?\u0007)|(?:(?:\d{1,4}(?:;\d{0,4})*)?[\dA-PR-TZcf-ntqry=><~])))'
```
davidkna, line 197: Is there a way to make this simpler? In practice, starship doesn't use all that many escapes. In shells that require it, only the `\x1b...m` escape is marked as zero-length.

docs/advanced-config/README.md
```diff
@@ -226,7 +240,7 @@ not explicitly used in either `format` or `right_format`.
 Note: The right prompt is a single line following the input location. To right align modules above
 the input line in a multi-line prompt, see the [`fill` module](/config/#fill).
 
-`right_format` is currently supported for the following shells: elvish, fish, zsh, xonsh, cmd.
+`right_format` is currently supported for the following shells: pwsh, elvish, fish, zsh, xonsh, cmd.
```
davidkna, line 243: Please add the warning I had in my PR regarding the drawbacks of the non-native approach in pwsh.

src/init/starship.ps1
```diff
@@ -136,22 +136,37 @@ $null = New-Module starship {
...
+            $rpromptText = if (Test-Path function:Invoke-Starship-TransientRightFunction) {
+                Invoke-Starship-TransientRightFunction
+            } else {
+                ""
+            }
         } else {
-            Invoke-Native -Executable ::STARSHIP:: -Arguments $arguments
+            $promptText = Invoke-Native -Executable ::STARSHIP:: -Arguments $arguments
+            $arguments += "--right"
+            $rpromptText = Invoke-Native -Executable ::STARSHIP:: -Arguments $arguments
```
davidkna, line 155: Looking at this, I should check if there's a notable performance difference between both approaches.

davidkna (changes requested):

src/init/starship.ps1
davidkna: This approach is about 20ms slower for a minimal rprompt on my Windows system, which makes me somewhat more inclined to go with https://github.com/starship/starship/pull/4208, given that all users will feel this slowdown.

docs/advanced-config/README.md
rashil2000: Will do

src/init/starship.ps1
rashil2000: Starship might not use them internally, but in theory users can specify anything in the starship.toml file, right? The regex is taken from https://github.com/chalk/ansi-regex/blob/main/index.js. I included it as-is for the sake of completeness. I don't suppose there'll be noticeable penalty for searching all the characters?

src/init/starship.ps1
rashil2000: I mean yes, two starship invocations will definitely be more time-consuming than one. 

I think we can still go with this for now, and when #3467 finally lands, we could go with the approach in #4208 as a performance improvement.

src/init/starship.ps1
rashil2000: > Do you know if there is a notable difference between this and the ANSI escape I used?

Is there any documentation on `1B7` and `1B8`? I couldn't figure out how to use it in PowerShell.

davidkna: https://en.wikipedia.org/wiki/ANSI_escape_code#Fp_Escape_sequences It does seem to be former private use, so it might be worse.

src/init/starship.ps1
davidkna: I'll try testing that, too. It might also be worth to compile this regex.

src/init/starship.ps1
davidkna: Is there an easy method to spawn both executables concurrently?

src/init/starship.ps1
rashil2000: I guess for clarity and familiarity, the `[s` and `[u` would be better, since all the Google search results for cursor control show these sequences.

src/init/starship.ps1
rashil2000: There are ways using Jobs and Threads, but they're ugly (see https://stackoverflow.com/questions/4016451/can-powershell-run-commands-in-parallel) and I'm not sure the overhead of these would justify their usage here.

docs/advanced-config/README.md
rashil2000: Done

src/init/starship.ps1
rashil2000: > Please also avoid printing the rprompt if it wouldn't fit like my PR did.

Done

rashil2000:
@davidkna I've made the requested changes. Anything remaining from my side?

rashil2000:
Bump?

davidkna:
Sorry for the delay, I wanted to explore the concurrency options but never got around to it. I dislike the slowdown associated with the rprompt for non-users, but for other shells with native support it's likely similar.
So I've thought of a way to do this concurrently, without resorting to the weird PS flavors of that: Please spawn both the normal and right starship processes together at once and wait for their outputs (like in `Invoke-Native`). The executables are already launching in the background.

src/init/starship.ps1
davidkna:
```suggestion
    $ansiRegex = [regex]'([\u001B\u009B][[\]()#;?]*(?:(?:(?:[a-zA-Z\d]*(?:;[-a-zA-Z\d\/#&.:=?%@~_]*)*)?\u0007)|(?:(?:\d{1,4}(?:;\d{0,4})*)?[\dA-PR-TZcf-ntqry=><~])))'
```

rashil2000:
> Please spawn both the normal and right starship processes together at once and wait for their outputs (like in `Invoke-Native`). The executables are already launching in the background.

Sorry I didn't get you. Are you talking about `Start-Job` and `Wait-Job`?

davidkna:
No just `[System.Diagnostics.Process]::Start` and
 ```powershell
$stdout = $process.StandardOutput.ReadToEndAsync()
$stderr = $process.StandardError.ReadToEndAsync()
[System.Threading.Tasks.Task]::WaitAll(@($stdout, $stderr))
```
But concurrently for both starship invocations.

**Comment by rashil2000** [comment]
Hi @davidkna!

Sorry, I was caught up in uni work so I couldn't attend to this. 


> No just `[System.Diagnostics.Process]::Start` and
> 
> ```powershell
> $stdout = $process.StandardOutput.ReadToEndAsync()
> $stderr = $process.StandardError.ReadToEndAsync()
> [System.Threading.Tasks.Task]::WaitAll(@($stdout, $stderr))
> ```
> 
> But concurrently for both starship invocations.

I'm still confused - if I use this, then I will be foregoing the use of `Invoke-Native`, right? And I will have to build up the `$startInfo` for both _normal_ and _right_ starship processes, separately (as is done in the function definition of `Invoke-Native`). This will make the script very long and complex.

**Comment by davidkna** [comment]
> I'm still confused - if I use this, then I will be foregoing the use of `Invoke-Native`, right? And I will have to build up the `$startInfo` for both _normal_ and _right_ starship processes, separately (as is done in the function definition of `Invoke-Native`). This will make the script very long and complex.

You could split off part of `Invoke-Native` into a separate function and for instance have it return the `startInfo` or the two  `async` readers. For compatibility with the continuation prompt, you can keep the old `Invoke-Native` around as a wrapper around the new split-off function.

**Comment by rashil2000** [comment]
@davidkna done!

P.S.: The `Invoke-Native` function is only used once now, [on line 239](https://github.com/starship/starship/blob/0c387dca400656b8a7366d4fa7268ab297fa1304/src/init/starship.ps1#L239).

File: src/init/starship.ps1 [hunk 5]
```diff
@@ -142,22 +146,50 @@ $null = New-Module starship {
...
+            if ($rserr.Result.Trim() -ne '') { $host.ui.WriteErrorLine($rserr.Result) }
+            $promptText  = $lsout.Result;
+            $rpromptText = $rsout.Result;
         }
 
+        $promptLines = $promptText.Split("`n")
+        # Calculate offset for printing on right side:
+        # offset = (console width) + (length of ANSI sequences on right_prompt) - (length of just text in last line of left_prompt)
+        $rpromptOffset = $Host.UI.RawUI.WindowSize.Width + ($rpromptText.Length - ($rpromptText -replace $ansiRegex).Length) - ($promptLines[-1] -replace $ansiRegex).Length
+
```
davidkna, line 180: I probably should have tested this earlier, but I just noticed that if the last prompt line contains an emoji or any double-width (or any non-single-width character), this calculation will break.
Does PowerShell have any nice Unicode-aware string width functions via .NET that we could use instead?

rashil2000: I tried some emojis and double-width characters (◌️, 😂, 😢) and they seem to be working well.

image

image

Do you have any characters I can test?

davidkna: `❌` from the status triggered a rendering issue in iTerm for me:
image

You can try this:
 ```
[character]
success_symbol = "❌❌❌❌❌❌❌❌❌❌❌"
```

rashil2000: Hmm, I can repro the issue using this symbol. I have tried searching online, but I can't find a proper method to find out the number of terminal boxes a character occupies, in C#. There are ways to find out code point or grapheme lengths, but they all mean different things 😢

davidkna: Starship itself uses the maximum code point width per grapheme to approximate character widths, so using something along those line should be fine.

rashil2000: Could you point me where the code for that lives? I can try translating it to PowerShell.

davidkna: It's defined in `print.rs`:
https://github.com/starship/starship/blob/master/src/print.rs#L21-L48

rashil2000: Okay, I'll admit I have no idea what's happening there 😅. It seems to be using a third-party library and its members.

Could we handle this in a new issue/PR?

davidkna: > Could we handle this in a new issue/PR?

I would rather not, we're already getting quite a few issues because `$fill` doesn't always calculate widths as expected.
At least this should use some kind of string width function like `wcswidth` verbatim.

rashil2000: Unfortunately Windows or .NET doesn't have a `wcswidth` equivalent. Most portable implementations of `wcswidth` I found online use lookup tables of code points. Even the unicode-width Rust library (used in `print.rs`) has a table. Implementing that in the PowerShell init script does not sound very feasible to me.

davidkna: In that case, I would rather go with #4208. I'll try to push the `--profile` PR forward and rendering custom target pairs after that to allow for transient rprompts in pwsh. Hopefully this can work out without a breaking change to transient prompt configuration.

rashil2000: Okay

PR #4161, opened 2022-07-07
feat(package): support cargo workspace versions

### LINKED ISSUE #3928: Cargo `version.workspace = true` support
#### Current Behavior
When upcoming feature `version.workspace = true` is set, no package version is shown

#### Expected Behavior
Package version is shown

See https://github.com/rust-lang/cargo/issues/8415 for progress on the cargo feature

#### Possible Solution
Detect this case and search for the workspace to get te version

#### Environment
- Starship version: 1.2.0
- bash version: GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)

- Operating system: Linux Mint 20.3
- Git Commit Hash: a53a99b1
- Branch/Tag: v1.2.0
- Rust Version: rustc 1.58.0 (02072b482 2022-01-11)
- Rust channel:  release
- Build Time: 2022-01-14 12:30:12 +00:00

```bash
# for examples

case $- in
    *i*) ;;
      *) return;;
esac

HISTCONTROL=ignoreboth

shopt -s histappend

HISTSIZE=1000
HISTFILESIZE=2000

shopt -s checkwinsize

#shopt -s globstar

fi

case "$TERM" in
esac

#force_color_prompt=yes

	color_prompt=yes
    else
	color_prompt=
    fi
fi

else
fi

case "$TERM" in
xterm*|rxvt*)
    ;;
*)
    ;;
esac

    alias ls='ls --color=auto'
    #alias dir='dir --color=auto'

fi

# some more ls aliases
alias ll='ls -alF'
alias la='ls -A'
alias l='ls -CF'

#   sleep 10; alert

# Alias definitions.

    . ~/.bash_aliases
fi

# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
    . /etc/bash_completion
  fi
fi
. "$HOME/.cargo/env"

export MANPAGER="vim -M +MANPAGER -"

export EDITOR=vim

export PATH=$PATH:$HOME/bin

eval "$(starship init bash)"

source /home/epage/.config/broot/launcher/bash/br
```

#### Starship Configuration

```toml
[git_status]
disabled = true

[rust]
disabled = true

[python]
disabled = true

```

davidkna:
#### Description

Newer versions of cargo support setting `version` indirectly via `version.workspace = true`. Support reading the version by querying the parent workspace and reading `workspace.package.version`. Additionally, if the initial `Cargo.toml` defines `workspace.package.version`, also try reading that.

#### Motivation and Context

Closes #3928

#### Checklist:

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -7,6 +7,8 @@ use quick_xml::events::Event as QXEvent;
 use quick_xml::Reader as QXReader;
 use regex::Regex;
 use serde_json as json;
+use std::fs;
+use std::io::Read;
 
 /// Creates a module with the current package version
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM

zao111222333:
I found that doesn't work for version 1.22.1 on both Linux and MacOS.
If there is no easy patch, I will try to fix that.

PR #4171, opened 2022-07-12
feat: add global plugin disable

kmcopper:
#### Description
add an option to disable all the plugins by default and selectively opt-in to enabling them.

#### Motivation and Context
Options are always a good thing in my opinion.

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1523,6 +1523,10 @@
     "add_newline": {
       "default": true,
       "type": "boolean"
+    },
+    "disable_all": {
+      "default": false,
+      "type": "boolean"
     }
   },
   "definitions": {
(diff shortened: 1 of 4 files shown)


matchai:
We recommend users use the [top-level `format`](https://starship.rs/config/#default-prompt-format) to selectively enable modules, as explained in the FAQ: https://starship.rs/faq/#do-top-level-format-and-module-disabled-do-the-same-thing

What does this do differently than the top-level `format`?

kmcopper:
As far as i can tell top level format does not prevent module execution. This patch fully prevents module execution unless they are enabled.

davidkna:
@kmcopper They both prevent module execution.

PR #4172, opened 2022-07-13
build: notify-rust now a mandatory dep

zeenix:
The reason to keep this an optional dep was a portability issue that was fixed in Feb:

https://github.com/NixOS/nixpkgs/issues/160876

#### Motivation and Context

tbh, as the author of zbus crate, I was happy to see that starship depeneds on it indirectly on Linux. This change would mean a mandatory dep on it (even if indirectly).

I've only checked if the build works but given that that `notify-rust` was default, I seriously doubt this can break anything.

#### Checklist:

`git grep notify-rust` didn't reveal any docs or tests to update. I did ensure tests build and run successfully though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,10 +30,9 @@ The minimal, blazing-fast, and infinitely customizable prompt for any shell! ☄
 """
 
 [features]
-default = ["battery", "notify"]
+default = ["battery"]
 battery = ["starship-battery"]
 config-schema = ["schemars"]
-notify = ["notify-rust"]
 
 [dependencies]
 ansi_term = "0.12.1"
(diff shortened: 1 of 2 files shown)


davidkna:
Have you confirmed that the issue was fixed? https://github.com/NixOS/nixpkgs/issues/160876 was closed because `notify-rust` was made optional in starship again.

zeenix:
@davidkna Thanks for the quick response!



> Have you confirmed that the issue was fixed? [NixOS/nixpkgs#160876](https://github.com/NixOS/nixpkgs/issues/160876) was closed because `notify-rust` was made optional in starship again.

Of course I didn't. That would have made sense. :) Sorry, I was still on my first cup of tea and have jetlag. :)

PR #4174, opened 2022-07-13
perf(rust): avoid calling `rustup` in more conditions

davidkna:
#### Description

This PR changes two things:
1. Instead of running `rustup run <toolchain> rustc --version` try if `~/.rustup/toolchains/<toolchain>/bin/rustc --version` works first
2. Instead of running `rustup default`, try reading `default_toolchain` from `~/.rustup/settings.toml`

I didn't add tests, because I'm not sure how to best approach that in this module.

On Windows, this improves performance from `~45ms` to `~15ms` in `starship timings` for me.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/modules/rust.rs
+++ b/src/modules/rust.rs
@@ -55,9 +55,10 @@ impl RustToolingEnvironmentInfo {
         // To display appropriate versions preventing `rustc` from downloading toolchains, we have to
         // check
         // 1. `$RUSTUP_TOOLCHAIN`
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM, not as dramatic a speed up for me (45ms down to 30ms) but still a nice improvement.

davidkna:
Thanks for reviewing again @andytom.

PR #4205, opened 2022-07-25
feat(zsh): Enable left transience

### LINKED ISSUE #888: Add transient prompt configuration option
Currently, every time a command is entered, the prompt is recreated on the line below. This is normal behavior, but arguably can look cluttery when the prompt extends beyond a single line.

The solution I propose is a configuration option for a transient prompt. The prompt would appear to stay on one line while previous commands would appear as normal. [Here's what this looks like in powerlevel10k](https://raw.githubusercontent.com/romkatv/powerlevel10k-media/master/transient-prompt.gif). I don't know how to implement this, nor do any particular drawbacks come to mind. Powerlevel10k's implementation could be a good place to look for inspiration.

rashil2000:
#### Description

Enables left and right transience for Zsh shell, similar to Cmd and PowerShell

#### Motivation and Context

Closes #888

image

#### Checklist:


rashil2000:
The reason I've marked this as draft is because there's a very stupid bug somewhere which leads to some additional padding on the right side of transient RPROMPT. I've scratched my head for hours but can't figure it out.

image

Any ideas @chipbuster @romkatv?

rashil2000:
I can remove the transient RPROMPT part in this PR for now, if we want to move forward.

rashil2000:
Doesn't seem to be a lot of interest in the transient right prompt, so I have removed it for now. This is ready for review.

max397574:
ofc there is interest
but I prefer having something over nothing

max397574:
wouldn't this be ready to be merged?

luxus:
would be nice if someone review it :D

ladydascalie:
I would very much be interested in RPROMPT transience as well. Happy to live with just left for the time being if this gets merged.

AndyP3r3z:
Who has to review this in order to merge?
I only use left prompt, so I'm totally fine with the progress made so far.
Please, merge this :(

sebastien-mcrae:
Agreed - would be great if someone could review this PR. cc: @matchai @davidkna

davidkna (changes requested):
This looks good to me, but I am not an expert in zsh.
[Given earlier comments](https://github.com/starship/starship/issues/888#issuecomment-580127661), implementing transience in zsh is difficult and will trigger bugs. Please document what steps where take to avoid this, and please document any known limitations and bugs with the used approach in the documentation.

romkatv:
I took a look at the implementation. It's essentially option 2 from [this comment](https://github.com/starship/starship/issues/888#issuecomment-580127661). It suffers from the issue described there: "[it] suppresses job monitor notifications as if `no_notify` option was in effect." Here's how you can reproduce it:

1. Start zsh.
2. Run `sleep 1&`.
3. Notice that after approximately one second the following text appears on the screen:
   ```text
   [1]  + done       sleep 1
   ```

This is the expected behavior. However, with this PR the text from the 3rd step won't appear on the screen until you press ENTER.

There is another issue with this code that I haven't mentioned earlier: the handling of Ctrl-D is incorrect. To reproduce:

1. Start zsh.
2. Run `zsh`.
3. Run `( exit 42 )`.
4. Press Ctrl-D.
5. Run `print $0`.
6. Notice that the output is "42".

This is the expected behavior. However, with this PR step 6 will print "0".

There may be more issues with this code. When working on transient prompt for Powerlevel10k I gave up on this approach in favor of another (option 1 from the comment linked earlier) because I didn't find a way to solve the job notification issue, which I considered a show stopper.

P.S.

I haven't actually tried running the code form this PR. What I wrote above about its behavior is based on a cursory reading of the code. I apologize if I misunderstood it.

rashil2000:
@davidkna how do you want me to mention the limitations of this feature? I'm no expert in zsh so I can link to this thread (and the corresponding issue) in the doc by saying "This method has some known limitations, see [link1] and [link2] for more details." Is that okay?

davidkna:
@rashil2000 I would prefer if the docs summarize the issues and perhaps link here for credit and a tracking issue.

Also, have you ever attempted to implement the first, more reliable approach to this? Is there any reason you didn't go with that?

rashil2000:
> @rashil2000 I would prefer if the docs summarize the issues and perhaps link here for credit and a tracking issue.

It would be difficult to summarize the bugs as they need detailed repro steps. 

> Also, have you ever attempted to implement the first, more reliable approach to this? Is there any reason you didn't go with that?

No. I don't think it's simpler - and as this thread mentions it breaks some stuff too: https://github.com/romkatv/powerlevel10k/issues/888#issuecomment-657969840

romkatv:
> It would be difficult to summarize the bugs as they need detailed repro steps.

Known bugs:

1. The shell behaves as if `notify` option was always off (note that it's on by default).
2. The exit status from the shell on Ctrl-D is aways 0.

> No. I don't think it's simpler - and as this thread mentions it breaks some stuff too: [romkatv/powerlevel10k#888 (comment)](https://github.com/romkatv/powerlevel10k/issues/888#issuecomment-657969840)

Keep in mind that [this comment](https://github.com/starship/starship/issues/888#issuecomment-580127661) has more info and explains how to get from 90% to 100% using the first approach. It is, however, quite difficult and requires a lot of code.

rashil2000:
> Keep in mind that [this comment](https://github.com/starship/starship/issues/888#issuecomment-580127661) has more info and explains how to get from 90% to 100% using the first approach. It is, however, quite difficult and requires a lot of code.

I'm not familiar with Zsh internals at all so I don't think I can fix the limitations in approach 1.

> Known bugs:
> 
> 1. The shell behaves as if `notify` option was always off (note that it's on by default).
> 2. The exit status from the shell on Ctrl-D is aways 0.

Thanks, I have updated the docs.

romkatv:
> I'm not familiar with Zsh internals at all so I don't think I can fix the limitations in approach 1.

That's very reasonable.

Not to split hairs but the word "limitations" is downplaying it. If transient prompt didn't shorten prompt on Mondays, that would've been a limitation. If it breaks unrelated fundamental features of the shell (such as replacing non-zero exit code with zero, or hiding notifications from background jobs -- which is what we are talking about here), that's a bug. To me personally these bugs looks very serious. If I were to launch a feature that exhibited such bugs, I would put a large disclaimer in the documentation in all caps and with blinking colors, and maybe even printed a warning on startup if the feature is enabled (with instructions on how the warning can be silenced).

I'm not the decision maker here, so it's nothing more than a bystander's opinion.

rashil2000:
The idea to print the info about the bugs seems okay to me, we could do it in the `enable_transience` function. @davidkna is that something that is acceptable in the init script?

docs/advanced-config/README.md
```diff
@@ -71,6 +71,34 @@ end
...
+enable_transience
+```
+
+Note: The current zsh transient prompt implementation has some known limitations:
+
+- The shell behaves as if `notify` option was always off (note that it's on by default).
+- The exit status from the shell on Ctrl-D is aways 0.
+
+Refer to the related [issue](https://github.com/starship/starship/issues/888) and
+[pull request](https://github.com/starship/starship/pull/4205) for more details.
```
davidkna, line 100: Let's put it on the top as an actual warning block:
````suggestion
::: warning

The current zsh transient prompt implementation has some known limitations:

- The shell behaves as if `notify` option was always off (note that it's on by default).
- The exit status from the shell on Ctrl-D is aways 0.

Refer to the related [issue](https://github.com/starship/starship/issues/888) and
[pull request](https://github.com/starship/starship/pull/4205) for more details.

:::

It is possible to replace the previous-printed prompt with a custom string. This
is useful in cases where all the prompt information is not always needed. To enable
this, run `enable_transience` in the shell session. To make it permanent, put
this statement in your `~/.zshrc`. Transience can be disabled on-the-fly with
`disable_transience`.

By default, the left side of input gets replaced with a bold-green `❯`. To customize this,
define a new function called `starship_transient_prompt_func`. For example, to
display Starship's `character` module here, you would do

```shell
function starship_transient_prompt_func {
  starship module character
}
eval "$(starship init zsh)"
enable_transience
```
````

File: src/init/starship.zsh [hunk 2]
```diff
@@ -62,6 +62,62 @@ starship_zle-keymap-select() {
...
+function disable_transience {
+  zle -D zle-line-init
+}
+
+function starship_left_prompt {
+    if [ "$is_prompt_transient" = true ] ; then
+        if whence -w starship_transient_prompt_func >/dev/null; then
+            starship_transient_prompt_func
+        else
+            echo "\e[1;32m❯\e[0m "
```
davidkna, line 106: On second look, we should probably wrap the color escapes with `%{` and `%}`.

davidkna:
```suggestion
            echo "%{\e[1;32m%}❯%{\e[0m%} "
```
I suppose this sort of thing may also warrant a warning in the docs.

rashil2000:
Done

davidkna (approved):

romkatv:
The way this code binds and deletes a zle widget won't play nice with plugins and things like oh-my-zsh and prezto. It's unfortunately cannot be assumed that there is no `zle-line-init` widget when startship is initializing, or that the existing `zle-line-init` widget is from starship when the user invokes `disable_transience`.

romkatv:
If there is someone among starship maintainers who knows zsh well, it would be a very good idea to ask them to help with this PR.

davidkna (changes requested):
@rashil2000 Given the high popularity of ohmyzsh and the likes, please fix the issue that @romkatv mentioned.

weilbith:
@romkatv thanks for all your input and sharing knowledge. 🙏
As it seems like you have done this already for p10k and @rashil2000 expressed already he doesn‘t feel knowledgable, couldn‘t you help with the implementation here? 🙂

romkatv:
If by "help" you mean implementing the whole thing, I'll respectfully decline.

weilbith:
Okay, other direction. Other underlying technologies like fish and cmd support this feature more natively. Is anyone more involved into the roadmap of Zsh to know if something into this direction is planned?

romkatv:
Zsh development discussions happen on zsh-workers@zsh.org. The archive is here: https://www.zsh.org/mla/. As far as I know, nothing like this feature has been discussed there.

weilbith:
Alright. 
I'm probably even much less knowledgable how things work in Zsh. But I currently try to learn and figure out how others to it. For example oh-my-posh seems to support transient prompts for Zsh too.

LunarLambda:
It's been nearly a year, has anything happened since?

AndyP3r3z:
> It's been nearly a year, has anything happened since?

It looks like not much happened... But for what I see, there are 19/19 checks passed, but I don't know if that means it is ready to merge or not.

Either way, from reading the comments, I think that what happens is that the instructions in [this comment](https://github.com/starship/starship/issues/888#issuecomment-580127661) must be implemented, but that triggers 3 bugs with zsh, that described @romkatv in a comment above:

> Known bugs:
> 
> 1. The shell behaves as if notify option was always off (note that it's on by default).
> 2. The exit status from the shell on Ctrl-D is aways 0.

But I also think that some people are OK if those bugs are written as comments or warnings in the code and/or documentation, which [is already done](https://github.com/starship/starship/pull/4205#issuecomment-1437301392).

LunarLambda:
yeah I'd be more than fine with those caveats. I actually use `nonotify` anyway and I've never seen anything I use care about the exit status of an interactive shell session.

ejerke:
> If by "help" you mean implementing the whole thing, I'll respectfully decline.

Would "help" meaning reviewing the PR once more be acceptable? My motivation here is pretty much:

> yeah I'd be more than fine with those caveats. I actually use `nonotify` anyway and I've never seen anything I use care about the exit status of an interactive shell session.

In short: would love to see this merged!

thespeck:
I'm currently using a solution from [powerlevel10k](https://github.com/romkatv/powerlevel10k/issues/888) for my transient prompt needs.

```zsh
# transient-prompt.zsh
zle-line-init() {
  emulate -L zsh

  [[ $CONTEXT == start ]] || return 0

  while true; do
    zle .recursive-edit
    local -i ret=$?
    [[ $ret == 0 && $KEYS == $'\4' ]] || break
    [[ -o ignore_eof ]] || exit 0
  done

  local saved_prompt=$PROMPT
  local saved_rprompt=$RPROMPT

  # Set prompt value from character module
  PROMPT=$(starship module character)
  RPROMPT=''
  zle .reset-prompt
  PROMPT=$saved_prompt
  RPROMPT=$saved_rprompt

  if (( ret )); then
    zle .send-break
  else
    zle .accept-line
  fi
  return ret
}

zle -N zle-line-init

```

```zsh
# .zshrc
source ~/transient-prompt.zsh
```

It works for me. But I cant figure out how to get `error_symbol` value from `starship`. It always prints `success_symbol` on transient prompt regardless of `$?` value.

LunarLambda:
> It works for me. But I cant figure out how to get error_symbol value from starship. It always prints success_symbol on transient prompt regardless of $? value.

`starship module -s $? character` works.

thespeck:
This is great. Thanks @LunarLambda.

PR #4204, opened 2022-07-25
feat(fish): Enable left and right transience

rashil2000:
#### Description

Enables left and right transience for Fish shell, similar to Cmd and PowerShell

#### Motivation and Context

Closes #888

#### Screenshots (if appropriate):
image

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [x] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #4208, opened 2022-07-26
feat: add support for non-native rprompt

### LINKED ISSUE #3177: Shell-agnostic right-prompt for shells that don't implement it themselves

### LINKED ISSUE #3982: Support right_format in nushell
Nushell supports dual input (left and right side), but when switching to starship I can't use it.

For example, this cannot be recreated in starship because the date/time cannot be displayed on the right in Nushell:

Screen Shot 2022-05-12 at 8 59 36 AM

davidkna:
#### Description

On shells that don't have native rprompt support enabled, use a fallback mechanism in the starship binary that print the rprompt after the main prompt via fill segments and ANSI sequences to save and restore the cursor positioning. The rprompt won't be printed if it would cause the line to overflow.

I've moved that code handles modules and `$all` in `get_prompt` for the active top-level format string into `StringFormatter::map_to_modules` to allow its use by both the main prompt and the rprompt. I also added a new `Segment`-type `Control` which holds a static string that gets ignored in width-calculations.

Compared to native rprompt this approach has some disadvantages:
> The right prompt will be printed in other shells as well, but it can behave differently. For instance, the right prompt won't be cleared once the input starts overlapping with the right prompt, and it might not line up correctly.

#### Motivation and Context

Mostly supersedes #4160
Closes #3177
Closes #3982
Related #4156

- [x] I have tested using **MacOS** (bash, ion, pwsh and tcsh)
- [x] I have tested using **Windows** (pwsh and nu)

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -261,7 +261,11 @@ not explicitly used in either `format` or `right_format`.
 Note: The right prompt is a single line following the input location. To right align modules above
 the input line in a multi-line prompt, see the [`fill` module](/config/#fill).
 
-`right_format` is currently supported for the following shells: elvish, fish, zsh, xonsh, cmd, nushell.
+`right_format` is currently supported for the following shells: pwsh, elvish, fish, zsh, xonsh, cmd, nushell.
+For pwsh, it is not implemented natively, and hence can result in two side-effects:
+
+- the right prompt won't be cleared once the input starts overlapping with the right prompt
+- the right prompt won't be reprinted if the input is cleared from the overlap region
 
 ### Example
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -697,6 +697,24 @@ pub enum Shell {
     Unknown,
 }
 
+impl Shell {
+    /// Returns `true` if starship should use the universal rprompt fallback.
+    /// If the shell natively supports rprompt or if the universal fallback is
+    /// not supported by the shell, returns `false`.
+    pub fn use_universal_prompt(&self) -> bool {
+        match self {
+            // Supported via universal fallback
+            Shell::PowerShell => true,
+            // Broken with universal fallback
+            Shell::Bash | Shell::Tcsh | Shell::Ion | Shell::Unknown => false,
+            // Supported natively
+            Shell::Elvish | Shell::Fish | Shell::Zsh | Shell::Xonsh | Shell::Cmd | Shell::Nu => {
+                false
+            }
+        }
+    }
+}
+
 /// Which kind of prompt target to print (main prompt, rprompt, ...)
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Target {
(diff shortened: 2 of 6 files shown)


rashil2000:
I tested this branch out and I like the idea of an in-built fallback universal RPROMPT in Starship. But there are a couple of differences compared to the implementation in #4160: 

- There's a padding of two columns on the right of RPROMPT, leading to misalignment. Somewhere in the calculation 2 needs to be added/subtracted. 

image
image

- #4160 will still be needed for transience RPROMPT printing, as transience is not available in Starship natively yet.

rashil2000:
> * [feat(pwsh): Support right_prompt #4160](https://github.com/starship/starship/pull/4160) will still be needed for transience RPROMPT printing, as transience is not available in Starship natively yet.

Another idea is to add two more prompt targets, `--transient` and `--transient --right`, which will eliminate the need for users to create custom functions (as done in #4143, #4204 and #4205)

davidkna:
> I tested this branch out and I like the idea of an in-built fallback universal RPROMPT in Starship. But there are a couple of differences compared to the implementation in #4160:
> 
> * There's a padding of two columns on the right of RPROMPT, leading to misalignment. Somewhere in the calculation 2 needs to be added/subtracted.

Fixed.

> Another idea is to add two more prompt targets, --transient and --transient --right, which will eliminate the need for users to create custom functions (as done in https://github.com/starship/starship/pull/4143, https://github.com/starship/starship/pull/4204 and https://github.com/starship/starship/pull/4205)

I would rather not add any more targets. I think #3467 (it could still use a hard-coded profile name) and a flag to specific a right prompt pair would work, and having an opt-in via the shell is still a good idea given the potential unreliability.

rashil2000:
I did not know about #3467, thanks! I guess then `--right` and `--continuation` will be converted into some (hard-coded, maybe) profiles too, right (once that PR is merged)?

My concern about #4160 still stands though. In shells that don't have native rprompt, printing of the main and right prompt is done simultaneously. This should be true for transient main and right prompts, too, which #3467 will not be able to handle (unless we right special code for that). So we will still need #4160.

davidkna:
> I did not know about https://github.com/starship/starship/pull/3467, thanks! I guess then --right and --continuation will be converted into some (hard-coded, maybe) profiles too, right (once that PR is merged)?

I think not, avoiding a breaking change would require adding migration code for little gain.

For the continuation prompt, I was envisioning something like `starship prompt --profile=transient --right-profile=transient-right`. This would also be handling module deduplication from `$all`, like the main and right prompt are doing.

For now, we should perhaps go with your solution, considering that the only shell this PR handles besides PowerShell is nu (though I feel like your solution moves a bit too much complexity into the shell script).

PR #4209, opened 2022-07-27
feat: add user-defined color palette

Anomalocaridid:
#### Description

I added the ability to define names for arbitrary colors in `starship.toml`. It adds a top-level config option, a table named `palette` where you can assign any name to any hex or ANSI color string. For example, if you want `#EEEEEE` to be named `#off-white`, then you would add this to `starship.toml`:

```toml
[palette]
off-white = "#EEEEEE"
```

Then you can use `off-white` to represent `#EEEEEE` wherever you can put a normal color string. User-defined colors take precedence over the 16 pre-defined colors.

Also, I added the `lazy_static` crate as a dependency so that I could define a static variable that holds the color palette in `parse_color_string()` instead of reading `starship.toml` once for every color string being parsed.

#### Motivation and Context

I was irritated that I had to write colors multiple times in different places. Not only would I have to change multiple things if I wanted to replace a color, it was difficult for me to remember which colorstring was which color. I got this idea from the `helix` text editor, which has a similar approach for defining colors for themes.

  - Distro: EndeavourOS Sway Edition
  - Shell: zsh 5.9.0.1-dev
  - Terminal: wezterm 20220624-141144-bd1b7c5d
  - Tested manually

#### Checklist:

  - I did not add any tests because I had difficulty implementing this feature in a way that allowed me to mock the config file.


davidkna (changes requested):
`Context` is available in `StringFormatter` already and contains the parsed config. Please try re-using the config from there, instead of loading it again.
The current approach doesn't seem to be testable. And even though it will likely no longer needed if you manage to use `Context`, please use the existing `once-cell` dependency instead of `lazy_static`.

Anomalocaridid:
How's this?

Rather than getting the `palette` by directly reading the config file in `parse_color_string()`, the `Context` gets passed from the `StringFormatter` all the way down to `parse_style_string()`, which passes the `palette` to `parse_color_string()`.

Also, this time I was able to add a test for the new color palette feature.

src/config.rs
```diff
@@ -300,7 +305,10 @@ pub fn parse_style_string(style_string: &str) -> Option<ansi_term::Style> {
                             None // fg:none yields no style.
                         } else {
                             // Either bg or valid color or both.
-                            let parsed = parse_color_string(color_string);
+                            let parsed = parse_color_string(
+                                color_string,
+                                &context.map(|x| x.root_config.palette.clone()),
```
davidkna, line 310:
```suggestion
                                context.map(|x| &x.root_config.palette),
```

src/config.rs
```diff
@@ -327,9 +335,12 @@ pub fn parse_style_string(style_string: &str) -> Option<ansi_term::Style> {
  There are three valid color formats:
   - #RRGGBB      (a hash followed by an RGB hex)
   - u8           (a number from 0-255, representing an ANSI color)
-  - colstring    (one of the 16 predefined color strings)
+  - colstring    (one of the 16 predefined color strings or a custom user-defined color)
 */
-fn parse_color_string(color_string: &str) -> Option<ansi_term::Color> {
+fn parse_color_string(
+    color_string: &str,
+    palette: &Option<HashMap<String, String>>,
```
davidkna, line 342:
```suggestion
    palette: Option<&HashMap<String, String>>,
```

src/config.rs
```diff
@@ -354,6 +365,16 @@ fn parse_color_string(color_string: &str) -> Option<ansi_term::Color> {
...
     }
 
+    // Check palette for a matching user-defined color
+    if let Some(palette_color) = palette.as_ref().and_then(|x| x.get(color_string)) {
+        log::trace!(
+            "Read user-defined color string: {} defined as {}",
+            color_string,
+            palette_color
+        );
+        return parse_color_string(palette_color, palette);
```
davidkna, line 375: To avoid potential stack overflows (this behavior should be documented):
```suggestion
        return parse_color_string(palette_color, None);
```

davidkna (changes requested):
It seems to work well, and I've made a few suggestions to the current code.

But there are features I feel should still be added:
* Replace palette with a map of palettes and a palette option to choose one of them. (changing this later would be a breaking change)
* Allow specifying numbered ANSI colors.

Anomalocaridid:
> Replace palette with a map of palettes and a palette option to choose one of them. (changing this later would be a breaking change)

I like that idea. That could allow for things like using different palettes depending on some condition.

> Allow specifying numbered ANSI colors (in junction with its name?).

I'm not entirely sure I understand what you mean by this. Do you mean adding the ability to assign colors numbers instead of or alongside a name? I assume that's so you can give colors in different palettes the same number and use the number in style strings to represent either color, depending on which palette is active, right?

I am not entirely against this idea, but would this even be necessary? I think something similar could be achieved by assigning corresponding colors in different palettes the same name. Something like this:

```
[palettes.cool-colors]
directory-fg = "#0000ff" # blue
directory-bg = "93"      # purple

[palettes.warm-colors]
directory-fg = "196"     # red
directory-bg = "#ffff00" # yellow
```

That way, you can match corresponding colors by name rather than keeping track of arbitrary numbers.

Any thoughts?

Anomalocaridid:
I have implemented support for multiple color palettes. Rather than a single table named `palette`, there is a table named `palettes` that maps names to tables representing individual color palettes. `palette` is now an option that specifies the name of the desired color palette from `palettes` to use. Now, color palettes are defined like my example from my previous comment. I also added a test to test retrieving a color palette from `palettes` by name.

davidkna:
@Anomalocaridid My concern regarding numbered colors is:
1. Some numbered colors also have a name (red and 1), I feel like both should be styled together.
2. Some default style strings use numbered colors. Allowing these to be remapped simplifies changing the whole color scheme via a palette, for instance in order to increase contrast on a light background (#2388).

Anomalocaridid:
@davidkna Oh I see, you want me to add the ability to override ANSI color codes. Sorry, I completely misunderstood you. That would make sense to add. My only concern is that it seems like sort of a hacky solution. My reasoning is that ANSI codes are usually, if not always, associated with specific colors, with the exception of 0-15, which have "default" colors, but are also determined by your terminal emulator's color scheme. It seems kind of strange to be able to override them, kind of like if you could override hex codes to other arbitrary colors.

As an alternative, I would like to suggest creating a default color palette and rewriting the default style strings to use it. I would be willing to do this myself and unless I am mistaken, that would not necessarily be a breaking change, especially if the styles are functionally the same, right? Although that approach brings up a few questions that would have to be answered:

- How should starship handle default color strings when using a custom palette that lacks a definition for them?
- What sort of naming convention should the default color palette use?

Regarding the first question, I personally think starship should fall back to using the default color palette. This way, users only have to worry about defining colors for custom style strings and can use whatever naming convention they want. I am less sure of how the second question should be handled, but maybe it should be something like the module name combined with the purpose the color serves.

So for example, the default foreground color for the `status` module would be something like `status_fg`. For special cases, such as the `user` module using a different style for the root user styles, its foreground colors could be `user_fg` and `user_root_fg`. It would require defining a lot of colors, but it's predictable, it would keep the default colors orthogonal, and it would theoretically allow adding extra colors to existing modules in the future without breaking compatibility.

Any thoughts?

davidkna:
@Anomalocaridid I think refactoring colors like that wouldn't be worth the effort. I suppose adding support for remapping the actual numbered colors could be unexpected. Given that, I think we'll just add the capability later if any users request it. Numbered colors seem to be more rare than I expected  in the default config, so it shouldn't restrict theming too much.

src/config.rs
```diff
@@ -354,6 +365,16 @@ fn parse_color_string(color_string: &str) -> Option<ansi_term::Color> {
...
     }
 
+    // Check palette for a matching user-defined color
+    if let Some(palette_color) = palette.as_ref().and_then(|x| x.get(color_string)) {
+        log::trace!(
+            "Read user-defined color string: {} defined as {}",
+            color_string,
+            palette_color
+        );
+        return parse_color_string(palette_color, None);
```
davidkna, line 375: Please add a test that hits this branch, something like:
```toml
red = "red"
```

docs/config/README.md
```diff
@@ -158,13 +158,15 @@ This is the list of prompt-wide configuration options.
...
+| `format`          | [link](#default-prompt-format) | Configure the format of the prompt.                                                                       |
+| `right_format`    | `""`                           | See [Enable Right Prompt](/advanced-config/#enable-right-prompt)                                          |
+| `scan_timeout`    | `30`                           | Timeout for starship to scan files (in milliseconds).                                                     |
+| `command_timeout` | `500`                          | Timeout for commands executed by starship (in milliseconds).                                              |
+| `add_newline`     | `true`                         | Inserts blank line between shell prompts.                                                                 |
+| `palette`         | `""`                           | Sets which color palette from `palettes` to use.                                                          |
+| `palettes`        | `{}`                           | Collection of color palettes that assign [colors](/advanced-config/#style-strings) to user-defined names. |
```
davidkna, line 169: Please mention that palette definitions can't reference their own definitions.

davidkna (commented):
It seems to work well. I just have two more things I'd like to see changed.

Anomalocaridid:
How's this?

davidkna (approved):

src/config.rs
```diff
@@ -384,6 +405,19 @@ fn parse_color_string(color_string: &str) -> Option<ansi_term::Color> {
...
 
+fn get_palette<'a>(
+    palettes: &'a HashMap<String, Palette>,
+    palette_name: &str,
+) -> Option<&'a Palette> {
+    let palette = palettes.get(palette_name);
+    if palette.is_some() {
+        log::trace!("Found color palette: {}", palette_name);
+    } else {
+        log::debug!("Could not find color palette: {}", palette_name);
```
andytom, line 416: Not a blocker but maybe it would be better to log this as a warning to let the user know that they have something wrong with their configuration?

src/config.rs
```diff
@@ -707,4 +741,70 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            get_palette(&palettes, "palette2")
+                .unwrap()
+                .get("test-color")
+                .unwrap(),
+            "#ABCDEF"
+        );
+    }
```
andytom, line 809: Not a blocker but should we also add a test here for getting an unknown pallet e.g `palette3`

Anomalocaridid:
@andytom I agree with both of your suggestions. How's this?

src/config.rs
```diff
@@ -413,7 +413,7 @@ fn get_palette<'a>(
     if palette.is_some() {
         log::trace!("Found color palette: {}", palette_name);
     } else {
-        log::debug!("Could not find color palette: {}", palette_name);
+        log::warn!("Could not find color palette: {}", palette_name);
```
davidkna, line 416: This will print a warning with the default starship config. (No palette defined and palette_name is an empty string). This should ignore empty strings or use `Option<&str>`.

The warning is also a bit noisy and will print multiple times depending on the configuration.

Anomalocaridid:
Sorry about that. Big oversights on my part.

I changed `palette` in `StarshipRootConfig` and `palette_name` in `get_palettes` to use `Option<Palette>`. I also changed the warning in `get_palettes` to be a debug message instead so it is not as noisy. Unfortunately, I cannot figure out how to have a warning for an undefined palette that only gets printed once.

src/config.rs
```diff
@@ -384,6 +408,24 @@ fn parse_color_string(color_string: &str) -> Option<ansi_term::Color> {
...
+fn get_palette<'a>(
+    palettes: &'a HashMap<String, Palette>,
+    palette_name: Option<&str>,
+) -> Option<&'a Palette> {
+    if let Some(palette_name) = palette_name {
+        let palette = palettes.get(palette_name);
+        if palette.is_some() {
+            log::trace!("Found color palette: {}", palette_name);
+        } else {
+            log::debug!("Could not find color palette: {}", palette_name);
```
davidkna, line 420: I think `warn` is still more appropriate here. If anything, I was thinking of to (if possible) reduce the number of times this code path is reached.

src/configs/starship_root.rs
```diff
@@ -12,8 +13,12 @@ pub struct StarshipRootConfig {
     pub scan_timeout: u64,
     pub command_timeout: u64,
     pub add_newline: bool,
+    pub palette: Option<String>,
+    pub palettes: HashMap<String, Palette>,
```
davidkna, line 17: Serializing `None` as TOML can cause issues, because it doesn't have a native type for that.
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub palettes: HashMap<String, Palette>,
```

Anomalocaridid:
How's this? I added `#[serde(skip_serializing_if = "Option::is_none")]` to the `palette` field in `StarshipRootConfig`. I also made sure `get_palette()` is only called once by saving the queried palette in a static  `OnceCell`, ensuring that the warning for a nonexistent palette only gets emitted once at most.

src/config.rs
```diff
@@ -299,8 +305,27 @@ pub fn parse_style_string(style_string: &str) -> Option<ansi_term::Style> {
                         if color_string == "none" && col_fg {
                             None // fg:none yields no style.
                         } else {
+                            // Ensure that get_palette is only called once so only one warning is emitted
+                            // when attempting to retrieve a nonexistent palette
+                            // Assumes that only one palette will be used for every time starship is run
+                            static PALETTE: once_cell::sync::OnceCell<Option<Palette>> =
+                                once_cell::sync::OnceCell::new();
+
```
davidkna, line 313: I'd still rather not have a `static` for this, because this might have hard-to-trace effects on tests. Let's just handle the duplicate messages later.

Anomalocaridid:
Understandable. I have removed that commit.

Anomalocaridid:
@davidkna Thanks for that.

@andytom Would you mind taking another look at this when you can get around to it?

andytom (approved):

andytom:
Sorry for the delay in getting to this @Anomalocaridid, but thank you for your contribution and thank you @davidkna for reviewing.

Anomalocaridid:
@andytom No worries. Thank you!

PR #4211, opened 2022-07-28
docs(directory): add note that fish_style_pwd_dir_length cannot be used with substitutions

zartre:
#### Description

Add a note to clarify that `fish_style_pwd_dir_length` option of the directory module will not produce any effect when `substitutions` is active.

#### Motivation and Context

I was trying to figure out why fish style PWD didn't work in my configuration but could not find any info in the doc.

#### Checklist:


amok:
Also, `truncate_to_repo` does not work if `fish_style_pwd_dir_length` is `> 0`

gutenye:
see also https://github.com/starship/starship/issues/3975

PR #4238, opened 2022-08-08
feat(k8s): Support for Kubernetes envvar_only lookups

saamalik:
#### Description
Add support for module activation to only look at `$KUBECONFIG` envvar instead of also the default `~/.kube/config`. 

#### Motivation and Context
A common K8s workflow is to have your `~/.kube/config` pointing to your default K8s cluster, but then override individual terminals/panes to another cluster (using a dedicated `$KUBECONFIG`). This minor feature enhancement adds support for Starship to highlight only _overriden_ panes/terminals. See screenshots below.

Also see article for more info: https://www.spectrocloud.com/blog/how-to-quickly-switch-between-kubernetes-clusters/.

Screen Shot 2022-08-07 at 7 05 03 PM

Top pane (and all other terminals) don't need to keep showing my default cluster (e.g: picard), whereas any pane/terminal explicitly overriding `$KUBECONFIG` shows the cluster name.

- Added Unit tests

#### Checklist:

- [ ] TODO: need help with creating a test that configures the `~/.kube/config` in a sandbox (_or we can drop this test, since we're currently don't have any tests for ~/.kube/config?_)


andytom:
I like this idea but I would prefer it to implement this in a more general way. I think if we could do this similar to the `detect_*` settings we use for files and directories. Something like `detect_envvars` and the logic inside the module would be if the list is empty the module is activated, if the list is non empty and one or more of the env vars  in the list are set the module is active and if the list is non empty and all the env vars in the list are not set then the module is not active.

This approach should be more flexible if, for example, there is another work flow that uses new/different env vars and something we could implement in other modules and keep a consistent configuration pattern across all modules.

saamalik:
@andytom - sounds good. I'll take a stab over the weekend.

jankatins:
This could also be accomplished in a more general way by the ignore feature in https://github.com/starship/starship/pull/4505/files

PR #4243, opened 2022-08-09
feat(container): Allow use of podman container name instead of image name

### LINKED ISSUE #3592: The container indicator should use the container name
https://github.com/starship/starship/pull/3304 added support for a container indicator but, at least in the case of podman, it uses the value of "image" from `/run/.containerenv`. Instead, it should use the value of "name" if set, to avoid that different containers derived from the same image have the same indicator.

Fuzzy-Math:
#### Description

This PR adds configuration in the container module to allow displaying the container name instead of the container image. This is useful for when many containers share the same image (ie. toolbox containers). To enable, add the following to starship.toml
```
[container]
use_container_name = true
```

Closes #3592

#### Checklist:

### DIFF
--- a/src/configs/container.rs
+++ b/src/configs/container.rs
@@ -8,6 +8,7 @@ pub struct ContainerConfig<'a> {
     pub symbol: &'a str,
     pub style: &'a str,
     pub disabled: bool,
+    pub use_container_name: bool,
 }
 
 impl<'a> Default for ContainerConfig<'a> {
@@ -17,6 +18,7 @@ impl<'a> Default for ContainerConfig<'a> {
             symbol: "⬢",
             style: "red bold dimmed",
             disabled: false,
+            use_container_name: false,
         }
     }
 }
--- a/src/modules/container.rs
+++ b/src/modules/container.rs
@@ -37,16 +37,30 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             // podman and others
 
             let image_res = read_file(container_env_path)
-                .map(|s| {
-                    s.lines()
-                        .find_map(|l| {
-                            l.starts_with("image=\"").then(|| {
-                                let r = l.split_at(7).1;
-                                let name = r.rfind('/').map(|n| r.split_at(n + 1).1);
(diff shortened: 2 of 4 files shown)


src/modules/container.rs
```diff
@@ -37,16 +40,30 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-                        .find_map(|l| {
-                            l.starts_with("image=\"").then(|| {
-                                let r = l.split_at(7).1;
-                                let name = r.rfind('/').map(|n| r.split_at(n + 1).1);
-                                String::from(name.unwrap_or(r).trim_end_matches('"'))
-                            })
+                .map(|buf| {
+                    buf.lines()
+                        .find_map(|line| {
+                            if config.use_container_name && line.starts_with("name") {
```
davidkna, line 46: Please change this to only iterate once over the lines.

src/modules/container.rs
```diff
@@ -37,16 +40,30 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                                let (_, name) = line.split_once('=').unwrap_or(("", "podman"));
+                                Some(String::from(name.trim_matches('"')))
```
davidkna, line 48: Would `shellwords::split`work here?

src/configs/container.rs
```diff
@@ -8,6 +8,7 @@ pub struct ContainerConfig<'a> {
     pub symbol: &'a str,
     pub style: &'a str,
     pub disabled: bool,
+    pub use_container_name: bool,
```
davidkna, line 11: Perhaps call this `prefer_container_name` instead, your code seems to include a fallback.

davidkna (changes requested):

spencergilbert:
I'd personally suggest the current behavior is a bug, or at least contradicting the documentation of the feature. I find the actual container name to be more useful to me than the name of the image, but it seemed like that is split between users - perhaps a configuration to toggle between the two is a good compromise.

andytom:
Closing this PR as it has been superceded by a few others.

PR #4258, opened 2022-08-15
fix: Disable multithreading in `jwalk` (via `gitoxide`) as workaround for #4251

### LINKED ISSUE #4251: Prompt hangs and uses high CPU usage
#### Current Behavior

When running `starship prompt`, it hangs and leads to high CPU usage.

#### Expected Behavior

It should show the prompt and not use high amounts of CPU.

#### Environment
- Starship version: 1.10.0
- bash version: GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)

- Operating system: Ubuntu 20.04
- Terminal emulator: vscode 1.70.1
- Git Commit Hash: 
- Branch/Tag: master
- Rust Version: rustc 1.63.0 (4b91a6ea7 2022-08-08)
- Rust channel:  release
- Build Time: 2022-08-14 20:13:29 +00:00

```bash
eval "$(starship init bash)"

nightly() {
    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain nightly
}
```

#### Starship Configuration

```toml
[time]
disabled = false
format = '[\[ $time \]]($style) '
time_format = "%T"

[nodejs]
disabled = true

[git_status]
disabled = true

[cmd_duration]
min_time = 500
format = "took [$duration](bold yellow)"

[battery]
disabled = true

[git_metrics]
disabled = false

[lua]
disabled = true
```

Byron:
#### Description

Even though that theoretically makes some filesystem iterations slower,
I don't think it matters much for typical repositories, so this fix
could be permanent.

One can explore to have this issue fixed upstream in `jwalk` as well
and I will bring it up there provided I can make a minimal reproducible
example.

With this fix, `STARSHIP_NUM_THREAD=1 starship prompt` won't hang anymore, 
which resolves #4251.

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1032,9 +1032,9 @@ dependencies = [
 
 [[package]]
 name = "git-features"
-version = "0.22.0"
+version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "57bfebf2c0a0d28414aa7c3baa058d5c1c2969e553806904343a2c11483f425a"
+checksum = "48289da362ef7ee1412a9a80bb459406d3045ad604484526210d594d10aa5268"
 dependencies = [
  "crc32fast",
  "crossbeam-channel",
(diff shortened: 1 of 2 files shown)


davidkna (approved):
Thanks for the quick fix!

ZetaNumbers:
This does not fix the issue for me (Linux)

davidkna:
@ZetaNumbers Are you also getting the issue in any repo and what steps are you using for reproducing the issue? Please also check which thread count starship ends up with.
What are the outputs of `starship bug-report`, `starship timings` and `STARSHIP_LOG=trace starship prompt`?

ZetaNumbers:
Ah, i am sorry. I have ran `starship prompt` instead of `cargo run -- prompt`. So actually it works.

PR #4264, opened 2022-08-15
feat(status): Support formatting of pipestatus separator

aaronkollasch:
#### Description

This PR adds formatting of the `pipestatus_separator` string in the "status" module before `pipestatus_separator` is used to join the pipestatus segments.

The `$style` variable can also be used in the format string, though I did not include the other variables as they seem irrelevant for formatting the pipestatus separator.

#### Motivation and Context
I could not find a way to format the pipestatus separator, so it sticks out from the rest of the status segments when they have a background color.

#### Screenshots:
before:
Screen Shot 2022-08-15 at 18 48 56

with status config:
```toml
[status]
style = "bg:57 fg:255 bold"
format = '[ $symbol $common_meaning$signal_name$maybe_int ]($style)'
disabled = false
pipestatus = true
pipestatus_format = '[ \[]($style)$pipestatus[\] => $symbol $common_meaning$signal_name$maybe_int ]($style)'
```
and after:
Screen Shot 2022-08-15 at 18 47 48

with the added status config line:
```toml
pipestatus_separator = '[|]($style)'
```

#### Checklist:

### DIFF
--- a/src/modules/status.rs
+++ b/src/modules/status.rs
@@ -722,17 +722,19 @@ mod tests {
 
     #[test]
     fn pipestatus_separator_format() {
-        let pipe_exit_code = &[0, 1];
-        let main_exit_code = 1;
+        let pipe_exit_code = &[0, 1, 2];
+        let main_exit_code = 2;
 
         let expected_style = Style::new().on(Color::Red).fg(Color::White).bold();
         let expected = Some(format!(
-            "{}{}{}{}{}",
+            "{}{}{}{}{}{}{}",
             expected_style.paint("["),
             expected_style.paint("0"),
             expected_style.paint("|"),
             expected_style.paint("1"),
-            expected_style.paint("] => <1>"),
+            expected_style.paint("|"),
+            expected_style.paint("2"),
+            expected_style.paint("] => <2>"),
         ));
         let actual = ModuleRenderer::new("status")
             .config(toml::toml! {
--- a/src/modules/status.rs
+++ b/src/modules/status.rs
@@ -55,38 +55,33 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     }
 
     let segment_format = config.pipestatus_segment_format.unwrap_or(config.format);
-
-    // Create pipestatus string
-    let pipestatus_separator =
-        match StringFormatter::new(config.pipestatus_separator).and_then(|formatter| {
-            formatter
-                .map_style(|variable| match variable {
-                    "style" => Some(Ok(config.style)),
-                    _ => None,
-                })
-                .parse(None, Some(context))
-        }) {
-            Ok(segments) => segments
-                .into_iter()
-                .map(|s| s.to_string())
-                .collect::<String>(),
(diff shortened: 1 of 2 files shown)


aaronkollasch:
Really, what I'd like be able to do is style the entire pipestatus as a single style block, e.g.
```toml
pipestatus_format = '[ \[$pipestatus\] => $symbol $common_meaning$signal_name$maybe_int ]($style)'
```
This would minimize the number of control characters that need to be printed.

However, this is currently not possible as `$pipestatus` comes pre-formatted – unless there's something I'm missing.

src/modules/status.rs
```diff
@@ -57,6 +57,22 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let segment_format = config.pipestatus_segment_format.unwrap_or(config.format);
 
     // Create pipestatus string
+    let pipestatus_separator =
+        match StringFormatter::new(config.pipestatus_separator).and_then(|formatter| {
+            formatter
+                .map_style(|variable| match variable {
+                    "style" => Some(Ok(config.style)),
+                    _ => None,
```
davidkna, line 65: I don't really like creating a separate `StringFormatter` for this. Maybe you could conditionally append the separator to the pipestatus format string?

aaronkollasch: Sure, I've added a commit that removes the extra StringFormatter and instead adds the separator to each pipestatus format string except for the last one.

aaronkollasch: Does the new version fit what you were suggesting?
https://github.com/starship/starship/pull/4264/files#diff-b377af853795c532ed1c949ed3014abe83237fb645587ff1681406d9fe726c0eR65-R71

src/modules/status.rs
```diff
@@ -55,22 +55,33 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
                         .map(|s| s.to_string())
                         .collect::<String>(),
                     Err(_) => "".to_string(),
-                },
-            )
+                }
+            })
             .collect::<Vec<String>>()
-            .join(config.pipestatus_separator),
+            .join(""),
```
davidkna, line 84: This would look a bit nicer:
```rust
            .enumerate()
            .filter_map(|(i, ec)| {
                format_exit_code(
                    ec.as_str(),
                    if i == pipestatus.len() - 1 {
                        segment_format
                    } else {
                        &segment_format_with_separator
                    },
                    None,
                    &config,
                    context,
                )
                .ok()
                .map(|segments| segments.into_iter().map(|s| s.to_string()))
            })
            .flatten()
            .collect::<String>(),
```

davidkna (approved):
I have made another small suggestion, but overall this looks good to me.

aaronkollasch:
Thanks @davidkna, this should be ready to go then. Checks are passing (though it took a few tries due to intermittent codecov errors).

davidkna:
Thanks for the contribution @aaronkollasch!

aaronkollasch:
Thanks for reviewing @davidkna!

PR #4269, opened 2022-08-16
feat(time): accept timezone names for offsets

davidkna:
#### Description

Instead of a static offset, also allow specifying timezones with their IANA time zone database names (Europe/Berlin) in the `time` module `utc_time_offset` option. This can help avoid issues with daylight saving time and can be more ergonomic.

#### Motivation and Context

Related #4173

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #4270, opened 2022-08-16
feat(schema): deny unknown keys

davidkna:
#### Description

Forbid the use of unknown keys in the config schema, which allows validators like taplo to give much better editing hints. While working on #4269 I noticed there is a method to specify serde attributes in such a manner that they are only read by `schemars` via specific `schemars()` attributes. Using `serde()` attributes instead, would have had a negative impact on runtime behavior here.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1545,6 +1545,7 @@
       "type": "boolean"
     }
   },
+  "additionalProperties": false,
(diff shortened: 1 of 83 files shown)


andytom (approved):

davidkna:

PR #4277, opened 2022-08-17
fix(git): upgrade `gitoxide` to v0.21

Byron:
This release comes with lenient configuration handling by default,
allowing to open repositories even their configuration values are
invalid (even for git), as long as there are viable defaults.

Furthermore this release adds the ability to open submodule repsitories.

Fixes https://github.com/starship/starship/issues/4266 and
fixes https://github.com/starship/starship/issues/4272
addresses performance in #4275

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -927,9 +927,9 @@ dependencies = [
 
 [[package]]
 name = "git-actor"
-version = "0.11.0"
+version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Byron:
I am not sure why [CI is failing in `git_status`](https://github.com/starship/starship/runs/7880132864?check_suite_focus=true#step:8:1606), I thought it was still using the `git` binary itself.

davidkna:
@Byron Some tests can fail intermittently (#886). In fact, the tests don't run on Windows CI at all without `core.fsyncObjectFiles` or `sleep` in-between git invocations.

andytom:
Thank you for the quick fix @Byron and thanks as always for reviewing @davidkna.

PR #4281, opened 2022-08-18
perf(git_commit): only use exact match for tag by default

### LINKED ISSUE #4275: git: starship still unusable in git repos (with submodules?)
## Bug Report
It looks like the point-release solved the hang in git repos, but starship still takes ~5 seconds to load inside a rust-lang/rust repository for me rendering it unusable. It used to take less than a second to load before 1.10.

#### Possible Solution

Can probably disable the git functionality or downgrade.

#### Environment
````
starship 1.10.1
branch:
commit_hash:
build_time:2022-08-16 00:27:55 +02:00
build_env:rustc 1.65.0-nightly (801821d15 2022-08-14),nightly-x86_64-unknown-linux-gnu
````
- Shell type: zsh 5.9
- Operating system: manjaro linux

```
# Your configuration here
```

#### Starship Configuration

```toml
# Inserts a blank line between shell prompts
add_newline = false

[time]
disabled = false

[status]
disabled = true

[cmd_duration]
min_time = 100

[directory]
truncate_to_repo = false
truncation_length = 20
```

davidkna:
#### Description

I noticed another major difference with the old git tag implementation is that only exact matches were considered, whereas this new one was looking at up to 10 commits. This restores performance to previous levels.

I also added an option `tag_max_candidates` to restore the previous behavior.

#### Motivation and Context

Closes #4275

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -525,6 +525,7 @@
         "only_detached": true,
         "style": "green bold",
         "tag_disabled": true,
+        "tag_max_candidates": 0,
         "tag_symbol": " 🏷  "
       },
       "allOf": [
@@ -2799,6 +2800,12 @@
         "tag_disabled": {
           "default": true,
           "type": "boolean"
+        },
+        "tag_max_candidates": {
+          "default": 0,
+          "type": "integer",
+          "format": "uint",
+          "minimum": 0.0
         }
       }
     },
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1558,15 +1558,16 @@ The `git_commit` module shows the current commit hash and also the tag (if any) 
 
 ### Options
 
-| Option               | Default                        | Description                                             |
-| -------------------- | ------------------------------ | ------------------------------------------------------- |
-| `commit_hash_length` | `7`                            | The length of the displayed git commit hash.            |
-| `format`             | `"[\\($hash$tag\\)]($style) "` | The format for the module.                              |
-| `style`              | `"bold green"`                 | The style for the module.                               |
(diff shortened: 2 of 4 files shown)


Byron:
I can confirm that this works as intended:

```sh
rust (14a459b) via 🐍
❯ starship timings

 git_status   -  507ms  -   ""
 git_metrics  -  434ms  -   ""
 git_commit   -  117ms  -   "(14a459b) "
 directory    -   <1ms  -   "rust "
 python       -   <1ms  -   "via 🐍 "
 line_break   -   <1ms  -   "\n"
 character    -   <1ms  -   "❯ "
```

andytom (approved):

Byron:
I took another look and noticed that `max-candidates = 0` effectively reduces the search to a map lookup, and actual describe functionality isn't desired then as all it would print is the tag name. `gitoxide` in the latest v0.21.1 point release contains the fix, so next time `starship` is built it would be picked up automatically.

This timing is more like it (2ms instead of 117ms):

```sh
rust (14a459b) via 🐍
❯ starship timings

 git_status   -  507ms  -   ""
 git_metrics  -  506ms  -   ""
 git_commit   -    2ms  -   "(14a459b) "
 directory    -   <1ms  -   "rust "
 python       -   <1ms  -   "via 🐍 "
 line_break   -   <1ms  -   "\n"
 character    -   <1ms  -   "❯ "
```

davidkna:
Thanks for reviewing this @andytom and @Byron! Also, thank you for the additional performance improvement @Byron! I think the lockfile will prevent a silent update, but dependabot already picked up the update.

PR #4313, opened 2022-08-24
docs(preset): Update nix_shell nerd font icon

patricksjackson:
#### Description

Update the Nerd Font symbol used for nix_shell to use the Nix project logo, instead of the similar looking snowflake.

#### Motivation and Context

The current snowflake symbol is an approximation for the real Nix project logo. Nerd Fonts includes the actual logo so we should use that instead. I couldn't find an exact release date for its inclusion in Nerd Fonts but it has been present in the Nerd Font codebase for at least 5 years, so this should be available for users on any reasonably recent Nerd Fonts version.

#### Checklist:

### DIFF
--- a/docs/.vuepress/public/presets/toml/nerd-font-symbols.toml
+++ b/docs/.vuepress/public/presets/toml/nerd-font-symbols.toml
@@ -50,7 +50,7 @@ symbol = " "
 symbol = " "
 
 [nix_shell]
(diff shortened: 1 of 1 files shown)


davidkna (approved):

davidkna:
Thank you for contributing the improved icon!

PR #4318, opened 2022-08-26
fix(cmd_duration): Fix cmd duration notifications

### LINKED ISSUE #4244: notification module get triggered twice or more on vi mode activation (zsh)
#### Current Behavior
I use vim keybindings a lot and in zsh I've noticed an unwanted behavior:

If I finish a long command and get notified then if I toggle on and off vi-mode in zsh, the same notifications get triggered again (can be more than twice).

#### Expected Behavior
To be notified only once

I can provide a video if necessary

#### Possible Solution
Implement some sort of a queue that delete previous notification calls once done.
This is kinda "pretentious" as I did not take a look at the code and how you guys implement all of this but still this behavior is unwanted in any kind of situation I can think about.
I don't even know how the various notification daemons works, maybe there's some sort of feedback from the daemon saying that the notification occurred, that you could use ?

#### Environment
- Starship version: starship 1.9.1, branch:master, commit_hash:4e9b8409, build_time:2022-06-27 19:07:32 +00:00, build_env:rustc 1.61.0 (fe5b13d68 2022-05-18),
- Shell type: zsh
- Shell version: zsh 5.8.1 (x86_64-ubuntu-linux-gnu)
- Shell plugin manager: oh-my-zsh
- Operating system: Linux pop-os 5.18.10-76051810-generic #202207071639~1659403207~22.04~cb5f582 SMP PREEMPT_DYNAMIC Tue A x86_64 x86_64 x86_64 GNU/Linux

```
# Your configuration here
```
Too sensitive

#### Starship Configuration

```toml
# ~/.config/starship.toml

[battery]
full_symbol = "🔋 "
charging_symbol = "⚡️ "
discharging_symbol = "💀 "

[[battery.display]]
threshold = 10
style = "bold red"

[[battery.display]] # "bold yellow" style and 💦 symbol when capacity is between 10% and 30%
threshold = 30
style = "bold yellow"
discharging_symbol = "💦"

[character]
vicmd_symbol = "[V](bold green) "

[cmd_duration]
show_notifications = true
min_time_to_notify = 45_000

[directory]
truncation_length = 3
truncate_to_repo = false

[git_metrics]
disabled = false
added_style = "bold blue"
format = '[+$added]($added_style)/[-$deleted]($deleted_style) '

[git_status]
#conflicted = "🏳"
#ahead = "🏎💨"
#behind = "😰"
#diverged = "😵"
#up_to_date = "✓"
#untracked = "🤷"
#stashed = "📦"
#modified = "📝"
#staged = '[++\($count\)](green)'
#renamed = "👅"
#deleted = "🗑"
ahead = "⇡${count}"
diverged = "⇕⇡${ahead_count}⇣${behind_count}"
behind = "⇣${count}"

[line_break]
disabled = true

#[memory_usage]
#disabled = false
#threshold = 90
#symbol = " "
#style = "bold dimmed red"
#format = "via $symbol [${ram}( \| ${swap})]($style) "

#[shell]
#disabled = false

#[status]
#style = "bg:blue"
#symbol = "🔴 "
#success_symbol = "🟢 SUCCESS"
#format = '[\[$symbol$common_meaning$signal_name$maybe_int\]]($style) '
#map_symbol = true
#disabled = false

[sudo]
style = "bold green"
symbol = "👩‍💻 "
disabled = false

#[time]
#disabled = false
#format = '🕙[\[ $time \]]($style) '
#time_format = "%T"
#utc_time_offset = "0"

#[username]
#style_user = "white bold"
#style_root = "black bold"
#format = "user: [$user]($style) "
#disabled = false
#show_always = true

```

Flygrounder:
#### Description
I've noticed that bug from #4244 happens also in fish so I've added fix for it as well. For other shells this module should behave the same as it used to. For zsh I use `precmd` hook, for fish `postexec` - both of these trigger only once after command finished even if prompt is rerendered multiple times.

#### Motivation and Context
Closes #4244

#### Checklist:


davidkna:
I would prefer if the shell integrations kept track of redraws and pass that information on to the binary directly.

Flygrounder:
In fish this can be done by setting global variable in `postexec` hook and updating it in `fish_prompt`. But I do not know how to do this reliably in ZSH: there are some hooks for specific cases where prompt is redrawn (`zle-keymap-select`, `clear-screen`, `TRAPWINCH`), but it's very hacky and I am not sure that it covers all the cases.

Flygrounder:
I've found a hack that allows counting redraws of `PROMPT` in ZSH: if I add `${//$((STARSHIP_REDRAW_COUNT++))}` to the beginning of `PROMPT`, then `STARSHIP_REDRAW_COUNT` will be incremented each time it's rendered and I can pass this variable to starship binary

src/init/starship.zsh
```diff
@@ -89,7 +90,7 @@ VIRTUAL_ENV_DISABLE_PROMPT=1
 
 setopt promptsubst
 
-PROMPT='$(::STARSHIP:: prompt --terminal-width="$COLUMNS" --keymap="${KEYMAP:-}" --status="$STARSHIP_CMD_STATUS" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --cmd-duration="${STARSHIP_DURATION:-}" --jobs="$STARSHIP_JOBS_COUNT")'
-RPROMPT='$(::STARSHIP:: prompt --right --terminal-width="$COLUMNS" --keymap="${KEYMAP:-}" --status="$STARSHIP_CMD_STATUS" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --cmd-duration="${STARSHIP_DURATION:-}" --jobs="$STARSHIP_JOBS_COUNT")'
+PROMPT='${//$((STARSHIP_REDRAW_COUNT++))}$(::STARSHIP:: prompt --terminal-width="$COLUMNS" --keymap="${KEYMAP:-}" --status="$STARSHIP_CMD_STATUS" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --cmd-duration="${STARSHIP_DURATION:-}" --jobs="$STARSHIP_JOBS_COUNT" --redraw-count="$STARSHIP_REDRAW_COUNT")'
```
davidkna, line 93: Please change this so that an empty command line does not set `STARSHIP_REDRAW_COUNT` to 0, so this information can be used elsewhere more easily. Though even without this change, zsh already seems to have some logic for handling this case.

Maybe also integrate `$((++STARSHIP_REDRAW_COUNT))}` directly into the starship prompt invocation.

Flygrounder: I do not seem to understand your first suggestion. What do you mean by an empty command line?

As for your second suggestion, if I change it like this:
`--redraw-count="$((++STARSHIP_REDRAW_COUNT))"`
then it will be executed inside of subshell and variable's state won't persist.

davidkna: I would prefer if it also incremented the value instead of resetting to zero for empty inputs in the prompt line that don't execute anything, instead of just vi-mode switches.
This sort of thing:
```
 ❯ 
 ❯ 
 ❯ 
```
As far as I can tell, they only don't send a notification because the `STARSHIP_DURATION` variable is unset just before `STARSHIP_REDRAW_COUNT` is reset to 0.

It has been requested to reset to set the displayed status (and duration) to 0 in such cases, though that approach may not work well with any vi-mode redraw. Unlike the zsh redraw-count approach, your fish init code does increment `redraw-count` for empty lines instead of resetting it.

Flygrounder: Sorry for disappearing for a couple of months - got busy and totally forgot about this PR. It looks like I can just move `STARSHIP_REDRAW_COUNT=0` to `preexec` instead of `precmd` and it will work as you described.

PR #4319, opened 2022-08-27
build: gitoxide upgrade

### LINKED ISSUE #4291: No ARM (32-bit) build for 1.10.2?
I have a GitHub Actions workflow which failed just now. It's because there is no expected ARM build `starship-arm-unknown-linux-musleabihf.tar.gz`.

Because there isn't anything mentioned in the release notes about builds for this architecture being dropped, I think a failure has occurred somewhere.

There are still a lot of 32-bit Raspberry Pis out there.

Byron:
Use the fixed `git-revision` crate via Cargo.lock to fix https://github.com/Byron/gitoxide/issues/503 .

Closes #4291

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,8 @@ dirs-next = "2.0.0"
 dunce = "1.0.2"
 gethostname = "0.2.3"
 # Addresses https://github.com/starship/starship/issues/4251
-git-features = { version = "0.22.2", features = ["fs-walkdir-single-threaded"] }
-git-repository = "0.21.1"
+git-features = { version = "0.22.3", features = ["fs-walkdir-single-threaded"] }
+git-repository = "0.22.1"
 indexmap = { version = "1.9.1", features = ["serde"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -43,8 +43,8 @@ dirs-next = "2.0.0"
 dunce = "1.0.2"
 gethostname = "0.2.3"
 # Addresses https://github.com/starship/starship/issues/4251
-git-features = { version = "0.22.2", features = ["fs-walkdir-single-threaded"] }
-git-repository = "0.21.1"
+git-features = { version = "0.22.1", features = ["fs-walkdir-single-threaded"] }
```
davidkna, line 46: Did you mean to downgrade this? If possible, please also update this to 0.22.3.

Byron: A great catch, that was unintentional. The latest version refers to `sha1` now, and it's good to see to see your PR landing here.

davidkna (approved):
Thanks for the quick fix and follow-up!

PR #4323, opened 2022-08-29
feat(azure): add username to azure module config

rjsab:
#### Description
Allows a user to show the currently logged in account to azure cli by adding $username as an available variable for the Azure Module configuration. 

#### Motivation and Context
Some companies/users work across many Azure tenants or leverage several accounts for separation of environment access, these changes help provide additional visibility to the context of the user you are currently working as.

#### Checklist:

### DIFF
--- a/src/modules/azure.rs
+++ b/src/modules/azure.rs
@@ -18,29 +18,14 @@ struct AzureProfile {
 #[derive(Serialize, Deserialize, Clone)]
 struct User {
     name: String,
-    #[serde(alias = "type")]
-    user_type: String,
-}
-
-#[derive(Serialize, Deserialize, Clone)]
-#[serde(rename_all = "camelCase")]
-struct ManagedByTenant {
-    tenant_id: String,
 }
 
 #[derive(Serialize, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
 struct Subscription {
-    id: String,
     name: String,
-    state: String,
     user: User,
     is_default: bool,
-    tenant_id: String,
-    environment_name: String,
-    home_tenant_id: String,
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    managed_by_tenants: Vec<ManagedByTenant>,
 }
 
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
@@ -93,28 +78,22 @@ fn get_azure_profile_info(context: &Context) -> Option<Subscription> {
     let mut config_path = get_config_file_location(context)?;
     config_path.push("azureProfile.json");
 
-    if config_path.exists() {
-        let azure_profile: AzureProfile = load_azure_profile(&config_path);
-        let subscription = azure_profile.subscriptions.iter().find_map(|s| {
-            if s.is_default {
-                Some(s.clone())
-            } else {
-                None
-            }
-        });
-        subscription
-    } else {
-        None
-    }
+    let azure_profile = load_azure_profile(&config_path)?;
+    azure_profile
+        .subscriptions
+        .into_iter()
+        .find(|s| s.is_default)
 }
 
-fn load_azure_profile(config_path: &PathBuf) -> AzureProfile {
-    let json_data = fs::read_to_string(&config_path).expect("Unable to open azureProfile.json");
+fn load_azure_profile(config_path: &PathBuf) -> Option<AzureProfile> {
+    let json_data = fs::read_to_string(config_path).ok()?;
     let sanitized_json_data = json_data.strip_prefix('\u{feff}').unwrap_or(&json_data);
-    let azure_profile: AzureProfile =
-        serde_json::from_str(sanitized_json_data).expect("Unable to parse json data.");
-
-    azure_profile
+    if let Ok(azure_profile) = serde_json::from_str::<AzureProfile>(sanitized_json_data) {
+        Some(azure_profile)
+    } else {
+        log::info!("Failed to parse azure profile.");
+        None
+    }
 }
 
 fn get_config_file_location(context: &Context) -> Option<PathBuf> {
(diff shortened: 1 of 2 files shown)


src/modules/azure.rs
```diff
@@ -25,6 +26,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     };
 
+    let user_name: Option<UserName> = get_azure_user_name(context);
+    if user_name.is_none() {
+        log::info!("Could not find Azure user name");
+        return None;
```
davidkna, line 32: Since this module can replace multiple things now, please handle missing value in `map` and remove the `unwrap` calls there. Since there seems to quite a bit of duplication, please also fetch the user name together with the subscription name.

src/modules/azure.rs
```diff
@@ -39,6 +46,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "subscription" => Some(Ok(subscription_name.as_ref().unwrap())),
                 _ => None,
             })
+            .map(|variable| match variable {
```
davidkna, line 49: Multiple variables are usually handled with a single closure:
```suggestion
```

src/modules/azure.rs
```diff
@@ -194,6 +227,76 @@ mod tests {
         dir.close()
     }
 
+    #[test]
+    fn user_name_set_correctly() -> io::Result<()> {
```
davidkna, line 231: Please also add a test with a missing username, but correctly set-up subscription and the reverse.

davidkna (changes requested):
Thanks for the PR! Sorry for only reviewing it now. Please fix the issues I described, so we can merge this.

rjsab:
@davidkna Thank you so much for the review! I apologies for the delay and making the changes. I've taken your feedback and made changes after reading up on Rust and soliciting feedback from a friend more versed in the language.

davidkna (changes requested):
Please only define keys in the `struct` definitions that are needed for the program and allow the program to continue if either the user or subscription name or missing by making those parts `Option<String>`.
As I mentioned in the last review, as part of this, please allow the module to work even if either the subscription name or user are missing.

rjsab:
> Please only define keys in the `struct` definitions that are needed for the program and allow the program to continue if either the user or subscription name or missing by making those parts `Option<String>`. As I mentioned in the last review, as part of this, please allow the module to work even if either the subscription name or user are missing.

For a bit of clarity, when you are asking to handle the subscription name or user missing, are you asking for me to handle them missing from the JSON structure entirely, or their values being empty?

davidkna:
> For a bit of clarity, when you are asking to handle the subscription name or user missing, are you asking for me to handle them missing from the JSON structure entirely, or their values being empty?

I was thinking of them missing entirely, but empty values would likely be handled the same way.

rjsab:
> > For a bit of clarity, when you are asking to handle the subscription name or user missing, are you asking for me to handle them missing from the JSON structure entirely, or their values being empty?
> 
> I was thinking of them missing entirely, but empty values would likely be handled the same way.

I chose to handle the parse failure on missing keys defined in the struct to allow the program to continue, instead of handling the individual keys as options. I think this would be a more appropriate handling of that situation since a missing name or user key would be more indicative of an issue with the azureProfile config file as a whole, barring any major updates from Microsoft to the format of that file; however, if Microsoft were to change the format of the file, it would likely require updates to this module regardless.

rjsab:
@davidkna - just nudging to see if you had any further recommendations or changes you would like to see. Thanks!

src/modules/azure.rs
```diff
@@ -53,28 +74,44 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    let azure_profile: Option<AzureProfile> =
+        if let Ok(azure_profile) = serde_json::from_str::<AzureProfile>(sanitized_json_data) {
+            Some(azure_profile)
+        } else {
+            log::info!("Failed to parse azure profile.");
+            None
+        };
+
+    azure_profile
+}
```
davidkna, line 113: No need to use `expect` here

```suggestion
fn load_azure_profile(config_path: &PathBuf) -> Option<AzureProfile> {
    let json_data = fs::read_to_string(&config_path).ok()?;
    let sanitized_json_data = json_data.strip_prefix('\u{feff}').unwrap_or(&json_data);
    if let Ok(azure_profile) = serde_json::from_str::<AzureProfile>(sanitized_json_data) {
        Some(azure_profile)
    } else {
        log::info!("Failed to parse azure profile.");
        None
    }
}
```

src/modules/azure.rs
```diff
@@ -53,28 +74,44 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
         } else {
             None
         }
-    });
-    if subscription_name.is_some() {
-        subscription_name
     } else {
-        log::info!("Could not find subscription name");
         None
     }
```
davidkna, line 98: Instead of checking for file existence, just try to open it:
```rust
    let azure_profile = load_azure_profile(&config_path)?;
    azure_profile
        .subscriptions
        .into_iter()
        .find(|s| s.is_default)
```

davidkna (changes requested):
I have two suggestions to clean up your code.

Given the supposed strong schema on the JSON-file, not being more flexible here is acceptable. Nevertheless, I don't really think there is much of a reason not to be more fault-tolerant, given that it wouldn't have much of an impact on complexity.

rjsab:
@davidkna thank you for the recommended changes! I've made those changes, but it appears the builds are failing on a new cargo clippy check for needless borrows. I've corrected those for this module, but build is still failing on the git_commit module.

I wanted to thank you again for your time and patience while working with me on this PR.

davidkna (approved):

andytom:
Thank you for your contribution @rjsab and thank you for reviewing @davidkna.

PR #4334, opened 2022-09-01
build: expose features for easier-to-build `git-repository` features

davidkna:
#### Description

This PR changes `git-repository` to only default to `max-performance-safe` which should be non-problematic to build in all environments. But by default, re-enable the current level of optimizations in the default feature.

In addition, add a new feature `git-repository-faster` specifically to opt into `zlib-stock` instead of `zlib-ng` from `max-performance-safe`, because it's easier to build (no cmake) and can link to zlib dynamically.

I also bumped `sha1` in `Cargo.lock` to v0.10.2, because it fixes a potential build issue.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1036,7 +1036,7 @@ dependencies = [
  "parking_lot",
  "prodash",
  "quick-error",
- "sha1 0.10.1",
+ "sha1 0.10.2",
  "sha1_smol",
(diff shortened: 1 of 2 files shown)


davidkna:
Closing for now, gitoxide just added a new zlib I want to look at first.

PR #3770, opened 2022-03-19
refactor: Replace ansi_term with owo-colors

jam1garner:
#### Description

* Replace ansi_term dependency with owo-colors 3.3.0
* Use strip-ansi-escapes in place of ansi_term's ANSI parser
  * This library is used by cargo and is quite popular, which I believe reflects well on its maintenance status relative to ansi_term's parsing
* Rework Module/Segments to merge TextSegments with identical formatting
  * This is necessary to avoid regressions in the output escape sequences due to the fact that owo-colors avoids the runtime cost of parsing/optimizing ANSI sequences
* Update module UI tests to use owo-colors

#### Motivation and Context

Copying from #3704:

> ansi_term appears to be unmaintained, having merged its last PR 3 years ago, and having had no response to its maintenance inquiry

Closes #3704 

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

Primary testing was performed via conformance to existing UI tests present throughout the modules, which were painfully thorough (great work, to be clear). Secondary testing was performed by installation and use on Linux (Ubuntu, Alacritty).

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

Sidenote/disclaimer: this was intended to be a small PR just for assessment purposes after seeing interest in owo-colors (a crate I am the author/maintainer of). However this turned out to be quite a rabbit hole. If owo-colors is determined to be a poor fit, I will 100% understand and it won't hurt my feelings \:)

PR #4339, opened 2022-09-02
refactor: replace `ansi_term` with `nu_ansi_term`

davidkna:
#### Description

Replaced `ansi_term` with the maintained nushell fork `nu_ansi_term`.
Bringing the owo-colors PR up to date was turning out to be difficult. Since `nu-ansi-term` has started to mature a bit, I think it will do as a replacement, at least until we consider brining color code escaping and printing ANSI codes more in sync.

The update is a simple search-and-replace, apart from using the bright colors defined in nu_ansi_term instead of using color numbers, which should require fewer characters to print.

#### Motivation and Context

Closes https://github.com/starship/starship/issues/3770
Closes https://github.com/starship/starship/issues/3704

#### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -180,7 +180,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 mod tests {
    use super::*;
    use crate::test::ModuleRenderer;
-   use ansi_term::Color;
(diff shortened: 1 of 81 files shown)


matchai (approved):
LGTM 👍

davidkna:

PR #4346, opened 2022-09-06
docs: fix typo

inashivb:
#### Description
Setting `disabled` to `false` means the module is enabled. Since the
setting and doc is about disabling the module, fix the typo.

#### Motivation and Context
It seemed like a typo when I was reading the docs.

#### Checklist:


davidkna:
This is not a typo. The table shows the default configuration values, and this module is not disabled by default.

inashivb:
> This is not a typo. The table shows the default configuration values, and this module is not disabled by default.

Apologies. I certainly missed the table titles. 😅

PR #4348, opened 2022-09-06
feat(logger): delete old logs & avoid more dup logs

davidkna:
#### Description

During startup, delete log files that haven't changed in 24 hours. In addition, the in-memory list of printed messages is updated in memory now. This prevents the same log message from showing up multiple times in a single prompt invocation.

As part of this refactoring, I removed all panics from the logging module, added some initial tests to the module and replaced one instance of `Lazy` with a more appropriate `Once` elsewhere.

#### Motivation and Context

Closes #

#### Checklist:


matchai (approved):
Looks good to me! Thank you for adding tests to the logger! 🙏

davidkna:

PR #4395, opened 2022-09-22
feat: add Haxe support

AlexHaxe:
#### Description
This adds a [Haxe](https://www.haxe.org) module that will display currently installed / active Haxe version when in a directory that contains Haxe related files and folders, e.g. files ending with `.hx` or `.hxml`, or a folder named `.haxelib` or `haxe_libraries` or specific file names used in Haxe development environments.

module detects Haxe version either by looking at contents of `.haxerc` file (which implies use of [lix package manager](https://github.com/lix-pm/lix.client)) or by calling `haxe --version`.

#### Motivation and Context
As a developer working on Haxe projects you might have different Haxe versions installed on your system (like current release version and a nightly build) and you switch between them to e.g. test compatibility.

This module is similar to e.g. nim module so I took that as a template. I rewrote and added test cases to reflect the differences of Haxe module. then I ran test cases to until all were green. I also ran a debug build during development in my local bash environment.

#### Checklist:

### DIFF
--- a/src/configs/haxe.rs
+++ b/src/configs/haxe.rs
@@ -28,9 +28,6 @@ impl<'a> Default for HaxeConfig<'a> {
             disabled: false,
             detect_extensions: vec!["hx", "hxml"],
             detect_files: vec![
-                "project.xml",
-                "Project.xml",
-                "application.xml",
                 "haxelib.json",
                 "hxformat.json",
                 ".haxerc",
--- a/src/configs/haxe.rs
+++ b/src/configs/haxe.rs
@@ -27,11 +27,7 @@ impl<'a> Default for HaxeConfig<'a> {
             style: "bold fg:202",
             disabled: false,
             detect_extensions: vec!["hx", "hxml"],
-            detect_files: vec![
-                "haxelib.json",
-                "hxformat.json",
-                ".haxerc",
-            ],
+            detect_files: vec!["haxelib.json", "hxformat.json", ".haxerc"],
             detect_folders: vec![".haxelib", "haxe_libraries"],
         }
     }
--- a/src/configs/haxe.rs
+++ b/src/configs/haxe.rs
@@ -27,7 +27,14 @@ impl<'a> Default for HaxeConfig<'a> {
             style: "bold fg:202",
             disabled: false,
             detect_extensions: vec!["hx", "hxml"],
-            detect_files: vec!["project.xml", "Project.xml", "application.xml", "haxelib.json", "hxformat.json", ".haxerc"],
+            detect_files: vec![
+                "project.xml",
+                "Project.xml",
+                "application.xml",
+                "haxelib.json",
+                "hxformat.json",
+                ".haxerc",
+            ],
             detect_folders: vec![".haxelib", "haxe_libraries"],
         }
     }
--- a/src/configs/haxe.rs
+++ b/src/configs/haxe.rs
@@ -1,0 +1,34 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Deserialize, Serialize)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+#[serde(default)]
+pub struct HaxeConfig<'a> {
+    pub format: &'a str,
+    pub version_format: &'a str,
+    pub symbol: &'a str,
+    pub style: &'a str,
+    pub disabled: bool,
+    pub detect_extensions: Vec<&'a str>,
+    pub detect_files: Vec<&'a str>,
+    pub detect_folders: Vec<&'a str>,
+}
+
+impl<'a> Default for HaxeConfig<'a> {
+    fn default() -> Self {
+        HaxeConfig {
+            format: "via [$symbol($version )]($style)",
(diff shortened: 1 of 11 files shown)


src/modules/haxe.rs
```diff
@@ -0,0 +1,252 @@
...
+    Some(module)
+}
+
+fn get_haxe_version(context: &Context) -> Option<String> {
+    let file_contents = context.read_file_from_pwd(".haxerc");
+    if file_contents.is_some() {
+        let raw_json = match file_contents {
+            Some(val) => val,
+            None => panic!("failed to read contents of .haxerc!"),
+        };
```
davidkna, line 70:
```suggestion
    if let Some(raw_json) = file_contents {
```

src/modules/haxe.rs
```diff
@@ -0,0 +1,252 @@
...
+        let haxerc_json: json::Value = json::from_str(&raw_json).ok()?;
+        let raw_version = haxerc_json.get("version")?.as_str()?;
```
davidkna, line 72: Does the `haxerc` always contain `version`? If not, it should fall back to `haxe --version`

davidkna (changes requested):

src/modules/haxe.rs
AlexHaxe: thanks for suggestion, it's my first dabble into rust territory :)

src/modules/haxe.rs
AlexHaxe: `.haxerc` always contains a version field.
however not all projects use lix package manager, so `.haxerc` is not present for all Haxe projects.

after checking lix sources there are multiple different values that `version` field can represent. it can either be a semver version, a git hash of a nightly build, fixed names like `stable` or `latest` or even a path to a local build of Haxe compiler. given these different possible values it doesn't make a lot of sense to use a regex to validate it's `version` field contents.

also `haxe --version` only prints a version string, so I might drop regex check entirely. unless there is a good reason to keep validating version information. 

regardless I will add a fallback to `haxe --version` in case of json parser issues.

src/configs/haxe.rs
```diff
@@ -0,0 +1,41 @@
...
+            format: "via [$symbol($version )]($style)",
+            version_format: "v${raw}",
+            symbol: "⌘ ",
+            style: "bold fg:202",
+            disabled: false,
+            detect_extensions: vec!["hx", "hxml"],
+            detect_files: vec![
+                "project.xml",
+                "Project.xml",
+                "application.xml",
```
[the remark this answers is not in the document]
AlexHaxe: understood, people who use it can still add it via their configuration files

src/modules/haxe.rs
davidkna: > it can either be a semver version, a git hash of a nightly build, fixed names like stable or latest or even a path to a local build of Haxe compiler.

In that case, I think there should be some verification that the field contents are version-like to avoid displaying long paths in the prompt.

AlexHaxe: I've added code to detect a path in a `.haxerc` version field and it will call the exectuable instead to get it's actual version number.

src/modules/haxe.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+fn get_haxerc_version(context: &Context) -> Option<String> {
+    let file_contents = context.read_file_from_pwd(".haxerc");
+    if let Some(raw_json) = file_contents {
+        let package_json: json::Value = json::from_str(&raw_json).ok()?;
+
+        let raw_version = package_json.get("version")?.as_str()?;
+        if raw_version == "null" {
```
davidkna, line 78: Is `"null"` (string) a possible value here? The line above should already handle regular non-string `null`.

src/modules/haxe.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+fn parse_haxe_version(raw_version: &str) -> Option<String> {
+    let re = Regex::new(HAXERC_VERSION_PATTERN).ok()?;
+    let captures = re.captures(raw_version)?;
+    let version = &captures["version"];
+
+    Some(version.to_string())
+}
```
davidkna, line 97: I think validating this doesn't hurt, but does this really need to use a capture group? Wouldn't verifying the whole string with `is_match()` work better here?

src/modules/haxe.rs
```diff
@@ -0,0 +1,325 @@
...
+    Some(module)
+}
+
+fn get_haxe_version(context: &Context) -> Option<String> {
+    if let Some(version) = get_haxerc_version(context) {
+        Some(version)
+    } else {
+        let cmd_output = context.exec_cmd("haxe", &["--version"])?;
+        parse_haxe_version(cmd_output.stdout.as_str())
+    }
```
davidkna, line 69: This would look better with `or_else`:
```suggestion
    get_haxerc_version(context).or_else(|| {
        context.exec_cmd("haxe", &["--version"])?;
        parse_haxe_version(cmd_output.stdout.as_str())
    })
```

davidkna (changes requested):

davidkna (approved):

AlexHaxe:
cheers, and thank you for taking the time to give feedback and do code reviews!

PR #4405, opened 2022-09-25
ci: replace dependabot with renovate

davidkna:
#### Description

Replace dependabot with [renovate](https://www.mend.io/free-developer-tools/renovate/). Compared to renovate, it has the following advantages:
- built-in auto-merging of PRs
- custom update managers (for `dprint` and `cargo install --version` shell commands in ci scripts)
- grouping of updates to related packages
- lock file maintenance (to avoid needing manual intervention in case of yanked packages)
- Shouldn't require manual intervention to disable running on forked repositories.

After merging this, the renovate GitHub app will need to be enabled for this repo to allow it to run. Alternatively, we could also run it in the GitHub Actions context. In order to simplify the dprint updater definition, I changed the URLs in the config file a bit to point directly at their release pages on GitHub.

#### Motivation and Context

Closes #

#### Checklist:


andytom (approved):
LGTM

PR #4414, opened 2022-09-29
feat(aws): add customizable `expiration_keyname`

liangyungong:
#### Description
For third party login solution like [saml2aws](https://github.com/Versent/saml2aws/blob/master/README.md#advanced-configuration-multiple-aws-account-access-but-saml-authenticate-against-a-single-sso-aws-account). It uses `x_security_token_expires` as the key name.

#### Motivation and Context
As description.

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -10,6 +10,7 @@
     "aws": {
       "default": {
         "disabled": false,
+        "expiration_keyname": "expiration",
         "expiration_symbol": "X",
         "force_display": false,
         "format": "on [$symbol($profile )(\\($region\\) )(\\[$duration\\] )]($style)",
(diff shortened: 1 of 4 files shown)


davidkna:
I think I would rather add `x_security_token_expires` as a fallback if `expiration` is missing. If that doesn't suit your needs, please allow configuring multiple keys and use the `VecOr` type.

liangyungong:
> I think I would rather add `x_security_token_expires` as a fallback if `expiration` is missing. If that doesn't suit your needs, please allow configuring multiple keys and use the `VecOr` type.

I made it a fallback in [another PR](https://github.com/starship/starship/pull/4455). This one can be closed later.

PR #4416, opened 2022-09-29
docs(java): Fix typo in file detection list

KingMob:
The Clojure tools.deps build file is called `deps.edn`, not `.deps.edn` (with a dot at the beginning). The Starship code itself looks for the correct file name, the typo is only in the docs.

#### Description
I replaced `.deps.edn` with `deps.edn` in the `docs/config/README.md` file. I did NOT replace it in any of the translated files, since, if I understand https://github.com/starship/starship/blob/d93074d0569db4bafb1788aa3f39136b734b5370/CONTRIBUTING.md#crowdin-translated-pages correctly, that could cause merge problems.

If you need someone to update the Crowdin files after this is merged, I can do that.

#### Motivation and Context
This fixes a minor, but slightly confusing, typo in the Java identification docs.

This has not been tested, but I'm not sure tests are warranted in this case.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1952,21 +1952,21 @@ disabled = false
 The `java` module shows the currently installed version of [Java](https://www.oracle.com/java/).
 By default the module will be shown if any of the following conditions are met:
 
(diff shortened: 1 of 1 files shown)


davidkna (approved):

davidkna:
Thanks for the fix!

PR #4423, opened 2022-09-30
feat(gradle): add gradle module

NyCodeGHG:
I added a module to display the version of [Gradle](https://gradle.org).

#### Description

I added 2 different strategies which are available via the config, because invoking the gradle executable is really slow.
Because almost every gradle project uses the [Gradle Wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) i made it the default option to parse the gradle version from the wrappers config file.

#### Motivation and Context

It adds support for Gradle, a popular choice for the build system in the Java and Kotlin Ecosystem.

image

I tested my changes on Windows and on a GitHub codespace, I don't own a device running macOS so I can't test that sadly.

#### Checklist:


src/configs/gradle.rs
```diff
@@ -0,0 +1,73 @@
...
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+#[serde(try_from = "String")]
+pub enum GradleVersionStrategy {
+    #[serde(rename = "properties")]
+    WrapperProperties,
+    #[serde(rename = "executable")]
+    Executable,
+}
```
davidkna, line 36: Instead of renaming every option, please use `rename_all=pascal_case`:

```suggestion
#[serde(try_from = "String")]
pub enum GradleVersionStrategy {
    #[serde(rename = "properties")]
    WrapperProperties,
    #[serde(rename = "executable")]
    Executable,
}
```

src/configs/gradle.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub detect_files: Vec<&'a str>,
+    pub detect_folders: Vec<&'a str>,
+    pub strategy: GradleVersionStrategy,
+}
+
+#[derive(Clone, Deserialize, Serialize)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
```
davidkna, line 28: This doesn't seem to do anything:
```suggestion
```

src/modules/gradle.rs
```diff
@@ -0,0 +1,243 @@
...
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "version" => {
+                    let gradle_version = match config.strategy {
+                        GradleVersionStrategy::Executable => parse_gradle_version_from_stdout(
+                            &context
+                                .exec_cmds_return_first(vec![
+                                    vec!["./gradlew", "--version"],
```
davidkna, line 44: I would avoid executing files from the current working directory unless explicitly configured to do so.

src/modules/gradle.rs
```diff
@@ -0,0 +1,243 @@
...
+        .1
+        .split_once("gradle-")?
+        .1
+        .split_once('-')?
+        .0;
+    Some(version.to_string())
+}
+
+/// Tries to find the gradle-wrapper.properties file recursively.
+fn find_wrapper_properties_file(directory: &Path) -> Option<PathBuf> {
```
davidkna, line 123: Please avoid searching files recursively, unless the module is configured to do this. For the base level, please try checking the scan results if a `gradle` folder exists.

davidkna (changes requested):

davidkna:
Please also update the presets in `docs/.vuepress/public/presets/toml`.

src/modules/gradle.rs
NyCodeGHG: 99.9% of gradle projects use the gradle wrapper, so most of the users don't even have gradle installed, also the versions can differ from project to project, so displaying the globally installed version would be wrong.

NyCodeGHG:
Now that I think about it, I think it would be better to drop the whole `(./)gradle(w) --version` execution thing and only provide the wrapper strategy, because the execution is so slow that it would time out most of the time and most gradle users use the wrapper anyway.

src/configs/gradle.rs
davidkna: Yes I meant that.

davidkna:
> Now that I think about it, I think it would be better to drop the whole `(./)gradle(w) --version` execution thing and only provide the wrapper strategy, because the execution is so slow that it would time out most of the time and most gradle users use the wrapper anyway.

That sounds good.

src/modules/gradle.rs
davidkna: I'm just concerned about running executables from potentially untrusted directories.

NyCodeGHG: Oh yeah, I didn't think of that.

src/modules/gradle.rs
```diff
@@ -0,0 +1,212 @@
...
+    if properties_path.exists() {
+        return Some(properties_path);
+    }
+
+    if recursive {
+        find_wrapper_properties_file(directory.parent()?, recursive)
+    } else {
+        None
+    }
+}
```
davidkna, line 100: I would prefer this sort of thing, making use of the scanned directory contents and reading the file straight away:
```suggestion
fn get_wrapper_properties_file(context: &Context, recursive: bool) -> Option<String> {
    let mut properties = None;
    if context
        .try_begin_scan()?
        .set_folders(&["gradle"])
        .is_match()
    {
        properties =
            utils::read_file(&context.current_dir.join("gradle/wrapper/gradle-wrapper.properties")).ok();
    };

    if recursive && properties.is_none() {
        for dir in context.current_dir.ancestors().skip(1) {
            properties =
                utils::read_file(&dir.join("gradle/wrapper/gradle-wrapper.properties")).ok();
            if properties.is_some() {
                break;
            }
        }
    }

    properties
}
```

src/modules/gradle.rs
```diff
@@ -0,0 +1,212 @@
...
+        zipStoreBase=GRADLE_USER_HOME
+        zipStorePath=wrapper/dists
+    */
+    let version = wrapper_properties
+        .split_once("distributionUrl=")?
+        .1
+        .split_once("gradle-")?
+        .1
+        .split_once('-')?
+        .0;
```
davidkna, line 84: On a second look, this may start failing if `gradle-` shows up elsewhere in the URL, this should be a bit less brittle:
```suggestion
    let version = wrapper_properties
        .lines()
        .find(|line| line.starts_with("distributionUrl="))?
        .split('/')
        .last()?
        .split('-')
        .nth(1)?;
```

davidkna (changes requested):
I just have two more minor suggestions, please also fix the conflicts that have arisen.

NyCodeGHG:
Hi! I finally found time and motivation to apply the requested changes and merge the conflicts.
I'm sorry that it took some time

NyCodeGHG:
I'm confused, when i generate the schema locally, there are no changes.

davidkna (approved):
LGTM

andytom:
Thank you for your contribution @NyCodeGHG and thank you for reviewing @davidkna.

PR #4426, opened 2022-10-01
feat: add show_detect_venv_prompt option in python module

### LINKED ISSUE #3632: Use virtual environment's `--prompt` value when venv directory detected

The following are true:

* Starship supports specifying a custom path to the executable for Python
* Starship supports specifying one or more directories to detect that determine whether to display the Python version

This works really well with [brettcannon/python-launcher](/brettcannon/python-launcher) and virtual environments:

```toml
[python]
python_binary = "py"
detect_folders = [".venv"]
```

The only shortcoming is that:

* python-launcher detects and uses the virtual environment's Python without having to activate the virtual environment, which is very convenient (and expressly intended)
* If the virtual environment created with `venv` and the `--prompt` flag, the value passed to `--prompt` is currently only displayed if the virtual environment is activated (added [here](https://github.com/starship/starship/pull/1747/files#diff-6a82ae9de73a3d1ea14bd519050b16337b03b9022d8d5dcc5e95db7c08f9db28R120))

If Starship detects one of the directories in `detect_folders` and that folder has a `pyvenv.cfg` present, display the `prompt` value from the config file alongside the Python version as is already done in the case where there's an active virtual environment. A visual indicator that a virtual environment's been picked up by the tools would go a long way in keeping a very efficient workflow.

* Manually activating the virtual environment—this is the main friction driving this pull request. Because Starship and python-launcher are both convenient and fast, this would add tedium to an otherwise smooth workflow.
* Automatically activating the virtual environment—this is an option in e.g. pyenv-virtualenv, but tends to be an unnecessary slowdown as well.
* Not doing anything—this leaves me frequently double checking that I'm actually in a directory that has a virtual environment directory in it before I install Python dependencies, also causing a slowdown.
* Propose/use a less specific solution like supplying a visual indicator _that_ a virtual environment's been found (via `detect_folders`) without doing any config file parsing. I might really be okay with this approach since it alleviates the previous bullet point, but I wasn't sure if it was possible.

t-mangoe:
#### Description

I added an option to show venv prompt value, regardless of whether the virtual environment is enabled or not.

#### Motivation and Context

Closes #3632 

#### Checklist:


jankatins:
I have trouble understanding (from the given help text) in which situations I would want that? If the venv is not active, what would it show?

t-mangoe:
Thank you for comment.
> If the venv is not active, what would it show?

-> It shows the value of `prompt` in `pyvenv.cfg` even if the venv is not active. Like this.
無題
> I have trouble understanding (from the given help text) in which situations I would want that?

-> @daneah, if you like, please describe about this situation.

daneah:
Tools like [brettcannon/python-launcher](https://github.com/brettcannon/python-launcher) can automatically use the virtual environment's Python without activating it.

PR #4429, opened 2022-10-01
Fix indent and use println instead of print (`starship init {shellname}`)

kyoheiu:
#### Description

- Use `println!` instead of  `print!` in `src/init/mod.rs` to add a line break properly.
- Remove indents of bash script to show the code properly. (indoc crate seems great to use in this case though)

#### Motivation and Context

1. First, the script printed by `starship init bash` is indented, which is not necessary.

2. Second, every script printed by `starship init {shellname}` does not have new line character at the end of it, which makes the next shell prompt appear without a line break.

Note that the line break issue appears differently according to the terminal emulator we use: For example, alacritty shows this issue, while wezterm does not.

- [x ] I have tested using **Linux**

#### Checklist:

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -101,7 +101,7 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
     let starship = StarshipPath::init()?;
 
     match shell_basename {
-        "bash" => println!(
+        "bash" => print!(
             /*
              * The standard bash bootstrap is:
              *      `source <(starship init bash --print-full-init)`
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -149,36 +149,36 @@ unset -f __main
             "#,
             starship.sprint_posix()?
         ),
-        "zsh" => println!(
+        "zsh" => print!(
             r#"source <({} init zsh --print-full-init)"#,
             starship.sprint_posix()?
         ),
-        "fish" => println!(
+        "fish" => print!(
             // Fish does process substitution with pipes and psub instead of bash syntax
             r#"source ({} init fish --print-full-init | psub)"#,
             starship.sprint_posix()?
         ),
-        "powershell" => println!(
+        "powershell" => print!(
             r#"Invoke-Expression (& {} init powershell --print-full-init | Out-String)"#,
             starship.sprint_pwsh()?
         ),
-        "ion" => println!("eval $({} init ion --print-full-init)", starship.sprint()?),
+        "ion" => print!("eval $({} init ion --print-full-init)", starship.sprint()?),
         "elvish" => print!(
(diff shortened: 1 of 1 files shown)


davidkna:
Have you tested your changes? I think some shells have issues evaluating multiple lines at once.

kyoheiu:
I cannot test on cmd and ion, but for other shells:

elvish:
elvish

fish:
fish

nushell:

tcsh:

xonsh:

zsh:

kyoheiu:
(Sorry for unintended close)

I cannot test on cmd and ion, but for other shells:

elvish:
elvish

fish:
fish

nushell:
nushell

tcsh:
tcsh

xonsh:
xonsh

zsh:
zsh

davidkna:
Unless someone objects, I would rather not merge the `println!` changes at this time, because there is little benefit, but a certain risk of something breaking - and testing all combinations of shells would be difficult. If you remove those changes, I wouldn't be opposed to merging the indent change.

kyoheiu:
Fixed to use `print!` instead of `println!`.

src/init/mod.rs
```diff
@@ -101,7 +101,7 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
     let starship = StarshipPath::init()?;
 
     match shell_basename {
-        "bash" => print!(
+        "bash" => println!(
```
davidkna, line 104:
```suggestion
        "bash" => print!(
```

src/init/mod.rs
```diff
@@ -178,7 +178,7 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
         ),
         "cmd" => print_script(CMDEXE_INIT, &StarshipPath::init()?.sprint_cmdexe()?),
         _ => {
-            eprintln!(
+            eprint!(
```
davidkna, line 181:
```suggestion
            eprintln!(
```

src/init/mod.rs
```diff
@@ -229,7 +229,7 @@ pub fn init_main(shell_name: &str) -> io::Result<()> {
 
 fn print_script(script: &str, path: &str) {
     let script = script.replace("::STARSHIP::", path);
-    print!("{}", script);
+    println!("{}", script);
```
davidkna, line 232: The files typically already have a final newline.
```suggestion
    printl!("{}", script);
```

davidkna (changes requested):

src/init/mod.rs
kyoheiu: You mean `print` right? Fixed it in my commit [06e1c62](https://github.com/starship/starship/pull/4429/commits/06e1c6296395a1e3e69e6d646ade59323372ac25)

davidkna (approved):

PR #4432, opened 2022-10-02
feat(package): added showing gradle version based on the gradle.properties file

### LINKED ISSUE #4410: extract package version from `gradle.properties`

The [starship docs](https://starship.rs/config/#package-version) for the package module only mention the Gradle build script (`build.gradle`) as a place where the package module looks for the package version. 

But in Gradle projects, the package version is often instead stored in a file called `/gradle.properties` in a format like this:

```
version=x.y.z
```

The package module indeed seems to ignore the version stored in `gradle.properties`.

It would be nice if the package module would be able to consider the version number found in `gradle.properties`.

Maybe make it configurable where the package module should look for the package version? This could potentially cover much more use cases.

BattleCh1cken:
Supplies an additional way to get gradle version.

#### Description

I added a condition in the get_gradle_version that checks if the version in build.gradle doesn't exist, and then gets the version from gradle.properties instead.
I'm fairly certain that the way I implemented this is very janky, but I don't know how to fix it.
#### Motivation and Context

Closes #4410 

I ran cargo test.

I created a new test that replicated a working directory with a gradle.properties file.

#### Checklist:


src/modules/package.rs
```diff
@@ -102,9 +102,24 @@ fn get_setup_cfg_version(context: &Context, config: &PackageConfig) -> Option<St
 }
 
 fn get_gradle_version(context: &Context, config: &PackageConfig) -> Option<String> {
-    let file_contents = context.read_file_from_pwd("build.gradle")?;
-    let re = Regex::new(r#"(?m)^version ['"](?P<version>[^'"]+)['"]$"#).unwrap();
-    let caps = re.captures(&file_contents)?;
+    let build_file_contents = match context.read_file_from_pwd("build.gradle") {
+        Some(contents) => contents,
+        None => "none".to_string(),
```
davidkna, line 107: Is `"none"` used in `build.gradle`? If not, please stick with `Option<String>`, and if it does use `"none"` please filter `Some("none")` into `None` instead and add a test that uses `version "none"`.

src/modules/package.rs
```diff
@@ -102,9 +102,24 @@ fn get_setup_cfg_version(context: &Context, config: &PackageConfig) -> Option<St
...
+    };
+
+    //check version from gradle.properties file
+    if build_file_contents == "none" {
+        let properties_file_contents = context.read_file_from_pwd("gradle.properties")?;
+        let re = Regex::new(r"/version=.*/gm").unwrap();
+        let caps = re.captures(&properties_file_contents)?;
+        return format_version(
+            &caps["version"].strip_prefix("version=").unwrap(),
```
davidkna, line 116: Please either use proper capture groups in the regex or just iterate over the lines and check if `strip_prefix` returns `Some(version)`. Please also add handling for whitespace if necessary.

src/modules/package.rs
```diff
@@ -960,6 +975,17 @@ java {
...
     }
+    #[test]
+    fn test_extract_grade_version_from_properties() -> io::Result<()> {
+        let config_name = "gradle.properties";
+        let config_content = "
+            version=1.2.3
+            ";
+        let project_dir = create_project_dir()?;
+        fill_config(&project_dir, config_name, Some(config_content))?;
+        expect_output(&project_dir, None, None);
```
davidkna, line 986: The test should show that the version number can be extracted:
```suggestion
        expect_output(&project_dir, Some("v1.2.3"), None);
```

davidkna (changes requested):

src/modules/package.rs
BattleCh1cken: I feel like I should probably stick with Option anyway, since that seems to be the intended Rust way.

src/modules/package.rs
BattleCh1cken: Capture groups was easier.

davidkna:
Please fix any clippy issues. I would also prefer if both files are considered in case the first doesn't have the version, but the other one does.
Try using something like this:
```rust
context.read_file_from_pwd("build.gradle")
    .and_then(|contents| {
        // ...
    }).or_else(|| {
        // try other file
    })
```

davidkna:
>  If the gradle.properties file returns none, it'll look in the other file.

It may exist, but not contain the version

davidkna:
I don't know how it works in detail either, just being prudent.

BattleCh1cken:
Alright it looks clean now.

moritzreiter:
> OOh. Sorry, I also don't really understand how the gradle build system works. Never used it. Thanks.

`build.gradle`/`build.gradle.kts` is the build script and `gradle.properties` is a place to store "properties" which are simple key/value pairs.

The project version is either stored in the build script directly or externalized to `gradle.properties` and then referenced from the build script like this (if using Kotlin):

```
project.properties["version"].toString()
```

To my understanding it wouldn't make sense to have a version literal in the build script *and* in the properties file as this would mean you'd have to maintain it in two places.

So I would look if there's a version to be found in `gradle.properties` and if nothing is found take it from `build.gradle`/`build.gradle.kts`.

By the way, thanks a lot for tackling this @BattleCh1cken!

BattleCh1cken:
No problem! Thanks for being a maintainer!

moritzreiter:
> No problem! Thanks for being a maintainer!

I'm not. I just felt let like giving my 2 cents as I had opened the issue requesting this little enhancement.

BattleCh1cken:
Welp. Thanks for helping me regardless.

src/modules/package.rs
```diff
@@ -102,11 +102,19 @@ fn get_setup_cfg_version(context: &Context, config: &PackageConfig) -> Option<St
...
+        .and_then(|contents| {
+            let re = Regex::new(r#"(?m)^version ['"](?P<version>[^'"]+)['"]$"#).unwrap(); /*dark magic*/
+            let caps = re.captures(&contents)?;
+            format_version(&caps["version"], config.version_format)
+        })
+        .or_else(|| {
+            let properties_file_contents = context.read_file_from_pwd("gradle.properties")?;
+            let re = Regex::new(r"version=(?P<version>.*)").unwrap();
+            let caps = re.captures(&properties_file_contents)?;
+            format_version(&caps["version"], config.version_format)
```
davidkna, line 116: Given the comments from [moritzreiter](https://github.com/moritzreiter), please read the properties file first.

flake.nix
```diff
@@ -0,0 +1,40 @@
...
+              "clippy"
+              "rust-src"
+              "rustc"
+              "rustfmt"
+              "rust-analyzer"
+            ])
+          ];
+        };
+      });
+}
```
davidkna, line 40: Please remove this file again.

davidkna (changes requested):

flake.nix
BattleCh1cken: Arg. I thought I didn't commit that file. Sorry about that.

BattleCh1cken: Do you also want me to squash these commits down into one, so that the commit history looks less messy?

davidkna: Not necessarily, PRs are squashed on merge.

BattleCh1cken: Alright, perfect.

davidkna (approved):

andytom (approved):

andytom:
Thank you for your contribution @BattleCh1cken and thank you for reviewing @moritzreiter and @davidkna.

PR #4439, opened 2022-10-03
feat: allow loading configuration values from the environment

### LINKED ISSUE #4808: ENV value as style (any config value)

I keep my config files in my private GIT-repo.

At a new host, I do `git pull` and the host is ready to use.

It is very useful, but I have one version of the config for each host.

From time to time I want to have different configs in different hosts, for example, I want to have different colours of PROMPT.

Could You add a way to use ENV values in the config (`.config/starship.toml`)?

example:
I do:
```bash
case hostname in
  host1)
     export HOSTNAME_COLOR='green dimmed bold'
  ;;
  host2)
     export HOSTNAME_COLOR='blue dimmed bold'
  ;;
  *)
     export HOSTNAME_COLOR='green dimmed bold'
  ;;
esac
```
before starting any terminal.

starship.toml could have these records:

```toml
[hostname]
ssh_only = false
trim_at = '.'
format = '[@$hostname]($style)'
style = { env = HOSTNAME_COLOR }
disabled = false
```

davidkna:
#### Description

This PR allows overwriting individual config keys with config keys like `STARSHIP_CONFIG__STATUS__DISABLED=false`. The code is very similar to the `starship config` sub-command. This allows making smaller overrides to the configuration without needing to create duplicate config files.
~~`toml_edit` is used for parsing the values, because `toml` does not support this, but it looks like both `toml` and `toml_edit` will use the same value types soon.~~ `toml` is used for parsing values now.

#### Motivation and Context

Related #4352
Closes #4808

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -44,6 +44,29 @@ Or for Cmd (Windows) would be adding this line to your `starship.lua`:
 os.setenv('STARSHIP_CONFIG', 'C:\\Users\\user\\example\\non\\default\\path\\starship.toml')
 ```
 
+### Loading additional configuration from the environment
+
+In addition to the configuration file, starship also allows you to set configuration values by using environment variables. The variables should be formatted like this: `STARSHIP_CONFIG__SECTION_NAME__KEY=value`. The configuration keys in the environment need to start with `STARSHIP_CONFIG__` and table keys need to be separated with `__`.
+
+In most POSIX-like shells, you can set the `disabled` key of the `status` module to `false` with the following:
+
+```sh
+export STARSHIP_CONFIG__STATUS__DISABLED=false
+```
+
+Equivalently in PowerShell (Windows) would be adding this line:
+
+```powershell
+$ENV:STARSHIP_CONFIG__STATUS__DISABLED = "false"
+```
+
+Other shells may use a different syntax to set environment variables.
+Any values set in the environment will be parsed as TOML-values, not as full documents. If the environment values cannot be directly parsed as TOML-values, they are treated as strings.
+
+This means that `true` is parsed as a boolean when it's not quoted, but is parsed as a string when it is quoted. In contrast, when a value that cannot be parsed as a TOML value, such as `starship` is found, it is treated as a string (`"starship"`).
+
+Because the values are TOML-values the `[table]` syntax will not work, but inline tables `{ a = 1 }` or arrays `[ 0, 1 ]`.
+
 ### Logging
 
 By default starship logs warnings and errors into a file named `~/.cache/starship/session_${STARSHIP_SESSION_KEY}.log`, where the session key is corresponding to an instance of your terminal.
--- a/src/config.rs
+++ b/src/config.rs
@@ -125,11 +125,18 @@ pub struct StarshipConfig {
 impl StarshipConfig {
     /// Initialize the Config struct
     pub fn initialize(config_file_path: Option<&OsStr>) -> Self {
-        Self::config_from_file(config_file_path)
+        #[cfg_attr(test, allow(unused_mut))]
+        let mut out = Self::config_from_file(config_file_path)
             .map(|config| Self {
                 config: Some(config),
             })
-            .unwrap_or_default()
+            .unwrap_or_default();
+
+        // Avoid tainting config in tests
+        #[cfg(not(test))]
+        out.load_from_env();
+
+        out
     }
 
     /// Create a config from a starship configuration file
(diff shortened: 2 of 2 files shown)


silvio:
Hi!

I'm the initiator of #4352.

I'm in favour of this PR, but I'm not sure if it should really close 4352, since the feature includes other content that was not implemented in this PR.

Thanks for your work!

onecrayon:
What's the status of this PR? I ran across it because I need a way to disable specific modules when using one shell vs. others and environment variables would perfectly solve my problem (because I can check for the shell and set appropriate environment variables in `.bashrc` just prior to loading Starship). If there's something I could do to help contribute to moving it toward being available to end users, I'd love to do so.

jankatins:
I would also love to see this land: I use direnv for setting up env vars per folder/projects and enabling the kubernetes module only in folders which need it would be really nice, especially together with [kubesess](https://github.com/Ramilito/kubesess) which can automatically set the right namespace/context per terminal session.

So: any help needed? I would love to help getting this over the finish line...

badloop:
Is this stagnated, or still being worked on?  Just hit a case today where this would be very helpful.  Anything I can do to assist?

silvio:
Yay! It does exactly what I suggested. Great! ❤️
I have tested the following:
```
«:starship on  feso/test is 📦 v1.17.1 via 🦀 v1.74.0 ➜  starship timings
                                                                  
 git_status   -  12ms  -   ""                                     
 git_metrics  -  10ms  -   ""                                     
 rust         -   9ms  -   "via 🦀 v1.74.0 "                      
 directory    -  <1ms  -   "«:starship "                          
 git_branch   -  <1ms  -   "on  feso/test "                      
 package      -  <1ms  -   "is 📦 v1.17.1 "                       
 character    -  <1ms  -   "➜  "                                  


«:starship on  feso/test is 📦 v1.17.1 via 🦀 v1.74.0 ➜  export STARSHIP__RUST__DISABLE=true
«:starship on  feso/test is 📦 v1.17.1 via 🦀 v1.74.0 ➜  export STARSHIP_CONFIG__RUST__DISABLED=true
«:starship on  feso/test is 📦 v1.17.1 ➜  export STARSHIP_CONFIG__GIT_METRICS__DISABLED=true
«:starship on  feso/test is 📦 v1.17.1 ➜  export STARSHIP_CONFIG__GIT_STATUS__DISABLED=true
«:starship on  feso/test is 📦 v1.17.1 ➜  export STARSHIP_CONFIG__GIT_BRANCH__TRUNCATION_LENGTH=4
«:starship on  feso… is 📦 v1.17.1 ➜  export STARSHIP_CONFIG__GIT_BRANCH__SYMBOL='xxx '
«:starship on xxx feso… is 📦 v1.17.1 ➜  starship timings

 battery     -   9ms  -   ""
 directory   -  <1ms  -   "«:starship "
 git_branch  -  <1ms  -   "on xxx feso… "
 package     -  <1ms  -   "is 📦 v1.17.1 "
 character   -  <1ms  -   "➜  "

```

Perfect!

nekowinston:
Seeing how both above-mentioned PRs got merged, please consider merging this PR as well.
I've been using this patch since [February 15](https://github.com/nekowinston/dotfiles/commit/c286c8b9636257947a2b0f6294225c5e66b53426#diff-82ed4160a9df744972aa59460615ff47a4d6b0b47179be464411d544a7e24b92R8-R13) with my config, and am using it to change the color scheme on the fly.

docs/config/README.md
```diff
@@ -44,6 +44,29 @@ Or for Cmd (Windows) would be adding this line to your `starship.lua`:
...
+```powershell
+$ENV:STARSHIP_CONFIG__STATUS__DISABLED = "false"
+```
+
+Other shells may use a different syntax to set environment variables.
+Any values set in the environment will be parsed as TOML-values, not as full documents. If the environment values cannot be directly parsed as TOML-values, they are treated as strings.
+
+This means that `true` is parsed as a boolean when it's not quoted, but is parsed as a string when it is quoted. In contrast, when a value that cannot be parsed as a TOML value, such as `starship` is found, it is treated as a string (`"starship"`).
+
+Because the values are TOML-values the `[table]` syntax will not work, but inline tables `{ a = 1 }` or arrays `[ 0, 1 ]`.
```
vergenzt, line 68: Could this eager parsing logic (and the expansiveness of the environment configurability) be a security risk?

If this PR were merged then an attacker with write access to environment variables would be able to escalate that access into remote code execution (RCE) by setting `STARSHIP_CONFIG__CUSTOM__MY_RCE_MODULE__WHEN=<...open_rce_shell_command>`.

Having write access to environment variables seems like pretty high privilege to me already so being able to escalate that into RCE may not be an issue, but I'm not sure. 🤷

vergenzt: Some food for thought: https://www.oreilly.com/library/view/secure-programming-cookbook/0596003943/ch01s01.html

src/config.rs
```diff
@@ -173,6 +180,68 @@ impl StarshipConfig {
         }
     }
 
+    /// Load additional config values from the environment
+    /// Variables are prefixed with `STARSHIP_` and delimited with `__`
+    /// e.g. `STARSHIP_CONFIG__JAVA__DISABLED`
+    pub fn load_from_env(&mut self) {
+        for (key, value) in std::env::vars() {
+            if let Some(name) = key.strip_prefix("STARSHIP_CONFIG__") {
+                log::debug!("Loading config from environment: {name}={value}");
```
vergenzt, line 189: I would suggest logging with the full environment variable name, for clarity:

```suggestion
                log::debug!("Loading config from environment: {key}={value}");
```

src/config.rs
```diff
@@ -173,6 +180,68 @@ impl StarshipConfig {
         }
     }
 
+    /// Load additional config values from the environment
+    /// Variables are prefixed with `STARSHIP_` and delimited with `__`
```
vergenzt, line 184:
```suggestion
    /// Variables are prefixed with `STARSHIP_CONFIG__` and delimited with `__`
```

nekowinston:
heya, i'm still compiling this branch as a patch on top of nixpkgs-unstable every few weeks.
I've cleaned up the conflicts on my fork here: https://github.com/nekowinston/starship/tree/set-from-env

PR #4441, opened 2022-10-03
feat: Open Policy Agent module (#1740)

### LINKED ISSUE #1740: [feature request] OPA(Open Policy Agent) Support
## Feature Request
OPA is a new way to describe policies across the stack in cloud-native, it would be nice to add support for this.

OPA uses Rego (DSL) language to author the policies, maybe we can track .rego extension to show up the OPA symbol.

Screen Shot 2020-10-06 at 09 43 51

Scotsguy:
#### Description

Adds a module for the [Open Policy Agent](https://www.openpolicyagent.org/) tool, `opa`.
This assumes that people put the opa binary on $PATH, as it seems to not really be packaged apart from Docker Hub.

Emoji was picked because it's the closest to the official logo, which is a horned hat.

Largely based on the bun module and other modules that just run a command and parse it.

Upstream will likely not change the output of `opa version`, as they're conscious of it potentially being parsed; see https://github.com/open-policy-agent/opa/blob/15b9a884cd5f5fd403cec9929fa58d86b0a9046f/cmd/version.go

#### Motivation and Context

Closes #1740

image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -259,6 +259,7 @@ $lua\
 $nim\
 $nodejs\
 $ocaml\
+$opa\
 $perl\
 $php\
 $pulumi\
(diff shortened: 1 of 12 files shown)


src/configs/opa.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub detect_files: Vec<&'a str>,
+    pub detect_folders: Vec<&'a str>,
+}
+
+impl<'a> Default for OpaConfig<'a> {
+    fn default() -> Self {
+        OpaConfig {
+            format: "via [$symbol($version )]($style)",
+            version_format: "v${raw}",
+            symbol: "🪖 ",
```
davidkna, line 26: VS-Code and iTerm require two spaces this for proper spacing, or are there any other options? The emoji was only standardized recently:
```suggestion
            symbol: "🪖  ",
```

davidkna (changes requested):
Thanks for the contribution! This PR mostly looks good, but please address my comments.

Please also update the presets as required in `docs/.vuepress/public/presets/toml`.

docs/config/README.md
```diff
@@ -2595,6 +2596,43 @@ By default the module will be shown if any of the following conditions are met:
...
+By default the module will be shown if the current directory contains a `.rego` file.
+
+### Options
+
+| Option              | Default                              | Description                                                               |
+| ------------------- | ------------------------------------ | ------------------------------------------------------------------------- |
+| `format`            | `"via [$symbol($version )]($style)"` | The format for the module.                                                |
+| `version_format`    | `"v${raw}"`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
+| `symbol`            | `"🪖 "`                               | A format string representing the symbol of OPA.                           |
+| `detect_extensions` | `["ropa"]`                           | Which extensions should trigger this module.                              |
```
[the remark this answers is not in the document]
Scotsguy: Whoops, typo

src/configs/opa.rs
Scotsguy: I can't test on iTerm, but:
Two spaces is the correct amount in VSCode, but
two spaces is too much on kitty

Same thing happens with an alternative emoji, ⛑️ "Helmet with White Cross", even though that's existed for far longer.

What's the consensus on other modules that this happens to, one space or two spaces?

Scotsguy:
Presets updated, at least the ones I think are relevant (?). There are no nerd fonts icons for OPA.

src/configs/opa.rs
davidkna: Kitty is better at determining proper widths, but other terminals will not match it, so please go with two spaces.

Scotsguy: Updated and pushed.

davidkna (approved):

PR #4450, opened 2022-10-05
fix(buf): fix spacing & harmonize docs with actual configuration

wilhelmeek:
#### Description

This change adds a trailing whitespace character after the printed version to align with behaviour of other modules. I've used the configuration for the `node` and `bun` modules as reference points.

Additionally:

- Updated documentation to specify `version` instead of incorrect `buf_version`
- Use bison emoji in-line with documentation.

#### Motivation and Context

The lack of trailing whitespace leads to rendering error when used with other modules.

Currently `buf` version is printed as:

image

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -73,7 +73,7 @@
         "disabled": false,
         "format": "with [$symbol($version )]($style)",
         "style": "bold blue",
-        "symbol": " ",
+        "symbol": "🦬 ",
         "version_format": "v${raw}"
       },
       "allOf": [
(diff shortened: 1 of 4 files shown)


davidkna (changes requested):
Your change looks good, but please also update the docs in `docs/config/README.md`.
In addition to this, please also fix a few other issues I just found (if this is an issue, I can make a follow-up pr):
In the docs, please document the version number as `version` instead of the incorrect name used in the docs (`buf_version`). Additionally, the symbol used in the docs is different to the actual symbol used in the code - but in this case, I would prefer if you changed the code to use the emoji from the docs.

wilhelmeek:
Hi @davidkna, thanks for pointing that out! I've updated the docs and symbol. I very much prefer the 🦬 as well.

andytom:
Thanks for your contribution @wilhelmeek.

PR #4455, opened 2022-10-08
feat(aws): add a fallback for `expiration` key

liangyungong:
#### Description
For third party login solution like [saml2aws](https://github.com/Versent/saml2aws/blob/master/README.md#advanced-configuration-multiple-aws-account-access-but-saml-authenticate-against-a-single-sso-aws-account). It uses `x_security_token_expires` as the expiration key name.

#### Motivation and Context
Add `x_security_token_expires` as a fallback, so both `expiration` and `x_security_token_expires` works.

#### Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -131,8 +131,10 @@ fn get_credentials_duration(
         let creds = get_creds(context, aws_creds)?;
         let section = get_profile_creds(creds, aws_profile)?;
 
-        section
-            .get("expiration")
+        let expiration_keys = ["expiration", "x_security_token_expires"];
+        expiration_keys
+            .iter()
+            .find_map(|expiration_key| section.get(expiration_key))
             .and_then(|expiration| DateTime::parse_from_rfc3339(expiration).ok())
     }?;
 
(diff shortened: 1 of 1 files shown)


src/modules/aws.rs
```diff
@@ -655,54 +657,58 @@ credential_process = /opt/bin/awscreds-retriever
...
+                .collect();
+
+            assert_eq!(
+                actual, actual_variant,
+                "both AWS_SHARED_CREDENTIALS_FILE and AWS_CREDENTIALS_FILE should work"
+            );
+
+            // In principle, "30m" should be correct. However, bad luck in scheduling
+            // on shared runners may delay it. Allow for up to 2 seconds of delay.
+            let possible_values = ["30m", "29m59s", "29m58s"];
```
davidkna, line 700: Despite #4455 this test has been failing intermittently, so let's increase the range a bit:
```suggestion
            // In principle, "30m" should be correct. However, bad luck in scheduling
            // on shared runners may delay it. Allow a few seconds of skew.
            let possible_values = [
                "30m2s", "30m1s", "30m", "29m59s", "29m58s", "29m57s", "29m56s", "29m55s",
            ];
```

src/modules/aws.rs
```diff
@@ -655,54 +657,58 @@ credential_process = /opt/bin/awscreds-retriever
...
+            let possible_values = possible_values.map(|duration| {
+                let segment_colored = format!("☁️  astronauts (ap-northeast-2) [{}] ", duration);
+                Some(format!(
+                    "on {}",
+                    Color::Yellow.bold().paint(segment_colored)
+                ))
+            });
+
+            assert!(possible_values.contains(&actual));
```
davidkna, line 709: To increase observability on failure:
```suggestion
            assert!(
                possible_values.contains(&actual),
                "time is in range: {actual:?}"
            );
```

davidkna (changes requested):
I have made two suggestions regarding the tests. Please also update the docs in `docs/config/README.md`.

davidkna (approved):

liangyungong:
@davidkna Would someone merge this soon? 🤔

PR #4463, opened 2022-10-09
feat(module): Add dbt module to display version and project name

tomasfarias:
#### Description

Adds a module for [dbt](https://github.com/dbt-labs/dbt-core) which displays the current dbt version and the current dbt project.

Fetching the dbt version was not as trivial as I originally imagined: Python requires all installed packages to include a [.dist-info directory](https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-dist-info-directory). This directory contains the version of a package in its name, and it's what I rely on to obtain the currently installed dbt version. We use [`importlib.resources.files`](https://docs.python.org/3.10/library/importlib.html?highlight=importlib%20resources%20files#importlib.resources.files) to fetch where dbt was installed, and then look for the corresponding `.dist-info` directory.

I felt this was required as the alternatives I considered are painfully slow:
* `dbt --version` does a network request which easily goes over the 500ms limit (and could be worse with slow/no internet connection).
* `pip freeze` is faster, but noticeably slow (450ms).
* `pip show` is slower than `pip freeze`.
* Running a Python script like `from dbt.version import __version__;print(__version__);` is slow due to the size of dbt.

All in all, I'm timing the dbt module at 50ms. This is still quite slow compared to other modules, so perhaps we should remove the `sql` extension from `detect_extensions` and let users configure this one themselves (as `sql` files are what it's most likely to trigger this module).

The dbt project name is fetched from a `dbt_project.yml` configuration file which resides in the root of every dbt project. I wanted to display the project name also when navigating sub-directories of the project. So I implemented a configuration variable `project_dirs` for users to specify directories that may contain a `dbt_project.yml` file. If we are in a directory that `starts_with` any of the directories in `project_dirs`, we use it to look for `dbt_project.yml`. I believed this was easier to implement than a solution that searches in all possible parent directories.

As dbt recommends using an organization name as project name, you could assume that they expect people to have a single project per organization, which means the list of `project_dirs` shouldn't be too large.

#### Motivation and Context

From a personal side, I work with a data team that uses dbt and during their onboarding process we instruct them how to setup their terminal. I love starship and having it support dbt would allow me to recommend it for all team members to install. dbt updates are frequent and folks have sometimes not realized which version they are on.

Moreover, more features can be added to this module in the future to be even more useful (from the top of my head):
* The display of the default profile/target.
* Display results of latest dbt execution.

From a more broader perspective, [dbt](https://github.com/dbt-labs/dbt-core) has around 5.8k stars on GitHub currently, and is well integrated into the data ecosystem: Data Warehouse companies like Snowflake and AWS (Redshift) have hosted articles/labs on working with dbt, and other tools in the ecosystem have integrations to work with it, like [materialize](https://materialize.com/blog/introducing-dbt-materialize/) or [Apache Airflow ](https://github.com/apache/airflow/tree/main/airflow/providers/dbt).

image

image

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -304,6 +304,32 @@
         }
       ]
     },
+    "dbt": {
+      "default": {
+        "detect_extensions": [
+          "sql"
+        ],
+        "detect_files": [
+          "dbt_project.yml"
+        ],
+        "detect_folders": [],
+        "disabled": false,
+        "format": "via [$symbol($version )(\\($project\\) )]($style)",
+        "project_dirs": [],
+        "python_binary": [
+          "python",
+          "python3"
+        ],
+        "style": "fg:#FF694A bold",
+        "symbol": "📊 ",
+        "version_format": "v${raw}"
+      },
+      "allOf": [
+        {
+          "$ref": "#/definitions/DbtConfig"
+        }
+      ]
+    },
     "deno": {
       "default": {
         "detect_extensions": [],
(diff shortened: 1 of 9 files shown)


tomasfarias:
Unfortunately, I don't have access to a Windows machine to look at the test failures in Windows. I believe it may have to do with the way I'm handling paths and directory creation during testing. I could try to get a Windows machine in a few days/a week, but in the meantime I'm open to any suggestions.

PR #4470, opened 2022-10-10
build(deps): update gitoxide crates

davidkna:
#### Description

`fs-walkdir-single-threaded` was replaced with an opposite `fs-walkdir-parallel` in the gitoxide update which is now part of the default `max-perfomance` feature level. As far as I know it's not possible to disable a single feature from a set, so to fix this and still benefit from most of the performance improvements, I've basically pulled in #4334. Alternatively, for the time being, just enabling `max-perfomance-safe` would likely be sufficient.

#### Motivation and Context

Closes #4480

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -396,9 +396,9 @@ dependencies = [
 
 [[package]]
 name = "compact_str"
-version = "0.4.1"
+version = "0.6.1"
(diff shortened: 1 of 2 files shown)


andytom (approved):
LGTM

PR #4486, opened 2022-10-15
feat(python): Add detect env vars option 

### LINKED ISSUE #4471: Python module does not show version in conda environment subdirectories
#### Current Behavior

The Python module uses the `VIRTUAL_ENV` (https://github.com/starship/starship/pull/584) variable to detect when it is in an activated virtual environment; but Conda uses `CONDA_DEFAULT_ENV`. This means that the Python module will not activate unless there is some other active detection mechanism (file etc).

I'm aware of the Conda module, but that doesn't display the Python version. 

#### Expected Behavior
It should use the CONDA_DEFAULT_ENV. 

#### Possible Solution
Use either of VIRTUAL_ENV or CONDA_DEFAULT_ENV

#### Environment
- Starship version: 1.10.3
- zsh version: zsh 5.8.1 (x86_64-ubuntu-linux-gnu)
- Operating system: Ubuntu 22.04
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.63.0
- Rust channel:  release
- Build Time: 2022-09-08 17:00:06 +00:00

### LINKED ISSUE #6197: Detect pyenv shell outside the python source tree
I would like to see the currently selected `pyenv` version activated via the `pyenv shell` command outside the Python project source tree. 

Detect the `PYENV_SHELL` environment variable and display the Python module if it is set. This feature can be opt-in and enabled via starship python module configuration (e.g. `detect_pyenv_shell = true`)

andytom:
#### Description

Allow the env vars that trigger the python module to be configured, if an empty list is passed then the module will fall back to just triggering based on the configured files, folders and extensions.

#### Motivation and Context

Closes #4471
Closes #6198
Closes #6197

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3672,6 +3672,7 @@ By default, the module will be shown if any of the following conditions are met:
 | `detect_extensions`  | `['py']`                                                                                                     | Which extensions should trigger this module                                            |
 | `detect_files`       | `['.python-version', 'Pipfile', '__init__.py', 'pyproject.toml', 'requirements.txt', 'setup.py', 'tox.ini']` | Which filenames should trigger this module                                             |
 | `detect_folders`     | `[]`                                                                                                         | Which folders should trigger this module                                               |
+| `detect_env_vars`    | `["VIRTUAL_ENV"]`                                                                                            | Which environmental variables should trigger this module                               |
 | `disabled`           | `false`                                                                                                      | Disables the `python` module.                                                          |
 
 ::: tip
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1396,6 +1396,9 @@
     },
     "python": {
       "default": {
+        "detect_env_vars": [
+          "VIRTUAL_ENV"
+        ],
         "detect_extensions": [
           "py"
         ],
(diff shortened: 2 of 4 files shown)


docs/config/README.md
```diff
@@ -2948,6 +2948,7 @@ By default the module will be shown if any of the following conditions are met:
 | `detect_extensions`  | `["py"]`                                                                                                     | Which extensions should trigger this module                                            |
 | `detect_files`       | `[".python-version", "Pipfile", "__init__.py", "pyproject.toml", "requirements.txt", "setup.py", "tox.ini"]` | Which filenames should trigger this module                                             |
 | `detect_folders`     | `[]`                                                                                                         | Which folders should trigger this module                                               |
+| `detect_env_vars`    | `["VIRTUAL_ENV"]`                                                                                            | Which environmental varialbes should trigger this module                               |
```
jankatins, line 2951: Typo: varialbes

andytom:
> If you want to add this to more modules, I would prefer if the env list would be integrated into `ScanDir` and maybe even add a macro to further reduce boilerplate.
> Alternatively, I just investigated if handling the `detect_` options via a flattened shared sub-`struct` works, but the serde implementation has some issues and doesn't properly propagate defaults.

I tried the same thing but I think I hit the same issue with the defaults but your suggestion has given me another idea.

one0oneChen:
starship 官方指南高级配置 搜索有关 $conda的配置

davidkna (approved):

PR #4488, opened 2022-10-15
feat(k8s): Add detect env vars option

### LINKED ISSUE #4898: Allow controlling Kubernetes module with environment variable

I would like to use the Kubernetes module but only have the prompt show when a specific environment variable is present. 

There are two approaches I can think of:

1. Add a `detect_env` configuration property where the value is the name of the environment variable to check for
2. Add a global method to disable modules using an environment variable

There's really no alternative that I've been able to find. I don't want Kubernetes context information always showing up in my prompt, and I'd prefer to use an environment variable to control it. In my specific use-case, I plan to use `direnv` and `.envrc.local` files to control when the context appears (i.e. only in repositories that are related to Kubernetes).

andytom:
#### Description

Have added the option to trigger the k8s module based on what env vars are set, this has been done in a backwards compatible way so if nothing is changed from the defaults the module will still behave the same way as before. This is similar to what I did in #4486 for the python module and if goes well I'd like to rollout to other modules.

#### Motivation and Context

Closes #4238
Closes #4898

#### Checklist:

### DIFF
--- a/src/modules/kubernetes.rs
+++ b/src/modules/kubernetes.rs
@@ -107,6 +107,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     };
 
+    // We check that detect_env_vars is not empty here first since calling any on an amptry
+    // iternator returns false, since we want to preserve backwards compatibiliTy we default to
+    // being enabled.
+    let have_env_vars = !config.detect_env_vars.is_empty()
+        && config
+            .detect_env_vars
+            .iter()
+            .any(|e| context.get_env(e).is_some());
+
     // If we have some config for doing the directory scan then we use it but if we don't then we
     // assume we should treat it like the module is enabled to preserve backward compatibility.
     let have_scan_config = [
@@ -127,7 +136,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         })
     });
 
-    if !is_kube_project.unwrap_or(true) {
+    if !is_kube_project.unwrap_or(true) && !have_env_vars {
         return None;
     }
 
@@ -320,7 +329,7 @@ users: []
     }
 
     #[test]
-    fn test_none_when_no_detected_files_or_folders() -> io::Result<()> {
+    fn test_none_when_no_detected_files_folders_or_env_vars() -> io::Result<()> {
         let dir = tempfile::tempdir()?;
 
         let filename = dir.path().join("config");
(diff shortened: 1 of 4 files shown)


Hustenbonbon:
Would you be open to set the default to `[$KUBEDONFIG]` ? Most people I know using kubernetes and starship set their kubernetes context through this env-var on a per-directory basis (with direnv mostly), so I assume it would be a good default

davidkna:
The way this is currently implemented, that could make it count as a breaking change.

davidkna (approved):
LGTM. There seems to be some demand for this specifically, so let's merge this one and sort [how this is implemented across the board later](https://github.com/starship/starship/pull/4486).

jankatins:
Needs a rebase now that #4550 is merged.

jankatins:
Will this be needed if https://github.com/starship/starship/pull/4439 is merged? @davidkna @andytom

src/modules/kubernetes.rs
```diff
@@ -361,6 +370,19 @@ users: []
...
+        let actual_env_var = ModuleRenderer::new("kubernetes")
+            .path(empty_dir.path())
+            .env("KUBECONFIG", filename.to_string_lossy().as_ref())
+            .env("TEST_K8S_ENV", "foo")
+            .config(toml::toml! {
+                [kubernetes]
+                disabled = false
+                detect_env_vars = ["TEST_K8S_ENV"]
+            })
+            .collect();
```
jankatins, line 384: Testing the negative case would probably also be nice: without the `.env("TEST_K8S_ENV", "foo")` it should be disabled.

andytom: Have updated the tests to cover this case with the other negative tests.

src/modules/kubernetes.rs
```diff
@@ -107,6 +107,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
     };
 
+    // We check that detect_env_vars is not empty here first since calling any on an amptry
+    // iternator returns false, since we want to preserve backwards compatibiliTy we default to
+    // being enabled.
+    let have_env_vars = !config.detect_env_vars.is_empty()
+        && config
+            .detect_env_vars
+            .iter()
+            .any(|e| context.get_env(e).is_some());
```
davidkna, line 117: This has a helper in `Context` now:
```suggestion
    let have_env_vars = context.detect_env_vars(&config.detect_env_vars);
```

andytom (commented):
Adding a comment to suggest fixes.

davidkna:
Thanks for the contribution @andytom!

PR #4490, opened 2022-10-15
feat(nu): enable right prompt

### LINKED ISSUE #3982: Support right_format in nushell

nibon7:
#### Description
This PR enables right prompt of nushell. It depends on nushell/reedline#492 and nushell/nushell#6781.

I think it's ready for review since both nushell/reedline#492 and nushell/nushell#6781 have been merged by upstream. @davidkna 

#### Motivation and Context
Closes #3982 

* `new_line` `line_break`
Screenshot from 2022-10-18 15-56-15

* `new_line`  ~~`line_break`~~
Screenshot from 2022-10-18 16-00-28

* ~~`new_line`~~  `line_break`
Screenshot from 2022-10-18 15-55-51

* ~~`new_line`~~ ~~`line_break`~~
Screenshot from 2022-10-18 15-55-27

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -261,7 +261,9 @@ not explicitly used in either `format` or `right_format`.
 Note: The right prompt is a single line following the input location. To right align modules above
 the input line in a multi-line prompt, see the [`fill` module](/config/#fill).
 
-`right_format` is currently supported for the following shells: elvish, fish, zsh, xonsh, cmd.
+`right_format` is currently supported for the following shells: elvish, fish, zsh, xonsh, cmd, nushell.
+
+Note: Nushell 0.71.0 or later is required
 
 ### Example
 
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -12,7 +12,28 @@ let-env PROMPT_COMMAND = {
     ^::STARSHIP:: prompt $"--cmd-duration=($env.CMD_DURATION_MS)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=($width)"
 }
 
-# Not well-suited for `starship prompt --right`.
-# Built-in right prompt is equivalent to $fill$right_format in the first prompt line.
-# Thus does not play well with default `add_newline = True`.
-let-env PROMPT_COMMAND_RIGHT = {''}
+# Whether we can show right prompt on the last line
+let has_rprompt_last_line_support = (version).version >= 0.71.0
+
+# Whether we have config items
+let has_config_items = (not ($env | get -i config | is-empty))
+
+if $has_rprompt_last_line_support {
+    let config = if $has_config_items {
+        $env.config | upsert render_right_prompt_on_last_line true
+    } else {
+        {render_right_prompt_on_last_line: true}
+    }
+    {config: $config}
+} else {
+    { }
+} | load-env
+
+let-env PROMPT_COMMAND_RIGHT = {
+    if $has_rprompt_last_line_support {
+        let width = (term size).columns
+        ^::STARSHIP:: prompt --right $"--cmd-duration=($env.CMD_DURATION_MS)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=($width)"
+    } else {
+        ''
+    }
+}


src/init/starship.nu
```diff
@@ -6,13 +6,16 @@ let-env PROMPT_MULTILINE_INDICATOR = (^::STARSHIP:: prompt --continuation)
 # TODO: Also Use starship vi mode indicators?
 let-env PROMPT_INDICATOR = ""
 
+# Render the right prompt on the last line of the prompt.
+let-env config = ($env.config | upsert render_right_prompt_on_last_line true)
```
davidkna, line 10: This will fail if the user hasn't set `$config` before sourcing the file.

davidkna (changes requested):
Thanks for making this work!
When testing this with a minimal nushell config, I found a small issue - please fix it.
Please also either check the nushell version before setting the config and enabling the right prompt, or else please update the minimally supported nushell version in the docs.

src/init/starship.nu
nibon7: Sorry for the late response. I've checked the version and `$env.config`.

src/init/starship.nu
```diff
@@ -12,7 +12,19 @@ let-env PROMPT_COMMAND = {
...
 
-# Not well-suited for `starship prompt --right`.
-# Built-in right prompt is equivalent to $fill$right_format in the first prompt line.
-# Thus does not play well with default `add_newline = True`.
-let-env PROMPT_COMMAND_RIGHT = {''}
+# The minimum required version is 0.71.0
+let has_right_prompt_support = ((version).version >= 0.71.0)
+
+# Whether we enable the config
+let config_enabled = (($env | get -i config.render_right_prompt_on_last_line) == true)
```
davidkna, line 19: I would prefer to attempt to set `render_right_prompt_on_last_line` to `true`, and to allow users to set it to `false` based on their preferences and keep the rprompt.

nibon7: So we just need to delete this line?

nibon7: Shall we set it to true in `starship.nu`？

nibon7: > and to allow users to set it to `false` based on their preferences and keep the rprompt.

Seems there's no need to check version too :joy:

davidkna: Yes.

docs/advanced-config/README.md
```diff
@@ -261,7 +261,9 @@ not explicitly used in either `format` or `right_format`.
 Note: The right prompt is a single line following the input location. To right align modules above
 the input line in a multi-line prompt, see the [`fill` module](/config/#fill).
 
-`right_format` is currently supported for the following shells: elvish, fish, zsh, xonsh, cmd.
+`right_format` is currently supported for the following shells: elvish, fish, zsh, xonsh, cmd, nushell.
+
+Note: If you want right prompt in nushell behaves like fish or zsh, you could upgrade nushell to 0.71.0 or later.
```
davidkna, line 266: Please also update the docs

src/init/starship.nu
```diff
@@ -12,7 +12,24 @@ let-env PROMPT_COMMAND = {
...
+        {render_right_prompt_on_last_line: true}
+    }
+    {config: $config}
+} else {
+    { }
+} | load-env
+
+let-env PROMPT_COMMAND_RIGHT = {
+    let width = (term size).columns
+    ^::STARSHIP:: prompt --right $"--cmd-duration=($env.CMD_DURATION_MS)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=($width)"
```
davidkna, line 34: I would prefer to enable this for nu 0.71+.
```suggestion
    if $has_rprompt_last_line_support {
        let width = (term size).columns
        ^::STARSHIP:: prompt --right $"--cmd-duration=($env.CMD_DURATION_MS)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=($width)"
    } else {
        ''
    }
```

davidkna (changes requested):

davidkna (approved):
LGTM. Thanks for handling this!

PR #4503, opened 2022-10-19
refactor(nu): simplify getting term width

verdayne:
#### Description

[starship.nu L11](https://github.com/starship/starship/blob/0b173c263e510be572db3ac88f3df7ca01aa486a/src/init/starship.nu#L11) uses multiple piping commands to pull the terminal width from `term size` and convert the result into a string. These actions can be condensed and combined into one of nushell's subexpressions, which makes the code cleaner and easier to read. The proposal is shown below.

Before:
```nu
let width = (term size | get columns | into string)
```
After:
```nu
let width = (term size).columns
```

#### Motivation and Context

Remove piping chain and reduce points of failure to only `term size` and the requested value `columns`.

#### Checklist:

### DIFF
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -8,7 +8,7 @@ let-env PROMPT_INDICATOR = ""
 
 let-env PROMPT_COMMAND = {
     # jobs are not supported
-    let width = (term size | get columns | into string)
+    let width = (term size).columns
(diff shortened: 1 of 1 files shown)


davidkna (approved):

mati865:
This does fix the error when  using the latest release of Nushell, can it be merged?

davidkna:
@mati865 It's already fixed on the master branch.

mati865:
Right, sorry. Mislooked at the diff.

davidkna:
Thanks for the clean-up @verdayne!

verdayne:
@davidkna thank you for accepting my suggestion

PR #4477, opened 2022-10-13
fix(nu): remove -c parameter from `term size`

nibon7:
#### Description
The -c parameter has been removed through nushell/nushell#6651

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

PR #4504, opened 2022-10-19
fix: command to get terminal width in nushell init script

chdalski:
Signed-off-by: Christoph Dalski <chdalski.coding@gmail.com>

#### Description
With version 0.70.0 of nushell the previous command (`(term size | get columns | into string)`), to get the terminal width, broke.
The replaced version (`(term size).columns`) works with 0.70.0 and below (tested with 0.69.1).

#### Motivation and Context
The change is required because starship prompt will fail to initialize with version 0.70.0, otherwise.

Changed the command in a term with version 0.70.0 were it breaks. After the change it obviously wasn't breaking anymore.
Tried the same thing on my laptop were still version 0.69.1 is running. Both implementations are working fine.

#### Checklist:

I didn't update the tests and the documentation because in this particular case there is no need, i guess?

### DIFF
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -8,7 +8,7 @@ let-env PROMPT_INDICATOR = ""
 
 let-env PROMPT_COMMAND = {
     # jobs are not supported
-    let width = (term size | get columns | into string)
+    let width = (term size).columns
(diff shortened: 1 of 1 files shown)


davidkna:
The expression broke because it used the `-c` flag, which was removed (#4477, https://github.com/nushell/nushell/pull/6651). This still looks like an improvement, but I'm closing this PR because your changes match the earlier PR #4503.

PR #4505, opened 2022-10-19
feat(kubernetes): Allow to filter context names

DanielKneipp:
Allow filtering of kubernetes context names

#### Description
This change should allow the user to specify a list of namespace names to be ignored via the variable `filter_contexts`. Any context that matches a name on this list will be ignored. 

#### Motivation and Context
This is to allow the user to ignore regular local contexts like `docker-desktop`. Basically, I want to have a way to ignore the default context I have.

Tested with `cargo test kubernetes`

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -792,6 +792,7 @@
         "detect_files": [],
         "detect_folders": [],
         "disabled": true,
+        "filter_contexts": [],
         "format": "[$symbol$context( \\($namespace\\))]($style) in ",
         "style": "cyan bold",
         "symbol": "☸ ",
@@ -3509,6 +3510,13 @@
           }
         },
         "detect_folders": {
+          "default": [],
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "filter_contexts": {
           "default": [],
           "type": "array",
           "items": {
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -792,8 +792,8 @@
         "detect_files": [],
         "detect_folders": [],
         "disabled": true,
-        "filter_contexts": [],
         "format": "[$symbol$context( \\($namespace\\))]($style) in ",
+        "ignore_contexts": [],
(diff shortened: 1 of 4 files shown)


davidkna (commented):
This seems fine, but please rename the option to `ignore_contexts` to better align with similar options elswhere.

davidkna (approved):
LGTM, but we may want to wait how https://github.com/starship/starship/pull/614 etc. progress before merging.

DanielKneipp:
No rush, but is that PR related somehow? I know it would allow conditional formatting based on the context name. I couldn't see a way to just ignore a context.

davidkna:
@DanielKneipp I was thinking it might make sense to move the `ignore` option into `display` alongside with the `style` options etc.

jankatins:
See https://github.com/starship/starship/pull/4550 for an updated/combined PR (#614 + #4505)

DanielKneipp:
@davidkna Agreed! this way we can have all context-related config unified. I might have some time to check this out later this weekend. Should I commit on #4550 or is a new PR preferred after this one gets merged? What do you guys think @davidkna @jankatins ?

jankatins:
@DanielKneipp I would appreciate a review of #4550 and it's probably best to wait for a comment by one of the starship maintainer on what direction #4550 should take. :-)

jankatins:
This PR would need a rebase (and probably a rework :-() now that #4550 is merged. I would also really love to have this. @DanielKneipp

jankatins:
@DanielKneipp Do you think you still find time (and interest) to finish this PR or would it be ok if I take this over to rebase and rework it (in a new PR with your name as coauthor in the commit).

DanielKneipp:
Oh fell free @jankatins! By the way, I also [added](https://github.com/DanielKneipp/starship/blob/master/src/configs/kubernetes.rs#L21-L23) some other `ignore` options to mix and match k8s contexts and namespaces

PR #4517, opened 2022-10-22
feat(preset): Add no-nerd-font preset

segevfiner:
#### Description
Add powerline-only-symbols preset.

Feel free to iterate on this and add actual icons for Erlang and Pulumi or chnage the suggested icons to ones you like better.

This will help until the project decides to release a major version with a clear separation between the default preset and Nerd Fonts preset. And will also give a chance to iterate on the new symbols.

cc @chipbuster

#### Motivation and Context

Related to #2563 & #3544

#### Checklist:


chipbuster:
@segevfiner Currently working out a few typos and other things. Two thoughts for this preset:

1. I think the name "powerline only" is a little misleading in this case, because it's really powerline + emoji. What do you think about either "no-nerd-font" or "powerline-emoji-only"?

2. I don't think we need a screenshot for this preset. This is partially because I managed to lose the script I used to take the last set of screenshots (oops),but also because only a few symbols are changed here, and none of those symbols are displayed in the standard preset screenshots. We'd need to create a custom set of commands to show it off, and I just don't think that's worth the confusion when you can read the entire preset in 5 seconds. Do you think this makes sense or would you prefer to do something else?

chipbuster:
Also, for the time being, I've decided to go with "letter 'e' in a circle" for Erlang and "ice cube" for pulumi. I'm not terribly happy with either, but I think it's preferable to keep the convention of 1-2 character widths for a module symbol in everything but the plain text preset.

```
[erlang]
symbol = "ⓔ "

[nodejs]
symbol = "[⬢](bold green) "

[pulumi]
symbol = "🧊 "
```

segevfiner:
> @segevfiner Currently working out a few typos and other things. Two thoughts for this preset:
> 
> 1. I think the name "powerline only" is a little misleading in this case, because it's really powerline + emoji. What do you think about either "no-nerd-font" or "powerline-emoji-only"?
> 2. I don't think we need a screenshot for this preset. This is partially because I managed to lose the script I used to take the last set of screenshots (oops),but also because only a few symbols are changed here, and none of those symbols are displayed in the standard preset screenshots. We'd need to create a custom set of commands to show it off, and I just don't think that's worth the confusion when you can read the entire preset in 5 seconds. Do you think this makes sense or would you prefer to do something else?

1. I'm fine with either name IMHO.
2. A screenshot for this is not really critical I guess. We can produce one manually if we have too. But this is something that can always be added later even if we don't add it now.

davidkna (approved):

andytom (approved):

PR #4520, opened 2022-10-22
docs(config): Add missing default values

Kissaki:
Improvement on inconsistencies and incompleteness #2935

PR #4521, opened 2022-10-22
docs(config): Resolve string inconsistencies

### LINKED ISSUE #2935: Configuration docs often omit double quotes on default values, which makes spaces vanish
#### Current Behavior
The [configuration docs](https://starship.rs/config/) usually display default values using correct TOML syntax, meaning strings are enclosed in quotes. For example, the [Erlang module](https://starship.rs/config/#erlang) does this.

Option | Default
-|-
`format` | `"via [$symbol($version )]($style)"`
`version_format` | `"v${raw}"`
`detect_files` | `["rebar.config", "elang.mk"]`

Some modules even use single-quoted strings (e.g. if there are backslashes), such as the [Singularity module](https://starship.rs/config/#singularity), which includes

Option | Default
-|-
`format` | `'[$symbol\[$env\]]($style) '`
`symbol` | `""`

But other modules omit the quotes. This is especially an issue when a value has a space after it (as the `format` key often does), as it means the space isn't visible. For example, the [Shell module](https://starship.rs/config/#shell) looks like

Option | Default
-|-
`bash_indicator` | `bsh`
`fish_indicator` | `fsh`
`unknown_indicator` | 
`format` | `$indicator`

Note how not only is the `format` key missing the space after it, but the `unknown_indicator` shows as having no value at all when its default value is actually an empty string. The space in the `format` key exists in the source, but is not preserved in the HTML rendering:

https://github.com/starship/starship/blob/40cb667b9dab7fb9b7ea31a32e24013af2ed4e09/docs/config/README.md#L2592-L2593

#### Expected Behavior
Every default config value specified in the documentation should use correct TOML value syntax, meaning it should use quotes for strings, including for empty strings. I should be able to take any default value and copy it directly into my `starship.toml`.

#### Other notes
I ran into this because I was trying to figure out where the space before my `character` was coming from. The configuration docs for `shell` don't show the space, so I didn't realize that's where it was coming from. `starship explain` looks like

```
 ~  (<1ms)  -  The current working directory
   (<1ms)   -  The currently used shell indicator
 ❯  (<1ms)  -  A character (usually an arrow) beside where the text is entered i
               n your terminal
```

and I wasn't sure if the "shell indicator" line was actually including any characters (though in retrospect, if it was truly empty it would have been omitted). I checked the docs and didn't see any space, and it wasn't until I thought to run `starship print-config` and double-check the `[shell]` module that I figured it out.

Kissaki:
Resolve inconsistencies in string value declarations.

Consistently use literal strings like we recommend, and like `starship print-config` outputs.

~Based on/includes PR #4520~

Fixes #2935

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -84,9 +84,9 @@ The name of a variable can only contain letters, numbers and `_`.
 
 For example:
 
-- `$version` is a format string with a variable named `version`.
-- `$git_branch$git_commit` is a format string with two variables named `git_branch` and `git_commit`.
-- `$git_branch $git_commit` has the two variables separated with a space.
+- `'$version'` is a format string with a variable named `version`.
+- `'$git_branch$git_commit'` is a format string with two variables named `git_branch` and `git_commit`.
+- `'$git_branch $git_commit'` has the two variables separated with a space.
 
 #### Text Group
 
@@ -99,9 +99,9 @@ In the second part, which is enclosed in a `()`, is a [style string](#style-stri
 
 For example:
 
-- `[on](red bold)` will print a string `on` with bold text colored red.
-- `[⌘ $version](bold green)` will print a symbol `⌘` followed by the content of variable `version`, with bold text colored green.
-- `[a [b](red) c](green)` will print `a b c` with `b` red, and `a` and `c` green.
+- `'[on](red bold)'` will print a string `on` with bold text colored red.
+- `'[⌘ $version](bold green)'` will print a symbol `⌘` followed by the content of variable `version`, with bold text colored green.
+- `'[a [b](red) c](green)'` will print `a b c` with `b` red, and `a` and `c` green.
 
 #### Style Strings
 
@@ -122,10 +122,10 @@ A conditional format string wrapped in `(` and `)` will not render if all variab
 
 For example:
 
-- `(@$region)` will show nothing if the variable `region` is `None` or empty string, otherwise `@` followed by the value of region.
-- `(some text)` will always show nothing since there are no variables wrapped in the braces.
-- When `$combined` is a shortcut for `\[$a$b\]`, `($combined)` will show nothing only if `$a` and `$b` are both `None`.
-  This works the same as `(\[$a$b\] )`.
+- `'(@$region)'` will show nothing if the variable `region` is `None` or empty string, otherwise `@` followed by the value of region.
+- `'(some text)'` will always show nothing since there are no variables wrapped in the braces.
+- When `$combined` is a shortcut for `\[$a$b\]`, `'($combined)'` will show nothing only if `$a` and `$b` are both `None`.
+  This works the same as `'(\[$a$b\] )'`.
 
 #### Special characters
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -177,7 +177,7 @@ For example:
 
 Many modules have `detect_extensions`, `detect_files`, and `detect_folders` variables. These take
 lists of strings to match or not match. "Negative" options, those which should not be matched, are
-indicated with a leading "!" character. The presence of _any_ negative indicator in the directory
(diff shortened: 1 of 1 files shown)


Kissaki:
Rebased onto merged #4520

docs/config/README.md
```diff
@@ -3477,14 +3498,14 @@ To enable it, set `disabled` to `false` in your configuration file.
...
-| `signal_symbol`             | `"⚡"`                                                                         | The symbol displayed on any signal                                    |
-| `style`                     | `"bold red"`                                                                  | The style for the module.                                             |
+| `format`                    | `'[$symbol$status]($style) '`                                                 | The format of the module                                              |
+| `symbol`                    | `'❌'`                                                                         | The symbol displayed on program error                                 |
+| `success_symbol`            | `''`                                                                          | The symbol displayed on program success                               |
+| `not_executable_symbol`     | `'🚫'`                                                                         | The symbol displayed when file isn’t executable                       |
```
davidkna, line 3504: As far as I can tell, isn't `'` the proper character to use for an apostrophe?

davidkna (changes requested):

docs/config/README.md
Kissaki: This is about the replacement of `'` with `’` as the supposedly correct character, right?

Looking at [Wikipedia Apostrophe](https://en.wikipedia.org/wiki/Apostrophe) and specifically [the section typographic form](https://en.wikipedia.org/wiki/Apostrophe#Typographic_form), it seems that both are useable and correct; where `’` is the explicit apostrophe, and `'` is a generalized character that can be used for an apostrophe amongst other meanings/symbol generalization.

> With the invention of the [typewriter](https://en.wikipedia.org/wiki/Typewriter), a "neutral" or "straight" shape quotation mark, ', was created to represent a number of different glyphs with a single keystroke: the apostrophe, both the opening and the closing single [quotation marks](https://en.wikipedia.org/wiki/Quotation_mark), the single [primes](https://en.wikipedia.org/wiki/Prime_(symbol)), and on some typewriters even the [exclamation point](https://en.wikipedia.org/wiki/Exclamation_point) (by backspacing and overprinting with a period). This is known as the typewriter apostrophe or 'vertical apostrophe'.

It seems I was misled a bit about only one being explicitly, typographically correctness, as [that is the case in German](https://de.wikipedia.org/wiki/Apostroph#Typografisch_korrekt).

I think there would still be value in using an apostrophe over the typewriter apostrophe for differentiation to the code single quote - but it is somewhat questionable to do so only in this document and not everywhere leading to inconsistencies, and when it is not explicitly incorrect to use either. And when one, presumably, sees `'` in en texts more often.

Shall I remove the apostrophe change commit form this PR?

davidkna: Please remove that change. I actually only looked up the Unicode definitions, and "apostrophe" seemed to be preferable over any "right quotation mark".

docs/config/README.md
```diff
@@ -71,6 +73,50 @@ os.setenv('STARSHIP_CACHE', 'C:\\Users\\user\\AppData\\Local\\Temp')
...
+```
+
+When using line breaks, multi-line declarations can be used.
+For example, if you want to print a `$` symbol on a new line, the following values for `format` are equivalent:
+
+```toml
+# with literal string (recommended)
+format = '''
+
+\$'''
```
davidkna, line 109: I feel like `"""` should be preferred over `'''`.  Using a construction like in the [`$all` explanation](https://starship.rs/config/#default-prompt-format) seems to be somewhat popular among users.

davidkna (changes requested):

docs/config/README.md
Kissaki: What do you think of this?

I dropped the recommendation, and described the possibility to use such escaped newlines.

davidkna (approved):

andytom:
Thank you for your contribution @Kissaki

PR #4527, opened 2022-10-24
fix(git): check `tag_disabled` option

### LINKED ISSUE #4526: `tag_disabled` option of `git_commit` don't take effect
#### Current Behavior
The `git_commit` module will always show tag info.

#### Expected Behavior
The tag info will be suppressed when `tag_disabled` is `true`.

I have a quick look at source of `git_commit`, I think the `tag_disabled` option was not be checked at all. And it seems to be removed in https://github.com/starship/starship/commit/ac55a01d0ffe907ef7af48c9597c0bca4dbd8c69#diff-bda7d97daa3f0d57f24f265115a7b6e87b59e54350a55ad21c1f3dc57b8cc202L23-L40.

#### Possible Solution
Check the `tag_disabled` option at the start of function `git_tag`.

#### Environment
- Starship version: 1.11.0
- fish version: fish, version 3.5.1
- Operating system: Mac OS 13.0.0
- Terminal emulator: iTerm.app 3.5.0beta7
- Git Commit Hash: 
- Branch/Tag: master
- Rust Version: rustc 1.64.0
- Rust channel:  release
- Build Time: 2022-10-15 01:40:28 +00:00

```fish
# starship initialize {{{
if type -q starship
  starship init fish | source
end
# starship initialize }}}
# vim:ts=2:sw=2:fdm=marker:et
```

#### Starship Configuration

```toml

# General {{{
format = """ \
$all\
$git_branch$git_commit$git_state$git_status\
$fill$status$cmd_duration\
$line_break $character\
"""
right_format = "$time"
add_newline = false

[character]
success_symbol = "[](bold green)"
error_symbol = "[](bold red)"
vimcmd_symbol = "[](bold green)"
vimcmd_visual_symbol = "[](bold green)"
vimcmd_replace_symbol = "[](bold purple)"
vimcmd_replace_one_symbol = "[](bold yellow)"

[cmd_duration]
format = "took [$duration]($style) "
min_time = 10_000

[directory]
format = "[$path]($style)[$read_only]($read_only_style) "
read_only = " "

[fill]
symbol = "-"

[hostname]
disabled = false

[jobs]
symbol = " "
style = "bold yellow"

[shell]
disabled = true

[shlvl]
disabled = false
symbol = " "

[status]
disabled = false

[sudo]
disabled = false

[time]
disabled = false
time_format = "%a %r"

[username]
disabled = false
# General }}}

# Git {{{
[git_branch]
symbol = " "

[git_commit]
format ="on [\\($hash$tag\\)]($style) "
tag_disabled = true
tag_symbol = " 笠"

[git_status]
format = "$ahead_behind$conflicted$stashed$renamed$modified$deleted$untracked$staged"
# all with trailing space is needed
ahead = "⇡ $count "
behind = "⇣ $count "
diverged = "⇕ $ahead_count $behind_count "
conflicted = " "
stashed = "⚑ $count "
staged = "● $count "
renamed = " $count "
deleted = " $count "
untracked = " $count "
modified = " $count "
# }}}
```

wangl-cc:
#### Description
Check `tag_disabled`, don't show tag info when it is true. I'm not familiar with rust, so the new test is just a modified copy of the below one.

#### Motivation and Context

Closes #4526 

The first starship is built from master, and the second is built from this PR.
截屏2022-10-25 02 35 43

#### Checklist:

### DIFF
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -51,6 +51,9 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn git_tag(repo: &Repo, config: &GitCommitConfig) -> Option<String> {
+    if config.tag_disabled {
+        return None;
+    }
     // allow environment variables like GITOXIDE_OBJECT_CACHE_MEMORY and GITOXIDE_DISABLE_PACK_CACHE to speed up operation for some repos
     let mut git_repo = repo.open().apply_environment();
     git_repo.object_cache_size_if_unset(4 * 1024 * 1024);
@@ -199,6 +202,42 @@ mod tests {
 
         assert_eq!(expected, actual);
         repo_dir.close()
+    }
+
+    #[test]
+    fn test_render_commit_hash_with_tag_disabled() -> io::Result<()> {
+        let repo_dir = fixture_repo(FixtureProvider::Git)?;
+
+        let mut git_commit = create_command("git")?
+            .args(&["rev-parse", "HEAD"])
+            .current_dir(&repo_dir.path())
+            .output()?
+            .stdout;
+        git_commit.truncate(7);
+        let commit_output = str::from_utf8(&git_commit).unwrap().trim();
+
+        create_command("git")?
+            .args(&["describe", "--tags", "--exact-match", "HEAD"])
+            .current_dir(&repo_dir.path());
+
+        let actual = ModuleRenderer::new("git_commit")
+            .config(toml::toml! {
+                [git_commit]
+                    only_detached = false
+                    tag_symbol = ""
+            })
+            .path(&repo_dir.path())
+            .collect();
+
+        let expected = Some(format!(
+            "{} ",
+            Color::Green
+                .bold()
+                .paint(format!("({})", commit_output.trim()))
+        ));
+
+        assert_eq!(expected, actual);
+        Ok(())
     }
 
     #[test]
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -29,7 +29,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             })
             .map(|variable| match variable {
                 "hash" => Some(Ok(git_hash(context.get_repo().ok()?, &config)?)),
-                "tag" => Some(Ok(format!(
+                "tag" if !config.tag_disabled => Some(Ok(format!(
                     "{}{}",
                     config.tag_symbol,
                     git_tag(context.get_repo().ok()?, &config)?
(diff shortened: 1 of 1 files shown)


src/modules/git_commit.rs
```diff
@@ -51,6 +51,9 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn git_tag(repo: &Repo, config: &GitCommitConfig) -> Option<String> {
+    if config.tag_disabled {
+        return None;
+    }
```
davidkna, line 56: I would prefer doing the check in `map` as `"tag" if !config.tag_disabled => `

src/modules/git_commit.rs
```diff
@@ -201,6 +204,42 @@ mod tests {
         repo_dir.close()
     }
 
+    #[test]
+    fn test_render_commit_hash_with_tag_disabled() -> io::Result<()> {
+        let repo_dir = fixture_repo(FixtureProvider::Git)?;
+
```
davidkna, line 210: The test needs to create a tag:
```suggestion
    create_command("git")?
            .args(&["tag", "v1", "-m", "Testing tags"])
            .current_dir(&repo_dir.path())
            .output()?;
```

src/modules/git_commit.rs
```diff
@@ -201,6 +204,42 @@ mod tests {
...
+        create_command("git")?
+            .args(&["describe", "--tags", "--exact-match", "HEAD"])
+            .current_dir(&repo_dir.path());
+
+        let actual = ModuleRenderer::new("git_commit")
+            .config(toml::toml! {
+                [git_commit]
+                    only_detached = false
+                    tag_symbol = ""
```
davidkna, line 227: `tag_disabled` does not default to `true`: 
```suggestion
                    tag_symbol = ""
                    tag_disabled=true
```

davidkna (changes requested):

src/modules/git_commit.rs
wangl-cc: Thanks for you suggestion. Updated.

src/modules/git_commit.rs
wangl-cc: I think it it default to `true`:
https://github.com/starship/starship/blob/c3cd499a30ccefcc14eaa60156da7ef98ab6390c/src/configs/git_commit.rs#L31

src/modules/git_commit.rs
wangl-cc: Sorry about missing this. I think the test `test_render_commit_hash_with_tag_enabled` also seems to miss this, so I add.

src/modules/git_commit.rs
davidkna: I think I would prefer changing the default to `false`, either way this is something of a breaking change from v0.10. Any thoughts for/against @starship/astronauts?

chipbuster: I concur that changing it to false would probably be better. I do think we should probably merge this now as a bugfix and change the default with the 2.0.0 release.

davidkna: Ok. Let's revisit this later then.

davidkna (approved):

andytom:
Thank you for your contribution @wangl-cc and thank you for reviewing @davidkna and @chipbuster.

PR #4535, opened 2022-10-26
fix: don't attempt to display cmd_duration notification if in TTY

### LINKED ISSUE #4148: Don't attempt to display cmd_duration notification if in TTY
#### Current Behavior
The notification from cmd_duration is attempted to be shown even while in a Linux TTY, where it cannot be displayed. This causes process to hang while it waits for the notification to be displayed even though it never will be (pressing Ctrl+C twice fixes it).

#### Expected Behavior
The notification will not be attempted to be displayed in a TTY.

#### Possible Solution
If on Linux, the notification should only be displayed if in a graphical session: either X11 or Wayland. X11 can be easily determined by checking if the `$DISPLAY` environment variable is set and non-empty, but I am not sure about Wayland.

Alternatively, it is also an option to not display the notification if `$TERM=linux`, which is what is used by the Linux TTY, though I am not sure whether this would work when using a multiplexer like tmux in the TTY.

Regardless of which option is used, it would still probably be good to have a configuration option to disable these checks in order to prevent this from becoming unusable in edge cases.

#### Environment
- Starship version: 1.9.1
- zsh version: zsh 5.9 (x86_64-pc-linux-gnu)
- Operating system: Arch Linux Rolling Release
- Terminal emulator: Alacritty 0.11.0-dev (28e3fc7c)
- Git Commit Hash: b8d8918b
- Branch/Tag: v1.9.1
- Rust Version: rustc 1.61.0 (Arch Linux rust 1:1.61.0-1)
- Rust channel:  release
- Build Time: 2022-06-27 19:26:59 +00:00

```bash
<unknown config>
```

#### Starship Configuration

```toml
format = """
$status$cmd_duration$directory\
$username$hostname\
$singularity\
$kubernetes\
$docker_context\
$terraform\
$nix_shell\
$aws\
$gcloud\
$openstack\
$sudo[❯](bold blue)[❯](bold cyan)[❯](bold green) ${custom.cursor}"""

right_format = """
$python\
$git_state$git_status$git_branch$hg_branch\
$jobs$shlvl$memory_usage$battery$time \
"""

add_newline = false
scan_timeout = 25
command_timeout = 200


[status]
format = '[↦](bold) [$status( \($signal_name\))]($style) [|](bold)'
style = 'bold red'
recognize_signal_code = true
map_symbol = true
disabled = false

[cmd_duration]
min_time = 2_500
format = ' [$duration]($style) [|](bold)'
style = 'bold yellow'
show_notifications = true
min_time_to_notify = 90_000
disabled = false

[directory]
format = ' [$path]($style)[$read_only]($read_only_style) '
style = 'blue'
read_only = ' ✕'
read_only_style = 'bold red'
truncation_length = 6
truncate_to_repo = false
truncation_symbol = '…/'
disabled = false

[username]
format = '[|](bold) [$user]($style) '
style_root = 'bold red'
style_user = 'bold bright-green'
show_always = false
disabled = false

[hostname]
format = 'at [$hostname]($style) '
trim_at = ''
style = 'bold green'
disabled = false

[sudo]
format = '[✦]($style)'
style = 'bold purple'
disabled = true

[custom.cursor]
command = "printf '\\033[5 q'"
description = 'Sets terminal cursor to line'
format = '$output'
when = 'true'
shell = 'sh'
disabled = false


[python]
format = '([$virtualenv]($style) [|](bold) )'
style = 'bold bright-yellow'
disabled = false

[git_state]
format = '[$state]($style)(\([$progress_current/$progress_total]($style)\) )'
rebase = 'RBS'
merge = 'MRG'
revert = 'RVR'
cherry_pick = 'CHR'
bisect = 'BSC'
am = 'AML'
am_or_rebase = 'A/R'
style = 'bold red'
disabled = false

[git_status]
format = '(\[[$all_status$ahead_behind]($style)\] )'
conflicted = '='
ahead = '↑'
behind = '↓'
diverged = '⇕'
untracked = '?'
stashed = '\$'
modified = '!'
staged = '+'
renamed = '»'
deleted = '-'
style = 'bold cyan'
disabled = false

[git_branch]
format = '[ $branch]($style)( [$remote_branch]($style)) [|](bold) '
style = 'bold purple'
truncation_length="10"
only_attached = false
disabled = false

[hg_branch]
format = '[ $branch]($style) [|](bold) '
style = 'bold bright-purple'
truncation_length="10"
disabled = false

[jobs]
format = '[$number ↺]($style) [|](bold) '
style = 'bold bright-blue'
symbol_threshold = 1
number_threshold = 1
disabled = false

[shlvl]
threshold = 2
format = '[$shlvl ↕]($style) [|](bold) '
disabled = false

[memory_usage]
format = '[$ram ]($style) [|](bold) '
style = 'bold red'
threshold = 80
disabled = false

[battery]
format = '[$percentage ]($style) [|](bold) '
disabled = false
[[battery.display]]
threshold = 25
style = 'bold yellow'

[time]
format = '[$time]($style)'
style = 'bright-yellow'
disabled = false

```

gabrielvictorcf:
#### Description
This PR adds the new option `tty_disabled` to the `cmd_duration` module. This option's default is `true` and changes the default behavior of the module when in linux - now, it'll check if user actually is in a graphical session before printing the duration. This new default behavior was added as an option in case it breaks someone's setup - just set `tty_disabled = false` to fallback to the old behavior.

The way a graphical session is detected is by checking if any of these environment variables are set: `["DISPLAY", "WAYLAND_DISPLAY", "MIR_SOCKET"]`. I may be missing some more obscure protocol, but i believe this already reaches most user's setups.

#### Motivation and Context
This change fixes the `starship` process from hanging when it tries to display duration in a TTY.
Closes #4148 

`cargo test --all-targets --all-features` runs ok (no failures). changes where made to the other tests in the `cmd_duration` module to include the `WAYLAND_DISPLAY = wayland-0` environment variable, as the default behavior now requires a graphical session to be in place.

I have not actually tried this with a multiplexer, like tmux, as i'm not sure exactly what corner case they actually generate (aren't the display variables enough?).

#### Checklist:


src/modules/cmd_duration.rs
```diff
@@ -12,6 +12,17 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("cmd_duration");
     let config: CmdDurationConfig = CmdDurationConfig::try_load(module.config);
 
+    if cfg!(target_os = "linux") && config.tty_disabled {
```
chipbuster, line 15: This block disables the entire module if not in a graphical environment, not just the notifications.

gabrielvictorcf: I've moved this block to the `undistract_me` fn where the notification happens and removed the option. What i'm not sure now is how we should test this, if it is even possible. Is it ok to leave the fix without them?

chipbuster:
A quick update: I managed to reproduce this on an Ubuntu 18.04 machine and can confirm that this patch fixes the tendency of the tty to hang indefinitely while trying to deliver a notification.

src/modules/cmd_duration.rs
chipbuster: As far as in-code testing goes, I don't think there's a way to test it with our in-code tests either. Code coverage metrics are not a hard blocker to merging for us, so that's not that important.

I'm of half a mind to move the notification feature to its own module in v2.0.0, just so that we can cleanly separate the code that is functionally untestable (the stuff with notifications) from the duration stuff, which should be 100% testable, but that's not for this PR.

gabrielvictorcf:
> I managed to reproduce the bug on an Ubuntu 18.04 machine and can confirm that this patch fixes the tendency of the tty to hang indefinitely while trying to deliver a notification.

Oh, that's great. I tried reproducing on Mint 19.3 but was unsuccessful. Any further changes or adjustments needed?

chipbuster (approved):
:+1: LGTM

chipbuster:
@gabrielvictorcf Sorry, nothing else needed, I just needed to find the time to review it again (also, finally got to try reviewing from VSCode directly instead of the web interface!).

src/modules/cmd_duration.rs
```diff
@@ -64,12 +65,24 @@ fn undistract_me<'a, 'b>(
...
     elapsed: u128,
 ) -> Module<'a> {
     use notify_rust::{Notification, Timeout};
     use nu_ansi_term::{unstyle, AnsiStrings};
 
     if config.show_notifications && config.min_time_to_notify as u128 <= elapsed {
+        if cfg!(target_os = "linux") {
+            let in_graphical_session = ["DISPLAY", "WAYLAND_DISPLAY", "MIR_SOCKET"]
+                .iter()
+                .map(|&var| context.get_env(var).filter(|val| !val.is_empty()))
```
andytom, line 78: Not a blocker but I think you can use a [find_map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find_map) here to simplify this a bit.

gabrielvictorcf: thanks for the heads-up, done!

gabrielvictorcf: for some reason the commit msg broke, but i meant 'refactor block to use `find_map`'

gabrielvictorcf:
The clippy fails seem unrelated to my commit, looks like clippy got updated to `0.1.65` (launched yesterday) but my branch is a bit behind.

andytom:
> The clippy fails seem unrelated to my commit, looks like clippy got updated to `0.1.65` (launched yesterday) but my branch is a bit behind.


I've merged the latest clippy fixes so if you rebase your branch that should (hopefully) fix them.

src/modules/cmd_duration.rs
```diff
@@ -64,12 +65,24 @@ fn undistract_me<'a, 'b>(
...
     elapsed: u128,
 ) -> Module<'a> {
     use notify_rust::{Notification, Timeout};
     use nu_ansi_term::{unstyle, AnsiStrings};
 
     if config.show_notifications && config.min_time_to_notify as u128 <= elapsed {
+        if cfg!(target_os = "linux") {
+            let in_graphical_session = ["DISPLAY", "WAYLAND_DISPLAY", "MIR_SOCKET"]
+                .iter()
+                .find_map(|&var| context.get_env(var).filter(|val| !val.is_empty()))
```
andytom, line 78: I think this can just be
```suggestion
                .find_map(|&var| context.get_env(var).is_some())
```

gabrielvictorcf: Can't we get a false positive for something like `DISPLAY=` , where the variable is set but the value is currently empty? Does `context.get_env` already check this? Might be a silly edge case but i just want to signal that the length check is intentional.

davidkna: IIRC on Windows it isn't possible to differentiate between empty and missing environment variables, at least in some cases.

chipbuster: In our case, will that present as an empty variable or a `None` variable?

At any rate, that probably won't affect this PR directly because this change is `cfg!`-ed to Linux-only.

chipbuster: I can find at least a few cases where `DISPLAY` is empty but set on the Arch forums. I'm not certain that these cases would impact us (as it appears that it's impossible to launch graphical programs like terminal emulators when this happens), but I think I'm okay with keeping the current code just in case. @andytom Would you object to merging in the current state?

andytom: Yep no problem from me we can always refactor later if we want.

PR #4547, opened 2022-10-30
fix(config): unrecognized config properties don't cause config error

### LINKED ISSUE #4481: Unrecognized sections in starship.toml prevent top-level format from being recognized
#### Current Behavior

When using a config as follows, the specified `format=` at the top level is completely ignored. Other sections work fine.

```toml
format = 'Hello! $character'

# This should be "golang".
[go]
disabled = true

[shell]
disabled = false
```

The prompt then appears as follows. Note that `bsh` is present, as the `[shell]` block in the config has been recognized, as expected.

format ignored

#### Expected Behavior

Only the unrecognized section is ignored and the prompt appears as follows.

expected format

If the `[go]` section in the config above is removed or corrected to `[golang]`, the top-level format is interpreted correctly.

#### Environment

- Starship version: 1.10.3
- bash version: GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin21)
Copyright (C) 2007 Free Software Foundation, Inc.
- Operating system: Mac OS 12.6.0
- Terminal emulator: iTerm.app 3.4.16
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.63.0
- Rust channel:  release
- Build Time: 2022-09-08 17:00:06 +00:00

```bash
export BASH_SILENCE_DEPRECATION_WARNING=1
export EDITOR=$(which nano)
export HISTIGNORE="&"
export PATH=$HOME/bin:/usr/local/bin:$PATH

[ -f /usr/local/etc/bash_completion ] && . /usr/local/etc/bash_completion

eval "$(starship init bash)"
```

#### Starship Configuration

```toml
format = 'Hello! $character'

[go]
disabled = true

[shell]
disabled = false

```

brahmlower:
Fixes an issue where the config was not being loaded when ignored properties are present. This still prints a warning log indicating a possible misconfiguration.

#### Description

~~The deserializer wrapper has been updated such that deserialization of ignored properties does not return an error, and instead returns the `visitor.visit_none()`. This function now includes a warning log statement since the [log statement for reporting the error](https://github.com/starship/starship/blob/master/src/config.rs#L34) as shown in the bug ticket is no longer reached.~~ (this described the initial implementation before PR review)

The `ValueDeserializer` wrapper has been updated so that warnings can be either ignored or cause an `Err` result. At this time, the only warning that gets emitted is the warning from ignored properties during deserialization. This can be used via the new `with_allow_unknown_keys` function.

The configurable warning behavior is used within the generic `from_config` implementation of `ModuleConfig`, wherein the deserializer wrapper is used to load the config. If it reports an "Unknown key" error, then the config is loaded again and deserialized without reporting warnings. This results in the warning be reported without preventing the otherwise valid config being loaded and returned.

#### Motivation and Context

Closes #4481 

In this screenshot, I'm showing that as I initialize starship with these changes, we can see I have not set a custom format (the line in the config is commented out), and the unexpected property "go" is detected and a warning is printed. After updating the config to enable the format using `sed`, the shell format is updated with the new prompt without resolving the unexpected config field.

image

1. set `~/.config/starship.toml` with an unexpected key like the following:
```toml
format = 'Hello! $character'

# This should be "golang".
[go]
disabled = true

[shell]
disabled = false
```
2. re-initialize starship using the build with `eval "$(cargo run init zsh)"` (I'm using `zsh`, but the shell type doesn't matter in this case)
3. Expect to see the prompt "Hello!" rather than the default shell name (in my case `zsh  ❯`

#### Checklist:

- [ ] I have updated the documentation accordingly. **(n/a)**

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -48,7 +48,17 @@ impl<'a, T: Deserialize<'a> + Default> ModuleConfig<'a, ValueError> for T {
     /// Create `ValueDeserializer` wrapper and use it to call `Deserialize::deserialize` on it.
     fn from_config(config: &'a Value) -> Result<Self, ValueError> {
         let deserializer = ValueDeserializer::new(config);
-        T::deserialize(deserializer)
+        T::deserialize(deserializer).or_else(|err| {
+            // If the error is an unrecognized key, print a warning and run
+            // deserialize ignoring that error. Otherwise, just return the error
+            if err.to_string().contains("Unknown key") {
+                log::warn!("{}", err);
+                let deserializer2 = ValueDeserializer::new(config).with_no_ignored_error();
+                T::deserialize(deserializer2)
+            } else {
+                Err(err)
+            }
+        })
     }
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -53,7 +53,7 @@ impl<'a, T: Deserialize<'a> + Default> ModuleConfig<'a, ValueError> for T {
             // deserialize ignoring that error. Otherwise, just return the error
             if err.to_string().contains("Unknown key") {
                 log::warn!("{}", err);
-                let deserializer2 = ValueDeserializer::new(config).with_no_ignored_error();
+                let deserializer2 = ValueDeserializer::new(config).with_allow_unknown_keys();
                 T::deserialize(deserializer2)
             } else {
                 Err(err)
--- a/src/config.rs
+++ b/src/config.rs
@@ -593,6 +593,24 @@ mod tests {
     }
 
     #[test]
+    fn test_load_unknown_key_config() {
+        #[derive(Clone, Default, Deserialize)]
+        #[serde(default)]
+        struct TestConfig<'a> {
+            pub foo: &'a str,
+        }
+
+        let config = toml::toml! {
+            foo = "test"
+            bar = "ignore me"
+        };
+        let rust_config = TestConfig::from_config(&config);
+
+        assert!(rust_config.is_ok());
+        assert_eq!(rust_config.unwrap().foo, "test");
+    }
+
+    #[test]
     fn test_from_string() {
         let config = Value::String(String::from("S"));
         assert_eq!(<&str>::from_config(&config).unwrap(), "S");
--- a/src/serde_utils.rs
+++ b/src/serde_utils.rs
@@ -47,7 +47,7 @@ impl<'de> ValueDeserializer<'de> {
         }
     }
 
-    pub fn with_no_ignored_error(self) -> Self {
+    pub fn with_allow_unknown_keys(self) -> Self {
         ValueDeserializer {
             error_on_ignored: false,
             ..self
@@ -354,7 +354,7 @@ mod test {
             "Error in 'StarshipRoot' at 'unknown_key': Unknown key"
         );
 
-        let deserializer2 = ValueDeserializer::new(&value).with_no_ignored_error();
+        let deserializer2 = ValueDeserializer::new(&value).with_allow_unknown_keys();
         let result = StarshipRootConfig::deserialize(deserializer2);
         assert!(result.is_ok());
     }
--- a/src/serde_utils.rs
+++ b/src/serde_utils.rs
@@ -13,6 +13,7 @@ pub struct ValueDeserializer<'de> {
     value: &'de Value,
     info: Option<StructInfo>,
     current_key: Option<&'de str>,
+    error_on_ignored: bool,
 }
 
 /// When deserializing a struct, this struct stores information about the struct.
@@ -28,14 +29,30 @@ impl<'de> ValueDeserializer<'de> {
             value,
             info: None,
             current_key: None,
+            error_on_ignored: true,
         }
     }
 
-    fn with_info(value: &'de Value, info: Option<StructInfo>, current_key: &'de str) -> Self {
+    fn with_info(
+        value: &'de Value,
+        info: Option<StructInfo>,
+        current_key: &'de str,
+        ignored: bool,
+    ) -> Self {
         ValueDeserializer {
             value,
             info,
             current_key: Some(current_key),
+            error_on_ignored: ignored,
+        }
+    }
+
+    pub fn with_no_ignored_error(self) -> Self {
+        ValueDeserializer {
+            value: self.value,
+            info: self.info,
+            current_key: self.current_key,
+            error_on_ignored: false,
         }
     }
 }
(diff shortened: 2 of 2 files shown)


davidkna:
It doesn't think we currently use `Either<A, B>` or any `enum`s like that, but I did it like that on purpose. For instance, if a config item has multiple variants, and if one of them fails to parse, this could print warnings; even if one of the items ends up being successful and doesn't have any extra keys.

brahmlower:
Ah okay, that's good foresight. Unless I've misunderstood though, I believe a scenario using `Either<A, B>` without my changes in this PR would still result in `deserialize_ignored_any` returning an `Err` if `B` was provided and `A` was checked first, meaning deserialization would still fail? I'm not sure how to rectify this. As I see it, we have a couple core facts:
- We shouldn't emit logs during deserialization
- We can't return warnings without using Err
- Warnings shouldn't cause deserialization to fail

So it would seem like we can't support the "Unknown key{did_you_mean}" feature without logging during serialization and without causing deserialization to fail.

One thought was we could give `ValueDeserializer` an extra config such that warnings do cause deserialization failures so that we could log them, and then we could just run it again without warnings as errors. That wouldn't solve the case you brought up though, because it would still potentially cause a warning to be emitted for an `Either` typed config.

Do you have any preferences or direction for this?

davidkna:
> Ah okay, that's good foresight. Unless I've misunderstood though, I believe a scenario using Either<A, B> without my changes in this PR would still result in deserialize_ignored_any returning an Err if B was provided and A was checked first, meaning deserialization would still fail? I

If I remember correctly, `serde(untagged)` works by trying each variant in turn, until the first deserialization succeeds (does not return an error).

> One thought was we could give ValueDeserializer an extra config such that warnings do cause deserialization failures so that we could log them, and then we could just run it again without warnings as errors. That wouldn't solve the case you brought up though, because it would still potentially cause a warning to be emitted for an Either typed config.

I guess that would be okay if it's simple to implement.

brahmlower:
> I guess that would be okay if it's simple to implement

Yeah I def agree on the "if it's simple part". The existing interface is really nice, and I don't want to undermine that 😅

I don't think I'll have a chance to get back to the computer this this evening, so I won't have any code updates for another 12 hours or so

brahmlower:
Alrighty! I've got the deserialization retry behavior implemented, so now we'll only print the warning when intended 👍

brahmlower:
Hey @davidkna, sorry to bother you with this but would you be comfortable adding the "hacktoberfest-accepted" label to this PR today? I don't at all want to rush review of these changes and I definitely intend to see this through to a state you're comfortable with, but this is my 4th and final PR for the event 🙏 Adding the label will give me credit without you having to approve or merge the changes in their current form

davidkna:
Sure.

brahmlower:
Awesome, thank you! I realized I still need to update the description to reflect the revised approach we came up with yesterday. I can take care of that this evening 👍

davidkna (changes requested):
I have made two suggestions, but apart from that this looks good to me.

src/serde_utils.rs
```diff
@@ -28,14 +29,30 @@ impl<'de> ValueDeserializer<'de> {
...
+        }
+    }
+
+    pub fn with_no_ignored_error(self) -> Self {
+        ValueDeserializer {
+            value: self.value,
+            info: self.info,
+            current_key: self.current_key,
+            error_on_ignored: false,
         }
```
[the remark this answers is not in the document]
brahmlower: Thanks for this suggestion! I wasn't aware of the struct update syntax, so I appreciate you teach me something new 😁

src/serde_utils.rs
```diff
@@ -28,14 +29,30 @@ impl<'de> ValueDeserializer<'de> {
...
+    ) -> Self {
         ValueDeserializer {
             value,
             info,
             current_key: Some(current_key),
+            error_on_ignored: ignored,
+        }
+    }
+
+    pub fn with_no_ignored_error(self) -> Self {
```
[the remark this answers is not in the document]
brahmlower: I definitely struggled to find a good name for this- iirc I flip flopped between a few names while working on it 😅 

How does `without_warnings` sound?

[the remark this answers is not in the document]
davidkna: How about `with_allow_ignored_keys()`, `with_allow_unknown_keys()` or `with_ignored_noerror()`?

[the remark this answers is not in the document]
brahmlower: ah yeah these are much better suggestions than mine 😅 Lets go with `with_allow_unknown_keys()`- I feel like that works best as part of the public interface for the struct whereas the other two lean more heavily on verbiage from the internal implementation 🤔

I'll make these changes and get them pushed in a minute

brahmlower:
Looks like that lint error is due to some clippy changes for rust 1.65.0, since that job is targeting stable

brahmlower:
Hmm, I tried fixing the lint errors but it turned into a bit of a slippery slope that seemed a little out of scope for this PR, so I've just rebased those commits out. 😬 I still have them locally if you'd like me to open a separate PR fixing the lint issues

andytom:
> Hmm, I tried fixing the lint errors but it turned into a bit of a slippery slope that seemed a little out of scope for this PR, so I've just rebased those commits out. grimacing I still have them locally if you'd like me to open a separate PR fixing the lint issues

I've merged the latest clippy fixes so if you rebase your branch that should (hopefully) fix them.

brahmlower:
Thanks for fixing those lint errors Thomas! With this branch rebased, we're all set for final review 👍

src/config.rs
```diff
@@ -48,7 +48,17 @@ impl<'a, T: Deserialize<'a> + Default> ModuleConfig<'a, ValueError> for T {
...
     fn from_config(config: &'a Value) -> Result<Self, ValueError> {
         let deserializer = ValueDeserializer::new(config);
-        T::deserialize(deserializer)
+        T::deserialize(deserializer).or_else(|err| {
+            // If the error is an unrecognized key, print a warning and run
+            // deserialize ignoring that error. Otherwise, just return the error
+            if err.to_string().contains("Unknown key") {
+                log::warn!("{}", err);
+                let deserializer2 = ValueDeserializer::new(config).with_allow_unknown_keys();
+                T::deserialize(deserializer2)
```
davidkna, line 57: Didn't you add or edit and existing test? Did it get lost during rebasing?

brahmlower: I just dug through my reflog to confirm, and it looks like I had only updated the test in `serde_utils.rs`. I thought the coverage reports showed this code branch was covered by that test change, but maybe I'm misremembering 🤔 I'll do some further testing/experimenting locally in the morning to see what's up

davidkna: In that case, would you mind adding a test that uses this path?

brahmlower: Hey! Sorry for the crazy delay here- I caught the flu a couple weeks ago and it ended up wiping me out pretty bad. I'm slowly getting back in the saddle here though 😅 I just wrapped up work today and will get this last test added this evening

brahmlower: Had to step away for a while while CI was running, but looks like that test has properly covered that block now 👍

davidkna (approved):

andytom:
Thank you for your contribution @brahmlower and thanks for reviewing @davidkna.

PR #1568, opened 2020-08-08
feat(kubernetes): Add styling based on current context

lht:
#### Description

Allow granular control of styles based on current Kubernetes context name. The config is inspired by the battery module. Multiple kubernetes.display option can be added with regular expressiones to match the current kubernetes context name. Style of the first matched display option will be used for the final style.

The display option can be extended to allow matching based on context aliases or namespace names, or allow changing the symbol.

#### Motivation and Context

Fixes #570

#### Screenshots (if appropriate):
starship-k8s-style

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #4550, opened 2022-10-30
feat(kubernetes): Add styling based on current context

jankatins:
#### Description
This PR implements the suggestion from https://github.com/starship/starship/issues/570 to add an ability to customize the configuration of the kubernetes module style, based on the current context.

A new variable is added to the config section, called environments, which is a list of possible customizations. Each such customization is an object with a context_pattern regex, which matches context name, and an optional style and icon that will override the global configuration, if the currently used context matched the context_pattern.

Based on multiple attempts to add per-context styling and symbols to the kubernetes module.

- https://github.com/starship/starship/pull/1568 by @lht  -> base (rebased and adjusted to current HEAD)
- https://github.com/starship/starship/pull/614 by @nomaed -> naming, symbol, some tests

Rebase, combined, and extended by @jankatins

#### Motivation and Context
It is common to work with development, testing, staging and production environments when deploying to kubernetes clusters, especially with continuous deployment. However, it is too easy to be using a production context (for example, for inspecting logs) and forgetting to switch to the personal development context when deploying any changes.

This change helps giving better visibility, by allowing to change the style of the kubernetes module in the prompt to draw attention, or changing the color to red (as an example) if using a context that should be used with care.

Closes: https://github.com/starship/starship/issues/570 (original request)
Closes: https://github.com/starship/starship/pull/1568 (superseded PR)
Closes: https://github.com/starship/starship/pull/614 (superseded PR)

image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2434,7 +2434,8 @@ kotlin_binary = 'kotlinc'
 Displays the current [Kubernetes context](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#context) name and, if set, the namespace, user and cluster from the kubeconfig file.
 The namespace needs to be set in the kubeconfig file, this can be done via
 `kubectl config set-context starship-context --namespace astronaut`.
-Similarly the user and cluster can be set with `kubectl config set-context starship-context --user starship-user` and `kubectl config set-context starship-context --cluster starship-cluster`.
+Similarly, the user and cluster can be set with `kubectl config set-context starship-context --user starship-user`
+and `kubectl config set-context starship-context --cluster starship-cluster`.
 If the `$KUBECONFIG` env var is set the module will use that if not it will use the `~/.kube/config`.
 
 ::: tip
@@ -2450,17 +2451,39 @@ case the module will only be active in directories that match those conditions.
 
 ### Options
 
+::: warning
+
+The `context_aliases` and `user_aliases` options are deprecated. Use `contexts` and the corresponding `context_alias`
+and `user_alias` options instead.
+
+:::
+
 | Option              | Default                                            | Description                                                           |
 | ------------------- | -------------------------------------------------- | --------------------------------------------------------------------- |
 | `symbol`            | `'☸ '`                                             | A format string representing the symbol displayed before the Cluster. |
 | `format`            | `'[$symbol$context( \($namespace\))]($style) in '` | The format for the module.                                            |
 | `style`             | `'cyan bold'`                                      | The style for the module.                                             |
-| `context_aliases`   | `{}`                                               | Table of context aliases to display.                                  |
-| `user_aliases`      | `{}`                                               | Table of user aliases to display.                                     |
+| `context_aliases`*  | `{}`                                               | Table of context aliases to display.                                  |
+| `user_aliases`*     | `{}`                                               | Table of user aliases to display.                                     |
 | `detect_extensions` | `[]`                                               | Which extensions should trigger this module.                          |
 | `detect_files`      | `[]`                                               | Which filenames should trigger this module.                           |
 | `detect_folders`    | `[]`                                               | Which folders should trigger this modules.                            |
+| `contexts`          | `[]`                                               | Customized styles and symbols for specific contexts.                  |
 | `disabled`          | `true`                                             | Disables the `kubernetes` module.                                     |
+
+*: This option is deprecated, please add `contexts` with the corresponding `context_alias` and `user_alias` options instead.
+
+To customize the style of the module for specific environments, use the following configuration as
+part of the `environments` list:
+
+| Variable          | Description                                                                              |
+| ----------------- | ---------------------------------------------------------------------------------------- |
+| `context_pattern` | **Required** Regular expression to match current Kubernetes context name.                |
+| `user_pattern`    | Regular expression to match current Kubernetes user name.                                |
+| `context_alias`   | Context alias to display instead of the full context name.                               |
+| `user_alias`      | User alias to display instead of the full user name.                                     |
+| `style`           | The style for the module when using this context. If not set, will use module's style.   |
+| `symbol`          | The symbol for the module when using this context. If not set, will use module's symbol. |
 
 ### Variables
 
@@ -2483,13 +2506,9 @@ case the module will only be active in directories that match those conditions.
 [kubernetes]
 format = 'on [⛵ ($user on )($cluster in )$context \($namespace\)](dimmed green) '
 disabled = false
-[kubernetes.context_aliases]
-'dev.local.cluster.k8s' = 'dev'
-'.*/openshift-cluster/.*' = 'openshift'
-'gke_.*_(?P<var_cluster>[\w-]+)' = 'gke-$var_cluster'
-[kubernetes.user_aliases]
-'dev.local.cluster.k8s' = 'dev'
-'root/.*' = 'root'
+contexts = [
+  { context_pattern = "dev.local.cluster.k8s", style = "green", symbol = "💔 " },
+]
 ```
 
 Only show the module in directories that contain a `k8s` file.
@@ -2502,29 +2521,28 @@ disabled = false
 detect_files = ['k8s']
 ```
 
-#### Regex Matching
+#### Kubernetes Context specific config
 
-Additional to simple aliasing, `context_aliases` and `user_aliases` also supports
-extended matching and renaming using regular expressions.
-
-The regular expression must match on the entire kube context,
-capture groups can be referenced using `$name` and `$N` in the replacement.
-This is more explained in the [regex crate](https://docs.rs/regex/1.5.4/regex/struct.Regex.html#method.replace) documentation.
-
-Long and automatically generated cluster names can be identified
-and shortened using regular expressions:
+The `contexts` configuration option is used to customise what the current Kubernetes context name looks
+like (style and symbol) if the name matches defined regular expression.
 
 ```toml
-[kubernetes.context_aliases]
-# OpenShift contexts carry the namespace and user in the kube context: `namespace/name/user`:
-'.*/openshift-cluster/.*' = 'openshift'
-# Or better, to rename every OpenShift cluster at once:
-'.*/(?P<var_cluster>[\w-]+)/.*' = '$var_cluster'
+# ~/.config/starship.toml
 
-# Contexts from GKE, AWS and other cloud providers usually carry additional information, like the region/zone.
-# The following entry matches on the GKE format (`gke_projectname_zone_cluster-name`)
-# and renames every matching kube context into a more readable format (`gke-cluster-name`):
-'gke_.*_(?P<var_cluster>[\w-]+)' = 'gke-$var_cluster'
+[[kubernetes.contexts]]
+# "bold red" style + default symbol when Kubernetes current context name contains "production"
+context_pattern = "production"
+user_pattern = "admin_user"
+style = "bold red"
+context_alias = "prod"
+user_alias = "admin"
+
+[[kubernetes.contexts]]
+# "green" style + a different symbol when Kubernetes current context name contains openshift
+context_pattern = ".*openshift.*"
+style = "green"
+symbol = "💔 "
+context_alias = "openshift"
 ```
 
 ## Line Break
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2474,7 +2474,7 @@ and `user_alias` options instead.
 *: This option is deprecated, please add `contexts` with the corresponding `context_alias` and `user_alias` options instead.
 
 To customize the style of the module for specific environments, use the following configuration as
-part of the `environments` list:
+part of the `contexts` list:
 
 | Variable          | Description                                                                              |
 | ----------------- | ---------------------------------------------------------------------------------------- |
@@ -2484,6 +2484,11 @@ part of the `environments` list:
 | `user_alias`      | User alias to display instead of the full user name.                                     |
 | `style`           | The style for the module when using this context. If not set, will use module's style.   |
 | `symbol`          | The symbol for the module when using this context. If not set, will use module's symbol. |
+
+Note that all regular expression are anchored with `^<pattern>$` and so must match the whole string. The `*_pattern`
+regular expressions may contain capture groups, which can be referenced in the corresponding alias via `$name` and `$N`
+(see example below and the
+[rust Regex::replace() documentation](https://docs.rs/regex/latest/regex/struct.Regex.html#method.replace)).
 
 ### Variables
 
@@ -2524,13 +2529,14 @@ detect_files = ['k8s']
 #### Kubernetes Context specific config
 
 The `contexts` configuration option is used to customise what the current Kubernetes context name looks
-like (style and symbol) if the name matches defined regular expression.
+like (style and symbol) if the name matches the defined regular expression.
 
 ```toml
 # ~/.config/starship.toml
 
 [[kubernetes.contexts]]
-# "bold red" style + default symbol when Kubernetes current context name contains "production"
+# "bold red" style + default symbol when Kubernetes current context name equals "production" *and* the current user
+# equals "admin_user"
 context_pattern = "production"
 user_pattern = "admin_user"
 style = "bold red"
@@ -2543,6 +2549,14 @@ context_pattern = ".*openshift.*"
 style = "green"
 symbol = "💔 "
 context_alias = "openshift"
+
+[[kubernetes.contexts]]
+# Using capture groups
+# Contexts from GKE, AWS and other cloud providers usually carry additional information, like the region/zone.
+# The following entry matches on the GKE format (`gke_projectname_zone_cluster-name`)
+# and renames every matching kube context into a more readable format (`gke-cluster-name`):
+context_pattern = "gke_.*_(?P<cluster>[\\w-]+)"
+context_alias = "gke-$cluster"
 ```
 
 ## Line Break
--- a/src/configs/kubernetes.rs
+++ b/src/configs/kubernetes.rs
@@ -18,6 +18,7 @@ pub struct KubernetesConfig<'a> {
     pub detect_extensions: Vec<&'a str>,
     pub detect_files: Vec<&'a str>,
     pub detect_folders: Vec<&'a str>,
+    pub contexts: Vec<KubernetesContextConfig<'a>>,
 }
 
 impl<'a> Default for KubernetesConfig<'a> {
@@ -32,6 +33,23 @@ impl<'a> Default for KubernetesConfig<'a> {
             detect_extensions: vec![],
             detect_files: vec![],
             detect_folders: vec![],
+            contexts: vec![],
         }
     }
+}
+
+#[derive(Clone, Deserialize, Serialize, Default)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+#[serde(default)]
+pub struct KubernetesContextConfig<'a> {
+    pub context_pattern: &'a str,
+    pub user_pattern: Option<&'a str>,
+    pub symbol: Option<&'a str>,
+    pub style: Option<&'a str>,
+    pub context_alias: Option<&'a str>,
+    pub user_alias: Option<&'a str>,
 }
--- a/src/modules/kubernetes.rs
+++ b/src/modules/kubernetes.rs
@@ -697,7 +697,9 @@ users: []
     }
 
     #[test]
-    fn test_multiple_config_files_with_ns() -> io::Result<()> {
+    fn test_multiple_config_files_with_context_defined_once() -> io::Result<()> {
+        // test that we get the current context from the first config file in the KUBECONFIG,
+        // no matter if it is only defined in the latter
         let dir = tempfile::tempdir()?;
 
         let filename_cc = dir.path().join("config_cc");
@@ -750,7 +752,7 @@ users: []
             })
             .collect();
 
-        // And tes with context and namespace first
+        // And test with context and namespace first
         let actual_ctx_first = ModuleRenderer::new("kubernetes")
             .path(dir.path())
             .env(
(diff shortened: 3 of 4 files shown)


src/modules/kubernetes.rs
```diff
@@ -955,4 +981,135 @@ users: []
...
+
+        let actual = ModuleRenderer::new("kubernetes")
+            .path(dir.path())
+            .env("KUBECONFIG", filename.to_string_lossy().as_ref())
+            .config(toml::toml! {
+                [kubernetes]
+                disabled = false
+                style = "bold red"
+                environments = [
+                    {context_pattern = "tests_(.*", style = "bold green", symbol = "§ "},
```
jankatins, line 1103: This misses a test for the warning, but my rust (beginner...) and google knowledge didn't find a possibility without adding another crate.

src/modules/kubernetes.rs
```diff
@@ -955,4 +981,135 @@ users: []
         assert_eq!(expected, actual);
         dir.close()
     }
+
```
jankatins, line 984: https://github.com/starship/starship/pull/614 contains more tests, I tried to port the most relevant over. Not sure if I should port more?

jankatins:
I guess https://github.com/starship/starship/pull/4505/files should be reworked to add a 'ignore' field to an environment? One could argue that the context alias should also be similarly be reworked as a per alias defined in the environment?

CC: @DanielKneipp @davidkna

jankatins:
The [windows failure](https://github.com/starship/starship/actions/runs/3357644660/jobs/5563604803) seems to be in an unrelated test:

```
failures:
    modules::git_status::tests::doesnt_show_ignored_file
```

docs/config/README.md
```diff
@@ -2236,6 +2247,9 @@ case the module will only be active in directories that match those conditions.
 [kubernetes]
 format = 'on [⛵ ($user on )($cluster in )$context \($namespace\)](dimmed green) '
 disabled = false
+environments = [
```
jankatins, line 2250: Alternative would be "contexts" (and renaming `context_pattern` to `pattern`)

jankatins:
Rebased on top of current master

jankatins:
Clippy failures seem unrelated, as the problems are in other modules.

jankatins:
Rebased and, wohoo, green builds!

src/configs/kubernetes.rs
```diff
@@ -32,6 +33,20 @@ impl<'a> Default for KubernetesConfig<'a> {
             detect_extensions: vec![],
             detect_files: vec![],
             detect_folders: vec![],
+            environments: vec![],
         }
     }
 }
+
+#[derive(Clone, Deserialize, Serialize, Default)]
```
chipbuster, line 41: I am not as familiar with schemars/serde as I'd like to be, so I'd like a quick check from another maintainer on this part to make sure the macros on this struct look correct.

src/modules/kubernetes.rs
```diff
@@ -955,4 +981,135 @@ users: []
...
+    #[test]
```
chipbuster, line 985: This set of tests looks good for the change. A number of the tests from #614 seem to be covered by other existing tests at this point.

chipbuster (commented):
The current PR looks good to me, though we should work out what we want to do with naming for the config options before we merge this, since that will be difficult to change later on.

I would also like another reviewer to check the macros on the new `KubernetesEvnironmentConfig` struct, since I'm not as familiar with schemars as I'd like to be.

As far as the proposed changes w.r.t putting alises into the environment config itself,  we can either include it in this PR, or merge this (after resolving naming) and have a new PR that includes that functionality. I will leave it to @jankatins to decide which they prefer.

docs/config/README.md
chipbuster: @jankatins I don't have lots of experience with Kubernetes, but some searching suggests that "contexts" is the less ambiguous (and more idiomatic) term for Kubernetes. Does that square with your experience? If so, I would vote for going with contexts and rename the pattern variable. The term "environment" is overloaded enough as it is.

Would also be interested in @nomaed's opinion on this.

As far as the aliases go, I think that sounds like a good idea (but again, I don't know much about Kubernetes). It seems to me that it basically eliminates the need for multiple sets regexes, which are a little fiddly to write. In the worst case, we can deprecate the user_alias and context_alias options now, then start emitting warnings for anyone using them in a future version, before finally getting rid of them for 2.0.

src/configs/kubernetes.rs
davidkna: The macros look correct to me.

docs/config/README.md
chipbuster: #4505 also seems to suggest that "contexts" is the preferred term.

jankatins: So:
-Rename environment to context
- Add user_pattern, user_alias, context_alias (given that we have user and context pattern, I would keep the `context_` prefix to `context_pattern` -> user_pattern would only be evaluated if set.
- Maker that the default way to set user and context aliases: if not set via the matched context
- look at the old way and omit a warning
- If can, also port the "set alias from regex group" feature.

jankatins:
- [x] Renamed "environments" to "contexts" and added the other three config items
- [x] Added context matching based on user_pattern (default: no matching happens, context_pattern alone is responsible for matching a context)
- [x] Added context_alias and user_alias (and deprecated the old way by removing the docs for it :-)

This has one regression: before it was possible to define a match group in the regex and then use that in the alias. I haven't implemented that.

docs/config/README.md
jankatins: Done now, apart from the regex group feature

jankatins:
Rebased... From my perspective it's ready for review again...

src/modules/kubernetes.rs
```diff
@@ -161,38 +198,97 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         None => false,
     });
 
+    // Parse config under `display`.
+    // Select the first style that matches the context_pattern and,
+    // if it is defined, the user_pattern
+    let matched_context_config = config.contexts.iter().find(|context_config| {
```
chipbuster, line 204: I'm find the computation of this variable's value hard to read. I don't know if the fact that I'm a little tired right now affects it, but there's a lot of error handling, logging, short-circuiting, and nested things mixed in here.

I will revisit this tomorrow and see if I still have problems with it (and if so, hopefully have some suggestions).

chipbuster (commented):
I'm not 100% happy with the body of the module function right now, but I'm also pretty fatigued. Let me get some sleep and get back to you on that.

Everything else looks alright to me, except for the loss of capture groups in the regexes. Is that not implemented just because of additional complexity, or is there some technical blocker that would prevent us from implementing that in the future?

I'm wary of deprecating existing functionality without having a suitable replacement or workaround ready, so if we aren't 100% certain that this can be written without too much hassle, I'm going to be a little nervous about merging this.

src/modules/kubernetes.rs
jankatins: Getting a better codestyle here (and also below with all the maps there) would be awesome. I'm a rust beginner, so in some places, I basically changed the code until it compiled and worked :-(

src/modules/kubernetes.rs
```diff
@@ -161,38 +198,97 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-                    })
-                }),
                 "cluster" => kube_cluster.and_then(|ctx| {
                     ctx.as_ref().map(|kube| {
                         // unwrap is safe as kube_cluster only holds kube.cluster.is_some()
                         Ok(Cow::Borrowed(kube.cluster.as_ref().unwrap().as_str()))
                     })
                 }),
-
+                "user" => display_user.as_ref().map(|user| Ok(user.clone())),
```
jankatins, line 291: I still have no idea why I need clone here, but that was the only way I could get this whole lines to compile :-(

jankatins:
> I'm wary of deprecating existing functionality without having a suitable replacement or workaround ready, so if we aren't 100% certain that this can be written without too much hassle, I'm going to be a little nervous about merging this.

So right now my problem with this feature is that it now would need two times running the regex: ones during the map which needs to return the bool to select an entry and the second time to get a match entry to fill out the context alias. Not sure if that is a performance penalty or not. If not, It's simply rerunning the regex.

src/modules/kubernetes.rs
chipbuster: With a little bit of fiddling, I've been able to get this down to this:

```
       let is_context_match = match_regex(&kube_ctx, &context_config.context_pattern);
        if !is_context_match {
            return false;
        };
        if let Some(user_pattern) = context_config.user_pattern {
            if let Some(user) = kube_user.as_ref() {
                return match_regex(user, user_pattern);
            } else {
                // If there is a user pattern but no user, the pattern cannot match
                return false;
            }
        } else {
            // If there is no user pattern, then this context config matches
            return true;
        }
```

with the following definition of `match_regex`:
```
fn match_regex(target: &str, pattern: &str) -> bool {
    match regex::Regex::new(pattern) {
        Ok(re) => re.is_match(&target),
        Err(error) => {
            log::warn!("Could not compile regex pattern `{}`:\n{}", pattern, error);
            false
        }
    }
}
```

I'm still not entirely happy with this, because:

1. The `match_regex` name is terrible.
2. I think the body of the closure in `find` can still be cleaned up a little bit
3. I think we can modify match_regex to do capture group parsing and output in a single pass instead of having to do two passes

but I think it's a starting point.

chipbuster: One downside here is that the error messages won't contain the context of whether it's a context-regex or a user-regex, but I doubt most users will have the same regex for both (and if they do, they both need to be fixed).

src/modules/kubernetes.rs
```diff
@@ -148,10 +176,19 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         .map(|filename| get_kube_ctx_component(filename, kube_ctx.clone()))
```
chipbuster, line 176: Something I just realized while poking around this module: it looks like in all subsequent uses of the `ctx_components` variable, we iterate over the components, check if they're `Some`, and then check if one of their components is a `Some`.

Why don't we just filter the vector here to only contain non-`None` components?

jankatins: you mean add a `.filter_map(|x| x)` here to filter out the `None` and declare it as `Vec<KubeCtxComponents>`?

chipbuster: On further reflection, I think we can change this `map` to `filter_map` and get the same result.

chipbuster:
@jankatins I might have a few spare hours this week to play around with this. Would you be okay with me applying some of the changes I've suggested here?

jankatins:
@chipbuster Please go ahead: I'm looking forward to see proper rust code and not my trial and error way of coding :⁠-⁠)

chipbuster:
@jankatins I've made some of the changes we discussed. Among the changes:

- Move the alias functions to a `deprecated` submodule, instead of changing their names.
- Filter out all `None` ContextComponents to greatly simplify processing of components
- Create a regex-matching function to simplify a lot of matching logic
- Make Clippy upset (whoops, should fix that!)

However, I'm not really a Kubernetes person, and I still don't understand a few of the things that are happening here.In particular, we currently

chipbuster:
@jankatins On further reflection, I think the nesting of options in the module makes it more confusing than it needs to be (it's probably also why I had trouble reading the PR initially and why you had to do a bit of compiler-driven type-tetris to get this thing building.

The changes I made in the previous commit do help this somewhat, but it also bloats the PR. I've made a separate PR with my cleanup changes (#4743) and assuming that it gets merged, I'll rebase your changes on top of that and we can go from there.

src/modules/kubernetes.rs
```diff
@@ -144,55 +128,87 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let kube_ctx = env::split_paths(&kube_cfg).find_map(get_kube_context)?;
 
-    let ctx_components: Vec<Option<KubeCtxComponents>> = env::split_paths(&kube_cfg)
-        .map(|filename| get_kube_ctx_component(filename, kube_ctx.clone()))
+    let ctx_components: Vec<KubeCtxComponents> = env::split_paths(&kube_cfg)
```
jankatins, line 131: The more I stare at this, the more this feels wrong: IMO this should end up in a `<Option<KubeCtxComponents>`  -> either the context is found in the contexts or not:

```yaml
current-context: work
contexts:
- context:
    cluster: work
    namespace: work-dev
    user: clusterUser
  name: 50hzk8s
- context:
    cluster: docker-desktop
    user: docker-desktop
  name: docker-desktop
```

i.e in `get_kube_ctx_component` we filter for the name, so there will at most be one KubeCtxComponents coming out of this, why should this become a vector...?

```rust
fn get_kube_ctx_component(filename: path::PathBuf, current_ctx: &str) -> Option<KubeCtxComponents> {
...
    let ctx_yaml = conf["contexts"].as_vec().and_then(|contexts| {
        contexts
            .iter()
            .filter_map(|ctx| Some((ctx, ctx["name"].as_str()?)))
            .find(|(_, name)| *name == current_ctx)  // <- HERE
    });

    let ctx_components = KubeCtxComponents {
...
    };

    Some(ctx_components)
}
```

And if it is None, the whole module should be aborted, IMO...

jankatins: E.g. this seems to have introduced the vec: https://github.com/starship/starship/pull/3569/files#diff-893760ffe78f91e7934d6da73675dbc83b63e62bbce7341f70310b0ab1187442L89

jankatins:
I can do the rebase, but I would probably squash all commits down and do it then as the rebase of all commits looks ... painful.

I'm not sure if I can get to it before xmas, though. Might be that I have some time tomorrow in the train, but otherwise only after...

jankatins:
@chipbuster I squash-rebased (added all authors as coauthors) and added a commit which removes some options, switches to `<Option<KubeCtxComponents>`, and renames some functions to make it more clearer what is actually going on (thats at least what I though...). 

This still doesn't add the match group feature back :-(

src/modules/kubernetes.rs
davidkna:
```suggestion
                "user" => display_user.as_deref().map(Cow::Borrowed).map(Ok),
```

davidkna:
I think some of the matches in this module can be simplified via [`let` `else` statements](https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#let-else-statements).

src/modules/kubernetes.rs
chipbuster: > in get_kube_ctx_component we filter for the name, so there will at most be one KubeCtxComponents coming out of this, why should this become a vector...?

The only condition I can think of that would cause this is if there are multiple Kubernetes configs that define the same context. If that isn't just a hard error, then you need some way to resolve conflicting settings.

Is this allowed by Kubernetes at all? If so, what's the correct behavior?

chipbuster: I think I understand now. The logic in using a vector of components is to implicitly (throughout the rest of the module) implement the merging logic detailed at https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#merging-kubeconfig-files

I'm not sure I'm a fan of having things spread out like this. Perhaps the move is to put the logic for implementing the merge procedure into `get_kube_ctx_components`? Right now, I think if you have incomplete context components e.g. one config that only defines a user, and another config that only defined the namespace, the module will incorrectly only pick up on the first one defined.

Also surprisingly, it seems we don't actually have a test case for config merging.

jankatins: I actually think it's now correct: 

> Never change the value or map key. Example: Preserve the context of the first file to set current-context. Example: If two files specify a red-user, use only values from the first file's red-user. Even if the second file has non-conflicting entries under red-user, discard them.

My reading is, that the last sentence implies that the first context found is used for everything, no matter if another config defines "non-conflicting" items (eg. your example: "one config that only defines a user, and another config that only defined the namespace").

chipbuster: Huh, I guess that's right. (And we did get [a comment](https://github.com/starship/starship/pull/3569#issuecomment-1102130317) that it wasn't working properly, though I don't really know what the issue was).

Is there a way we can add a test for this?

jankatins: Ok, just verified this behaviour: if there is already a context with that name, only the config from that file is used and not "added" or overwritten from a second file: both ` one.yml` and `two.yml` contain a context with `name: work`, but the values only come from the context which was defined first:

```
λ  KUBECONFIG=$(pwd)/one.yml:$(pwd)/two.yml kubectl config view # no "namespace" in one.yml, but in two.yml
apiVersion: v1
clusters: null
contexts:
- context:
    cluster: docker-desktop
    user: docker-desktop
  name: docker-desktop
- context:
    cluster: work-cluster1
    user: clusterUser1
  name: work
current-context: work
kind: Config
preferences: {}
users: null

λ  KUBECONFIG=$(pwd)/two.yml:$(pwd)/one.yml kubectl config view # no user in two.yml -> gets set to an empty string!
apiVersion: v1
clusters: null
contexts:
- context:
    cluster: docker-desktop
    user: docker-desktop
  name: docker-desktop
- context:
    cluster: work-cluster2
    namespace: work-dev2
    user: ""
  name: work
current-context: work
kind: Config
preferences: {}
users: null
```

jankatins:
> I think some of the matches in this module can be simplified via [let else statements](https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#let-else-statements).

@davidkna I tried to find some places where the pattern mapped, but in all the places I looked, we do not simple "unpack" a option, but do something to it, too. From my (beginner) understanding, I couldn't make it work with `let else` statements. Any chance you have an example how it would works?

Sorry :-(

davidkna:
@jankatins I agree, on second look there aren't as many opportunities as I thought. Because the `else` branches often don't return or look at the error inside the result. There is some opportunity to improve in the deprecated functions, but that might not be worth the effort to refactor, i.e.:
```rust
    pub fn get_kube_user_alias<'a>(
        config: &'a KubernetesConfig,
        kube_user: &'a str,
    ) -> Cow<'a, str> {
        let Some(alias) = get_alias(&config.user_aliases, kube_user) else {
            return Cow::Borrowed(kube_user)
        }
        log::warn!("Usage of 'user_aliases' is deprecated and will be removed in 2.0; Use 'contexts' with 'user_alias' instead. (`{kube_user}` -> `{alias}`)");
        alias
    }

**Comment by jankatins** [comment]
> Edit: If pulled into a separate function, the context_config.user_pattern match could also benefit.

Fun, this worked even without a new function as it was in a closure...

Never looked at that part :-) Thanks, I learned something new :-)

**Comment by jankatins** [comment]
I added back the "use capture groups in alias" feature. for my taste I had to use a lot of `as_str/to_string` due to me and clippy having long conversations until it compiled and passed the tests...

@chipbuster @davidkna In my opinion it's ready for review again...

File: src/modules/kubernetes.rs [hunk 10]
```diff
@@ -80,30 +86,44 @@ fn get_kube_ctx_component(filename: path::PathBuf, current_ctx: &str) -> Option<
...
-        match replaced {
-            Cow::Owned(replaced) => Some(Cow::Owned(replaced)),
-            _ => None,
+    let re = match regex::Regex::new(&format!("^{pattern}$")) {
+        Ok(re) => re,
+        Err(error) => {
+            log::warn!(
+                "Could not compile regular expression `{}`:\n{}",
+                &format!("^{pattern}$"),
+                error
```
jankatins, line 116: This woul have been a nice let-else example, but I haven'T found a way to access the error (to print) in that form :-(

jankatins:
@chipbuster @davidkna Rebased on latest master. Any chance to get a review to keep this going?

src/modules/kubernetes.rs
```diff
@@ -80,30 +86,44 @@ fn get_kube_ctx_component(filename: path::PathBuf, current_ctx: &str) -> Option<
...
-    return get_alias(&config.context_aliases, kube_ctx).unwrap_or(Cow::Borrowed(kube_ctx));
-}
-
-fn get_alias<'a>(
-    aliases: &'a HashMap<String, &'a str>,
-    alias_candidate: &'a str,
-) -> Option<Cow<'a, str>> {
-    if let Some(val) = aliases.get(alias_candidate) {
-        return Some(Cow::Borrowed(val));
+fn get_aliased_name<'a>(
```
chipbuster, line 89: I suspect this is partially because I'm tired, but I've had a hell of a time trying to parse how this function is used. I've been staring at this and its call sites for almost 20 minutes now and there are still things I don't understand:

- Why do we take two `Option<&str>` and then an `Option<&String>`?
- What do these strings represent? (Every time I think I have it, one of them falls out of my head and I have to start thinking about it again).
- Why does the one of the call sites have a `Some(Some(something, _, _))` followed immediately by an `unwrap_or_else`?

I'm afraid I don't have a concrete suggestion right now (maybe will have one this weekend) but my gut is that this is too complex to merge as it is. Perhaps there's no better way to write it, in which case I'll grumble a little and approve, but I suspect there's a way to make this simpler than it currently is.

jankatins: I think it's more likely due to my inexperience with rust :-/

> Why do we take two Option<&str> and then an Option<&String>?

Mostly because the inputs were this and this version compiled :-/ I still haven't completely wrapped my head around when to use &str vs String vs &String...

> What do these strings represent? (Every time I think I have it, one of them falls out of my head and I have to start thinking about it again).

I think of this function as a wrapper around a regex search and replace: have a pattern, have a current value, have a replacement. The alias is the replacement. Only difference: if the pattern doesn't match, return None instead of current value.

> Why does the one of the call sites have a Some(Some(something, _, _)) followed immediately by an unwrap_or_else?

Again, probably because I got it compiled like that when using the filter_map (if you mean line 201) :-(

Just to clarify, the part around line 201:

```
   let (matched_context_config, display_context, display_user) = config.contexts.iter().find_map(|context_config| {
        let Some(context_alias) = get_aliased_name(Some(context_config.context_pattern), Some(&current_kube_ctx_name), context_config.context_alias) else {
            return None;
        };

        let user_alias = ... 
...
        Some((Some(context_config), context_alias, user_alias))
    }).unwrap_or_else(|| (None, current_kube_ctx_name.clone(), ctx_components.user));

```

So `Some(Some(something, _, _))` is actually `Some(Some(something), _, _)`,, so `Option<Tuple<Option<..>,..>> which hopefully is a bit more sane than an `Option<Option<Tuple>>`

>  but I suspect there's a way to make this simpler than it currently is.

I sure hope, but I'm currently out of my wits/experience for anything to make this easier. :-(

jankatins: Ok, came across this: https://rust-unofficial.github.io/patterns/idioms/coercion-arguments.html which argues that you should always go with `&str` instead of `&String`. Will make the adjustments after I understood that chapter...

jankatins: Did some refinement on the function... Hope it's clearer now. Also got rid of the `&String`

jankatins:
I cleared up the alias function and renamed it to `replace_with_alias_if_pattern_matches()`. Hopefully this function is now clearer and a bit easier to understand.

jankatins:
Rebased and force pushed on top of current master. 

@chipbuster I'm out of ideas what to improve here after the changes I did to the function which irritated you (see last commit in the PR). Any chance to get a review from you?

jankatins:
@davidkna @chipbuster Rebased and resolved the conflict. any chance to get this reviewed?

davidkna:
I've made an attempt to reduce the complexity a bit more, please have a look @jankatins: https://github.com/davidkna/starship/commit/ccfff3243a65dc1503b7bce69d1a7f20f90d2667

jankatins:
@davidkna Wow, thank you, that commit was a great learning experience! Especially the pattern `let value = current_value?` was totally unknown to me.


I pulled it into my branch, added a `refactor(kubernetes):` prefix and pushed it. 

The windows test failure seems unrelated:

```
Error: Custom { kind: Uncategorized, error: PathError { path: "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\.tmpTT07Ia", err: Os { code: 32, kind: Uncategorized, message: "The process cannot access the file because it is being used by another process." } } }


failures:
    modules::git_status::tests::doesnt_show_ignored_file

test result: FAILED. 994 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 48.27s
```

jankatins:
Rebased and force pushed -> All tests are green

jankatins:
Rebased and force pushed and all tests green: @davidkna @chipbuster Any chance to get this reviewed and over the finish line?

jankatins:
@davidkna @chipbuster any chance to get this over the finish line?

felipesere:
I would love to have this. @davidkna @chipbuster any way I can help get this merged? 💜

jankatins:
@davidkna @chipbuster @matchai Any chance to get a review of the latest review fixes?

chipbuster (approved):
:+1: from me, this is looking great!

I'm very happy with this new interface for specifying context-specific styling. Assuming we don't encounter issues with it once it's out in the wild, we might consider porting this into other modules where people have been requesting changes, e.g. in some of the git modules.

chipbuster:
And it looks like all our other contributors are listed as coauthors in at least one commit, so this PR should be ready to go.

@jankatins Thank you so much for the patience to stick it out on this PR!

andytom:
Thank you for your contribution @jankatins and thank you reviewing @chipbuster.

PR #4554, opened 2022-11-02
feat(sudo): use new -N option if available

### LINKED ISSUE #4538: Sudo module without updating credential cache

The sudo module is updating the credential cache time as it is using the `sudo -nv` command.
This is forcing us to kill the shell to get out of sudo as any new command on the shell will result in resetting the cache.

Sudo 1.9.12 has introduced the -N option: https://www.sudo.ws/posts/2022/10/running-sudo-without-updating-cached-credentials/

So if the module is running `sudo -Nnv` then it will check the cache without resetting it.

A drawback is that we may need to check the version of sudo before adding the -N option.

None.

regit:
This implements the concept described in https://github.com/starship/starship/issues/4538. It improves the sudo module by not updating the cache when sudo is not used.

#### Description

Patch detect if the -N option is available in sudo then update command to avoid cache update.

Closes #4538

Tested with sudo 1.1.12.

#### Checklist:

### DIFF
--- a/src/modules/sudo.rs
+++ b/src/modules/sudo.rs
@@ -18,7 +18,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let is_sudo_cached = context.exec_cmd("sudo", &["-n", "true"]).is_some();
+    let mut sudo_option = "-nv";
+
+    /* let's check if -N option is supported */
+    let is_sudo_nocreds_update = context.exec_cmd("sudo", &["-NV"]).is_some();
+    if is_sudo_nocreds_update {
+        sudo_option = "-Nnv"
+    }
+
+    let is_sudo_cached = context.exec_cmd("sudo", &[sudo_option]).is_some();
 
     if !is_sudo_cached {
         return None;


src/modules/sudo.rs
```diff
@@ -18,7 +18,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let is_sudo_cached = context.exec_cmd("sudo", &["-n", "true"]).is_some();
+    let mut sudo_option = "-nv";
+
+    /* let's check if -N option is supported */
+    let is_sudo_nocreds_update = context.exec_cmd("sudo", &["-NV"]).is_some();
```
jankatins, line 24: wouldn't this run sudo twice on each run of starship?

davidkna: Maybe try inspecting the error output?

chipbuster: In principle I think we could avoid calling sudo twice on the optimistic path by using `-Nnv`, then falling back to `-nv` only if the first command fails, but I don't see a way to avoid calling it twice in general unless we cache the observation that the system sudo is an old version.

chipbuster: I've done some testing on this. As far as I can tell, it is never faster to get the version from sudo than it is to have it barf on an invalid flag. We would additionally need to parse the output (which would be fast, but adds a small amount of additional work) to decide what to do with that output.

I haven't observed a system where parsing the version output would be significantly faster than what this is currently doing. As for the solution I proposed, it could save ~8ms on a ~30ms operation, but only when the sudo version is >= 1.9.12 **and** the user already has cached credentials. In all other cases, the performance would be roughly the same (and the 8ms savings is based on the best-case scenario over all the systems I tried. In most cases, it would be more like 3ms on a 20ms operation).

Given these results, I'm willing to say that the performance issues here are probably fine, especially given that on most systems I've tested, the cost of a `sudo -NV` with a old sudo is somewhere between 0.5-2ms.

src/modules/sudo.rs
```diff
@@ -18,7 +18,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let is_sudo_cached = context.exec_cmd("sudo", &["-n", "true"]).is_some();
+    let mut sudo_option = "-nv";
```
chipbuster, line 21: We don't need to have a `mut` here. Instead, you could do something like

```
let has_N_option = context.exec_cmd("sudo", &["-NV"]).is_some();
let sudo_option = if has_N_option {
  "-Nnv"
} else {
  "-nv"
}
```

You could even fold the `has_N_option` directly into the `if` statement, though I'm usually a little bit wary about conditionals with potential side effects.

chipbuster (commented):
We can get rid of the mutable-then-change of `sudo_option` by directly evaluating the command in the conditional and doing the assignment. Aside from that, I think this PR is good to go.

EDIT: And we would need to fix the current crop of test failures, which would definitely be a merge blocker.

jorhett:
This remains an issue, what is currently blocking integration?

PR #4588, opened 2022-11-10
build(deps): upgrade to git-repository v0.27

Byron:
Please note that the only fix was for obtaining the name of a remote, which is now
strongly typed.

Even though it works as before, it's now made obvious that this remote may also be
a URL which might be relevant to `starship`.


andytom (approved):
LGTM

andytom:
Thank you for your help @Byron.

PR #4593, opened 2022-11-12
fix(container): avoid detecting WSL as a systemd-container

### LINKED ISSUE #4555: WSL shows ⬢ [Systemd] prompt
#### Current Behavior
On WSL with zsh, the prompt constantly shows `⬢ [Systemd]`.

#### Expected Behavior
Don't show that.

Seems similar to, but not the same as https://github.com/starship/starship/issues/3821.

#### Possible Solution
Assuming this is trying to show that systemd is in error, it is running in WSL but perhaps the liveliness is different.

#### Environment
- Starship version: 1.11.0
- Shell type: zsh
- Shell version: 5.8
- Shell plugin manager: None
- Operating system: Windows 11

#### Starship Configuration
Empty.

davidkna:
#### Description

WSL with systemd triggers the container module, because it creates a `/run/systemd/container` file. For WSL with systemd this file contains `wsl`, so this PR disables container detection if the file contains that.
I also included some test-refactorings.

#### Motivation and Context

Closes #4555

#### Checklist:

### DIFF
--- a/src/modules/container.rs
+++ b/src/modules/container.rs
@@ -10,7 +10,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     use super::ModuleConfig;
     use crate::configs::container::ContainerConfig;
     use crate::formatter::StringFormatter;
-    use crate::utils::read_file;
+    use crate::utils::{self, read_file};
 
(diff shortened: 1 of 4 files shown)


chipbuster (approved):
LGTM!

PR #4594, opened 2022-11-12
feat(loadavg): add module for load average

cgzones:
Example configuration:

    [loadavg]
    disabled = false

    [[loadavg.display]]
    threshold_one = 8.0
    style = "bold red"

    [[loadavg.display]]
    threshold_one = 4.0
    style = "bold yellow"

    [[loadavg.display]]
    threshold_one = 2.0
    style = "bold white"

    [[loadavg.display]]
    threshold_one = 0.0
    style = "dimmed white"

#### Description
Add a module to show the system load average.

#### Motivation and Context

Inspired by https://github.com/starship/starship/discussions/1252#discussioncomment-347145

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -912,6 +912,27 @@
         }
       ]
     },
+    "loadavg": {
+      "default": {
+        "disabled": true,
+        "display": [
+          {
+            "style": "white bold",
+            "symbol": null,
+            "threshold_fifteen": null,
+            "threshold_five": null,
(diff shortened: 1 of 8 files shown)


davidkna:
I think this should support Windows before merging.

cgzones:
What would be a good replacement metric on Windows?

cgzones:
Since the module is disabled by default, is it necessary to support Windows? See for example the sudo module.

PR #4604, opened 2022-11-14
fix(aws): enable when using .aws/credentials

### LINKED ISSUE #4603: AWS module not displayed when credential_process is defined in `.aws/credentials`
#### Current Behavior

I define my AWS `credential_process` in the `~/.aws/credentials` file, for each profile. The AWS CLI and SDK both support this behaviour.

When the `credential_process` is set there, the AWS module isn't enabled.

#### Expected Behavior

When I have a complete AWS configuration and an AWS profile is set in my environment, the `aws` module should appear in my prompt.

#### Possible Solution
I will be submitting a PR for this in a moment or two.

#### Environment
- Starship version: 1.11.0
- fish version: fish, version 3.5.0
- Operating system: Mac OS 13.0.1
- Terminal emulator: iTerm.app 3.5.0beta7
- Git Commit Hash: 
- Branch/Tag: master
- Rust Version: rustc 1.64.0 (a55dd71d5 2022-09-19)
- Build Time: 2022-10-15 01:46:07 +00:00

```fish
if status is-interactive
end

# Setting PATH for Python 3.10
# The original version is saved in /Users/offby1/.config/fish/config.fish.pysave
set -x PATH "/Library/Frameworks/Python.framework/Versions/3.10/bin" "$PATH"
```

#### Starship Configuration

```toml
add_newline = true
format = """
$username\
$hostname\
$shlvl\
$kubernetes\
$directory\
$git_branch\
$git_commit\
$git_state\
$git_status\
$hg_branch\
$docker_context\
$package\
$cmake\
$dart\
$dotnet\
$elixir\
$elm\
$erlang\
$golang\
$helm\
$java\
$julia\
$kotlin\
$nim\
$nodejs\
$ocaml\
$perl\
$php\
$purescript\
$python\
$ruby\
$rust\
$scala\
$swift\
$terraform\
$vagrant\
$zi

### PR DESCRIPTION
**Pull request by offbyone**
# fix(aws): enable when using .aws/credentials

-------

#### Description

Check in the AWS credentials file for the `credential_process` in order to enable
the `aws` module, not just in the `.aws/config` file (and all related configured 
locations for same.)

#### Motivation and Context

Closes #4603

#### Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -63,7 +63,7 @@ fn get_creds<'a>(context: &Context, config: &'a OnceCell<Option<Ini>>) -> Option
 // Get the section for a given profile name in the config file.
 fn get_profile_config<'a>(
     config: &'a Ini,
-    profile: &Option<Profile>,
+    profile: Option<&Profile>,
 ) -> Option<&'a ini::Properties> {
     match profile {
         Some(profile) => config.section(Some(format!("profile {}", profile))),
@@ -74,11 +74,11 @@ fn get_profile_config<'a>(
 // Get the section for a given profile name in the credentials file.
 fn get_profile_creds<'a>(
     config: &'a Ini,
-    profile: &Option<Profile>,
+    profile: Option<&Profile>,
 ) -> Option<&'a ini::Properties> {
     match profile {
         None => config.section(Some("default")),
-        _ => config.section(profile.as_ref()),
+        _ => config.section(profile),
     }
 }
 
@@ -88,7 +88,7 @@ fn get_aws_region_from_config(
     aws_config: &AwsConfigFile,
 ) -> Option<Region> {
     let config = get_config(context, aws_config)?;
-    let section = get_profile_config(config, aws_profile)?;
+    let section = get_profile_config(config, aws_profile.as_ref())?;
 
     section.get("region").map(std::borrow::ToOwned::to_owned)
 }
@@ -118,7 +118,7 @@ fn get_aws_profile_and_region(
 
 fn get_credentials_duration(
     context: &Context,
-    aws_profile: &Option<String>,
+    aws_profile: Option<&Profile>,
     aws_creds: &AwsCredsFile,
 ) -> Option<i64> {
     let expiration_env_vars = ["AWS_SESSION_EXPIRATION", "AWSUME_EXPIRATION"];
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/modules/aws.rs [hunk 1]
```diff
@@ -148,13 +148,38 @@ fn alias_name(name: Option<String>, aliases: &HashMap<String, &str>) -> Option<S
...
+        None => &empty_section,
+    };
 
-    let section = get_profile_config(config, aws_profile)?;
-    Some(section.contains_key("credential_process") || section.contains_key("sso_start_url"))
+    Some(
+        config_section.contains_key("credential_process")
+            || config_section.contains_key("sso_start_url")
+            || credential_section.contains_key("credential_process")
+            || credential_section.contains_key("sso_start_url"),
```
davidkna, line 181: `has_credential_process_or_sso` has been mostly concerned with the config file and `has_defined_credentials` with the credentials file. Given that, please move the `has_defined_credentials` into `has_defined_credentials` or change this function to first do the check on the config file and either return on success, or follow up with the credentials file check.

src/modules/aws.rs
```diff
@@ -148,13 +148,38 @@ fn alias_name(name: Option<String>, aliases: &HashMap<String, &str>) -> Option<S
...
+    aws_creds: &AwsCredsFile,
 ) -> Option<bool> {
     let config = get_config(context, aws_config)?;
+    let credentials = get_creds(context, aws_creds);
+
+    let empty_section = ini::Properties::new();
+    // We use the aws_profile here because `get_profile_config()` treats None
+    // as "special" and falls back to the "[default]"; otherwise this tries
+    // to look up "[profile default]" which doesn't exist
+    let config_section = get_profile_config(config, aws_profile).or(Some(&empty_section))?;
```
davidkna, line 162: Try using `get_profile_creds` instead. Either way, in the non-success case, I would prefer just working with `None` instead of an empty config section.

davidkna (changes requested):

src/modules/aws.rs
offbyone: That doesn't quite feel right to me. That checks for an actual active credential, which in this case would require me to _invoke_ the credential process under the hood... _not_ something we want to do in the prompt. This is just looking for the configuration that _enables_ credentials, in the second place where it can be found. The AWS CLI allows for it to be configured in two places (which is a bad idea, for sure!) but right now Starship only looks in one of them.

If I'm going to change that, I think the name of those methods should change as well. I'll update the PR with the changes from the below, but I think this method's the right place for the code. If you insist otherwise, I'll move it, but it feels wrong to me.

davidkna: I would still prefer If you changed this code to do the checks sequentially, to only read each file based on necessity. But since it looks like they both will be read in most cases either way, keeping it this way would be fine too.

offbyone: That was my thinking too; the most common configuration will have both files read, so I don't think there's a lot of benefit or optimization in making that more complex. Doubly so because starship caches the read, so it's still pretty safe.

davidkna (approved):
more or less lgtm

andytom:
Thank you for your contribution @offbyone and thank you for reviewing @davidkna.

PR #4614, opened 2022-11-16
fix(localip): avoid external network connections

cgzones:
#### Description
The currently used crate local_ipaddress opens a network connection to 8.8.8.8 (Google DNS) on each prompt.  This might invade the users privacy.  Use the crate local_ip_address instead, which uses local system APIs.

#### Motivation and Context
Avoid external network connectio to Google on each promt (if the module `localip` is enabled).
See https://github.com/egmkang/local_ipaddress/blob/d64c66b9593537b3d2b1bff11dfac61891a7845a/src/lib.rs#L13

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1617,10 +1617,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d4d2456c373231a208ad294c33dc5bff30051eafd954cd4caae83a712b12854d"
 
 [[package]]
-name = "local_ipaddress"
-version = "0.1.3"
+name = "local-ip-address"
+version = "0.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6a104730949fbc4c78e4fa98ed769ca0faa02e9818936b61032d2d77526afa9"
+checksum = "9eab7f092fb08006040e656c2adce6670e6a516bea831a8b2b3d0fb24d4488f5"
+dependencies = [
+ "libc",
+ "neli",
+ "thiserror",
+ "windows-sys 0.42.0",
+]
 
 [[package]]
 name = "lock_api"
@@ -1748,6 +1754,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "neli"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9053554eb5dcb7e10d9cdab1206965bde870eed5d0d341532ca035e3ba221508"
+dependencies = [
+ "byteorder",
+ "libc",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


chipbuster:
There was a brief discussion on this [when the localip module was first proposed](
https://github.com/starship/starship/pull/3289), though more from a performance standpoint than a privacy one. Most casual resources claim that `connect()` on a UDP socket only sets up the destination addresses, but doesn't actually send data.

If you're aware of an example where the OS sends data to the remote on `connect()` please let us know.

Even without that, I wouldn't be opposed to changing dependencies. The ability to pull IPv4 and IPv6 addresses from the same dependency would be nice. However, I do have an additional concern: [this file](https://github.com/egmkang/local_ipaddress/blob/d64c66b9593537b3d2b1bff11dfac61891a7845a/src/lib.rs) is essentially the entirety of the current dependency. It can be audited trivially, and only relies on the Rust standard library.

The proposed dependency requires [somewhat more intimate knowledge](https://github.com/EstebanBorai/local-ip-address/blob/ac6c96e5536de113346005e459902e47f5d2f3d3/src/linux.rs#L146) of how networking works on each specific platform it supports. Even if we ignore this altogether and assume the author(s) know exactly what they're doing (which may be a fair assumption!), there appear to be known [memory safety issues](https://github.com/EstebanBorai/local-ip-address/issues/24) and [problems with FreeBSD support](https://github.com/EstebanBorai/local-ip-address/issues/40) which would block us from using it for the time being.

chipbuster:
Looks like OpenBSD support is being merged soon, and they're hopeful that this will also fix FreeBSD support, but we'd need to check for both FreeBSD and NetBSD before being sure.

davidkna:
[systemstat](https://crates.io/crates/systemstat) seems to have fairly complete cross-platform support for listing connected networks with their addresses, and it seems to use the same library surfaces that this library uses. On the other hand, this data would still need to be evaluated correctly to produce the expected primary local ip address.

cgzones:
An alternative might be to continue the approach from `local_ipaddress`, but use a different destination IP address, e.g. one of the *Documentation* ones from https://en.wikipedia.org/wiki/IPv4#Special-use_addresses.

chipbuster:
@cgzones Just to confirm, are you aware of an example where connecting a UDP socket causes a packet transmission to the remote, or is this a "just in case" type of change?

> An alternative might be to continue the approach from `local_ipaddress`, but use a different destination IP address, e.g. one of the _Documentation_ ones from https://en.wikipedia.org/wiki/IPv4#Special-use_addresses.

I think this change would probably best be made upstream. The 192.168.*.* family looks like a good candidate for the remote in these cases, since it should never be routed outside of the local network.

cgzones:
> Just to confirm, are you aware of an example where connecting a UDP socket causes a packet transmission to the remote, or is this a "just in case" type of change?

No, I am not aware of any.
I just stumbled over the connect syscall with the Google IP.

andytom:
Replaced by #4648

PR #4616, opened 2022-11-16
fix(java): Improved regex for Java version (starship#4610)

### LINKED ISSUE #4610: Java module displays incorrect version number
#### Current Behavior
The version shown for java is not the version number users would expect to see.  It appears to be a vendor-specific version.

For example when I run "java -version", this is the output:

openjdk version "17.0.4" 2022-07-19 LTS
OpenJDK Runtime Environment Zulu17.36+13-CA (build 17.0.4+8-LTS)
OpenJDK 64-Bit Server VM Zulu17.36+13-CA (build 17.0.4+8-LTS, mixed mode, sharing)

The version reported by the java module in starship is "17.36".  I expected to see "17.0.4" not the "Zulu" version.

Note that for JDK 8 it gets the version right, even though the output of "java -version" is not significantly different:

openjdk version "1.8.0_332"
OpenJDK Runtime Environment (Zulu 8.62.0.19-CA-macosx) (build 1.8.0_332-b09)
OpenJDK 64-Bit Server VM (Zulu 8.62.0.19-CA-macosx) (build 25.332-b09, mixed mode)

In this case the java module shows "1.8.0" rather than the "8.62" "Zulu" version, so it isn't even consistent.

#### Expected Behavior
Java module should show the JDK version rather than the vendor specific build version.

#### Environment
- Starship version: 1.11.0
- zsh version: zsh 5.8.1 (x86_64-apple-darwin21.0)
- Operating system: Mac OS 12.6.2
- Terminal emulator: WarpTerminal v0.2022.11.08.08.07.stable_02
- Git Commit Hash: 
- Branch/Tag: master
- Rust Version: rustc 1.64.0 (a55dd71d5 2022-09-19)
- Build Time: 2022-10-15 01:46:07 +00:00

```bash
# If you come from bash you might have to change your $PATH.
export PATH=$HOME/bin:/usr/local/bin:$PATH

# Path to your oh-my-zsh installation.
export ZSH=$HOME/.oh-my-zsh

# Set name of the theme to load --- if set to "random", it will
# load a random theme each time oh-my-zsh is loaded, in which case,
# to know which specific one was loaded, run: echo $RANDOM_THEME
# See https://github.com/robbyrussell/oh-my-zsh/wiki/Themes
ZSH_THEME="af-magic"

plugins=(mercurial osx)

source $ZSH/oh-my-zsh.sh

alias jfind='find . -name "*.java" | xargs grep -i '


# Need a prompt that uses hg_prompt_info
# PROMPT='${ret_status}%{$fg_bold[green]%}%p %{$fg[cyan]%}%c %{$fg_bold[blue]%}$(git_prompt_info)$(hg_prompt_info)%{$fg_bold[blue]%} % %{$reset_color%}'
eval "$(starship init zsh)"

#THIS MUST BE AT THE END OF THE FILE FOR SDKMAN TO WORK!!!
export SDKMAN_DIR="/Users/scott/.sdkman"
[[ -s "/Users/scott/.sdkman/bin/sdkman-init.sh" ]] && source "/Users/scott/.sdkman/bin/sdkman-init.sh"
```

#### Starship Configuration

```toml

[hg_branch]
disabled = false

[aws]
disabled = true

[java]
version_format = "v${raw}"

```

swpalmer:
#### Description
Modified regex for extracting Java version to get the version number expected by most users rather than a vendor-specific version.

#### Motivation and Context

Closes #4610

Ran "./target/debug/starship module java" with multiple versions of JDK from Azul.

#### Checklist:
No doc or test changes needed.

### DIFF
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -5,7 +5,7 @@ use crate::utils::get_command_string_output;
 use std::path::PathBuf;
 
 use regex::Regex;
-const JAVA_VERSION_PATTERN: &str = "(?:JRE.*\\()(?P<version>[\\d\\.]+)";
+const JAVA_VERSION_PATTERN: &str = "(?:JRE.*\\(|OpenJ9 )(?P<version>[\\d]+(?:\\.\\d+){0,2})";
 
 /// Creates a module with the current Java version
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
@@ -120,15 +120,17 @@ mod tests {
     fn test_parse_java_version_zulu() {
         let java_8 = "OpenJDK 64-Bit Server VM (25.222-b10) for linux-amd64 JRE (Zulu 8.40.0.25-CA-linux64) (1.8.0_222-b10), built on Jul 11 2019 11:36:39 by \"zulu_re\" with gcc 4.4.7 20120313 (Red Hat 4.4.7-3)";
         let java_11 = "OpenJDK 64-Bit Server VM (11.0.4+11-LTS) for linux-amd64 JRE (Zulu11.33+15-CA) (11.0.4+11-LTS), built on Jul 11 2019 21:37:17 by \"zulu_re\" with gcc 4.9.2 20150212 (Red Hat 4.9.2-6)";
+        let java_17 = "OpenJDK 64-Bit Server VM (17.0.5+8-LTS) for bsd-amd64 JRE (17.0.5+8-LTS) (Zulu17.38+21-CA), built on Oct  7 2022 06:03:12 by \"zulu_re\" with clang 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.17)";
         assert_eq!(parse_java_version(java_8), Some("1.8.0".to_string()));
         assert_eq!(parse_java_version(java_11), Some("11.0.4".to_string()));
+        assert_eq!(parse_java_version(java_17), Some("17.0.5".to_string()));
     }
 
     #[test]
     fn test_parse_java_version_eclipse_openj9() {
         let java_8 = "Eclipse OpenJ9 OpenJDK 64-bit Server VM (1.8.0_222-b10) from linux-amd64 JRE with Extensions for OpenJDK for Eclipse OpenJ9 8.0.222.0, built on Jul 17 2019 21:29:18 by jenkins with g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)";
         let java_11 = "Eclipse OpenJ9 OpenJDK 64-bit Server VM (11.0.4+11) from linux-amd64 JRE with Extensions for OpenJDK for Eclipse OpenJ9 11.0.4.0, built on Jul 17 2019 21:51:37 by jenkins with g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)";
-        assert_eq!(parse_java_version(java_8), Some("1.8.0".to_string()));
+        assert_eq!(parse_java_version(java_8), Some("8.0.222".to_string()));
         assert_eq!(parse_java_version(java_11), Some("11.0.4".to_string()));
     }
 
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -5,7 +5,7 @@ use crate::utils::get_command_string_output;
 use std::path::PathBuf;
 
 use regex::Regex;
-const JAVA_VERSION_PATTERN: &str = "(?P<version>[\\d\\.]+)[^\\s]*\\s(?:built|from)";
+const JAVA_VERSION_PATTERN: &str = "(?:JRE.*\\()(?P<version>[\\d\\.]+)";
 
 /// Creates a module with the current Java version
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {


davidkna:
Please update the tests, and as far as I can tell this regex wouldn't be able to satisfy all the existing test cases.

swpalmer:
Working on it.  Sorry, I'm not a Rust programmer so I have to get used to the tools and how run the tests, where to find the test code, etc.

swpalmer:
Would it be acceptable to change the output for Eclipse J9  JDK 8?

Currently from this:
Eclipse OpenJ9 OpenJDK 64-bit Server VM (1.8.0_222-b10) from linux-amd64 JRE with Extensions for OpenJDK for Eclipse OpenJ9 8.0.222.0, built on Jul 17 2019 21:29:18 by jenkins with g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)

The reported version is currently "1.8.0"
However I have a regex that will extract "8.0.222"  Which is actually more useful as it would communicate JDK 8u222.  The update number is actually the only useful number that might be changing for a JDK 8 install.  For later JDKs from Eclipse J9 the appropriate triplet is extracted, e.g.:

Eclipse OpenJ9 OpenJDK 64-bit Server VM (11.0.4+11) from linux-amd64 JRE with Extensions for OpenJDK for Eclipse OpenJ9 11.0.4.0, built on Jul 17 2019 21:51:37 by jenkins with g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)

results in "11.0.4"

Eclipse J9 is the only one of the tests that requires tweaking to the regex.  My revised regex that outputs the above values is:

(?:JRE.*\(|OpenJ9 )(?P<version>[\d]+(?:\.\d+){0,2})

swpalmer:
Actually a better regex is:

(?:JRE.*\(|OpenJ9 )(?P\<version\>[\d]+(?:\.\d+){0,2}(?:(_|u)\d+)?)

Which captures the "update version" for JDK 8 from various vendors. Eg.:

OpenJDK 64-Bit GraalVM CE 19.2.0.1 (25.222-b08-jvmci-19.2-b02) for linux-amd64 JRE (8u222), built on Jul 19 2019 17:37:13 by \"buildslave\" with gcc 7.3.0

results in "8u222" instead of just "8"

OpenJDK 64-Bit Server VM (25.222-b10) for linux-amd64 JRE (Zulu 8.40.0.25-CA-linux64) (1.8.0_222-b10), built on Jul 11 2019 11:36:39 by \"zulu_re\" with gcc 4.4.7 20120313 (Red Hat 4.4.7-3)

results in "1.8.0_222" instead of just "1.8.0"

Eclipse OpenJ9 OpenJDK 64-bit Server VM (1.8.0_222-b10) from linux-amd64 JRE with Extensions for OpenJDK for Eclipse OpenJ9 8.0.222.0, built on Jul 17 2019 21:29:18 by jenkins with g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)

results in "8.0.222" instead of "1.8.0" 
I fee this change from 1.8.0 to 8.0.222 is less confusing that the current issue that captures the Zulu-specific number that does not reflect the JDK version for JDK 17 and 19.

If this is acceptable, I will change the tests to match and go with the above regex.

src/modules/java.rs
```diff
@@ -120,15 +120,17 @@ mod tests {
...
         assert_eq!(parse_java_version(java_11), Some("11.0.4".to_string()));
+        assert_eq!(parse_java_version(java_17), Some("17.0.5".to_string()));
     }
 
     #[test]
     fn test_parse_java_version_eclipse_openj9() {
         let java_8 = "Eclipse OpenJ9 OpenJDK 64-bit Server VM (1.8.0_222-b10) from linux-amd64 JRE with Extensions for OpenJDK for Eclipse OpenJ9 8.0.222.0, built on Jul 17 2019 21:29:18 by jenkins with g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)";
         let java_11 = "Eclipse OpenJ9 OpenJDK 64-bit Server VM (11.0.4+11) from linux-amd64 JRE with Extensions for OpenJDK for Eclipse OpenJ9 11.0.4.0, built on Jul 17 2019 21:51:37 by jenkins with g++ (GCC) 7.3.1 20180303 (Red Hat 7.3.1-5)";
-        assert_eq!(parse_java_version(java_8), Some("1.8.0".to_string()));
+        assert_eq!(parse_java_version(java_8), Some("8.0.222".to_string()));
```
chipbuster, line 133: As someone who recently got bamboozled by Java version numbers, there's no potential ambiguity here, correct? 1.8.0 and 8.0.x are always the same version?

If so, I think this change is fine (esp. as Java 8 seems to be mostly EOL except for legacy at this point).

chipbuster (commented):
It feels a little weird having different outputs for OpenJ9 and the other JVMs, but I suppose this can't be helped. Just need to verify one thing about version numbers (see question in review comment) and we should be good to approve this.

Performance testing on my system shows somewhere between a 0-4ms change to parse all Java versions in the test. The numbers seem to differ depending on what I'm running in the background, and are pretty insignificant anyways.

src/modules/java.rs
swpalmer: Yes that's right. They started calling it "Java 8" before they fixed the reported version numbers to actually reflect how people referred to it.

swpalmer:
@chipbuster yes, I agree that it isn't great to have it different, but the only way around it seems to be to use the much slower "-version" option. Java 8 is sort of legacy. There is a commitment to support it for several more years I think, but it is ancient at this point. The point being that people using it won't be very confused to see 8.0 instead of 1.8, at least compared to the weirdness that was happing with the later Zulu versions.

chipbuster (approved):

andytom:
Thank you for your contribution @swpalmer and thank you for reviewing @chipbuster.

PR #4619, opened 2022-11-17
fix(ci): cache after selecting the toolchain

MarcoIeni:
I noticed the CI initializes the cache BEFORE installing the toolchain.
However, the documentation of [rust-cache](https://github.com/Swatinem/rust-cache#example-usage) says you should call the rust-cache action AFTER you install the toolchain


andytom (approved):
LGTM

chipbuster:
Thanks for the catch!

PR #4648, opened 2022-11-25
feat(localip): use reserved remote address

cgzones:
#### Description
Instead of the remote address of 8.8.8.8 (Google DNS) in the crate `local_ipaddress` use a reserved IPv4 address, that should never be assigned.
Also forward the underlying error on failure.

#### Motivation and Context
The current used crate `local_ipaddress` connects an UDP socket to 8.8.8.8 (Google DNS), which might trigger some privacy concerns (although no actual packet should be send). 

Supersedes: #4614

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1631,12 +1631,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8f9f08d8963a6c613f4b1a78f4f4a4dbfadf8e6545b2d72861731e4858b8b47f"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


chipbuster (changes requested):
I'll want to check this on my Windows and MacOS machines to make sure nothing breaks, but I think this is in pretty good form at the minute. A few small tweaks.

chipbuster (approved):
👍

andytom:
Thank you for your contribution @cgzones and thank you for reviewing @chipbuster.

PR #4656, opened 2022-11-27
feat: Add ssh-agent module

bwestlin:
#### Description
Adds a `ssh_agent` module that shows if one or more private key identities are added to the [OpenSSH authentication](https://en.wikipedia.org/wiki/Ssh-agent) agent. Check is done through `ssh-agent -l` which list added key identities or exits with a non-zero exit code if no keys are added.

#### Motivation and Context
This is a feature request. It helps quickly identifying if key identities are added to the agent using the prompt.

image

Tested using Konsole with BitstreamVeraSansMono Nerd font & Vscode terminal on Linux with various configuration options and symbols. 

#### Checklist:


src/configs/ssh_agent.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub disabled: bool,
+}
+
+impl<'a> Default for SshAgentConfig<'a> {
+    fn default() -> Self {
+        Self {
+            format: "[ssh $symbol]($style)",
+            symbol: "🔑 ",
+            symbol_multi: "🔑➕ ",
+            none_symbol: "❌ ",
```
jankatins, line 22: `symbol_none`, to fit the other two names?

bwestlin: For some reason the `Format + Docs Workflow / Dprint [Docs Formatter] (pull_request) ` worklow is now failing but I'm unable to reproduce the issue locally.

PR #4688, opened 2022-12-05
feat: Add Fig completions

nihaals:
#### Description
This adds [Fig](https://github.com/withfig/autocomplete) completions to the existing `completions` command. To avoid adding boilerplate to support `clap_complete_fig`, this also refactors the `completions` command to use [`clap_complete_command`](https://crates.io/crates/clap_complete_command) (full disclosure, I'm the author of this crate), which wraps the completion crates.

#### Motivation and Context

This makes it easier to add Starship suggestions to tools that use Fig's completion specs, such as [Fig itself](https://fig.io).

- [x] I have tested using **macOS**

This isn't OS-specific so I've only tested it on macOS.

#### Checklist:

I don't see the specific shells supported mentioned in either the docs or tests.


Cargo.toml
```diff
@@ -44,7 +44,7 @@ git-repository-faster = ["git-features/zlib-stock", "git-repository/fast-sha1"]
 [dependencies]
 chrono = { version = "0.4.23", features = ["clock", "std"] }
 clap = { version = "4.0.29", features = ["derive", "cargo", "unicode"] }
-clap_complete = "4.0.6"
+clap_complete_command = "0.4.0"
```
matchai, line 47: I'm reluctant to replace `clap_complete` to avoid the little boilerplate that would be needed to add Fig support.
Would we be able to use `clap_complete_fig` instead?

nihaals: We could, but it gets a bit annoying and messy and you lose basically the whole nice API of `clap_complete`. Just needing to add to the `ValueEnum` means you now need to make your own enum and implement the `clap_complete` and `clap` traits, which also needs to be updated if you want to use other `clap_complete` crates too. When I was doing this exact thing and realised I would need to use it in multiple projects, I ended up making the crate, which also allowed me to add some extra things that would be a bit annoying to maintain across projects, like changing the API to be more consistent and simpler and hand-rolling one of the traits.

You can see an example of what it looks like with the extra boilerplate vs. what it looks like with `clap_complete_command`: https://github.com/PyO3/maturin/commit/e41cd23146cb06497b14966fc9dd1f01edbbf9b7

Since the wrapper is already fairly thin, the extra boilerplate needed to support both `clap_complete` and `clap_complete_fig` would already look fairly similar to `clap_complete_command`, and `clap_complete_command` supports `clap`/`clap_complete`/`clap_complete_fig` v4.x, so updating `clap_complete` is fairly simple. Outside of Clap v5 releasing, I'm not sure there's much of a disadvantage to using the wrapper crate, and since the update to Starship's boilerplate would look very similar to the update needed for `clap_complete_command` to support a Clap v5 as well, maybe it makes sense to focus those efforts to help others with the same end goal to update too instead of everyone with similar boilerplate needed to update themselves. An added bonus of concerted efforts is extra things like https://github.com/nihaals/clap-complete-command/issues/10, which I haven't seen many CLIs add themselves, but might be more likely to use if it's with the same API as shell completions they already have.

Is there another aspect you're concerned about?

davidkna: I would also prefer not adding a crate just for simplifying the use of fig completions, and compared to maturin, I think we could implement this with a bit less code using an `Either<clap_complete::Shell, Other>` construct.

In addition, IIRC  `clap_complete_fig` is supposed to get merged into `clap_complete` at some point, and we could just wait for that to happen. Also, for fig, completions are usually pulled from a central repository as far as I know. Thus, adding a publishing workflow to the release workflow and generating the fig completion in `build.rs` might be more appropriate than to add a command to generate them at runtime. But I'm not convinced that it's worth implementing that option.

nihaals: > - Cost of keeping the dependency up-to-date
> - Risk that the dependency isn't secure
> - Risk that the dependency won't continue to be maintained

I definitely agree here.

> IIRC `clap_complete_fig` is supposed to get merged into `clap_complete` at some point

I wasn't aware of this but I think it makes more sense than the current separate crate and it will make it much easier to use both. Since adding Fig isn't a priority, I think waiting until it's merged does make more sense unless it becomes a higher priority request.

> Also, for fig, completions are usually pulled from a central repository as far as I know. Thus, adding a publishing workflow to the release workflow and generating the fig completion in `build.rs` might be more appropriate than to add a command to generate them at runtime. But I'm not convinced that it's worth implementing that option.

This is true and has given me mixed feelings on it too. All the Fig completions used by default are part of the main repo, and there isn't an easy way of importing local ones. The only way as far as I'm aware is uploading it to your account, which is more meant for private tools than public ones that aren't in the main repo. Fig specs are technically specs, and so another tool could use them and import them in a completely different way, and I personally think they fit well with the other shell completions, but another approach would absolutely be making it at compile time or making it some kind of dev command which is built upon as opposed to blindly imported (more for advanced features like dynamic suggestions).

I think it makes sense to close this now and then potentially revisit it in the future if it's more requested and with another approach that fits the project better.

PR #4694, opened 2022-12-07
feat(c++): Implemented c++ module

### LINKED ISSUE #4583: Add a module for C++
Starship shows the GCC version if there is a `.c` or `.h` file in the directory, but it doesn't detect `.cpp` and other C++ files. 

Starship should show the gcc version (since g++ is a part of the gcc project, it always has the same version number as gcc) if there is a `.cpp`, `.cxx`, `.cc` or `.hpp` file in the working directory. This shouldn't be too difficult, since the C module can already detect the gcc version and we can build something off that module.

Funinja:
#### Description

Created c++ module to allow the user to customize their prompt when in a folder with c++ like files. Also created test cases and documentation for the module.

#### Motivation and Context

Starship does not have a c++ module.

Closes #4583

#### Checklist:


davidkna:
See also the discussion in #3880

PR #4696, opened 2022-12-07
chore: Apply clippy suggestions

amaanq:
#### Description

Applied a few clippy suggestions, namely inlining format variables

#### Motivation and Context

Better styling/readability

Doesn't really have changes that could break on other platforms

#### Checklist:

N/A


davidkna (changes requested):
I actually already fixed most instances of this after the last rust release, but we've had some recent additions of non-inlined `format!` arguments.

This looks good to me if you fix the compilation issue.

PR #4713, opened 2022-12-12
fix(nodejs): apply `style` even if node version is unavailable

Andree37:
#### Description
Allowed `nodejs_version` to be unset (node isn't installed) and assign the node configuration

#### Motivation and Context
Closes https://github.com/starship/starship/issues/4705

image

#### Checklist:

I don't believe they are applicable.

### DIFF
--- a/src/modules/nodejs.rs
+++ b/src/modules/nodejs.rs
@@ -92,18 +92,18 @@ fn get_engines_version(context: &Context) -> Option<String> {
     Some(raw_version.to_string())
 }
 
-fn check_engines_version(nodejs_version: Option<&String>, engines_version: Option<String>) -> bool {
+fn check_engines_version(nodejs_version: Option<&str>, engines_version: Option<String>) -> bool {
     let (nodejs_version, engines_version) = match (nodejs_version, engines_version) {
         (Some(nv), Some(ev)) => (nv, ev),
         _ => return true,
     };
-    let r = match VersionReq::parse(&engines_version.unwrap()) {
+    let r = match VersionReq::parse(&engines_version) {
         Ok(r) => r,
         Err(_e) => return true,
     };
     let re = Regex::new(r"\d+\.\d+\.\d+").unwrap();
     let version = re
-        .captures(nodejs_version.unwrap())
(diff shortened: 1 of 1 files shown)


src/modules/nodejs.rs
```diff
@@ -92,8 +92,8 @@ fn get_engines_version(context: &Context) -> Option<String> {
     Some(raw_version.to_string())
 }
 
-fn check_engines_version(nodejs_version: &str, engines_version: Option<String>) -> bool {
-    if engines_version.is_none() {
+fn check_engines_version(nodejs_version: Option<&String>, engines_version: Option<String>) -> bool {
+    if engines_version.is_none() || nodejs_version.is_none() {
         return true;
     }
```
davidkna, line 98: I would prefer not calling `unwrap` unless strictly necessary to avoid issue in case of potential refactorings, Perhaps use something like this?

```suggestion
    let (nodejs_version, engines_version) = match (nodejs_version, engines_version) {
        (Some(nv), Some(ev)) => (nv, ev),
        _ => return true,
    };
```

src/modules/nodejs.rs
```diff
@@ -45,7 +45,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "style" => {
                     let engines_version = get_engines_version(context);
                     let in_engines_range =
-                        check_engines_version(nodejs_version.deref().as_ref()?, engines_version);
+                        check_engines_version(nodejs_version.deref().as_ref(), engines_version);
```
davidkna, line 48: `Option<&str>` will usually work better.

```suggestion
                        check_engines_version(nodejs_version.as_deref(), engines_version);
```

davidkna (changes requested):
Thanks for the PR!

Please also write a test case for this problem, and if possible, please also fix the other things I pointed out.

Andree37:
I've added an unit test and applied the commits that you mentioned.
:)

davidkna (approved):

andytom:
Thank you for your contribution @Andree37

PR #4655, opened 2022-11-26
feat(release): handle choco portable & install variants

davidkna:
#### Description

In addition to starship on chocolatey, also handle `starship.install` (MSI installer) and `starship.portable` and make `starship` an empty meta-package that only depends on `starship.install`. MSI/installer packages seem to be preferred over zip-based installers on chocolatey. Proper virtual packages that allow choosing either a `portable` or `install` variant aren't implemented in chocolatey yet.
~~I made the installation script handle all possible variants dynamically, based on the parent package name.~~

#### Motivation and Context

Closes #4637

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [x] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #4723, opened 2022-12-14
feat(release): handle chocolatey starship.portable and starship.install pkg publishing

davidkna:
#### Description

> In addition to starship on chocolatey, also handle `starship.install` (MSI installer) and `starship.portable` and make `starship` an empty meta-package that only depends on `starship.install`. MSI/installer packages seem to be preferred over zip-based installers on chocolatey. Proper virtual packages that allow choosing either a `portable` or `install` variant aren't implemented in chocolatey yet.

I think this is ready now, and I also added a potential fix for the push failure from the last release.

#### Motivation and Context

Related #4655

#### Checklist:


matchai (approved):
LGTM! 👍

PR #4724, opened 2022-12-15
feat(nix): support new `nix shell` command

### LINKED ISSUE #2950: Support new `nix shell` command
When entering a nix-shell with the [new cli command](https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-shell.html) the prompt is not updated to give any indication as it does when using the older `nix-shell` command.

I'd like to have some form of indication like the old `nix-shell` command has. Compare the prompt when using the new vs. the old command:
```
/tmp/tmp.tPUR3PZHrx
» nix-shell -p hello

/tmp/tmp.tPUR3PZHrx via ❄  impure (shell)
» exit

/tmp/tmp.tPUR3PZHrx took 3s
» nix shell 'nixpkgs#hello'

/tmp/tmp.tPUR3PZHrx
» exit
```

jjant:
#### Description
Closes #2950.

Adds two options for `nix_shell`:
- `heuristic`: Enables detection of `nix shell` with the heuristic described below (default: disabled)
- `unknown_msg`: Like `pure/impure_msg` for when we can't detect which shell type it is

#### Motivation and Context
The current `nix_shell` module only works with `nix-shell`, not `nix shell`.

There's no straightforward way to detect if we're in a `nix shell` yet, so I've used a workaround that checks if we have a nix store path in our `PATH` to detect if we're in a `nix shell`.

I've added tests in the `nix_shell` module.

#### Checklist:

### DIFF
--- a/src/modules/nix_shell.rs
+++ b/src/modules/nix_shell.rs
@@ -41,14 +41,17 @@ impl NixShellType {
 /// determine if it's inside a nix-shell and the name of it.
 ///
 /// The following options are availables:
-///     - `impure_msg` (string) // change the impure msg
-///     - `pure_msg` (string)   // change the pure msg
+///     - `impure_msg` (string)  // change the impure msg
+///     - `pure_msg` (string)    // change the pure msg
+///     - `unknown_msg` (string) // change the unknown message
 ///
 /// Will display the following:
 ///     - pure (name)    // $name == "name" in a pure nix-shell
 ///     - impure (name)  // $name == "name" in an impure nix-shell
 ///     - pure           // $name == "" in a pure nix-shell
 ///     - impure         // $name == "" in an impure nix-shell
+///     - unknown (name) // $name == "name" in an unknown nix-shell
+///     - unknown        // $name == "" in an unknown nix-shell
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("nix_shell");
     let config: NixShellConfig = NixShellConfig::try_load(module.config);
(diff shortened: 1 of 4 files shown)


src/modules/nix_shell.rs
```diff
@@ -3,6 +3,36 @@ use super::{Context, Module, ModuleConfig};
...
+
+        shell_type_format_from_env.or_else(|| Self::in_new_nix_shell(context).map(|_| Unknown))
+    }
+
+    // Hack to detect if we're in a `nix shell` (in constrast to a `nix-shell`).
+    // A better way to do this will be enabled by https://github.com/NixOS/nix/issues/6677.
+    fn in_new_nix_shell(context: &Context) -> Option<()> {
+        let path = context.get_env("PATH")?;
+
+        path.contains("/nix/store").then_some(())
```
davidkna, line 32: To reduce false positive, please use `env::split_paths` and check if any of the paths _starts_ with `/nix/store/`.

davidkna (changes requested):
I think I would prefer disabling this by default and gating it with a `use_heuristics`-option since `nix shell` not being detectable (as opposed to `nix develop`) seems to be an intentional upstream choice.

Please also document that there is now a third unknown nix-shell state. (Having the message for the unknown state be configurable might be good too, but I think we could wait for someone to make a feature request first.)

src/modules/nix_shell.rs
jjant: Thank you, done in https://github.com/starship/starship/pull/4724/commits/5d369fb50adf8fc58e3929e41e338970f0489edd.

jjant:
@davidkna I've added a `heuristic` flag (default: disabled) in https://github.com/starship/starship/pull/4724/commits/f5f90899e6742e2ad18953dc58475e45d09bad3d.

I used `heuristic` instead of `use_heuristic` to match the existing flag in the `dotnet` module's config.

davidkna (approved):

andytom:
Thank you for your contribution @jjant and thanks for reviewing @davidkna.

PR #4729, opened 2022-12-17
docs(config): clarify git-commit tag variable

### LINKED ISSUE #4640: Git tags not displayed + Git Commit module doesn't have a tag variable (conflicted docs)
#### Current Behavior

The default format for the [Git Commit](https://starship.rs/config/#git-commit) module is `'[\($hash$tag\)]($style) '`
However, the 'Variables' section doesn't list a variable called tag.

image

Furthermore, it's not just missing from the documentation. But also in the functionality.
I have tagged a commit and then did a git checkout for it, the commit hash is displayed but not the tag (not even the tag symbol by itself).

And yes, I did have the option to disable it set to false in my .toml file for Starship:
`tag_disabled = false`
In fact, I tried setting it to false and true and no tag was displayed either way.

#### Expected Behavior

I expect to see the tag symbol along with the tag name whenever I perform `git checkout` for a commit that I have tagged, as long as I have the option `tag_disabled` set to `false` in the `git_commit` module.

Essentially: 1. there is a conflict in the docs (a tag variable shown in the format string, but not the variables listed), and 2. there is a missing functionality in that module (commit tags now displayed).

#### Environment
- Starship version: 1.11.0
- zsh version: zsh 5.9 (arm-apple-darwin21.3.0)
- Operating system: Mac OS 13.0.1
- Terminal emulator: iTerm.app 3.4.18
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.64.0
- Rust channel:  release
- Build Time: 2022-10-15 01:40:28 +00:00

```bash
echo '.zshrc stuff are loaded!'
pfetch

# ===== Change ZSH Options =====
# man zshoptions

# ===== Create Aliases =====
alias ls=exa
alias exa='exa -lah --git --group-directories-first --no-time'
alias man=batman
alias grep='grep --color'
alias bbd='brew bundle dump --force --describe --file=~/.dotfiles/Homebrew/Brewfile'  # Update the Brewfile in ~/.dotfiles/Homebrew
# Using parameter expansion, apply the newline-separator option to the lowercase array version of 'path' and redirect the result to standard output using hereword (the default for which we've changed to bat).
alias trail='<<<${(F)path}'
# Don't get accustomed to a more forgivable version of rm if you do server work.
# alias rm=trash

# ===== Add Locations to $path Array =====
# This enforces uniqueness on the array:
# Duplicate items added to the right are removed,
# while duplicate items added to the left are kept and
# the equivalents on the right are removed.
typeset -U path
path=(
  $path
  "

cmdaltent:
The `git_commit` module uses a `tag` variable in its format string, which is not explained in the Variables section of this module.

Missing clarification of this `tag` variable is added to the documentation of the `git_commit` module.

Closes #4640

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1640,10 +1640,11 @@ The `git_commit` module shows the current commit hash and also the tag (if any) 
 
 ### Variables
 
-| Variable | Example   | Description                         |
-| -------- | --------- | ----------------------------------- |
-| hash     | `b703eb3` | The current git commit hash         |
-| style\*  |           | Mirrors the value of option `style` |
+| Variable | Example   | Description                                  |
+| -------- | --------- | -------------------------------------------- |
+| hash     | `b703eb3` | The current git commit hash                  |
+| tag      | `v1.0.0`  | The tag name if showing tag info is enabled. |
+| style\*  |           | Mirrors the value of option `style`          |
 
 *: This variable can only be used as a part of a style string
 


andytom (approved):
LGTM

chipbuster:
@cmdaltent Thanks for the catch + fix!

@andytom As always, appreciate the review.

PR #4734, opened 2022-12-18
fix(git_commit): fix potential test failure

Foster0123:
Starship git_commit latest tag lightweight test case failed, i fixed the error to pass the test case

### DIFF
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -428,7 +428,7 @@ mod tests {
             .output()?;
 
         create_command("git")?
-            .args(["tag" , "v2"])
+            .args(["tag", "v2"])
             .env("GIT_COMMITTER_DATE", "2022-01-01 00:00:01 +0000")
             .current_dir(repo_dir.path())
             .output()?;
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -373,16 +373,7 @@ mod tests {
             .output()?;
 
         let git_tag = create_command("git")?
-            .args([
-                "for-each-ref",
-                "--contains",
-                "HEAD",
-                "--sort=-taggerdate",
-                "--count=1",
-                "--format",
-                "%(refname:short)",
-                "refs/tags",
-            ])
+            .args(["describe", "--tags"])
             .current_dir(repo_dir.path())
             .output()?
             .stdout;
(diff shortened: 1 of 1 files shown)


src/modules/git_commit.rs
```diff
@@ -437,7 +437,7 @@ mod tests {
             .output()?;
 
         create_command("git")?
-            .args(["tag", "v2"])
+            .args(["describe" , "tag" , "v2"])
```
davidkna, line 440: What sort of failure are you seeing without this? The code here is meant to create a tag, not to run `describe`. The git tests can have intermittent failures.

Foster0123: i ran the cargo tests, and saw an a test case failed, after adding the describe tag it passes the test case

davidkna: But does it pass without your change?

Foster0123: error
before making changes

Foster0123: fix
after making changes

davidkna: Does it work if you replace:
```rust
 let git_tag = create_command("git")?
            .args([
                "for-each-ref",
                "--contains",
                "HEAD",
                "--sort=-taggerdate",
                "--count=1",
                "--format",
                "%(refname:short)",
                "refs/tags",
            ])
```
with this?
```rust
 let git_tag = create_command("git")?
            .args(["describe", "--tags"])
```

If not, please close this and open a bug report, because in my opinion it would mean that the behavior and not the test is incorrect somehow.

Foster0123: yes it worked after i replaced the code as u mentioned

davidkna: In that case, please change your PR to use that instead. I would also appreciate if you changed the other tag tests to use `git describe --tags` as the baseline instead.

Foster0123: if u mean changing the block of code u mentioned , then yes, i changed it. there was another instance of same code, so changed that as well with git describe --tags. please review the changes

Foster0123: ran the test cases, and all of those passed without failures

src/modules/git_commit.rs
```diff
@@ -437,22 +428,13 @@ mod tests {
             .output()?;
 
         create_command("git")?
-            .args(["tag", "v2"])
+            .args(["tag" , "v2"])
```
andytom, line 431: This change is making cargo fmt fail is you change this to

```suggestion
            .args(["tag", "v2"])
```

It should pass

davidkna (approved):
LGTM

PR #4743, opened 2022-12-19
refactor(kubernetes): Remove nested Options

chipbuster:
#### Description
Kubernetes module was previously a bit messy, with lots of unnecessarily nested options. Clean this up using filter_map and find_map, with two major results:

1. `ctx_components` is now a Vec<KubeCtxComponent> instead of a Vec<Option<KubeCtxComponent>>. This greatly simplified downstream processing of these context components.

2. Instead of storing a partial computation of the namespace in variables `kube_ns`, etc, compute them directly in the formatter mapping. This is made simpler (read: actually doable) by change 1. This change also removes nonlocal commenting (i.e. "This is safe because `var` contains a `Some` value") and replaces it with operations that do not require `unwrap` at all. 

#### Motivation and Context

Nested options made it difficult to understand and manipulate the operations of the Kubernetes module. I suspect this is partially why I initially found #4550 so difficult to understand.

Removing these nested options will hopefully make it easier to understand the code and modify it moving forward.

#### Checklist:

### DIFF
--- a/src/modules/kubernetes.rs
+++ b/src/modules/kubernetes.rs
@@ -160,7 +160,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
                 "namespace" => ctx_components
                     .iter()
-                    .find_map(|kube| kube.user.as_deref())
+                    .find_map(|kube| kube.namespace.as_deref())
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM

PR #4750, opened 2022-12-21
fix: Remove vulnerable time-0.1.x chrono dependency

### LINKED ISSUE #3163: RUSTSEC-2020-0071: Potential segfault in the time crate
> Potential segfault in the time crate

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `time`                      |
| Version             | `0.1.44`                   |
| URL                 | [https://github.com/time-rs/time/issues/293](https://github.com/time-rs/time/issues/293) |
| Date                | 2020-11-18                         |
| Patched versions    | `>=0.2.23`                  |
| Unaffected versions | `=0.2.0,=0.2.1,=0.2.2,=0.2.3,=0.2.4,=0.2.5,=0.2.6`               |

### Impact

Unix-like operating systems may segfault due to dereferencing a dangling pointer in specific circumstances. This requires an environment variable to be set in a different thread than the affected functions. This may occur without the user&#39;s knowledge, notably in a third-party library.

The affected functions from time 0.2.7 through 0.2.22 are:

- `time::UtcOffset::local_offset_at`
- `time::UtcOffset::try_local_offset_at`
- `time::UtcOffset::current_local_offset`
- `time::UtcOffset::try_current_local_offset`
- `time::OffsetDateTime::now_local`
- `time::OffsetDateTime::try_now_local`

The affected functions in time 0.1 (all versions) are:

- `at`
- `at_utc`

Non-Unix targets (including Windows and wasm) are unaffected.

### Patches

Pending a proper fix, the internal method that determines the local offset has been modified to always return `None` on the affected operating systems. This has the effect of returning an `Err` on the `try_*` methods and `UTC` on the non-`try_*` methods.

Users and library authors with time in their dependency tree should perform `cargo update`, which will pull in the updated, unaffected code.

Users of time 0.1 do not have a patch and should upgrade to an unaffected version: time 0.2.23 or greater or the 0.3. series.

### Workarounds

No workarounds are known.

### References

time-rs/time#293

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2020-0071.html) for additional details.

ixti:
The dependency is optional for chrono and enabled by default for backward compatibility only.

RUSTSEC-2020-0071: https://rustsec.org/advisories/RUSTSEC-2020-0071

Closes #3163

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,7 +1,0 @@
-[advisories]
-ignore = [
-  # Potential segfault in the time crate
-  # chrono dependency, but vulnerable function is never called
-  # Tacked in #3163
-  "RUSTSEC-2020-0071",
-]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,7 +309,6 @@ dependencies = [
  "js-sys",
  "num-integer",
  "num-traits",
- "time 0.1.44",
  "wasm-bindgen",
  "winapi",
 ]
@@ -1001,7 +1000,7 @@ dependencies = [
  "bstr",
  "itoa",
  "thiserror",
- "time 0.3.14",
+ "time",
 ]
 
 [[package]]
@@ -1660,7 +1659,7 @@ dependencies = [
  "dirs-next",
  "objc-foundation",
  "objc_id",
- "time 0.3.14",
+ "time",
 ]
 
 [[package]]
@@ -2662,7 +2661,7 @@ checksum = "6c401e795850edb4e9fdde5940f856364f0fbab573e8dea58f6ee5f85fcf471d"
 dependencies = [
  "const_format",
  "is_debug",
- "time 0.3.14",
+ "time",
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


syphar (commented):
you probably also can remove the `ignore` from `.cargo/audit.toml`, right?

davidkna:
> you probably also can remove the `ignore` from `.cargo/audit.toml`, right?

Agreed, but please just remove the whole file. Looks like I forgot to add the default-features toggle when attempting to restrict the feature set.

davidkna (approved):

ixti:
@davidkna you said:

> I forgot to add the default-features toggle when attempting to restrict the feature set

Does it mean we should remove `wasmbind` feature too?

davidkna:
@ixti I think it wasn't a feature yet when I changed it. You can keep it, but starship doesn't build for wasm targets either way.

ixti:
> @ixti I think it wasn't a feature yet when I changed it. You can keep it, but starship doesn't build for wasm targets either way.

Okay. Either way, I believe such cleanup belongs to a different PR ;))

davidkna:
Thanks for the fix!

PR #4760, opened 2022-12-26
fix(package): Improve regex for extracting gradle version from gradle.properties

### LINKED ISSUE #4759: Getting a package version from gradle.properties is not implemented correctly
#### Current Behavior
Commented out lines in gradle.properties with 'version=' will be matched as defining a version. versions defined with whitespace between 'version' and '=' will not be extract properly.

#### Expected Behavior
Comment lines should be ignored. Whitespace should be allowed on either side of '='.

E.g. this content in a gradle.properties file should NOT set the version to 0.0.1
```
# Generally it is better to set the version only for release builds, build time
# from command line using -Pversion=0.0.1
```

#### Possible Solution
Fix the regex on this line https://github.com/starship/starship/blob/master/src/modules/package.rs#L108

#### Environment
- Starship version: 1.12.0
- zsh version: zsh 5.8.1 (x86_64-apple-darwin22.0)
- Operating system: Mac OS 13.2.0
- Terminal emulator: WarpTerminal v0.2022.12.13.08.04.stable_01
- Git Commit Hash: 
- Branch/Tag: master
- Rust Version: rustc 1.65.0 (897e37553 2022-11-02)
- Build Time: 2022-12-13 16:12:55 +00:00

swpalmer:
#### Description
Changed regex to ```(?m)^\s*version\s*=\s*(?P<version>.*)``` to account for comment lines and possible whitespace.

#### Motivation and Context
Closes #4759

Added test with build.gradle file that contains commented out ```version=``` that was picked up by previous regex and with whitespace around the '='

- [x ] I have tested using **MacOS**

#### Checklist:

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -105,7 +105,7 @@ fn get_gradle_version(context: &Context, config: &PackageConfig) -> Option<Strin
     context
         .read_file_from_pwd("gradle.properties")
         .and_then(|contents| {
-            let re = Regex::new(r"version=(?P<version>.*)").unwrap();
+            let re = Regex::new(r"(?m)^\s*version\s*=\s*(?P<version>.*)").unwrap();
             let caps = re.captures(&contents)?;
             format_version(&caps["version"], config.version_format)
         }).or_else(|| {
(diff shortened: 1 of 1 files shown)


davidkna (approved):

andytom:
Thanks for your contribution @swpalmer and thank you for reviewing @davidkna.

PR #4765, opened 2022-12-27
feat: add pijul_channel module

IslandUsurper:
#### Description

Adds a module to detect [Pijul](https://pijul.org/) repos and display the channel name.

#### Motivation and Context

When a project uses Pijul as its VCS, it is nice to have some indication of it the way Git and Mercurial do. This module is basically a copy of the `hg_branch` module. It has the same settings and functionality, except there is no concept of bookmarks in Pijul as there is in Mercurial, so it gets only.

Screenshot

As a VCS integration module, the tests are marked as `#[ignore]` so that developers don't have to have Pijul installed to run tests. In addition to automatic unit tests, I have installed Starship modified with this module and switched channels on a Pijul repository, and the channel name was reflected on the command line.

#### Checklist:


IslandUsurper:
It seems [there are problems](https://nest.pijul.com/pijul/pijul/discussions/748) with building pijul from source, so I'll stop burning GH actions on this for now.

IslandUsurper:
Well, I don't have a Windows machine to test on, and I just learned about vcpkg last night, so I've probably done as much as I can here. I copied the workflow actions for Windows from https://github.com/boringcactus/pijul-windows-builds/ (linked from the official Pijul docs, despite being unofficial), so they _should_ work, but I probably missed something and can't see it.

src/modules/pijul.rs
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    let mut module = context.new_module("pijul");
+    let config: PijulConfig = PijulConfig::try_load(module.config);
+
+    // We default to disabled=true, so we have to check after loading our config module.
+    if config.disabled {
+        return None;
+    };
+
```
davidkna, line 28: For truncation, please follow the approach in the meson module, using the functions from `crate::utils::truncate`.

.github/workflows/workflow.yml
```diff
@@ -207,6 +207,38 @@ jobs:
...
+        run: |
+          export VCPKG_ROOT=$VCPKG_INSTALLATION_ROOT
+          export LIB=$VCPKG_ROOT/installed/x64-windows-static-md/lib
+          export OPENSSL_DIR=$VCPKG_ROOT/installed/x64-windows-static-md
+          cargo install pijul --version "~1.0.0-beta" --locked
+
+
+      - name: Setup | Pijul
+        if: matrix.os == 'macOS-latest' || matrix.os == 'ubuntu-latest'
+        run: cargo install pijul --version "~1.0.0-beta" --locked
```
davidkna, line 240: In think this much set-up code and compilation time isn't worth it to test this one module more thoroughly. Please either switch to using compiled binary or please remove testing with real binaries completely.

davidkna (changes requested):

roland-rollo:
> It seems [there are problems](https://nest.pijul.com/pijul/pijul/discussions/748) with building pijul from source, so I'll stop burning GH actions on this for now.

Do `cargo install pijul --version 1.0.0-beta.2 --locked`.

src/modules/pijul.rs
IslandUsurper: I thought there ought to have been a better way, but I didn't look much outside the hg_branch module to find it.

IslandUsurper:
I did use `--locked` in [5301faf](https://github.com/starship/starship/pull/4765/commits/5301faf7010d3287cd2b7f1d4b64c129ffc17902)7010d3287cd2b7f1d4b64c129ffc17902). :wink:

docs/config/README.md
```diff
@@ -3162,6 +3163,21 @@ By default the module will be shown if any of the following conditions are met:
...
+The `pijul` module shows the active channel of the repo in your current directory.
+
+### Options
+
+| Option              | Default                           | Description                                                                          |
+| ------------------- | --------------------------------- | ------------------------------------------------------------------------------------ |
+| `symbol`            | `' '`                            | The symbol used before the pijul channel name of the repo in your current directory. |
+| `style`             | `'bold purple'`                   | The style for the module.                                                            |
+| `format`            | `'on [$symbol$channel]($style) '` | The format for the module.                                                           |
+| `truncation_length` | `2^63 - 1`                        | Truncates the pijul channel name to `N` graphemes                                    |
```
davidkna, line 3177: To be more compatible with 32-bit systems:
```suggestion
| `truncation_length` | `2^32 - 1`                        | Truncates the pijul channel name to `N` graphemes                                    |
```

src/configs/pijul.rs
```diff
@@ -0,0 +1,30 @@
...
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+#[serde(default)]
+pub struct PijulConfig<'a> {
+    pub symbol: &'a str,
+    pub style: &'a str,
+    pub format: &'a str,
+    pub truncation_length: i64,
```
davidkna, line 14:
```suggestion
    pub truncation_length: u32,
```

src/configs/pijul.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub disabled: bool,
+}
+
+impl<'a> Default for PijulConfig<'a> {
+    fn default() -> Self {
+        PijulConfig {
+            symbol: " ",
+            style: "bold purple",
+            format: "on [$symbol$channel]($style) ",
+            truncation_length: std::i64::MAX,
```
davidkna, line 25:
```suggestion
            truncation_length: u32::MAX,
```

src/modules/mod.rs
```diff
@@ -273,6 +275,7 @@ pub fn description(module: &str) -> &'static str {
         "package" => "The package version of the current directory's project",
         "perl" => "The currently installed version of Perl",
         "php" => "The currently installed version of PHP",
+        "pijul" => "The current channel of the repo in the current directory",
```
davidkna, line 278: Maybe it should be called `pijul_branch` in case someone wants to do a `status` module in the future.

davidkna (changes requested):

docs/config/README.md
IslandUsurper: Both `git_branch` and `hg_branch` take `i64` for `truncation_length`. While I agree that this is an easy change for 32-bit compatibility, I think consistency is more important. All of the modules that use truncation need to be using `usize` instead of `i64`.

davidkna: Either way, I feel like an unsigned integer type should be preferred and meson also uses `u32`.

IslandUsurper: Sure, but let's do a follow-up change to handle all of the affected modules, instead of just polishing this one.

davidkna: I've avoided doing this, because it could be a breaking change. If for instance they are switched to `u32` the former default would be outside the accepted value range, and negative values would lead to errors instead of warnings.

davidkna (approved):

docs/config/README.md
IslandUsurper: At the risk of causing [XKCD 1172](https://xkcd.com/1172/), I doubt anything would _actually_ break. Defaults shouldn't be stored in any TOML files, and anyone who actually sets `truncation_length` for any module is most likely to set it less than 200. As for handling negative values, I wonder if there's a way to adjust the deserializer to emit warnings with defaults instead of hard errors. Possibly we could make the default `0` instead of `INT_MAX`, as at these sizes, there's no practical difference.

davidkna: I think some people will have based their config-file on `starship print-config --default`, it was even recommended in the docs for a few weeks.

IslandUsurper: Oof. That does make a mess of things then.

andytom:
Thank you for your contribution @IslandUsurper and thanks for reviewing @davidkna.

PR #4769, opened 2022-12-28
feat(aws): Handle SSO session for AWS

### LINKED ISSUE #4697: the AWS module doesn't handle SSO default profiles
#### Current Behavior
when using aws sso (so no env vars, no AWS_PROFILE or anything) 
aws module displays items from .aws/config [default] key if it exists (and displays nothing if it doesn't)
instead of the correct [profile default] or currently logged in [profile xxxxx] entry.

ie, having the following in .aws/config

[default] 
region = eu-west-1

[profile default]
region = us-west-2
sso_session = starship
sso_account_id = ...
sso_role_name = ...

[profile tokyo]
region = ap-northeast-1
sso_session = starship
sso_account_id = ...
sso_role_name = ...

[sso-session starship]
sso_start_url = https://xxxxxxxxxx.awsapps.com/start
sso_region = ...
sso_registration_scopes = ...

would always show `eu-west-1`, 
whereas the `aws configure get region` actually use and returns `us-west-2`

#### Expected Behavior
- it should display `default` and corresponding data if there is a `[profile default]`
- it should handle having both `[default]` and `[profile default]` in the `.aws/config` file
- it should display information consistent with `aws configure get region` in all cases.
- it should display either nothing, or the `expiration_symbol` (X), if the current profile (either `default` or `AWS_PROFILE`) has no valid token in `.aws/sso/cache`

#### Possible Solution
use [profile default] over [default] values when available and no `AWS_PROFILE` is defined...
see the .aws/sso/cache or other for token status

#### Environment
- Starship version: 1.11.0
- Shell type: zsh
- Shell version: zsh 5.8.1 (x86_64-apple-darwin22.0)
- Shell plugin manager: none
- Operating system: macOS 13.0.1 (22A400)

#### Starship Configuration
```toml
[aws]
force_display = true
disabled = false
```

alexeyshockov:
#### Description
Activate AWS module for profiles that uses SSO authentication.

#### Motivation and Context
Closes #4697

#### Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -170,6 +170,7 @@ fn has_credential_process_or_sso(
 
     Some(
         config_section.contains_key("credential_process")
+            || config_section.contains_key("sso_session")
             || config_section.contains_key("sso_start_url")
             || credential_section?.contains_key("credential_process")
             || credential_section?.contains_key("sso_start_url"),


src/modules/aws.rs
```diff
@@ -170,6 +170,7 @@ fn has_credential_process_or_sso(
 
     Some(
         config_section.contains_key("credential_process")
+            || config_section.contains_key("sso_session")
```
davidkna, line 173: Please add a test for this. Can the credentials file also contain this? If so, please also add it below.

jaklan:
Hi guys, any chance to push it forward?

raylu:
it looks like this was done by 297176b0b8b9da34176d7b278837f77f960799b1 (#5170)

PR #4771, opened 2022-12-28
feat(hg_branch): Add support for mercurial topics and find hg root dir

### LINKED ISSUE #870: hg_branch module doesn't show up when I am in subdirectory

### LINKED ISSUE #1384: [mercurial] support for topic (lightweight branches)

nicoe:
#### Description
This change introduce a new variable that store the active topic name and also it finds the root directory of the mercurial repo.

I am aware that there is already a similar PR : https://github.com/starship/starship/pull/3970 but I noticed it only after I wrote my code :smile:. Anyway I think the remark that @davidkna made on this PR can be fixed by using https://github.com/speedy37/mountpoints-rs/ but I don't know what is the policy regarding including libraries that are quite young and seldomly used.

#### Motivation and Context
`hg_branch` does not work when the directory is not at the root of the repository.
I introduced the topic in this change because we started using it in our project and it's a nice way to have lightweight branches in mercurial.

Closes #3970
Closes #1461 
Closes #1384
Closes #870

screenshot

I ran this version on my computer.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2638,14 +2638,14 @@ The `hg_branch` module shows the active branch and topic of the repo in your cur
 
 ### Options
 
-| Option              | Default                                 | Description                                                                                  |
-| ------------------- | --------------------------------------- | -------------------------------------------------------------------------------------------- |
-| `symbol`            | `' '`                                  | The symbol used before the hg bookmark or branch name of the repo in your current directory. |
-| `style`             | `'bold purple'`                         | The style for the module.                                                                    |
+| Option              | Default                                   | Description                                                                                  |
+| ------------------- | ----------------------------------------- | -------------------------------------------------------------------------------------------- |
+| `symbol`            | `' '`                                    | The symbol used before the hg bookmark or branch name of the repo in your current directory. |
+| `style`             | `'bold purple'`                           | The style for the module.                                                                    |
 | `format`            | `'on [$symbol$branch(:$topic)]($style) '` | The format for the module.                                                                   |
-| `truncation_length` | `2^63 - 1`                              | Truncates the hg branch / topic name to `N` graphemes                                        |
-| `truncation_symbol` | `'…'`                                   | The symbol used to indicate a branch name was truncated.                                     |
-| `disabled`          | `true`                                  | Disables the `hg_branch` module.                                                             |
+| `truncation_length` | `2^63 - 1`                                | Truncates the hg branch / topic name to `N` graphemes                                        |
+| `truncation_symbol` | `'…'`                                     | The symbol used to indicate a branch name was truncated.                                     |
+| `disabled`          | `true`                                    | Disables the `hg_branch` module.                                                             |
 
 ### Variables
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2634,26 +2634,27 @@ style = 'bold dimmed green'
 
 ## Mercurial Branch
 
-The `hg_branch` module shows the active branch of the repo in your current directory.
+The `hg_branch` module shows the active branch and topic of the repo in your current directory.
 
 ### Options
 
-| Option              | Default                          | Description                                                                                  |
-| ------------------- | -------------------------------- | -------------------------------------------------------------------------------------------- |
-| `symbol`            | `' '`                           | The symbol used before the hg bookmark or branch name of the repo in your current directory. |
-| `style`             | `'bold purple'`                  | The style for the module.                                                                    |
-| `format`            | `'on [$symbol$branch]($style) '` | The format for the module.                                                                   |
-| `truncation_length` | `2^63 - 1`                       | Truncates the hg branch name to `N` graphemes                                                |
-| `truncation_symbol` | `'…'`                            | The symbol used to indicate a branch name was truncated.                                     |
-| `disabled`          | `true`                           | Disables the `hg_branch` module.                                                             |
+| Option              | Default                                 | Description                                                                                  |
+| ------------------- | --------------------------------------- | -------------------------------------------------------------------------------------------- |
+| `symbol`            | `' '`                                  | The symbol used before the hg bookmark or branch name of the repo in your current directory. |
+| `style`             | `'bold purple'`                         | The style for the module.                                                                    |
+| `format`            | `'on [$symbol$branch:$topic]($style) '` | The format for the module.                                                                   |
+| `truncation_length` | `2^63 - 1`                              | Truncates the hg branch / topic name to `N` graphemes                                        |
+| `truncation_symbol` | `'…'`                                   | The symbol used to indicate a branch name was truncated.                                     |
+| `disabled`          | `true`                                  | Disables the `hg_branch` module.                                                             |
 
 ### Variables
 
-| Variable | Example  | Description                          |
-| -------- | -------- | ------------------------------------ |
-| branch   | `master` | The active mercurial branch          |
-| symbol   |          | Mirrors the value of option `symbol` |
-| style\*  |          | Mirrors the value of option `style`  |
+| Variable | Example   | Description                          |
+| -------- | --------- | ------------------------------------ |
+| branch   | `master`  | The active mercurial branch          |
+| topic    | `feature` | The active mercurial topic           |
+| symbol   |           | Mirrors the value of option `symbol` |
+| style\*  |           | Mirrors the value of option `style`  |
 
 *: This variable can only be used as a part of a style string
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2642,7 +2642,7 @@ The `hg_branch` module shows the active branch and topic of the repo in your cur
 | ------------------- | --------------------------------------- | -------------------------------------------------------------------------------------------- |
 | `symbol`            | `' '`                                  | The symbol used before the hg bookmark or branch name of the repo in your current directory. |
 | `style`             | `'bold purple'`                         | The style for the module.                                                                    |
-| `format`            | `'on [$symbol$branch:$topic]($style) '` | The format for the module.                                                                   |
+| `format`            | `'on [$symbol$branch(:$topic)]($style) '` | The format for the module.                                                                   |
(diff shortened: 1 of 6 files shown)


davidkna:
> Anyway I think the remark that @davidkna made on this PR can be fixed by using https://github.com/speedy37/mountpoints-rs/

I actually implemented both `GIT_CEILING_DIRS` and limiting discovery across file systems in gitoxide `git-discover`, you could try re-using some of the code. Getting the `device_id` of a directory is fairly simple without any extra libraries. https://github.com/Byron/gitoxide/blob/main/git-discover/src/upwards/util.rs#L34-L75

src/modules/hg_branch.rs
```diff
@@ -74,13 +95,38 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+
+fn get_hg_branch_name(hg_root: &PathBuf) -> String {
+    std::fs::read_to_string(hg_root.join(".hg").join("branch"))
         .map_or_else(|_| "default".to_string(), |s| s.trim().into())
 }
 
-fn get_hg_current_bookmark(ctx: &Context) -> Option<String> {
-    std::fs::read_to_string(ctx.current_dir.join(".hg").join("bookmarks.current"))
+fn get_hg_current_bookmark(hg_root: &PathBuf) -> Option<String> {
+    std::fs::read_to_string(hg_root.join(".hg").join("bookmarks.current"))
```
davidkna, line 123: Try using `crate::utils::read_file`

nicoe:
The topic test can only pass if the topic extension is installed for the user running the `hg` command of the test. It works on my box (but you have to trust me :stuck_out_tongue_closed_eyes:)

nicoe:
Also I don't get why the config-schema test fails while looking at the diff the config schema is updated …

davidkna:
> The topic test can only pass if the topic extension is installed for the user running the `hg` command of the test. It works on my box (but you have to trust me 😝)

Per the docs, it looks like it can be enabled via `.hgrc`? Would creating a temporary config file that enables it, and setting `HGRC` to its location work? Alternatively, please create the topic file directly.

docs/config/README.md
```diff
@@ -2634,26 +2634,27 @@ style = 'bold dimmed green'
...
-| `truncation_symbol` | `'…'`                            | The symbol used to indicate a branch name was truncated.                                     |
-| `disabled`          | `true`                           | Disables the `hg_branch` module.                                                             |
+| Option              | Default                                 | Description                                                                                  |
+| ------------------- | --------------------------------------- | -------------------------------------------------------------------------------------------- |
+| `symbol`            | `' '`                                  | The symbol used before the hg bookmark or branch name of the repo in your current directory. |
+| `style`             | `'bold purple'`                         | The style for the module.                                                                    |
+| `format`            | `'on [$symbol$branch:$topic]($style) '` | The format for the module.                                                                   |
```
davidkna, line 2645: The `$topic` may be unset:
```suggestion
| `format`            | `'on [$symbol$branch(:$topic)]($style) '` | The format for the module.                                                                   |
```

src/modules/utils/path.rs
```diff
@@ -100,6 +102,20 @@ impl PathExt for Path {
...
+
+    #[cfg(target_os = "linux")]
+    fn device_id(&self) -> u64 {
+        use std::os::linux::fs::MetadataExt;
+        let m = self.metadata().unwrap();
+        m.st_dev()
+    }
+
+    #[cfg(all(unix, not(target_os = "linux")))]
+    pub fn device_id(&self) -> u64 {
```
davidkna, line 114:
```suggestion
    fn device_id(&self) -> u64 {
```

src/configs/hg_branch.rs
```diff
@@ -21,7 +21,7 @@ impl<'a> Default for HgBranchConfig<'a> {
         HgBranchConfig {
             symbol: " ",
             style: "bold purple",
-            format: "on [$symbol$branch]($style) ",
+            format: "on [$symbol$branch:$topic]($style) ",
```
davidkna, line 24:
```suggestion
            format: "on [$symbol$branch(:$topic)]($style) ",
```

src/modules/hg_branch.rs
```diff
@@ -34,17 +46,27 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         config.truncation_length as usize
     };
 
-    let branch_name =
-        get_hg_current_bookmark(context).unwrap_or_else(|| get_hg_branch_name(context));
```
davidkna, line 38: I would prefer to keep using lazy evaluation here.

src/modules/hg_branch.rs
```diff
@@ -34,17 +46,27 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    let topic_graphemes = if let Some(topic) = &repo.topic {
+        let truncated_topic_graphemes = get_graphemes(topic, len);
+        if len < graphemes_len(topic) {
+            truncated_topic_graphemes + truncation_symbol.as_str()
+        } else {
+            truncated_topic_graphemes
+        }
+    } else {
+        String::from("")
+    };
```
davidkna, line 69: Since you are changing truncation quite a bit, please also try moving it to the `utils::truncate` functions, like the meson module.

davidkna (changes requested):

nicoe:
> Per the docs, it looks like it can be enabled via `.hgrc`? Would creating a temporary config file that enables it, and setting `HGRC` to its location work? Alternatively, please create the topic file directly.

It needs to install an additional python module, I'll just create the topic file directly then.

Thanks for the review!

src/modules/hg_branch.rs
```diff
@@ -74,26 +73,36 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     Some(module)
 }
 
-fn get_hg_branch_name(ctx: &Context) -> String {
-    std::fs::read_to_string(ctx.current_dir.join(".hg").join("branch"))
-        .map_or_else(|_| "default".to_string(), |s| s.trim().into())
+fn get_hg_repo_root<'a>(ctx: &'a Context) -> Result<&'a Path, Error> {
+    let dir = ctx.current_dir.as_path();
+    let dev_id = dir.device_id();
```
davidkna, line 78: You will need to sprinkle `#[cfg(unix)]` here or add a no-op for other targets.

nicoe: It won't compile under windows if `device_id` is not defined in the Path extension, so I guess a stub is the best way to go.

src/modules/hg_branch.rs
```diff
@@ -74,26 +73,36 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+    for root_dir in dir.ancestors() {
+        if dev_id != root_dir.device_id() {
+            break;
+        }
+        if root_dir
+            .read_dir()?
+            .any(|e| e.unwrap().file_name() == ".hg")
+        {
```
davidkna, line 86: Please avoid `unwrap`, please just use this:
```suggestion
        if root_dir.join(".hg").is_dir() {
```

src/modules/utils/path.rs
```diff
@@ -100,6 +107,20 @@ impl PathExt for Path {
...
+        let m = self.metadata().unwrap();
+        m.st_dev()
+    }
+
+    #[cfg(all(unix, not(target_os = "linux")))]
+    fn device_id(&self) -> u64 {
+        use std::os::unix::fs::MetadataExt;
+        let m = self.metadata().unwrap();
+        m.dev()
+    }
```
davidkna, line 123: Please avoid `unwrap`, instead please just return `Option<u64>` or maybe `unwrap_or(u64::MAX)`.

src/modules/utils/path.rs
```diff
@@ -80,6 +82,11 @@ impl PathExt for Path {
         let (_, path) = normalize::normalize_path(self);
         path
     }
+
+    fn device_id(&self) -> u64 {
+        // Maybe it should us unimplemented!
```
davidkna, line 87: Typo:
```suggestion
        // Maybe it should be unimplemented!
```

I wonder if [std::os::windows::fs::MetadataExt::volume_serial_number](https://doc.rust-lang.org/std/os/windows/fs/trait.MetadataExt.html#tymethod.volume_serial_number) will ever get stabilized

davidkna (changes requested):
I have a few more minor suggestions, but apart from that this PR looks great and ready to merge.

src/modules/utils/path.rs
nicoe: In fact I meant that it might use `unimplemented!` and do the `cfg` stuffs needed to still be usable under windows.

But indeed I looked at `volume_serial_number` and that's what we need but the last comment of the linked issue did not made me hopeful.

src/modules/hg_branch.rs
nicoe: BTW why should unwrap be avoided? I'm new to rust and I'd like to know why unwrap should be avoided (just like the other change in `device_id`).

davidkna (approved):
LGTM

andytom:
Thank you for your contribution @nicoe and thank you for reviewing @davidkna.

PR #4757, opened 2022-12-25
feat(config): update nushell config command usage from `is-empty` to `length`

niruhsa:
Update the Nushell `$HOME/.cache/starship/starship.nu` config file on line 19 to use `(not ($env | get -i config | length) == 0)` instead of `(not ($env | get -i config | is-empty))` since `is-empty` is no longer a valid external command in Nushell according to latest issues on the Nushell repo indicating that commands with `-` dashes in them no longer exist.

Got this error when creating a fresh install of Nushell & Starship. Version information is listen below.
```
❯ version
╭────────────────────┬─────────────────────────────────────╮
│ version            │ 0.61.0                              │
│ branch             │ master                              │
│ tag                │                                     │
│ build_os           │ macos-aarch64                       │
│ build_os           │ aarch64-apple-darwin                │
│ rust_version       │ rustc 1.59.0 (9d1b2106e 2022-02-23) │
│ rust_channel       │ stable-aarch64-apple-darwin         │
│ cargo_version      │ cargo 1.59.0 (49d8809dc 2022-02-10) │
│ pkg_version        │ 0.61.0                              │
│ build_time         │ 2022-12-25 09:10:29 +11:00          │
│ build_rust_channel │ release                             │
│ features           │ default, trash, which, zip          │
│ installed_plugins  │                                     │
╰────────────────────┴─────────────────────────────────────╯

```
#### Description
According to multiple issues & PRs in the Nushell repo, commands that feature a `-` dash are no longer valid commands, but the referencing documentation isn't updated to show these changes. Installing a fresh install of Nushell & starship, the check in the file `$HOME/.cache/starship/starship.nu` on line 19 uses `is-empty` which is invalid on new installs currently, instead I've updated it to use `(not ($env | get -i config | length) == 0)` which is valid in the latest release of Nushell.

#### Motivation and Context
What does it solve? Deprecation issue relating to Nushell command references in the Nushell configuration file for Starship.

#### How Has This Been Tested?
Tested & working once changing the `starship.nu` config file with the change, with the command `source $HOME/.cache/starship/starship.nu` no longer throwing an external command not found error.
- [X] I have tested using **MacOS**
- [X] I have tested using **Linux (Debian - WSL2)**

#### Checklist:
- [X] I have updated the documentation accordingly.
  - No documentation changes needed as far as I can tell
- [ ] I have updated the tests accordingly.
  - Unsure what tests would be required to change, if any, so leaving this for anyone that has a clue about it.

PR #4777, opened 2022-12-30
docs(nu): Add `-f` flag

### LINKED ISSUE #4701: initializing nushell with save
We just wanted to let you know that in our upcoming nushell release on Dec 20th, we have changed the way the nushell `save` command works. If people using nushell initialize starship with each instance like `starship init nu | save ~/.cache/starship/init.nu`, which overwrites the `init.nu` each time, this will break because our `save` command now requires `save -f` to "force" it to overwrite.

Thank you for supporting nushell in starship!

cassis163:
#### Description

I noticed that I had to edit an instruction to make the integration with NuShell work (see https://github.com/starship/starship/issues/4701). All that was needed was to add a '-f' flag, but the documentation did not include that. Without this flag, a file would not be overwritten, which causes an error on startup.

Closes #4701
Closes #4757

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ Add the following to the end of your Nushell env file (find it by running `$nu.e
 
 ```sh
 mkdir ~/.cache/starship
-starship init nu | save ~/.cache/starship/init.nu
+starship init nu | save -f ~/.cache/starship/init.nu
 ```
 
 And add the following to the end of your Nushell configuration (find it by running `$nu.config-path`):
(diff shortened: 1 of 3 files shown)


davidkna:
Thanks for the contribution.

Please do not update the translated files, they are updated automatically via crowdin and manual updates can lead to conflicts. Please also update the installation script (like in #3908). Because `-f` is not supported by older versions of nu, please also update the minimum supported nu version, and remove the fall-back for older versions introduced in #4490 and the `Note: Nushell 0.71.0 or later is required`.
In addition I would prefer putting the `-f` directly after the `save`.

davidkna (approved):

andytom:
Thank you for your contribution @cassis163 and thank you for reviewing @davidkna. I will open a separate issue for cleaning up the nu init script.

PR #4788, opened 2023-01-03
docs(presets): fix the scope of tip block

KaoruMuta:
#### Description

Here is a PR to change the scope of tip block in presets docs except in English.
This part is not able to be edited as far as I get touched in `clowdin` so I created PR (Sorry for inconvenience if we can fix easily via `clowdin`)

#### Motivation and Context

Currently, presets docs except in English have a different design from docs in English due to invalid usage of tip block. (Example is shown in screenshots)
I think the design except in English may cause the confusion as if `No Nerd Fonts` involves in the following presets such as `Bracketed Segments` to user.

|As Is|To Be|
|---|---|
|スクリーンショット 2023-01-03 14 02 02|スクリーンショット 2023-01-03 14 06 17|

#### Checklist:

### DIFF
--- a/docs/ar-SA/presets/README.md
+++ b/docs/ar-SA/presets/README.md
@@ -14,9 +14,7 @@ This preset changes the symbols for each module to use Nerd Font symbols.
 
 This preset changes the symbols for several modules so that no Nerd Font symbols are used anywhere in the prompt.
 
-::: tip
-This preset will become the default preset [in a future release of starship](https://github.com/starship/starship/pull/3544).
-:::
+::: tip This preset will become the default preset [in a future release of starship](https://github.com/starship/starship/pull/3544). :::
 
 [Click to view No Nerd Font preset](./no-nerd-font)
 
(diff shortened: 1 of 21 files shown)


matchai:
Hi @KaoruMuta 👋
Unfortunately, these files are automatically generated by Crowdin from the translations provided by the community.
If we change them like you have in the PR, the change will be undone by Crowdin automatically in this PR: #4719.

I think the safest thing to do would be to update the source file (`docs/presets/README.md`) to have extra whitespace, like this:
```diff
 ::: tip
+
 This preset will become the default preset [in a future release of starship](https://github.com/starship/starship/pull/3544).
+
 :::
```

Would you be

KaoruMuta:
>  If we change them like you have in the PR, the change will be undone by Crowdin automatically in its next automated PR: #4719.

Got it.

> I think the safest thing to do would be to update the source file (`docs/presets/README.md`) to have extra lines of whitespace, like this:
> 
> ```diff
>  ::: tip
> +
>  This preset will become the default preset [in a future release of starship](https://github.com/starship/starship/pull/3544).
> +
>  :::
> ```
> 
> How does that sound to you?

It looks good to me, so I revert previous change and reflect your suggestion.

matchai (approved):
LGTM! 👍

PR #4790, opened 2023-01-03
chore: Don't run crowdin workflow on forks

jankatins:
#### Description

I get a weekly email that the crowdin workflow has failed in my fork. This change should prevent that from happening by only running the task on the primary repo (starship/starship)


davidkna (approved):

davidkna:
Thanks for the fix @jankatins.

PR #4794, opened 2023-01-05
docs: use literal toml strings in `continuation_prompt` docs

Marzal:
#### Description
continuation_prompt should be set to a literal string so it should have ' instead of " in the doc/examples.

UPDATE:
Using " even without any variables doesn't work . -> not true, a shell refresh was needed

#### Motivation and Context

Closes #

- [ x] I have tested using **Linux**

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -287,7 +287,7 @@ Produces a prompt like the following:
 
 Some shells support a continuation prompt along with the normal prompt. This prompt is rendered instead of the normal prompt when the user has entered an incomplete statement (such as a single left parenthesis or quote).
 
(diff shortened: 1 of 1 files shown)


davidkna (approved):
Actually, both variants should work. Did you refresh your shell after setting the option?

Either way, changing this ensures that `'` strings are used more consistently, so thanks for the PR!

Marzal:
> Actually, both variants should work. Did you refresh your shell after setting the option?
> 
> Either way, changing this ensures that `'` strings are used more consistently, so thanks for the PR!

That's true, normally changes are applied without interaction, but closing and opening `konsole` did work with `"` as with `'`.

I

PR #4796, opened 2023-01-07
feat(php): Add `symfony-cli` support

### LINKED ISSUE #3759: PHP Version should use check more source to be determined
As a PHP developer I can use different version of PHP version between project (ignore php in docker container).
I use Homebrew to install different version, and the use the Symfony CLI to handle which version I want to use.
Also PHP version is (can be) set in composer.json

In project managed by the Symfony CLI, I setup version in a `.php-version` file => display this version, or better but can lead to timedout issue, use the Symfony CLI command to retrieve version: `symfony php -v`
In PHP project I specify the php version in composer.json => display the version available in `require.php`

image

I create an alias for php: `alias php="symfony php` but doesn't work as expected (placed in my `~/.zshrc` before the starship initialization)
image

Nsbx:
#### Description
I added the support of `symfony-cli` for `php` module.
With this new option when you work with a PPA and multiple version of php. The symfony-cli is used to get the current php version used in the current project and not the default version used by the system.

#### Motivation and Context
Like the author of the issue, I encountered a bad version displayed in a project.
Before I check the doc but nothings about PPA or symfony-cli.
So I checked the issues and see someone has the same.
I never coded in rust before but with the tests and codespace I successfully added the option.
Closes #3759

I coded the change in github codespace, i simply installed rust, made the change, write new tests, and build the project to test it live.
I added a PPA installed multiple version of PHP, installed symfony-cli, and define specific version inside .php-version file.
And the version displayed was not the same as the system default one

#### Checklist:


drupol:
Isn't this issue related to https://github.com/symfony-cli/phpstore/pull/10 ?

Nsbx:
Like my other PR (4837).
I forgot about it, and no time to update it.
Sorry.

PR #4806, opened 2023-01-10
feat(fossil_branch): add fossil_branch module

VegardSkui:
#### Description

Add a module to print the active branch of a [Fossil](https://www.fossil-scm.org/) check-out, if the current directory is a Fossil check-out. The module is analogous to the `git_branch` and `hg_branch` modules.

#### Motivation and Context

This is part of adding initial Fossil support to Starship, as this was currently missing. This lets users of Fossil repositories have some of the same functionality available in their check-outs as they already enjoy in their Git repositories. While Fossil users usually don't switch branches for a single directory as often as e.g. Git users, this module provides a nice indicator that the directory is version controlled and would work nicely together with a `fossil_metrics` module. The `fossil_metrics` module will be implemented in a follow-up pull request.

Contributes to the closing of #4036.

Manual tests were ran both with and without Fossil installed and in multiple check-outs. I have also tested the module on FreeBSD.

NA

#### Checklist:

### DIFF
--- a/src/modules/fossil_branch.rs
+++ b/src/modules/fossil_branch.rs
@@ -77,7 +77,6 @@ mod tests {
     use nu_ansi_term::{Color, Style};
 
     use crate::test::{fixture_repo, FixtureProvider, ModuleRenderer};
-    use crate::utils::create_command;
 
     enum Expect<'a> {
         BranchName(&'a str),
@@ -102,10 +101,9 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
     fn test_fossil_branch_disabled_per_default() -> io::Result<()> {
         let tempdir = fixture_repo(FixtureProvider::Fossil)?;
-        let checkout_dir = &tempdir.path().join("checkout");
+        let checkout_dir = tempdir.path();
         expect_fossil_branch_with_config(
             checkout_dir,
             Some(toml::toml! {
@@ -119,7 +117,6 @@ mod tests {
     }
 
     #[test]
-    #[ignore]
     fn test_fossil_branch_autodisabled() -> io::Result<()> {
         let tempdir = tempfile::tempdir()?;
         expect_fossil_branch_with_config(tempdir.path(), None, &[Expect::Empty]);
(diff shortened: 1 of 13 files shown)


davidkna (changes requested):
I haven't used fossil yet, but it looks like it would be reasonable to scan for `.fslckout` before running the external command. Please also add more tests, including a branch-available test.

VegardSkui:
Agreed. I've added more test in the style of `hg_branch` and `pijul_channel` now. I don't know which method is preferred so I first committed a set of tests which initializes a Fossil repository+checkout and calls out to the fossil binary, and then a commit which alters the tests to use command mocking. The 9c52cb1c600dda2cd42e094f4577bdfb334b4807-commit can be reverted if calling the binary is preferred.

src/test/mod.rs
```diff
@@ -167,13 +167,22 @@ impl<'a> ModuleRenderer<'a> {
...
 }
 
 pub fn fixture_repo(provider: FixtureProvider) -> io::Result<TempDir> {
     match provider {
+        FixtureProvider::Fossil => {
+            let path = tempfile::tempdir()?;
+            fs::OpenOptions::new()
+                .create(true)
+                .write(true)
+                .open(path.path().join(".fslckout"))?;
```
davidkna, line 183: To reduce intermittent test failures:
```suggestion
                .open(path.path().join(".fslckout"))?
                .sync_all()?;
```

src/modules/fossil_branch.rs
```diff
@@ -0,0 +1,215 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let is_checkout = context
+        .try_begin_scan()?
+        .set_files(&[".fslckout"])
+        .is_match();
+
+    if !is_checkout {
+        return None;
+    }
+
```
davidkna, line 19: Please move this after the `config.disabled` check, because it should be cheaper.

davidkna (changes requested):

src/modules/fossil_branch.rs
VegardSkui: Changed the order of the checks now. This should probably be changed in the `pijul_channel` module as well to keep consistency between the modules.

davidkna (approved):
LGTM

davidkna:
Thanks for the contribution @VegardSkui!

PR #4819, opened 2023-01-13
feat(profiles): add right prompt, set via env & add docs

### LINKED ISSUE #4786: transient rprompt often not working in Fish
#### Current Behavior
I run the following code to not mix my configuration for both fish and startship:
```shell
$ sh -c 'env HOME=$(mktemp -d) fish'                                                      3.9.13
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
jonathf@kadesh /h/jonathf> mv .config/starship.toml{,_}
jonathf@kadesh /h/jonathf> function starship_transient_prompt_func
                               starship module character
                           end
jonathf@kadesh /h/jonathf> function starship_transient_rprompt_func
                                 starship module time
                           end
                           starship init fish | source
                           enable_transience
❯ echo 123
123

/home/jonathf via 🐍 v3.9.13 (.39)
❯
```

#### Expected Behavior
The line with `echo 123` should have a timestamp on the right.
As seen on the top line, rprompt works, just not in transient.

#### Additional context

There is something weird going on.
I can't get either prompt nor rprompt to work, but I seem to be able to get both to work as if I use no-config `fish -N` oposed to the default config `sh -c 'env HOME=$(mktemp -d) fish'`. The latter is the recommended approach to test by fish as far as I can tell.

#### Possible Solution
No idea.

#### Environment
- Starship version:
```
starship 1.12.0
branch:
commit_hash:
build_time:2022-12-13 20:48:05 +01:00
build_env:rustc 1.65.0 (897e37553 2022-11-02),stable-x86_64-unknown-linux-gnu
```
- Shell type: fish
- Shell version: fish, version 3.5.1
- Shell plugin manager: None.
- Operating system: Ubuntu 22.10.

Code above is run without configuration

### LINKED ISSUE #4812: Document `--profile` feature
#3467 added support for the `--profile` flag but was merged without corresponding documentation.

Request: documentation for this feature (the code is fairly short) along with a few examples of how it could be used.

### LINKED ISSUE #4953: Regression: right_prompt includes content from prompt
#### Current Behavior
After upgrading to 1.13.1, my RPROMPT includes the content of PROMPT. Content which should be in RPROMPT does appear as it should (e.g. Python version, git info), but it is then wrapped by my PROMPT content.

This started showing up after a `brew upgrade`; I've not touched my shell or starship configurations.

#### Expected Behavior
RPROMPT only includes the content it should, as before.

Screenshot showing the prompt is duplicated to the right, as well as other details from my configuration. I created a minimal Starship config file based on my main one to reproduce.

Screenshot 2023-03-01 at 14 10 22

Screenshot showing the directory name and "character" wrapping the RPROMPT.

Screenshot 2023-03-01 at 14 16 30

#### Possible Solution
None yet.

#### Environment
- Starship version: 1.13.1 (installed from homebrew)
- Shell type: zsh
- Shell version: zsh 5.9 (arm-apple-darwin21.3.0)
- Shell plugin manager: n/a
- Terminal emulator: iTerm2 and wezterm; inside and outside tmux
- Operating system: macOS 13.2.1

This function is called late in my `.zshrc`; it is a holdover from when I was running a custom build of starship. Now it just uses the installed one.

```zsh
launch_starship() {
	local _latest_starship
	local _starship_time

	if [[ -x $HOME/src/starship/target/release/starship ]]; then
		_latest_starship=$HOME/src/starship/target/release/starship
		_starship_time=$( stat -f %m "$_latest_starship" )
	fi

	if [[ -x $HOME/src/starship/target/debug/starship ]]; then
		if [[ -n $_latest_starship ]]; then
			if [[ $( stat -f %m $HOME/src/starship/target/debug/starship ) -gt $_starship_time ]]; then
				_load_debug "debug starship is the newer binary"
				_latest_starship=$HOME/src/starship/target/debug/starship
			else
				_load_debug "release starship is the newer binary"
			fi
		else
			_latest_starship=$HOME/src/starship/target/debug/starship
		fi
	fi

	if [[ -z $_latest_starship ]]; then
		if has starship; then
			_load_debug "using installed starship"
			eval $( starship init zsh )
		fi
	else
		_load_debug "using starship at $_latest_starship"
		eval $( $_latest_starship init zsh)
	fi
}
```

#### Starship Configuration

```toml
add_newline = false
right_format = "$all"
```

davidkna:
#### Description

First in addtion to `profiles.name = "format-string"` also allow defining profiles via a table that include a `right_format` string:
```toml
[profiles.name]
format = "a"
right_format = "b"
```
Thus, `--profile` becomes usable together with `--right` (and `--continuation`), and because it has moved `starship explain` will work with `--profile`.

To properly allow switching between profiles, I also made the `profile` settable via the `STARSHIP_PROFILE` env var, by enabling the `clap` `env` feature.

I also amended the transient-prompt docs a bit to make use of a `transient` feature and included such a profile in the default config (and fix #4786).

#### Motivation and Context

Closes #4786
Closes #4812
~~Closes #4953~~
Required to allow #4208 to be used to define transient rprompts(as supported by #4160)

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -210,16 +210,17 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Option            | Default                        | Description                                                                                                                                                                        |
-| ----------------- | ------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `format`          | [link](#default-prompt-format) | Configure the format of the prompt.                                                                                                                                                |
-| `right_format`    | `''`                           | See [Enable Right Prompt](../advanced-config/#enable-right-prompt)                                                                                                                 |
-| `scan_timeout`    | `30`                           | Timeout for starship to scan files (in milliseconds).                                                                                                                              |
-| `command_timeout` | `500`                          | Timeout for commands executed by starship (in milliseconds).                                                                                                                       |
-| `add_newline`     | `true`                         | Inserts blank line between shell prompts.                                                                                                                                          |
-| `palette`         | `''`                           | Sets which color palette from `palettes` to use.                                                                                                                                   |
-| `palettes`        | `{}`                           | Collection of color palettes that assign [colors](../advanced-config/#style-strings) to user-defined names. Note that color palettes cannot reference their own color definitions. |
-| `follow_symlinks` | `true`                         | Follows symlinks to check if they're directories; used in modules such as git.                                                                                                     |
+| Option            | Default                                   | Description                                                                                                                                                                        |
+| ----------------- | ----------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `format`          | [link](#default-prompt-format)            | Configure the format of the prompt.                                                                                                                                                |
+| `right_format`    | `''`                                      | See [Enable Right Prompt](../advanced-config/#enable-right-prompt)                                                                                                                 |
+| `scan_timeout`    | `30`                                      | Timeout for starship to scan files (in milliseconds).                                                                                                                              |
+| `command_timeout` | `500`                                     | Timeout for commands executed by starship (in milliseconds).                                                                                                                       |
+| `add_newline`     | `true`                                    | Inserts blank line between shell prompts.                                                                                                                                          |
+| `palette`         | `''`                                      | Sets which color palette from `palettes` to use.                                                                                                                                   |
+| `palettes`        | `{}`                                      | Collection of color palettes that assign [colors](../advanced-config/#style-strings) to user-defined names. Note that color palettes cannot reference their own color definitions. |
+| `follow_symlinks` | `true`                                    | Follows symlinks to check if they're directories; used in modules such as git.                                                                                                     |
+| `profiles`        | `{ transient: { format: "$character" } }` | Collection of prompt strings that can be used instead of the main `format` and `right_format`. For more information please consult the profile section below.                      |
 
 ::: tip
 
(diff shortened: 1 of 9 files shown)


docs/config/README.md
```diff
@@ -349,6 +350,29 @@ modules you explicitly add to the format will not be duplicated. Eg.
...
+[profiles]
+# Instead of a map you can also set a profile to a simple string instead
+# Starship includes the definition below for use with transient prompts
+transient = "$character"
+
+[profiles.as_map_with_right_format]
+format = "$character"
+right_format = "$git_branch"
+```
+
```
mickimnet, line 375: Hello David,

While reading issue #4828, I was contemplating, where to add further documentation about the `[palettes]` option. In my opinion, the `Prompt` section of the _Configuration_ documentation is already quite extensive. Therefore, I'll add it to the _Advanced Documentation_ section.

Following my above reasoning, I would suggest to also put the _Profiles_ part to the _Advanced Configuration_ section.

Just my 2 cents
Mick

davidkna: I feel like the _Advanced Configuration_ docs aren't very visible, which is why I put it in the main config docs, but I will consider moving it.

jamesoff:
Happy to test a patch :)

docs/config/README.md
mickimnet: @davidkna, should the new `[profiles]` section already work in the v1.13.1 release, and you just add the documentation?

docs/config/README.md
```diff
@@ -350,6 +351,29 @@ modules you explicitly add to the format will not be duplicated. Eg.
 format = '$all$directory$character'
 ```
 
+### Profiles
+
+Profiles can be used to set a key-value-map with alternate versions of the top-level `format` (the main prompt format string) and `right_format` (the same for the `right_prompt`). To quickly switch to a different profile of the current prompt, you can set the `STARSHIP_PROFILE` environment variable to the name of the profile, for instance by running `export STARSHIP_PROMPT=my_profile` or `$ENV:STARSHIP_PROMPT="my_profile"` in PowerShell. Alternatively, it can also be set via an `--profile=my_profile` flag for commands like `starship prompt` and `starship explain`.
```
mickimnet, line 356: If I read `src/config.rs` correctly, the variable is `STARSHIP_PROFILE` and **not** `STARSHIP_PROMPT`. Is this true, or did I miss something?

davidkna: Thanks, I've fixed this.

docs/config/README.md
Ch4ni: Apologies for weighing in late. I am on mobile, but wanted to capture a thought that I can't currently confirm.

I agree @mickimnet, the advanced configuration docs are rather extensive, and generally amazing. To @davidkna's point, the whole reason for https://github.com/starship/starship/issues/4828 was the lack of concept linking between the color/pallette feature in the main docs and the advanced configuration.

We could likely satisfy both desires (to keep advanced topics where they are, but provide visibility) by including links to the advanced configuration topics in the basic documentation.

As a general practice, I think that this improves UX, indicates that desired capabilities already exist, and also implicitly let's users know that they're venturing into advanced territory - it also acts as an `apropos` of sorts to connect the advanced configuration to its (more searchable/navigable) basic counterparts.

TL;DR - if we leave the advanced config where it is and just link basic config to further reading in "advanced config" we may empower users who lack the intuition/context on where to find the info they need - without additional human intervention.

PR #4826, opened 2023-01-16
fix(fish): enable transient prompt when in vi mode

accidentaldevelopment:
#### Description
Fix prompt transience in `fish` if using vi binding.

#### Motivation and Context
The current `enable_transience` function only works when using emacs/default bindings. This is because a command (`transient_execute`) is bound to `\r`. However with vi binding enabled, the bindings are modal. This PR updates `enable_transience` to also bind `\r` in `insert` mode.

This also updates the `disable_transience` function to erase the new bindings rather than redefine them back to `execute`. This will cause the default "preset" bindings to be effective again. Not strictly required to get transience working in vi mode, but it is closer to what is trying to be accomplished. It also doesn't depend on fish keeping `execute` as the default command.

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -70,12 +70,18 @@ function transient_execute
     commandline -f execute
 end
 
-function enable_transience
-    bind \r transient_execute
+# --user is the default, but listed anyway to make it explicit.
+function enable_transience --description 'enable transient prompt keybindings'
+    bind --user \r transient_execute
+    bind --user -M insert \r transient_execute
 end
 
-function disable_transience
-    bind \r execute
+# Erase the transient prompt related key bindings.
+# --user is the default, but listed anyway to make it explicit.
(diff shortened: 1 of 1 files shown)


davidkna (approved):
I am not very experienced with fish internals, so perhaps somebody else should also have a look at this, but this appears to work and looks good to me.

accidentaldevelopment:
Thanks @davidkna! I did test it and all was good, but I'm obviously biased 😛. Let me know if I can provide any help, but otherwise I'll just be patient :).

rashil2000 (approved):
LGTM, but I do feel the `--user` should be omitted. Since it's the default, it's just adding noise.

davidkna:
Thanks for the contribution @accidentaldevelopment, and thanks for the follow-up review @rashil2000!

PR #4829, opened 2023-01-17
feat: add typechange to git_status module

cdenyar:
#### Description
Adds tracking/counting of `typechange` changes in `git_status`.

#### Motivation and Context

Displays any unstaged typechanges on the prompt and includes any staged typechanges as `is_staged` to avoid surprises on `git commit`.

image

Using the following `git_status` config to include the count for testing:
```
[git_status]
typechange = '[⇢\($count\)](bold green)'
format = '([\[$all_status$ahead_behind\]]($style) )'
```

1.  Create an empty repo with two files committed
2. Replace one of the two files with a symlink to the other file
3. Confirm the the prompt displayed the unstaged typechange as expected
4. Staged the symlink with `git add`
5. Confirmed the staged change displayed as a normal staged change/addition

- [x] I have tested using **MacOS** (`test result: ok. 941 passed; 0 failed; 32 ignored; 0 measured; 0 filtered out; finished in 8.80s`)

#### Checklist:

- [x] I have run `cargo fmt` and `dprint fmt`

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1823,6 +1823,7 @@ You can disable the module or use the `windows_starship` option to use a Windows
 | `staged`            | `'+'`                                         | The format of `staged`                                                                                      |
 | `renamed`           | `'»'`                                         | The format of `renamed`                                                                                     |
 | `deleted`           | `'✘'`                                         | The format of `deleted`                                                                                     |
+| `typechange`        | `'⇢'`                                         | The format of `typechange`                                                                                  |
 | `style`             | `'bold red'`                                  | The style for the module.                                                                                   |
 | `ignore_submodules` | `false`                                       | Ignore changes to submodules.                                                                               |
 | `disabled`          | `false`                                       | Disables the `git_status` module.                                                                           |
@@ -1843,6 +1844,7 @@ The following variables can be used in `format`:
 | `staged`       | Displays `staged` when a new file has been added to the staging area.                                         |
 | `renamed`      | Displays `renamed` when a renamed file has been added to the staging area.                                    |
 | `deleted`      | Displays `deleted` when a file's deletion has been added to the staging area.                                 |
+| `typechange`   | Displays `typechange` when a file's type has been changed in the staging area.                                |
 | style\*        | Mirrors the value of option `style`                                                                           |
 
 *: This variable can only be used as a part of a style string
--- a/src/configs/git_status.rs
+++ b/src/configs/git_status.rs
@@ -21,6 +21,7 @@ pub struct GitStatusConfig<'a> {
     pub modified: &'a str,
     pub staged: &'a str,
     pub untracked: &'a str,
+    pub typechange: &'a str,
     pub ignore_submodules: bool,
     pub disabled: bool,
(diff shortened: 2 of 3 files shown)


src/configs/git_status.rs
```diff
@@ -43,6 +44,7 @@ impl<'a> Default for GitStatusConfig<'a> {
             modified: "!",
             staged: "+",
             untracked: "?",
+            typechange: "⇢",
```
davidkna, line 47: For consistency, I think I would prefer calling this `typechanged`. Please also disable it by default by replacing the default to an empty string or `None` to avoid confusing users with additional symbols.

src/modules/git_status.rs
```diff
@@ -275,7 +284,13 @@ impl RepoStatus {
 
     fn is_staged(short_status: &str) -> bool {
         // is_index_modified || is_index_added
```
davidkna, line 286: Please update this comment.

src/modules/git_status.rs
```diff
@@ -275,7 +284,13 @@ impl RepoStatus {
...
-        short_status.starts_with('M') || short_status.starts_with('A')
+        short_status.starts_with('M')
+            || short_status.starts_with('A')
+            || short_status.starts_with('T')
```
davidkna, line 289: Please add a test for this condition.

cdenyar: added test `shows_staged_typechange_with_count()`

docs/config/README.md
```diff
@@ -1860,6 +1860,7 @@ You can disable the module or use the `windows_starship` option to use a Windows
 | `staged`            | `'+'`                                         | The format of `staged`                                                                                      |
 | `renamed`           | `'»'`                                         | The format of `renamed`                                                                                     |
 | `deleted`           | `'✘'`                                         | The format of `deleted`                                                                                     |
+| `typechanged`       |                                               | The format of `typechange`                                                                                  |
```
davidkna, line 1863: The changes LGTM, but please list the true default in the table:
```suggestion
| `typechanged`       | `""`                                          | The format of `typechange`                                                                                  |
```

davidkna (approved):

davidkna:
Thanks for the contribution @cdenyar!

docs/config/README.md
```diff
@@ -1860,6 +1860,7 @@ You can disable the module or use the `windows_starship` option to use a Windows
 | `staged`            | `'+'`                                         | The format of `staged`                                                                                      |
 | `renamed`           | `'»'`                                         | The format of `renamed`                                                                                     |
 | `deleted`           | `'✘'`                                         | The format of `deleted`                                                                                     |
+| `typechanged`       | `""`                                          | The format of `typechange`                                                                                  |
```
adamchainz, line 1863: So typechanges aren't shown by default? That's a bit of a shame to me, since the reason I reported it was I found it surprising that starship wasn't summarizing something that I had to commit by default.

(It also looks like this would be better off using a single-quoted string for consistency)

adamchainz (commented):
Thanks for this feature!

PR #4831, opened 2023-01-17
perf:  Use `elixir --short-version` for better performance

### LINKED ISSUE #4825: Read the "elixir" shell script for its version number before running it

Address the long spin-up time to get Elixir's version number, caused by spinning up the Erlang VM.

When inspecting my Elixir installation, I noticed that `elixir` in my `PATH` was a shell script with its version number explicitly defined near the top of the file:

```sh
#!/bin/sh
set -e

ELIXIR_VERSION=1.14.3
```

On my machine, it looks like this is the case for both my nix and Homebrew installations of Elixir. I'd be interested in hearing if it's the same for Windows users as well, but at the very least, this is a way we can eliminate all spin-up time on Unix-based OSes.

aselder:
#### Description
This change the elixir version helper to use `elixir --short-version`. This command is about 5x faster than `elixir --version`.

#### Motivation and Context

Often then existing elixir module times out trying to get the version. This speeds it up considerably.

Closes #4825 

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -410,7 +410,7 @@
         ],
         "detect_folders": [],
         "disabled": false,
-        "format": "via [$symbol($version \\(OTP $otp_version\\) )]($style)",
+        "format": "via [$symbol($version )]($style)",
         "style": "bold purple",
         "symbol": "💧 ",
         "version_format": "v${raw}"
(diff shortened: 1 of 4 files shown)


matchai (commented):
It looks like this would be a functional regression, removing support for OTP version. This would necessitate a breaking change.
I would elect to continue using `--version` if the `$otp_version` variable is being used, otherwise using `--short-version`.

aselder:
I can see where you're coming from, but I'll leave this hear for other to pick up, as my Rust isn't good enough to handle that logic :)

PR #4853, opened 2023-01-25
build(deps): update toml crates

davidkna:
#### Description

This PR updates the TOML crates. With this release, they should be using the same parser implementation.

`toml::toml!` was changed to return a `toml::Table` instead of a `toml::Value` containing one. I replaced many of the `toml::Value` uses with `toml::Table`.

To still allow for constructing `Value::ValueDeserializer` with both `&Table` and `&toml::Value`, I added a `ValueRef` type, because `&Value` can't be constructed from a `&Table`.

#### Motivation and Context

Closes #

#### Checklist:


matchai (approved):
Nice! LGTM 👍

davidkna:

PR #4857, opened 2023-01-26
ci: use `reviewdog/action-suggester` for config-schema check

davidkna:
#### Description

Instead of `git` commands, this will use the reviewdog action-suggester action to give feedback on required config changes to try it out as a replacement for the `actions-rs` `cargo fmt` action.

#### Motivation and Context

Related #4725

#### Checklist:

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -137,6 +137,10 @@ jobs:
     name: Check if config schema is up to date
     needs: cargo_check
     runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      contents: read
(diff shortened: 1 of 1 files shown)


andytom:
The plugin is complaining that it doesn't have [write access to the review api ](https://github.com/starship/starship/actions/runs/4027009108/jobs/6932048922#step:6:220) is this something we need to worry about?

andytom (approved):
LGTM just one question about the api key.

davidkna:
I've tried to increase the token permissions to the expected level, but as far as I can tell this will only slightly change the way the check results are displayed.

PR #4863, opened 2023-01-27
feat(cmd_duration): add the possibility to format the duration

### LINKED ISSUE #3034: Change the format from "$duration" in the module "cmd_duration"
Normally, when a command takes some amount of minutes, seconds, ... the prompt is like this:
`<dir> took 1m46s`

But there is no way to change it for example to
`<dir> took 1m 46s`
or
`<dir> took 1 Minute and 46 Seconds`

Maybe there is a format string where there are some values fixed and the rest can be modded.
`$hours, $minutes, $seconds, $milliseconds  # fixed values`
The first example would look like this:
`duration_format = '[$hours]h [$minutes]m [$seconds]s [$milliseconds]ms'`

There is no alternative solution to this feature.

m19c:
#### Description

This pull request introduces the possibility to easily format the elapsed time provided by cmd_duration.
Previously it was only possible to display the time in the form of `$duration` (e.g. `1s37ms`). 

The following variables have been introduced:
- `dd` to represent days preceded by 0
- `d` to represent days
- `hh` to represent hours with a preceding 0
- `h` to represent hours
- `mm` to represent minutes with a preceding 0
- `m` to represent minutes
- `ss` to represent seconds preceded by a 0
- `s` to represent seconds

The previous configuration `show_milliseconds` is still available, but is used exclusively for the already existing variable `$duration`.

#### Motivation and Context

Closes #3034

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -335,11 +335,11 @@ impl<'a> Context<'a> {
     }
 
     // TODO: This should be used directly by clap parse
-    pub fn get_cmd_duration(&self) -> Option<u128> {
+    pub fn get_cmd_duration(&self) -> Option<u64> {
         self.properties
             .cmd_duration
             .as_deref()
-            .and_then(|cd| cd.parse::<u128>().ok())
+            .and_then(|cd| cd.parse::<u64>().ok())
     }
 
     /// Execute a command and return the output on stdout and stderr if successful
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -8,9 +8,9 @@ use once_cell::unsync::OnceCell;
 
 use super::{Context, Module, ModuleConfig};
 
+use super::utils::duration::Duration;
 use crate::configs::aws::AwsConfig;
 use crate::formatter::StringFormatter;
-use crate::utils::render_time;
 
 type Profile = String;
 type Region = String;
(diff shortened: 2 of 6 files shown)


davidkna:
I think this feature should include some way to allow mapping `0` time values to `None` like #1444.

m19c:
Is this really necessary? I am just thinking about a use case.

Assume the format `${hh}:${mm}` is used. In this case the user expects hours and minutes. To remove an element like `${hh}` (e.g. `12` instead of `00:12`) at this point would mean that the desired format is not satisfied, right? I personally do not know this kind of behavior, at least not in this form.

An alternative would be to introduce absolute values in the form of `${mmmm}` (or something like that) to not just use the leftover. But also here I am unsure ¯\_(ツ)_/¯ 

Anyway, if it really makes sense I'll gladly upgrade the pull request. Although, even if necessary, it's not exactly nice to just move the values around as is the case in #1444 (ms to s, s to m and so on). 
Maybe at this point it would be nice to teach the `StringFormatter` to delete elements enclosed in certain characters if they do not output a value (e.g. `?${hh}?`). But even that would not be enough for this particular problem.

davidkna:
>  I am just thinking about a use case.

For instance, in order to keep the existing format, but split the different unit-values with spaces.

jeaye:
Just found this since I was looking for a way to put a space between `9s603ms`. It's pretty tough to read that way, compared to `9s 603ms`. ForthesamereasonthatIwouldn'ttypelikethis.

barrettruth:
Please!! Love this PR. Yet another reason to push this forward: this is the ONLY difference between [pure-prompt](https://github.com/sindresorhus/pure) and starship left until they can be configured identically!!!

PR #4874, opened 2023-01-31
feat(fossil_metrics): add fossil_metrics module

VegardSkui:
#### Description

Add a module to show the count of currently added and deleted lines in a [Fossil](https://www.fossil-scm.org/) check-out in the current directory. The module is analogous to the `git_metrics` module.

#### Motivation and Context

This is part of adding initial Fossil support to Starship. This lets users of Fossil repositories have some of the same functionality available in their check-outs as they already enjoy in their Git repositories. Initial support for Fossil was included in #4806, which added the `fossil_branch` module.

Contributes to the closing of #4036.

NA

Manual tests were ran both with and without Fossil installed and in multiple check-outs. I have also tested the module on FreeBSD.

#### Checklist:


src/modules/fossil_metrics.rs
```diff
@@ -0,0 +1,267 @@
...
+    deleted: &'a str,
+}
+
+impl<'a> FossilDiff<'a> {
+    fn try_parse(diff_numstat: &'a str, only_nonzero_diffs: bool) -> Option<Self> {
+        let mut split = match diff_numstat.lines().last() {
+            Some(line) => line.split_whitespace(),
+            None => {
+                log::warn!("Error in module `fossil_metrics`:\nLast line of numstat diff missing.");
+                return None;
```
davidkna, line 70: ``Error in module `fossil_metrics`:\n`` is a bit redundant, and the duplicate log message detector may have trouble with this. I would prefer having `try_parse` return a `Result<Self, &str>` instead. Please also add tests for some of the different error conditions.

davidkna: I didn't mean to say that you should be removing all error handling in the module, I think that is nice to have. The ``Error in module `fossil_metrics`:\n``-prefix was redundant, because the log message will already contain `WARN` and the module name.

VegardSkui: Okay, I see. I tried to move closer to the implementation of `git_metrics` where an unexpected output leads to 0 changes reported, e.g. if the format of the command output changes in an update or the `git` binary is not actually Git. The different error conditions explicitly checked for previously should not be able to occur if a Fossil version later than v2.3 (2017-07-21) is installed (which introduced the `--numstat` option). I can revert the latest commit and fix the log messages if preferable, but I'm uncertain about what other input to test when it cannot be produced by Fossil.

VegardSkui:
A possible error condition is if the user has installed a version of Fossil >=v2.3 (2017-07-21), but older than v2.14 (2021-01-20). The final line with the total changes was introduced in v2.14, such that this module would then incorrectly report only the changes in whatever file is listed last. My initial thinking was that checking the installed version of Fossil on each run would be too much overhead in Starship (especially since the number of users of these versions is vanishingly small, for reference none of the supported Debian repos ships any of these versions as far as I can see, and those users likely won't install a new version of Starship either), though the format of the last line could be checked before possibly looping over the earlier output and do the summation manually.

davidkna:
Please create another PR and try to have it share its code with the hg upwards search. You can put it into one of the `utils` modules or `context.rs`.

davidkna:
> A possible error condition is if....

Please document this in the module docs.

src/modules/fossil_metrics.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        let added = match split.next() {
+            Some("0") | None => zero,
+            Some(s) => s,
+        };
+
+        let deleted = match split.next() {
+            Some("0") | None => zero,
+            Some(s) => s,
+        };
```
davidkna, line 91: Since this operation isn't supported on some versions, perhaps try sanity-check the returned values.

VegardSkui: Solved by using a regex to only match on the expected format.

davidkna (changes requested):
This looks good to me if you document the range of supported fossil versions.

davidkna (approved):

andytom:
Thank you for your contribution @VegardSkui and thanks for reviewing @davidkna.

PR #4881, opened 2023-02-02
chore: test git's `GIT_CEILING_DIRECTORIES`

### LINKED ISSUE #1996: Starship git prompts should respect `GIT_CEILING_DIRECTORIES`.
#### Current Behavior
If I have `GIT_CEILING_DIRECTORIES=$HOME/src` set in my environment, and I run `starship prompt` in `$HOME/src`,
it includes the branch status for a repo in `$HOME`.

#### Expected Behavior
I expect that `starship prompt` won't include any git status.

#### Possible Solution
It looks like `git_repository_discover` (which underlies `Repository::discover`) has an option that takes the value of that variable (though the rust wrapper doesn't currently expose it). Alternatively, `Repository::open_ext(..., RepositoryOpenFlags::FROM_ENV)` might have the expected behavior.

#### Environment
- Starship version: 0.44.0

Nerixyz:
#### Description

This PR adds a test for `GIT_CEILING_DIRECTORIES` - the handling for this on Windows will be fixed in the next gitoxide release (https://github.com/Byron/gitoxide/pull/718).

#### Motivation and Context

Closes #1996

#### Checklist:

  _I don't think this should be explicitly mentioned since it can be assumed, that starship respects this variable._

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -280,7 +280,7 @@ impl<'a> Context<'a> {
                 let shared_repo =
                     match ThreadSafeRepository::discover_with_environment_overrides_opts(
                         &self.current_dir,
-                        Default::default(),
+                        git::discover::upwards::Options::default().apply_environment(),
                         git_open_opts_map,
                     ) {
                         Ok(repo) => repo,
--- a/src/context.rs
+++ b/src/context.rs
@@ -280,7 +280,7 @@ impl<'a> Context<'a> {
                 let shared_repo =
                     match ThreadSafeRepository::discover_with_environment_overrides_opts(
                         &self.current_dir,
-                        git::discover::upwards::Options::default().apply_environment(),
+                        Default::default(),
                         git_open_opts_map,
                     ) {
                         Ok(repo) => repo,
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -109,6 +109,7 @@ mod tests {
         std::fs::create_dir(&level1)?;
         std::fs::create_dir(&level2)?;
         std::fs::create_dir(&level3)?;
+        let level1 = dunce::canonicalize(level1)?;
 
         // Set ceiling directory
         const CEILING_DIRECTORY_VAR: &str = "GIT_CEILING_DIRECTORIES";
@@ -121,13 +122,15 @@ mod tests {
             }
             None => OsString::new(),
         };
-        let parent_dir = repo_dir
-            .path()
-            .parent()
-            .expect("tempdir should always have a parent");
-        ceiling.push(parent_dir);
+        let parent_dir = dunce::canonicalize(
+            repo_dir
+                .path()
+                .parent()
+                .expect("tempdir should always have a parent"),
+        )?;
+        ceiling.push(&parent_dir);
         ceiling.push(CEILING_DIRECTORY_SEP);
-        ceiling.push(parent_dir);
+        ceiling.push(&parent_dir);
         ceiling.push(CEILING_DIRECTORY_SEP);
         ceiling.push(&level1);
 
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -130,6 +130,8 @@ mod tests {
         ceiling.push(parent_dir);
         ceiling.push(CEILING_DIRECTORY_SEP);
         ceiling.push(&level1);
+
+        #[allow(clippy::disallowed_methods)]
         env::set_var(CEILING_DIRECTORY_VAR, ceiling);
 
         let assert_discovery = |cwd: &Path| -> io::Result<()> {
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -99,6 +99,38 @@ mod tests {
     }
 
     #[test]
+    fn test_ceiling_directory() -> io::Result<()> {
+        let repo_dir = fixture_repo(FixtureProvider::Git)?;
+        let level1 = repo_dir.path().join("level1");
+        let level2 = level1.join("level2");
+        std::fs::create_dir(&level1)?;
+        std::fs::create_dir(&level2)?;
+
+        const CEILING_DIRECTORY_VAR: &str = "GIT_CEILING_DIRECTORIES";
+
+        let prev_ceiling = std::env::var_os(CEILING_DIRECTORY_VAR);
+        std::env::set_var(CEILING_DIRECTORY_VAR, level1.as_os_str());
+
+        let actual = ModuleRenderer::new("git_commit")
+            .config(toml::toml! {
+                [git_commit]
+                    only_detached = false
+            })
+            .path(level2)
+            .collect();
+
+        let expected = None;
+
+        match prev_ceiling {
+            Some(ceiling) => std::env::set_var(CEILING_DIRECTORY_VAR, ceiling),
+            None => std::env::remove_var(CEILING_DIRECTORY_VAR),
+        };
+
+        assert_eq!(expected, actual);
+        repo_dir.close()
+    }
+
+    #[test]
     fn test_render_commit_hash() -> io::Result<()> {
         let repo_dir = fixture_repo(FixtureProvider::Git)?;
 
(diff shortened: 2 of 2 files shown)


src/modules/git_commit.rs
```diff
@@ -98,6 +98,38 @@ mod tests {
...
+        let repo_dir = fixture_repo(FixtureProvider::Git)?;
+        let level1 = repo_dir.path().join("level1");
+        let level2 = level1.join("level2");
+        std::fs::create_dir(&level1)?;
+        std::fs::create_dir(&level2)?;
+
+        const CEILING_DIRECTORY_VAR: &str = "GIT_CEILING_DIRECTORIES";
+
+        let prev_ceiling = std::env::var_os(CEILING_DIRECTORY_VAR);
+        std::env::set_var(CEILING_DIRECTORY_VAR, level1.as_os_str());
```
Nerixyz, line 112: I'm not really sure how else I should do this, since gitoxide [uses `std::env::var_os`](https://github.com/Byron/gitoxide/blob/78517a616b4b1b6fefc2367057a9b3b786d93750/git-discover/src/upwards/types.rs#L81-L87).

src/context.rs
```diff
@@ -280,7 +280,7 @@ impl<'a> Context<'a> {
                 let shared_repo =
                     match ThreadSafeRepository::discover_with_environment_overrides_opts(
```
davidkna, line 281: Per the docs, `discover_with_environment_overrides_opts` should already apply `GIT_CEILING_DIRECTORIES`. At a glance, it may not actually be doing this right now. Please try fixing the behavior in gitoxide first (or align the docs with the reality).

src/modules/git_commit.rs
davidkna, line 112: I would rather not add any test that changes the environment, because it can cause various side effects, but as far as I can tell this particular instance might be fine. Please just add `#[clippy(allow)]`.

davidkna (changes requested):

src/context.rs
Nerixyz: My bad, it actually reads the environment variables [here](https://github.com/Byron/gitoxide/blob/78517a616b4b1b6fefc2367057a9b3b786d93750/git-repository/src/discover.rs#L84), so this PR effectively only adds a test.

src/modules/git_commit.rs
```diff
@@ -98,6 +98,40 @@ mod tests {
...
+        let level1 = repo_dir.path().join("level1");
+        let level2 = level1.join("level2");
+        std::fs::create_dir(&level1)?;
+        std::fs::create_dir(&level2)?;
+
+        const CEILING_DIRECTORY_VAR: &str = "GIT_CEILING_DIRECTORIES";
+
+        let prev_ceiling = std::env::var_os(CEILING_DIRECTORY_VAR);
+        #[allow(clippy::disallowed_methods)]
+        std::env::set_var(CEILING_DIRECTORY_VAR, level1.as_os_str());
```
davidkna, line 113: I think it might better to append the env value instead of replacing and restoring it, maybe even adding it twice to test your upstream PR (thanks for fixing the upstream bug I introduced!).
Beyond that, shouldn't the tests already pass on Windows? You are only adding a single path (so there is no platform-specific delimiter), there might be more issues here.

Nerixyz: > Beyond that, shouldn't the tests already pass on Windows?

No, since the paths on Windows include the Drive letter (e.g. `C:\Users\User` → `["C", "\Users\User"]`).

> You are only adding a single path (so there is no platform-specific delimiter), there might be more issues here.

I could add more test-cases. If the tests are running concurrently, this could introduce flaky tests.

> I think it might better to append the env value instead of replacing and restoring it, maybe even adding it twice to test your upstream PR (thanks for fixing the upstream bug I introduced!).

Wouldn't that make the tests dependent on the outside environment?

davidkna: Ok, that sounds fine.

> Wouldn't that make the tests dependent on the outside environment?

They were already included by the environment to begin with, and only changing the environment once feels less disruptive and makes any potential impact more observable.

Nerixyz: Hmm, I can't really get the appending to work.

<details>
<summary>My current approach</summary>

```diff
diff --git a/src/modules/git_commit.rs b/src/modules/git_commit.rs
index f9a18bc2..ada24996 100644
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -98,6 +98,22 @@ mod tests {
         repo_dir.close()
     }
 
+    #[allow(clippy::disallowed_methods)]
+    fn append_ceiling_directory(directory: &std::ffi::OsStr) {
+        const CEILING_DIRECTORY_VAR: &str = "GIT_CEILING_DIRECTORIES";
+
+        let ceiling: std::borrow::Cow<std::ffi::OsStr> =
+            match std::env::var_os(CEILING_DIRECTORY_VAR) {
+                Some(mut var) => {
+                    var.push(if cfg!(windows) { ";" } else { ":" });
+                    var.push(directory);
+                    var.into()
+                }
+                None => directory.into(),
+            };
+        std::env::set_var(CEILING_DIRECTORY_VAR, ceiling);
+    }
+
     #[test]
     fn test_ceiling_directory() -> io::Result<()> {
         let repo_dir = fixture_repo(FixtureProvider::Git)?;
@@ -106,11 +122,8 @@ mod tests {
         std::fs::create_dir(&level1)?;
         std::fs::create_dir(&level2)?;
 
-        const CEILING_DIRECTORY_VAR: &str = "GIT_CEILING_DIRECTORIES";
-
-        let prev_ceiling = std::env::var_os(CEILING_DIRECTORY_VAR);
-        #[allow(clippy::disallowed_methods)]
-        std::env::set_var(CEILING_DIRECTORY_VAR, level1.as_os_str());
+        append_ceiling_directory(level1.as_os_str());
+        append_ceiling_directory(level1.as_os_str());
 
         let actual = ModuleRenderer::new("git_commit")
             .config(toml::toml! {
@@ -122,11 +135,40 @@ mod tests {
 
         let expected = None;
 
-        match prev_ceiling {
-            #[allow(clippy::disallowed_methods)]
-            Some(ceiling) => std::env::set_var(CEILING_DIRECTORY_VAR, ceiling),
-            None => std::env::remove_var(CEILING_DIRECTORY_VAR),
-        };
+        assert_eq!(expected, actual);
+        repo_dir.close()
+    }
+
+    #[test]
+    fn test_ceiling_directory_in_current_dir() -> io::Result<()> {
+        let repo_dir = fixture_repo(FixtureProvider::Git)?;
+        let level1 = repo_dir.path().join("level1");
+        let level2 = level1.join("level2");
+        std::fs::create_dir(&level1)?;
+        std::fs::create_dir(&level2)?;
+
+        append_ceiling_directory(repo_dir.path().as_os_str());
+
+        let mut git_output = create_command("git")?
+            .args(["rev-parse", "HEAD"])
+            .current_dir(repo_dir.path())
+            .output()?
+            .stdout;
+        git_output.truncate(7);
+        let expected_hash = str::from_utf8(&git_output).unwrap();
+
+        let actual = ModuleRenderer::new("git_commit")
+            .config(toml::toml! {
+                [git_commit]
+                    only_detached = false
+            })
+            .path(repo_dir.path())
+            .collect();
+
+        let expected = Some(format!(
+            "{} ",
+            Color::Green.bold().paint(format!("({expected_hash})"))
+        ));
 
         assert_eq!(expected, actual);
         repo_dir.close()
```
</details>

After setting the ceiling directory to `C:\Users\User\AppData\Local\Temp\.tmpaSiHpn`, no other git repository can be discovered - even if the tempdir has a different name (e.g. `C:\Users\Johannes\AppData\Local\Temp\.tmpDHMU0f`), when running `cargo test --lib -- modules::git_commit::tests::test_render_commit_hash modules::git_commit::tests::test_ceiling_directory_in_current_dir --exact --nocapture `

davidkna: I think that's a different issue (Byron/gitoxide#721). Still, I would prefer only setting the variable once to avoid any possible race-conditions between reading and writing the environment variable.

Nerixyz: You mean there should be a single test that sets the variable and tests multiple things with it?

davidkna: Yes.

Nerixyz:
I have no clue why the test only fails mac.

davidkna:
> I have no clue why the test only fails mac.

It doesn't work because the path is canocialized, you need to pre-canocialize or add an empty entry to the env var.

davidkna:
Closing as stale.

PR #4892, opened 2023-02-08
ci: reduce renovate PR noise

matchai:
#### Description

Enables ["Branch Automerging"](https://docs.renovatebot.com/noise-reduction/#branch-automerging) in Renovate.
Configure Renovate to try performing version updates and status checks on a branch rather than in a PR, only opening a PR if the status checks fail. Passing version bumps will be committed directly to `master`.

#### Motivation and Context

Reduces the amount of PR noise that we get from Renovate, which doesn't require intervention.

#### Checklist:

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -9,6 +9,7 @@
         "workarounds:all",
         // automerge minor updates
         ":automergeMinor",
+        ":automergeBranch"
     ],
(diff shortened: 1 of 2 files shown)


davidkna:
I think we should also set `rebaseWhen` to `conflicted` to avoid running CI all the time.

PR #4893, opened 2023-02-09
fix: let-env warning when using nushell

mslot:
#### Description

The nushell warns about not using `let-env` for configuring `config`. I have removed the `load-env` and just added a `let-env` for the `config` as the documentation recommends: https://www.nushell.sh/book/environment.html#let-env.

#### Motivation and Context

The fix solves a warning that is visible when opening nushell with starship. I haven't seen any issues  opened explaining this.

I haven't contributed to anything of this size before, and my motivation of getting nushell and starship was to try to help out on a couple of rust projects. I actually just compiled and set up starship and nushell, to get started using these tools, when i saw this error. I am therefore _not_ a nushell expert, so i have spent a couple of days reading up on nushell, before trying out this fix.

It would be advised to get a nushell expert to look this through, because i am new to Nushell and the Nu programming language. I feel like the `let-env` is a new contstruct to ease the loading of environment variables, because the other solution to the problem would be to use `load-env` with a predefined `let config` variable, that was used here.

Warning looks like this:
image

When applying my fix, ithe warning goes away
image

I have tested this using:
1. Fedora 37 workstation edition - 64 bit
2. Gnome 43.2
3. Surfacebook 2 - with modded kernel
4. Starship 1.12.0 - parent commit hash 93c7eca5 - build from master
5. Nushell 0.75.1
6. Alacritty 0.12.0-dev (2d27fff)

#### Checklist:

### DIFF
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -15,12 +15,11 @@ let-env PROMPT_COMMAND = {
 # Whether we have config items
 let has_config_items = (not ($env | get -i config | is-empty))
 
-let config = if $has_config_items {
+let-env config = if $has_config_items {
     $env.config | upsert render_right_prompt_on_last_line true
 } else {
     {render_right_prompt_on_last_line: true}
 }
-{config: $config} | load-env
 
 let-env PROMPT_COMMAND_RIGHT = {
     let width = (term size).columns


davidkna:
Thanks for the fix @mslot.

muzhawir:
In Nushell v0.103.0, this warning has appeared again. According to the [documentation](https://www.nushell.sh/commands/docs/let-env.html), `let-env FOO = ...` has been removed — we should use `$env.FOO = ...` instead.

Screenshot_20250423_050025

davidkna:
@muzhawir Please ensure that you are using the latest version of starship. `let-env` is not used anymore.

muzhawir:
Thanks! yeah that’s right, my distro tends to update nushell faster than starship

PR #4902, opened 2023-02-13
feat(bash): Support right prompt and transience

rashil2000:
#### Description

Just like Clink for Cmd, [Ble.sh](https://github.com/akinomyoga/ble.sh) is an extension framework for Bash that does a lot of fun, amazing and frankly magical things. In the context of Starship, it can add right prompt, transient left prompt and transient right prompt to Bash. This PR tells user how to configure these. As evident, it does not affect non-Ble.sh users in any way.

Detailed documentation for transience can be found in Ble.sh's wiki: https://github.com/akinomyoga/ble.sh/wiki/Manual-%C2%A74-Editing

#### Motivation and Context

Relates to #888 

##### Right Prompt

image

##### Transience

image

#### Checklist:


akinomyoga (commented):
I came from https://github.com/akinomyoga/ble.sh/issues/239#issuecomment-1428565807. Here are my comments as the developer of ble.sh.

Also, ble.sh has PREEXEC & POSTEXEC hooks. I haven't tested it but a suggested change is

```diff
-if [[ "${__bp_imported:-}" == "defined" || $preexec_functions || $precmd_functions ]]; then
+if [[ ${BLE_VERSION-} && _ble_version -ge 400 ]]; then
+  blehook PREEXEC!='starship_preexec "$_"'
+  blehook POSTEXEC!='starship_precmd'
+elif [[ "${__bp_imported:-}" == "defined" || $preexec_functions || $precmd_functions ]]; then
```

src/init/starship.bash
```diff
@@ -58,8 +58,14 @@ starship_precmd() {
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
         PS1="$(::STARSHIP:: prompt --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="$NUM_JOBS" --cmd-duration=$STARSHIP_DURATION)"
         unset STARSHIP_START_TIME
+        if [[ ${BLE_VERSION-} ]]; then
```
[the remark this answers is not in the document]
rashil2000: Made the change

src/init/starship.bash
```diff
@@ -58,8 +58,14 @@ starship_precmd() {
...
+            bleopt prompt_rps1="$nlns$(::STARSHIP:: prompt --right --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="$NUM_JOBS" --cmd-duration=$STARSHIP_DURATION)"
+        fi
     else
         PS1="$(::STARSHIP:: prompt --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="$NUM_JOBS")"
+        if [[ ${BLE_VERSION-} ]]; then
+            bleopt prompt_rps1="$nlns$(::STARSHIP:: prompt --right --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="$NUM_JOBS")"
```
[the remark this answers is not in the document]
rashil2000: Done!

src/init/starship.bash
```diff
@@ -58,8 +58,14 @@ starship_precmd() {
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
         PS1="$(::STARSHIP:: prompt --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="$NUM_JOBS" --cmd-duration=$STARSHIP_DURATION)"
         unset STARSHIP_START_TIME
+        if [[ ${BLE_ATTACHED-} ]]; then
+            bleopt prompt_rps1="${PS1//[!$'\n']}$(::STARSHIP:: prompt --right --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="$NUM_JOBS" --cmd-duration=$STARSHIP_DURATION)"
```
akinomyoga, line 62: Thanks for the updates! Could you check here again?

When the user sets `shopt -u extquote`, `"${PS1//[!$'\n']}"` (which is quoted inside `"..."`) fails to properly remove the characters with Bash <= 5.1 (I think Bash-5.2's behavior is a bug). It leaves `$`, `'`, `\`, and `n` and removes everything other than them.

An easy way is to use variable assignment such as `nlns=${...}` because the right-hand side of the assignments is not subject to pathname expansions and word splitting so does not need to be quoted (unless it contains the shell special delimiters like `;|&<>`).

```suggestion
            local nlns=${PS1//[!$'\n']}
            bleopt prompt_rps1="$nlns$(::STARSHIP:: prompt --right --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="$NUM_JOBS" --cmd-duration=$STARSHIP_DURATION)"
```

If you don't like it, maybe there is another approach, e.g., to save and restore the state of `shopt -u extquote`, etc.

rashil2000: Sure, no problem, I have used the `nlns` variable now.

rashil2000:
> Also, ble.sh has PREEXEC & POSTEXEC hooks. I haven't tested it but a suggested change is
> 
> ```diff
> -if [[ "${__bp_imported:-}" == "defined" || $preexec_functions || $precmd_functions ]]; then
> +if [[ ${BLE_VERSION-} && _ble_version -ge 400 ]]; then
> +  blehook PREEXEC!='starship_preexec "$_"'
> +  blehook POSTEXEC!='starship_precmd'
> +elif [[ "${__bp_imported:-}" == "defined" || $preexec_functions || $precmd_functions ]]; then
> ```

I tried using this, but with this change, Starship's prompt does not get displayed at all. The default PS1 is visible.

akinomyoga:
Ah, sorry `starship_precmd` was not postexec but precmd. Could you see if it works with the following additional change?

```diff
-  blehook POSTEXEC!='starship_precmd'
+  blehook PRECMD!='starship_precmd'
```

akinomyoga:
Thanks!

akinomyoga (approved):

davidkna (changes requested):
Sorry for only testing this now.

In general, I am happy with this, but the way the starship arguments are handled needs to be changed. `${STARSHIP_PIPE_STATUS[*]}` has to be quoted for argument parsing not to error and other arguments like `${NUM_JOBS}` can also contain spaces in some cases, so the quoted arguments need to remain quoted.

src/init/starship.bash
```diff
@@ -52,21 +52,30 @@ starship_precmd() {
 
     eval "$_PRESERVED_PROMPT_COMMAND"
 
+    local ARGS="--terminal-width=${COLUMNS} --status=${STARSHIP_CMD_STATUS} --pipestatus=${STARSHIP_PIPE_STATUS[*]} --jobs=${NUM_JOBS}"
```
akinomyoga, line 55:
```suggestion
    local -a ARGS=(--terminal-width="${COLUMNS}" --status="${STARSHIP_CMD_STATUS}" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="${NUM_JOBS}")
```

I'd suggest turning `ARGS` into an array. The later `${ARGS}` should be replaced by `"${ARGS[@]}"`. Also, `ARGS+=...` should be `ARGS+=(...)` with proper quoting.

rashil2000: Hey, thanks for the suggestions - I have added them.

davidkna (changes requested):
I wanted to test if the changes worked, but I discovered that `$PIPESTATUS` is not forwarded to starship when using ble.sh.

akinomyoga:
That is described in README. I recommend referencing `BLE_PIPESTATUS` inside ble.sh. If you really need to use the name `PIPESTATUS`, you can use the setting `bleopt exec_restore_pipestatus=1`.

src/init/starship.bash
```diff
@@ -32,7 +32,11 @@ starship_preexec() {
 # Will be run before the prompt is drawn
 starship_precmd() {
     # Save the status, because commands in this pipeline will change $?
-    STARSHIP_CMD_STATUS=$? STARSHIP_PIPE_STATUS=(${PIPESTATUS[@]})
+    if [[ ${BLE_ATTACHED-} ]]; then
+        STARSHIP_CMD_STATUS=$? STARSHIP_PIPE_STATUS=(${BLE_PIPESTATUS[@]})
```
rashil2000, line 36: @akinomyoga is this approach fine?

akinomyoga: It doesn't work because running `[[ ... ]]` washes out both `$?` and `"${PIPESTATUS[@]}"` You need to store those values in the very first command of the precmd function (as in the current master) and then run `STARSHIP_PIPE_STATUS=("${BLE_PIPESTATUS[@]}")` in the case of ble.sh.

src/init/starship.bash
```diff
@@ -33,6 +33,9 @@ starship_preexec() {
 starship_precmd() {
     # Save the status, because commands in this pipeline will change $?
     STARSHIP_CMD_STATUS=$? STARSHIP_PIPE_STATUS=(${PIPESTATUS[@]})
+    if [[ ${BLE_ATTACHED-} ]]; then
```
davidkna, line 36: Maybe for consistency this would be more appropriate? 
```suggestion
    if [[ "${#BLE_PIPESTATUS[@]}" -gt "${#STARSHIP_PIPE_STATUS[@]}" ]]; then
```

davidkna (approved):
LGTM

src/init/starship.bash
akinomyoga: I'd say it doesn't work.

Inside ble.sh, `PIPESTATUS` contains a pipe status of random commands, so its content is unpredictable. With the current implementation of ble.sh, `PIPESTATUS` seems to be single-valued within the hook functions, but that can change in the future. Even if `PIPESTATUS` is ensured to be single-valued, the above condition doesn't work when the real pipe status of the previous user command is also single-valued (which is the majority of the cases).

Also, as far as the hook functions are called by ble.sh, `BLE_PIPESTATUS` is ensured to exist. Ah, but wait. An old version of ble.sh may not have the support. I've now checked the git blame, and `BLE_PIPESTATUS` was added two years ago. If we would care about it, I think the check could be

```suggestion
    if [[ ${BLE_ATTACHED-} && ${#BLE_PIPESTATUS[@]} -gt 0 ]]; then
```

rashil2000: Added

davidkna:
Thanks for the contribution @rashil2000, and thanks for all the help @akinomyoga!

PR #4905, opened 2023-02-15
docs(kubernetes): Remove extra backspace from regex in example

fenech:
In the example, `[\\w-]` would match a literal backspace `\`, the character `w` or a dash `-`. By removing the backspace, instead it matches any "word character" `\w` or a dash `-`.

#### Motivation and Context

Closes #

After removing the extra `\`, I could use a similar regular expression to match the context name of a kubernetes cluster in my local environment.

#### Checklist:


maxsxu (approved):
Thanks for this fix.

Tested and verified in macOS.

davidkna (approved):

davidkna:
Thanks for the fix @fenech, and thanks for the extra review @maxsxu!

PR #4912, opened 2023-02-17
fix: handle variable bash $SHLVL behavior with explicit option

### LINKED ISSUE #2407: $SHLVL seems to be off by one on bash
#### Current Behavior
The `$SHLVL` component seems to be off by one when using `bash` (`fish` displays the level correctly). When I use bash, if my `threshold` is set to 2, I have to go to down to `$SHLVL` 3 in order to get the prompt to show 2.

```sh
❯ bash

~
❯ bash

↴2 ~
❯ echo $SHLVL
3
```

#### Expected Behavior
Correct `$SHLVL` to be displayed regardless of shell.

#### Environment
- Starship version: 0.50.0
- fish version: fish, version 3.2.0
- Operating system: Mac OS 10.14.6
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.49.0
- Rust channel:  release
- Build Time: 2021-02-02 20:11:21

#### Starship Configuration

```toml
[directory]
# Use shortened path names
fish_style_pwd_dir_length = 1

[hostname]
ssh_only = true
format = " on [$hostname]($style) in "

[username]
# show_always = true
format = "[$user]($style)"

[status]
disabled = false
format = "[$symbol $status]($style) "

[aws]
disabled = true

[shlvl]
disabled = false
threshold = 2
symbol = "↴"

```

wlonkly:
#### Description
In #2407 we discovered that bash has two different ways of handling `$SHLVL` within a `$()` depending on version. In bash 3.x or 4.x.

```
$ echo $SHLVL; X=$( perl -e 'print $ENV{"SHLVL"}' ); echo $X
1
1
```

but in bash 5.x: 

```
$ echo $SHLVL; X=$( perl -e 'print $ENV{"SHLVL"}' ); echo $X
1
0

```

Is this a bash bug? Who knows, but it only happens when the command in the `$()` requires a fork/exec... like `starship prompt` does in `starship.bash`. 

So rather than introduce some complicated logic to handle bash versions, and rather than have a conversation about what bash versions to support -- we know that `$SHLVL` is always correct _outside_ the `$()`, so let's just pass it in via a `--shlvl` option, just like we pass in `--jobs`. 

The `--shlvl` option is optional, so that the startup scripts for all the other shells can be left untouched. 

#### Motivation and Context

Closes #2407

I didn't write a new test for this as I wasn't sure how to represent the source of `shlvl` in the test. But as for interactive testing, I used this config:

```
format = "$shlvl"

[shlvl]
disabled = false
format = "$shlvl"
```

and tested the following cases:

```
# $SHLVL set, no `--shlvl` option (non-bash shells)
SHLVL=100 STARSHIP_CONFIG=shlvl.toml target/debug/starship prompt

100

# $SHLVL set, with `--shlvl` option (bash shells)
SHLVL=100 STARSHIP_CONFIG=shlvl.toml target/debug/starship prompt --shlvl 200

200

# No $SHLVL, with `--shlvl` option (edge case) 
env STARSHIP_CONFIG=shlvl.toml target/debug/starship prompt --shlvl 200

200

# No $SHLVL, no `--shlvl` option (edge case) 
env STARSHIP_CONFIG=shlvl.toml target/debug/starship prompt
[no output]
```

Also I verified that with this fix and the updated `starship.bash`, the original mis-counting bug reported in #2407 (minus the `fish` digression!) is fixed. 

#### Checklist:

Since this is an internal-to-starship option I didn't update any documentation; happy to take advice on testing.

**I should mention that this is the first Rust code I have ever written so please review  it carefully!**


zanna-37:
@davidkna I'm sorry to bother you, but merging this PR would be amazing. Any chance this can be reviewed? It has been here for 6 months now, and it's pretty small.

davidkna (commented):
I've made two suggestions that should make this more reliable. As far as the implementation goes, I think it would make sense to use the `clap` `env` feature here, but I since this PR has been open for a long time, let's go with this approach for now.

zanna-37:
@wlonkly Sorry to bother you, any chances to make the modifications and get this merged?

wlonkly:
@zanna-37 Had forgotten all about this, on it!

wlonkly:
@davidkna as a rust newbie I could use some help with the failing tests, if you don't mind? Otherwise ready to go.

wlonkly:
(Ah, I see the -nightly builds are failing on all PRs. Phew.)

davidkna (approved):

davidkna:
Thanks for the contribution [wlonkly](https://github.com/wlonkly).

PR #4913, opened 2023-02-17
chore: use updated gitoxide crate names

davidkna:
#### Description

`git-repository` was renamed to `gix` and the other `git-*` crates were renamed to `gix-*`.

#### Motivation and Context

Closes #

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #4915, opened 2023-02-19
upgrade from `git-repository` 0.35 to `gix` 0.36

Byron:
The major change here is merely in name, allowing for better usability when 'use'ing `gitoxide` types. These are now accessible under the `gix` namespace, with all plumbing crates naturally being prefixed with `gix-`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -867,75 +867,118 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


davidkna:
Thanks for the pr. I already made the #4913 PR for this, but I pulled in your changes there to include your more extensive code changes.

PR #4918, opened 2023-02-21
Add powerpc64le release

rohanmclure:
Modern distros for powerpc are all little endian, and so this is the most common ppc subarchitecture.
Provide a binary release accordingly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,6 +56,10 @@ jobs:
             os: ubuntu-latest
             name: starship-arm-unknown-linux-musleabihf.tar.gz
 
+          - target: powerpc64le-unknown-linux-musl
+            os: ubuntu-latest
+            name: starship-powerpc64le-unknown-linux-musl.tar.gz
+
           - target: x86_64-apple-darwin
             os: macOS-11
             name: starship-x86_64-apple-darwin.tar.gz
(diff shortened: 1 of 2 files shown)


davidkna (changes requested):
As far as I can tell, this currently does not build with `cross`. In addition, this would need a change in the `install.sh` script.

rohanmclure:
Changed it to build for `linux-musl` abi, and added arch parsing and an assert to say that the target is supported to install.sh. Are we able to cross compile now?

davidkna:
There is no `powerpc64le-unknown-linux-musl` target in rust. Please make at least some effort to test your pull requests.

rohanmclure:
Fair enough. `rustc --print target-list` displays support for both `libc` and `musl`, but in actual fact powerpc is regarded as a tier two platform, so only standalone binaries for powerpc compilers, you may not be able to download `rust-std` for cross compilation. I was getting a build on my power box but understand this is tricker when cross-compiling. I'll see what modifications are needed in interfacing with `cross`.
https://doc.rust-lang.org/nightly/rustc/platform-support.html

davidkna:
Are you still working on this, or can this PR be closed?

davidkna:
I'm closing this PR, but please feel free to pick up work again if you want to continue working on this.

PR #4941, opened 2023-02-27
docs(install): fix typo and update indentation and whitespaces

hezhizhen:
#### Description

fix typo in `install/` and update indentation (2 spaces) and other whitespaces

#### Motivation and Context

Closes #

#### Checklist:


chipbuster:
Appreciate the typo fixes.

Are the whitespace fixes here done with some sort of automated tool? If so, could you share what the tool is and the invocation to use it? Would be useful if we ever break it in the future and have to fix it.

hezhizhen:
> Are the whitespace fixes here done with some sort of automated tool? If so, could you share what the tool is and the invocation to use it? Would be useful if we ever break it in the future and have to fix it.

Yes, and I use VS Code to do that. The extension I use to format shell scripts is https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format .

chipbuster:
Okay, it looks like it uses `shfmt` to do that, so we should have something reproducible in the future. Thanks!

chipbuster (approved):

hezhizhen:
@chipbuster I need some help to pass the failed test.

chipbuster:
@hezhizhen Something funky is going on with the tests. I can't seem to reproduce this error locally, but I get *other* errors locally, and it doesn't make sense for this change to have caused these tests to fail.

My guess is that there was a bug from a previous release that's being exposed here, but I'm not 100% sure what's going on---either way, that's not something you should have to worry about. I'll either work out what happened or just merge this as-is eventually.

hezhizhen:
@chipbuster I rebased `master` and pushed it to trigger tests. Hope all tests will pass.

Btw, do you mind if I add a spell checker https://github.com/crate-ci/typos to your github workflow for every pull request?

chipbuster:
Well, that works.

As far as `typos` goes, I wouldn't be opposed, but I'd want agreement from at least one other maintainer. @davidkna @andytom @matchai What are our thoughts about adding a typo-detection tool to our CI? (It seems that it's designed to work on a list of common typos instead of functioning as a spellcheck, so in principle it seems it shouldn't have many false positives).

hezhizhen:
> Well, that works.
> 
> As far as `typos` goes, I wouldn't be opposed, but I'd want agreement from at least one other maintainer. @davidkna @andytom @matchai What are our thoughts about adding a typo-detection tool to our CI? (It seems that it's designed to work on a list of common typos instead of functioning as a spellcheck, so in principle it seems it shouldn't have many false positives).

Move the discussion to a new issue and then merge it?

andytom:
> Well, that works.
> 
> As far as `typos` goes, I wouldn't be opposed, but I'd want agreement from at least one other maintainer. @davidkna @andytom @matchai What are our thoughts about adding a typo-detection tool to our CI? (It seems that it's designed to work on a list of common typos instead of functioning as a spellcheck, so in principle it seems it shouldn't have many false positives).

I have no problem trying it out if we find it has too many false positives or it is problematic we can always disable it.

hezhizhen:
> > Well, that works.
> > As far as `typos` goes, I wouldn't be opposed, but I'd want agreement from at least one other maintainer. @davidkna @andytom @matchai What are our thoughts about adding a typo-detection tool to our CI? (It seems that it's designed to work on a list of common typos instead of functioning as a spellcheck, so in principle it seems it shouldn't have many false positives).
> 
> I have no problem trying it out if we find it has too many false positives or it is problematic we can always disable it.

Then I'll add it in the next PR.

PR #4966, opened 2023-03-06
fix(java): wrong version number when using Android Studio JDK

agateau:
#### Description

When using the JDK provided by Android Studio, the Java version number is `v7284624`. This is because the regular expression used to extract the version catch an SVN revision found later in the returned string.

Add `, built on` to the end of the regex to ensure it stops earlier.

Screenshot below shows the prompt before and after the fix.

image

- Extended unit-test
- Built starship with the changes and used it

#### Checklist:

### DIFF
--- a/src/modules/java.rs
+++ b/src/modules/java.rs
@@ -5,7 +5,8 @@ use crate::utils::get_command_string_output;
 use std::path::PathBuf;
 
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


andytom (approved):
LGTM

davidkna:
Thanks for the contributions @agateau, and thanks for the review @andytom!

PR #4568, opened 2022-11-05
fix: check that BP_PIPESTATUS is defined

raffaem:
#### Description
Check that BP_PIPESTATUS is defined before operating on it.

#### Motivation and Context

If `set -u` to treat undefined parameters as error during parameters expansion, you will get an error at every command.

Closes #4567
Closes #4540

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [X] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #4972, opened 2023-03-07
fix(bash): Handle Unbound Variables Errors in Bash

tricktron:
#### Description

If you run starship with `set -u` in Bash you get several unbounded variables errors. This pr fixes that by first checking if the variables are set or empty and continues accordingly.

#### Motivation and Context

Closes https://github.com/starship/starship/issues/4540

Built from source using Nix and then tested with:
```bash
set -u
run any command
observe no more unbounded variable errors
```

#### Checklist:

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -71,7 +71,7 @@ starship_precmd() {
 
     local -a ARGS=(--terminal-width="${COLUMNS}" --status="${STARSHIP_CMD_STATUS}" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="${NUM_JOBS}")
     # Prepare the timer data, if needed.
-    if [[ $STARSHIP_START_TIME ]]; then
+    if [[ -n "${STARSHIP_START_TIME-}" ]]; then
         STARSHIP_END_TIME=$(::STARSHIP:: time)
         STARSHIP_DURATION=$((STARSHIP_END_TIME - STARSHIP_START_TIME))
         ARGS+=( --cmd-duration="${STARSHIP_DURATION}")
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -92,7 +92,7 @@ if [[ ${BLE_VERSION-} && _ble_version -ge 400 ]]; then
     blehook PRECMD!='starship_precmd'
 # If the user appears to be using https://github.com/rcaloras/bash-preexec,
 # then hook our functions into their framework.
-elif [[ "${__bp_imported:-}" == "defined" || $preexec_functions || $precmd_functions ]]; then
+elif [[ "${__bp_imported:-}" == "defined" || -n "${preexec_functions-}" || -n "${precmd_functions-}" ]]; then
     # bash-preexec needs a single function--wrap the args into a closure and pass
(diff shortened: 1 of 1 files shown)


tricktron:
@chipbuster Maybe you could have a look at this, since you reviewed the last [closed pull request](https://github.com/starship/starship/pull/4568) regarding this issue.

aarondill:
Any update on this? I just ran into this same issue today and would be overjoyed to see this merged

tricktron:
Bump. Maybe @andytom or @davidkna do you have some time for a review?

davidkna:
@tricktron I will try to look at this this week, if you would please resolve the conflicts.

davidkna (changes requested):
I tried this out with `set -u` and discovered some issues.
Please also handle unset `preexec_functions`, `precmd_functions` and `PROMPT_COMMAND`.

tricktron:
@davidkna How can I reproduce these issues? How did you test?

davidkna:
I built the binary and used the usual init sequence to source it in a fresh shell (`bash --noprofile --norc -ux`). Instead of building a binary, you could also temporarily replace `::STARSHIP::` with `starship` and source the file directly.

tricktron:
> I tried this out with `set -u` and discovered some issues. Please also handle unset `preexec_functions`, `precmd_functions` and `PROMPT_COMMAND`.

Should be fixed now.

src/init/starship.bash
```diff
@@ -111,9 +113,9 @@ else
 
     # Finally, prepare the precmd function and set up the start time. We will avoid to
     # add multiple instances of the starship function and keep other user functions if any.
-    if [[ -z "$PROMPT_COMMAND" ]]; then
+    if [[ -n "${PROMPT_COMMAND-}" ]]; then
```
davidkna, line 116: This needs to remain `-z`
```suggestion
    if [[ -z "${PROMPT_COMMAND-}" ]]; then
```

src/init/starship.bash
```diff
@@ -111,9 +113,9 @@ else
...
         PROMPT_COMMAND="starship_precmd"
-    elif [[ "$PROMPT_COMMAND" != *"starship_precmd"* ]]; then
+    elif [[ -n "${PROMPT_COMMAND-}" && "$PROMPT_COMMAND" != *"starship_precmd"* ]]; then
```
davidkna, line 118: I would prefer this or keeping it as it was since this is already checked above.
```suggestion
    elif [[ "${PROMPT_COMMAND-}" != *"starship_precmd"* ]]; then
```

davidkna (changes requested):

src/init/starship.bash
tricktron: You are absolutely right, thanks!

src/init/starship.bash
tricktron: @davidkna I kept it as it is because you are right again, it is checked already.

davidkna (approved):
LGTM

tricktron:
@andytom @davidkna I think this ready to merge😃.

chipbuster:
@tricktron thanks for the PR and thanks @davidkna for reviewing!

PR #4974, opened 2023-03-09
fix: Rename var AWS_SESSION_EXPIRATION for aws-vault v7+

bruno:
In [aws-vault v7.0.0](https://github.com/99designs/aws-vault/releases/tag/v7.0.0) the authors removed the `AWS_SESSION_EXPIRATION` env var and recommend the use of `AWS_CREDENTIAL_EXPIRATION`. This commit just renames the variable name.

#### Description

A simple rename of the old ENV var that was expected to contain the expiration of the  session to the new name.

#### Motivation and Context

As described in the commit message, [aws-vault v7.0.0](https://github.com/99designs/aws-vault/releases/tag/v7.0.0) no longer sets the variable named `AWS_SESSION_EXPIRATION` with the details of when the session will expire, so `starship` wasn't displaying it anymore.

Closes: #4783

I've run the `cargo test` locally on an Apple Silicon and x86 Linux (Ubuntu 22.04 LTS).

`test result: ok. 951 passed; 0 failed; 32 ignored; 0 measured; 0 filtered out; finished in 0.82s`

#### Checklist:


davidkna (changes requested):
I think adding support for the other environment variable is good, and it sounds like it should have a higher priority than the current value, but please keep the support of it for the time being. This new aws-vault release has only been out for 3 days, and other tools could using the variable.
Please also avoid making any changes to the non-English documentation.

bruno:
Thanks for the feedback @davidkna. All good suggestions, sorry about the changes to the non-English docs. I will try to get to this in the near future ...

PR #4975, opened 2023-03-09
chore: add spell checker to workflows

hezhizhen:
#### Description

Add `typos` to `workflows` and fix existing typos.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/configs/fennel.rs
+++ b/src/configs/fennel.rs
@@ -13,6 +13,7 @@ pub struct FennelConfig<'a> {
     pub symbol: &'a str,
     pub style: &'a str,
     pub disabled: bool,
+    #[serde(alias = "detect_extentions")] // TODO: remove it after breaking change releases
     pub detect_extensions: Vec<&'a str>,
     pub detect_files: Vec<&'a str>,
     pub detect_folders: Vec<&'a str>,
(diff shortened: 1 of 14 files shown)


src/configs/fennel.rs
```diff
@@ -13,7 +13,7 @@ pub struct FennelConfig<'a> {
     pub symbol: &'a str,
     pub style: &'a str,
     pub disabled: bool,
-    pub detect_extentions: Vec<&'a str>,
+    pub detect_extensions: Vec<&'a str>,
```
davidkna, line 16: I would prefer not to break compatibility with older configs with this change:
```suggestion
    #[serde(alias = "detect_extentions")]
    pub detect_extensions: Vec<&'a str>,
```

hezhizhen: How about keeping the old config and adding a TODO?

hezhizhen: PTAL

davidkna: We have some alias for similar cases, I would prefer using that here; but please add a TODO, to ensure it will be remembered on the next breaking release.

hezhizhen: Done. PTAL

Btw, this typo only happens in fennel.

davidkna (approved):

hezhizhen:
@davidkna It's not merged.

chipbuster:
@hezhizhen We have a soft agreement among maintainers (in the sense that I don't think it's been explicitly discussed, but we try to stick to it when possible) that approving and merging are done by different parties.

PR #4978, opened 2023-03-09
fix(fossil_branch): fossil checkout database file name on windows

stefanc:
#### Description

When running on Windows, use `_FOSSIL_` as checkout database file name instead of `.fslckout`.

#### Motivation and Context

On Windows, the checkout database file name is `_FOSSIL_` (see [this document](https://www.fossil-scm.org/home/doc/trunk/www/tech_overview.wiki) for details). As a result, the `fossil_branch` module is not able to detect fossil checkout directories.

This PR sets the correct checkout database file name for all platforms, fixing the issue.

#### Checklist:

### DIFF
--- a/src/modules/fossil_branch.rs
+++ b/src/modules/fossil_branch.rs
@@ -17,9 +17,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     };
 
+    let checkout_db = if cfg!(windows) {
+        "_FOSSIL_"
+    } else {
+        ".fslckout"
+    };
+
     let is_checkout = context
         .try_begin_scan()?
-        .set_files(&[".fslckout"])
+        .set_files(&[checkout_db])
         .is_match();
 
     if !is_checkout {
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -176,11 +176,16 @@ pub enum FixtureProvider {
 pub fn fixture_repo(provider: FixtureProvider) -> io::Result<TempDir> {
     match provider {
         FixtureProvider::Fossil => {
+            let checkout_db = if cfg!(windows) {
+                "_FOSSIL_"
+            } else {
+                ".fslckout"
+            };
             let path = tempfile::tempdir()?;
             fs::OpenOptions::new()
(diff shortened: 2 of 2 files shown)


davidkna:
Thanks for the fix!
Do you happen to know how strict fossil is about only supporting one option or the other? It might be useful to enable both options on all operating systems to aid compatibility with WSL/cygwin or shared file-systems.

stefanc:
Hey @davidkna, thanks for the quick reply!

> Do you happen to know how strict fossil is about only supporting one option or the other? 

That's a good point I have not consider. Looking at fossil's source code, it seems `_FOSSIL_` is used for Windows and Cygwin and `.fslckout` for everything else. From [src/db.c](https://www.fossil-scm.org/home/file?name=src/db.c&ci=trunk):

```cpp
#if defined(_WIN32) || defined(__CYGWIN__)
# define LOCALDB_NAME "./_FOSSIL_"
#else
# define LOCALDB_NAME "./.fslckout"
#endif
```

I am not sure how Rust handles Cygwin but I think it's covered by `if cfg!(windows)`. Please correct me if I'm wrong.

>  It might be useful to enable both options on all operating systems to aid compatibility with WSL/cygwin or shared file-systems.

I don't think fossil's checkout database was intended to be used this way. I think the repository database is intended as the portable bit and the checkout database is platform specific.

I tried to access a repository opened on Windows from WSL and while it found the `_FOSSIL_` checkout database, I got an error because the path to the repository database (`project.fossil`) was an absolute windows path. I then opened another copy on Windows with a relative path to the repository database, and this time managed to access it from WSL but all files in the repo appeared as changed because they were marked as `EXECUTABLE`.

In conclusion, while fossil looks for both file names on all platform, I don't think it's a good idea to access a checkout from one platform on another. To avoid the kind of issues mentioned above I'd rather check just for the file name fossil uses to create a checkout database on the current platform and ignore the rest. Let me know what you think.

davidkna (approved):
LGTM, but please rebase/merge master to get the clippy check to run.

andytom:
Thank you for your contribution @stefanc and thanks for reviewing @davidkna.

PR #5002, opened 2023-03-15
feat(aws): Adds support for AWS_CREDENTIAL_EXPIRATION environment variable

hhtpcd:
Adds support for the AWS_CREDENTIAL_EXPIRATION environment variable which was adopted as the standard way to set the expiration for temporary credentials. The existing AWS_SESSION_EXPIRATION environment variable is not dropped for backwards compatibility.

See https://github.com/aws/aws-cli/pull/7398

#### Description

The expiration timer does not show when the `AWS_SESSION_EXPIRATION` variable is not set. the aws-cli has adopted `AWS_CREDENTIAL_EXPIRATION` as the default variable for storing the expiration time of temporary credentials. This change updates the aws prompt to consider both environment variables. 

#### Motivation and Context

[99design/aws-vault at v7.0.0 switched](https://github.com/99designs/aws-vault/pull/1124) to using the new variable. Users on this version of `aws-vault` will not see a timer, as per #4783

Closes: #4783
Closes: #4974

image

Tested with version 7.1.0 of `aws-vault` and compiled starship binary from this branch. Requested temporary credentials and got duration time in the prompt again.

I can see there is a [test case that allows for some drift in timings](https://github.com/starship/starship/blob/0ccb9e4bbf9d1382e1e60fb43a7209c37f5774c5/src/modules/aws.rs#L717-L719), I've put the timestamp init at the top of the iter block instead to try and avoid timing issues.

#### Checklist:


davidkna (approved):

andytom:
Thank you for your contribution @hhtpcd and thank you for reviewing @davidkna.

PR #5005, opened 2023-03-18
Add support for excluding seconds in render_time and aws module duration

senilio:
#### Description

Adding support for rendering time without seconds.
Adding support in AWS module to display session duration without seconds.

#### Motivation and Context

Motivation for doing this is to be able to reduce the length of the prompt as much as possible, which is nice when displaying the prompt on a single line. In my case, an AWS session lasts a full working day, so displaying seconds in the prompt is not interesting and wastes valuable horizontal line space.

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -10,6 +10,7 @@
     "aws": {
       "default": {
         "disabled": false,
+        "duration_show_seconds": true,
         "expiration_symbol": "X",
         "force_display": false,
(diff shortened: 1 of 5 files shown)


davidkna:
I would prefer if you added a `total_seconds` variable or something like that, an option like this may not play well with future changes.

davidkna:
There have been attempts to make this more flexible like #4863 which I would prefer as a more general solution to this sort of problem.

PR #5016, opened 2023-03-21
fix(nu): upgrade closure syntax

### LINKED ISSUE #5060: [nushell] 0.78 breaking change for closure
A breaking change was introduce in [v0.78](https://www.nushell.sh/blog/2023-04-04-nushell_0_78.html) of NuShell:

> ### `||` now required in closures
>
> To help differentiate between blocks (which can mutate variables) and closures (which can be used in a pipeline), we've changed the syntax of closures to require `||`. This means the simplest closure now looks like `{|| }`

dmatos2012:
#### Description

This PR only updates syntax for the up and coming nushell releases. I believe 0.77.2 will be the first release(currently 0.77.1). Check out the related [PR](https://github.com/nushell/nushell/pull/8290)

#### Motivation and Context
Otherwise its syntax error while running `init.nu`

Closes #5054
Closes #5061
Closes #5060

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,7 +341,7 @@ And add the following to the end of your Nushell configuration (find it by runni
 source ~/.cache/starship/init.nu
 ```
 
-Note: Only Nushell v0.73+ is supported
+Note: Only Nushell v0.78+ is supported
 
 </details>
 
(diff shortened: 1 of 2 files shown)


dmatos2012:
Currently after 0.78, which might get released in 3 weeks or so, the syntax is no longer supported. Currently, `nu` is in 0.77 which does support this syntax. 

Another possibility is maybe leave this as draft till `0.78` gets released. I am not sure of the release cycle here. 
 Thanks for quick response @davidkna

davidkna:
Please also update the version number in `docs/README.md` and `install/install.sh`. Also, if v0.77.2 is going to be an intermediate release that supports both syntaxes, please use that version for the bump.

PR #5020, opened 2023-03-23
fix(bash): use `eval` instead of a procsub for the POSIX mode

### LINKED ISSUE #1674: Lightdm session won't start because of failed eval of starship
#### Current Behavior
Lightdm greeter goes through authentication and goes back to login screen as soon as it evaluates starships line in `.bashrc` making it impossible to login. Manually sourcing `.bashrc` from terminal with starship works correctly.

#### Expected Behavior
Starship correctly evaluating and lightdm session starting.

Content of `~/.xsession-errors`
```
/home/wojtek/.bashrc: eval: line 3: syntax error near unexpected token `('
/home/wojtek/.bashrc: eval: line 3: `source <("/usr/bin/starship" init bash --print-full-init)'
```
#### Environment
- Starship version: 0.44.0
- bash version: GNU bash, version 5.0.18(1)-release (x86_64-pc-linux-gnu)

- Operating system: Arch Linux
- Terminal emulator: Alacritty 0.6.0-dev

```bash
eval "$(starship init bash)"
```

#### Starship Configuration

```toml
<unknown config>
```

### LINKED ISSUE #5382: Error on launching with Kitty Terminal
#### Current Behavior

getting the below error when launching kitty. Works as per expected with iterm2.

I am using bash shell.

```
 -bash: eval: line 8: syntax error near unexpected token `('
-bash: eval: line 8: `                    source <(/opt/homebrew/bin/starship init bash --print-full-init)'
```

#### Environment
- Starship version: 1.16.0
- Operating system: Mac OS 13.5.0
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.71.0 (8ede3aae2 2023-07-12) (built from a source tarball)
- Rust channel:  release
- Build Time: 2023-07-30 13:30:05 +00:00

```bash
<unknown config>
```

#### Starship Configuration

```toml
add_newline = true

[git_status]
disabled = false

[git_commit]
commit_hash_length = 7
tag_disabled = false
only_detached = false

[battery]
disabled = true

[python]
disabled = true

[cmake]
disabled = true

[directory]
read_only = "(read only)"

[git_metrics]
disabled = false
format = '([+$added]($added_style)) ([-$deleted]($deleted_style)) '

```

akinomyoga:
#### Description

Use `eval -- "$(/path/to/starship init bash --print-full-init)"`, which works in any version of Bash and also in the POSIX mode, instead of the current version of `source <(...)` which does not work when the POSIX mode is turned on in Bash <= 5.0. I have added a detailed explanation in the code comments.

#### Motivation and Context

~~In Bash 5.0 and below~~ (**edit**: After I submitted this PR, #5735 introduced the approach using `PS0` for Bash 4.4 and above, so the problem arises only in Bash 4.3 and below as of 2024-04-01), Starship initialization failed when the POSIX mode is turned on. This PR fixes the issue.

Closes #1674
Closes #3660
Closes #5382

With the following `bashrc`:

```bash
# bashrc
set -o posix
eval "$(starship init bash)"
```

Before the fix:
image

After the fix:
image

#### Checklist:

There doesn't seem to be any documentation about the internal implementation of the original approach `<(...)`, so I judged we do not have to add documentation for the new approach either.

I'm not familiar with Rust, but `cargo check` just seems to pass.

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -9,11 +9,11 @@ use which::which;
 shell. This command evaluates a more complicated script using `source` and
 process substitution.
 
-Directly using `eval` on a shell script causes it to be evaluated in
-a single line, which sucks because things like comments will comment out the
-rest of the script, and you have to spam semicolons everywhere. By using
-source and process substitutions, we make it possible to comment and debug
-the init scripts.
+Directly using `eval` on a shell script without proper quoting causes it to be
+evaluated in a single line, which sucks because things like comments will
+comment out the rest of the script, and you have to spam semicolons
+everywhere. By using source and process substitutions, we make it possible to
+comment and debug the init scripts.
 
 In the future, this may be changed to just directly evaluating the initscript
 using whatever mechanism is available in the host shell--this two-phase solution
@@ -106,8 +106,13 @@ pub fn init_stub(shell_name: &str) -> io::Result<()> {
 
     match shell_basename {
         "bash" => print!(
-            /*
-             * The standard bash bootstrap is:
+            /* We now use the following bootstrap:
+             *      `eval -- "$(starship init bash --print-full-init)"`
+             * which works in any version of Bash from 3.2 to the latest
+             * version and also works in the POSIX mode.
+             *
+             * ----
+             * Historically, the standard bash bootstrap was:
              *      `source <(starship init bash --print-full-init)`
              *
              * Unfortunately there is an issue with bash 3.2 (the MacOS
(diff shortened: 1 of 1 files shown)


akinomyoga:
@davidkna Could you take a look at this PR? I just wanted to submit another PR to solve the issues with the `DEBUG` trap, but to explain the problem the coming PR tries to solve, I need this PR.

Maybe the code coverage error in the CI blocks the reviewing but I'd like some advice about how to fix it (or whether I should fix it in the first place because I think the part I modified wasn't covered by the test even before this change. In the part to which `codecov` issues an error, I haven't added any code but only changed the content of the string literal).

akinomyoga:
I think this PR also solved the problem that PR #3660 tried to solve. The problem PR #3660 attempts to solve is that `trap -p DEBUG` is called inside a function `_main` and `source`, so the DEBUG trap at the top level is not captured by `trap -p DEBUG`. To solve the problem, PR #3660 tried to temporarily set `set -o functrace` to make all the `DEBUG` trap handlers inherited by the child function calls, which is a global behavioral change of the entire shell. However, the problem disappears with the PR because this PR makes `trap -p DEBUG` executed at the top level.

davidkna (approved):
LGTM

akinomyoga:
Thank you for your quick response! If there is anything additional needed, please let me know.

andytom:
Thank you for your contribution @akinomyoga and thanks for reviewing @davidkna.

akinomyoga:
Thank you!

PR #5028, opened 2023-03-26
feat(aws): add aws-sso-cli support.

### LINKED ISSUE #3650: Support AWS SSO information via aws-sso-cli
## Feature Request

When using AWS SSO with a tool like [aws-sso-cli](https://github.com/synfinatic/aws-sso-cli) it would be helpful to support seeing my AWS_SSO_PROFILE in the prompt. 

I think it might be ideal to have a separate module for this, which would allow getting access to a series of key details:
* AWS_SSO_PROFILE
* AWS_SSO_ACCOUNT_ID
* AWS_SSO_ROLE_NAME
* Calculated seconds to token expiration
* A simple flag to indicate a session is present, and if present if it is active or expired (since most of the above values can be long)

Most simply, this would be supported via an $sso_profile variable that could be handled by the user as desired.

frap:
#### Description
Add aws-sso-cli support to the aws module

#### Motivation and Context
I use aws-sso and like to know what environment Im running in in terminal

Closes #3650

#### Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -97,8 +97,14 @@ fn get_aws_profile_and_region(
     context: &Context,
     aws_config: &AwsConfigFile,
 ) -> (Option<Profile>, Option<Region>) {
-    let profile_env_vars = ["AWSU_PROFILE", "AWS_VAULT", "AWSUME_PROFILE", "AWS_PROFILE"];
-    let region_env_vars = ["AWS_REGION", "AWS_DEFAULT_REGION"];
+    let profile_env_vars = [
+        "AWS_SSO_PROFILE",
+        "AWSU_PROFILE",
+        "AWS_VAULT",
+        "AWSUME_PROFILE",
+        "AWS_PROFILE",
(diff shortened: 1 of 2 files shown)


src/modules/aws.rs
```diff
@@ -122,6 +128,7 @@ fn get_credentials_duration(
     aws_creds: &AwsCredsFile,
 ) -> Option<i64> {
     let expiration_env_vars = [
+        "AWS_SSO_SESSION_EXPIRATION",
         "AWS_CREDENTIAL_EXPIRATION",
         "AWS_SESSION_EXPIRATION",
         "AWSUME_EXPIRATION",
```
davidkna, line 134: I'm not sure if `AWS_SSO_SESSION_EXPIRATION` should be first here, because `AWSUME_EXPIRATION` is last. On the other hand, other options prefer the non-aws vars.

JoshuaFurman:
would be so awesome to have this - crazy that this fix has been out there for so long without getting merged

PR #5038, opened 2023-03-28
fix(pulumi): Fix formatting on pulumi module when using version

rjsab:
#### Description
Sanitize `pulumi version` output to remove leading 'v' character and trailing newlines.

#### Motivation and Context
The output of `pulumi version` currently contains a leading 'v' character and trailing newline character in version 3.60.0. This creates a duplicate v and unexpected newline when leveraging $version in the pulumi module configuration. I'm not certain when the behavior change occurred in the `pulumi version` command, but I believe the change proposed in this PR preserves backwards compatibility.

Updated test cases in the PR.

#### Checklist:

### DIFF
--- a/src/modules/pulumi.rs
+++ b/src/modules/pulumi.rs
@@ -81,24 +81,28 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     }
 }
 
-/// Parse the output of `pulumi version` into just the version string.
+/// Parse and sanitize the output of `pulumi version` into just the version string.
 ///
 /// Normally, this just means returning it. When Pulumi is being developed, it
 /// can return results like `3.12.0-alpha.1630554544+f89e9a29.dirty`, which we
 /// don't want to see. Instead we display that as `3.12.0-alpha`.
 fn parse_version(version: &str) -> &str {
+    let new_version = version.strip_prefix('v').unwrap_or(version);
+
+    let sanitized_version = new_version.trim_end();
+
     let mut periods = 0;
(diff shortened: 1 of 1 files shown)


src/modules/pulumi.rs
```diff
@@ -81,24 +81,31 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
 /// Normally, this just means returning it. When Pulumi is being developed, it
 /// can return results like `3.12.0-alpha.1630554544+f89e9a29.dirty`, which we
 /// don't want to see. Instead we display that as `3.12.0-alpha`.
 fn parse_version(version: &str) -> &str {
+    let new_version = version.strip_prefix('v').unwrap_or(version);
+
+    let sanitized_version = new_version
+        .strip_suffix("\r\n")
+        .or_else(|| new_version.strip_suffix('\n'))
+        .unwrap_or(new_version);
```
davidkna, line 95: This should also work, and would be a bit simpler:
```suggestion
    let sanitized_version = new_version.trim_end();
```

davidkna (approved):
I made a small suggestion, but in general this looks good to me.

src/modules/pulumi.rs
rjsab: Great suggestion! I completely missed fine print of trim_end() including new lines when looking through the docs. I've updated the PR to reflect the suggestion!

andytom:
Thank you for your contribution @rjsab and thank you for reviewing @davidkna.

PR #5049, opened 2023-04-01
feat(pwsh): Support vi command mode indicator

Nemo157:
#### Description
Updates the `character` module to support showing vicmd-mode when specified for powershell, and changes `starship.ps1` to detect the mode when drawing the prompt and trigger a prompt redraw when needed. The triggering of prompt redraw doesn't work on PS5's builtin version of PSReadLine, so an error when setting it up is silently ignored.

#### Motivation and Context
Increases feature parity of the PowerShell support with other shells.

Tested on PowerShell 7.3.2 with PSReadLine 2.2.6 and PowerShell 5.1 with PSReadLine 2.2.6 and 2.0.0 (which doesn't _work_ because you can't get it to redraw the prompt, but it still runs as it did previously).

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -141,6 +141,10 @@ $null = New-Module starship {
 
         $arguments += "--status=$($lastExitCodeForPrompt)"
 
+        if ([Microsoft.PowerShell.PSConsoleReadLine]::InViCommandMode()) {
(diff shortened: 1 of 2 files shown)


davidkna (approved):

davidkna:
Thanks for implementing this!

PR #5052, opened 2023-04-01
feat(git_metrics): add option to ignore submodules

cdmistman:
#### Description

There's an option in [`git_status`](https://starship.rs/config/#git-status) to ignore git submodules, this does the same for the `git_metrics` module.

#### Motivation and Context

I have a repository with a Very Large Git Submodule™️ causing the git_metrics module to not show in my starship prompt due to the starship timeout configuration. It's inconvenient having the prompt not paint for ~300ms (my timeout configuration) so this adds a configuration option to the `git_metrics` module to pass the `--ignore-submodules` flag to `git diff`.

N/A

  - (I'm using Nix, so I'm omitting irrelevant steps/info)
  - Add `git_metrics.ignore_submodules = true` to my starship config
  - Build my edited version of starship
  - `export PATH="/path/to/starship/repo/target/debug:$PATH"`
  - `eval $(starship init zsh)`
  - went to my repo with the Very Large Git Submodule™️
  - noticed presence of git metrics module
  - `starship timings` showed the module took 15ms to render

#### Checklist:

### DIFF
--- a/.devenv/profile
+++ b/.devenv/profile
@@ -1,1 +1,0 @@
-/nix/store/kzwr00qay6njqzj3ypw7wc2zpm1ajzj7-devenv-profile
--- a/.devenv/profile
+++ b/.devenv/profile
@@ -1,0 +1,1 @@
+/nix/store/kzwr00qay6njqzj3ypw7wc2zpm1ajzj7-devenv-profile
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -575,8 +575,8 @@
         "deleted_style": "bold red",
         "disabled": true,
         "format": "([+$added]($added_style) )([-$deleted]($deleted_style) )",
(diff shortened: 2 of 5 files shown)


davidkna (changes requested):
Please try adding a test for this, even if it only confirms that the module still works after the change.

cdmistman:
Added a test that's redundant of another test, except adds the `ignore_submodules = true` configuration. I'll note that the `git_status` module doesn't test at all with `ignore_submodules = true`.

For posterity: Git bans local repositories from being submodules (submodules must have a valid networked remote), so this feature can't be tested without pulling the starship repo on every `cargo test`, which will add *lots* of time to the test suite.

davidkna (approved):

davidkna:
Thanks for implementing this @cdmistman!

PR #5054, opened 2023-04-02
chore(nu): use updated closure syntax

### LINKED ISSUE #5060: [nushell] 0.78 breaking change for closure

SamirTalwar:
#### Description

Nushell recently made a change to require that all closures have an explicit parameter list, even if it's empty, in https://github.com/nushell/nushell/pull/8290.

This updates starship.nu to conform to this requirement.

#### Motivation and Context

Without this change, Starship will break on Nushell when the next version of Nushell is released.

Closes #5016
Closes #5061
Closes #5060

I have casually tested this against both the latest released version of Nushell, and the latest version on HEAD; the changed code works well (for me) on both.

#### Checklist:

### DIFF
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -6,7 +6,7 @@ let-env PROMPT_MULTILINE_INDICATOR = (^::STARSHIP:: prompt --continuation)
 # TODO: Also Use starship vi mode indicators?
 let-env PROMPT_INDICATOR = ""
 
-let-env PROMPT_COMMAND = {
+let-env PROMPT_COMMAND = { ||
     # jobs are not supported
     let width = (term size).columns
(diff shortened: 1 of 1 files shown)


davidkna (approved):
LGTM. It looks like it indeed works without bumping the minimum version.

andytom:
Thank you for your contribution @SamirTalwar and thanks for reviewing @davidkna.

presidento:
Thank you for fixing it.

SamirTalwar:
Thanks for merging!

happysalada:
Would it be possible to do a minor release including this patch ?

PR #5072, opened 2023-04-07
fix(pwsh): avoid extraneous first newline on terminal on clear/open

### LINKED ISSUE #560: Empty line when opening terminal
#### Current Behavior
When `add_line` is enabled a new line is added at the beginning of every new terminal window/tab.

#### Expected Behavior
Using `add_line` should not add a new line above the first line of every window / tab.

new-line

#### Environment
- Starship version: starship 0.25.0
- Shell type: zsh
- Shell version: zsh 5.3 (x86_64-apple-darwin18.0)
- Shell plugin manager: antibody
- Terminal emulator: iTerm2

```
eval "$(starship init zsh)"

source <(antibody init)
antibody bundle < ~/.zsh_plugins.txt
```

#### Starship Configuration
```
add_newline = true
```

#### Possible Solution
```
if config.add_newline {
        writeln!(buf).unwrap();
}
```
The above if-statement from https://github.com/starship/starship/blob/master/src/print.rs#L23 could be changed into something inspired by Spaceship's solution:
https://github.com/maximbaz/spaceship-prompt/commit/815b68c70f21c5512765c87985897b98b8ed232c found in  https://github.com/denysdovhan/spaceship-prompt/issues/462

domsleee:
#### Description
When `add_newline` is true, a newline is added at the top when opening the terminal for the first time, and when clearing the terminal.
This change fixes it by passing in the terminal row into the `starship` invocation, and only printing a newline if the row is not zero.

#### Motivation and Context
Fixes https://github.com/starship/starship/issues/560 for pwsh and powershell
There are workarounds mentioned for zsh and fish using hooks like fish's `fish_postexec`
I couldn't find any such hook for powershell

image

Testing steps
1. `cargo install --path . --force`
2. Re-open your terminal (so that the profile is reloaded)

Testing scenarios
Scenario | Behaviour
---------|----------
1 Opening a new terminal | No newline in any circumstance
2 Clearing the terminal with `ctrl+l` | No newline in any circumstance
3 Clearing the terminal with `clear` | No newline in any circumstance
4 Using `ctrl+c` | Adds a newline if `add_newline = true`
5 Pressing enter after some text |  Adds a newline if `add_newline = true`
6 Pressing enter with no text | Adds a newline if `add_newline = true`

- [x] I have tested using **MacOS** (pwsh)
- [x] I have tested using **Windows** (powershell + pwsh)

#### Checklist:

#### Additional notes
I did try using `crossterm`, eg `mybin/src/main.rs`:
```rust
fn main() {
    println!("crossterm::cursor::position {:?}", crossterm::cursor::position());
}
```

This works correctly (on windows) when the executable is invoked directly, i.e. `&mybin`
It always returns (0, 0) when invoked using [`Invoke-Native -Executable mybin` from starship.ps1](https://github.com/starship/starship/blob/master/src/init/starship.ps1).


davidkna (changes requested):
The approach seems reasonable, though this method won't work if there is anything being printed before the first prompt is drawn.
If possible, please add integration of this `--row` parameter for some other supported shells, given that giving a common experience was expressed as a concern to other solutions. Please also document the new `add_newline` behaviour.

src/print.rs
```diff
@@ -119,7 +119,10 @@ pub fn get_prompt(context: Context) -> String {
     let module_strings = root_module.ansi_strings_for_shell(context.shell, Some(context.width));
     if config.add_newline && context.target != Target::Continuation {
         // continuation prompts normally do not include newlines, but they can
-        writeln!(buf).unwrap();
+        let should_print_newline = context.row != Some(0);
+        if should_print_newline {
+            writeln!(buf).unwrap();
+        }
```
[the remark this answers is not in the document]
domsleee: we can move `context.row() != Some(0)` to the if condition above, maybe that would make more sense?

domsleee:
Ok so I can't really think of a "zero cost" solution for the other shells

There are some ideas though

Idea 1: We could use `"\033[6n"` before executing the shell [ex](https://stackoverflow.com/questions/43911175/get-current-cursor-position-while-drawing-zsh-prompt), but render time may take a hit

Idea 2: We could move the `add_newline` logic for some shells to hooks e.g. `fish_postexec`. I think this would require getting the config value of `add_newline` during init and saving it somewhere, e.g. `STARSHIP_ADD_NEWLINE`. Startup time would take a hit because we would be calling `starship` on init

Idea 1 seems more promising, I'll look into it 👍

src/print.rs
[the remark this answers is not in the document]
domsleee: actually, what does this comment even mean
>         // continuation prompts normally do not include newlines, but they can

davidkna:
> Ok so I can't really think of a "zero cost" solution for the other shells

I expect other shells to have similar options that would allow them to forward the current row, or at least if it is the first row.

domsleee:
> 1 Opening a new terminal
2 Clearing the terminal with ctrl+l
3 Using ctrl+c

I tested these three scenarios with `add_newline = true` on powershell, zsh & bash
I tried solving it for fish using `fish_postexec`, but I ran into issues with ctrl+c not printing a newline.
I also tried using the ansi escape approach in zsh, but it returned the cursor before the shell was cleared, so it isn't an option

davidkna:
I want this kind of config file parsing to be added to the init scripts.

davidkna:
I apologize for only looking at this now, there have been so many changes in this PR that I've lost track of the current progress. Please remove or split out the clear changes, I don't feel comfortable reviewing those, and I would prefer handling the initial line case first.

domsleee:
Gday @davidkna, no worries for a late review, thanks for your ongoing feedback 👍 
I think it would make more sense to break it up per shell implementation, so I removed the zsh/bash ones from this PR.

Of course we split it by "open" and "clear" functionality if you prefer.

I think the bash/zsh approach in this PR of calling `starship` on clear might be undesirable for some users, there may be a better solution for these shells. So it makes sense to split them off to another PR I reckon.

aradalvand:
@davidkna Can you please take a look at @domsleee's [latest comment](https://github.com/starship/starship/pull/5072#issuecomment-1593221115)?
Can this PR be merged now?

domsleee:
I have been using the `pwsh` solution here for a few months on windows (native, not wsl) and mac with no issues, so this PR is still relevant 😄

If this is merged, it should be a squash merge 👍 
Otherwise if you want me to rebase it, just let me know.

One concern I had is the performance of getting the cursor position, so I built [PowerShell/PowerShell](https://github.com/PowerShell/PowerShell) and debugged it.

How pwsh `$Host.UI.RawUI.CursorPosition.Y` works:

- On windows, [CursorPosition](https://github.com/PowerShell/PowerShell/blob/18350bc9ff28532d3d75cf4f07642d35fd8c490b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs#L157C9-L194C10) calls win32 [GetConsoleScreenBufferInfo](https://learn.microsoft.com/en-us/windows/console/getconsolescreenbufferinfo), and I would assume `powershell` calls the same
- On unix, [CursorPosition](https://github.com/PowerShell/PowerShell/blob/18350bc9ff28532d3d75cf4f07642d35fd8c490b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs#L1368-L1383) uses `CursorTop` from System.Console, which calls [TryGetCursorPosition](https://github.com/dotnet/runtime/blob/8f5364695b107f37af899aba3ad1b9f98d577c33/src/libraries/System.Console/src/System/ConsolePal.Unix.cs#L437-L672). This function parses the Device Status Report escape sequence (e.g. `echo -ne "\033[6n"` in bash)

I think this benchmark is a reasonable indication that it is _reasonably_ fast (<1-2ms or so, if the results are even valid). I would expect a shell hook to outperform querying the cursor position though.
```bash
 hyperfine -L script '$Host.UI.RawUI.CursorPosition.Y','0' "pwsh -NoProfile -NoLogo -Command {{script}}" --warmup 5 --runs 500 --export-markdown out.md
```

<details>
<summary>Benchmark results</summary>

### Mac (iTerm2 + m1 pro)

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `pwsh -NoProfile -NoLogo -Command {'$Host.UI.RawUI.CursorPosition.Y'}` | 147.6 ± 3.8 | 140.5 | 186.4 | 1.00 |
| `pwsh -NoProfile -NoLogo -Command {'0'}` | 147.8 ± 2.6 | 141.6 | 163.0 | 1.00 ± 0.03 |

### Windows (Windows terminal + 5600X)

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `pwsh -NoProfile -NoLogo -Command {'$Host.UI.RawUI.CursorPosition.Y'}` | 147.6 ± 3.8 | 140.5 | 186.4 | 1.00 |
| `pwsh -NoProfile -NoLogo -Command {'0'}` | 147.8 ± 2.6 | 141.6 | 163.0 | 1.00 ± 0.03 |

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `powershell -NoProfile -NoLogo -Command {'$Host.UI.RawUI.CursorPosition.Y'}` | 105.9 ± 2.2 | 101.8 | 126.2 | 1.00 ± 0.05 |
| `powershell -NoProfile -NoLogo -Command {'0'}` | 105.6 ± 4.5 | 101.6 | 175.4 | 1.00 |

</details>

aradalvand:
@davidkna Just another ping...

src/init/starship.fish
```diff
@@ -43,8 +44,9 @@ function fish_right_prompt
             printf ""
         end
     else
-        ::STARSHIP:: prompt --right --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="$STARSHIP_CMD_PIPESTATUS" --keymap=$STARSHIP_KEYMAP --cmd-duration=$STARSHIP_DURATION --jobs=$STARSHIP_JOBS
+        ::STARSHIP:: prompt $STARSHIP_ADD_NEWLINE_FLAGS --right --terminal-width="$COLUMNS" --status=$STARSHIP_CMD_STATUS --pipestatus="$STARSHIP_CMD_PIPESTATUS" --keymap=$STARSHIP_KEYMAP --cmd-duration=$STARSHIP_DURATION --jobs=$STARSHIP_JOBS
     end
+    set -g STARSHIP_ADD_NEWLINE_FLAGS
```
davidkna, line 49: This approach is incompatible with vi-mode (I assume the same goes for pwsh).

src/init/starship.fish
```diff
@@ -84,6 +87,26 @@ function disable_transience --description 'remove transient prompt keybindings'
...
+
+function clear
+    starship_clear
+    if functions -q __original_clear
+        __original_clear
+    else
+        command clear
+    end
+end
+
```
davidkna, line 109: I actually meant what I said, and I will not be reviewing clearing.

davidkna (changes requested):

src/init/starship.fish
domsleee: 😢 Thanks for finding this, I agree it does not work for fish
Reproduction is `fish_vi_key_bindings`, clearing the screen, and then pressing escape.
Unfortunately this clear hook approach doesn't really work then.

I tested the same with powershell using `Set-PSReadlineOption -EditMode vi` and it works fine, since it always retrieves the cursor's position 👍

davidkna: Just flipping between vi modes causes an extra line to appear for me, even without clearing. Pwsh might only be working for you thanks to the y-position check.

src/init/starship.ps1
```diff
@@ -123,6 +123,10 @@ $null = New-Module starship {
             "--jobs=$($jobs)"
         )
 
+        if ($Host.UI.RawUI.CursorPosition.Y -eq 0 -or $script:firstRender) {
```
domsleee, line 126: Hi @davidkna, I'm not sure if "reviewing clearing" (https://github.com/starship/starship/pull/5072#discussion_r1274604551) also refers to this line. Hopefully this is not seen as ignoring your comment.

Please also consider that starship currently uses [CursorPosition.Y in the transient prompt code](https://github.com/starship/starship/blob/c88aeeaf15acd95df649cb07c167cdd0056c2d1c/src/init/starship.ps1#L89C124-L89C155). So there is some precedent for querying the cursor position 👍

src/init/starship.fish
domsleee: Yeah, and the y-position check could work for fish and perhaps other shells (won't work for zsh/bash on clear since the prompt is not re-evaluated on clear, iirc).

The way pwsh on mac queries the position [seems quite involved](https://github.com/dotnet/runtime/blob/8f5364695b107f37af899aba3ad1b9f98d577c33/src/libraries/System.Console/src/System/ConsolePal.Unix.cs#L437-L672), it might be an idea to query the cursor position from rust on nix based machines in a similar way.

Cmd would probably use win32 the same as pwsh/powershell on windows I think.
Many ideas 🤔

aradalvand:
giphy

PR #5074, opened 2023-04-07
fix(unix): improve readonly check reliability

guoci:
#### Description

Reimplement check if directory is writable with `access` syscall.

#### Motivation and Context

The current implementation does not always accurately determine if the directory is writable.

#### Checklist:

### DIFF
--- a/src/modules/utils/directory_nix.rs
+++ b/src/modules/utils/directory_nix.rs
@@ -1,52 +1,15 @@
-use nix::sys::stat::Mode;
-use nix::unistd::{Gid, Uid};
-use std::fs;
-use std::os::unix::fs::MetadataExt;
-use std::os::unix::fs::PermissionsExt;
+use nix::errno::Errno::ENOENT;
+use nix::unistd::access;
+use nix::unistd::AccessFlags;
 use std::path::Path;
 
 /// Checks if the current user can write to the `folder_path`.
-///
-/// It extracts Unix access rights from the directory and checks whether
-/// 1) the current user is the owner of the directory and whether it has the write access
-/// 2) the current user's primary group is the directory group owner whether if it has write access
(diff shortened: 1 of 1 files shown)


src/modules/utils/directory_nix.rs
```diff
@@ -1,54 +1,17 @@
...
-        Ok(perms & u32::from(Mode::S_IWOTH.bits()) != 0)
-    }
-}
-
-#[cfg(all(unix, not(any(target_os = "macos", target_os = "ios"))))]
-fn get_supplementary_groups() -> Vec<u32> {
-    match nix::unistd::getgroups() {
-        Err(_) => Vec::new(),
-        Ok(v) => v.into_iter().map(nix::unistd::Gid::as_raw).collect(),
+    match access(folder_path, AccessFlags::W_OK) {
```
davidkna, line 8: Where did the previous implementation have issues? Also, I feel like `eaccess` should be used instead.

guoci: I had issues where a directory is writable but that function determined it was not writable. That directory is mounted by `sshfs`.
It is easy to switch to `eaccess` or `faccessat` with `AT_EACCESS` flag. Both of them have some unsupported `target_os` or `target_env`.
https://github.com/nix-rust/nix/blob/master/src/fcntl.rs#L26-L38
https://github.com/nix-rust/nix/blob/master/src/unistd.rs#L3418
Why do we need effective ids for this use case?

davidkna: Wouldn't switching to regular ids be a regression compared to the previous behavior? It was measuring the effective ability to write before. I would rather not keep around the old code, but I feel like `eaccess` should be used on those platforms that support it.

guoci: `eaccess` doesn't work on `macOS`.

PR #5078, opened 2023-04-08
chore(choco): remove chocolatey dependency

davidkna:
#### Description

Following chocolatey moderator review comments, this PR removes the dependency on chocolatey, since it is safe to assume that no user is using a version older than required.

#### Motivation and Context

Closes #

#### Checklist:


davidkna:

PR #5079, opened 2023-04-09
refactor(Context): `set_config` method for `Context`

marcybelardo:
#### Description

Since the `ModuleRenderer` config method simply manipulates the context config fields, I added a `.set_config()` method within `Context` that directly overrides the config fields, and let the `config` method in `ModuleRenderer` call that method instead. 

This way, any tests that use a module renderer will continue functioning like normal, and now maintainers can use `set_config` to directly set the configuration of `Context` structs.

I've added a test to show that the configuration is completely overwritten by calling the method. Let me know if this seems like a good idea or not!

#### Motivation and Context

Implemented an idea from @davidkna about giving a `.config()` method to `Context` to match `ModuleRenderer`. Mentioned in #5077.

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -951,16 +951,6 @@ mod tests {
     }
 
     #[test]
-    fn set_config_method_overwrites_constructor() {
-        let context = default_context();
-        let mod_context = default_context().set_config(toml::toml! {
-            add_newline = true
-        });
-
-        assert_ne!(context.config.config, mod_context.config.config);
-    }
-
-    #[test]
     fn context_constructor_should_fall_back_to_tilde_replacement_when_canonicalization_fails() {
         use utils::home_dir;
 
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -940,6 +949,30 @@ mod tests {
         assert_eq!(expected_logical_dir, &context.logical_dir);
     }
 
+    #[test]
+    fn set_config_method_overwrites_constructor() {
+        let test_path = Path::new("/test_path").to_path_buf();
+        let context = Context::new_with_shell_and_path(
```
domsleee, line 955: would it be better to use `default_context()` instead? i.e.
```rust
let context = default_context();
let mod_context = default_context().set_config(...
```

marcybelardo: Sounds good! Let me do that now~

src/print.rs
```diff
@@ -493,21 +493,18 @@ fn preset_list() -> String {
...
                 format="$character"
                 [character]
                 format=">\n>"
-            }),
-        };
+        });
+
         context.root_config.format = "$character".to_string();
         context.target = Target::Main;
         context.root_config.add_newline = false;
```
davidkna, line 510: `set_config()` can handle `root_config` now here and elsewhere:
```suggestion
        context.target = Target::Main;
```

davidkna (approved):

andytom:
Thank you for your contribution @marcybelardo and thanks for reviewing @davidkna.

PR #5080, opened 2023-04-09
fix: Make STARSHIP_PREEXEC_READY work in 'strict mode'

NickSdot:
#### Description
This PR changes the behavior so that `STARSHIP_PREEXEC_READY` falls back to an empty bound var if the var is not yet bound.

#### Motivation and Context
I am sourcing Starship through a script that uses 'strict mode'  (`set -o nounset`) to exit if an var is unbound.  `STARSHIP_PREEXEC_READY` is unbound if not yet set by `starship_precmd`. 

image

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

#### Checklist:

Not required
There is no test for the behavior and imho this change neither requires any.

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -21,7 +21,7 @@ starship_preexec() {
     local PREV_LAST_ARG=$1
 
     # Avoid restarting the timer for commands in the same pipeline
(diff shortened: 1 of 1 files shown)


NickSdot:
Fixed in https://github.com/starship/starship/pull/5438

PR #5081, opened 2023-04-09
feat(nodejs): Add `expected_version` variable

### LINKED ISSUE #5066: Add variable for the expected Node.js version
## Feature Request

For Node.js, Starship provides the [variable](https://starship.rs/config/#node-js) `version` and the [options](https://starship.rs/config/#node-js) `style` and `not_capable_style`. `version` is the version of the installed Node.js and it is displayed using the format specified in `style` when it matches the version specified in `package.json` (property `$.engines.node`) or using the format specified in `not_capable_style` when it does not match.

This is helpful in itself but it could be more helpful than than. When the version of the interpreter does not match the project requirements, the first execution of any `yarn` command in the directory tells not only that the current version of Node.js is not appropriate for the project but it also tells what version is expected.

Starship could bring the same information without even running `yarn`.
It already knows what is written in `package.json` and has the information in memory. This is how it detects if the current version of Node.js is good or not and decides to render it using `style` or `not_capable_style`. All it needs is to provide a variable similar to `version` that contains the Node.js version required by the project (the value of property `$.engines.node` from `package.json`).

Please add a new variable (it could be named `expected_version`) that contains the value of property `$.engines.node` from `package.json`).

---

A similar solution can be implemented for other languages that declare the version of the interpreter that they need to run (PHP, Python, Ruby etc.)

marcybelardo:
#### Description

Allow users to use `$expected_version` as a variable for `nodejs`. Mirrors the value of `$.engines.node` in the `package.json` file. Currently set to show only if the node version does not match the parameters set in `package.json`, will be hidden otherwise.

#### Motivation and Context

Closes #5066. As raised by the person who opened the issue, a similar implementation of this can also be used for other modules that have declarative versioning.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2832,7 +2832,7 @@ By default the module will be shown if any of the following conditions are met:
 | Variable        | Example    | Description                                                                                                                                                             |
 | --------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | version         | `v13.12.0` | The version of `node`                                                                                                                                                   |
-| engines_version | `>=12.0.0` | `node` version as set in the engines property of `package.json`. Will show if installed node version does not match with the set engines value, is otherwise invisible. |
+| engines_version | `>=12.0.0` | `node` version requirement as set in the engines property of `package.json`. Will only show if the version requirement does not match the `node` version. |
 | symbol          |            | Mirrors the value of option `symbol`                                                                                                                                    |
 | style\*         |            | Mirrors the value of option `style`                                                                                                                                     |
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2829,12 +2829,12 @@ By default the module will be shown if any of the following conditions are met:
 
 ### Variables
 
-| Variable         | Example    | Description                                                              |
-| ---------------- | ---------- | ------------------------------------------------------------------------ |
-| version          | `v13.12.0` | The version of `node`                                                    |
-| expected_version | `>=12.0.0` | Optional `node` version as set in the engines property of `package.json` |
-| symbol           |            | Mirrors the value of option `symbol`                                     |
-| style\*          |            | Mirrors the value of option `style`                                      |
+| Variable        | Example    | Description                                                                                                                                                             |
+| --------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| version         | `v13.12.0` | The version of `node`                                                                                                                                                   |
+| engines_version | `>=12.0.0` | `node` version as set in the engines property of `package.json`. Will show if installed node version does not match with the set engines value, is otherwise invisible. |
+| symbol          |            | Mirrors the value of option `symbol`                                                                                                                                    |
+| style\*         |            | Mirrors the value of option `style`                                                                                                                                     |
 
 *: This variable can only be used as a part of a style string
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2829,11 +2829,12 @@ By default the module will be shown if any of the following conditions are met:
 
 ### Variables
 
-| Variable | Example    | Description                          |
-| -------- | ---------- | ------------------------------------ |
-| version  | `v13.12.0` | The version of `node`                |
-| symbol   |            | Mirrors the value of option `symbol` |
-| style\*  |            | Mirrors the value of option `style`  |
+| Variable         | Example    | Description                                                              |
+| ---------------- | ---------- | ------------------------------------------------------------------------ |
+| version          | `v13.12.0` | The version of `node`                                                    |
+| expected_version | `>=12.0.0` | Optional `node` version as set in the engines property of `package.json` |
+| symbol           |            | Mirrors the value of option `symbol`                                     |
+| style\*          |            | Mirrors the value of option `style`                                      |
 
 *: This variable can only be used as a part of a style string
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2829,12 +2829,12 @@ By default the module will be shown if any of the following conditions are met:
 
 ### Variables
 
-| Variable        | Example    | Description                                                                                                                                                             |
-| --------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| version         | `v13.12.0` | The version of `node`                                                                                                                                                   |
+| Variable        | Example    | Description                                                                                                                                               |
+| --------------- | ---------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| version         | `v13.12.0` | The version of `node`                                                                                                                                     |
 | engines_version | `>=12.0.0` | `node` version requirement as set in the engines property of `package.json`. Will only show if the version requirement does not match the `node` version. |
-| symbol          |            | Mirrors the value of option `symbol`                                                                                                                                    |
-| style\*         |            | Mirrors the value of option `style`                                                                                                                                     |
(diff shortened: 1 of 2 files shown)


src/modules/nodejs.rs
```diff
@@ -71,6 +71,20 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                    let versions_in_range =
+                        check_engines_version(version, get_engines_version(context));
+
+                    match (expected_version, versions_in_range) {
+                        (Some(ver), false) => Some(Ok(ver)),
+                        _ => None,
+                    }
+                }
+                _ => None,
+            })
```
davidkna, line 87: Please re-use the `map` handling `$version`

docs/config/README.md
```diff
@@ -2829,11 +2829,12 @@ By default the module will be shown if any of the following conditions are met:
...
 
-| Variable | Example    | Description                          |
-| -------- | ---------- | ------------------------------------ |
-| version  | `v13.12.0` | The version of `node`                |
-| symbol   |            | Mirrors the value of option `symbol` |
-| style\*  |            | Mirrors the value of option `style`  |
+| Variable         | Example    | Description                                                              |
+| ---------------- | ---------- | ------------------------------------------------------------------------ |
+| version          | `v13.12.0` | The version of `node`                                                    |
+| expected_version | `>=12.0.0` | Optional `node` version as set in the engines property of `package.json` |
```
davidkna, line 2835: Please properly document when this is shown and it is not.

davidkna (changes requested):

src/modules/nodejs.rs
marcybelardo: So instead of a separate variable, it'll show the expected version if the installed one is wrong? Like it'll replace the normal version variable?

davidkna: Please just move the `match`-arm for `$expected_version` into the existing `match` used by `$version`.

marcybelardo: Oh! Okay sure thing

docs/config/README.md
```diff
@@ -2829,11 +2829,12 @@ By default the module will be shown if any of the following conditions are met:
...
 
-| Variable | Example    | Description                          |
-| -------- | ---------- | ------------------------------------ |
-| version  | `v13.12.0` | The version of `node`                |
-| symbol   |            | Mirrors the value of option `symbol` |
-| style\*  |            | Mirrors the value of option `style`  |
+| Variable        | Example    | Description                                                                                                                                                             |
+| --------------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| version         | `v13.12.0` | The version of `node`                                                                                                                                                   |
+| engines_version | `>=12.0.0` | `node` version as set in the engines property of `package.json`. Will show if installed node version does not match with the set engines value, is otherwise invisible. |
```
davidkna, line 2835:
```suggestion
| engines_version | `>=12.0.0` | `node` version requirement as set in the engines property of `package.json`. Will only show if the version requirement does not match the `node` version. |
```

src/modules/nodejs.rs
```diff
@@ -56,19 +57,29 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
                 }
+                "engines_version" => {
+                    let node_ver = nodejs_version.as_deref();
+                    let eng_ver = get_engines_version(context);
+                    let in_engines_range = check_engines_version(node_ver, eng_ver.clone());
+
+                    match (eng_ver, in_engines_range) {
+                        (Some(ver), false) => Some(Ok(ver)),
+                        _ => None,
+                    }
```
davidkna, line 81: On second look, this would look a bit better:
```suggestion
                    let eng_ver = get_engines_version(context)?;
                    let in_engines_range = check_engines_version(node_ver, eng_ver.as_deref());

                    (!in_engines_range).then_some(Ok(ver))
```

To replace the `clone` with `as_deref` please change the signature to `Option<&str>`.

docs/config/README.md
marcybelardo: Got it, committing this!

src/modules/nodejs.rs
marcybelardo: Implemented this!

src/modules/nodejs.rs
```diff
@@ -56,19 +58,26 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
                         module.get_name(),
-                        version,
+                        node_ver,
                         config.version_format,
                     )
                     .map(Ok)
                 }
+                "engines_version" => {
+                    let node_ver = nodejs_version.as_deref();
+                    let eng_ver = get_engines_version(context)?;
```
axiac, line 76: The function `get_engines_version()` reads a file and parses a JSON. Both these operations are expensive. Call it only once, store the value and use it any number of times it is needed.
Also, `check_engines_version()` should be called only once per execution.

davidkna: Compared to stuff like getting the software version, it should be relatively cheap, but I agree that it could be a worthwhile optimization to store the result with a `Lazy<_>` variable.

marcybelardo: I can try these both, gimme a sec

marcybelardo:
I changed the `(!in_engines_range)` check to a match statement because it has to check for `in_engines_range` to be false and for `eng_ver` to be a Some value, so I thought a match would be the best way to do that

src/modules/nodejs.rs
```diff
@@ -35,6 +35,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             .exec_cmd("node", &["--version"])
             .map(|cmd| cmd.stdout)
     });
+    let engines_version = Lazy::new(|| get_engines_version(context));
+    let in_engines_range =
+        check_engines_version(nodejs_version.as_deref(), engines_version.as_deref());
```
davidkna, line 40: Please only get the node/engine version if required in `map` and `map_style`. The `check_engines_version` shouldn't be doing anything expensive, so it's fine for it not to use an extra `Lazy`, just move it down again.

davidkna:
> I changed the `(!in_engines_range)` check to a match statement because it has to check for `in_engines_range` to be false and for `eng_ver` to be a Some value, so I thought a match would be the best way to do that

The version range check function already handles that, there is no need to do that.

marcybelardo:
> > I changed the `(!in_engines_range)` check to a match statement because it has to check for `in_engines_range` to be false and for `eng_ver` to be a Some value, so I thought a match would be the best way to do that
> 
> The version range check function already handles that, there is no need to do that.

Ahhhhh I see what you mean now. Okay, I did it like before, I was just having trouble with unwrapping the eng_ver option, but I think it's okay now

src/modules/nodejs.rs
marcybelardo: Sorry, I'm having trouble understanding this, I'll ask a couple things. `check_engines_version` isn't using a Lazy right now, do you mean for `get_engines_version`? As for getting the engines version only if requirements aren't met, I don't know what it means to get the version only if its required because we need to get the version to check if its required in the first place. Thanks for your time!

davidkna: You should only access `nodejs_version` and `engines_version` in a `Lazy`-context or in `map*` as to only compute their values as required.

marcybelardo: Okay, changed these! Sorry, it's embarrassing to admit but I just didn't fully understand how Lazy worked but after I looked it up it all made sense :P

davidkna (approved):

PR #5112, opened 2023-04-15
refactor(git_status): use `git status --show-stash` instead of `git stash list`

domsleee:
#### Description

`git status --porcelain=v2` handles the `--show-stash` option since [git 2.35.0 (released Jan 2022)](https://github.com/git/git/blob/5d01301f2b865aa8dba1654d3f447ce9d21db0b5/Documentation/RelNotes/2.35.0.txt#L28)
Note that the `--show-stash` option was introduced in [2.14.0](https://github.com/git/git/commit/c1b5d0194b98bd3d0acc9cec7070808fbbe6a740), but it didn't have porcelain v2 handling until later.

`git status` porcelain v2 was first documented in [2.11.0](https://github.com/git/git/commit/1cd828ddc8a61f6bec93303f4bdbcdbf1c261cb7)

git status option | introduced | added to porcelain v2
-- | -- | --
`--show-stash` | [2.14.0](https://github.com/git/git/commit/c1b5d0194b98bd3d0acc9cec7070808fbbe6a740) | [2.35.0](https://github.com/git/git/blob/5d01301f2b865aa8dba1654d3f447ce9d21db0b5/Documentation/RelNotes/2.35.0.txt#L28)
`--branch` | ? | [2.11.0](https://github.com/git/git/commit/d9fc746cd77910a7dec53abfec36df5c699b33c2)
`--ignore-submodules` | [1.7.2](https://github.com/git/git/commit/46a958b3daa1da336683ec82d7f321d0f51b39c8) | 2.11.0?
`--untracked-files=no` | [1.8.2?](https://github.com/git/git/commit/5823eb2b28696bf0eb25f6ca35b303447869f85c) | 2.11.0?

There is also the `--no-ahead-behind` option introduced in [2.17.0](https://github.com/git/git/commit/d0db9edba0050ada6f6eac68061599690d2a4333), perhaps we could leverage that for some configs?

Summary:
* Currently supported git version seems to be >=2.11.0
* With this change, >=2.14.0 would be fine for all except stash, and >=2.35.0 would have full support
* Maybe we should introduce a config in the `git_status` module to not break compatibility? 2.35.0 (Jan 2022) is quite recent. Perhaps a `git_version` string entry, which we can use to determine which features can be used

#### Motivation and Context
`git_status` performance https://github.com/starship/starship/issues/4305
It improves performance by a small amount (5ms)

#### Checklist:
   - We might want to add `git` version info in https://starship.rs/config/#git-status
   - There are existing stash checks, see `shows_stashed` and `shows_stashed_with_count`

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1877,6 +1877,7 @@ You can disable the module or use the `windows_starship` option to use a Windows
 | `ignore_submodules` | `false`                                       | Ignore changes to submodules.                                                                               |
 | `disabled`          | `false`                                       | Disables the `git_status` module.                                                                           |
 | `windows_starship`  |                                               | Use this (Linux) path to a Windows Starship executable to render `git_status` when on Windows paths in WSL. |
+| `git_version`       | `'2.11.0'`                                      | The assumed git version. This module uses `git status --porcelain=2`, introduced in `2.11.0`. Git `2.35.0` has `git status --show-stash`, which is a bit faster. |
 
 ### Variables
 
--- a/src/configs/git_status.rs
+++ b/src/configs/git_status.rs
@@ -26,6 +26,7 @@ pub struct GitStatusConfig<'a> {
     pub disabled: bool,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub windows_starship: Option<&'a str>,
+    pub git_version: &'a str,
 }
 
 impl<'a> Default for GitStatusConfig<'a> {
@@ -48,6 +49,7 @@ impl<'a> Default for GitStatusConfig<'a> {
             ignore_submodules: false,
             disabled: false,
             windows_starship: None,
+            git_version: "2.11",
         }
     }
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -11,6 +11,7 @@ use gix::{
     sec::{self as git_sec, trust::DefaultForLevel},
     state as git_state, Repository, ThreadSafeRepository,
 };
+use log::debug;
 use once_cell::sync::OnceCell;
 #[cfg(test)]
 use std::collections::HashMap;
@@ -251,6 +252,7 @@ impl<'a> Context<'a> {
     pub fn get_repo(&self) -> Result<&Repo, Box<gix::discover::Error>> {
         self.repo
             .get_or_try_init(|| -> Result<Repo, Box<gix::discover::Error>> {
+                debug!("get_repo start");
                 // custom open options
                 let mut git_open_opts_map =
                     git_sec::trust::Mapping::<gix::open::Options>::default();
@@ -297,6 +299,7 @@ impl<'a> Context<'a> {
                 let branch = get_current_branch(&repository);
                 let remote = get_remote_repository_info(&repository, branch.as_deref());
                 let path = repository.path().to_path_buf();
+                debug!("get_repo end");
                 Ok(Repo {
                     repo: shared_repo,
                     branch,
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -10,6 +10,7 @@ use std::{
     io::Write,
     path::{Path, PathBuf},
     sync::{Mutex, RwLock},
+    time::Instant,
 };
 
 pub struct StarshipLogger {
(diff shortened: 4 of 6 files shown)


davidkna:
Generally, I would prefer implementing stash support in (or with) gitoxide.

domsleee:
What do you mean by other library features?

I think we could write a competitive replacement for the `git_status` information in the future, but for the time being, I think `--show-stash` is the most performant option

I'm looking into performance improvements for the git_status module, here are my logs using `STARSHIP_TIMINGS=1` that I added :)
```
[DEBUG  0.000044] - (starship): running with 8 threads
[DEBUG  0.005418] - (starship::context): get_repo start
[DEBUG  0.012600] - (starship::modules::git_status): waiting for get_repo
[DEBUG  0.056914] - (starship::context): get_repo end
[DEBUG  0.057620] - (starship::modules::git_status): get_repo OK
[DEBUG  0.058114] - (starship::modules::git_status): git status start
[DEBUG  0.058126] - (starship::modules::git_status): git stash list start
[DEBUG  0.106295] - (starship::modules::git_status): git stash list end
[DEBUG  0.120216] - (starship::modules::git_status): git status end
```

So my performance ideas:
* ~50ms: Instead of using `context.get_repo()`, write a more efficient `context.is_git_repo()`. It might be a new method in `gitoxide` that uses `GIT_DIR` / `GIT_DISCOVERY_ACROSS_FILESYSTEM` etc, but without actually loading the repo. All the time is spent handling `refs/replacements` in `ThreadSafeRepository#open_from_paths`
* 7ms: Change the order that the modules are executed, it is more efficient in my use case if `git_status` is executed first. Other people might have different slowest modules, maybe this could be configurable?
* 3ms: Use `--show-stash`

There may be some optimisation for gitoxide 
```
d0 | main                     | region_enter | r1  |  0.029968 |           | index        | label:do_read_index .git/index
d0 | main                     | region_leave | r1  |  0.045798 |  0.015830 | index        | label:do_read_index .git/index
```

I would also note that `starship timings` is misleading here, because `context.get_repo()` has already been evaluated in another module (e.g. `directory`). So the time reported in `starship timings` doesn't represent what happens when the prompt is run with more than one thread. Separate problem though

davidkna:
50ms for just `context.get_repo` seems unusually long to me.

I believe the gitoxide project was aiming to provide the features to allow powering a `git_status` replacement in the first half of this year. I would rather not merge a feature like this, that even includes a new compatibility option, which could end up being thrown away in just a few months. Likewise, I expect most of the speed-up of the changes in this PR can be attained with gitoxide by likewise avoiding the extra git executable launch.

> What do you mean by other library features?

I am not very familiar with the inner workings of the stash, but I simply meant that the stash count is perhaps already accessible using the existing set of gitxoide library features, albeit less conveniently than an explicit stash API. On the other hand, this sort of thing would be a good candidate for upstreaming either way.

domsleee:
> 50ms for just context.get_repo seems unusually long to me.

The repo I'm using is perhaps unreasonably large (the `.git` folder is 38GB), which is why I'm seeing such performance issues :)

To me the `git_status` module is such a pain point for performance that it may be worth maintaining two versions of the module, a `gitoxide` and a `git` cli version. The `git` cli is still under development and there have been new features coming out that have helped performance a lot in large repos like `core.fsmonitor` and `feature.manyFiles`, it seems reasonable to play both sides here?

> Likewise, I expect most of the speed-up of the changes in this PR can be attained with gitoxide by likewise avoiding the extra git executable launch.

I mean, maybe? The `git` instance has a lot loaded from the git status call, `gitoxide` would need to load everything itself right? I guess my focus has shifted a bit to that 50ms that is from `context.get_repo()`

Anyway I would be interesting in helping out with gitoxide as well, these are just my observations 👍

davidkna:
> To me the `git_status` module is such a pain point for performance that it may be worth maintaining two versions of the module, a `gitoxide` and a `git` cli version. The `git` cli is still under development and there have been new features coming out that have helped performance a lot in large repos like `core.fsmonitor` and `feature.manyFiles`, it seems reasonable to play both sides here?

I would rather not carry any extra maintenance burden, and performance so far has mostly been competitive with git as far as I can tell (with some exceptions, as observed here). Beyond that, I also remember seeing references to things like untracked cache and `core.fsmonitor` in the gitoxide roadmap.

PR #5114, opened 2023-04-18
Add `elixir --short-version` for better performance

### LINKED ISSUE #4825: Read the "elixir" shell script for its version number before running it

darwin67:
#### Description

Add `elixir --short-version` as a way to obtain the elixir version if `$otp_version` is not included in the format.

#### Motivation and Context

Closes #4825 & #4831

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1302,7 +1302,7 @@ heuristic = false
 
 ## Elixir
 
-The `elixir` module shows the currently installed version of [Elixir](https://elixir-lang.org/) and [Erlang/OTP](https://erlang.org/doc/).
+The `elixir` module shows the currently installed version of [Elixir](https://elixir-lang.org/) and [Erlang/OTP](https://erlang.org/doc/) if `$otp_version` variable is set on format.
 By default the module will be shown if any of the following conditions are met:
 
 - The current directory contains a `mix.exs` file.
--- a/src/modules/elixir.rs
+++ b/src/modules/elixir.rs
@@ -23,7 +23,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let versions = Lazy::new(|| get_elixir_version(context));
+    let versions = Lazy::new(|| get_elixir_version(context, &config));
 
     let parsed = StringFormatter::new(config.format).and_then(|formatter| {
         formatter
@@ -51,7 +51,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "otp_version" => versions
                     .deref()
                     .as_ref()
-                    .map(|(otp_version, _)| otp_version.to_string())
+                    .map(|(otp_version, _)| match otp_version {
+                        None => "".to_string(),
+                        Some(version) => version.to_string(),
+                    })
                     .map(Ok),
                 _ => None,
             })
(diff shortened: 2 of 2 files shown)


src/modules/elixir.rs
```diff
@@ -23,7 +23,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let versions = Lazy::new(|| get_elixir_version(context));
+    let versions = Lazy::new(|| get_elixir_version(context, &config));
```
davidkna, line 26: Please just handle both paths separately.

docs/config/README.md
```diff
@@ -1302,7 +1302,7 @@ heuristic = false
 
 ## Elixir
 
-The `elixir` module shows the currently installed version of [Elixir](https://elixir-lang.org/) and [Erlang/OTP](https://erlang.org/doc/).
+The `elixir` module shows the currently installed version of [Elixir](https://elixir-lang.org/) and [Erlang/OTP](https://erlang.org/doc/) if `$otp_version` variable is set on format.
```
davidkna, line 1305: Please remove this change and just update the default format string (in the table!).

src/modules/elixir.rs
```diff
@@ -69,22 +72,35 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
 }
 
-fn get_elixir_version(context: &Context) -> Option<(String, String)> {
-    let output = context.exec_cmd("elixir", &["--version"])?.stdout;
-
+fn get_elixir_version(
+    context: &Context,
+    config: &ElixirConfig,
+) -> Option<(Option<String>, String)> {
+    let output = match config.format.contains("$otp_version") {
```
davidkna, line 79: The `StringFormatter` has an API for checking if a variable is used. `$otp_version` is not the only valid synatx.

src/modules/elixir.rs
```diff
@@ -69,22 +72,35 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+        true => context.exec_cmd("elixir", &["--version"])?.stdout,
+        false => context.exec_cmd("elixir", &["--short-version"])?.stdout,
```
davidkna, line 81: It would be nice to make an attempt at supporting older `elixir` versions in some manner. A possible method that has been raised is to just read the version string from the `elixir` executable script file.

davidkna (changes requested):

src/modules/elixir.rs
darwin67: Which one? `get_variables`?

davidkna: That should be it.

docs/config/README.md
darwin67: Update the default format to remove `$otp_version` or are you suggesting something else?

davidkna: I'm not completely, but your PR description read like you were going to do that.

darwin67: Yeah, I was initially going to remove `$otp_version` and was playing with it but might change my mind.
That's why I put it in draft mode, so wasn't expecting to get feedback yet.

src/modules/elixir.rs
darwin67: It becomes like this though unless I'm missing something.

```
let versions = Lazy::new(|| match StringFormatter::new(config.format) {
        Ok(fmt) => match fmt.get_variables().contains("otp_version") {
            true => get_elixir_version(context),
            false => get_elixir_version_short(context),
        },
        Err(_) => return None,
    });

```

Which seems over complicated for just trying to check if a variable exists in the format string or not.

src/modules/elixir.rs
darwin67: I'll need to take a closer look at this.
Don't recall that being a thing.

src/modules/elixir.rs
davidkna: You can skip using the `Lazy<_>` `versions` at all if `!fmt.get_variables().contains("otp_version")`.

docs/config/README.md
darwin67: Not changing the default format for now.

darwin67:
@davidkna the older version support and checking variables is making this unnecessarily complex.

I found another PR attempt here, and adding another variable would be easier.
https://github.com/starship/starship/pull/3884#issuecomment-1140482060

What do you think?

matchai:
I'd rather we have added complexity in our code than in our docs or in user configuration.

darwin67:
Alright, let me ask at the Elixir forum and see if people know anything about doing some kind of backward compatible version of `--short-version`.

Reading from file is a last resort since that will blow up the complexity exponentially.

darwin67:
Closing this as I no longer have time to look into this in depth.
Considering that Elixir only supports until the last 5 minor version, it might be easier to just wait till < 1.13 versions are no longer supported.

PR #5115, opened 2023-04-18
ci: remove actions-rs actions

### LINKED ISSUE #4725: Migrating away from action-rs
Sadly, it seems at this point that [actions-rs is not coming back](https://github.com/actions-rs/toolchain/issues/216), having not seen a release in almost 1000 days and no activity for over two years. While we're not obligated to migrate *right now*, it seems that we'll have to move off it sooner or later.

This issue is meant to serve as a record of information of discussions held in other locations (e.g. the Discord) and to hopefully provide enough information that whoever ends up tackling this doesn't have to go on a weekend-long internet expedition just to be able to create a feasible solution.

Right now, we use three actions from actions-rs:
  - actions-rs/toolchain
  - actions-rs/cargo
  - actions-rs/audit-check

The `toolchain` action is almost always used to install a toolchain or toolchain components. The `cargo` action is used to run cargo commands, but with easy cross-compilation (which we do take advantage of), and with appropriate output being displayed in the GitHub web interface. We will likely need to replace each instance of this action with a separate or specialized action. Finally, audit-check has already been forked by the Rustsec team, so we can likely just do a drop-in replacement there.

@davidkna has suggested several alternatives for these actions:
  - https://github.com/dtolnay/rust-toolchain for simple toolchain installation
  - https://github.com/rustsec/audit-check for security (this is forked from actions-rs/audit-check by the Rustsec team, but it's unclear how many resources are being dedicated to it at the moment)
  - https://github.com/giraffate/clippy-action for replacing clippy
  - https://github.com/marketplace/actions/reviewdog-suggester-suggest-any-code-changes-based-on-diff-with-reviewdog for formatting
  - https://github.com/EmbarkStudios/cargo-deny-action as an alternate replacement for audit-check and also to spot license violations (which we have been doing manually)

#### workflow.yml

- Used [on line 32](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L32) to install a minimal cargo to run `cargo fmt --all -- --check`. Similarly used on lines [79](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L79), [101](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L101), [123](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L123), and [145](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L145) to install toolchains to run basic command line applications.
- Used [on line 54](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L54) to install a toolchain to [use the clippy action with](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L65) on line 65. The clippy action is run as a command of actions-rs/cargo and likely cannot be replicated by just running cargo clippy (I think it uploads results into the GitHub web interface, but I'm not 100% certain at the moment).
- Used on [line 182](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L182) to install a toolchain on multiple platform with the `llvm-tools-preview` components (which I believe are used for code coverage testing)
- Used on [line 225](https://github.com/starship/starship/blob/master/.github/workflows/workflow.yml#L225) to run `cargo wix`, which builds a windows installer.

#### security-audit.yml

Used on lines [21](https://github.com/starship/starship/blob/master/.github/workflows/security-audit.yml#L21) and [27](https://github.com/starship/starship/blob/master/.github/workflows/security-audit.yml#L27) to run the audit-check action.

#### release.yml

- Used on [line 95](https://github.com/starship/starship/blob/master/.github/workflows/release.yml#L95) to install the platform-specific toolchain.
- Used on [line 112](https://github.com/starship/starship/blob/master/.github/workflows/release.yml#L112) to run `cargo build` (with cross) and [line 121](https://github.com/starship/starship/blob/master/.github/workflows/release.yml#L121) to build the installer.
- Used on [line 287](https://github.com/starship/starship/blob/master/.github/workflows/release.yml#L287) to install a toolchain to run `cargo publish`.

davidkna:
#### Description

This PR replaces the remaining actions-rs actions with the alternatives discussed in #4725.

The clippy action is set to the latest revision to pull in a commit, which makes setting the GitHub token optional, pulling it from the environment instead. Using something like a personal access token would allow it to make reviews with comments containing suggestions, but as far as I can tell from reading the documentation, any secret would be in danger of being leaked.

A major difference with the new clippy action is that it is only considering the added lines (by default). I feel like this is an improvement, because it should prevent new rust releases from causing clippy CI failures in PRs that don't touch any related lines.

#### Motivation and Context

Closes #4725

#### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -92,11 +92,9 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Setup | Rust
-        uses: actions-rs/toolchain@v1.0.7
+        uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
-          override: true
-          profile: minimal
           target: ${{ matrix.target }}
 
       - name: Setup | Install cargo-wix [Windows]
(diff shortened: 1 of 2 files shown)


andytom (approved):
LGTM

PR #5129, opened 2023-04-21
feat(rust): show date for nightly toolchain

### LINKED ISSUE #5111: rust module: show date for nightly toolchain
## Feature Request

For nightly rust users, it's common to pin a specific date of the toolchain and don't care about the version number. So it might be better to show date for nightly.

image

jgoday:
#### Description

Show toolchain date for nightly instead of rust version. 

Would it be better to have an option in RustConfig to enable this behaviour ? , something like:
```
pub struct RustConfig<'a> {
...
     pub show_nightly_date: bool,
...
}
```

#### Motivation and Context

Closes #5111

#### Checklist:

### DIFF
--- a/src/modules/rust.rs
+++ b/src/modules/rust.rs
@@ -11,6 +11,7 @@ use crate::configs::rust::RustConfig;
 use crate::formatter::{StringFormatter, VersionFormatter};
 use crate::utils::create_command;
 use home::rustup_home;
+use regex::Regex;
 
 use once_cell::sync::OnceCell;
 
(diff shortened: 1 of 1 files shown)


src/modules/rust.rs
```diff
@@ -707,7 +715,7 @@ version = "12"
         let rustc_nightly = "rustc 1.34.0-nightly (b139669f3 2019-04-10)";
         assert_eq!(
             format_rustc_version(rustc_nightly, config.version_format),
-            Some("v1.34.0-nightly".to_string())
+            Some("nightly-2019-04-10".to_string())
```
taiki-e, line 718: I think this is not correct because the actual rustup toolchain version and commit date are consistently off by one. (https://github.com/rust-lang/rust/issues/51533)

src/modules/rust.rs
```diff
@@ -384,17 +385,24 @@ fn execute_rustc_version(context: &Context) -> Option<String> {
...
+    if rustc_version.contains("nightly") {
+        let re = Regex::new(r"rustc .* \(.* (?P<date>.*)\)").unwrap();
+        let caps = re.captures(rustc_version)?;
 
-    match VersionFormatter::format_version(version, version_format) {
-        Ok(formatted) => Some(formatted),
-        Err(error) => {
-            log::warn!("Error formatting `rust` version:\n{}", error);
-            Some(format!("v{version}"))
+        caps.name("date").map(|x| format!("nightly-{}", x.as_str()))
```
xxchan, line 392:
```suggestion
        caps.name("date").map(|x| {
            let date = x.as_str();
            let date = chrono::NaiveDate::parse_from_str(date, "%Y-%m-%d")
                .expect(&format!("invalid date format: {date}"))
                + chrono::Duration::days(1);
            format!("nightly-{}", date.format("%Y-%m-%d"))
        })
```

PR #5141, opened 2023-04-26
build(deps): upgrade gitoxide to v0.44 for performance improvements during discovery

Byron:
Please note that there is a new `dot_git_only` option which would further
speedup discovery. On even moderatly fast disks that probably not going
to make a difference, but it will on incredibly slow (networked) disks.

See https://github.com/helix-editor/helix/issues/6867 for reference.


davidkna:
LGTM, but would you mind bumping to the `.1`?

davidkna (approved):

PR #5175, opened 2023-05-08
build(deps): disable unnecessary/unused regex features

jirutka:
#### Description

Disable unnecessary/unused regex features.

#### Motivation and Context

To reduce the binary size (by ~400 kB).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,7 @@ pest_derive = "2.6.0"
 quick-xml = "0.28.2"
 rand = "0.8.5"
 rayon = "1.7.0"
-regex = "1.8.1"
+regex = { version = "1.8.1", default-features = false, features = ["perf", "std", "unicode-perl"] }
 rust-ini = "0.19.0"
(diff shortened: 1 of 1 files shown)


andytom (approved):

Cargo.toml
```diff
@@ -67,7 +67,7 @@ pest_derive = "2.6.0"
 quick-xml = "0.28.2"
 rand = "0.8.5"
 rayon = "1.7.0"
-regex = "1.8.1"
+regex = { version = "1.8.1", default-features = false, features = ["perf", "std", "unicode-perl"] }
```
davidkna, line 70: Is there a reason you have kept `unicode-perl` enabled, unlike the other unicode features?

jirutka: Unfortunately, this is needed even for `\d`, `\w` etc. that are used in starship code.

davidkna: The docs hint at that, but is that really true? The tests run fine without `unicode-perl`.

PR #5178, opened 2023-05-11
add Flamer preset

atahabaki:
I'm a bit lazy

#### Motivation and Context

Closes #

ss

#### Checklist:


mickimnet:
@atahabaki, currently the _Presets_ section, only contains _Starship_ configurations, that are also included in the `docs/.vuepress/public/presets` folder. Those are automatically included as preset in the _Starship_ binary.

In my opinion, there are two ways to go from here:
1. either wait for the core maintainers to decide about including external configurations,
2. or extend your pull request to include your configuration in the preset section.

Take care,
Mick

PR #5188, opened 2023-05-17
refactor: update the nushell init file and make it valid module and overlay

amtoine:
i've been using Nushell for quite some time now and started combining it with `starship` not so long a go :blush: 

i thought i could try to *improve* the init file generated by `starship init nu` to make it more aligned with the current feel of Nushell and the new features.

> **Note**
> i've tried to make the commits as atomic and clear as possible!
> they should be best reviewed one at a time :+1: 

#### Description
this PR 
- breaks the long commands into multiple lines for readability and maintainability
- fixes some of the format of commands and closures
- simplifies the `render_right_prompt_on_last_line` block with the new `?` syntax
- wrap all the `let-env`s in an `export-env { load-env { ... }}` block

the last item in the list above is particularily interesting for the following reason: **we are thinking of removing the `source` command from Nushell** :scream: 

with this PR, `init.nu` would be both a valid (1) script, (2) module and (3) overlay :muscle: 

> **Note**  
> in my config, i save the output of `starship init nu` as `starship.nu` and add the directory in which i save it to `NU_LIB_DIRS`, so that i can just `<cmd> starship.nu` without giving a path :ok_hand: 
> > see my [`env.nu`](https://github.com/goatfiles/dotfiles/blob/3e9b8c1/.config/nushell/env.nu#L190-L195)
> the following still applies by adding the path to the `init.nu` file :thumbsup: 
 
the last bullet point (and commit) allows, without any loss of features compared to before this PR, to
- `source starship.nu` as before
- `use starship.nu`
- `overlay use starship.nu` which is particularly nice for two reasons
    - we can see `starship` being activated in `overlay list`
    - we can `overlay hide starship` to get the default prompt temporarily and reactivate it once done with `overlay use starship` from the REPL of Nushell

#### Motivation and Context
i did not see any related issue nor PR, this is just an improvement attempt :wink: 

```
$nothing
```
this should not affect the look and feel of the prompt at all :relieved: 

i've just
```
cargo install --path /path/to/amtoine/starship
```
and noticed it did not change the look and feel of the prompt as expected.

#### Checklist:
- [x] I have updated the tests accordingly. (the functionality of `init.nu` should not have changed at all)

### DIFF
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -1,6 +1,8 @@
 let-env STARSHIP_SHELL = "nu"
 let-env STARSHIP_SESSION_KEY = (random chars -l 16)
-let-env PROMPT_MULTILINE_INDICATOR = (^::STARSHIP:: prompt --continuation)
+let-env PROMPT_MULTILINE_INDICATOR = (
+    ^::STARSHIP:: prompt --continuation
+)
 
 # Does not play well with default character module.
 # TODO: Also Use starship vi mode indicators?
@@ -8,8 +10,12 @@ let-env PROMPT_INDICATOR = ""
 
 let-env PROMPT_COMMAND = { ||
     # jobs are not supported
-    let width = (term size).columns
-    ^::STARSHIP:: prompt $"--cmd-duration=($env.CMD_DURATION_MS)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=($width)"
+    (
+        ^::STARSHIP:: prompt
+            $"--cmd-duration=($env.CMD_DURATION_MS)"
+            $"--status=($env.LAST_EXIT_CODE)"
+            $"--terminal-width=((term size).columns)"
+    )
 }
 
 # Whether we have config items
@@ -22,6 +28,11 @@ let-env config = if $has_config_items {
 }
 
 let-env PROMPT_COMMAND_RIGHT = { ||
-    let width = (term size).columns
-    ^::STARSHIP:: prompt --right $"--cmd-duration=($env.CMD_DURATION_MS)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=($width)"
+    (
+        ^::STARSHIP:: prompt
+            --right
+            $"--cmd-duration=($env.CMD_DURATION_MS)"
+            $"--status=($env.LAST_EXIT_CODE)"
+            $"--terminal-width=((term size).columns)"
+    )
 }
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -1,33 +1,35 @@
-let-env STARSHIP_SHELL = "nu"
-let-env STARSHIP_SESSION_KEY = (random chars -l 16)
-let-env PROMPT_MULTILINE_INDICATOR = (
-    ^::STARSHIP:: prompt --continuation
-)
-
-# Does not play well with default character module.
-# TODO: Also Use starship vi mode indicators?
-let-env PROMPT_INDICATOR = ""
-
-let-env PROMPT_COMMAND = {||
-    # jobs are not supported
-    (
-        ^::STARSHIP:: prompt
-            --cmd-duration $env.CMD_DURATION_MS
-            $"--status=($env.LAST_EXIT_CODE)"
-            --terminal-width (term size).columns
+export-env { load-env {
+    STARSHIP_SHELL: "nu"
+    STARSHIP_SESSION_KEY: (random chars -l 16)
+    PROMPT_MULTILINE_INDICATOR: (
+        ^::STARSHIP:: prompt --continuation
     )
-}
 
-let-env config = ($env.config? | default {} | merge {
-    render_right_prompt_on_last_line: true
-})
+    # Does not play well with default character module.
+    # TODO: Also Use starship vi mode indicators?
+    PROMPT_INDICATOR: ""
 
-let-env PROMPT_COMMAND_RIGHT = {||
-    (
-        ^::STARSHIP:: prompt
-            --right
-            --cmd-duration $env.CMD_DURATION_MS
-            $"--status=($env.LAST_EXIT_CODE)"
-            --terminal-width (term size).columns
-    )
-}
+    PROMPT_COMMAND: {||
+        # jobs are not supported
+        (
+            ^::STARSHIP:: prompt
+                --cmd-duration $env.CMD_DURATION_MS
+                $"--status=($env.LAST_EXIT_CODE)"
+                --terminal-width (term size).columns
+        )
+    }
+
+    config: ($env.config? | default {} | merge {
+        render_right_prompt_on_last_line: true
+    })
+
+    PROMPT_COMMAND_RIGHT: {||
+        (
+            ^::STARSHIP:: prompt
+                --right
+                --cmd-duration $env.CMD_DURATION_MS
+                $"--status=($env.LAST_EXIT_CODE)"
+                --terminal-width (term size).columns
+        )
+    }
+}}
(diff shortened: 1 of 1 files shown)


amtoine:
regarding the documentation part, i've located https://github.com/starship/starship/blob/master/docs/README.md and https://github.com/starship/starship/blob/master/README.md, are these the files to modify? :yum:

src/init/starship.nu
```diff
@@ -1,27 +1,35 @@
...
-    let width = (term size).columns
-    ^::STARSHIP:: prompt $"--cmd-duration=($env.CMD_DURATION_MS)" $"--status=($env.LAST_EXIT_CODE)" $"--terminal-width=($width)"
-}
+    PROMPT_COMMAND: {||
+        # jobs are not supported
+        (
+            ^::STARSHIP:: prompt
+                --cmd-duration $env.CMD_DURATION_MS
+                $"--status=($env.LAST_EXIT_CODE)"
+                --terminal-width (term size).columns
```
davidkna, line 18: The `--flag=` syntax has the advantage that it may handle empty values, better I would prefer to keep using it.

davidkna:
Does this have any impact on the supported nushell version range, due to the use of `?`?

> regarding the documentation part, i've located https://github.com/starship/starship/blob/master/docs/README.md and https://github.com/starship/starship/blob/master/README.md, are these the files to modify? 😋

The installation script `install/install.sh` also contains some instructions.

src/init/starship.nu
amtoine: i think Nushell handles *empty values* out-of-the-box.
there is the `null` value which is passed if an argument is not defined, e.g. with the following mockup of the init script,
```bash
def starship [-a: any, -b: any, -c: any] {
    print $"a: ($a)"
    print $"b: ($b)"
    print $"c: ($c)"
}

def init [-a: any, -b: any, -c: any] {
    starship -a $a -b $b -c $c
}
```
`init ...` with any combination of options works just fine :relieved: 

also i think `$env.CMD_DURATION_MS`, `$env.LAST_EXIT_CODE` and `term size` will always return something :+1: 

i think this is safe and more clean, you tell me of course :yum:

amtoine:
> Does this have any impact on the supported nushell version range, due to the use of `?`?
> 
i think it does, yes, very good remark :open_mouth: 

it's related to https://github.com/nushell/nushell/pull/8379 and https://github.com/nushell/nushell/pull/8554, so you'd need Nushell version `0.78.0` or revision nushell/nushell@6a274b860 :+1: 

> The installation script `install/install.sh` also contains some instructions.
>
thanks, i'll have a look right now :partying_face:

amtoine:
> The installation script `install/install.sh` also contains some instructions.
>
i do not see what `install/install.sh` tells about the website of Starship? :open_mouth: 

my original question about the doc was: where should i edit to update the [Nushell section](https://starship.rs/#nushell) of the doc? :yum:

davidkna:
> my original question about the doc was: where should i edit to update the [Nushell section](https://starship.rs/#nushell) of the doc? 😋

You should update the nushell docs in all the files:
- `install/install.sh`
- `docs/README.md`
- `README.md`

amtoine:
@davidkna 
gotcha :yum:
doing it :relieved:

amtoine:
i've changed the version requirement in cda3f01f68a8d83d2016eee0b030851f800bbf69 and updated the documentation in d903115f4b28ef294315ed024c4cc6ad58bbc1aa :muscle: 

i hope you will like the changes :innocent:

davidkna:
If going forward `overlay use` will require use to keep up with one less breaking nushell change, I would prefer just changing the docs to use that. That being said, I would prefer to keep any compatibility measures for `source` you implemented for now.

amtoine:
> If going forward `overlay use` will require use to keep up with one less breaking nushell change, I would prefer just changing the docs to use that.
>
do you mean we should only mention `overlay use` in the doc and let the Nushell pros use `use` or `source` on their own if they prefer? :yum: 

> That being said, I would prefer to keep any compatibility measures for `source` you implemented for now.
>
as long as we keep the `source` command in Nushell, this will be 100% compatible with either `source`, `use` and `overlay use`, no breaking change on that front :+1: 

it's just the much cleaner `?` notation that's a bit more recent :wink:

davidkna:
> do you mean we should only mention `overlay use` in the doc and let the Nushell pros use `use` or `source` on their own if they prefer? 😋

Yes, but it might make sense to add some explanations for that in the init file comments.

amtoine:
mm i'm actually super confused here :sweat_smile: 

i've just pushed https://github.com/amtoine/starship/commit/2206296cfcbc97c8d0847d1b7222b119f0a5aa96 to the PR branch but it does not appear here :eyes:

davidkna:
[Apparently, Pull-Requests are having a major outage right now](https://www.githubstatus.com). The commit will hopefully appear soon.

davidkna:
LGTM, but please also update the installation script again.

amtoine:
> LGTM, but please also update the installation script again.
>
it took me some time to understand how the `elvish` / Nushell part of `install.sh` worked.
i tried to follow as much as possible what's written in the README, i hope it'll be ok :relieved:

install/install.sh
```diff
@@ -337,12 +337,15 @@ print_install() {
       nushell )
         # shellcheck disable=SC2088
         config_file="${BOLD}your nu config file${NO_COLOR} (find it by running ${BOLD}\$nu.config-path${NO_COLOR} in Nushell)"
-        config_cmd="mkdir ~/.cache/starship
-        starship init nu | save -f ~/.cache/starship/init.nu
-        source ~/.cache/starship/init.nu"
+        config_cmd="\`use ~/.cache/starship/init.nu\`"
```
davidkna, line 340:
```suggestion
        config_cmd="use ~/.cache/starship/init.nu"
```

install/install.sh
```diff
@@ -337,12 +337,15 @@ print_install() {
...
         warning="${warning} This will change in the future.
-  Only Nushell v0.73 or higher is supported.
-  Add the following to the end of ${BOLD}your Nushell env file${NO_COLOR} (find it by running ${BOLD}\$nu.env-path${NO_COLOR} in Nushell): \"mkdir ~/.cache/starship; starship init nu | save -f ~/.cache/starship/init.nu\""
+  Only Nushell v0.78 or higher is supported.
+
+  Add the following to the end of ${BOLD}your Nushell env file${NO_COLOR} (find it by running ${BOLD}\$nu.env-path${NO_COLOR} in Nushell):
+  \`\`\`nu
+  mkdir ~/.cache/starship
+  starship init nu | save -f ~/.cache/starship/init.nu
+  \`\`\`"
```
davidkna, line 348:
```suggestion
  Only Nushell v0.78 or higher is supported.
  Add the following to the end of ${BOLD}your Nushell env file${NO_COLOR} (find it by running ${BOLD}\$nu.env-path${NO_COLOR} in Nushell):

	mkdir ~/.cache/starship
	starship init nu | save -f ~/.cache/starship/init.nu
```

davidkna (commented):
I have added some suggestions to keep the styling consistent with other shells.

install/install.sh
amtoine: should be good in 8d43f2ac2254bd054ddac4d806799703bcfa6512 :yum:

amtoine:
> I have added some suggestions to keep the styling consistent with other shells.
>
of course, my bad :relieved: 

`crates-ci/typos` is going nuts these days, isn't it? :sweat_smile: 
it's also failing for crate names in Nushell, saying the `ratatui` crate should be spelled "ratatouille" :laughing:

davidkna (approved):

amtoine:
:partying_face: 

thanks @davidkna for the review :blush:

davidkna:
Thanks for contributing the init changes!

PR #5193, opened 2023-05-20
feat(container): Allow use of podman container name instead of image name (continuation of original PR)

davidscholberg:
Since progress on https://github.com/starship/starship/pull/4243 seems to have stalled, I decided to try to move this feature forward in this new PR.

#### Description
These changes add a configuration parameter to the container module that allows users to specify starship to use the podman container name instead of the image name.

I took the commits from https://github.com/starship/starship/pull/4243 and made some minor changes/additions on top of them:

1. I resolved the merge conflicts with the master branch.
2. I implemented one of the changes requested in the review by @davidkna in https://github.com/starship/starship/pull/4243 (I have questions about the other two requests, which I ask below).
3. I updated the image name template in the unit test to more accurately reflect image names in the wild.

Regarding the other changes requested by @davidkna:

> Would `shellwords::split` work here?

I don't see why it wouldn't work, but I don't know why this would be preferred over merely trimming the quotes off the ends in this particular use case.

> Perhaps call this `prefer_container_name` instead, your code seems to include a fallback.

I don't understand the reasoning here. The meaning of `use_container_name` seems clear enough to me for what it does. Is there some standard or convention for using "prefer" when there's a fallback?

#### Motivation and Context
Closes: #3592
Closes: #4243
Closes: #5195

I tested with the new option absent, as well as with the new option set to true and false.

#### Checklist:


jeaye:
I'd love to have this, for distrobox usage. It's so seamless that I often forget which terminals I have are in a container. The existing container check helps point out that it's in a container now, but my recent workflows have me in multiple containers at a time, so remembering which is which is now a problem. Thank you for making these changes.

Is there any interest from starship in merging it?

PR #5196, opened 2023-05-21
feat(hostname): add detect_env_vars as option

mickimnet:
Hello everybody,

#### Motivation and Context

I don't use the default `hostname` module and instead use a custom module because I frequently use terminal multiplexers like `tmux`, `screen`, or `Zellji`. These multiplexers display the hostname in the status bar, so I don't want or need to see it in the terminal as well.

#### Description

To achieve this, I added the `context::detect_env_vars` function to the `hostname` module and also added a new `context::has_no_negative_env_vars` function to it.

This is the first time I've written anything other than shell scripts, so please bear with me🫣 I couldn't have done it without the invaluable help of my friend and experienced Rustacean, @dsander 💛 

#### Option combinations

I created a truth table to outline the expected behavior of the changes:

_Legend:_

* `detect_env_vars`:
  + `positive`: a name of an environment variable is configured and the variable is set
  + `empty`: the array is empty
* `detect_env_vars = none:` none of the entered environment variables are set
* `detect_env_vars = negative`: the name of an environment variable is prefixed with an "!" and the variable is set

| ssh_only | SSH_CONNECTION | detect_env_vars | result        |
| :------: | :------------: | :-------------- | :------------ |
|   true   |      true      | positive        | show hostname |
|   true   |      true      | empty           | show hostname |
|   true   |      true      | none            | show hostname |
|   true   |      true      | negative        | show hostname |
|          |                |                 |               |
|   true   |     false      | positive        | hide hostname |
|   true   |     false      | empty           | hide hostname |
|   true   |     false      | none            | hide hostname |
|   true   |     false      | negative        | hide hostname |
|          |                |                 |               |
|  false   |      true      | positive        | show hostname |
|  false   |      true      | empty           | show hostname |
|  false   |      true      | none            | hide hostname |
|  false   |      true      | negative        | hide hostname |
|          |                |                 |               |
|  false   |     false      | positive        | show hostname |
|  false   |     false      | empty           | show hostname |
|  false   |     false      | none            | hide hostname |
|  false   |     false      | negative        | hide hostname |

**Hint:** If `ssh_only` is set to `false` and you still want to see the `hostname` in an `ssh` connection, you will need to add `SSH_CONNECTION` to the `detect_env_vars` array. My use case can be achieved with the following configuration:

```toml
[hostname]
ssh_only = false
detect_env_vars = [ '!TMUX', 'SSH_CONNECTION' ]
```

#### Checklist:

Thank you and have fun!
Mick

_Edith says:_ exchanged the truth table with an actual table 😄


davidkna (changes requested):

mickimnet:
As, I prefer to have it documented here, I made a new truth table for [David's review comment](https://github.com/starship/starship/pull/5196#discussion_r1202716043):

#### Truth table for combining `ssh_only` with `detect_env_vars`

_Legend:_

* `detect_env_vars`:
  * `positive`: a name of an environment variable is configured and the variable is set
  * `empty`: the array is empty
  * `none:` none of the entered environment variables are set
  * `negative`: the name of an environment variable is prefixed with an "!" and the variable is set
  * **bold** shows changes to the truth table in the first post

| ssh_only | SSH_CONNECTION | detect_env_vars | result                   |
| :------: | :------------: | :-------------- | :----------------------- |
|   true   |      true      | empty           | show hostname (like now) |
|   true   |      true      | positive        | show hostname            |
|   true   |      true      | none            | ❓ hostname              |
|   true   |      true      | negative        | **hide** hostname        |
|          |                |                 |                          |
|   true   |     false      | empty           | hide hostname (like now) |
|   true   |     false      | positive        | **show** hostname        |
|   true   |     false      | none            | ❓ hostname              |
|   true   |     false      | negative        | hide hostname            |
|          |                |                 |                          |
|  false   |      true      | empty           | show hostname (like now) |
|  false   |      true      | positive        | show hostname            |
|  false   |      true      | none            | hide hostname            |
|  false   |      true      | negative        | hide hostname            |
|          |                |                 |                          |
|  false   |     false      | empty           | show hostname (like now) |
|  false   |     false      | positive        | show hostname            |
|  false   |     false      | none            | hide hostname            |
|  false   |     false      | negative        | hide hostname            |

I'm unsure how to handle the two conditions, marked with ❓, in the above table.

docs/config/README.md
```diff
@@ -2189,16 +2189,22 @@ format = 'via [⎈ $version](bold white) '
 
 The `hostname` module shows the system hostname.
 
+::: warning
+
+The `detect_env_vars` is only recognized, if `ssh_only` is set to `false`.
+You can still check for a ssh connection, add `SSH_CONNECTION` to `detect_env_vars`.
```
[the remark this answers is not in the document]
mickimnet: There is one combination condition, I'm unsure how to handle: what happens if an environment variable is configured for `detect_env_vars` and it is not set in the shell environment?

With `ssh_only = "false"`, my logic would say: hide the hostname. With `ssh_only = "true"` and an active `ssh` connection, what should take precedence: the `ssh_only` or the `detect_env_vars` option?

See [my comment](https://github.com/starship/starship/pull/5196#issuecomment-1561171181) for a new truth table.

davidkna:
I think it makes the most sense to consider both options mutually independent boolean variables.
```math
\begin{align*}
s &= \lnot \text{ssh\_only} \lor \text{SSH\_CONNECTION} \in \text{ENV} \\
w &= \begin{cases}
  t & \text{if empty} \\
  f & \text{there is any negative match} \\
  t & \text{there are only negatives matchers} \\
  w & \text{there is a positive match} \\
  f & \text{otherwise}
\end{cases}
\end{align*}
```

In that case, I think it makes the most sense to enable the module if $`s \land w`$

davidkna:
The only part ("matchers") is concerned with the matchers themselves, so having only `!`-variables in the config.

The other one ("matches") is concerned with the evaluated matcher-matches.

src/context.rs
```diff
@@ -235,8 +235,20 @@ impl<'a> Context<'a> {
...
+            .any(|env_var| env_var.starts_with('!') && self.get_env(&env_var[1..]).is_some())
+    }
+
+    /// Returns true if 'detect_env_vars' is empty,
+    /// or if at least one environment variable is set and no negated environment variable is set
     pub fn detect_env_vars(&'a self, env_vars: &'a [&'a str]) -> bool {
-        env_vars.is_empty() || (env_vars.iter().any(|e| self.get_env(e).is_some()))
+        env_vars.is_empty()
+            || ((env_vars.iter().any(|e| self.get_env(e).is_some()))
+                && self.has_no_negative_env_var(env_vars))
```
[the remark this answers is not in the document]
mickimnet: Because of the supposed formula in https://github.com/starship/starship/pull/5196#issuecomment-1566228913 we decided to leave it in.

mickimnet:
Additionally, to the test on _macOS_, I did test the new feature on _NixOS_ as well.

src/context.rs
```diff
@@ -235,8 +235,31 @@ impl<'a> Context<'a> {
...
+        if self.has_negated_env_var(env_vars) {
+            return false;
+        }
+
+        // Returns true if all environment variables start with a "!",
+        // or if at least one environment variable is set
+        env_vars.iter().all(|env_var| env_var.starts_with('!'))
+            || env_vars
+                .iter()
+                .any(|env_var| self.get_env(env_var).is_some())
```
davidkna, line 262: What do you think about this?

```suggestion
        let mut iter = env_vars
            .iter()
            .filter(|env_var| !env_var.starts_with('!'))
            .peekable();

        iter.peek().is_none() || iter.any(|env_var| self.get_env(env_var).is_some())
```

src/context.rs
```diff
@@ -235,8 +235,31 @@ impl<'a> Context<'a> {
         disabled == Some(true)
     }
 
+    /// Returns true when a negated environment variable is defined in `env_vars` and is present
+    fn has_negated_env_var(&self, env_vars: &'a [&'a str]) -> bool {
+        env_vars
+            .iter()
+            .any(|env_var| env_var.starts_with('!') && self.get_env(&env_var[1..]).is_some())
```
davidkna, line 242: `strip_prefix` can be used here to combine the `starts_with` and `[1..]` part:
```suggestion
        env_vars
            .iter()
            .filter_map(|env_var| env_var.strip_prefix('!'))
            .any(|env_var| self.get_env(env_var).is_some())
```

mickimnet: Thank you! Very good recommendation. A lot better to read and probably faster as well.

src/context.rs
mickimnet: Faster and more readable as well. I just need to rewrite the comments.

docs/config/README.md
```diff
@@ -2200,16 +2200,22 @@ format = 'via [⎈ $version](bold white) '
 
 The `hostname` module shows the system hostname.
 
+::: warning
+
+The `detect_env_vars` is only used, if `ssh_only` is set to `false` or there
+is an environment variable `SSH_CONNECTION` set.
+
```
davidkna, line 2207: I think this warning can be removed
```suggestion
```

mickimnet: True! Very good point.

davidkna:
Thanks for the contribution @mickimnet!

mickimnet:
Thank you so much, for your support @davidkna

PR #5211, opened 2023-05-29
fix(package): support non-virtual cargo workspaces

tobyvin:
#### Description
Fix the logic for finding cargo package's version for workspaces that contain a root project.

If `package.version.workspace = true`, also check current directory's for `workspace.package.version` (non-virtual workspace) before moving up the ancestor tree.

#### Motivation and Context
Cargo workspace version was not detected when in a workspace with a root project.

I extended and ran the existing cargo workspace unit tests.

#### Checklist:

### DIFF
--- a/src/modules/package.rs
+++ b/src/modules/package.rs
@@ -7,8 +7,6 @@ use quick_xml::events::Event as QXEvent;
 use quick_xml::Reader as QXReader;
 use regex::Regex;
 use serde_json as json;
-use std::fs;
-use std::io::Read;
 
 /// Creates a module with the current package version
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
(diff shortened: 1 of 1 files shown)


src/modules/package.rs
```diff
@@ -223,44 +221,35 @@ fn get_sbt_version(context: &Context, config: &PackageConfig) -> Option<String>
...
             }
         }
-        version?
     } else {
-        // This might be a workspace file
-        cargo_toml
-            .get("workspace")?
-            .get("package")?
-            .get("version")?
-            .as_str()?
```
davidkna, line 263: I think this case still needs to be handled.

src/modules/package.rs
```diff
@@ -223,44 +221,35 @@ fn get_sbt_version(context: &Context, config: &PackageConfig) -> Option<String>
...
+                .get("workspace")
+                .and_then(|w| w.get("package"))
+                .and_then(|p| p.get("version"))
+                .and_then(|v| v.as_str())
+            {
+                Some(v) => break v,
+                None => {
+                    let cargo_path = ancestors.next()?;
+                    cargo_toml = std::fs::read_to_string(cargo_path).ok()?.parse().ok()?;
+                }
```
davidkna, line 248: I prefer the way this was handled previously, please just handle the initial case manually instead.

davidkna (changes requested):

PR #5217, opened 2023-06-05
Add godot module

davidscholberg:
The godot module tells whether or not the current directory is within a [Godot](https://godotengine.org/) project folder and optionally shows the version of a godot installation on the user's system.

#### Description

In order to check if the current directory is within a godot project folder, this module looks for the presence of a `project.godot` file in the current directory or one of its ancestors.

This module can also optionally show the version of a godot installation on the user's system, but this feature is disabled by default because it's not uncommon for godot to not be in the user's path, and the executable is not guaranteed to have a predictable name. The user can configure a custom command to get the version from.

#### Motivation and Context
Godot users may find this feature useful.

image

I've tested this with the default configuration, as well as with a custom `version_command`. I've tested both inside and outside of a godot project folder.

#### Checklist:

### DIFF
--- a/src/configs/godot.rs
+++ b/src/configs/godot.rs
@@ -1,0 +1,32 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Deserialize, Serialize)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+#[serde(default)]
+pub struct GodotConfig<'a> {
+    pub format: &'a str,
+    pub version_format: &'a str,
+    pub style: &'a str,
+    pub symbol: &'a str,
+    pub disabled: bool,
+    pub show_version: bool,
+    pub version_command: Vec<&'a str>,
+}
+
+impl<'a> Default for GodotConfig<'a> {
+    fn default() -> Self {
+        GodotConfig {
+            format: "via [$symbol($version )]($style)",
+            version_format: "v${raw}",
+            style: "bold blue",
+            symbol: "godot ",
+            disabled: false,
+            show_version: false,
+            version_command: vec!["godot", "--version"],
+        }
+    }
+}
--- a/src/modules/godot.rs
+++ b/src/modules/godot.rs
@@ -1,0 +1,192 @@
+use crate::configs::godot::GodotConfig;
+
+use super::{Context, Module};
+
+/// Creates a module that tells if the current directory is in a Godot project.
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    use super::ModuleConfig;
+    use crate::formatter::StringFormatter;
+    use crate::formatter::VersionFormatter;
+
+    let godot_project_root = context
+        .begin_ancestor_scan()
+        .set_files(&["project.godot"])
+        .scan();
+    godot_project_root?;
+
+    let mut module = context.new_module("godot");
+    let config: GodotConfig = GodotConfig::try_load(module.config);
+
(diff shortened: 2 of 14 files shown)


src/modules/godot.rs
```diff
@@ -0,0 +1,169 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    use super::ModuleConfig;
+    use crate::formatter::StringFormatter;
+    use crate::formatter::VersionFormatter;
+
+    let godot_project_root = context
+        .begin_ancestor_scan()
+        .set_files(&["project.godot"])
+        .scan();
+    godot_project_root?;
```
davidkna, line 15: So far, we have only really used this for version control systems, I would prefer to keep this with the regular scanner and the related options.

src/configs/godot.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl<'a> Default for GodotConfig<'a> {
+    fn default() -> Self {
+        GodotConfig {
+            format: "via [$symbol($version )]($style)",
+            version_format: "v${raw}",
+            style: "bold blue",
+            symbol: "godot ",
+            disabled: false,
+            show_version: false,
```
davidkna, line 28: There is no need for an explicit `show_version` because users can just remove `$version` from the format string instead. If the command is too slow, I would prefer disabling the module by default instead.

src/modules/godot.rs
```diff
@@ -0,0 +1,169 @@
...
+    for (index, char) in version_output.char_indices() {
+        if char == '.' {
+            period_count += 1;
+            if period_count == 3 {
+                let (version_string, _) = version_output.split_at(index);
+                return Some(String::from(version_string));
+            }
+        }
+    }
+    None
```
davidkna, line 74: I would prefer a regex or `split` & `take_while` here. Instead of only counting dots, please also make sure the text in-between is only compromised of digits.

davidkna (changes requested):

src/configs/godot.rs
davidscholberg: The reason for not showing the version by default is not because of slowness but because the godot executable is not guaranteed to be in the user's path or have a consistent name.

Would it be acceptable to just remove `$version` from the default format string?

src/modules/godot.rs
davidscholberg: Is the ancestor scan reserved for VCS due to performance reasons? If so, would it be better to just have this module disabled by default?

I'd prefer to keep the ancestor scan for a couple reasons:

1. One of the common godot file extensions, `.gd`, is apparently shared by another project called [GD Graphics Library](https://libgd.github.io/), so searching for this extension could lead to false positives.
2. Personally, I like to know that I'm within a godot project folder even if I'm currently in an organizational subdirectory that doesn't contain any godot-specific file formats.

src/modules/godot.rs
davidscholberg: I pushed an update that uses `split`, `take`, and `take_while` here, and I also updated the unit tests to more thoroughly cover this function.

src/configs/godot.rs
davidkna: For other modules with slow CLIs, we have disabled them by default, so I would prefer doing that for consistency.

src/modules/godot.rs
davidkna: If the module is disabled by default, adding `gd` should be less of an issue.

davidscholberg: If the module was disabled by default, wouldn't the ancestor scan also be less of an issue?

PR #5253, opened 2023-06-15
fix(bash): Clear out completed jobs before counting NUM_JOBS

ajeetdsouza:
#### Description

Fixes https://github.com/starship/starship/issues/5159

#### Motivation and Context

Due to a bug in certain Bash versions, any external process launched inside $PROMPT_COMMAND will be reported by `jobs` as a background job:

```bash
   [1]  42135 Done                    /bin/echo
```

This is a workaround - we run `jobs` once to clear out any completed jobs first, and then we run it again and count the number of jobs.

More context: https://github.com/starship/starship/issues/5159
Original bug: https://lists.gnu.org/archive/html/bug-bash/2022-07/msg00117.html

#### Checklist:

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -37,6 +37,18 @@ starship_precmd() {
         STARSHIP_PIPE_STATUS=(${BP_PIPESTATUS[@]})
     fi
 
+    # Due to a bug in certain Bash versions, any external process launched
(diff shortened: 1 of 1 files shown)


chipbuster (approved):
:+1: LGTM

davidkna:
Thanks for the fix @ajeetdsouza, and thanks for handling the review @chipbuster!

PR #5260, opened 2023-06-19
feat(separator): support adding separator line with full context width

yjpark:
#### Description
Support adding separator line with full context width.

#### Motivation and Context
I'd like to have a clear visual way to separate each command and result in terminal, as shown in the screenshot, the config for the screen shot is:

```
add_newline = false
add_separator = true
```

You can also define `separator_pattern` in the config (type is String, will be repeated to fill the whole context width), note that the way to calculate the display size is assuming each character take only single char space (so wider unicode characters such as Chinese characters will have wrong size), since I don't know how to calculate unicode display size without using extra crates, and I don't think this is really needed for normal usages, so it's not a good idea to add an extra dependency.

image

#### Checklist:

Haven't updated the documentation yet, not sure this is a useful feature to other people, or my implementation is in an acceptable way.

### DIFF
--- a/src/configs/starship_root.rs
+++ b/src/configs/starship_root.rs
@@ -18,6 +18,8 @@ pub struct StarshipRootConfig {
     pub scan_timeout: u64,
     pub command_timeout: u64,
     pub add_newline: bool,
+    pub add_separator: bool,
+    pub separator_pattern: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub palette: Option<String>,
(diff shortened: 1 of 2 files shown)


davidkna:
Can't this already be achieved with the `fill` module?

yjpark:
Thanks for the comment, that's right, not notice the `fill` module, which is perfect for my needs, here is my config using it for the same result.

```
add_newline = false
format = "$fill\n$all"

[fill]
symbol = "═"
style = "gray"
```

PR #5282, opened 2023-06-28
feat: support xdg-conformant config file location

### LINKED ISSUE #896: Use the correct configuration directory for your platform
#### Current Behavior
Currently, starship always places its configuration in `$HOME/.config/starship.toml`

#### Expected Behavior
I'd expect starship to follow the platform conventions for where to store application specific configuration. These are generally to identify the user-specific configuration directory, create an application specific subdirectory, and store configuration in there.

Specifically, I'd expect:

- Linux: `$XDG_CONFIG_HOME/starship/config.toml` (defaults to `$HOME/.config/starship/config.toml`)
- macOS: `$HOME/Library/Application Support/starship/config.toml`
- Windows: `{FOLDERID_RoamingAppData}/starship/config.toml`

#### Possible Solution

I'd recommend using the [`directories`](https://docs.rs/crate/directories/2.0.2) library, which handles all these conventions automatically. This library is by the same author as `dirs`, which is what it currently used.kan

#### Environment
- Starship version: 0.33.1
- powershell version: 10.0.18
- Terminal emulator: Windows Terminal Preview, 0.8.10261.0

cpea2506:
#### Description

Users now can have an ability to put their default config in `starship/config.toml`. If users have both `starship.toml` and `starship/config.toml`, the former should take higher priority even if it's invalid.
 
#### Motivation and Context

I've used this repo for a long time and wait for the feature to put the configuration file in a folder like `.config/starship/config.toml` by default so that my `.config` folder could be more tidy and manageable (especially when working with `dotfiles`).

Closes #896

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -8,13 +8,13 @@ To get started configuring starship, create one of the following file:
 mkdir -p ~/.config && touch ~/.config/starship.toml
 ```
 
-- `~/.config/starship/config.toml`
+- `~/.config/starship/config.toml` or `$XDG_CONFIG_HOME/starship/config.toml` if `$XDG_CONFIG_HOME` is set.
 
 ```sh
 mkdir -p ~/.config/starship && touch ~/.config/starship/config.toml
 ```
 
-Note: if you have both files available, the `starship.toml` will take higher priority.
+Note: if you have both files available, the `~/.config/starship.toml` will take higher priority.
 
 All configuration for starship is done in this [TOML](https://github.com/toml-lang/toml) file:
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -1,10 +1,20 @@
 # Configuration
 
-To get started configuring starship, create the following file: `~/.config/starship.toml`.
+To get started configuring starship, create one of the following file:
+
+- `~/.config/starship.toml`.
 
 ```sh
 mkdir -p ~/.config && touch ~/.config/starship.toml
 ```
+
+- `~/.config/starship/config.toml`
+
+```sh
+mkdir -p ~/.config/starship && touch ~/.config/starship/config.toml
+```
+
+NOTE: if you have both files available, the `starship.toml` will take higher priority.
 
 All configuration for starship is done in this [TOML](https://github.com/toml-lang/toml) file:
 
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -14,7 +14,7 @@ mkdir -p ~/.config && touch ~/.config/starship.toml
 mkdir -p ~/.config/starship && touch ~/.config/starship/config.toml
 ```
 
-NOTE: if you have both files available, the `starship.toml` will take higher priority.
+Note: if you have both files available, the `starship.toml` will take higher priority.
 
 All configuration for starship is done in this [TOML](https://github.com/toml-lang/toml) file:
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -431,7 +431,16 @@ fn get_config_path_os(env: &Env) -> Option<OsString> {
     if let Some(config_path) = env.get_env_os("STARSHIP_CONFIG") {
         return Some(config_path);
     }
-    Some(home_dir(env)?.join(".config").join("starship.toml").into())
+
+    let config_dir = home_dir(env)?.join(".config");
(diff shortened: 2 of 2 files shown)


src/context.rs
```diff
@@ -431,7 +431,16 @@ fn get_config_path_os(env: &Env) -> Option<OsString> {
     if let Some(config_path) = env.get_env_os("STARSHIP_CONFIG") {
         return Some(config_path);
     }
-    Some(home_dir(env)?.join(".config").join("starship.toml").into())
+
+    let config_dir = home_dir(env)?.join(".config");
+    let config_file_path = config_dir.join("starship.toml");
+    let folder_based_config_file_path = config_dir.join("starship/config.toml");
```
davidkna, line 437: For this, I would prefer following the XDG-spec (and things like `$XDG_CONFIG_HOME`), or use the platform-native directory via `dirs`. Please consult the discussions in #896.

davidkna (changes requested):

src/context.rs
cpea2506: I've updated to follow the XDG-spec. It should respect the `$XDG_CONFIG_HOME` and fallback to `$HOME/.config` if the env is not set. 
I'm not sure if it could break current user config in case they have `$XDG_CONFIG_HOME` in different place but `starship.toml` still in `~/.config/`.

davidkna: I think only `~/.config/starship/config.toml` should use `$XDG_CONFIG_HOME` and `~/.config/starship.toml` should only use `$HOME` to keep compatibility.

docs/config/README.md
```diff
@@ -1,11 +1,21 @@
...
 mkdir -p ~/.config && touch ~/.config/starship.toml
 ```
 
+- `~/.config/starship/config.toml`
+
+```sh
+mkdir -p ~/.config/starship && touch ~/.config/starship/config.toml
+```
+
+Note: if you have both files available, the `starship.toml` will take higher priority.
```
davidkna, line 17:
```suggestion
Note: if you have both files available, the `~/.config/starship.toml` will take higher priority.
```

davidkna (commented):
LGTM apart from the docs, but I would prefer an additional review from another team member before merging, if this is a step we want to take.

docs/config/README.md
```diff
@@ -1,11 +1,21 @@
...
 mkdir -p ~/.config && touch ~/.config/starship.toml
 ```
 
+- `~/.config/starship/config.toml` or `$XDG_CONFIG_HOME/starship/config.toml` if `$XDG_CONFIG_HOME` is set.
+
+```sh
+mkdir -p ~/.config/starship && touch ~/.config/starship/config.toml
+```
+
+Note: if you have both files available, the `~/.config/starship.toml` will take higher priority.
```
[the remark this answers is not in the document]
davidkna: `~/.config/starship.toml` has higher priority.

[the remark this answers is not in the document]
mickimnet: May bad. Sorry, misread it 😊

LionC:
Landed here from the linked discussion - what work is still missing here to get it merged? Would be happy to do whatever needed just to get my own config dir clean ;-)

pantheraleo-7:
After reading the opinions on the discussion, here's what I would recommend:
- @\matchai wants a single path [in the docs], to keep things simple for non-technical users, let this path be `~/.config/starship/config.toml`. So let's just update the path, no other doc changes there (optionally add a `Note:` alerting users to see the below section for advance config file path configuration).
- Deprecate the old `~/.config/starship.toml` (but keep it to maintain backwards compatibility )
- Support `$XDG_CONFIG_HOME`, and give it the highest priority (if set).
- Mention all the "technicalities" i.e. the deprecation, the `$XDG_CONFIG_HOME` support, the priority, and the `STARSHIP_CONFIG` env var in a separate section (currently we have `Config File Location` section in the docs, we can update this)

The order of precedence [for the 3 paths]:
- `$XDG_CONFIG_HOME`/starship/config.toml (if set)
- `~/.config/starship/config.toml` (the "default" path)
- `~/.config/starship.toml` (will only be effective if the other 2 aren't available )

pantheraleo-7:
[#6690](https://github.com/starship/starship/pull/6690)

PR #5303, opened 2023-07-11
build(refactor) - Refactor platform detection for builds

### LINKED ISSUE #5022: "x86_64 builds for linux are not yet available for Starship" (possibly due to Debian?)
This strikes me as extremely odd, as this is THE most common open-source OS lol.

So I'm guessing it has more to do with the fact that it's...
```
bash>> uname -a
Linux truenas 5.15.79+truenas #1 SMP Tue Dec 13 12:40:04 UTC 2022 x86_64 GNU/Linux
```
(which is a Debian)

I just wanted the nice PS1 from my main machine also available on my TrueNAS box. >..<

DynamicApproach:
Attempt to refactor platform detection for builds in order to better detect WSL/Distributions that return "linux"

#### Description
Some WSL platforms aren't currently detected and detection causes issues with some x86_64 platforms. This PR adds support for WSL detection with current detection as a fallback. Uses OS-release files to try to find OS name.

#### Motivation and Context
This should fix 5022 as well as improve detection for platforms which return "linux" as the platform.
Closes #5022 (I believe?)

https://github.com/starship/starship/issues/5022

Ubuntu WSL:
image

Only in WSL currently. 

#### Checklist:

- [ x ] I have updated the documentation accordingly.
- [ x ] I have updated the tests accordingly.

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -18,7 +18,7 @@ SUPPORTED_TARGETS="x86_64-unknown-linux-gnu x86_64-unknown-linux-musl \
                   arm-unknown-linux-musleabihf x86_64-apple-darwin \
                   aarch64-apple-darwin x86_64-pc-windows-msvc \
                   i686-pc-windows-msvc aarch64-pc-windows-msvc \
-                  x86_64-unknown-freebsd"
+                  x86_64-unknown-freebsd x86_64-debian x86_64-ubuntu"
 
 info() {
   printf '%s\n' "${BOLD}${GREY}>${NO_COLOR} $*"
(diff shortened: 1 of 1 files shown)


davidkna:
Isn't this still just handling the `linux`value? MUSL was preferred to avoid versioning glibc and other dependency issues, I'm uncertain if the binary will run on all platforms that you are redirecting to GNU targets. I would expect this to require something similar to #2612.
Furthermore, have you verified if the FreeBSD case you added is sensible and works in Linux Binary Compatibility environments on FreeBSD?

DynamicApproach:
> Isn't this still just handling the `linux`value? 
Yes, it was just a quick fix by using release files. I wasn't aware of that PR (last activity seems to be two years old?) Is it still being actively worked on?

> MUSL was preferred to avoid versioning glibc and other dependency issues, I'm uncertain if the binary will run on all platforms that you are redirecting to GNU targets.

I highly doubt it would - I didn't do any sort of through testing, only checking that it worked with WSL (debian and ubuntu only so a very limited test) 

 > I would expect this to require something similar to #2612. 
 Very much agreed, it's essentially a simple version of it. 
 
> Furthermore, have you verified if the FreeBSD case you added is sensible and works in Linux Binary Compatibility environments on FreeBSD?
Nope! Sorry, once again, I did minimal testing on this. Is there any specific issue that needs work on?

davidkna:
I am not aware of any activity on the other PR.

If we were to distribute the glibc version to some users, I think it should have a level of checking similar to #2612 (though personally, I would not want to review that kind of thing at this time). And regarding the FreeBSD check, it should be verified if the binary is executable in the given conditions.

DynamicApproach:
> I am not aware of any activity on the other PR.

Is there anything left to be done on it? 

> If we were to distribute the glibc version to some users, I think it should have a level of checking similar to #2612 (though personally, I would not want to review that kind of thing at this time). 
And regarding the FreeBSD check, it should be verified if the binary is executable in the given conditions.

What would be best moving forward?  Is it worth checking that, or should I just close the PR? 🙂

 My main aim with the PR was just to be a quick fix for WSL, a few of the WSL distros work fine it seemed. I tested fedora but it failed pretty gracefully.

davidkna:
> Is there anything left to be done on it?

I think it wasn't quite done yet.

> My main aim with the PR was just to be a quick fix for WSL

I'm not sure how this PR would help WSL, can you elaborate what WSL-issue this PR fixes exactly? A more targeted fix for WSL in particular might be a better approach in case there is a problem.

davidkna:
@DynamicApproach Are you planning to continue work on this or can I close this PR?

DynamicApproach:
> Are you planning to continue work on this or can I close this PR?

I'll check tonight to see if it or something else is needed and I'll update accordingly; sorry for the delay, I've been busy

PR #5313, opened 2023-07-14
refactor(directory): avoid unnecessary string copies

cgzones:
#### Description
Avoid some unnecessary string copies by using references.

#### Checklist:
- [/] I have updated the documentation accordingly.

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -35,7 +35,6 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("directory");
     let config: DirectoryConfig = DirectoryConfig::try_load(module.config);
 
-    let home_symbol = String::from(config.home_symbol);
     let home_dir = context
         .get_home()
         .expect("Unable to determine HOME_DIR for user");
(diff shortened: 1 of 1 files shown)


src/modules/directory.rs
```diff
@@ -149,12 +147,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
-                "path" => Some(Ok(&path_vec[2])),
-                "before_root_path" => Some(Ok(&path_vec[0])),
-                "repo_root" => Some(Ok(&path_vec[1])),
+                "path" => Some(Ok(path_vec[2].as_str())),
+                "before_root_path" => Some(Ok(path_vec[0].as_str())),
+                "repo_root" => Some(Ok(path_vec[1].as_str())),
                 "read_only" => {
                     if is_readonly_dir(physical_dir) {
-                        Some(Ok(&lock_symbol))
+                        Some(Ok(config.read_only))
```
davidkna, line 155: Shouldn't `as_str` usually match `&`?

davidkna (changes requested):

src/modules/directory.rs
cgzones: No, since `&` converts into `&String` (not `&str`), which `config.read_only` would not fit.

davidkna (approved):

PR #5314, opened 2023-07-14
feat(logger): add option to disable logging to file

cgzones:
#### Description
Add the option to disable logging to file completely.

#### Motivation and Context
If starship does not have permissions to create the log directory it will complain.

#### Checklist:

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -19,6 +19,10 @@ pub struct StarshipLogger {
     log_level: Level,
 }
 
+struct StarshipStderrLogger {
+    log_level: Level,
+}
+
 /// Returns the path to the log directory.
 pub fn get_log_dir() -> PathBuf {
     env::var_os("STARSHIP_CACHE")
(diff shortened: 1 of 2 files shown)


src/logger.rs
```diff
@@ -236,8 +245,51 @@ impl log::Log for StarshipLogger {
...
+    }
+
+    fn flush(&self) {
+        std::io::stderr().flush().unwrap();
+    }
+}
+
 pub fn init() {
-    log::set_boxed_logger(Box::<StarshipLogger>::default()).unwrap();
+    if env::var("STARSHIP_DISABLE_LOG_FILE").is_ok() {
```
davidkna, line 288:
```suggestion
    if env::var_os("STARSHIP_DISABLE_LOG_FILE").is_some() {
```

davidkna (changes requested):
Technically, this is already possible by changing the log dir to `/dev/null` or `nul`. 

Now for the main feedback: Please try to reduce duplication with the regular logger.

PR #5317, opened 2023-07-16
feat: allow battery styling to differ based on whether the battery is charging

doy:
#### Description
adds configuration options `charging_style` and `discharging_style` to the battery display configuration.

#### Motivation and Context
it's distracting to have the battery display in bright red if i've already addressed the issue by plugging my laptop in.

i wrote automated tests and also ran it locally with `charging_style` and `discharging_style` configuration options set and observed the desired behavior.

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -45,7 +45,9 @@
         "discharging_symbol": "󰂃 ",
         "display": [
           {
+            "charging_style": null,
             "charging_symbol": null,
+            "discharging_style": null,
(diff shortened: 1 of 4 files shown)


davidkna:
I prefer the approach discussed in #3849, but the original plan is blocked by #3941. I saw some other PRs avoid the need for #3941 by implementing `Deserialize` manually, so given the lack of progress on #3941 that might be the best way forward.

PR #5320, opened 2023-07-16
feat(git_metrics): add option to include staged files

f10et:
Add an option to include staged files for the git_metrics module

#### Description

Make git metrics module using `git diff HEAD` (+ other options) when the config is set to include staged files. That way, it's possible to see the metrics for all files not committed.

#### Motivation and Context

I often staged changes during my work to easily reset errors, but I would still like to know how many lines would be added to the commit, when I decide to commit.

#### Checklist:


davidkna (commented):
The changes themselves look good, but I would prefer handling the option with an enum/string option to make other diff styles easier to implement. Don't hesitate to reach out if you need help with that, currently automatic  `serde` `enum` de-serialization doesn't work.

f10et:
Hi @davidkna,

Sorry, I didn't notice your feedback until now. Would you have an example of another module using what you propose? I have to say this was my first time touching Rust, so I definitely expect to have a better way to implement this.

PR #5325, opened 2023-07-18
feat(scanner): add option not to follow symlinks

lutzky:
Never consider a symlink to be a subdirectory; this prevents hanging when the symlink points to a slow/inaccessible filesystem.

#### Description

Instead of using `is_dir`, which follows symlinks, use [`symlink_metadata`](https://doc.rust-lang.org/std/fs/fn.symlink_metadata.html), which doesn't.

#### Motivation and Context

If the current directory has a symlink to a slow filesystem (e.g. an inaccessible network filesystem), `is_dir` hangs indefinitely, disregarding `scan_timeout`.

This should help with some instances of #312.

n/a

I've tested by running `starship timings` with network disconnected in a directory which includes a symlink to a network filesystem. Without my change it hangs, with my change it doesn't.

#### Checklist:

- [x] I have updated the documentation accordingly
- [x] I have updated the tests accordingly

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1794,6 +1794,10 @@
       "default": true,
       "type": "boolean"
     },
+    "follow_symlinks": {
+      "default": true,
+      "type": "boolean"
+    },
     "palette": {
       "type": [
         "string",
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -219,6 +219,14 @@ This is the list of prompt-wide configuration options.
 | `add_newline`     | `true`                         | Inserts blank line between shell prompts.                                                                                                                                        |
 | `palette`         | `''`                           | Sets which color palette from `palettes` to use.                                                                                                                                 |
 | `palettes`        | `{}`                           | Collection of color palettes that assign [colors](/advanced-config/#style-strings) to user-defined names. Note that color palettes cannot reference their own color definitions. |
+| `follow_symlinks` | `true`                         | Follows symlinks to check if they're directories; used in modules such as git.                                                                                                   |
+
+::: tip
+
+If you have symlinks to networked filesystems, consider setting
+`follow_symlinks` to `false`.
+
+:::
 
 ### Example
 
--- a/src/configs/starship_root.rs
+++ b/src/configs/starship_root.rs
@@ -18,6 +18,7 @@ pub struct StarshipRootConfig {
     pub scan_timeout: u64,
     pub command_timeout: u64,
     pub add_newline: bool,
+    pub follow_symlinks: bool,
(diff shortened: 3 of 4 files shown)


davidkna:
Following symlinks does seem somewhat useful, I assume there is no option to ignore cross-device symlinks only?

lutzky:
The trick is determining whether or not a symlink is cross-device *without* calling `stat` on the destination (because that's the bit that might fail). https://github.com/stemjail/mnt-rs seems to be able to do it, but doesn't appear very portable. WDYT?

davidkna:
> The trick is determining whether or not a symlink is cross-device _without_ calling `stat` on the destination (because that's the bit that might fail). https://github.com/stemjail/mnt-rs seems to be able to do it, but doesn't appear very portable. WDYT?

Try using `systemstat` which we already depend on, relevant functions will likely return an error on unsupported operating systems which should be handled.

lutzky:
I've tried out some stuff in https://github.com/starship/starship/pull/5325/commits/a99335a127a4b2369dffa60b829b4fe05fccbc14; WDYT? Reasonable or too brittle?

lutzky:
There's an issue with `systemstat`: [`mounts`](https://docs.rs/systemstat/latest/x86_64-unknown-linux-gnu/src/systemstat/platform/linux.rs.html#641-654) proceeds to immediately `stat` each filesystem, causing it to wait on network. :disappointed: 

I guess one path forward is submitting a patch to `systemstat` to make a no-stat-just-the-files option... WDYT?

davidkna:
> I guess one path forward is submitting a patch to `systemstat` to make a no-stat-just-the-files option... WDYT?

The crate is somewhat minimalistic, so I'm uncertain if the maintainers will be receptive to such a change.

lutzky:
So, given these limitations, how about we go back to the "don't follow symlinks in this case" option? Perhaps as a configurable?

davidkna:
I think making it configurable is fine, unless you want to try making the timeout itself more reliable.

lutzky:
Making the timeout itself more reliable is more challenging (`D` is called "uninterruptable" for a reason, you'd end up accumulating stale threads, it might be worth doing but would be messy). For now, I've made it configurable. PTAL

lutzky:
Finally added tests. I believe that the failures on nightly are unrelated to my changes. Do my changes seem OK now?

davidkna (approved):
LGTM

lutzky:
Still blocked on https://github.com/starship/starship/pull/5467 for passing the checks.

davidkna:
Thanks for the contribution @lutzky!

PR #5327, opened 2023-07-20
Add attribution of subcommand `--print-full-init`

### LINKED ISSUE #5278: Lack of attribution of description for `--print-full-init`
v.: **1.15.0** | Hello. Lack of attribution of description for `--print-full-init` that is printed by `starship init --help`.

Amazingkenneth:
#### Description

Add attribution of subcommand `--print-full-init`

#### Motivation and Context

Closes #5278 

#### Checklist:

There are no documentation or tests that need to be updated for this fix.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -50,6 +50,7 @@ enum Commands {
     ///  Prints the shell function used to execute starship
     Init {
         shell: String,
+        /// Prints out the main initialization script
         #[clap(long)]
         print_full_init: bool,
     },


src/main.rs
```diff
@@ -50,6 +50,7 @@ enum Commands {
     ///  Prints the shell function used to execute starship
     Init {
         shell: String,
+        /// Prints out the main initialization script
```
davidkna, line 53: I feel like this by itself isn't a great explanation of what the options do. The explanation should mention how this is different from not passing the option, it may not be clear to users what the main script is without context. Another option would be to hide this option, but I think I prefer keeping it visible.

Amazingkenneth: What about adding a redirect to [the page which introduces it](https://github.com/starship/starship/pull/168)? I think the option without explanation might make the user relatively more confused.

PR #5335, opened 2023-07-23
feat: add random module

### LINKED ISSUE #5998: Fortune equivalent for starship?
I configured `fortune | lolcat` to run on shell startup, and I think it would be fun if starship had a module that selects and random word and just inserts it into the prompt...so a couple of questions.

1. Am I dumb, and this is already implemented?
2. If I made a PR on this, does it have a chance of being merged?

For example, if this is the format for the prompt: `$randword$character`, then the output would be something like this:
`Pneumonoultramicroscopicsilicovolcanoconiosis $`. Obviously, the word would probably be much shorter. Perhaps it could be used for fun easter eggs or things like that.

0x00002a:
This adds a new module "random"

#### Description

A new module type "random" is added which lets the user specify a list of style and symbol variables, a random one of which will be chosen when the prompt is generated

#### Motivation and Context

I was doing this before with the `custom` module and a `fish` function. This has the disadvantage of needing `fish` installed and it can be quite slow (see below for comparisons)

Closes #5998

fish: Screenshot_20230723_175406

native: Screenshot_20230723_180008

I'm using Arch Linux in fish (although I also tested with bash). I ran the entire test suite and added my own tests

#### Checklist:

## Current issues

~~There's a problem I can't track down the source of where you will get a warning from starship when doing the initial init:~~
```
[WARN] - (starship::config): Error in 'StarshipRoot' at 'random': Unknown key
```
~~The prompt still works fine and I'm guessing I'm just missing a key somewhere but I can't work out where~~
 resolved by f14a35a04ad3f9b49d7161bfa42dda5ed7039f6f


docs/config/README.md
```diff
@@ -4528,3 +4528,33 @@ detect_extensions = ['pst'] # filters *.pst files
...
+## Random
+
+`random` modules randomise their output every time the prompt is displayed.
+
+### Options
+
+| Option        | Default              | Description                          |
+| ------------- | -------------------- | ------------------------------------ |
+| `styles`      | `['>', '$', '%']`    | List of style variables to pick from |
+| `symbols`     | `['bold', 'italic']` | List of symbols to pick from         |
```
mickimnet, line 4541: The option column doesn't match the default and description columns.

docs/config/README.md
```diff
@@ -4528,3 +4528,33 @@ detect_extensions = ['pst'] # filters *.pst files
...
+| `format`      | `'[$symbol]($style)` | The format for the module            |
```
mickimnet, line 4542: There is missing an apostrophe at the end of the default configuration: `'[$symbol]($style)'`

0x00002a:
Thanks for your review and apologies for not responding earlier. I should have time to look at this this evening if not then tomorrow

docs/config/README.md
```diff
@@ -4528,3 +4528,33 @@ detect_extensions = ['pst'] # filters *.pst files
 shell = ['pwsh.exe', '-NoProfile', '-Command']
 use_stdin = false
 ```
+
+## Random
+
+`random` modules randomise their output every time the prompt is displayed.
```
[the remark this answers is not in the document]
0x00002a: Well it's one module but like `custom` you can define multiple versions of it

docs/config/README.md
0x00002a: I'm not sure what you mean?

0x00002a: I'm not sure what you mean?

mickimnet: Are `['>', '$', '%']` styles or symbols?

src/configs/starship_root.rs
```diff
@@ -119,6 +119,7 @@ pub const PROMPT_ORDER: &[&str] = &[
     "os",
     "shell",
     "character",
+    "random",
```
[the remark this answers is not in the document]
mickimnet: Hm … for that, I would like to have the opinion of someone from the core maintainers.

docs/config/README.md
0x00002a: Oh My God :woman_facepalming:  :woman_facepalming:  :woman_facepalming: well spotted I guess xD

mickimnet: You're welcome.

0x00002a:
@chipbuster sorry just would like to bump this as its been a while

PR #5339, opened 2023-07-26
docs(presets): add new preset parrot-os

snaeil:
#### Description
Add new preset: ParrotOS

#### Motivation and Context
ParrotOS comes with a very special shell styling that should be in starship.
The aim is to provide a beautiful preset with high contrast with high compatibilty (no nerd fonts).

parrot-os

#### Checklist:

- [ x] I have updated the documentation accordingly.


docs/.vuepress/public/presets/toml/parrot-os.toml
```diff
@@ -0,0 +1,53 @@
...
+format = '[\[](fg:red)[$symbol]($style)[\]](fg:red)'
+disabled = false
+
+[character]
+success_symbol = '[\$](fg:yellow)'
+error_symbol = '[\$](fg:yellow)'
+vimcmd_symbol = '[\$](fg:yellow)'
+vimcmd_replace_one_symbol = '[\$](fg:yellow)'
+vimcmd_replace_symbol = '[\$](fg:yellow)'
+vimcmd_visual_symbol = '[\$](fg:yellow)'
```
mickimnet, line 18: I'm uncertain if it is a good idea to use the same symbol for all different character values. In my opinion, that could lead to confused users.

I would prefer if someone from the maintainers would voice their opinion.

docs/presets/parrot-os.md
```diff
@@ -0,0 +1,18 @@
+[Return to Presets](./README.md#parrot-os)
+
+# ParrotOS Preset
+
+This preset is inspired by the zsh look that comes with the [ParrotOS](https://www.parrotsec.org) operating system.
+It doesn't need a nerd font to be installed.
```
mickimnet, line 6: This is not true. You're using the standard branch symbol (``)  in the `[git_branch]` module, which may be available in non _Nerd Fonts_, like fonts with _Powerline_ symbols enabled.

Therefore, please either change the symbol of `[git_branch]` or add the requirement for a _Nerd Font_.

mickimnet (changes requested):

docs/.vuepress/public/presets/toml/parrot-os.toml
snaeil: This is a very valid point. It would be great if there was another solution, maybe there is one I don't know about?

mickimnet: Use different characters and colors?

snaeil: What I meant is, it would be great to have something like a wrapper, so that I can specify the surrounding for all variants of character. I wanted do it as much ParrotOS'ish as possible, that's why I chose exactly `$` and these colours.
Am I misunderstanding your comment?

docs/presets/parrot-os.md
snaeil: Thanks I replaced it

docs/.vuepress/public/presets/toml/parrot-os.toml
snaeil: I removed the usage of `character` now since it's not necessary to imitate ParrotOS' shell style.

mickimnet (approved):
LGTM.

PR #5352, opened 2023-07-31
build: Sort toml preset files

bmwiedemann:
#### Description

Sort toml input file list for reproducible build results.

#### Motivation and Context
Without this patch, the resulting `starship` binary varied, depending on random filesystem readdir order.
See https://reproducible-builds.org/ for why this matters.

This patch was done while working on reproducible builds for openSUSE.

- [X] I have tested using **Linux** (build only)

#### Checklist:

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,5 @@
 use std::fs::{self, File};
+use std::io;
 use std::io::Write;
 
 use shadow_rs::SdResult;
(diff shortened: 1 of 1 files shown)


davidkna (approved):

andytom:
Thank you for your contribution @bmwiedemann and thank you as always for reviewing @davidkna.

PR #5381, opened 2023-08-16
feat(module): added iterm2_mark module (continuation of #3018)

andrewgazelka:
Continuation of #3018


andrewgazelka:
@chipbuster I can't for the life of me figure out why I am getting the current compile error when near-identical code 

https://github.com/starship/starship/blob/b5d3d8fcf331cdff6d0e687dcdbac77351731475/src/configs/nodejs.rs#L17-L19

does not get it. Any help?

andrewgazelka:
I tested this locally and it still seems like it has the same problem as https://gitlab.com/gnachman/iterm2/-/issues/10537 (typing does not work properly as shown in the video)

davidkna:
I think a number of other terminal applications supports this now (VS Code, Windows terminal, kitty?), so I think this should be tied less strongly to iterm. Concerning the spacing issues, the sequence may need to be escaped to let the shell know not to use it in width calculations (see [`wrap_seq_for_shell`](https://github.com/starship/starship/blob/f29416094b45ce3419b5f053ab7529a98f854d21/src/utils.rs#L484)). In general the escapes could perhaps use something like `Segment::Control` from #4208.

andrewgazelka:
> I think a number of other terminal applications supports this now (VS Code, Windows terminal, kitty?), so I think this should be tied less strongly to iterm. Concerning the spacing issues, the sequence may need to be escaped to let the shell know not to use it in width calculations (see [`wrap_seq_for_shell`](https://github.com/starship/starship/blob/f29416094b45ce3419b5f053ab7529a98f854d21/src/utils.rs#L484)). In general the escapes could perhaps use something like `Segment::Control` from #4208.

Thanks. I'll look into the code more a little later today.

davidkna:
@andrewgazelka Are you still planning on continuing work on this PR?

andrewgazelka:
> @andrewgazelka Are you still planning on continuing work on this PR?

@davidkna I am not; I was just trying to get the PR to not have merge conflicts, but it doesn't even fix the issue when I tried.

andrewgazelka:
going to let others carry this on

PR #5426, opened 2023-09-08
fix: improve reliability of config-file writing

davidkna:
#### Description

When writing to the config file, instead of writing directly to the file, create a temporary file in the same directory and attempt to move the file to the destination. This should make writing more reliable in cases multiple concurrent writes happen at the same time by ensuring write operation happen in a more atomic manner. In addition, `starship preset -o` now requires a `--force` flag to replace existing files. ~~The creation of the temporary files is not done with `tempfile` since it sets very restrictive permissions (`600`) on files it creates, with no option to disable it.~~ Edit: `tempfile` allows disabling this behavior now, the PR has been changed to use the library.

#### Motivation and Context

Closes #

#### Checklist:


davidkna:
@cubic-dev-ai review

Cargo.toml
```diff
@@ -118,8 +119,7 @@ dunce = "1.0.5"
 winres = "0.1.12"
 
 [dev-dependencies]
-mockall = "=0.14.0"
-tempfile = "=3.27.0"
+mockall = "0.14.0"
```
davidkna:
```suggestion
mockall = "=0.14.0"
```

PR #5429, opened 2023-09-08
feat: $gemset variable for Ruby module

### LINKED ISSUE #5305: Support gem sets in the Ruby module
Ruby developers that use tools like RVM (or RBEnv with a plugin) often use gemsets to isolate gems between different projects. With OMZ themes like powerlevel9k/10k, the built-in RVM plugin can (optionally) display the gemset alongside the Ruby version. 

An option to display the current gemset (preferably only if it isn't `default`) and the respective format variable, so we can customize the display (default format could be similar to the RVM output: `version@gemset`, eg. `3.2.2@my-project`).

None, other than switching back to pl10k.

marcybelardo:
#### Description

Adding an assignable variable called `$gemset` to the Ruby module. This will read the path provided by the environment variable `GEM_HOME` and capture the gemset information at the end of the path.

#### Motivation and Context

Closes #5305

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3681,11 +3681,12 @@ Starship gets the current Ruby version by running `ruby -v`.
 
 ### Variables
 
-| Variable | Example  | Description                          |
-| -------- | -------- | ------------------------------------ |
-| version  | `v2.5.1` | The version of `ruby`                |
-| symbol   |          | Mirrors the value of option `symbol` |
-| style\*  |          | Mirrors the value of option `style`  |
+| Variable | Example      | Description                                                                |
+| -------- | ------------ | -------------------------------------------------------------------------- |
+| version  | `v2.5.1`     | The version of `ruby`                                                      |
+| symbol   |              | Mirrors the value of option `symbol`                                       |
+| style\*  |              | Mirrors the value of option `style`                                        |
+| gemset   | `2.0.0@test` | Reads the Gemset and its version from the environment variable `GEM_HOME`. |
 
 *: This variable can only be used as a part of a style string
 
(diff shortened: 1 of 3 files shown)


docs/config/README.md
```diff
@@ -3681,11 +3681,12 @@ Starship gets the current Ruby version by running `ruby -v`.
...
-| -------- | -------- | ------------------------------------ |
-| version  | `v2.5.1` | The version of `ruby`                |
-| symbol   |          | Mirrors the value of option `symbol` |
-| style\*  |          | Mirrors the value of option `style`  |
+| Variable | Example      | Description                                                                |
+| -------- | ------------ | -------------------------------------------------------------------------- |
+| version  | `v2.5.1`     | The version of `ruby`                                                      |
+| symbol   |              | Mirrors the value of option `symbol`                                       |
+| style\*  |              | Mirrors the value of option `style`                                        |
+| gemset   | `2.0.0@test` | Reads the Gemset and its version from the environment variable `GEM_HOME`. |
```
rafasoares, line 3689: I think the gemset and version should be separate, so people can format it however they like it, like using an icon instead of the `@` sign.

marcybelardo: Ah, so like `gemset-version` and `gemset-name`?

rafasoares: As far as I know, gemsets themselves don't have versions, they belong to an installed ruby version[^1]. Gemset `2.5.0@my-gemset`, for instance, is part of the ruby 2.5.0 install.

Here's an example:
image

marcybelardo: I see, so it'd be better off to just grab the name on its own, and use the already-fetched `version` variable and allow the users to format it themselves

tchiaspko:
any update on this?

davidkna (approved):

andytom:
Thank you for your contribution @marcybelardo and thank you for reviewing @davidkna.

PR #5431, opened 2023-09-09
build(deps): upgrade to `gix` 0.53

Byron:
This also reduces the binary size from

-rwxr-xr-x  1 byron  staff  6482184 Sep  9 08:52 ./target/release/starship

to

-rwxr-xr-x  1 byron  staff  6399544 Sep  9 08:56 ./target/release/starship


davidkna (approved):

davidkna:
@Byron Thanks for handling the upgrade!

PR #5442, opened 2023-09-14
fix(install): do not use curl installed through snap

### LINKED ISSUE #5403: I Can't Download Starship on my Ubuntu System.
#### Current Behavior

When I try to download starship using the curl command. It shows this error saying it can't
open a tmp file and then it forcely exits the installation process.

#### Expected Behavior

it is meant to run the normal installation process.

image

#### Environment
- Shell type: [fish, zsh]
- Shell plugin manager: [if present, e.g. oh-my-fish, oh-my-zsh, fisher, antigen]
- Terminal emulator: [e.g. iTerm, Hyper, Terminator]
- Operating system: [e.g. macOS 10.13.4, Windows 10]

```
# Your configuration here
```

#### Starship Configuration

```toml
# Your configuration here
```

chipbuster:
#### Description
Changes the install script to check if the resolved `curl` is within `/snap`. If it is, the script prints a warning and acts as if `curl` is not installed, searching for other download programs.

#### Motivation and Context

For some reason, `curl` which comes from the default package on snapcraft is badly broken. Among its deficiencies are:

  1. Attempting to download from a GitHub URL to a location in the home directory results in an empty file being created, even though cURL reports success.
  2. Attempting to download from a GitHub URL to a location in `/tmp` results in *no* file being created at all, even though cURL reports success.
  3. [Attempting to download to a path with a hidden directory results in spurious failures.](https://askubuntu.com/questions/1356327/cant-write-to-a-hidden-path-using-curl)

While the broken version of the curl program is capable of downloading the install script and piping it to `sh`, we have had strange issues when it is used within the installer (#5403). Given the severity and unpredictability of these problems, I think it is safer to disallow the usage of snap-installed curl until all the outstanding issues are fixed.

Closes #5403 

Tested on a fresh install of Ubuntu 23.04 installed with the non-legacy installer. Script works in default setting (since wget is installed by default), correctly bypassing curl if it is snap-installed and using it if it is apt-installed.

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -40,6 +40,14 @@ has() {
   command -v "$1" 1>/dev/null 2>&1
 }
 
+curl_is_snap() {
+  curl_path="$(command -v curl)"
+  case "$curl_path" in
+    (/snap/*) return 0 ;;
+    (*) return 1 ;;
+  esac
+}
+
 # Make sure user is not using zsh or non-POSIX-mode bash, which can cause issues
 verify_shell_is_posix_or_exit() {
   if [ -n "${ZSH_VERSION+x}" ]; then
@@ -55,7 +63,6 @@ verify_shell_is_posix_or_exit() {
   fi
 }
 
-# Gets path to a temporary file, even if
 get_tmpfile() {
   suffix="$1"
   if has mktemp; then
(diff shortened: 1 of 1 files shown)


davidkna (approved):
LGTM

davidkna:
Thanks for the fix @chipbuster!

PR #5443, opened 2023-09-14
ci: Fix how version is obtained for pkgbuild

chipbuster:
#### Description
Modify how the starship version is obtained in CI to be more strict.

#### Motivation and Context

#5137 notes that the version in the distribution XML is broken. This PR changes how the starship version is obtained to be more robust to the errors that occurred.

1. Use Release Please to set the version as an environment variable, and use this environment variable if it exists, filtering out just the numeric portion.
2. If the first step does not work, try to run the starship binary to obtain a version. This line previously had a bug with the ordering of redirect statements which caused an error message to be printed if this step failed. The version here should suppress all errors.
3. If neither of the two steps work, try to grab the first key from `Cargo.toml` using a grep command. This PR fixes two bugs from the previous version: bash generates output on pushd/popd which needs to be suppressed, and we need to choose the first version tag, since there are now multiple in the Cargo.toml.

Tested locally, but unclear if this will work in the actual CI pipeline.

Potentially addresses #5137, but we cannot test this until we actually fire a release.

#### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -176,6 +176,7 @@ jobs:
     env:
       KEYCHAIN_FILENAME: app-signing.keychain-db
       KEYCHAIN_ENTRY: AC_PASSWORD
+      STARSHIP_VERSION: ${{ needs.release_please.outputs.tag_name }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4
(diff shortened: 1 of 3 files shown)


davidkna (approved):
LGTM. Beyond these changes, I noticed that `shellcheck` finds an issue in `build_component_package.sh` line 90 (unquoted variable) we could fix alongside this.

davidkna (approved):

davidkna:
Thanks for the fix @chipbuster!

PR #5448, opened 2023-09-18
feat(status): add success_symbol_rootuser option

eatradish:
#### Description

This commit will allow user to set a different success symbol for root users than for non-root users in some command-line topics (typically `#` for the former and `$` for the latter).

#### Motivation and Context

In some shell theme, the status success symbol is different for root and non-root. For example, bash-config for AOSC OS:

图片

success_symbol in the first and success_symbol_rootuser in the second.

#### Checklist:


MingcongBai:
Any chance that someone would please review this pull request...?

PR #5469, opened 2023-09-30
feat: add user overrides for username and hostname modules

oddlama:
#### Description

This PR adds a very simple but powerful override table, that allows to override some attributes based on the currently logged-in user. This allows you to:

- Color the hostname in the same (or any other) color as the user, so for example red for root
- Only show the username if it isn't the currently logged in user OR the user is root. So you can additionally hide the username if it is conveyed by the colored hostname.
- It allows to change the format based on the user, in case you want to style the "@" in user@hostname based on the hostname's color scheme instead of the username.

#### Motivation and Context

I always missed the possibility to set different colors, styles and formats for the username and hostname component, based on what user is logged in. There's rudimentary support to set a different color for the root user, and a compound option to hide the username under certain conditions, but I'd personally like to be able to make more specific format changes based on the current user. This PR adds this functionality by introducing a new override map and without adding any new "complex logic" that would need to be maintained (username detection is reused from the username module).

- The current design allows this to be extended later for usecases such as #680 or #607

image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2355,6 +2355,7 @@ The `hostname` module shows the system hostname.
 | `format`          | `'[$ssh_symbol$hostname]($style) in '` | The format for the module.                                                                                                            |
 | `style`           | `'bold dimmed green'`                  | The style for the module.                                                                                                             |
 | `disabled`        | `false`                                | Disables the `hostname` module.                                                                                                       |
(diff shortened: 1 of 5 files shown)


oddlama:
Evidently this hasn't gained any traction for almost a year now, so I'm closing this.
If anyone is interested in upstreaming this, feel free to ping me and I'll update the PR.

PR #5031, opened 2023-03-27
distinguish between pwsh and powershell

wmmc88:
#### Description
pwsh.exe and powershell.exe are currently both identified as powershell. This PR intends to distinguish between the two in the shell module (as well as make the necessary context changes to allow for that)

#### Motivation and Context

Distinguishes between pwsh and powershell environments

Closes #

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #5156, opened 2023-04-30
Distinguish between pwsh and powershell

HeyItsGilbert:
#### Description
pwsh.exe and powershell.exe are currently both identified as powershell. This PR intends to distinguish between the two in the shell module (as well as make the necessary context changes to allow for that)

This is a copy of #5031 which was in draft. Hopefully I did this right....

#### Motivation and Context
Distinguishes between pwsh and powershell shell instances. A Windows machine can have both and having a clear distinction can avoid issues.

Closes #5031 

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #5478, opened 2023-10-03
feat(shell): allow distinguishing between pwsh and powershell

HeyItsGilbert:
#### Description
pwsh.exe and powershell.exe are currently both identified as powershell. This PR intends to distinguish between the two in the shell module (as well as make the necessary context changes to allow for that)

~~This is a copy of #5031 which was in draft. Hopefully I did this right....~~
 This is a redo of #5156. 

#### Motivation and Context
Distinguishes between pwsh and powershell shell instances. A Windows machine can have both and having a clear distinction can avoid issues.

Closes #5031 

#### Checklist:


davidkna (changes requested):
To avoid unexpected changes to user-configured shell-module configs, please default `powershell_indicator` to the value of `powershellcore_indicator`. I have provided some suggestions to that end, but please also implement tests for the behavior.

Please also update the English shell module documentation.

HeyItsGilbert:
> To avoid unexpected changes to user-configured shell-module configs, please default `powershell_indicator` to the value of `powershellcore_indicator`. 

I'm assuming you mean default `powershellcore` to `powershell`. PowerShell is what most folks call Windows PowerShell. Where as the open source .Net core version (that's cross platform) is called PowerShell Core. In my previous PR I actually renamed `powershell_indicator` to Windows PowerShell, but I think there was some concern with that. I'll bring this back up in the issue, to make sure we're all on the same page before I put this back up for review, etc.

HeyItsGilbert:
Rebased to hopefully resolve the nightly failed tests which are unrelated to my change but ¯\_(ツ)_/¯. This is ready @davidkna.

docs/config/README.md
```diff
@@ -3814,6 +3814,7 @@ To enable it, set `disabled` to `false` in your configuration file.
 | `fish_indicator`       | `'fsh'`                   | A format string used to represent fish.                      |
 | `zsh_indicator`        | `'zsh'`                   | A format string used to represent zsh.                       |
 | `powershell_indicator` | `'psh'`                   | A format string used to represent powershell.                |
+| `pwsh_indicator`       | `'psh'`                   | A format string used to represent pwsh.                      |
```
davidkna, line 3817: `None` defaults are usually empty in the table, but perhaps add a comment with an explanation. Now that I am thinking about it, setting the default to `$powershell_indicator` instead of `None` should also work in this case.

src/modules/shell.rs
```diff
@@ -200,6 +202,35 @@ mod tests {
...
+            .config(toml::toml! {
+                [shell]
+                pwsh_indicator = "[pwsh](bold cyan)"
+                disabled = false
+            })
+            .collect();
+
+        assert_eq!(expected, actual);
+    }
+
```
davidkna, line 233: Please add a test for the fallback:
```rust
    #[test]
    fn test_pwsh_custom_format_fallback() {
        let expected = Some(format!("{} ", Color::Cyan.bold().paint("pwsh")));
        let actual = ModuleRenderer::new("shell")
            .shell(Shell::Pwsh)
            .config(toml::toml! {
                [shell]
                powershell_indicator = "[pwsh](bold cyan)"
                disabled = false
            })
            .collect();

        assert_eq!(expected, actual);
    }

```

HeyItsGilbert: Good call. Thanks!

docs/config/README.md
```diff
@@ -3814,6 +3814,7 @@ To enable it, set `disabled` to `false` in your configuration file.
 | `fish_indicator`       | `'fsh'`                   | A format string used to represent fish.                      |
 | `zsh_indicator`        | `'zsh'`                   | A format string used to represent zsh.                       |
 | `powershell_indicator` | `'psh'`                   | A format string used to represent powershell.                |
+| `pwsh_indicator`       | `''`                      | A format string used to represent pwsh.                      |
```
davidkna, line 3817:
```suggestion
| `pwsh_indicator`       |                           | A format string used to represent pwsh. The default value mirrors the value of `powershell_indicator `. |
```

docs/config/README.md
```diff
@@ -3825,6 +3826,8 @@ To enable it, set `disabled` to `false` in your configuration file.
 | `style`                | `'white bold'`            | The style for the module.                                    |
 | `disabled`             | `true`                    | Disables the `shell` module.                                 |
 
+`pwsh_indicator` is empty by default and falls back to using `powershell_indicator`.
+
```
davidkna, line 3830:
```suggestion
```

davidkna (commented):
I made some suggestions to the docs, but otherwise this looks good.

davidkna:
LGTM if you run `dprint fmt` on the config file.

HeyItsGilbert:
Checking in @davidkna. Let me know if I need anything else.

davidkna (approved):

davidkna:
Thanks for adding this feature @HeyItsGilbert!

PR #5480, opened 2023-10-04
docs(config): Add quote to unquoted strings, unifying quotes

Macilenta:
docs(config): Add quote to unquoted strings, unifying quotes

There are some unquoted default value strings in module's "options" table, I've quote them with proper quotes.

Looking at all modules' "options" table, there are some default value strings wrapped in double quotes and others that are wrapped in single quotes, unifying them to use single quotes.

#### Description

It looks like all values in `Option` table are quote by backtick, for all types of strings. I've found an unquoted string in C module's option table, and quote it using backtick.

"string" type in TOML file should be quoted using either single slash or double slash, There are some unquoted string in "Default" column of some Modules' `Option` table. I tried to find all of them and have quote them properly.

Most of the "string" type in `Option` table are quoted by single slash, but there are some quoted by double slash, I've unifying them to single slash.

#### Motivation and Context

if "string" type are unquoted in TOML file then the TOML file will be invalid. All the key-value value in "Option" table should form a valid TOML which presents default behavior.

Code should be quoted by backtick in Markdown files, all default values are TOML values, which are either a string, arrays, bool, or something else.

For unifying quotes, the config file which formed by all modules' default value table should follow certain criteria, like linting rules.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -383,7 +383,7 @@ falls back to the `x_security_token_expires` key.
 | `region_aliases`    | `{}`                                                              | Table of region aliases to display in addition to the AWS name.                                             |
 | `profile_aliases`   | `{}`                                                              | Table of profile aliases to display in addition to the AWS name.                                            |
 | `style`             | `'bold yellow'`                                                   | The style for the module.                                                                                   |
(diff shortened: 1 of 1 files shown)


davidkna (approved):

davidkna:
Thanks for the fixes @Macilenta!

PR #5500, opened 2023-10-16
fix: Remove printing the ASCII clear screen code in fish

pe:
#### Description

Remove [printing the ASCII clear screen code in fish](https://github.com/starship/starship/pull/865/commits/eb24540a18a64d250e97dc1c4bca8428b43bb277). It's not needed any more since fish 3.2.

#### Motivation and Context

The clear screen code breaks the highlighting of successful/erroneous return codes in the [shell integration of iTerm2](https://iterm2.com/documentation-shell-integration.html). Removing this workaround works with newer versions of fish. According to https://github.com/fish-shell/fish-shell/issues/3550#issuecomment-751476180 the issue was fixed in March 2021 with version 3.2.0.

This code was added with https://github.com/starship/starship/pull/739 to fix https://github.com/starship/starship/issues/279 and restricted to fish with https://github.com/starship/starship/pull/865.

I hope dropping workarounds for shells older than 2021 to re-enable another feature is OK?

_Workaround_

Unset `STARSHIP_SHELL` after initializing Starship and before loading iterm's shell integration in config.fish.

```fish
starship init fish | source
set -e STARSHIP_SHELL # Breaks iterm shell integration
source {$HOME}/.iterm2_shell_integration.fish
```

SCR-20231016-lgsy

Regressions that I tested:

- <kbd>ALT</kbd>-<kbd>LEFT</kbd>/<kbd>RIGHT</kbd> to change dir in fish works with and without `line_break` and with and without `add_newline = false` (all four combinations) (#279)
- `ls`<kbd>TAB</kbd> works in zsh without `line_break` (#834)
- Multiline editing works in zsh (#824)
- `cd`<kbd>TAB</kbd> works in zsh (#824)
- <kbd>CTRL</kbd>-<kbd>R</kbd> in bash without `line_break` works (#808)

Tested using zsh 5.9, bash 5.2.15 and fish 3.6.1.

#### Checklist:

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -76,12 +76,6 @@ pub fn get_prompt(context: Context) -> String {
         _ => {}
     }
 
-    // A workaround for a fish bug (see #739,#279). Applying it to all shells
-    // breaks things (see #808,#824,#834). Should only be printed in fish.
-    if Shell::Fish == context.shell && context.target == Target::Main {
-        buf.push_str("\x1b[J"); // An ASCII control code to clear screen
-    }
-
     let (formatter, modules) = load_formatter_and_modules(&context);
 
     let formatter = formatter.map_variables_to_segments(|module| {


pe:
Anything I can do to get this PR further?

davidkna:
I am reluctant to merge this at this time, because for example Ubuntu 20.04 still ships affected versions. There have been some discussions about explicit shell integration through starship itself, hopefully this will also resolve this issue.

pe:
OK. I'll close the PR.

ikalnytskyi:
@davidkna FWIW, this erase display hack also breaks OSC 133 in Tmux, and who knows how many other broken terminals out there? Being unable to opt-out of this hack is quite annoying and requires me to patch starship manually. If you are against of removing this hack, can you at least consider adding an opt-out configuration flag?

PR #5507, opened 2023-10-19
fix(zsh): clear exit code between prompts

pstadler:
#### Description
Using Zsh with Status module enabled, the exit code sticks around for as long as there's another command being run.
This change clears the exit code between prompts if no command was being executed, resulting in the correct behavior.

```sh
# BEFORE                                                          # exit code in rprompt
~/GitHub/starship zsh-fix-status-code ❯ return 1
~/GitHub/starship zsh-fix-status-code ❯                           ↵ 1
~/GitHub/starship zsh-fix-status-code ❯                           ↵ 1 # not cleared

# AFTER
~/GitHub/starship zsh-fix-status-code ❯ return 1
~/GitHub/starship zsh-fix-status-code ❯                           ↵ 1
~/GitHub/starship zsh-fix-status-code ❯                               # cleared
```

#### Motivation and Context

Having an irrelevant status code sticking around is slightly annoying.

#### Checklist:

### DIFF
--- a/src/init/starship.zsh
+++ b/src/init/starship.zsh
@@ -23,23 +23,22 @@ else
     }
 fi
 
-
 # The two functions below follow the naming convention `prompt_<theme>_<hook>`
 # for compatibility with Zsh's prompt system. See
 # https://github.com/zsh-users/zsh/blob/2876c25a28b8052d6683027998cc118fc9b50157/Functions/Prompts/promptinit#L155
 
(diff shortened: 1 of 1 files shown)


davidkna (approved):

davidkna:
Thanks for fixing this @pstadler!

pstadler:
Thanks for caring!

PR #5514, opened 2023-10-23
feat: show module when typing command on ZSH for custom command modules

### LINKED ISSUE #5509: Show on command for modules
When working with a lot of tools like `terraform`, `kubectl`, `gcloud`, `aws`, etc there is no way (as far as I'm aware) to dynamically dynamically display the corresponding module. So you either display all modules all the time making the prompt too crowded which is not suited for some smaller terminals like inside an IDE or display none at all. It makes it a lot harder to grab the right info directly.

Powerlevel10k implements a feature called `Show on command`, you provide a list of executables that are tied to a module and the module dynamically appears and disappears when you start typing a command with the binary (or alias to the binary) in it. Starship could support a similar feature.

None

florentinl:
#### Description
This PR implements a `Show on command` feature like Powerlevel10k. It allows some modules to appear when certain keywords appear and disappear from the command line buffer.  

The PR adds a `detect_keywords` field in the configuration file for the custom command module and a `--keywords` flag for the `prompt` subcommand. The custom command module is then showed only when one of the keywords passed in the prompt command is also in the `detect_keywords` entry of the custom command module. Additionally, the prompt is redrawn each time the line editor sees a change to the buffer.

For now, I've only made the implementation for the `custom` module and in the `zsh` init file. But I plan to implement it for other shells, and maybe some other modules in subsequent pull requests.   

Because this change makes the prompt redraw on each input, it definitely has a big impact on performance. It cannot be used in conjunction with modules that take a long time to redraw (for example git_status in large repositories, or custom modules).

To avoid causing problems for people who don't need this feature, I propose to make it optional through a flag (`--enable-detect-keywords`) passed to the init command.

#### Motivation and Context

When working with a lot of tools like terraform, kubectl, gcloud, aws, etc there is no way (as far as I'm aware) to dynamically display the corresponding module. So you either display all modules all the time making the prompt too crowded which is not suited for some smaller terminals like inside an IDE or display none at all. It makes it a lot harder to grab the right info directly.

https://github.com/starship/starship/assets/101594032/370f3c4e-adc3-4ca5-bf52-32003cdd2dc0

I added the `$custom` variable to my prompt as well as the following toml block in the starship.toml:
```toml
[custom.foo]
command = 'echo foo' # shows output of command
detect_keywords= [ 'foo' ]
format = '$output'
```
In my zshrc, I added the following line:
```zsh
eval $(starship init zsh --enable-detect-keywords)
```

I have tested on Windows Terminal with zsh in wsl and in iTerm2 on MacOS as well as Alacritty on Linux

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -6142,6 +6142,13 @@
             "type": "string"
           }
         },
+        "detect_keywords": {
+          "default": [],
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
         "os": {
           "type": [
             "string",
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -4564,6 +4564,7 @@ These modules will be shown if any of the following conditions are met:
 - The current directory contains a file whose name is in `detect_files`
 - The current directory contains a directory whose name is in `detect_folders`
 - The current directory contains a file whose extension is in `detect_extensions`
+- The buffer contains a keyword which is in `detect_keywords` (zsh only)
 - The `when` command returns 0
 - The current Operating System (std::env::consts::OS) matches with `os` field if defined.
 
@@ -4613,6 +4614,7 @@ Format strings can also contain shell specific prompt sequences, e.g.
 | `detect_files`      | `[]`                            | The files that will be searched in the working directory for a match.                                                                                                                                                                                                                         |
 | `detect_folders`    | `[]`                            | The directories that will be searched in the working directory for a match.                                                                                                                                                                                                                   |
 | `detect_extensions` | `[]`                            | The extensions that will be searched in the working directory for a match.                                                                                                                                                                                                                    |
+| `detect_keywords`   | `[]`                            | The keywords that will matched against the content of the command line buffer.                                                                                                                                                                                                                |
 | `symbol`            | `''`                            | The symbol used before displaying the command output.                                                                                                                                                                                                                                         |
 | `style`             | `'bold green'`                  | The style for the module.                                                                                                                                                                                                                                                                     |
 | `format`            | `'[$symbol($output )]($style)'` | The format for the module.                                                                                                                                                                                                                                                                    |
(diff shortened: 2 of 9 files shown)


jankatins:
Wow, I didn't even know this was possible. 

A few curious questions: how often will a redraw happen? E.g. if I type in a regular speed, how many times will it redraw? And: dio the starship calls happen in the frontend or in the background, i.e. what happens if I'm in a big git repo and starship takes half a second to respond, will that mean I don't see typing for half a second each time it's called?

florentinl:
In this first proof of concept, the prompt is redrawn each time a char is inserted. This would mean that if one of your modules takes ages to load the typing experience will be indeed degraded.

I see a few options here to mitigate this:
- Only refresh modules on the rprompt on each input, so that these dynamic modules could be placed there and refreshed easily while the main prompt (containing the more time-consuming modules) would only be refreshed in a new line.
  That's the choice that the OhMyPosh project has made. 
  However this would mean that dynamic modules would necessarily be on the bottom line for multi-line prompts so that doesn't really cover my use-case. (I want the line I'm typing on to not be "free")
- Find a way to cache some modules across starship invocations and invalidate the cache after each new command execution. (far from the scope of this PR I would say)
- Find a way to render starship prompt asynchronously.  (separate problem but would be very nice I think independently of this feature)
- Make this opt-in through some flag in the config file
  I only very rarely work in very large git repos and my custom modules are not very long to compute, so I've yet to encounter the problem after a few days using this version of starship. I guess I must not be the only one for whom it is not a problem.
 
  
  As for the environment config, it would definitely be useful but it wouldn't cover this specific use-case at least for me. I often work in the single repo for the infrastructure in which I will use terraform, kubernetes and one of aws, az or gcloud. I would love to be able to dynamically show one or the other even within the same directory.

florentinl:
To ensure that this feature does not deteriorate the experience of the users that don't want it. I added an optional flag at the init script generation step, to make it optional.

However, this update to the code makes the code coverage check fail. Since there are no tests for this part of the code, I am not sure what to test. I welcome any suggestions.

carlfriedrich:
@florentinl Why has this been closed?

florentinl:
To be honest I have given up on this. It is not that much of a pressing need, and I have no wish to work on this anymore. 
Feel free to update it and reopen it if you want, but I don't think It is useful to have a stale PR open. Sorry

PR #5521, opened 2023-10-25
refactor(modules): use whoami crate to get username

bestgopher:
#### Description

1. get username by `whoami` crate, because the `USER` env may be changed.
2. fix some clippy warning
3. get hostname by `whoami`, and remove the `gethostname` crate from dependencies.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/modules/username.rs
+++ b/src/modules/username.rs
@@ -2,21 +2,15 @@ use super::{Context, Module, ModuleConfig};
 
 use crate::configs::username::UsernameConfig;
 use crate::formatter::StringFormatter;
-
-#[cfg(not(target_os = "windows"))]
-const USERNAME_ENV_VAR: &str = "USER";
-
-#[cfg(target_os = "windows")]
+#[cfg(test)]
 const USERNAME_ENV_VAR: &str = "USERNAME";
 
 /// Creates a module with the current user's username
-///
-/// Will display the username if any of the following criteria are met:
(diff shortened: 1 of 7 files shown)


src/modules/username.rs
```diff
@@ -2,22 +2,15 @@ use super::{Context, Module, ModuleConfig};
...
+#[cfg(not(test))]
+use whoami::username;
 
-#[cfg(not(target_os = "windows"))]
-const USERNAME_ENV_VAR: &str = "USER";
-
-#[cfg(target_os = "windows")]
-const USERNAME_ENV_VAR: &str = "USERNAME";
+#[cfg(test)]
+use tests::username;
```
davidkna, line 9: I would prefer handling this with a small wrapper function, and it seems like a good idea to keep around the possibility of overriding the username in the tests:
```rust
fn username() -> String {
    #[cfg(test)]
    if let Some(user) = context.get_env("USER") {
        return user;
    };
    whoami::username()
}

bestgopher: Thanks for your suggest. I have  optimized the codes, and review again please.

davidkna (approved):

bestgopher:
@davidkna Hi. Is there anything else here that I need to fix?

davidkna:
Thanks for the contribution @bestgopher!

PR #5565, opened 2023-11-10
doc: add Stand With Palestine badge

Hephaestus14089:
Simply adding a 'Stand With Palestine' badge in README.md file.


matchai:
While I personally support the cause, we're we won't be adding more badges to the repo at this time.

PR #5567, opened 2023-11-11
style(character): Trailing whitespace

Sunglas:
#### Description

Removed the trailing whitespace in the `config::chararacter` module.
I also changed the tests for the character module so they would pass with this change.
I can recognize that this is a bit counterintuitive to the tests but the changes were just removing the hard-coded spaces in the format strings.

#### Motivation and Context

Every starship module has a trailing space after it.
I can recognize why this consistency is very valuable in keep formatting the same when composing modules together.
Nevertheless, I can also recognize that the `character` module is usually used in a different context than the other modules.
Since it is used to show a symbol next to where text is entered on a terminal, the trailing whitespace that every module has gets in the way of configuring the symbol  in certain contexts.
Furthermore, from what I have been able to read from past PR's, this trailing whitespace is used to work around emojis acting strangely on different systems.
Because of this, I believe the trailing whitespace in the `character` should be removed since this is a very particularly customizable part of the prompt and leaving the hardcoded whitespace would reduce the configurability of the prompt.

Prompt before change:
before-change

Prompt after change:
after-change

#### Checklist:

PS. Since the documentation doesn't mention the trailing whitespace I have not updated it. As I saw in other PR's, adding a comment explaining why the space was or wasn't removed might also be a good option going forward.

### DIFF
--- a/src/configs/character.rs
+++ b/src/configs/character.rs
@@ -22,7 +22,7 @@ pub struct CharacterConfig<'a> {
 impl<'a> Default for CharacterConfig<'a> {
     fn default() -> Self {
         CharacterConfig {
-            format: "$symbol ",
+            format: "$symbol",
             success_symbol: "[❯](bold green)",
             error_symbol: "[❯](bold red)",
             vimcmd_symbol: "[❮](bold green)",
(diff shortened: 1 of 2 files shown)


davidkna:
The current default works well with the default theme and many of the included presets, if a custom configuration benefits from removing the space it can be removed in the specific configuration.

PR #5614, opened 2023-12-03
feat: Added possibility to have different last character of prompt if the user is "root".

### LINKED ISSUE #5397: Last char distinguish login user and no space?
Traditional shell prompt will have last char that either `%` as normal user or `#` as root. I like starship able to preserve this way

and there's no space between directory and last char, i.e:
```
<shell-prompt>% _      --> traditional shell
<shell-prompt> % _     --> starship if try to mimic traditional one
```
two problems with last char:
1. can't distinguish normal user or root
2. always has a space between other module

ironiq:
#### Description
If sudo is not run as login shell (eg: "sudo -s") and the HOME is kept in sudo, it is hard to recognize the commands are running as root or a simple user. In Linux the default behaviour is to have different last character if user is root. With this change it is possible to set the root symbols separately. 

#### Motivation and Context

Closes #5397 

The idea and a piece of code was taken from module "username". I tried to write proper test for it, but this is my first Rust code, so i'm sure this is not enough. The code was built on linux and i use it now.

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -143,6 +143,8 @@
         "error_symbol": "[❯](bold red)",
         "format": "$symbol ",
         "success_symbol": "[❯](bold green)",
+        "root_error_symbol": "[❯](bold purple)",
+        "root_success_symbol": "[❯](bold blue)",
         "vimcmd_replace_one_symbol": "[❮](bold purple)",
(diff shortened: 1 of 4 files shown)


borice:
This would be great to have.  I think merging is blocked due to test coverage.

PR #5616, opened 2023-12-04
fix(docker_context): ignore unix domain socket path from Docker Context

### LINKED ISSUE #5548: DOCKER_HOST contains unix:// should be skipped
#### Current Behavior
export DOCKER_HOST="unix://$(podman info --format '{{.Host.RemoteSocket.Path}}')"
touch docker-compose.yaml

the shell showup:

via  unix:///run/user/1001/podman/podman.sock

#### Possible Solution

if the docker_host contains `unix://` the display should be skipped

#### Environment
pkg_version:1.16.0
branch:master
commit_hash:
build_time:2023-07-30 13:37:26 +00:00
build_env:rustc 1.71.0 (8ede3aae2 2023-07-12),

bash
#### Starship Configuration
default config

ananta:
fix(modules): ignore unix domain socket path from Docker Context

#### Description
Implemented functionality to slice long Unix domain socket paths in Docker Context within the terminal prompt. This optimization ensures the prompt remains concise and user-friendly, especially in environments where Unix domain socket paths are lengthy.

#### Motivation and Context
This enhancement addresses the issue of excessively long terminal prompts when Unix domain sockets are used in Docker Contexts. The change is aimed at improving the user experience by maintaining a manageable prompt length. Relevant issue:
Closes #5548

Before:
image
After: 
image

#### Checklist:

### DIFF
--- a/src/modules/docker_context.rs
+++ b/src/modules/docker_context.rs
@@ -72,7 +72,9 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 _ => None,
             })
             .map(|variable| match variable {
-                "context" => Some(Ok(ctx.as_str())),
+                "context" => { 
+                    Some(Ok(if ctx.contains("unix://") { "" } else { ctx.as_str() }))
+                },
                 _ => None,
             })
             .parse(None, Some(context))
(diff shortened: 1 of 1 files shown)


src/modules/docker_context.rs
```diff
@@ -72,7 +72,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 _ => None,
             })
             .map(|variable| match variable {
-                "context" => Some(Ok(ctx.as_str())),
+                "context" => Some(Ok(if ctx.contains("unix://") {
+                    ""
+                } else {
+                    ctx.as_str()
+                })),
```
davidkna, line 79: I would prefer handling it like this:
```suggestion
                "context" if !ctx.starts_with("unix://") => ctx.as_str(),
```

davidkna: On the other hand, it might make sense to avoid displaying the module at all and returning `None` like when the value is `default`.

ananta: Thank you for reviewing the PR 😄 . I've updated the codebase accordingly. Let me know if you have any more suggestions!

davidkna: I thought a bit about this, and please do move the `ctx.starts_with("unix://")` above and combine it with the `ctx == "default"` check.

ananta: Sure @davidkna,
But, if we merge the check for the "default" context with the current condition, the `via 🐳` text will be omitted from our prompt when a Docker context is present. Is this an acceptable outcome?

davidkna: Yes.

ananta: @davidkna  Refactored! 
I've squashed the irrelevant commits and retained only the first two for clarity. This ensures that if we need to revert in the future, the necessary changes are readily accessible. Thank you for reviewing the PR!

davidkna (approved):

andytom:
Thank you for your contribution @ananta and thank you for reviewing @davidkna.

PR #5633, opened 2023-12-12
Fix typo in german translation

philipp-schuetz:
#### Description
Fixed a typo in the german translation of the quick install section for powershell.

#### Motivation and Context
Change improves docs

Has not been tested but only changes documentation, so it should not break anything.

#### Checklist:
--


matchai:
Thank you for your contribution, @philipp-schuetz! 😊
We currently manage translations with Crowdin at http://translate.starship.rs/
Submitting a translation on Crowdin will auto-generate a PR with the updates that are made.

PR #5640, opened 2023-12-16
feat: Adding the AWS SSO CLI env variable to profile list

02strich:
#### Description
Adding support for the profile env variable used by https://github.com/synfinatic/aws-sso-cli

#### Motivation and Context
I am using https://github.com/synfinatic/aws-sso-cli as CLI for AWS SSO, which uses a different env variable name than the tools already supported. So I added the additional variable name to the existing list.

I have tested this on MacOS with a local build of starship, together with the most recent version of aws-sso-cli. Tests included correct rendering with sso context as well as with context.

#### Checklist:


davidkna (approved):

andytom:
Thank you for your contribution @02strich and thank you for reviewing @davidkna.

PR #5641, opened 2023-12-16
removed unwrap because it is not so safe

BogdanCiocea:
#### Description
I have replaced unwrap because it is unsafe.

#### Motivation and Context

Closes #

#### Checklist:

- [] I have updated the documentation accordingly.
- [] I have updated the tests accordingly.


davidkna:
[Those `unwrap` calls will never actually fail](https://doc.rust-lang.org/src/std/io/impls.rs.html#393-425). For better clarity and keeping the code more simple, perhaps use `let _ =` instead.

davidkna:
Closing as stale.

PR #5645, opened 2023-12-20
chore(context): explicitly avoid erroring on no git-ceiling-dir-match

davidkna:
#### Description

See: https://github.com/Byron/gitoxide/pull/1191

The default behavior will change with the next version of `gix`, explicitly set `match_ceiling_dir_or_error` to avoid opting into the change.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -319,7 +319,10 @@ impl<'a> Context<'a> {
                 let shared_repo =
                     match ThreadSafeRepository::discover_with_environment_overrides_opts(
                         &self.current_dir,
-                        Default::default(),
+                        gix::discover::upwards::Options {
+                            match_ceiling_dir_or_error: false,
+                            ..Default::default()
+                        },
                         git_open_opts_map,
                     ) {
                         Ok(repo) => repo,


kdarkhan:
@davidkna thanks for handling this. A feature which I want to have for my usecase is to have `match_ceiling_dir_or_error` set to `true`. This would allow me to use starship on slow network drives.

To provide an example, I have a very slow network drive located in dir `/mnt/some/slow/drive`. I can set `GIT_CEILING_DIRECTORIES=/mnt/some/slow/drive`. It works as expected if my `cwd` is in the root of the drive. However when I am somewhere deep in the nested directory, like `/mnt/some/slow/drive/very/deep/nested/path`, `starship` keeps walking up until the root of the dir.

With `match_ceiling_dir_or_error` set to true, I will be able to explicitly define a list of dirs where I expect the git repos to be in, like `/home/user` or `/home/user/workdir`. With that approach, `starship` will ignore early any dir not in `GIT_CEILING_DIRECTORIES`.

Perhaps, I can make `match_ceiling_dir_or_error` configurable in `starship.toml`. Are you ok with that? If yes, I can work on that PR. Or feel free to implement it within your PR as well.

davidkna:
I am fine with a contribution for this, but I will not implement this myself. That said, I'm not sure if changing the semantics of `GIT_CELING_DIRS` is the best approach for solving this.
Beyond that, git isn't the only thing in starship that scans up the directory tree by default, and right now, these don't support `GIT_CELING_DIRS` at all yet.

PR #5657, opened 2023-12-28
fix(direnv): update to work with direnv v2.33

### LINKED ISSUE #5656: Failling to load direnv module for direnv v2.33.0
## Bug Report
Direnv v2.33.0 changed the status of `allowed` from `true` and `false` to a `0, 1, 2` state as explained by this [comment](https://github.com/direnv/direnv/pull/1142#issuecomment-1848677532). To support the current version in a backward compatible way, parsing would need to take into account both ways to expose the `allowed` state. Note that a [new status output](https://github.com/direnv/direnv/issues/1211) might even be considered in the future, which would then require a new fix. 

#### Current Behavior
Won't load direnv module for direnv v2.33.0 because it fails to parse the `allowed` status.

#### Expected Behavior
Successfully load direnv module for direnv v2.33.0.

```bash
[WARN] - (starship::modules::direnv): invalid allow status
```

#### Possible Solution
Check both ways to expose the `allowed` state when parsing from `direnv --status` ouput.

#### Environment
- Starship version: 1.17.0

#### Configuration
```toml
[direnv]
disabled = false
```

camflan:
direnv now returns int enum instead of boolean, https://github.com/direnv/direnv/pull/1158

#### Description
Direnv now returns an int enum value instead of true/false for AllowStatus, https://github.com/direnv/direnv/pull/1158

This allows both `true/false` as well as the `"0"/"1"/"2"` values to work

#### Motivation and Context

Closes #5656

#### Checklist:


docs/config/README.md
```diff
@@ -1225,6 +1225,7 @@ The `direnv` module shows the status of the current rc file if one is present. T
 | `detect_files`      | `['.envrc']`                           | Which filenames should trigger this module.           |
 | `detect_folders`    | `[]`                                   | Which folders should trigger this module.             |
 | `allowed_msg`       | `'allowed'`                            | The message displayed when an rc file is allowed.     |
+| `not_allowed_msg`   | `'not_allowed'`                        | The message displayed when an rc file is not_allowed. |
```
davidkna, line 1228: The config has `not allowed`. Which value is it supposed to be?

camflan: Ah, good catch. Should be “not allowed”

camflan:
```suggestion
| `not_allowed_msg`   | `'not allowed'`                        | The message displayed when an rc file is not_allowed. |
```

src/modules/direnv.rs
```diff
@@ -45,6 +45,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "rc_path" => Some(Ok(state.rc_path.to_string_lossy())),
                 "allowed" => Some(Ok(match state.allowed {
                     AllowStatus::Allowed => Cow::from(config.allowed_msg),
+                    AllowStatus::NotAllowed => Cow::from(config.denied_msg),
```
davidkna, line 48: Alternatively, we could skip adding `not_allowed_msg` for now
```suggestion
                    AllowStatus::NotAllowed => Cow::from(config.not_allowed_msg),
```

camflan: 🤦‍♂️ dang, another good catch

I'd be fine with that too - I don't remember the exact distinction of NotAllowed/Denied but I _think_ it's when you have explicitly denied a file vs not having been granted. 

_Seems_ worthwhile for people to decide if they care about the distinction but for my use, I have them configured in a way that it doesn't matter 🤷‍♂️

davidkna:
Please ensure the tests pass with the recent changes.

camflan:
> Please ensure the tests pass with the recent changes.

good to go

davidkna (approved):
LGTM

andytom:
Thank you for your contribution @camflan and thank you for reviewing @davidkna.

PR #5676, opened 2024-01-02
feat(git_status): add symbol for local branches without an upstream

waldyrious:
#### Description
This PR adds a new symbol called `no_upstream` to the `git status` module, to allow explitly indicating when a branch doesn't have an upstream. By default it is an empty string so the current behavior is maintained.

As this is my first time programming Rust, I based my changes on #2945 and on studying the module code (hence the separate commit making a few changes that make some bits of the code a little bit more readable for someone looking into this codebase for the first time).

#### Motivation and Context
I often get hiccups with git when trying to e.g. push to a branch whose upstream has not been set (or has been unset). I would like to have a symbol in the command line to remind me when that is the case, so that I can save the extra step of running a command, getting an error, setting the upstream and then re-running the original command.

I searched for issues related to this but only found this #2927, which was fixed in PR #2945. Since the changes didn't seem too daunting, I decided to take a stab at implementing the fix myself rather than merely reporting an issue :)

#### Checklist:


src/modules/git_status.rs
```diff
@@ -665,6 +676,25 @@ mod tests {
...
+        let actual = ModuleRenderer::new("git_status")
+            .config(toml::toml! {
+                [git_status]
+                no_upstream="🥷"
+            })
+            .path(repo_dir.path())
+            .collect();
+        let expected = format_output("🥷");
+
+        assert_eq!(expected, actual);
```
waldyrious, line 694: This test is failing (the value of `actual` is `None` rather than the expected string), and I don't know how to fix it.

Is it a bug in the implementation itself, or in the test? I built this test based on the other ones, so I assumed it would be similar, but I never coded Rust before, so it's possible that the test is incorrect. Any help appreciated! :sweat:

image

davidkna: I would want the test to pass before reviewing this in more detail.

waldyrious: Right, me too, but that's precisely where I'm struck at and asking for help with. Would it help if I described what I tried to accomplish with the code changes in natural language or pseudocode?

orhun: Seems like that the failure is directly related to the functionality not being available. Have you tried these changes locally and got the expected result?

PR #5679, opened 2024-01-03
feat: Add `opt_in_modules` config option

### LINKED ISSUE #4312: Add option to disable all modules by default to explicitly enable wanted modules later
I've disabled most of the modules, retaining only a few that I'm interested in. 
That is already quite verbose, but doable.
However, when new modules are added they are enabled by default, forcing me to explicitly disable them again.

Basically, I'd like to opt-in instead of opt-out for modules.

I'd like to have a top-level options like `modulesDisabledByDefault=true`, which inverts the current behavior, and allows me to only mention the modules I'm interested in, in the config.

Alternatively, section would also make sense.
```toml
["*"]
disabled=true
```

I'm aware that I can override the `format` in the [prompt](https://starship.rs/config/#prompt), but as the [faq](https://starship.rs/faq/#do-top-level-format-and-module-disabled-do-the-same-thing) states:

> Disabling modules is more explicit than omitting them from the top level `format`

I would still like the explicitness, and better readability of the module config.

mandx:
#### Description
When `true`, it turns module selection into opt-in (all modules are disabled by default, with some exceptions), instead of opt-out as it is now.

Technically, if `opt_in_modules` is true, the default value for any `disabled` module key will then be false.

The default value for this new option is `false` meaning the default value for any `disabled` module key will be true, as it is now.

#### Motivation and Context
Starship frequently adds new modules to cover the diverse projects and environments of all most/users. Now, from the point of view of individual users, their environment seldom change, and those who care about performance (like me) only disable all modules we know we are not going to use.

In my case, every time I update `starship` I go through the chore of scanning the release note(s) take note of all the newly added modules, determine which I would like to use (which is almost always none), then disable the rest by adding them all to `~/.config/starship.toml` with `disabled = true`.

Closes #4312

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1892,6 +1892,10 @@
       "additionalProperties": {
         "type": "string"
       }
+    },
+    "opt_in_modules": {
+      "default": false,
+      "type": "boolean"
     }
   },
   "additionalProperties": false,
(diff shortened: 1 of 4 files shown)


acuteaangle:
The commit message states:

> Technically, if opt_in_modules is true, the default value for any disabled module key will then be false.

However, I believe this is intended to say the opposite.

PR #5690, opened 2024-01-08
chore: bump libz-sys crates to 1.1.13

ognevnydemon:
so `aarch64-pc-windows-gnullvm` will become buildable again

#### Description
update libz-sys deps

#### Motivation and Context
this is needed for supporting `aarch64-pc-windows-gnullvm` as it contains important fix

patch bumping, I don't think it's needed to be tested :)

#### Checklist:

that's surely not about deps...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1770,9 +1770,9 @@ dependencies = [
 
 [[package]]
 name = "libz-ng-sys"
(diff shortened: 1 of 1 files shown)


davidkna (approved):

davidkna:
Thanks for the contribution!

PR #5734, opened 2024-01-28
fix(bash): improve integration with bash-preexec

davidkna:
#### Description

1. For bash-preexec detection, a check for the official `bash_preexec_imported` variable in addition to the private `__bp_imported` for prior versions (see https://github.com/rcaloras/bash-preexec/issues/146 and https://github.com/starship/starship/issues/5686) I also removed the check for the explicit value `defined` since that matches the official sample check more strongly.
2. Always use `BP_PIPESTATUS` if its length is greater than 0, because the value of `PIPESTATUS` could be unpredictable (https://github.com/starship/starship/pull/4902#discussion_r1456513201).

#### Motivation and Context

Closes #

#### Checklist:


andytom (approved):

davidkna:

PR #5735, opened 2024-01-28
feat(bash): use PS0 for preexec hook

davidkna:
#### Description

If ble.sh and bash-preexec are not available, use `PS0` instead of debug traps on bash 4.4+. This may help performance a bit and will hopefully lead to fewer conflicts with other shell scrips (kitty shell integration?).

#### Motivation and Context

#4973

#### Checklist:


andytom:
I think this might also help with some cases of #1991

andytom (approved):

davidkna:
Thanks for the review again @andytom!

PR #5743, opened 2024-01-31
fix: Implement a better way to check for sudo on Windows

funlennysub:
#### Description
Changes the way starship checks for sudo on Windows. Uses the same [method](https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/src/platform/shell_windows.go#L16-L5 ) as oh-my-posh 

#### Motivation and Context
This method removes cmd window popping out every time you press enter while in sudo, and constant gsudo popouts when you're not.

- Replaced original exe in binary folder and checked if sudo module worked as expected.
- Changes don't affect anything else.
- Tested on Windows 11 23H2

#### Checklist:

_Not sure if these 2 are needed, but lmk_

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,7 +110,8 @@ features = [
   "Win32_UI_Shell",
   "Win32_Security",
   "Win32_System_Threading",
-  "Win32_Storage_FileSystem",
+  "Win32_System_SystemServices",
+  "Win32_Storage_FileSystem"
 ]
 
 [target.'cfg(not(windows))'.dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,7 +111,7 @@ features = [
   "Win32_Security",
   "Win32_System_Threading",
   "Win32_System_SystemServices",
-  "Win32_Storage_FileSystem"
+  "Win32_Storage_FileSystem",
 ]
 
 [target.'cfg(not(windows))'.dependencies]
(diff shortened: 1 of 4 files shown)


davidkna:
The `sudo` module is concerned whether credentials for the `sudo`-CLI utility are cached (`sudo` has been run recently). It looks like this instead checks if the current user has admin permissions. This is already being shown and determined in the `username` module via `is_root_user`:
https://github.com/starship/starship/blob/master/src/modules/username.rs#L73-L94

funlennysub:
Is there a reason then not to use `is_root_user` for `sudo` module on windows since from my understanding it's the same thing anyway but without constant cmd popouts? I haven't used starship on linux with sudo module to see how it behaves there but on windows they seem to be showing the same thing.

davidkna:
> Is there a reason then not to use `is_root_user` for `sudo` module on windows since from my understanding it's the same thing anyway but without constant cmd popouts? I haven't used starship on linux with sudo module to see how it behaves there but on windows they seem to be showing the same thing.

It's not the same thing, and it is supposed to inform about the credential cache, not if the user is root. 
To avoid this sort of issue on Windows, `allow_windows` is `False` by default. That said, I think we had requests asking for more freedom in displaying things if a user is root, but I feel like this module isn't the place for that.

funlennysub:
Makes sense I guess, will wait for "more freedom" then.

PR #5749, opened 2024-02-02
feat(zsh): display named directories

LunarLambda:
Pass Zsh's [`%~` prompt specifier](https://man.archlinux.org/man/zshmisc.1#Shell_state) to `--logical-path` in `PROMPT`/`RPROMPT`

#### Description

Display the current directory using zsh's `%~` specifier, which shortens paths if they refer to a [named directory](https://man.archlinux.org/man/zshexpn.1#Static_named_directories).

#### Motivation and Context

Named directories are a great feature of zsh and most common zsh prompts use the `%~` specifier to display the current directory in shortened form when possible.

Before (With `directory.truncate_to_repo = false` for emphasis)
image
After (named directory set with `hash -d nvim=~/cfg/nvim`)
image

#### Checklist:

Unsure what testing/documentation this needs, since it's just passing an existing flag to `starship prompt`. I would be happy to make any necessary adjustments.

However, I did notice that passing `--logical-path` overrides `directory.truncate_to_repo` completely, I'm not sure if that's intentional/desired.

### DIFF
--- a/src/init/starship.zsh
+++ b/src/init/starship.zsh
@@ -87,6 +87,6 @@ VIRTUAL_ENV_DISABLE_PROMPT=1
 
 setopt promptsubst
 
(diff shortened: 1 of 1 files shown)


LunarLambda:
Alternatively, here's a hack-ish way to do it with `starship init zsh`:

```sh
eval "$(starship init zsh | sed "/^R\?PROMPT=/s/)'$/ --logical-path \"\${(%):-%~}\"&/")"
```

PR #5760, opened 2024-02-06
feat(line_break): add break_below_width option

OddBloke:
#### Description

When specified, this option will change `line_break`'s behaviour to only insert a line break when the width of the terminal is below the given value.

#### Motivation and Context

I value being able to see as much of my terminal history as possible, so my current starship configuration disables `line_break`. This is disadvantageous in narrow terminals, as commands I type are more likely to wrap and my right prompt may not be displayed at all.

Setting an appropriate `break_below_width` value allows users like me to increase the vertical space used by their prompt only when horizontal space is at a premium: I don't "lose" lines of history off the top of my screen except when doing so will give me more space to type and/or mean that my right prompt will be displayed.

These screenshots are of the same prompt with only the width of the terminal changed:

startship-bbw-wide
starship-bbw-narrow

I have built `starship` locally with this change (on an Ubuntu 20.04 system) and tested it by:
* setting `break_below_width` and resizing my terminal across the boundary to observe its behaviour
* unsetting `break_below_width` and confirming that a line break is always emitted

Additionally, there are `cargo test` tests to ensure that the correct output is being emitted.

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -970,6 +970,7 @@
     },
     "line_break": {
       "default": {
+        "break_below_width": null,
         "disabled": false
       },
       "allOf": [
@@ -4325,6 +4326,15 @@
         "disabled": {
           "default": false,
           "type": "boolean"
+        },
+        "break_below_width": {
+          "default": null,
+          "type": [
+            "integer",
+            "null"
+          ],
+          "format": "uint",
+          "minimum": 0.0
         }
       },
       "additionalProperties": false
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2724,9 +2724,10 @@ The `line_break` module separates the prompt into two lines.
 
 ### Options
 
-| Option     | Default | Description                                                        |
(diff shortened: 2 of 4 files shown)


OddBloke:
This change was inspired by my current custom zsh prompt, which will insert a newline if my prompt will leave me below a certain amount of space on-screen:

```sh
    PROMPT_PREFIX="<my actual prompt value>"
    # I want 20 characters of space available assuming 24 characters for RPROMPT
    MAX_PROMPT_LENGTH=$((COLUMNS-24-20))
    PROMPT="$(print "$PROMPT_PREFIX%$MAX_PROMPT_LENGTH(l.\n.)%# ")"
```

(The magic here is in zsh prompt escaping: `%<number>(l.\n.)%` means (approximately; this is my first time revisiting this part of my .zshrc in about a decade :sweat_smile:) "if at least <number> characters have already been printed on this line, then emit `\n` otherwise emit nothing".)

PR #5761, opened 2024-02-07
docs(preset): add conda config to gruvbox-rainbow

ickc:
#### Description

For the gruvbox-rainbow theme, add conda, following the approach of docker context there.

#### Motivation and Context

Provide default including conda. Following the example of docker context there because they have similar meaning. For example, it will show something like ` v3.12.1  🅒  jupyterlab`.

#### Checklist:

### DIFF
--- a/docs/.vuepress/public/presets/toml/gruvbox-rainbow.toml
+++ b/docs/.vuepress/public/presets/toml/gruvbox-rainbow.toml
@@ -21,6 +21,7 @@ $haskell\
 $python\
 [](fg:color_blue bg:color_bg3)\
 $docker_context\
+$conda\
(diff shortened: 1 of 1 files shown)


ickc:
I guess the reason `ignore_base` is defaulted to true is because it is confusing when one is working in another language, and it so happened is loaded (for example in an rc file. I discourage this pattern but many people do this.)

Personally, I wouldn't regard the base as a dev environment such that the context is necessary.

Anyway, the proposed change is minimal and supposed to mimic whatever the theme is treating docker already, and uses the defaults otherwise. Whether the default is good is another discussion.

davidkna (approved):

davidkna:
Thanks for the addition @ickc!

PR #5762, opened 2024-02-07
fix: combine ANSI color codes before wrapping them

OddBloke:
#### Description

The existing code wraps each individual module's output for `context.shell`, concatenates all that output together and passes it to `AnsiStrings` to merge ANSI color codes.  However, the wrapping obscures ANSI color codes, meaning that no merging is possible.

This commit changes the shell-specific wrapping to happen right before output, once all modules' output has been concatenated together.  This results in ANSI color codes being correctly merged, as well as reducing the number of calls to `wrap_colorseq_for_shell` to one.

With a minimal `starship.toml`:

```
format = """$directory"""

[directory]
format = '[a]($style)[b]($style)'
```

The current code produces[0]:

```
\n%{\x1b[31m%}a%{\x1b[0m%}%{\x1b[31m%}b%{\x1b[0m%
```

And this commit's code:

```
\n%{\x1b[31m%}ab%{\x1b[0m%}
```

You can see that the current code emits an additional reset and repeated color code between "a" and "b" compared to the new code.

[0] Produced in a Python shell with:

```
subprocess.check_output(
    "./target/debug/starship prompt", shell=True,
    env={"STARSHIP_CONFIG": "./starship.toml", "STARSHIP_SHELL": "zsh"}
)
```

#### Motivation and Context

This simplifies starship's output on shells which wrap color sequences, aligning them with the behaviour of shells which do not.

I performed the specific testing described above, and am using my locally-built `starship` including the change in my `zsh` shell.

#### Checklist:

(There appear to be neither documentation or tests which cover this behaviour, so I am marking these as done.)


andytom (approved):

davidkna:
Thanks for the contribution @OddBloke, and thanks for reviewing @andytom!

PR #5765, opened 2024-02-10
Allow user to set special colour for exit code 130 (ctrl-c)

Sciencentistguy:
#### Description
This change allows the user to set a colour not only for exit code failure and success, but also exit code 130 (ctrl-c). This is set to yellow by default. This also fixes an issue where even the existing red for exit code failure was cleared by entering vim mode

#### Motivation and Context
I have been using this in a fork for a long time (see the master branch of the source fork of this PR if you really want) and i think others may desire this feature

image

#### Checklist:


Sciencentistguy:
Marking as draft as the tests are not currently passing, and I would like to solicit feedback before putting energy into resolving them

PR #5768, opened 2024-02-11
fix(git_branch): fall back to "HEAD" when there is no current branch

fraserli:
#### Description

According to [the docs](https://github.com/starship/starship/blob/fed7445ecf8959a32a8690a5fa36c4d9765663ba/docs/config/README.md?plain=1#L1799), the `branch` variable in the git branch module should fall back to `HEAD` if there's no current branch (detached HEAD). The code has been changed to do this.

#### Motivation and Context

The existing code sets `branch` to be empty in the case that there isn't a current branch.

#### Checklist:

### DIFF
--- a/src/modules/git_branch.rs
+++ b/src/modules/git_branch.rs
@@ -30,13 +30,13 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         return None;
     }
 
-    let branch_name = repo.branch.as_ref()?;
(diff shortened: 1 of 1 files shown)


davidkna (commented):
LGTM, but I would prefer if you added a test case for this.

fraserli:
Okay, it's done

davidkna (approved):

andytom:
Thank you for your contribution @fraserli and thank you for reviewing @davidkna.

PR #5772, opened 2024-02-12
feat(jj): add jj module to display repository status

### LINKED ISSUE #6076: [FR] Support Jujutsu VCS (git backed)

Add a subsystem for jujutsu

0xdeafbeef:
#### Description
Add jj status module, which shows basics info for [jj](https://github.com/martinvonz/jj) repos.

Closes #6076

Change without branch
image
Change set to some branch
image
Change with description
image

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -275,6 +275,7 @@ $directory\
 $vcsh\
 $fossil_branch\
 $fossil_metrics\
+$jj_status\
 $git_branch\
 $git_commit\
 $git_state\
@@ -2443,6 +2444,78 @@ then the module will also show when there are 0 jobs running.
 symbol = '+ '
 number_threshold = 4
 symbol_threshold = 0
+```
+
+## JJ
+
+The `jj` module reflects the current status of your Jujutsu repository. It shows
+different symbols and formats depending on the state of your work in progress,
+including changes, commits, branch, and more.
+
+The `jj` module is only shown when you're in a directory that is part of a
+Jujutsu repository.
+
+The default functionality includes displaying:
+
+- current change id
+- current commit id
+
+::: warning
+
+This module requires the Jujutsu command line tool to be installed and
+accessible in your system's PATH for it to function correctly.
+
+:::
+
+### Options
+
+| Option                   | Default                                                                                                                                                                                                                                                             | Description                                                                                                                     |
+| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `symbol`                 | `'🍐 '`                                                                                                                                                                                                                                                             | The string used before displaying the repository status.                                                                        |
+| `no_description_symbol`  | `' 📝'`                                                                                                                                                                                                                                                             | Is displayed when change description is empty.                                                                                  |
+| `divergent_symbol`       | `' 💥'`                                                                                                                                                                                                                                                             | Is displayed when commit is in [divergent](https://github.com/martinvonz/jj/blob/main/docs/glossary.md#divergent-change) state. |
+| `format`                 | `'\[$symbol[$change_id_prefix]($change_id_prefix_style)[$change_id_suffix]($change_id_suffix_style) [$commit_id_prefix]($commit_id_prefix_style)[$commit_id_suffix]($commit_id_suffix_style) on [$branch]($branch_style)$no_description_symbol$divergent_symbol\]'` | The format for rendering the module. `symbol`, `style`, and other variables can be included.                                    |
+| `disabled`               | `false`                                                                                                                                                                                                                                                             | Disables the `jj` module.                                                                                                       |
+| `change_id_prefix_style` | `'purple'`                                                                                                                                                                                                                                                          | The style for uniq part of change id.                                                                                           |
+| `change_id_suffix_style` | `'bright-black'`                                                                                                                                                                                                                                                    | The style for the rest of change id.                                                                                            |
+| `commit_id_prefix_style` | `'blue'`                                                                                                                                                                                                                                                            | The style for unique part of commit hash.                                                                                       |
+| `commit_id_suffix_style` | `'bright-black'`                                                                                                                                                                                                                                                    | The style for the rest of commit hash.                                                                                          |
+| `truncation_length`      | `8`                                                                                                                                                                                                                                                                 | Truncates the commit and change id to `truncation_length` characters.                                                           |
+| `branch_style`           | `'purple'`                                                                                                                                                                                                                                                          | The style for the branch name.                                                                                                  |
+
+### Variables
+
+```
+[🍐 rnyynlop 811ca1a4 on push-rnyynlopprno*]
+```
+
+| Variable           | Example   | Description                                                       |
+| ------------------ | --------- | ----------------------------------------------------------------- |
+| `symbol`           | `🍐`      | Mirrors the value of option `symbol`.                             |
+| `change_id_prefix` | `r`       | Unique part of the change id, styled by `change_id_prefix_style`. |
+| `change_id_suffix` | `nyynlop` | The rest of the change id, styled by `change_id_suffix_style`.    |
+| `commit_id_prefix` | `811`     | Unique part of the commit id, styled by `commit_id_prefix_style`. |
+| `commit_id_suffix` | `ca1a4`   | The rest of the commit id, styled by `commit_id_suffix_style`.    |
+| `branch`           | `master`  | The current branch name.                                          |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[jj]
+symbol = "🍐 "
+style = "purple"
+format = "\\[$symbol[$change_id_prefix]($change_id_prefix_style)[$change_id_suffix]($change_id_suffix_style) [$commit_id_prefix]($commit_id_prefix_style)[$commit_id_suffix]($commit_id_suffix_style) on [$branch]($branch_style)$no_description_symbol$divergent_symbol\\]"
+disabled = false
+change_id_prefix_style = "purple"
+change_id_suffix_style = "bright-black"
+commit_id_prefix_style = "blue"
+commit_id_suffix_style = "bright-black"
+truncation_length = 8
+no_description_symbol = "📝"
+divergent_symbol = "💥"
+branch_style = "purple"
 ```
 
 ## Julia
--- a/src/configs/jj.rs
+++ b/src/configs/jj.rs
@@ -1,0 +1,40 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Deserialize, Serialize, Debug)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+#[serde(default)]
+pub struct JJConfig<'a> {
+    pub symbol: &'a str,
+    pub format: &'a str,
+    pub disabled: bool,
+    pub change_id_prefix_style: &'a str,
+    pub change_id_suffix_style: &'a str,
+    pub commit_id_prefix_style: &'a str,
+    pub commit_id_suffix_style: &'a str,
+    pub truncation_length: u8,
+    pub no_description_symbol: &'a str,
+    pub divergent_symbol: &'a str,
+    pub branch_style: &'a str,
+}
+
+impl<'a> Default for JJConfig<'a> {
+    fn default() -> Self {
+        JJConfig {
+            symbol: "🍐 ",
+            format: "\\[$symbol[$change_id_prefix]($change_id_prefix_style)[$change_id_suffix]($change_id_suffix_style) [$commit_id_prefix]($commit_id_prefix_style)[$commit_id_suffix]($commit_id_suffix_style) on [$branch]($branch_style)$no_description_symbol$divergent_symbol\\]",
+            disabled: false,
+            change_id_prefix_style: "purple", // magenta
+            change_id_suffix_style: "bright-black",
+            commit_id_prefix_style: "blue",
+            commit_id_suffix_style: "bright-black",
+            truncation_length: 8,
+            no_description_symbol: " 📝",
+            divergent_symbol: " 💥",
+            branch_style: "purple",
+        }
+    }
+}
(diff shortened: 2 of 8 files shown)


0xdeafbeef:
The only unresolved message is how to make jj module mutual exclusive to git module - show only jj status if it's available, otherwise show git* modules.

0xdeafbeef:
> My suggestion: provide an option to prefer status from `jj` or from `git` than the other.
> 
> Implementation: we can get the repo root with `jj root`. Inside that directory, if both `.jj` and `.git` directories exist, the user (using the above configuration option) will get either the status from 1 of the 2 VCS backends.

I think it's possible to filter the git module's output if jj is present [here](https://github.com/starship/starship/blob/1e4e967f96bca8c311b05a829c123ad466f74553/src/print.rs#L291-L310), but a better mechanism can be implemented.

davidkna:
Is it possible to get this jj status through git, to avoid duplication and perhaps external command spawning?

0xdeafbeef:
> Is it possible to get this jj status through git, to avoid duplication and perhaps external command spawning?

Sadly, no. JJ maintains its own database in the `.jj` directory, and manually parsing it is not a trivial task. Moreover, JJ can be used without a Git repository. However, JJ can also be used as a library.

0xdeafbeef:
I'm daily driving it for a week, haven't found any problems. It resolves in 90ms even it linux repo.
 
  ```
 git_status  -  106ms  -   "[?]"
 jj_status   -   87ms  -   "[🍐 kyysvqks 8fbd179b on <no branch> 📝]"
 directory   -   <1ms  -   "linux "
 git_commit  -   <1ms  -   "(374a7f4) "
 line_break  -   <1ms  -   "\n"
 character   -   <1ms  -   "❯ "
 ```

0xdeafbeef:
Hi, I was wondering if someone could take a look at this PR when you have a chance. Thank you!

src/modules/jj.rs
```diff
@@ -0,0 +1,288 @@
...
+        &[
+            "log",
+            "-r",
+            "@",
+            "-T",
+            template.as_str(),
+            "--no-graph",
+            "--ignore-working-copy",
+        ],
+    )?;
```
poliorcetics, line 28: probably needs a `--color never` too, I've seen badly configured terminals be confused by that

src/modules/jj.rs
```diff
@@ -0,0 +1,288 @@
...
+                    None
+                }
+            }
+            "no_description_symbol" => {
+                if parsed.description.is_empty() {
+                    Some(config.no_description_symbol)
+                } else {
+                    None
+                }
+            }
```
poliorcetics, line 53:
```suggestion
                parsed.description.is_empty().then_some(config.no_description_symbol)
            }
```

src/modules/jj.rs
```diff
@@ -0,0 +1,288 @@
...
+    divergent: &'s str,
+    description: &'s str,
+    branches: &'s str,
+}
+
+impl<'s> JJStatus<'s> {
+    fn new(input: &'s str) -> Option<JJStatus<'s>> {
+        let parsed: Vec<&str> = input.split("#,").collect();
+        if parsed.len() != 7 {
+            log::error!("expected at least 5 parts, got {}", parsed.len());
```
poliorcetics, line 102: I think the log should say 7 ?

src/modules/jj.rs
```diff
@@ -0,0 +1,288 @@
...
+            change_id_prefix,
+            change_id_suffix,
+            commit_id_prefix,
+            commit_id_suffix,
+            divergent,
+            description,
+            branches,
+        ) = (
+            parsed[0], parsed[1], parsed[2], parsed[3], parsed[4], parsed[5], parsed[6],
+        );
```
poliorcetics, line 115: the whole check here could be done using `let else`:

```rust
let [var1, var2] = parsed else {
    // log error
    return None;
};
```

or even `match`:

```rust
match parsed {
    [var1, var2] => Some(S { var1, var2 }),
    _ => {
        // log error
        None
    }
}
```

src/modules/jj.rs
poliorcetics, line 28: On my machine, when a little deep in a repo (not at the root), directly passing the root is faster by ~10ms.

Since it has already been computed in the `jj root` above, it could be passed in the command below:

```rust
let out = context.exec_cmd("jj", &["root", "--ignore-working-copy"])?.stdout;
let root = out.lines().next()?;

// and then
&["-R", &root, "log", ...]
```

poliorcetics: the issue is that it's annoying to mock the command for tests then

src/modules/jj.rs
```diff
@@ -0,0 +1,288 @@
...
+    module.set_segments(formatted);
+
+    Some(module)
+}
+
+fn jj_template(len: u8) -> String {
+    let template = format!(
+        r##"change_id.shortest({len}).prefix() ++ "#," ++ change_id.shortest({len}).rest() ++ "#," ++ commit_id.shortest({len}).prefix() ++ "#," ++ commit_id.shortest({len}).rest() ++ "#," ++ divergent ++ "#," ++ description ++ "#," ++ branches.join(",")"##,
+    );
+    template
```
poliorcetics, line 84:
```suggestion
    format!(
        r##"change_id.shortest({len}).prefix() ++ "#," ++ change_id.shortest({len}).rest() ++ "#," ++ commit_id.shortest({len}).prefix() ++ "#," ++ commit_id.shortest({len}).rest() ++ "#," ++ divergent ++ "#," ++ description ++ "#," ++ branches.join(",")"##,
    )
```

src/modules/jj.rs
```diff
@@ -0,0 +1,288 @@
...
+    };
+    let formatted = formatter
+        .map_meta(|s, _| match s {
+            "symbol" => Some(config.symbol),
+            "divergent_symbol" => {
+                if parsed.divergent == "true" {
+                    Some(config.divergent_symbol)
+                } else {
+                    None
+                }
```
poliorcetics, line 45:
```suggestion
                (parsed.divergent == "true").then_some(config.divergent_symbol)
```

docs/config/README.md
```diff
@@ -2445,6 +2445,78 @@ number_threshold = 4
...
+| `symbol`                 | `'🍐 '`                                                                                                                                                                                                                                                             | The string used before displaying the repository status.                                     |
+| `no_description_symbol`  | `' 📝'`                                                                                                                                                                                                                                                             | Is displayed when change description is empty.                                               |
+| `divergent_symbol`       | `' 💥'`                                                                                                                                                                                                                                                             | Is displayed when commit in divergent state.                                                 |
```
poliorcetics, line 2475:
```suggestion
| `divergent_symbol`       | `' 💥'`                                                                                                                                                                                                                                                             | Is displayed when commit is in divergent state.                                                 |
```

what's "divergent" here ? conflicted ?

docs/config/README.md
```diff
@@ -2445,6 +2445,78 @@ number_threshold = 4
...
+| `format`                 | `'\[$symbol[$change_id_prefix]($change_id_prefix_style)[$change_id_suffix]($change_id_suffix_style) [$commit_id_prefix]($commit_id_prefix_style)[$commit_id_suffix]($commit_id_suffix_style) on [$branch]($branch_style)$no_description_symbol$divergent_symbol\]'` | The format for rendering the module. `symbol`, `style`, and other variables can be included. |
```
poliorcetics, line 2476: The issue here is that some parts will be displayed even if there are no associated data. For example, on a change with no branch, a description and no divergence, this prompt will end with ` on `, which is not wanted

poliorcetics: Ah I see you worked around it with `<no branch>` but given that's the default state in JJ, I don't think adding 14 characters (`on <no-branch>`) for this case is worth it. If there is no branch, this whole part should probably no be displayed

davidkna:
> However, JJ can also be used as a library.

I would rather avoid that as long as it still depends on `git2`.

In general regarding the module, I would prefer if the module here was named and structured in line with other starship vcs modules for git/hg/fossil.
Starship doesn't provide general vcs modules at this time, but it would be nice for the modules to follow a common structure.

poliorcetics:
> However, JJ can also be used as a library.

Even if gitoxide was used instead inside JJ, there would still be the problem of future or private backends, which `jj-lib` won't be aware of if statically compiled in `starship`. As long as there is no dynamic loading or discovery of some sort, using the CLI is more future-proof

src/modules/jj.rs
0xdeafbeef: done

src/modules/jj.rs
0xdeafbeef: done

src/modules/jj.rs
0xdeafbeef: nice catch!

src/modules/jj.rs
0xdeafbeef: done

src/modules/jj.rs
0xdeafbeef: done

src/modules/jj.rs
0xdeafbeef: done

src/modules/jj.rs
0xdeafbeef: done

docs/config/README.md
0xdeafbeef: https://github.com/martinvonz/jj/blob/dce99cf1f83bfc54a4130c4fe2b1c1d0a81a980c/docs/glossary.md#divergent-change

docs/config/README.md
0xdeafbeef: The issue is that you need to apply styling to both `on` and the branch name, showing `on` conditional based on the presence of the branch name.

docs/config/README.md
timcki, line 1: I suggest disabling the formatter here to avoid cluttering the PR.

src/configs/jj.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub no_description_symbol: &'a str,
+    pub divergent_symbol: &'a str,
+    pub branch_style: &'a str,
+}
+
+impl<'a> Default for JJConfig<'a> {
+    fn default() -> Self {
+        JJConfig {
+            symbol: "🍐 ",
+            // format: "[[$change_id_prefix]($change_id_prefix_style)[$change_id_suffix]($change_id_suffix_style) [$commit_id_prefix]($commit_id_prefix_style)[$commit_id_suffix]($commit_id_suffix_style)$branch]($style)",
```
timcki, line 28: is the comment here left intentionally?

src/formatter/model.rs
timcki, line 1: I don't think these structs need to derive `Debug`

src/modules/jj.rs
```diff
@@ -0,0 +1,290 @@
...
+            }
```
timcki, line 53: I suggest to replace this block with
```suggestion
           "divergent_symbol" => parsed
                .divergent
                .eq("true")
                .then_some(config.divergent_symbol),
```
to keep it more consistent with the comparison below

src/configs/jj.rs
0xdeafbeef: Removed this

0xdeafbeef:
@davidkna, do you want me to refactor this PR to match the structure of #5814? I can do that, but I'm still unsure about how to handle output that conflicts with Git. For example, if we break down the `jj` status into several modules like we do with Git, the Starship prompt will display `[master][⎇ master]` for both `jj` and Git outputs. With the current grouping, it doesn't look as strange:
`[🍐 tknkszzp 56b48ad0 on master*][⎇ master]`

poliorcetics:
I think we could introduce a config to say "disable git if jj is present", since if both are there, jj is very probably the one wanted by users.

Something like

```toml
[jj]
disable-git = true
```

there are several cases as a result:

1. No Git, no JJ: nothing
2. Git, no JJ: normal git modules
3. No Git, JJ enabled, JJ: normal jj module(s)
4. Git, JJ enabled, JJ: both git and jj modules
5. Git, JJ enabled, JJ, git-disabling: only jj module(s)

poliorcetics:
If we make the JJ module disabled by default (or not in the default status line) the cost of checking should be one easily cached boolean

0xdeafbeef:
> @davidkna, do you want me to refactor this PR to match the structure of #5814? I can do that, but I'm still unsure about how to handle output that conflicts with Git. For example, if we break down the `jj` status into several modules like we do with Git, the Starship prompt will display `[master][⎇ master]` for both `jj` and Git outputs. With the current grouping, it doesn't look as strange: `[🍐 tknkszzp 56b48ad0 on master*][⎇ master]`

@davidkna can you please comment on this?

poliorcetics:
Maybe what we need is not separate modules per VCS but instead a single "VCS" module with a behavior like this:

```toml
[vcs]
# Order in which the VCSs are tried for the current directory.
# The first successful one is picked
order = [
  "jj",
  "git",
  "pijul",
  # ...
] # The default would be git first probably, to make the first check successful for most people ?

# and the final status line would replace the whole "$vcs" module with the relevant format
jj_format = "jj: $jj_change$jj_status"
git_format = "git: $git_branch$git_state"
```

People could then still configure their modules independently and (de)activate one by one but the final statusline would only ever display one VCS.

If they want several VCS, they could still directly use the relevant modules in the directly in the overall statusline

What do you all think ?

davidkna:
@0xdeafbeef Please do split the module up.

docs/config/README.md
```diff
@@ -2445,6 +2446,78 @@ number_threshold = 4
...
+| `commit_id_prefix` | `811`     | Unique part of the commit id, styled by `commit_id_prefix_style`. |
+| `commit_id_suffix` | `ca1a4`   | The rest of the commit id, styled by `commit_id_suffix_style`.    |
+| `branch`           | `master`  | The current branch name.                                          |
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[jj]
```
Nemo157, line 2506:
```suggestion
[jj_status]
```

The docs here are out of sync with the code. Also needs changing in the initial prose.

teto:
I've tried applying this on top of v1.22.1, the module is recognized  and everything looks fine but I can't output anything `starship module jj_status` shows nothing. 
I've tried
```
[jj_status]
format = "TOTO"
disabled = false
```
to force an output but nothing. The patch is applied on top of starship 1.22.1

poliorcetics:
As per discord messages with the author, consider this MR abandoned. I'm working on #6388 and then I'll add `jj_*` modules on top of it in their own MRs

phucisstupid:
any update?

poliorcetics:
This can be closed and considered replaced by #7612 now that #6388 has been merged

PR #5775, opened 2024-02-14
fix(character): also handle vi edit mode in pwsh

HeyItsGilbert:
After landing #5478 vi edit mode doesn't update the character as expected. This adds the new Shell type to properly show it on Pwsh.

#### Description
Mentioned in https://github.com/starship/starship/pull/5478#issuecomment-1886829331 that this broke. This works correctly in powershell.exe, but char doesn't update in pwsh.exe

#### Motivation and Context
This will fix the character change for Pwsh.

#### Checklist:


davidkna (approved):

davidkna:
Thanks for fixing this @HeyItsGilbert!

PR #5814, opened 2024-03-02
feat(jj): add module for Jujutsu current change ID

poliorcetics:
#### Description

Adds support for getting and displaying the current Jujutsu change ID (when in a Jujutsu repo)

#### Motivation and Context

Closes https://github.com/starship/starship/issues/6076

[Jujutsu](https://github.com/martinvonz/jj) is a new VCS that is change-based.

I want to do several PRs to add various parts of its informations to starship, starting with the  simplest one, the current change ID.

Explained prompt, with the JJ change ID module:

Screenshot 2024-03-02 at 13 45 29

Result of explained prompt above:

Screenshot 2024-03-02 at 13 45 50

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -52,6 +52,10 @@ pub struct Context<'a> {
     /// Private field to store Git information for modules who need it
     repo: OnceCell<Repo>,
 
+    /// Private field containing the JJ root, if one was found. Useful to skip computing JJ modules
+    /// when unnecessary because a previous one already checked.
+    jj_repo: OnceCell<Option<Box<str>>>,
+
     /// The shell the user is assumed to be running
     pub shell: Shell,
 
@@ -168,6 +172,7 @@ impl<'a> Context<'a> {
             logical_dir,
             dir_contents: OnceCell::new(),
             repo: OnceCell::new(),
+            jj_repo: OnceCell::new(),
             shell,
             target,
             width,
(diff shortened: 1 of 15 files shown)


poliorcetics:
I see #5772 exists, it uses the approach of computing everything at once and only displaying the wanted elements, I'm not sure which starship prefers

necauqua:
I'd strongly prefer the jj calls would have `--ignore-working-copy` to avoid snapshots being made literally on every prompt. Somebody might want it I guess, but at least make it configurable then.

Not snapshotting works fine for change ids and branches, for diffs you'd either give up and do snapshots or wait for `--no--commit-transaction` option to be a thing

poliorcetics:
@necauqua you're right, and `--ignore-working-copy` is much faster too.

I also added the `--repository`, I though it would interfere with testing but since I mocked `jj root` too, it doesn't :tada:

The step has gone from 48ms before to 35ms with those

ckoehler:
Would love to also be able to indicate whether the change is empty or missing a description. Thanks! Happy to help move this along...

src/context.rs
```diff
@@ -363,6 +368,19 @@ impl<'a> Context<'a> {
             })
     }
 
+    pub fn get_jj_repo(&self) -> Option<&str> {
+        self.jj_repo
+            .get_or_init(|| {
+                let out = self.exec_cmd("jj", &["root", "--ignore-working-copy"])?;
```
davidkna, line 374: I think it might be better to do `begin_ancestor_scan` first for better performance.

davidkna (changes requested):

poliorcetics:
Closing because I started work on https://github.com/starship/starship/issues/5932 instead

PR #5825, opened 2024-03-06
fix(cmd_duration): Make render_time format more consistent

### LINKED ISSUE #5822: Make `render_time` format more consistent
## Feature Request

Not a feature not a bug, just an enhancement:

Currently, the `cmd_duration` module does not render time components which equal `0`, meaning that the format is inconsistent in its length when multiple components are involved, making it hard to know what I am looking at at a glance:

- `7260` seconds will render as `2h1m`
- `7201` seconds will render as `2h1s`
- `7261` seconds will render as `2h1m1s`

- Render all components after the first non-zero component even if they are zero, so the examples above would become to make them scannable at a glance:
    - `7260` seconds would render as `2h1m0s`
    - `7201` seconds would render as `2h0m1s`
    - `7261` seconds would render as `2h1m1s`

- Always reduce the output to the two (?) highest components to make it consistent but compact without loosing too much significant information. Probably viable but would kind of break the `show_milliseconds` config
    - `7290` seconds would render as `2h1m`
    - `90100` seconds would render as `1d1h`
- Leave it as it is. Has the advantage of being no change and no work, but the disadvantage of being inconsistent to scan for no real gain

I would be happy to implement this

LionC:
#### Description
- Make timing format always show enabled components after te\he first nonzero component (see #5822 )
- Reduce the dynamic string allocation part of the `render_time` function to a single string allocation by pre-allocating the (pretty much fixed size for a given input) string
- Make `render_time` return `0s` instead of `0ms` when `show_millis` is `false` (that result is not rendered anyways though)
- Prefix the tests for the `render_time` function with `render_time_` to make them runnable via `cargo test render_time`

#### Motivation and Context

Closes #5822

- Passed the old tests except the ones it shouldn't
- Added new tests

#### Checklist:
- [x] I have updated the documentation accordingly. (n/a)

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -705,7 +705,7 @@ credential_process = /opt/bin/awscreds-retriever
                 .collect();
 
             let possible_values = [
-                "30m2s", "30m1s", "30m", "29m59s", "29m58s", "29m57s", "29m56s", "29m55s",
+                "30m2s", "30m1s", "30m0s", "29m59s", "29m58s", "29m57s", "29m56s", "29m55s",
             ];
             let possible_values = possible_values.map(|duration| {
                 let segment_colored = format!("☁️  astronauts (ap-northeast-2) [{duration}] ");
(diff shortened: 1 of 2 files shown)


andytom (approved):

andytom:
Thank you for your contribution @LionC

tchajed:
Under the `cmd_duration` config I had `min_time = 200` and I believe this PR changed slow commands between 200-999ms to display as "0s" with that config.

In case anybody else runs into this: I had to add `show_milliseconds = true` to fix it.

PR #5826, opened 2024-03-07
docs: Fix git status $all_status documentation

brandondong:
#### Description

Fix incorrect documentation for the git status `$all_status` variable value.

#### Motivation and Context

After https://github.com/starship/starship/pull/4829/, the documentation diverged from the actual variable value in code.

#### Checklist:


davidkna:
Thanks for the updating the documentation, but please only update the English docs. This helps avoid conflicts with crowdin, the translated files will be automatically updated.

davidkna:
Thanks for the fix @brandondong!

PR #5830, opened 2024-03-08
docs(python): Remove a bad config example in the python

### LINKED ISSUE #5582: Documentation should not promote unsafe practice of running arbitrary "python" executable
The config example in https://github.com/starship/starship/blob/ba9f209a5fa739e6d183d896ed22550fb6a38f34/docs/config/README.md?plain=1#L3550-L3555
could lead to execution of arbitrary code by running `./venv/bin/python --version`.
Just because an executable resides in a venv/bin/ folder and is called python it shouldn't automatically be assumed that it's an actual version of Python, the interpreter.

This example should either be removed, or be kept with a warning to turn it into kind of a counter example.

jankatins:
Running a unknown binary in a subfolder (in this case `.venv/bin/python`) has at least two drawbacks with some security implication:

- In case of a unknown directory, simply cd'ing into a subdirectory could lead to starship executing a unknown binary (usually the binary MUST be in path so it's a decision taken by the user, e.g. by activating the venv)
 - starship would show the result of the venv python even if the venv is not active but calling `python3` on the cli would NOT run that `python3` in the venv but a different one (most likely the system python3).

Activating the env will of course display the python version within the venv. But that was a conscious decision of the user to activate the venv.

Closes #5582

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3508,7 +3508,7 @@ current [Python virtual environment](https://docs.python.org/tutorial/venv.html)
 If `pyenv_version_name` is set to `true`, it will display the pyenv version
 name. Otherwise, it will display the version number from `python --version`.
 
-By default the module will be shown if any of the following conditions are met:
+By default, the module will be shown if any of the following conditions are met:
 
 - The current directory contains a `.python-version` file
 - The current directory contains a `Pipfile` file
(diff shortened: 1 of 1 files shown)


andytom (approved):

davidkna:
Thanks for updating the docs @jankatins and thanks for reviewing @andytom!

PR #5833, opened 2024-03-09
feat(username): add detect_env_vars as option

mickimnet:
Hello everybody,

#### Motivation and Context

I don't use the default `username` module and instead use a custom module because I frequently use terminal multiplexers like `tmux`, `screen`, or `Zellji`. These multiplexers display the hostname in the status bar, so I don't want or need to see it in the terminal as well.

#### Description

To achieve this, I added the `context::detect_env_vars` function to the `username` module, analogous to the `hostname` module.

- [X] I have tested using **macOS**

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1717,6 +1717,7 @@
     },
     "username": {
       "default": {
+        "detect_env_vars": [],
         "disabled": false,
         "format": "[$user]($style) in ",
         "show_always": false,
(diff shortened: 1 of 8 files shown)


jankatins:
This use case could be handled by https://github.com/starship/starship/pull/4439, at least I would assume that one would anyway need a custom env var to enable it (assuming the default is to disable it when in a multiplexer).

mickimnet:
> This use case could be handled by #4439, at least I would assume that one would anyway need a custom env var to enable it (assuming the default is to disable it when in a multiplexer).

The difference is, that with #4439 one would need an external tool—like `direnv` to set an additional environment variable. This configuration addition works without any external tool. And follows the implementation of #5166 and #5196.

davidkna (approved):

andytom:
Thank you for your contribution @mickimnet and thank you for reviewing @davidkna.

PR #5834, opened 2024-03-09
docs(bun): Add nodejs override to bun config documentation

alper:
As discussed in #5805 using the bun module creates a "double prompt" situation which is easily remedied in user config.

Adding it to the default shipped config I think does not make sense.


davidkna:
Please put the example after the variables sections and add a heading, like the examples for other modules.

alper:
Done.

davidkna (approved):

davidkna:
Thanks for improving the documentation @alper!

PR #5844, opened 2024-03-16
[docs] Simplify Nushell integration

texastoland:
#### Description

Simplifies Nushell integration instructions *(docs only)* in the following ways:

- Saves the script to a standard cross-platform path
- Just `use starship.nu` instead of arbitrary path
- Provides runnable commands
- Explains why it's necessary 
- Explains how to update

#### Motivation and Context

Motivated by jdx/mise-docs#46 for consistency between tools.

#### Checklist:

- [ ] I have updated the tests accordingly *(N/A)*.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -137,24 +137,24 @@ description: Starship is the minimal, blazing fast, and extremely customizable p
 
    #### Nushell
 
-   ::: warning
+   Since Nu does [not support `eval`](https://www.nushell.sh/book/how_nushell_code_gets_run.html#eval-function) you must save the initialization script:
 
-   This will change in the future.
-   Only Nushell v0.78+ is supported.
-
-   :::
-
-   Add the following to the end of your Nushell env file (find it by running `$nu.env-path` in Nushell):
-
-   ```sh
-   mkdir ~/.cache/starship
-   starship init nu | save -f ~/.cache/starship/init.nu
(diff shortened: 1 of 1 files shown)


davidkna (changes requested):
It hasn't been more stable lately, but nushell has had many breaking changes and our init script also gets updated now and then, so I would like to avoid having the init script static at this time.

texastoland:
> I would like to avoid having the init script static at this time.

It was already static. That's the same. I added a link to explain why:

```diff
+   Since Nu does [not support `eval`](https://www.nushell.sh/book/how_nushell_code_gets_run.html#eval-function) you must save the initialization script:
```

davidkna:
FWIW, it is not necessarily static between different versions of starship (or with different starship binary locations).

More importantly, if the file doesn't update itself, it will make handling user issues more difficult and will cause additional strife on breaking nushell updates. Given all that, I'd rather not accept this PR.

texastoland:
> * Saves the script to a standard cross-platform path
> * Just `use starship.nu` instead of arbitrary path
> * Provides runnable commands
> * Explains why it's necessary
> * Explains how to update

@davidkna I think you're misunderstanding. It is and always was "static" because Nu doesn't have an "eval" feature. My documentation only explains that and simplifies the process. Request to reopen CC @matchai.

PR #5845, opened 2024-03-17
docs(nu): Simplify Nushell integration instructions 

texastoland:
Implements @davidkna's astute feedback in #5844. Always runs unless non-interactive (a script). Clarifies and simplifies installation while reducing (possibly concurrent) file writes. RFC @amtoine (previous committer) @fdncred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,21 +344,15 @@ Since Nu does [not support `eval`](https://www.nushell.sh/book/how_nushell_code_
 ```nushell
 '
 let starship_path = $nu.default-config-dir | path join scripts starship.nu
-if $nu.is-interactive {
-  starship init nu | save $starship_path --force
-}' | save $nu.env-path --append
-'
-if $nu.is-interactive {
-  use starship.nu
-}' | save $nu.config-path --append
+starship init nu | save $starship_path --force
+' | save $nu.env-path --append
+"\nuse starship.nu" | save $nu.config-path --append
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -339,21 +339,28 @@ eval $(starship init ion)
...
+```nushell
+'
+let starship_path = $nu.default-config-dir | path join scripts starship.nu
+if $nu.is-interactive {
+  starship init nu | save $starship_path --force
+}' | save $nu.env-path --append
+'
+if $nu.is-interactive {
+  use starship.nu
+}' | save $nu.config-path --append
```
amtoine, line 353: as far as i know, this does not work :thinking: 
did it change on the latest `main` of Nushell?

amtoine (commented):
i think i find the
```nushell
'
...
' | save $nu.env-path --append
```
quite harder to read :confused:

README.md
texastoland: Yes you can run it directly to try 👍🏼

texastoland:
> quite harder to read 😕

Because it follows a convention of providing a command to run instead of code to add 👍🏼 The previous instructions presume a user is already familiar with the [configuration docs](https://www.nushell.sh/book/configuration.html#configuration). This "just works" but still provides enough information for expert users like us.

README.md
texastoland: Removed the conditional check based on additional feedback that it doesn't work in all cases.

PR #5855, opened 2024-03-19
feat(username): add aliases option

alper:
As described in #5784 a way to translate the displayed username into something else.


andytom:
Thanks for your contribution @alper and thank you for reviewing @davidkna.

PR #5859, opened 2024-03-21
fix: replace all remaining paths referring to vuepress

davidkna:
#### Description

Some files were still referencing the old locations used by vuepress instead of those used by vitepress. The incorrect in `Cargo.toml` is likely the cause for the failed publishing to crates.io.

#### Motivation and Context

Related #5865

#### Checklist:

### DIFF
--- a/.dprint.json
+++ b/.dprint.json
@@ -14,7 +14,7 @@
   ],
   "excludes": [
     "CHANGELOG.md",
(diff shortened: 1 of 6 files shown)


andytom (approved):

davidkna:
Thanks for all the reviews @andytom!

PR #5873, opened 2024-03-24
feat(odin): Add Odin Lang module

error-27:
#### Description

This adds a new module that shows the current Odin version when in an Odin project. This includes a separate option to show commit numbers, and doesn't include a version format string because of Odin's irregular version numbering.

#### Motivation and Context

I noticed Odin was not supported and I felt like it should be added.

Without commit number:
image
With commit number:
image

#### Checklist:


src/utils.rs
```diff
@@ -326,6 +326,10 @@ active boot switches: -d:release\n",
             stdout: String::from("4.10.0\n"),
             stderr: String::default(),
         }),
+        "odin version" => Some(CommandOutput {
+            stdout: String::from("odin version dev-2024-03:fc587c507"),
```
davidkna, line 330:
```suggestion
            stdout: String::from("odin version dev-2024-03:fc587c507\n"),
```

src/modules/odin.rs
```diff
@@ -0,0 +1,89 @@
...
+        let expected = None;
+        assert_eq!(expected, actual);
+        dir.close()
+    }
+
+    #[test]
+    fn folder_with_odin_file() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        File::create(dir.path().join("main.odin"))?.sync_all()?;
+        let actual = ModuleRenderer::new("odin").path(dir.path()).collect();
```
davidkna, line 81: Please add a test that mocks the odin command to an output without the extra hash (my build does not have it), e.g. `"odin version dev-2024-03\n"`. You can use `cmd` from `ModuleRenderer`

davidkna (changes requested):

src/modules/odin.rs
error-27: I think it's done now in the latest commit

davidkna (approved):
LGTM

andytom:
Thank you for your contribution @error-27 and thank you for reviewing @davidkna.

PR #5887, opened 2024-03-26
fix: replace unmaintained crates `yaml-rust`, `dirs-next`

davidkna:
#### Description

`yaml-rust` was declared unmaintained, the suggested alternative is `yaml-rust`.

`dirs-next` is not declared unmaintained outright, but seems close to being unmaintained. At least it looks like `dirs` again has more maintenance right now.

#### Motivation and Context

Closes #

#### Checklist:


andytom (approved):

davidkna:
Thanks for the review!

PR #5900, opened 2024-03-30
feat(nats): Add NATS Context

hooksie1:
#### Description
Adds a module for getting the current NATS context. Uses the nats cli to get the current context name.

#### Motivation and Context
This is similar to having the Kubernetes context displayed so it's always known which context you are in.

Screenshot 2024-03-30 at 17 41 46

Added the mock nats cli and then created tests to make sure styling matched.
Also built locally and tested changing contexts and styles.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2904,7 +2904,7 @@ The `nats` module shows the name of the current [NATS](https://nats.io) context.
 
 | Option     | Default                    | Description                                                  |
 | ---------- | -------------------------- | ------------------------------------------------------------ |
-| `symbol`   | `'✉️ '`                       | The symbol used before the NATS context (defaults to empty). |
+| `symbol`   | `'✉️ '`                     | The symbol used before the NATS context (defaults to empty). |
 | `style`    | `'bold purple'`            | The style for the module.                                    |
 | `format`   | `'[$symbol$name]($style)'` | The format for the module.                                   |
 | `disabled` | `false`                    | Disables the `nats` module.                                  |
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -338,6 +338,7 @@ $aws\
 $gcloud\
 $openstack\
 $azure\
+$nats\
 $direnv\
 $env_var\
 $crystal\
@@ -2893,6 +2894,35 @@ The `hg_branch` module shows the active branch and topic of the repo in your cur
 format = 'on [🌱 $branch](bold purple)'
 truncation_length = 4
 truncation_symbol = ''
+```
+
+## NATS
+
+The `nats` module shows the name of the current [NATS](https://nats.io) context.
+
+### Options
+
+| Option     | Default                    | Description                                                  |
+| ---------- | -------------------------- | ------------------------------------------------------------ |
+| `symbol`   | `''`                       | The symbol used before the NATS context (defaults to empty). |
+| `style`    | `'bold purple'`            | The style for the module.                                    |
+| `format`   | `'[$symbol$name]($style)'` | The format for the module.                                   |
+| `disabled` | `false`                    | Disables the `nats` module.                                  |
+
+### Variables
+
+| Variable | Example     | Description                          |
+| -------- | ----------- | ------------------------------------ |
+| name     | `localhost` | The name of the NATS context         |
+| symbol   |             | Mirrors the value of option `symbol` |
+| style\*  |             | Mirrors the value of option `style`  |
+
+### Example
+
+```toml
+[nats]
+format = '[$symbol]($style)'
+style = 'bold purple'
 ```
 
 ## Nim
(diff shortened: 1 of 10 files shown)


hooksie1:
I'm not sure what's going on with the print all tests.

hooksie1:
Ah, helps if I read the test output correctly.

src/configs/nats.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub symbol: &'a str,
+    pub style: &'a str,
+    pub disabled: bool,
+}
+
+impl<'a> Default for NatsConfig<'a> {
+    fn default() -> Self {
+        NatsConfig {
+            format: "[$symbol$name]($style)",
+            symbol: "",
```
davidkna, line 21: I think this should have a symbol by default.

src/modules/nats.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    if config.disabled {
+        return None;
+    };
+
+    let ctx_str = context
+        .exec_cmd("nats", &["context", "info", "--json"])?
+        .stdout;
+    let nats_context: json::Value = json::from_str(&ctx_str).ok()?;
+    let name = nats_context.get("name")?.as_str()?;
```
davidkna, line 19: I feel like failure to parse the JSON should be logged here.
Additionally, I would prefer to move the `nats_context.get("name")` into the `name =>` `match`, but you can skip that if that causes any issues.

src/configs/nats.rs
hooksie1: Yeah I'm not sure what to put here lol.

src/modules/nats.rs
hooksie1: yeah I can do that. I think I looked at another module to see how others did it and assumed the error bubbled up.

src/configs/nats.rs
hooksie1: I used an envelope for "messages"

docs/config/README.md
```diff
@@ -2895,6 +2896,35 @@ truncation_length = 4
...
 
+## NATS
+
+The `nats` module shows the name of the current [NATS](https://nats.io) context.
+
+### Options
+
+| Option     | Default                    | Description                                                  |
+| ---------- | -------------------------- | ------------------------------------------------------------ |
+| `symbol`   | `''`                       | The symbol used before the NATS context (defaults to empty). |
```
davidkna, line 2907:
```suggestion
| `symbol`   | `'✉️ '`                    | The symbol used before the NATS context (defaults to empty). |
```

src/modules/nats.rs
```diff
@@ -0,0 +1,99 @@
...
+                toml::from_str(&format!(
+                    "
+                [nats]
+                format = \"[$symbol$name](bold purple)\"
+                symbol = \"\"
+                disabled = false
+                "
+                ))
+                .unwrap(),
+            )
```
davidkna, line 73:
```suggestion
            .config(toml::toml! {
                [nats]
                format = "[$symbol$name](bold purple)"
                symbol = ""
                disabled = false
            })
```

src/modules/nats.rs
```diff
@@ -0,0 +1,99 @@
...
+                toml::from_str(&format!(
+                    "
+                [nats]
+                format = \"[$symbol$name](bold red)\"
+                symbol = \"N \"
+                disabled = false
+                "
+                ))
+                .unwrap(),
+            )
```
davidkna, line 93:
```suggestion
            .config(toml::toml! {
                [nats]
                format = "[$symbol$name](bold red)"
                symbol = "N "
                disabled = false
            })
```

src/configs/nats.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub format: &'a str,
+    pub symbol: &'a str,
+    pub style: &'a str,
+    pub disabled: bool,
+}
+
+impl<'a> Default for NatsConfig<'a> {
+    fn default() -> Self {
+        NatsConfig {
+            format: "[$symbol$name]($style)",
```
davidkna, line 20: This should have more space, please also update the docs and config schema:
```suggestion
            format: "[$symbol($name )]($style)",
```

davidkna (commented):
I have made a few more suggestions, but this is already looking fairly good.

src/modules/nats.rs
hooksie1: I switched out the N since we have an icon.

hooksie1:
I'm not sure what the config is failing on. I ran `cargo run --features config-schema -- config-schema > .github/config-schema.json` and it's not showing any changes locally.

davidkna (approved):
LGTM

davidkna:
Thanks for the contribution @hooksie1!

PR #5908, opened 2024-04-05
fix(bash): fix handling of the preserved DEBUG trap (used in Bash <= 4.3)

### LINKED ISSUE #3749: Starship init appears to **execute** the existing bash DEBUG trap rather than capture it
#### Current Behavior
When setting the trap *before* `starship init` as documented in https://starship.rs/advanced-config/#custom-pre-prompt-and-pre-execution-commands-in-bash and having a `function set_win_title()` a number of errors appear.  I tried with the `functrace` settings as well and restarted my terminal app. 

What appears to happen is that the trap capture code executes the existing trap rather than capturing the command.

#### Expected Behavior

It should do as documented and capture the existing function beforehand.

* https://github.com/starship/starship/issues/1524
* [My solution to get the trap working to display the currently running process](https://stackoverflow.com/questions/71459823/how-to-change-the-terminal-title-to-currently-running-process/71467884#71467884).

#### Possible Solution
Per [my solution to get the trap working to display the currently running process](https://stackoverflow.com/questions/71459823/how-to-change-the-terminal-title-to-currently-running-process/71467884#71467884).  The key part is

```sh
trap "$(trap -p DEBUG |  awk -F"'" '{print $2}');set_win_title \${BASH_COMMAND}" DEBUG
```

But in starship's case the solution would be

```sh
# assuming I set
trap 'set_win_title \${BASH_COMMAND}' DEBUG

# Then
trap "starship_preexec \"$_\";$(trap -p DEBUG |  awk -F"'" '{print $2}')" DEBUG
```

However what would be better is if starship passed the command line into the handler.

#### Environment
- Starship version: 
```
starship 1.4.2
tag:v1.4.2
commit_hash:68d3f932
build_time:2022-03-10 18:16:43 +00:00
build_env:rustc 1.59.0 (9d1b2106e 2022-02-23),stable-x86_64-pc-windows-msvc
```
- Shell type: bash
- Shell plugin manager: [if present, e.g. oh-my-fish, oh-my-zsh, fisher, antigen]
- Operating system: Windows 11

```
# Your configuration here
```

#### Starship Configuration

```toml
# Your configuration here
```

akinomyoga:
### Description

> Describe your changes in detail

Three changes are included in this PR. Two of them are fixes to the handling of the DEBUG trap (used in Bash <= 4.3 for `starship_preexec`), and the other is a related variable name change. I include these changes in one PR here because they overlap with each other. However, if you would like separate PRs, I can happily separate them into four PRs and submit them one by one (because otherwise, they would produce conflicts).

- a46625ad This commit fixes the problem that the original `DEBUG` trap containing space is not correctly preserved. For example, when the user sets a custom DEBUG trap e.g. `trap 'echo hello' DEBUG` before Starship is initialized, Starship only preserves the first word, `echo`, for *the preserved DEBUG trap*. This is because the current code tries to extract the third word of the output of `trap -p DEBUG` (which will write `trap -- 'echo hello' DEBUG` to `stdout`) using `cut -d' ' -f3`. In this PR, we evaluate the output of `trap -p DEBUG` within the array assignment as `eval "DEBUG_TRAP=($(trap -p DEBUG))"` because `trap -p DEBUG` is produced in a format which can be evaluated for the shell execution. Then, we can pick the third element of the array.
- 13ed3563 This commit fixes the problem that the quote removal and other shell expansions in the preserved DEBUG trap string are not executed by Starship's DEBUG handler. The trap string is the general shell command that can contain shell expansions and quoting, so just running `$STARSHIP_DEBUG_TRAP` is not the right way to execute it. In this PR, we use `eval --
  "$STARSHIP_DEBUG_TRAP"`. The first option `--` is needed in case `$STARSHIP_DEBUG_TRAP` is `--`.
- 47286817 This is a refactoring commit. It is better to save the original DEBUG trap in `STARSHIP_DEBUG_TRAP` or another variable with its name starting with `_starship_` or `starship_` instead of putting them in a global variable named `dbg_trap`.

### Motivation and Context

> Why is this change required? What problem does it solve?

- a46625ad This change is required because, without this change, the current Starship integration does not correctly preserve the original DEBUG trap string containing spaces.  This PR solves the problem that only the first word in the trap string is recorded and executed by Starship's DEBUG trap handler.
- 13ed3563 This change is required because, without this change, shell expansions and quote removal that are expected to be executed for the original `DEBUG` trap string are not performed.  This PR solves the problem that processing of those expansions and quote removal is not performed.
- 47286817 This change is required because, without this change, the variable `dbg_trap` used to save the original `DEBUG` trap string may conflict with the user's shell variable or the variable set by other scripts.  This PR solves the conflicts by prefixing `STARSHIP_` to the variables used by Starship and separating their namespace.

> If it fixes an open issue, please link to the issue here.

Closes #3749 -- This fixes the issue reported in #3749 that the existing DEBUG trap string `"set_win_title \${BASH_COMMAND}"` is not correctly preserved.  This is partially fixed by PR #5020 and is finally fixed by commits cbb0da16 and 1adae699 provided by this PR.  On the other hand, the reported behavior that the existing DEBUG trap is executed is not an issue because the DEBUG trap is expected to be executed for any commands including the ones to initialize Starship.

### Screenshots (if appropriate):

For demonstration, let us consider the situation where a new Bash-4.3 session is started using the following `bashrc`:

```bash
# bashrc

trap 'echo "Hello World"' DEBUG
eval "$(starship.master init bash)"
```

#### In the current master

The result looks like this:

> 

We see that the messages `Hello World` are initially printed as expected. However, they stop at some point and start to produce empty lines, which is not expected.

#### After the first fix for the trap string containing spaces

The result becomes this:

> 

Now, we see that the string `Hello World` is preserved in the trap string, but extra quoting `""` is also printed.

#### After the second fix for the not-evaluated trap string

The result becomes this (which is expected):

> 

This is the expected behavior.

### How Has This Been Tested?

> Please describe in detail how you tested your changes.

### Checklist:

These are internal changes and fixes, so I think we do not need to update the documentation. If there would be places to update, please let me know. I'll happily update the necessary parts.

I checked `CONTRIBUTING.md`, but it seems to mention only the testing of the Rust codes. This patch only touches the Bash code and does not change the Rust code. There don't seem to be existing tests for the shell codes, so I didn't include tests. If the tests are needed, I'd appreciate it if you could point me to which place to look to prepare the tests for the shell integration.

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -65,8 +65,8 @@ starship_precmd() {
     # command pipeline, which may rely on it.
     _starship_set_return "$STARSHIP_CMD_STATUS"
 
-    if [[ -n "${_PRESERVED_PROMPT_COMMAND-}" ]]; then
-        eval "$_PRESERVED_PROMPT_COMMAND"
+    if [[ -n "${STARSHIP_PROMPT_COMMAND-}" ]]; then
+        eval "$STARSHIP_PROMPT_COMMAND"
     fi
 
     local -a ARGS=(--terminal-width="${COLUMNS}" --status="${STARSHIP_CMD_STATUS}" --pipestatus="${STARSHIP_PIPE_STATUS[*]}" --jobs="${NUM_JOBS}")
@@ -110,12 +110,12 @@ else
         # We want to avoid destroying an existing DEBUG hook. If we detect one, create
         # a new function that runs both the existing function AND our function, then
         # re-trap DEBUG to use this new function. This prevents a trap clobber.
-        dbg_trap="$(trap -p DEBUG | cut -d' ' -f3 | tr -d \')"
-        if [[ -z "$dbg_trap" ]]; then
+        STARSHIP_DEBUG_TRAP="$(trap -p DEBUG | cut -d' ' -f3 | tr -d \')"
+        if [[ -z "$STARSHIP_DEBUG_TRAP" ]]; then
             trap 'starship_preexec "$_"' DEBUG
-        elif [[ "$dbg_trap" != 'starship_preexec "$_"' && "$dbg_trap" != 'starship_preexec_all "$_"' ]]; then
+        elif [[ "$STARSHIP_DEBUG_TRAP" != 'starship_preexec "$_"' && "$STARSHIP_DEBUG_TRAP" != 'starship_preexec_all "$_"' ]]; then
             starship_preexec_all() {
-                local PREV_LAST_ARG=$1 ; $dbg_trap; starship_preexec; : "$PREV_LAST_ARG";
+                local PREV_LAST_ARG=$1 ; $STARSHIP_DEBUG_TRAP; starship_preexec; : "$PREV_LAST_ARG";
             }
             trap 'starship_preexec_all "$_"' DEBUG
         fi
@@ -130,7 +130,7 @@ else
         # Prepending to PROMPT_COMMAND breaks "command duration" module.
         # So, we are preserving the existing PROMPT_COMMAND
         # which will be executed later in the starship_precmd function
-        _PRESERVED_PROMPT_COMMAND="$PROMPT_COMMAND"
+        STARSHIP_PROMPT_COMMAND="$PROMPT_COMMAND"
         PROMPT_COMMAND="starship_precmd"
     fi
 fi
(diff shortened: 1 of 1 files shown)


davidkna (approved):
LGTM

PR #5910, opened 2024-04-07
feat(AzureRm) Adds a basic Azure Powershell module. 

theonlyames:
#### Description
Adds a alternative version of the Azure module for use with Azure PowerShell, as they store their context files separately using different formats. ~It still needs some tests to be written up for it.~

#### Motivation and Context
I use both the Azure PowerShell module and Azure CLI for my work, and the existing Azure module only supports Azure CLI. Originally I wanted to augment the existing module, I decided to make it a separate module instead as the context file structure is more complicated than what AzureCLI uses. 

image

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -38,6 +38,20 @@
         }
       ]
     },
+    "azurerm": {
+      "default": {
+        "disabled": false,
+        "format": "on [$symbol($subscription)]($style) ",
+        "style": "blue bold",
+        "subscription_aliases": {},
+        "symbol": "󰠅 "
+      },
+      "allOf": [
+        {
+          "$ref": "#/definitions/AzureRMConfig"
+        }
+      ]
+    },
     "battery": {
       "default": {
         "charging_symbol": "󰂄 ",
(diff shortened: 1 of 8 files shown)


theonlyames:
I am still writing tests for this, but am curious as to whether people think it is preferable to do this as a separate module or whether I should try to integrate the functionality directly in the `azure` module similar to how [Oh My Posh does things with its Az module's `source` property](https://ohmyposh.dev/docs/segments/az).

davidkna:
> I am still writing tests for this, but am curious as to whether people think it is preferable to do this as a separate module or whether I should try to integrate the functionality directly in the `azure` module similar to how [Oh My Posh does things with its Az module's `source` property](https://ohmyposh.dev/docs/segments/az).

I think it might be a good idea to combine these, but I am not familiar with the distinction between the two clients.

theonlyames:
> > I am still writing tests for this, but am curious as to whether people think it is preferable to do this as a separate module or whether I should try to integrate the functionality directly in the `azure` module similar to how [Oh My Posh does things with its Az module's `source` property](https://ohmyposh.dev/docs/segments/az).
> 
> I think it might be a good idea to combine these, but I am not familiar with the distinction between the two clients.

I have a WIP branch for that, but I feel like it might complicate tests and limit potential configurations more than I'd prefer.

PR #5952, opened 2024-05-01
perf(scala): use scala-cli first to avoid timeout

wahtique:
#### Description

- update Scala module to try first to retrieve the current version with `scala-cli` and fallback to current ( slower ) `scclac` 
- update `utils.rs` mock commands to call `scala-cli` and get back the current Scala version `3.4.1` before `scalac` which returns `... 2.13.5 ...` . Keep them distinct for clearer testing.

#### Motivation and Context

Partial fix to #5583

`scalac --version` has some noticeable ( most likely jvm ) startup time. Depending on the computer, this can cause a timeout. `scala-cli version --scala` usually has little to no startup time since it is either compiled to a native binary or use its own embedded GraalVM. It also outputs exactly and only the version number which avoids some string parsing.

image

This can only be considered a partial fix as `scala-cli` is not a hard requirement to work with Scala whereas `scalac` is.

image

#### Checklist:

### DIFF
--- a/src/modules/scala.rs
+++ b/src/modules/scala.rs
@@ -57,10 +57,17 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 }
 
 fn get_scala_version(context: &Context) -> Option<String> {
-    let command = context.exec_cmd("scalac", &["-version"])?;
-    let scala_version_string = get_command_string_output(command);
-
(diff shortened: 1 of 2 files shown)


davidkna (approved):
LGTM

andytom:
Thank you for your contribution @wahtique and thank you for reviewing @davidkna.

PR #5958, opened 2024-05-04
feat(container): Add more ways of finding container name when in docker.

Zshoham:
#### Description
When using distrobox with docker and starship I noticed that the container name is not displayed, instead only "Docker". 

This adds tow checks to find the actual container name: 
* Docker usually sets the HOSTNAME variable to the container name.
* Distrobox sets the CONTAINER_ID variable to the container name.

#### Checklist:

### DIFF
--- a/src/modules/container.rs
+++ b/src/modules/container.rs
@@ -66,7 +66,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
         if context_path(context, "/.dockerenv").exists() {
             // docker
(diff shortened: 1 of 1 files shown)


Zergoma:
Thx for the fix, I hope it will be merge soon !

klmcwhirter:
@Zshoham is this PR going to be resolved? I need the CONTAINER_ID env var support (distrobox).

Or is there anyone from the project who can weigh in? I do not see any comments in the diff of his changes.

PR #5962, opened 2024-05-05
Fixed Winget Installation Error

Esai-Keshav:
The Winget installation command is missing the source flag I have added that 

#### Description
The installation command is causing the error
`winget install --id Starship.Starship`

```cmd

C:\Users\user>winget install --id Starship.Starship
Failed when searching source: msstore
An unexpected error occurred while executing the command:
WinHttpSendRequest: 12030: The connection with the server was terminated abnormally


The following packages were found among the working sources.
Please specify one of them using the --source option to proceed.
Name     Id                Source
---------------------------------
starship Starship.Starship winget

```

#### Motivation and Context
When I tried to install using the winget command it caused source not defined error. If the source is defined it will easy for installation. 

#### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ Install the latest version for your system with the MSI-installers from the [rel
 Install Starship using any of the following package managers:
 
 | Repository      | Instructions                                            |
-| --------------- | --------------------------------------------------------|
+| --------------- | ------------------------------------------------------- |
 | **[crates.io]** | `cargo install starship --locked`                       |
 | [Chocolatey]    | `choco install starship`                                |
 | [conda-forge]   | `conda install -c conda-forge starship`                 |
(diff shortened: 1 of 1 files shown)


davidkna:
Did you run some sort of utility to slim down Windows that could have uninstalled the MS store?

Esai-Keshav:
The installation the source is `winget` not Ms store

davidkna:
It just seems that rather than a general issue, this could be caused by issues with your system's configuration (even if winget is perhaps acting overzealously here).

Esai-Keshav:
> It just seems that rather than a general issue, this could be caused by issues with your system's configuration (even if winget is perhaps acting overzealously here).

While installing using winget it is compulsory to define the source in the new update of winget

davidkna:
There does not seem to be the case on my system, both with the current stable and preview releases. I assume there is some issue around the Window Store on your system.

PR #5987, opened 2024-05-21
fix(dotnet): Remove duplicate `v` in dotnet version

berndverst:
Every language version provider returns the version without a leading 'v', and the shared code then prepends the 'v'. Take a look at Python for example.

The Dotnet provider prepends a 'v' however, causing the version to be printed like 'vv8.0.203'.

This happens for example with this configuration

```
[dotnet]
symbol = '🥅 '
style = 'bg:color_blue'
format = '[[$symbol($version)](fg:color_fg0 bg:color_blue)]($style)'
heuristic = false
```

Where simply `$dotnet\` would be used somewhere in the prompt. Then navigate to a folder contain a dotnet project.

#### Description
Removes the string formatting which prepends the 'v' character.

#### Motivation and Context
Removes duplicate `v` in dotnet version string.

image

#### Checklist:

### DIFF
--- a/src/modules/dotnet.rs
+++ b/src/modules/dotnet.rs
@@ -294,7 +294,7 @@ fn map_str_to_lower(value: Option<&OsStr>) -> Option<String> {
 
 fn get_version_from_cli(context: &Context) -> Option<String> {
     let version_output = context.exec_cmd("dotnet", &["--version"])?;
-    Some(format!("v{}", version_output.stdout.trim()))
+    Some(version_output.stdout.trim())
 }
 
 fn get_latest_sdk_from_cli(context: &Context) -> Option<String> {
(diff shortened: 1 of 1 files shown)


src/modules/dotnet.rs
```diff
@@ -294,7 +294,7 @@ fn map_str_to_lower(value: Option<&OsStr>) -> Option<String> {
 
 fn get_version_from_cli(context: &Context) -> Option<String> {
     let version_output = context.exec_cmd("dotnet", &["--version"])?;
-    Some(format!("v{}", version_output.stdout.trim()))
+    Some(version_output.stdout.trim())
```
berndverst, line 297:
```suggestion
    Some(version_output.stdout.trim().to_string())
```

davidkna (commented):
Please add a test for this.

PR #5989, opened 2024-05-22
fix: improve non-msvc windows support by bumping stable dependencies

### LINKED ISSUE #5984: Bump gitoxide version to re-enable windows-gnu builds

Currently gitoxide, and thus all dependents such as starship, cannot be built with `*-windows-gnu` or any custom non-MSVC toolchains. This is due to a mismatch in how `gitoxide-features` tracks the safety of enabling the `sha1/asm` feature (MSVC checking only) and a more recent change in the corresponding ASM implementation due to results being incorrect on Windows (target OS checking). RustCrypto/asm-hashes is working on fixing this by dropping the whole ASM feature and instead making the ASM inlined (and I assume feature and/or platform gated?). Byron/gitoxide#917 is currently tracking this problem on their end and I proposed a one line fix for the issue that I assume will land in the next minor bump.

Once the linked issue is closed and a new version bump has been done, pushing a patch update for starship should allow it to be built on any Windows toolchain again without issue. Should only be a one line fix unless gitoxide has a breaking change.

I would not want to implement this fix, for the same reason I wouldn't want to PR a fix to gitoxide. A single line commit as sole reason for being a contributor is laughable at best, I'd rather not.

None whatsoever. I'm currently building updates for starship with a manual toolchain override which special cases it as the only package in my cargo/bin folder. I don't mind doing it for such a nice and useful utility, but I'd love to see it build with the other packages when I run `cargo install-update`

RivenSkaye:
#### Description
Updates `gix` and `gix-features` in order to re-enable building on all windows targets using a toolchain other than `*-windows-msvc`.
Updates `shadow-rs` to latest stable.
Tests have been run locally, the only failure relates to the global UTF-8 support flag in Windows, which I have off due to breakage in some legacy work stuff. Turning it on makes the test pass, turning it off makes the test fail on the second assert, where all of the wide characters get turned into question marks. Please run CI to confirm everything still works as intended just in case.

#### Motivation and Context
This change is required to fix building of direct dependencies `gix` and `gix-features` for the Rust tier 1 targets `i686-pc-windows-gnu` and `x86_64-pc-windows-gnu`, as well as several tier 2 and 3 `*-windows-gnullvm` targets and any custom toolchains that do not target `*-windows-msvc` explicitly. This means it also has a direct impact on the ability to build Starship itself for these targets.
The `shadow-rs` bump is included because `cargo update` mentioned it explicitly and checking the changelog, it only fixes building for projects that globally set `deny(missing_docs)`.
Neither are breaking changes, and both are limited to `Cargo.{toml,lock}`, as such I figured they're the same type of non-breaking change.
Closes #5984 

  - [x] x64 `msvc`
  - [x] x64 `gnu`
  - [ ] i686 `msvc`
  - [ ] i686 `gnu`

#### Checklist:

I don't think either of these apply? If they do, please let me know and I'll modify what needs to be modified.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,18 +22,18 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.16"
+version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0942ffc6dcaadf03badf6e6a2d0228460359d5e34b57ccdc720b7382dfbd5ec5"
+checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
 name = "android-tzdata"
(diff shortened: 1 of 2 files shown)


davidkna (approved):
LGTM

davidkna:
Thanks for the fix @RivenSkaye!

RivenSkaye:
Would a follow-up PR to add at least `windows-gnu` to the CI compilation checks be welcome? If so, I'll whip one up tonight

davidkna:
I don't particularly care about the `windows-gnu` targets, but adding it to CI should be fine (if possible, I would prefer only testing with the stable toolchain).

RivenSkaye:
Considering `*-windows-gnu` is tier 1, expect a PR soon to add tests on stable then

PR #5991, opened 2024-05-22
feat(hostname): add option for aliases

ekorchmar:
#### Description

* Extend HostnameConfig to include a table to store key-value pair for aliases
* Read the field from starship.toml; default to an empty table 
* Replace hostname with an alias if match is found in the table
* Includes a test case

#### Motivation and Context

Adds an option for users to setup shorthand aliases for their usual daily driver machines.

#### Checklist:


ekorchmar:
https://github.com/starship/starship/pull/6097

PR #5995, opened 2024-05-23
docs: fix copy-paste typo in config docs

tchajed:
The symbol config line for Gleam was copied from Go, for Haxe was copied from Helm, and for Typst was copied from Daml.

#### Description
The lines describing `symbol` config value for Gleam and Haxe were incorrect; they referenced Go and Helm respectively, likely a copy-paste error from those languages.

#### Motivation and Context

Fix the documentation to refer to the right language.

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**
Documentation-only change, no tests

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #6001, opened 2024-05-30
Update README.md

kilna:
Fix Helm->Haxe

#### Description
Helm was used instead of Haxe in this section, probably a copy/paste/search/replace error.


kilna:
Looks like this is a dupe of #5995 , shoulda checked first!

PR #6025, opened 2024-06-08
Docs: Updated gleam documentation to reference gleam in symbols section instead of go

matthewrobinsondev:
In the gleam documentation it was referencing go under the symbol instead of gleam. Missed during copy & paste I imagine.

#### Description
Updates all the language files that was using the word go in the gleam section to reference the symbol correctly.

#### Motivation and Context

Closes #

Documentation-only change, no tests

#### Checklist:


davidkna:
Please only update the English docs to avoid conflicts with Crowdin. The translated docs will be updated automatically.

matthewrobinsondev:
> Please only update the English docs to avoid conflicts with Crowdin. The translated docs will be updated automatically.

No problem, will fix

matthewrobinsondev:
I am slightly confused though because `docs/config/README.md` appear to be correct but the translations all seem wrong if thats where they are generated from?

davidkna:
It looks like we already fixed this. The fixed translated docs are already in `i18n_master` (#5992).

PR #6032, opened 2024-06-11
perf(k8s): Improve performance of kubeconfig module

alvaroaleman:
This module currently takes about 200 ms when using our ~10MiB kubeconfig. This change improves its performance by:
* Only parsing the file once: Reduces time to around 130 ms
* (Naively) checking if the content is yaml or json and potentially parse as the latter, as that seems to be much faster, reducing the runtime to ~30ms

All timings with warm page cache.

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/modules/kubernetes.rs
+++ b/src/modules/kubernetes.rs
@@ -1,8 +1,8 @@
-use yaml_rust2::YamlLoader;
+use serde_json::Value as JsonValue;
+use yaml_rust2::{Yaml, YamlLoader};
 
 use std::borrow::Cow;
 use std::env;
-use std::path;
 
 use super::{Context, Module, ModuleConfig};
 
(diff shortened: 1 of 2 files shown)


alvaroaleman:
> Main workflow / Check if config schema is up to date (pull_request) Failing after 33s
How can this be fixed?

src/modules/kubernetes.rs
```diff
@@ -97,6 +112,11 @@ fn get_aliased_name<'a>(
     }
 }
 
+enum JsonOrYaml {
+    Json(JsonValue),
+    Yaml(Yaml),
+}
```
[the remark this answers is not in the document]
alvaroaleman: Done

src/modules/kubernetes.rs
```diff
@@ -17,50 +17,65 @@ struct KubeCtxComponents {
...
-        .find(|(_, name)| name == &current_ctx_name)?;
-
-    let ctx_components = KubeCtxComponents {
-        user: ctx_yaml["context"]["user"]
-            .as_str()
-            .filter(|s| !s.is_empty())
-            .map(String::from),
-        namespace: ctx_yaml["context"]["namespace"]
-            .as_str()
+fn get_current_kube_context_name(document: &JsonOrYaml) -> Option<String> {
```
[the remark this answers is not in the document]
alvaroaleman: Done

src/modules/kubernetes.rs
```diff
@@ -17,50 +17,65 @@ struct KubeCtxComponents {
...
             .map(String::from),
-    };
+    }
+}
 
-    Some(ctx_components)
+fn get_kube_ctx_components(
+    document: &JsonOrYaml,
+    current_ctx_name: &str,
+) -> Option<KubeCtxComponents> {
```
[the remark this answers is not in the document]
alvaroaleman: I don't think the tiny, likely not even measurable performance improvement is worth the added complexity

src/modules/kubernetes.rs
```diff
@@ -97,6 +108,12 @@ fn get_aliased_name<'a>(
     }
 }
 
+#[derive(Debug)]
+enum Document {
```
jankatins, line 112: I'm not really happy that this ends up in two different codepath everywhere, just to get a bit more speed in an edge case (That's at least how I perceive this: how often do you have such big kube configs?) :-(

Personally, I would also have expected to get a common data structure from both parser.

If this is acceptable, then I would expect an explanation in the code why this was chose, so one does not have to look up the original commits to understand why two different parsers are used, either here where the data structure is defined or in the parser function.

davidkna: Sadly, the yaml-rust2 crate doesn't support serde yet and serde-yaml is unmaintained, but [there are plans](https://www.github.com/saphyr-rs/saphyr/issues/1) for this at the more actively developed branch of yaml-rust2.

Nevertheless, it might be possible to reduce duplication a bit with `macro_rules!` at least.

alvaroaleman: I've tried to make this better by using generics to not need to duplicate the logic that finds the relevant fields for the yaml and json case and also left a comment as to why to have the two codepaths.

I do agree kubeconfigs of this size are unusual but it is not "a bit" more speed, its about 77% faster so this makes a really huge difference once you do have such a kubeconfig.

src/modules/kubernetes.rs
```diff
@@ -239,6 +280,26 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+fn parse_kubeconfigs<I>(raw_kubeconfigs: I) -> Vec<Document>
+where
+    I: Iterator<Item = Option<String>>,
+{
+    raw_kubeconfigs
+        .filter_map(|content| match content {
+            Some(value) => match value.chars().next() {
+                // Parsing as json is about an order of magnitude faster than parsing
+                // as yaml, so do that if possible.
+                Some('{') => serde_json::from_str(&value).ok().map(Document::Json),
```
davidkna, line 292: I would prefer to fall back to `YamlLoader::load_from_str` if `serde_json::from_str` fails, even if the first character is `{`.

alvaroaleman: Added a fallback and test, ptal

alvaroaleman:
@davidkna gentle ping, any chance you could give this another look, please? Really appreciate it!

davidkna (approved):
LGTM

alvaroaleman:
@davidkna what is the process to get a change like this actually merged?

andytom:
Thank you for your contribution @alvaroaleman and thanks for reviewing @davidkna.

PR #6039, opened 2024-06-17
feat(gcloud): Disabled gcloud module when active config hasn't been set

### LINKED ISSUE #3658: Feature Request: Allow gcloud to be hidden when the active config is NONE

jpiggg:
This change will disable Google Cloud module when there is no active config set

#### Motivation and Context
It has been suggested to implement there https://github.com/starship/starship/issues/3658

This PR is based on those:
closes #4092
closes #3692
closes #3658

#### Checklist:

### DIFF
--- a/src/modules/gcloud.rs
+++ b/src/modules/gcloud.rs
@@ -467,6 +467,16 @@ project = very-long-project-name
     }
 
     #[test]
+    fn no_active_config() {
+        let actual = ModuleRenderer::new("gcloud")
+            .env("CLOUDSDK_ACTIVE_CONFIG_NAME", "NONE")
+            .collect();
+        let expected = None;
+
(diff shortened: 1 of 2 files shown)


src/modules/gcloud.rs
```diff
@@ -89,6 +89,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     }
 
     let (config_name, config_path) = get_current_config(context)?;
+
+    if config_name == "NONE" {
+        return None;
+    }
```
davidkna, line 95: I think I would prefer putting this into `get_current_config`, but this is fine too.

davidkna (approved):
LGTM

andytom:
Thank you for your contribution @jpiggg and thanks for the review @davidkna.

PR #6060, opened 2024-06-30
fix(dotnet): Remove duplicate v in dotnet version 

davidkna:
This PR matches #5987 (rebased) with a test added.

Closes #5987

> Every language version provider returns the version without a leading 'v', and the shared code then prepends the 'v'. Take a look at Python for example.
> 
> The Dotnet provider prepends a 'v' however, causing the version to be printed like 'vv8.0.203'.
> 
> This happens for example with this configuration
> 
> ```
> [dotnet]
> symbol = '🥅 '
> style = 'bg:color_blue'
> format = '[[$symbol($version)](fg:color_fg0 bg:color_blue)]($style)'
> heuristic = false
> ```
> 
> Where simply `$dotnet\` would be used somewhere in the prompt. Then navigate to a folder contain a dotnet project.
> 
> 
> 
> 
> 
> #### Description
> Removes the string formatting which prepends the 'v' character.
> 
> #### Motivation and Context
> Removes duplicate `v` in dotnet version string.
> 
> #### Screenshots (if appropriate):
> 
> image

#### Checklist:

### DIFF
--- a/src/modules/dotnet.rs
+++ b/src/modules/dotnet.rs
@@ -356,6 +356,7 @@ mod tests {
     use std::fs::{self, OpenOptions};
     use std::io::{self, Write};
     use tempfile::{self, TempDir};
+    use utils::{write_file, CommandOutput};
 
     #[test]
     fn shows_nothing_in_directory_with_zero_relevant_files() -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


andytom (approved):

davidkna:

berndverst:
Thanks for fixing. I'm not a rust dev (I'm a Go / Python / C# dev) and I didn't have time to ramp up on learning more rust to write the test unfortunately :)

PR #6066, opened 2024-07-03
fix: plain-text-symbols preset to not show 🌐 emoji when using SSH

thebigdalt:
#### Description

Added the following to keep the globe (🌐) emoji from showing when remoting into a machine that uses starship:

```toml
[hostname]
ssh_symbol = ""
```

Since no other emojis are present in this preset and the default format string seems descript enough, I figured this can just be an empty string.  

#### Motivation and Context

Closes #

The globe 🌐 emoji looks out of place when using the plain-text-symbols preset since every other symbol is character-based or ASCII. It kind of goes against "plain text".

#### Checklist:

I don't think these are applicable.

### DIFF
--- a/docs/public/presets/toml/plain-text-symbols.toml
+++ b/docs/public/presets/toml/plain-text-symbols.toml
@@ -88,6 +88,9 @@ symbol = "guix "
 [hg_branch]
 symbol = "hg "
 
+[hostname]
+ssh_symbol = ""
+
 [java]
 symbol = "java "
 


docs/public/presets/toml/plain-text-symbols.toml
```diff
@@ -88,6 +88,9 @@ symbol = "guix "
 [hg_branch]
 symbol = "hg "
 
+[hostname]
+ssh_symbol = ""
```
davidkna, line 92: I would prefer adding a suitable ASCII replacement instead.

PR #6084, opened 2024-07-12
fix(docs): correct username options

mrgum:
These seem to have been copied from hostname and did not make sense

#### Description
These username config options seem to have been copied from hostname and did not make sense

#### Motivation and Context
Make the docs less confusing

its a doc change

#### Checklist:
it is a doc change and there are no tests


andytom (approved):

davidkna:

PR #6094, opened 2024-07-17
Remove `🌐` symbol in plain text symbols preset

yvan-sraka:
#### Description

Before:
```
yvan in 🌐 X270 in guix-config on git master [x!?>] took 9s
```

After:
```
yvan in X270 in guix-config on git master [x!?>] took 9s
```

#### Motivation and Context

From https://starship.rs/presets/plain-text:

> This preset changes the symbols for each module into plain text. Great if you don't have access to Unicode.

So using emojis such as `🌐` should be avoided :)

#### Checklist:

### DIFF
--- a/docs/public/presets/toml/plain-text-symbols.toml
+++ b/docs/public/presets/toml/plain-text-symbols.toml
@@ -232,3 +232,6 @@ symbol = "terraform "
 
 [zig]
 symbol = "zig "
+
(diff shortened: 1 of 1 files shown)


yvan-sraka:
Oops, didn't notice that https://github.com/starship/starship/pull/6066 was already opened, closing as duplicated ;)

PR #6098, opened 2024-07-18
docs(configuration): 'symbol' as Option and Variable should not be translated

nanianlisao:
#### Description
`symbol` as Option and Variable should not be translated

#### Motivation and Context

Closes #

#### Checklist:


nanianlisao:
Sorry I don't know how to use crowdin

PR #6107, opened 2024-07-20
fix(fish): improve fish transient prompt

st1020:
#### Description

This PR fixes the problem that the transient prompt in fish does not work when pressing Enter without entering a command.
In addition, it also fixes the problem that the transient prompt is incorrectly triggered when pressing Enter to confirm an entry in Tab Completion.

#### Motivation and Context

Improve transient prompt in fish.

Before:
old

After:
new

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -19,6 +19,7 @@ function fish_prompt
         else
             printf "\e[1;32m❯\e[0m "
         end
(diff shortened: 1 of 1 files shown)


andytom:
Thank you for your contribution @st1020 and thank you for reviewing @davidkna.

PR #6108, opened 2024-07-20
feat(mojo): add module for Mojo language

DanielKneipp:
Add module to support [Mojo programming language](https://www.modular.com/mojo)

#### Description
This new module should show the Mojo version installed and its commit hash optionally whenever a Mojo file is identified.

#### Motivation and Context
I started playing with this language and saw the opportunity to add it to Starship.

Without commit hash version:
Screenshot 2024-07-20 at 15 46 06

With commit hash version:
Screenshot 2024-07-20 at 15 46 44

Executed `cargo test mojo` and added a mock to the command `mojo --version` as well

#### Checklist:


src/modules/mojo.rs
```diff
@@ -0,0 +1,155 @@
...
+                    let mojo_version_output = context.exec_cmd("mojo", &["--version"])?.stdout;
+                    let version_items = mojo_version_output.split(' ').collect::<Vec<&str>>();
+
+                    if version_items.len() <= 1 {
+                        return None;
+                    }
+
+                    if version_items.len() == 2 || !config.show_commit {
+                        return Some(Ok(version_items[1].trim().to_string()));
+                    }
```
davidkna, line 43: Instead of manual indexing, I would prefer something like this:
```
                    let (version, hash) = match version_items[..] {
                        [_, version] => (version.trim(), None),
                        [_, version, hash, ..] => (version, Some(hash)),
                        _ => {
                            log::debug!("Unexpected `mojo --version` output: {mojo_version_output}");
                            return None;
                        }
                    }
```

src/modules/mojo.rs
```diff
@@ -0,0 +1,155 @@
...
+
+                    let mojo_version = version_items[1].trim();
+                    let mojo_hash = version_items[2].trim();
+
+                    Some(Ok(format!("{} {}", mojo_version, mojo_hash)))
```
davidkna, line 48: I think it might be better to have separate `$version` and `$hash` variables. To cache the output/parsed version you can use `once_cell::sync::Lazy`.

davidkna (changes requested):

src/modules/mojo.rs
DanielKneipp: Hey, thanks a lot for the suggestions. I implemented the hash as a variable and used the match pattern. Much better indeed!

However, I'm struggling to understand how `Lazy` will be beneficial here. Since the module function will need to be executed anyway when `parsed` is used, wouldn't the function inside `Lazy` be called the same amount of times with or without `Lazy`? I see only a delay in the function call

davidkna: The version is not necessarily included in the format string, and in that case getting the version can be skipped altogether.

DanielKneipp: Oh well, I wouldn't expect someone to use this without showing the version. But since it's a possibility, I believe I have it implemented now 👍.

davidkna (approved):
LGTM

andytom:
Thank you for your contribution @DanielKneipp and thank you for reviewing @davidkna.

PR #6109, opened 2024-07-20
feat(fish): support async prompt

st1020:
#### Description

This PR makes prompt load asynchronously in fish.

The idea comes from [fish-async-prompt](https://github.com/acomagu/fish-async-prompt), which basically runs starship in a new fish process and notifies fish to update prompt by sending a signal after loading is complete.

This makes starship very fast in fish. The difference is noticeable, especially in large git repositories. With async prompt enabled, the delay is almost unnoticeable.

#### Motivation and Context

Try to make prompt load asynchronously in fish. 

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -1,31 +1,54 @@
 function fish_prompt
-    switch "$fish_key_bindings"
-        case fish_hybrid_key_bindings fish_vi_key_bindings
-            set STARSHIP_KEYMAP "$fish_bind_mode"
-        case '*'
-            set STARSHIP_KEYMAP insert
-    end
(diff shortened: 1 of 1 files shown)


davidkna:
Are there any notable differences to [fish-async-prompt](https://github.com/acomagu/fish-async-prompt)? This should probably be optional. On the technical side, I'm not sure if I like the use of temporary files for updating the prompt (injecting strings there could lead to a security issue), and I would prefer doing the signal handling in rust.

st1020:
@davidkna Yes, you are right, this is no notable differences to [fish-async-prompt](https://github.com/acomagu/fish-async-prompt) except starship integration. Doing the signal handling in rust is better, but I'm not sure if we need to do something special for fish only.

PR #6145, opened 2024-07-29
fix(python): improve parsing of `pyvenv.cfg` files

### LINKED ISSUE #6125: Error parsing python virtual environment on Windows if path name startswith `n`
Hi! Thank you for your awesome work and I have an issue here.

Since there are so many issues here, I can't be sure that a similar issue still exists after I search for it. So forgive me if my issue is a duplicate issue and feel free please close it, and of course let me know how to track this down :)

#### Current Behavior
```console
acmbot on  main [?] is 📦 v1.0.0 via  v21.6.1 via 🐍 v3.12.2 (C:Users
octiDesktopWorkspace
onebot-plugin-swpuacm) took 14s
```

It seems that when there's a directory named startswith a literal `n`, starship will consider this `\n` to a line break on Windows systems. There's something wrong with the path parsing.

#### Expected Behavior
```console
acmbot on  main [?] is 📦 v1.0.0 via  v21.6.1 via 🐍 v3.12.2 (C:Users\noctis\DesktopWorkspace\nonebot-plugin-swpuacm) took 14s
```

VSCode Terminal:

QQ_1722004018049

#### Possible Solution
Fix this parser to escape `\n` to avoid unexpected next line.

#### Environment
- Starship version: 1.19.0
- pwsh version: PowerShell 7.4.3
- Git Commit Hash: 
- Branch/Tag: master
- Rust Version: rustc 1.78.0 (9b00956e5 2024-04-29)
- Build Time: 2024-05-15 17:12:02 +00:00

```pwsh
```

#### Starship Configuration

No more configuration, I was using default config.

fu050409:
#### Description

Some python virtual environments have `prompt` set to a path, which in Windows would include the `\` symbol here. We will accidentally escape something like `\n` to a line break here, and similarly symbols like `\r` will be converted. I got around this by rewriting `\` to `\\` before parsing the `pyvenv.cfg` file.

#### Motivation and Context

Closes #6125

#### Checklist:

### DIFF
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -1,5 +1,4 @@
 use ini::Ini;
-use std::io::Read;
 use std::path::Path;
 
 use super::{Context, Module, ModuleConfig};
@@ -124,19 +123,11 @@ fn get_python_virtual_env(context: &Context) -> Option<String> {
 }
 
 fn get_prompt_from_venv(venv_path: &Path) -> Option<String> {
-    let mut buf = String::new();
-    if std::fs::File::open(venv_path.join("pyvenv.cfg"))
-        .ok()?
-        .read_to_string(&mut buf)
-        .is_err()
-    {
-        return None;
-    }
-    Ini::load_from_str(buf.replace("\\", "/").as_str())
+    Ini::load_from_file_noescape(venv_path.join("pyvenv.cfg"))
         .ok()?
         .general_section()
         .get("prompt")
-        .map(|prompt| String::from(prompt.trim_matches(&['(', ')'] as &[_])))
+        .map(|prompt| String::from(prompt.replace("\\", "/").trim_matches(&['(', ')'] as &[_])))
 }
 
 #[cfg(test)]
--- a/src/modules/python.rs
+++ b/src/modules/python.rs
@@ -1,4 +1,5 @@
 use ini::Ini;
+use std::io::Read;
 use std::path::Path;
 
 use super::{Context, Module, ModuleConfig};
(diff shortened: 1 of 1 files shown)


davidkna:
Perhaps just use [`load_from_file_noescape`](https://docs.rs/rust-ini/0.21.0/ini/struct.Ini.html#method.load_from_file_noescape) instead.

fu050409:
> Perhaps just use [`load_from_file_noescape`](https://docs.rs/rust-ini/0.21.0/ini/struct.Ini.html#method.load_from_file_noescape) instead.

Thanks for your review and I have made changes according to your suggestions 😊

src/modules/python.rs
```diff
@@ -121,12 +121,13 @@ fn get_python_virtual_env(context: &Context) -> Option<String> {
...
 }
+
 fn get_prompt_from_venv(venv_path: &Path) -> Option<String> {
-    Ini::load_from_file(venv_path.join("pyvenv.cfg"))
+    Ini::load_from_file_noescape(venv_path.join("pyvenv.cfg"))
         .ok()?
         .general_section()
         .get("prompt")
-        .map(|prompt| String::from(prompt.trim_matches(&['(', ')'] as &[_])))
+        .map(|prompt| String::from(prompt.replace("\\", "/").trim_matches(&['(', ')'] as &[_])))
```
davidkna, line 130: This should not be necessary with `load_from_file_noescape`
```suggestion
        .map(|prompt| String::from(prompt.trim_matches(&['(', ')'] as &[_])))
```

src/modules/python.rs
```diff
@@ -424,6 +425,33 @@ prompt = '(foo)'
...
+        venv_cfg.sync_all()?;
+
+        let actual = ModuleRenderer::new("python")
+            .path(dir.path())
+            .env("VIRTUAL_ENV", dir.path().join("my_venv").to_str().unwrap())
+            .collect();
+
+        let expected = Some(format!(
+            "via {}",
+            Color::Yellow.bold().paint("🐍 v3.8.0 (foo/nbar) ")
```
davidkna, line 448:
```suggestion
            Color::Yellow.bold().paint(r"🐍 v3.8.0 (foo\nbar) ")
```

src/modules/python.rs
fu050409: Yes, you're right. At first, I wanted to convert the prompt to a posix style, but that doesn't seem necessary now.

davidkna (approved):

davidkna:
Thanks for the fix @fu050409!

PR #6154, opened 2024-07-31
feat(directory): Fallback to original string when contract failed

soraxas:
#### Description

There was a bug where `[directory]` produced an empty directory string when `contract_repo_path` failed.

This changes fixes it. The changes involves avoiding the `?` at: (current behaviour will made it returns none, i.e. empty string)
```rs
let contracted_path = contract_repo_path(display_dir, repo_root)?;
```
The original approach also has multiple code path that leads to the default
```rs
[String::new(), String::new(), prefix + dir_string.as_str()]
```
and this implementation centralises it to only provides default value at the very end.

#### Motivation and Context

**Minimum reproducible config:**
```toml
[directory]
repo_root_style = ""
truncate_to_repo = true
```
Currently, `[directory]` can shorten path to repo root. When user is using custom `repo_root_style` style, and when the path contract fails (e.g. using a symlink to a subdirectory of a repo), the current behaviour produces an empty "shorten" string.

Creating a test scenario:
```sh
# make an empty test repo
$ mkdir my_repo && cd my_repo && git init && mkdir subdir && cd ..
# link to repo root
$ ln -s my_repo link_to_repo     
# link to repo root subdir
$ ln -s my_repo/subdir link_to_repo_subdir    
```
Current behaviour:
```sh
$ cd link_to_repo
$ starship prompt

link_to_repo on  master         # <=== This is correct
$ cd ../link_to_repo_subdir
$ starship prompt

on  master                      # <=== This is an EMPTY "shorten" path
```

New behaviour (falling back to un-shorten path, instead of unwrapping empty option):
```sh
$ cd ../link_to_repo_subdir

git-repos/other_folders/link_to_repo_subdir on  master         # <=== This is the original un-shorten path
```

#### Checklist:

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -103,25 +103,37 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         String::new()
     };
 
-    let path_vec = match &repo.and_then(|r| r.workdir.as_ref()) {
-        Some(repo_root) if config.repo_root_style.is_some() => {
-            let contracted_path = contract_repo_path(display_dir, repo_root)?;
-            let repo_path_vec: Vec<&str> = contracted_path.split('/').collect();
-            let after_repo_root = contracted_path.replacen(repo_path_vec[0], "", 1);
-            let num_segments_after_root = after_repo_root.split('/').count();
+    // the usage of 2 takes in the following is to avoid borrowing issues. They would never overlaps, so the ? are safe
+    let mut maybe_prefix = Some(prefix);
+    let path_vec = repo
+        .and_then(|r| r.workdir.as_ref())
+        .and_then(|repo_root| {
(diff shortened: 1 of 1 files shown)


davidkna:
Please ensure that the tests still pass with your changes.

cappyzawa:
@soraxas Sorry for the delayed review. I verified this locally - all directory module tests pass and the symlink issue is fixed. Could you rebase on master to trigger CI?

soraxas:
it's good that clippy keeps getting better with new rules getting added after a year.

@cappyzawa I dont understand what's the [Check if config schema is up to date](https://github.com/starship/starship/actions/runs/21192067364/job/60960381894?pr=6154#logs) is about.

soraxas:
@cappyzawa can we get a review to merge this? Thanks

PR #6164, opened 2024-08-03
feat: ignore configured commands for cmd_duration notify

### LINKED ISSUE #1933: Don't show cmd_duration notifications for certain commands

I'd love to automatically get notifications after long-running tasks, but it's frustrating when every time I exit vim I get a popup.

I'd like to be able to specify a set of commands that would be ignored for the purposes of `cmd_duration.show_notifications`, e.g.

```toml
[cmd_duration]
show_nofitications = true
ignore_commands = ['vim', 'sudoedit']
```

Drawbacks:
- configuration gets more complicated
- might not actually be possible for starship to tell what command just ran
- what should happen when the command isn't a single simple command, but e.g. a pipeline?

Don't make the change, in which case I keep doing <kbd>^Z</kbd>, `fg; notify-send done` by hand every time a command starts taking longer than I anticipated

samuelallan72:
WIP!  I've implemented this because I was wanting this feature for a while, then I discovered it was already a popular request (#1933), so I attempted to implement it.  No tests or documentation updates yet, as the api is likely to change as this is reviewed and discussed.

#### Description

Add a `notify_ignore_commands` option to the `cmd_duration` module, and configuration for nushell support.

For example, this provides support for enabling notifications for long running commands, while ignoring a configurable list of long running commands:

```
[cmd_duration]
show_notifications = true
min_time_to_notify = 5_000
notify_ignore_commands = ["vim", "mpv", "mutt"]
```

This is useful to avoid many false positives for this notification when exiting editors, etc.

I use nushell, so have included an initial nushell configuration for getting the last run command.  This may not be possible for all shells.

#### Motivation and Context

Closes #1933

#### Checklist:

### DIFF
--- a/src/configs/cmd_duration.rs
+++ b/src/configs/cmd_duration.rs
@@ -15,6 +15,7 @@ pub struct CmdDurationConfig<'a> {
     pub disabled: bool,
     pub show_notifications: bool,
     pub min_time_to_notify: i64,
+    pub notify_ignore_commands: Vec<&'a str>,
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub notification_timeout: Option<u32>,
@@ -31,6 +32,7 @@ impl<'a> Default for CmdDurationConfig<'a> {
             show_notifications: false,
             min_time_to_notify: 45_000,
             notification_timeout: None,
+            notify_ignore_commands: vec![],
         }
     }
 }
--- a/src/init/starship.nu
+++ b/src/init/starship.nu
@@ -2,37 +2,53 @@
 # - overlay which can be loaded with `overlay use starship.nu`
 # - module which can be used with `use starship.nu`
 # - script which can be used with `source starship.nu`
-export-env { $env.STARSHIP_SHELL = "nu"; load-env {
-    STARSHIP_SESSION_KEY: (random chars -l 16)
-    PROMPT_MULTILINE_INDICATOR: (
-        ^::STARSHIP:: prompt --continuation
-    )
+export-env {
+    $env.STARSHIP_SHELL = "nu"
 
-    # Does not play well with default character module.
-    # TODO: Also Use starship vi mode indicators?
-    PROMPT_INDICATOR: ""
-
-    PROMPT_COMMAND: {||
-        # jobs are not supported
-        (
-            ^::STARSHIP:: prompt
-                --cmd-duration $env.CMD_DURATION_MS
-                $"--status=($env.LAST_EXIT_CODE)"
-                --terminal-width (term size).columns
-        )
+    $env.config.hooks.pre_execution ++= {||
+        let cmd = commandline | parse --regex '\s*(:?sudo\s+)?(?<cmd>\w+).*' | get cmd
+        if ($cmd | is-not-empty) {
+            export-env {
+                $env.LAST_CMD = $cmd | first
+            }
+        }
     }
 
-    config: ($env.config? | default {} | merge {
-        render_right_prompt_on_last_line: true
-    })
-
-    PROMPT_COMMAND_RIGHT: {||
-        (
-            ^::STARSHIP:: prompt
-                --right
-                --cmd-duration $env.CMD_DURATION_MS
-                $"--status=($env.LAST_EXIT_CODE)"
-                --terminal-width (term size).columns
+    load-env {
+        STARSHIP_SESSION_KEY: (random chars -l 16)
+        PROMPT_MULTILINE_INDICATOR: (
+            ^::STARSHIP:: prompt --continuation
         )
+
+        # Does not play well with default character module.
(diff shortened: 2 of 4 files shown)


src/init/starship.nu
```diff
@@ -2,37 +2,53 @@
...
-        ^::STARSHIP:: prompt --continuation
-    )
+export-env {
+    $env.STARSHIP_SHELL = "nu"
 
-    # Does not play well with default character module.
-    # TODO: Also Use starship vi mode indicators?
-    PROMPT_INDICATOR: ""
+    $env.config.hooks.pre_execution ++= {||
+        let cmd = commandline | parse --regex '\s*(:?sudo\s+)?(?<cmd>\w+).*' | get cmd
```
samuelallan72, line 9: A basic heuristic to get the command - this works for `sudo cmd args` and `cmd args`.  For my case this is good enough, as it's mostly to avoid too many unneeded notifications.  It's fine if some commands are missed.

src/init/starship.nu
```diff
@@ -2,37 +2,53 @@
...
-    PROMPT_MULTILINE_INDICATOR: (
-        ^::STARSHIP:: prompt --continuation
-    )
+export-env {
+    $env.STARSHIP_SHELL = "nu"
 
-    # Does not play well with default character module.
-    # TODO: Also Use starship vi mode indicators?
-    PROMPT_INDICATOR: ""
+    $env.config.hooks.pre_execution ++= {||
```
samuelallan72, line 8: I added this in a pre_execution hook because getting history in nushell may or may not be scoped to the same session.

src/configs/cmd_duration.rs
```diff
@@ -15,6 +15,7 @@ pub struct CmdDurationConfig<'a> {
     pub disabled: bool,
     pub show_notifications: bool,
     pub min_time_to_notify: i64,
+    pub notify_ignore_commands: Vec<&'a str>,
```
samuelallan72, line 18: This is a simple list of commands to ignore.  I'm not sure if we want to be more flexible for pattern matching (regex? list of regexes? regexes based on the entire commandline (command + args)?).

ekorchmar:
Would love this feature! Have you considered offloading the detection to the user provided patterns? To have `notify_ignore_command_patterns` config option instead of `notify_ignore_commands`?

samuelallan72:
@ekorchmar I'm not sure yet about how users would most like to use this; I'm curious to hear feedback and thoughts on the design.  What would you envision you would set as the value to `notify_ignore_command_patterns`?

I based this off the original request in #1933 and my own use case, which is simply: for certain commands I don't want to be notified on exit because they're interactive long running commands (ie. I'm explicitly closing them myself, so I don't need to be notified that they closed).  And it's only based on the command name; I can't think of cases yet where I might want to be notified for some patterns of a certain command but not others.  For me it would be a list of commands like vim, mutt, bash, sudoedit, etc.  I'm interested to hear how others would like to use it though.

I guess there are several options to consider:

- list of base commands (and do we ignore known 'prefixes' like `sudo`?)
- list of patterns/regexes over the entire command line
- list of patterns/regexes over just the base commands
- single regex string (eg. `^(sudo\s+)?(vim|mutt|sudoedit|bash)(\s+|$)`)
- ?

ekorchmar:
To follow the precedent, the only module that I can find that exposes regex in config is https://starship.rs/config/#kubernetes, and there is not much. So what I would suggest is:
1. Have a variable list of base commands that should be ignored; default would be like
```toml
notify_ignore_commands=["vim","nvim","mpv","mutt","mc","lazygit","nano","bash","zsh","fish","pwsh","tcsh"]
```
2. Have another variable list for known prefixes, default to just `known_prefixes = ["sudo"]`
These two would match on a very simple level, only looking at the first characters in command string before encountering space (allowing for optional prefix)

In code that would look something like:
```rust
regex_prefixes = known_prefixes.join("|");
ignore_commands = notify_ignore_commands.join("|")
resulting_regex = format!("^({} )?\\?({})[ $]", regex_prefixes, ignore_commands)
```

Note the backslash that can be used to escape aliased commands in most shells.

In addition, have another variable, default empty, that would contain a list of verbatim regexs that would be applied independently in a separate round; so if I would want to exclude all `git commit` commands that do not pass an explicit `-m` flag (and therefore start an `$EDITOR` session) I could do that by populating the variable `notify_ignore_command_patterns` with overengineered regex with negative group matching and what not. Simplified case:
```toml
# Default
notify_ignore_command_patterns =  []
# Commit without a message key (simplified) 
notify_ignore_command_patterns = ["^git commit$"]
```
Regexs in this array would be applied independently of the default regex.

samuelallan72:
This is rather outdated now, and I don't use the notify feature any more, so I'll close this.

PR #6173, opened 2024-08-08
fix(git_status): read proper name for core.fsmonitor flag

james-thorne-canva:
## Context

It seems like this is referencing the wrong config value, so it is unsetting the config value when running git statuses

#### Description
`core.fs_monitor` is not a valid config value, the proper name is `core.fsmonitor`

#### Motivation and Context

This should help speed up git status by using the `core.fsmonitor` git config

See the git status is overriding the `core.fsmonitor` to empty

image

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -345,7 +345,7 @@ impl<'a> Context<'a> {
 
                 let fs_monitor_value_is_true = repository
                     .config_snapshot()
-                    .boolean("core.fs_monitor")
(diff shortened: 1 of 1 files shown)


davidkna (approved):
LGTM

james-thorne-canva:
Hey @davidkna, what is the process of getting this released? Just wondering as it would help speed up my git status' commands alot

davidkna:
Thanks for the fix @james-thorne-canva!

PR #6178, opened 2024-08-09
feat(directory): Add separate repo_truncation_symbol to directory module

### LINKED ISSUE #6177: Separate truncation symbol for git repositories in directory module

Currently if both `truncation_length` and `truncate_to_repo` share the same `truncation_symbol`. It would be good to have a different symbol for git rep truncation.

Add a `repo_truncation_symbol` config option, which provides a different symbol for git repo truncation as opposed to 'long path' truncation. The end result would be something like:

```
trunction_symbol = ".../"
repo_truncation_symbol = "G"
```

```
.../ share/icons/hicolor > Truncated path by length

G my_repo > Truncated repo
```

mrichar1:
#### Description
Adds `repo_truncation_symbol` to directory module to allow a different symbol to be displayed if repo truncation rather than 'path length' truncation.

#### Motivation and Context
This avoids confusion as to whether the path is truncated for length or a git repo, and also allows no symbol if desired for either case.

Closes #6177

#### Screenshots

* Current release (with `truncate_to_repo = true`, `truncation_length = 3` and `truncation_symbol = .../`) - both types of truncation share the same symbol:
2024-08-09-130759_441x81_scrot

* New PR release (with `repo_truncation_symbol` set as well) - different symbol for different truncation types:
2024-08-09-130836_420x80_scrot

Tests updated to cover new behaviour. All tests passing with `cargo test`

#### Checklist:


src/configs/directory.rs
```diff
@@ -23,6 +23,7 @@ pub struct DirectoryConfig<'a> {
     pub disabled: bool,
     pub read_only: &'a str,
     pub read_only_style: &'a str,
+    pub repo_truncation_symbol: &'a str,
```
davidkna, line 26: To avoid breaking existing configurations, I would prefer defaulting to `truncation_symbol` if `repo_truncation_symbol` was not set

```suggestion
    pub repo_truncation_symbol: Option<&'a str>,
```

src/configs/directory.rs
```diff
@@ -44,6 +45,7 @@ impl<'a> Default for DirectoryConfig<'a> {
             disabled: false,
             read_only: "🔒",
             read_only_style: "red",
+            repo_truncation_symbol: "",
```
davidkna, line 48:
```suggestion
            repo_truncation_symbol: None,
```

davidkna (changes requested):

taylorjonl:
What are the chances this gets merged?

Not sure if this is related but I think it would be interesting to also add a custom truncation_length when it is a repo and you have `truncate_to_repo` set to true.  Today when that is set to true you get just the repo name but that can be confusing if you are working with forks.  Normally my folder structure for my repos always has the repos in a folder based on their organization so I can keep track of when I am in my forked repo or in the upstream repo.

PR #6186, opened 2024-08-12
docs(git_commit): add extra trailing space to `tag_symbol`

injust:
#### Description
The [documented default](https://starship.rs/config/#options-32) for `tag_symbol` has one trailing space, but it actually has 2 trailing spaces. This removes the extra trailing space.

#### Motivation and Context

Removes an extra trailing space to match the documented default.

#### Checklist:


davidkna:
As a multi-character symbol, most terminals display the width incorrectly, so this was likely intentional. Would you mind updating the docs to match instead?

injust:
> As a multi-character symbol, most terminals display the width incorrectly, so this was likely intentional. Would you mind updating the docs to match instead?

@davidkna Makes sense. Fixed!

davidkna (approved):

PR #6196, opened 2024-08-16
feat(direnv): allow direnv to detect environment variables

### LINKED ISSUE #6199: direnv: module is not being displayed in subdirectories, despite direnv being active
#### Current Behavior
direnv module does not activate on any of subdirectories containing `.envrc` file.

#### Expected Behavior
Module is always displayed while envrc is loaded, even in subdirectories.

image

#### Possible Solution
https://github.com/starship/starship/blob/171746a58408cbc0d48dedc52e21327bd35fccd4/src/modules/direnv.rs#L19
`direnv_applies` variable should not depend on directory content, as leaving a directory for a subdirectory breaksde detection. There are two better ways to check if direnv is active:
1. Monitoring any of env variables direnv sets (`$DIRENV_FILE`, `$DIRENV_DIR`)
2. Calling `direnv --status`, which is also done elsewhere in the code; may be noticeably slower than looking up env variables.

Adding an option for the module to detect environment variables (prepopulated with `[$DIRENV_FILE, $DIRENV_DIR]`) may be a clean enough solution in itself, but I would argue that looking at the directory content at all should not be even attempted for simplicity.

#### Environment
- Starship version: 1.20.1
- zsh version: zsh 5.9 (x86_64-pc-linux-gnu)
- Operating system: Arch Linux Rolling Release
- Git Commit Hash: f505324da
- Branch/Tag: v1.20.1
- Rust Version: rustc 1.80.0 (051478957 2024-07-21) (Arch Linux rust 1:1.80.0-1)
- Rust channel:  release
- Build Time: 2024-07-27 12:27:58 +00:00

```bash
# Starsip prompt
source <(/usr/bin/starship init zsh --print-full-init)
# If we are in TTY, use legacy color config
if [[ $XDG_SESSION_TYPE == "tty" ]]; then
    export STARSHIP_CONFIG=~/.config/starship_tty.toml
fi
```

nickconway:
#### Description
Allows the direnv module to detect environment variables such as DIRENV_FILE. 

#### Motivation and Context

This allows for the moodule to work properly in subdirectories when direnv is sill active. 
Closes #6199

image

#### Checklist:


src/modules/direnv.rs
```diff
@@ -13,9 +13,10 @@ use serde::Deserialize;
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("direnv");
     let config = DirenvConfig::try_load(module.config);
+    let has_detected_env_var = context.detect_env_vars(&config.detect_env_vars);
 
-    let direnv_applies = !config.disabled
-        && context
+    let direnv_applies = !config.disabled && has_detected_env_var
+        || context
```
davidkna, line 19: I think the `||` expression should be wrapped in braces.

silentroach:
@nickconway could you please the review comments?

nickconway:
> @nickconway thank you for this PR, could you please fix the review comments?

Apologies for the delay, I've implemented the change now!

silentroach:
@nickconway could you please fix the typo from cicd check? Looks like nobody will recheck this PR until it is green :{

davidkna (approved):

davidkna:
Sorry, I just did not notice the first update.

davidkna:
Thanks for the contribution @nickconway!

PR #6200, opened 2024-08-19
docs(config, preset): fix Buf module config docs and No Empty Icons preset

injust:
#### Description
Update config docs because the Buf module is shown even if the `buf` CLI is not installed.

Also update the No Empty Icons preset to make the Buf module consistent with other modules. I noticed this because (unlike all but one of the other modules in this preset) this one doesn't have `$version ` in parentheses.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -579,10 +579,7 @@ discharging_symbol = '💦 '
 
 ## Buf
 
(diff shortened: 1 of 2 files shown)


andytom (approved):

davidkna:
Thanks for fixing this @injust, and thanks for the review @andytom.

PR #6225, opened 2024-09-01
feat: PowerShell Init - Combine ViModeChangeHandler

### LINKED ISSUE #6217: PSReadLine mode indicator
#### Current Behavior

1. Launch powershell (with starship)
2. On PSReadLine mode change the whole prompt blinks and cursor does not change
3. Comment out line with starship in the config
4. Launch powershell
5. See cursor changing its shape depending on the editing mode
6. Prompt does not blink

#### Expected Behavior

1. Launch powershell (with starship)
2. On PSReadLine mode change the whole prompt does not blink and cursor changes shapes according to the `$PROFILE`

#### Possible Solution

Maybe I have missed some configuration option. The cursor shouldn't be altered unless configured so in the config.

#### Environment
- Starship version: 1.20.1
- pwsh version: PowerShell 7.4.4
- Terminal emulator: WezTerm 20240203-110809-5046fc22
- Git Commit Hash: f505324
- Branch/Tag: master
- Rust Version: rustc 1.80.0 (051478957 2024-07-21)
- Build Time: 2024-07-27 09:12:42 +00:00

```pwsh
$env:POWERSHELL_TELEMETRY_OPTOUT = 1
$env:POWERSHELL_UPDATECHECK_OPTOUT = 1
$env:POWERSHELL_UPDATECHECK = "Off"
$OutputEncoding = [Console]::InputEncoding = [Console]::OutputEncoding = New-Object System.Text.UTF8Encoding

Import-Module PSReadLine

function OnViModeChange
{
  if ($args[0] -eq 'Command')
  {
    Write-Host -NoNewLine "`e[2 q"
  } else
  {
    Write-Host -NoNewLine "`e[6 q"
  }
}

$PSReadLineOptions = @{
  EditMode = "vi"
  HistoryNoDuplicates = $true
  HistorySearchCursorMovesToEnd = $true
  ViModeIndicator = "Script"
  ViModeChangeHandler = $Function:OnViModeChange
}
Set-PSReadLineOption @PSReadLineOptions
Set-PSReadLineKeyHandler -Chord Ctrl+w -Function BackwardDeleteWord
Set-PSReadLineKeyHandler -Chord Ctrl+k -Function HistorySearchBackward
Set-PSReadLineKeyHandler -Chord Ctrl+j -Function AcceptLine
Set-PSReadLineKeyHandler -Chord Ctrl+y -Function AcceptSuggestion

Set-Item -Path Env:RUSTC_WRAPPER -Value sccache

# Import-Module posh-git
# $GitPromptSettings.BranchColor.BackgroundColor = [ConsoleColor]::Magenta

Invoke-Expression (& { (zoxide init --cmd cd powershell | Out-String) })
```

#### Starship Configuration

```toml
"$schema" = "https://starship.rs/config-schema.json"

format = "$directory$git_branch$git_status$git_state$cmd_duration$status$character"
# right_format = "$cmd_duration$status"
add_newline = false

[cmd_duration]
format = "[$duration]($style) "
style = "yellow"

[username]
style_root = "red"
style_user = "yellow"
show_always = true
format = "[$user]($style) "
aliases = { shirshov = "%", pinbraerts = "%", dima = "%", root = "$", wheel = "$", admin = "$" }

[git_branch]
format = "[$symbol$branch(:$remote_branch)]($style)"
style = "purple"

[git_status]
modified = "[*](yellow)"
staged = "[+](green)"
deleted = "[-](red)"
untracked = "?"
conflicted = "[=](red)"
ahead = "[↑](green)"
behind = "[↓](red)"
renamed = "[→](yellow)"
format = "$all_status$ahead_behind "

[directory]
use_os_path_sep = false
truncation_length = 100
truncate_to_repo = false
style = "cyan"

[status]
format = "[$status]($symbol) "
disabled = false
success_symbol = "green"
symbol = "red"

```

HeyItsGilbert:
If a user has defined a ViModeChangeHandler setting this would respect that as well as redraw the prompt (to allow rendering things like the character change).

#### Description
Users don't realize that the `ViModeChangeHandler` is redrawing the prompt. If the user sets their own, as show in the PSReadLine example [to change cursor type](https://learn.microsoft.com/en-us/powershell/module/psreadline/set-psreadlineoption?view=powershell-7.4#example-6-use-vimodechangehandler-to-display-vi-mode-changes), they don't realize it'll be overwritten by the `starship init powershell` call.

#### Motivation and Context

Closes #6217 

#### Checklist:

### DIFF
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -213,10 +213,11 @@ $null = New-Module starship {
     try {
         # Combine user defined ViModeChangeHandler if it exists
         if((Get-PSReadLineOption).ViModeChangeHandler){
-            Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler [scriptblock]::Create(
+            Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler $([scriptblock]::Create(
                 "[Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()" +
+                "`n" +
                 (Get-PSReadLineOption).ViModeChangeHandler.ToString()
-            )
+            ))
         } else {
             Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler {
                 [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
--- a/src/init/starship.ps1
+++ b/src/init/starship.ps1
@@ -211,8 +211,16 @@ $null = New-Module starship {
     )
 
     try {
-        Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler {
-            [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
+        # Combine user defined ViModeChangeHandler if it exists
+        if((Get-PSReadLineOption).ViModeChangeHandler){
+            Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler [scriptblock]::Create(
+                "[Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()" +
+                (Get-PSReadLineOption).ViModeChangeHandler.ToString()
+            )
+        } else {
+            Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler {
+                [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
+            }
         }
     } catch {}
 


src/init/starship.ps1
```diff
@@ -211,8 +211,17 @@ $null = New-Module starship {
...
 
     try {
-        Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler {
-            [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()
+        # Combine user defined ViModeChangeHandler if it exists
+        if((Get-PSReadLineOption).ViModeChangeHandler){
+            Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler $([scriptblock]::Create(
+                "[Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt()" +
+                "`n" +
+                (Get-PSReadLineOption).ViModeChangeHandler.ToString()
```
davidkna, line 219: Would it be possible to use and copy and execute the old handler without stringification?

HeyItsGilbert: The ViModeChangeHandlers needs to be a script block. AFAIK there isn't a good way to to join scriptblocks without concatenating them this way. I'll ask around though just to be certain.

Do you have a specific concern? I suspect this is more of an edge case then not. And the resource cost would be minimal since it would only happen once at profile load. The stringification does not mean it's interpolated.

romero126: This is going to create unexpected behavior, I recommend changing it to the following code

```pwsh
 Set-PSReadLineOption -ViModeIndicator script -ViModeChangeHandler $([scriptblock]::Create(
                [Microsoft.PowerShell.PSConsoleReadLine]::InvokePrompt() +
                "`n" +
                (Get-PSReadLineOption).ViModeChangeHandler.ToString()
```

davidkna: I was mostly concerned with strings causing not preserving the full semantics of any given handler and potentially issues with string concatenation itself.

@romero126 Do you still have concerns with the current approach?

davidkna (changes requested):
I tried this PR with the linked sample `ViModeChangeHandler`, and it did not work correctly because the handler function arguments are not forwarded to the preserved handler. Please update the code to forward any function arguments.

davidkna (approved):
LGTM

HeyItsGilbert:
@davidkna I don't have access to merge. Are we waiting on something?

davidkna:
Thanks for contribution @HeyItsGilbert!

PR #6228, opened 2024-09-03
feat(python): detect pixi and ipynb files

probaku1234:
add support detecting pixi.toml and .ipynb file for python module

#### Description

update default Python module config to detect pixi.toml file and ipynb file extension

#### Motivation and Context

Closes https://github.com/starship/starship/issues/6194

#### Checklist:


davidkna (approved):

.github/config-schema.json
andytom, line 1: Looks like something has gone wrong with the schema, the github app is telling me this is a binary file now and the preset check is failing. Can you try running the following to see if that fixes it

```
cargo run --locked --features config-schema -- config-schema > .github/config-schema.json
```

probaku1234: It seems the command you posted make the json file as bin file, I don't know why. 
So I had to manually create and push the file.

andytom (approved):

davidkna:
Thanks for the contribution @probaku1234, and thanks for spotting the issue in the second review @andytom!

pavelzw:
iiuc this will still search for python on PATH, right?

pixi usually stores its python in `$PWD/.pixi/envs/default/bin/python` on unix (somewhere else on windows) (and puts it on PATH only when executing `pixi run ...` similar to node, ...)

Should starship look there as well?

pavelzw:
also, there are use cases where pixi is used for non-python projects as well. pixi and conda are technically language-agnostic

davidkna:
> pixi usually stores its python in `$PWD/.pixi/envs/default/bin/python` (you can replace `default` with other env names as well if you use [multi environments](https://pixi.sh/latest/features/multi_environment/#feature-environment-set-definitions)) on unix (somewhere else on windows) (and puts it on PATH only when executing `pixi run ...` similar to node, ...)
> 
> Should starship look there as well?

Implicitly running executables relative to the CWD should be avoided in starship due to the security risk when entering untrusted directories. I wager that pixi has some mechanism to put its tools into `$PATH` like a python venv would? direnv could be used to handle that automatically.

pavelzw:
yes when using direnv the issue won't occur ([docs](https://pixi.sh/latest/features/environment/#using-pixi-with-direnv)) but i'm not sure whether everybody uses it

PR #6231, opened 2024-09-03
feat(git_commit): add a option to toggle hash in git_commit

### LINKED ISSUE #6219: Show git tag without showing commit

I want `starship` to show current git tags without the commit hash

Either add a `[git_commit]` module or add `git_commit.only_show_hash_when_detached` field or add `$hash_if_detached` variable.

This configuration shows git tags, but also commit hashes, which is not exactly what I want:

```toml
[git_commit]
tag_disabled = false
only_detached = false
```

gr8mtyr:
#### Description

Added a configuration option to toggle the commit hash in the *git_commit* module.

screenshot-Tue-2024-09--03-20-59-31
screenshot-Tue-2024-09--03-20-59-53

#### Checklist:

### DIFF
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -17,7 +17,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let git_head = git_repo.head().ok()?;
 
     let is_detached = git_head.is_detached();
-    if config.only_detached && !is_detached {
+    if config.tag_disabled && config.only_detached && !is_detached {
         return None;
     }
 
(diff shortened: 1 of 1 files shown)


davidkna:
The `hash`-variable can already be hidden altogether by removing `$hash` from the module format-string. The original issue is asking for an option that is also dependent on the detached-state.

MahmoudESSE:
Thats a way, but I think it would be easier for biginners if there was an option to disable the hash been visible without having the overhead of making their own format string.

For advanced users they can tweak their format, for example like the advanced configuration in the website.

jacobeturpin:
This _almost_ achieves what I would want. I want to always display tags but only display the hash in detached mode. This seems like it could be achieved with an entity-specific `only_detached` for both hash and tag.

gr8mtyr:
Hi, I think i understand the issue now, i can make something so that tags can be shown if the `tag_disabled` is set to false, so that tags are independant of `only_detached` while keeping them dependant on the `tag_max_candidates`, and the hashcode will still show only when detached, would it be ok to change current logic for that `tag_disabled` or should another option be added to trigger this state.

gr8mtyr:
@davidkna, I fixed my implementation and removed code that was not needed from my part, now the tag does not depend on detached state to be displayed, the hashcode also will displayed if it is in a detached state.

PR #6234, opened 2024-09-04
fix: Fixed "Click to download TOML" links

zaira-bibi:
#### Description
I have added the {download} attribute to all presets' downlink links.

#### Motivation and Context
Previously, the download links were broken. They all navigated to 404 page instead of downloading the TOML file. 
Closes: #6063 

#### Checklist:

### DIFF
--- a/docs/ar-SA/presets/bracketed-segments.md
+++ b/docs/ar-SA/presets/bracketed-segments.md
@@ -12,6 +12,6 @@ This preset changes the format of all the built-in modules to show their segment
 starship preset bracketed-segments -o ~/.config/starship.toml
 ```
 
-[Click to download TOML](/presets/toml/bracketed-segments.toml)
+[Click to download TOML](/presets/toml/bracketed-segments.toml){download}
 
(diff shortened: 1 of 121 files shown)


davidkna:
LGTM, but if possible, revert the changes to the translated files to avoid conflicts with [crowdin](https://translate.starship.rs), which will update the automatically.

zaira-bibi:
> LGTM, but if possible, revert the changes to the translated files to avoid conflicts with [crowdin](https://translate.starship.rs), which will update the automatically.

Appreciate the feedback! Will push in a commit soon.

davidkna (approved):

revolunet:
Hello, looks like some issue persist in Firefox

In https://starship.rs/presets/plain-text, the link to download the TOML `https://starship.rs/presets/toml/plain-text-symbols.toml` is correct but triggers a 404 when i click :thinking:

if i open in a new tab the urls works as expected


Kapture 2024-09-30 at 00 42 52 2

PR #6265, opened 2024-09-16
fix(fish): use correct input function in transient execute

kidonng:
#### Description

#6107 (unreleased) introduced a bug that mistakenly make `Enter` `accept-autosuggestion` when pager is open:

https://github.com/user-attachments/assets/fdd3b897-4421-4916-957c-0d05bafa87a6

This pull request uses the check from tide's transient prompt implementation, which in essence falls back to `execute` to only accept the selection in pager:

https://github.com/IlanCosman/tide/blob/44c521ab292f0eb659a9e2e1b6f83f5f0595fcbd/functions/fish_prompt.fish#L154-L164

#### Motivation and Context

Closes #

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -64,12 +64,7 @@ function reset-transient --on-event fish_postexec
 end
 
 function transient_execute
-    if commandline --paging-mode
-        commandline -f accept-autosuggestion
-        return
-    end
-    commandline --is-valid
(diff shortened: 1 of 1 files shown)


andytom (approved):

davidkna:
Thanks for the fix before this got released @kidonng, and thanks for review @andytom.

PR #6269, opened 2024-09-18
feat(status): add `{success,failure}_style` options

T0mstone:
#### Description
Add two new config options for the `style` module that allow the user to style the status differently based on whether it was successful or not.

#### Motivation and Context
The color red usually indicates that something went wrong, so when using `pipestatus = true` or `success_symbol`, you wouldn't want a status of zero to show up in red.
By adding this feature, the module can style the status based on whether it was successful or not, allowing for easy discernment when looking at the prompt.

Screenshot 2024-09-18_16h33

#### Checklist:

I'm unsure which interactions to write unit tests for, since this feature interacts a lot with the previously existing stuff.


davidkna (approved):

davidkna:
Thanks for the contribution @T0mstone!

PR #5943, opened 2024-04-28
 fix(container): Corrected systemd-nspawn identification

brewedBorrows:
#### Description

I have added a check, that will read the contents of "container-manger" file before deciding to show 'OCI' or 'nspawn'.

#### Motivation and Context

Fix to Issue : https://github.com/starship/starship/issues/5937

#### Screenshots (if appropriate):
BEFORE : 
Screenshot from 2024-04-25 14-34-33

AFTER : 
Screenshot from 2024-04-28 10-51-40

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

PR #6287, opened 2024-09-29
feat(container): identify systemd-nspawn environment

### LINKED ISSUE #5937: nspawn wrongly detected as OCI
#### Current Behavior
While running in a systemd-nspawn, the `container` module report `OCI` 

#### Expected Behavior
`systemd-nspawn` or `nspawn` to be reported

When `/run/host/container-manager` exist, `OCI` is hardcoded https://github.com/starship/starship/blob/cce9d335095a10e7f53cb5c6e54d27d3ce8ac1e1/src/modules/container.rs#L24-L27
But systemd-nspawn also create this file with `systemd-nspawn` inside https://github.com/systemd/systemd/blob/8f0de5c433929782f7582a743d3b3d88bea9696d/src/nspawn/nspawn.c#L4148-L4149
This check is made early, bypassing the `/run/systemd/container` one.

#### Possible Solution
Check the file content.

frectonz:
#### Description

This PR implements support for identifying `systemd-nspawn` based environments. This change was initially implemented in #5943 but since that PR can't be merged due to failing tests i have decided to create a new PR to address the issue.

#### Motivation and Context

Closes #5937 
Closes #5943

A `systemd-nspawn` container gets wrongly detected as `OCI`.

The following screenshots were taken inside a [NixOS](https://nixos.org/) machine. The [`nixos-container`](https://wiki.nixos.org/wiki/NixOS_Containers) command is just a wrapper around `systemd-nspawn` to make using it easier.

Before
image

After
image

#### Checklist:

### DIFF
--- a/src/modules/container.rs
+++ b/src/modules/container.rs
@@ -21,19 +21,6 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             return Some("OpenVZ".into());
         }
 
-        if context_path(context, "/run/host/container-manager").exists() {
-            // systemd-nspawn
-            if read_file("/run/host/container-manager")
-                .map(|c| c.contains("systemd-nspawn"))
-                .unwrap_or_default()
-            {
-                return Some("nspawn".into());
-            }
-
-            // OCI
-            return Some("OCI".into());
-        }
-
         let container_env_path = context_path(context, "/run/.containerenv");
 
         if container_env_path.exists() {
@@ -75,6 +62,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         if context_path(context, "/.dockerenv").exists() {
             // docker
             return Some("Docker".into());
+        }
+
+        if context_path(context, "/run/host/container-manager").exists() {
+            // OCI
+            return Some("OCI".into());
         }
 
         None
--- a/src/modules/container.rs
+++ b/src/modules/container.rs
@@ -22,6 +22,14 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         }
 
         if context_path(context, "/run/host/container-manager").exists() {
+            // systemd-nspawn
+            if read_file("/run/host/container-manager")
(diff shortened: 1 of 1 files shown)


frectonz:
Should i add tests for this change? If so how can i do that, can `read_file` be mocked?

davidkna:
> Should i add tests for this change? If so how can i do that, can `read_file` be mocked?

Please add tests. You can use `context_path` to mock the path. For the test itself, please try doing something in the manner of `containerenv_systemd`.
I would also suggest combining `exists` and `read_file` into a single `if let Ok(c) = read_file` statement.

frectonz:
So i did some more investigating and found out there's actually some code that already exists to detect systemd environments. The current problem of nspawn being wrongly detected as `OCI` can be fixed by deprioritizing the OCI check, that is making `/run/host/container-manager` be the last file checked.

image

If it is still worth it to detect nspawn based environments i believe the best way to add that functionality would be inside this [match statement](https://github.com/starship/starship/blob/0c32314f86e9666a178e579ea7b4510aa50be952/src/modules/container.rs#L60) that checks the contents of `/run/systemd/container`.

image

So my question is should i add specific detection logic for `nspawn` or is detecting `Systemd` enough?

frectonz:
> So i did some more investigating and found out there's actually some code that already exists to detect systemd environments. The current problem of nspawn being wrongly detected as `OCI` can be fixed by deprioritizing the OCI check, which means making `/run/host/container-manager` be the last file checked.
> 
> image
> 
> If it is still worth it to detect nspawn based environments i believe the best way to add that functionality would be inside this [match statement](https://github.com/starship/starship/blob/0c32314f86e9666a178e579ea7b4510aa50be952/src/modules/container.rs#L60) that checks the contents of `/run/systemd/container`.
> 
> image
> 
> So my question is should i add specific detection logic for `nspawn` or is detecting `Systemd` enough?

Referring to my last question, if detecting just `Systemd` environment is enough, i think this PR can be merged.

PR #6315, opened 2024-10-12
refactor: bump msrv to remove once-cell, improve logger test & handle clippy warnings

davidkna:
#### Description

This PR bumps the MSRV to 1.80. This enables replacing `once_cell` with std-provided `LazyLock` and `OnceLock`. The std version does not provide `get_or_try_init yet ` so I switched relevant code to store `Result<_, _>`instead. This might be the better anyway because the documentation sounds like any errors would not be cached.

I also made the `test_cleanup` for the logger target-agnostic, but it still does not run as-is on Windows due to requiring elevated permissions to change file dates.

I also fixed a few clippy warnings to get ahead of the upcoming rust v1.82 release.

Edit: also handled the recent dprint TOML-formatter upgrade.

#### Motivation and Context

Closes #

#### Checklist:


matchai (approved):

davidkna:
Thanks for the review @andytom and @matchai!

PR #6335, opened 2024-10-19
feat: Add support for pixi

pavelzw:
#### Description

[pixi](https://pixi.sh) is a modern package manager that builds on top of the conda ecosystem.
A pixi project consists of a pixi.toml (or pyproject.toml) as well as a pixi.lock.
You can be inside of an activated pixi environment using `pixi shell` or using [pixi with direnv](https://pixi.sh/latest/features/environment/#using-pixi-with-direnv). This can be detected by looking at the `PIXI_ENVIRONMENT_NAME` variable.

#### Motivation and Context

I would like to have better pixi support next to conda support in starship.

I tested all use cases I could think of on my machine.

- inside of a pixi environment (using `pixi shell`)
- in a directory but not inside of a pixi environment
- ensure that pixi envs don't clash with conda envs

#### Checklist:

---

Example walkthrough

starship-pixi


docs/config/README.md
```diff
@@ -3570,6 +3571,57 @@ The `pijul_channel` module shows the active channel of the repo in your current
...
+
+[python]
+# customize python binary path for pixi
+python_binary = [
+  # this is the python from PATH if in a pixi shell
+  # (assuming you don't have python on your global PATH)
+  "python",
+  # fall back to pixi's python if it's available
+  ".pixi/envs/default/bin/python",
+]
```
pavelzw, line 3622: this is a bit hacky but works... not sure whether you want this workaround in your docs or not. i am also fine with removing it again

src/configs/python.rs
```diff
@@ -44,7 +44,6 @@ impl<'a> Default for PythonConfig<'a> {
                 "tox.ini",
                 "setup.py",
                 "__init__.py",
-                "pixi.toml",
```
pavelzw, line 47: pixi is a language-agnostic package manager. there are many valid use cases where you have a non-python project but still use pixi. since we now have a proper pixi module, i would be in favor of removing this (as most of the time you should have `.py` files anyway somewhere in your cwd

src/modules/conda.rs
```diff
@@ -20,6 +20,9 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     if config.ignore_base && conda_env == "base" {
         return None;
     }
+    if config.ignore_pixi_envs && context.get_env("PIXI_ENVIRONMENT_NAME").is_some() {
+        return None;
+    }
```
pavelzw, line 25: pixi sets `CONDA_DEFAULT_ENV` as well thus we need to deactivate the conda module in this case

pavelzw:
@andytom could you please take a look at this?

pavelzw:
i published pre-built binaries for `linux-x86_64`, `linux-aarch64`, `macos-x86_64`, `macos-aarch64` and `win_x86_64` on my own conda channel: https://prefix.dev/channels/yolo-forge/packages/starship-fork-pavelzw

you can install it on your system by running

```
pixi global install -c https://prefix.dev/yolo-forge -c conda-forge starship-fork-pavelzw
```

(packages are being built in CI here: https://github.com/pavelzw/yolo-forge)

docs/config/README.md
```diff
@@ -3603,6 +3604,57 @@ The `pijul_channel` module shows the active channel of the repo in your current
...
+| `symbol`                   | `'📦 '`                                                 | The symbol used before the environment name.                                          |
+| `style`                    | `'yellow bold'                                          | The style for the module.                                                             |
+| `show_default_environment` | `true`                                                  | Whether to indicate that the `default` environment of your project is activated.      |
+| `pixi_binary`              | `['pixi']`                                              | Configures the python binaries that Starship should execute when getting the version. |
+| `detect_files`             | `['pixi.toml']`                                         | Which filenames should trigger this module                                            |
+| `disabled`                 | `false`                                                 | Disables the `conda` module.                                                          |
```
lucascolley, line 3628:
```suggestion
| `disabled`                 | `false`                                                 | Disables the `pixi` module.                                                          |
```

README.md
```diff
@@ -221,21 +221,21 @@ curl -sS https://starship.rs/install.sh | sh
...
-| Manjaro            |                         | `pacman -S starship`                                          |
-| NixOS              | [nixpkgs]               | `nix-env -iA nixpkgs.starship`                                |
-| openSUSE           | [OSS]                   | `zypper in starship`                                          |
-| Ubuntu 25.04+      | [Ubuntu Universe]       | `apt install starship`                                        |
-| Void Linux         | [Void Linux Packages]   | `xbps-install -S starship`                                    |
+| Distribution       | Repository              | Instructions                                                                  |
+| ------------------ | ----------------------- | ----------------------------------------------------------------------------- |
+| **_Any_**          | **[crates.io]**         | `cargo install starship --locked`                                             |
+| _Any_              | [conda-forge]           | `conda install -c conda-forge starship` <br /> `pixi global install starship` |
```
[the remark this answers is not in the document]
pavelzw: [`bf08d39` (#6335)](https://github.com/starship/starship/pull/6335/commits/bf08d39149f543658ae98d389d6c9cebc8e359bd)

docs/config/README.md
```diff
@@ -3603,6 +3604,57 @@ The `pijul_channel` module shows the active channel of the repo in your current
...
+| `format`                   | `'via [$symbol($version )(\($environment\) )]($style)'` | The format for the module.                                                            |
+| `version_format`           | `'v${raw}'`                                             | The version format. Available vars are `raw`, `major`, `minor`, & `patch`.            |
+| `symbol`                   | `'📦 '`                                                 | The symbol used before the environment name.                                          |
+| `style`                    | `'yellow bold'                                          | The style for the module.                                                             |
+| `show_default_environment` | `true`                                                  | Whether to indicate that the `default` environment of your project is activated.      |
+| `pixi_binary`              | `['pixi']`                                              | Configures the python binaries that Starship should execute when getting the version. |
```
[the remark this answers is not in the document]
pavelzw: [`5c0d51f` (#6335)](https://github.com/starship/starship/pull/6335/commits/5c0d51fdb0f3811cb22bd9512105c4fc8cccda59)

docs/config/README.md
```diff
@@ -909,6 +909,7 @@ If you use [pixi](https://pixi.sh), you can disable pixi's prompt modifier by ru
 | `style`             | `'bold green'`                         | The style for the module.                                                                                                                                                                                   |
 | `format`            | `'via [$symbol$environment]($style) '` | The format for the module.                                                                                                                                                                                  |
 | `ignore_base`       | `true`                                 | Ignores `base` environment when activated.                                                                                                                                                                  |
+| `ignore_pixi_envs`  | `true`                                 | Ignores pixi environments when activated (in favor of the pixi starship module). They are recognized by the `PIXI_ENVIRONMENT_NAME` environemnt variable.                                                   |
```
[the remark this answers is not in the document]
pavelzw: didn't know about this feature, thanks!

[`d413dd0` (#6335)](https://github.com/starship/starship/pull/6335/commits/d413dd021327990aa0e458a26afb5dfd8ed56200)

docs/config/README.md
pavelzw: fair, then let's leave it out [`4246c3e` (#6335)](https://github.com/starship/starship/pull/6335/commits/4246c3ee67cc633dbad7487dd76afe319b4fbbd2)

docs/config/README.md
```diff
@@ -3603,6 +3604,57 @@ The `pijul_channel` module shows the active channel of the repo in your current
...
+| `version_format`           | `'v${raw}'`                                             | The version format. Available vars are `raw`, `major`, `minor`, & `patch`.            |
+| `symbol`                   | `'📦 '`                                                 | The symbol used before the environment name.                                          |
+| `style`                    | `'yellow bold'                                          | The style for the module.                                                             |
+| `show_default_environment` | `true`                                                  | Whether to indicate that the `default` environment of your project is activated.      |
+| `pixi_binary`              | `['pixi']`                                              | Configures the python binaries that Starship should execute when getting the version. |
+| `detect_files`             | `['pixi.toml']`                                         | Which filenames should trigger this module                                            |
```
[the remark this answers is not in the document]
pavelzw: so `detect_extensions` and `detect_folders`? [`cac81f0` (#6335)](https://github.com/starship/starship/pull/6335/commits/cac81f05de1e79011f38d4cf23180112e03997d5)

pavelzw:
thanks for your review @davidkna! i addressed your comments

src/modules/pixi.rs
```diff
@@ -0,0 +1,140 @@
...
+        return None;
+    }
+
+    let pixi_environment_name = if !config.show_default_environment
+        && pixi_environment_name == Some("default".to_string())
+    {
+        None
+    } else {
+        pixi_environment_name
+    };
```
davidkna, line 33: I would prefer fully disabling the module if the default environment is detected like the conda module, at least by default.

davidkna (changes requested):

src/modules/pixi.rs
pavelzw: the way pixi and conda work in that regard are very different. in conda, the `base` environment is usually _always_ active when you are in your shell (imo stupid design decision but okay). there, it definitely makes sense to skip showing this in starship as it's present always anyway.

for pixi, on the other hand, you are not always inside of the `default` environment but only if you run `pixi shell` (which is equivalent to `conda activate <my-env>`).
for me, it makes more sense to see the `(default)` because it indicates me that i'm actually inside of a pixi shell (which isn't always the case).

that being said, if you feel very strongly about this, i can also set the default behavior to `show_default_environment: false`

davidkna: Sorry for not being clearer, I was referring to using `return None` instead of only hiding the environment name. Though, if base envs are actively enabled the current behaviour could be fine, even if the option isn't quite consistent with the conda module.

pavelzw: ah okay but i still think that `return None` would be a bit weird here:

if you are not in a pixi shell but in a pixi project (i.e. `pixi.toml` exists), you will still get `🧚 v0.45.0` displayed. once you would go into the `pixi shell`, you would get nothing displayed anymore.

pavelzw: > if base envs are actively enabled the current behaviour could be fine, even if the option isn't quite consistent with the conda module

there is no "base env" in pixi and going inside of the `default` environment is still an active decision by the user so i think it makes sense that this isn't consistent with the conda module

pavelzw:
@davidkna thanks for your review :) i incorporated your suggestions

davidkna (approved):

davidkna:
Thanks for the contribution @pavelzw!

PR #6346, opened 2024-10-23
feat(nodejs): disable if bun project files detected

### LINKED ISSUE #5805: Questionable prompt behaviour when both bun and node are installed
#### Current Behavior
- Only node installed (`npm init`), correctly treated as a Node.js project.
image

&nbsp;
- Only bun installed (`bun init`), bun doesn't automatically add a `version` to `package.json` so both I've included two examples. You could argue this is inaccurate as the Node.js logo is still displayed even when Node.js is not installed on the system.
image
image

&nbsp;
- Both bun and node installed (`npm init`), correctly treated as a Node.js project.
image

&nbsp;
- Both bun and node installed (`bun init`), incorrectly shows irrelevant Node.js version.
image
image

#### Expected Behavior
Since bun is first and foremost a drop-in replacement for Node.js, I would expect it to override any mention of Node.js if `bun.lockb` or `bunfig.toml` are present.

Technically, bun can be used as a standalone package manager and is compatible with Node.js but I'd wager the vast majority of its users are using it as a runtime as well. Once bun is supported by corepack and the bun team add support for the `engines` field in `package.json` it will be possible to establish which specific information to display but for now a guess needs to be made.

N/A

#### Possible Solution
I assume Starship already has certain detections that are able to take precedence over others, or the capability to match file negatively. If `bun.lockb` or `bunfig.toml` are present, only display the bun runtime info and not Node.js info.

#### Environment
- Starship version: 1.17.1
- zsh version: zsh 5.9 (x86_64-apple-darwin23.0)
- Operating system: Mac OS 14.0.0
- Terminal emulator: iTerm.app 3.4.23
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.75.0 (82e1608df 2023-12-21) (Homebrew)
- Rust channel:  release
- Build Time: 2024-01-02 18:29:33 +00:00

```bash
eval "$(starship init zsh)"
```

#### Starship Configuration

```toml
N/A
```

alex-grover:
#### Description

Hide the Node.js prompt by default if Bun project files are detected.

#### Motivation and Context

The current default prompt shows both the Node.js and Bun prompts in Bun projects, because both runtimes use `package.json` to specify dependencies. However, in Bun projects, this file doesn't actually indicate that Node.js is used. This is a common enough use case that it was added to the docs in #5834.

It was suggested that this should be the default behavior in #4298 but was never implemented.

Closes #5805

I have not tested this but have tested the equivalent behavior via config file

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1135,7 +1135,10 @@
         "detect_files": [
           "package.json",
           ".node-version",
-          ".nvmrc"
+          ".nvmrc",
+          "!bunfig.toml",
+          "!bun.lock",
+          "!bun.lockb"
         ],
         "detect_folders": [
           "node_modules"
@@ -4737,7 +4740,10 @@
           "default": [
             "package.json",
             ".node-version",
-            ".nvmrc"
+            ".nvmrc",
+            "!bunfig.toml",
+            "!bun.lock",
+            "!bun.lockb"
           ],
           "type": "array",
           "items": {
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -645,7 +645,7 @@ By default the module will be shown if any of the following conditions are met:
 
 *: This variable can only be used as a part of a style string
 
-### Examples
+### Example
 
 #### Customize the format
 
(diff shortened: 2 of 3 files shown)


davidkna (changes requested):
LGTM, but please update the docs at `docs/config/README.md`.

alex-grover:
@davidkna done!

tharakadesilva:
Could you also include `bun.lock` here, please? This is with bun's new [text based lock file](https://bun.sh/blog/bun-lock-text-lockfile) (https://github.com/starship/starship/pull/6441)

alex-grover:
added, thanks for the suggestion @tharakadesilva!

davidkna (approved):

davidkna:
Thanks for the contribution @alex-grover!

PR #6350, opened 2024-10-24
feat: Export VIRTUAL_ENV_DISABLE_PROMPT on zsh

viccie30:
#### Description
Export VIRTUAL_ENV_DISABLE_PROMPT on zsh

#### Motivation and Context
This is already done on Fish and PowerShell. It suppresses the prefixing
of the Python virtualenv name to the prompt in Visual Studio Code's
terminal.

#### Checklist:


jankatins (approved):
LGTM (but I'm not a maintainer), I do the same in my zshrc

I just went through the rabbit hole why vscode would need this and an not-exported env var is not enough: apparently it opens a shell and in that shell runs a python process which dumps the envionment to stdout which vscode then intpretes and sets it config from. My interpretation is tat the non-exported env var does not reach the python process and therefore it doesn't show up and gets act upon by vscode.

- Where the python interpreter is spun up to print stuff: https://github.com/microsoft/vscode-python/blob/main/src/client/interpreter/activation/service.ts#L191 -> https://github.com/microsoft/vscode-python/blob/main/src/client/common/process/internal/scripts/index.ts#L74-L83
- Where the prompt is disabled based on that (PR which introduced that): https://github.com/microsoft/vscode-python/pull/22053/files 

On the other hand, fish does not set -x which seems to indicate an exported env var: https://fishshell.com/docs/current/cmds/set.html and bash doesn't even have this set. Might maybe a good idea to add that feature to all init scripts?

davidkna:
Is there an upstream issue on this? I would prefer not to export this permanently to avoid making changes to child-shells which may not have a built-in virtualenv prompt.

jankatins:
I don'T use vscode so, no idea. I  guess documenting that in certian scenarious, a user has to also `export VIRTUAL_ENV_DISABLE_PROMPT` after running the starship init is also quite reasonable...

PR #6358, opened 2024-10-29
test(username): Default to `false` for `is_root_user()` regardless of the OS

### LINKED ISSUE #6330: modules::username::tests::show_always_false fails when ran as root
#### Current Behavior

```
thread 'modules::username::tests::show_always_false' panicked at src/modules/username.rs:335:9:
  left: None
 right: Some("astronaut in ")


failures:
    modules::username::tests::show_always_false

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1117 filtered out; finished in 0.00s
```

#### Expected Behavior

https://github.com/starship/starship/blob/4ecd327fb174f518f2d9f5bad94dec60f5dce79f/src/modules/username.rs#L319-L335

https://github.com/starship/starship/blob/4ecd327fb174f518f2d9f5bad94dec60f5dce79f/src/modules/username.rs#L40-L48

When ran as root, this check is short-circuited at `is_root` being true. With `has_detected_env_var` also being true, `return None` is not executed. The code then goes on to produce two text segments, one `"astronaut"` and one `" in "`, resulting in `Some("astronaut in ")` seen above.

#### Possible Solution

Depends on whether the test or the check should be right.

#### Environment
- Starship version: git tag 1.21.1 (commit 4ecd327fb174)

Not relevant.

#### Starship Configuration
Not relevant.

Chiffario:
#### Description
Makes `is_root()` default to `false` for tests on all OSes instead of just Windows

#### Motivation and Context
First contribution with a tiny fix. Note that the [comment from davidkna](https://github.com/starship/starship/issues/6330#issuecomment-2423931927) was ignored, as running `cargo test` with elevated privileges (tested via `sudo -E cargo test`) does change the username to `root`, and also technically the name of the superuser can be changed, which is an edge case
Closes #6330

Testing on Windows is irrelevant

#### Checklist:
Updating documentation and tests seems irrelevant

### DIFF
--- a/src/modules/username.rs
+++ b/src/modules/username.rs
@@ -109,12 +109,12 @@ fn is_root_user() -> bool {
     )
 }
 
-#[cfg(all(target_os = "windows", test))]
+#[cfg(test)]
 fn is_root_user() -> bool {
     false
 }
 
-#[cfg(not(target_os = "windows"))]
+#[cfg(all(not(target_os = "windows"), not(test)))]
 fn is_root_user() -> bool {
     nix::unistd::geteuid() == nix::unistd::ROOT
 }


davidkna:
> First contribution with a tiny fix. Note that the [comment from davidkna](https://github.com/starship/starship/issues/6330#issuecomment-2423931927) was ignored, as running `cargo test` with elevated privileges (tested via `sudo -E cargo test`) does change the username to `root`, and also technically the name of the superuser can be changed, which is an edge case

That is true, but (almost) all the tests already mock the username, so this wouldn't be an issue in practice, and it would open up the possibility of adding tests for the `root` case.

davidkna (approved):

davidkna:
Thanks for the fix @Chiffario!

PR #6359, opened 2024-10-30
docs: Use inline code for Odin module name

JakeWharton:
This matches all other modules.


davidkna (approved):

davidkna:
Thanks for the fix @JakeWharton!

PR #6366, opened 2024-11-02
feat(completions): Offer Nushell completions

Kissaki:
Clap offers Nushell completions in a crate clap_complete_nushell separate from the crate clap_complete.

clap_complete::Shell and clap_complete_nushell::Nushell share the clap_complete::Generator trait.

I'm not too familiar with Rust, and all attempts to use a more direct match to common trait and call generate in one place only failed.

1. Introduce enum `CompletionShell` to unify the options for our command line across these
2. On command, match CompletionShell to their respective counterparts and call generate with them

---

The completions can be used in Nushell with:

```nu
mkdir $nu.cache-dir
starship completions nushell | save -f $'($nu.cache-dir)/starship-completions.nu'
use $'($nu.cache-dir)/starship-completions.nu' *
```

and can be set up in the env.nu and config.nu for automatic loading respectively.

---

#### Motivation and Context

Nushell can offer extensive, typed command completion, including external commands.

Starship already offers completions generation through the `completions` command, implemented through the clap library. The clap library also natively, but as a separate crate, offers to generate Nushell completions.

    *  Local `cargo build`, nutshell commands as described above, and checking that the Nushell completions show up as expected
    * `starship completions -h`
    * `starship completions nushell`
    * `starship completions bash`
    * `starship completions fish`

#### Checklist:

- [ ] ~~I have updated the documentation accordingly.~~ - I don't see any docs about completions
- [ ] ~~I have updated the tests accordingly.~~ - I don't see any tests about completions


src/main.rs
```diff
@@ -28,6 +28,44 @@ struct Cli {
...
+    )
+}
+fn generate_nushell(shell: clap_complete_nushell::Nushell) {
+    generate(
+        shell,
+        &mut Cli::command(),
+        "starship",
+        &mut io::stdout().lock(),
+    )
+}
```
davidkna, line 56: These can share a single definition:
```suggestion
fn generate_main_shell(shell: impl clap_complete::Generator) {
    generate(
        shell,
        &mut Cli::command(),
        "starship",
        &mut io::stdout().lock(),
    )
}
```

davidkna (changes requested):
Thanks for the PR. I do not like all the additional boilerplate this requires, but given the relative stability of nushell I think it's safe to offer these completions.

I have made a suggestion to reduce the boilerplate somewhat, please make the necessary changes.

Kissaki:
I combined the functions, and rebased to resolve a cargo.toml merge conflict.

I ran the completions again and it seems to generate fine still.

davidkna (approved):

davidkna:
Thanks for the contribution @Kissaki!

gfgafn:
use with nusehll [Autoload directories](https://www.nushell.sh/blog/2024-07-23-nushell_0_96_0.html#autoload-directories-for-package-managers-toc)
```nushell
let starship_autoload_dir = ($nu.vendor-autoload-dirs | filter {$in | str contains $env.USERNAME} | get 0 | path join 'starship')
mkdir $starship_autoload_dir
starship completions nushell | save --force ($starship_autoload_dir | path join 'completions.nu')
```

Kissaki:
@gfgafn thanks for the pointer. Made me fix that linked release notes (had outdated var name), and set it up for myself too.

Given the auto-load nature of those folders, I don't think there's a need to create a subfolder and to append it to the dirs list, though. Using a distinct filename like `starship-completions.nu` seems more appropriate.

For my own gen, I [only generate them once per day](https://github.com/Kissaki/nushell-config/blob/e6227cd37f51e21fc9b6504c78b580a8ee408918/env.nu#L12-L19) to speed up shell startup.

chawyehsu:
@Kissaki 
This PR introduced a break change, the shell `powershell` is renamed to `power-shell` which is an unexpected change. Please consider reverting the change as `powershell` is the conventional naming and most other clap cli tools use the name `powershell` for PowerShell completion generation.

```pwsh
error: invalid value 'powershell' for '<SHELL>'
  [possible values: bash, elvish, fish, nushell, power-shell, zsh]

  tip: a similar value exists: 'power-shell'

For more information, try '--help'.

NOTE:
    passed arguments: ["completions", "powershell"]
Invoke-Expression : Cannot bind argument to parameter 'Command' because it is an emp
ty string.
At [redacted]\Documents\WindowsPowerShell\profile.ps1:671 char:58
+ ...    (& $_ 'completions' 'powershell') | Out-String | Invoke-Expression
+                                                         ~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:PSObject) [Invoke-Expression], Paramet
   erBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationErrorEmptyStringNotAllowed
   ,Microsoft.PowerShell.Commands.InvokeExpressionCommand
```

```pwsh
# Tab-completions autoload for tools written with clap-rs
@('rustup', 'deno', 'starship', 'moonup') | ForEach-Object {
    if (Test-Command $_) {
        (& $_ 'completions' 'powershell') | Out-String | Invoke-Expression
    }
}
```

src/main.rs
```diff
@@ -28,6 +28,36 @@ struct Cli {
     command: Commands,
 }
 
+#[derive(clap::Parser, ValueEnum, Debug, Clone, PartialEq, Eq)]
+enum CompletionShell {
+    Bash,
+    Elvish,
+    Fish,
+    Nushell,
+    PowerShell,
```
chawyehsu, line 37: The variant should be named `Powershell` to prevent `-` from being added to the `<SHELL>` arg value.

Kissaki: Will you create a PR for the fix?

Kissaki: Looking at clap_complete the enum is also named `PowerShell` https://docs.rs/clap_complete/latest/clap_complete/aot/enum.Shell.html

PR #6388, opened 2024-11-11
feat(vcs): Introduce the VCS module

### LINKED ISSUE #5932: Feature/Refacto: Conflate all VCSes under an umbrella `vcs` module
## Feature Request

Context: original idea came from https://github.com/starship/starship/pull/5772#issuecomment-2019200784

A few months ago I started using [`jj`](https://github.com/martinvonz/jj) a lot. `jj` as a mode where its `.jj/` folder is colocated with the `.git/` folder. Using https://github.com/starship/starship/pull/5814 I then end up with both `git` and `jj` in my prompt when I only want the later.

I suspect this can also happen for people in *very* specific setups where they use `hg` for one mirror and `git` for another but I don't know of any real world case for that. If you have one I'm interested!

The solution I'm thinking of is the following:

```toml
[vcs] # New module

# VCS detection is attempted in this order.
# The first one returning a successful response is the one that will be used in the final output
#
# They should be in the default order from: "most likely to be used" to "least likely"
# so that most people don't pay the cost of tools they don't use
order = [
    "git",
    "hg",
    "jj",
    "pijul",
]

# The format to use for each VCS
format.git = "$git_state$git_branch$git_status"
format.hg = "$hg_branch"

[vcs.modules.git_branch]
# Same config options as the current toplevel `git_branch`

[vcs.modules.git_status]
# Again
```

Top level VCS modules would disappear and only `$vcs` would be available instead of `$git_state$git_branch` for example.

##### Advantages

- It would be possible for the `vcs` module to detect that for example both `$git_state` and `$git_branch` are used and so do one request to get all the information instead of two. For most repos it's not a big deal but it could help the `git_status` module be less costly on bigger ones maybe.
  - For `jj` for example, where using a library is not possible, it would be very helpful to do that.
- It would avoid unnecessary costs: if a user activates both `hg_branch` and `git_branch` currently, `starship` has to check for both. With `vcs`, once it finds `git` (with the order above), it stops, avoiding paying the cost for `hg`

##### Disadvantages

- It forces using only one VCS at a time in the status line: I can't think of a situation where I want two VCSes in my statusline. In the `jj` case, if I want `git` infos, I do `git` commands since it's very rare: paying the cost on every prompt seems strange
- It's a breaking change
  - That's easily solved by simply saying `vcs` uses the config from the top level modules (`git_branch` instead of `vcs.modules.git_branch) instead of having its own subconfigs. `vcs` would then end up with only `order` and `format.<vcs>` as members.
  - I don't like that solution as much because it forces us to support both use case in starship which would be annoying (caches would have to be shared to avoid duplicate costs, different VCSes would have to be tested despite `$vcs` trying to fix that, ...), especially considering it opens the option of `$vcs$git_branch` in the prompt config. We could disallow having "top level free VCSes modules" in the same prompt as `$vcs` but I think that would be a first in `starship` and not a good experience for a user trying to build their own prompt.

There is https://github.com/starship/starship/issues/818. The problem with that is that we don't have a `Repo` type for all VCSes nor even a Rust lib to use. We could "fudge" it by having our own `Repo` type for those missing ones but that's probably not a repo anymore but just a cache of relevant infos and would be confusing to document.

This issue also mentions having `$vcs_branch` instead of different `$<vcs>_branch` for each VCS: that's wrong because it implies all VCSes will have a concept of branches that map to each other but that's not the case at all: `git`, `hg`, `pijul`, `jj` and `fossil` are the VCSes I'm thinking of and they don't all have the concept of branches, sometimes it's not under that name (e.g. "channels" in `pijul`) and while branches are one of the most portable concept, lots of others are not: `change_id` vs `commit_id` in `jj`, topics in `hg`, tags from `git`

#### Working on it

I'm fully willing to do the work on all of this! This is an issue is to confirm maintainers are interested and hash out issues before I spend hours working on it :)

poliorcetics:
#### Description

As presented in #5932, `starship` supports a list of VCS that is only getting longer over time. It works-*ish* currently because of two reasons:

- Only Git is active by default
- Is it rare for two repos to be active in the same dir

Both of those are unlikely to work out well over time: new VCS tend to come with git compat when living in parallel in the same repo (Jujutsu already does it, I know Pijul plans to from discussions with the author)

This Pull Request implements an ordering (configured by the user) in which VCSes are displayed, stopping at the first one found, which neatly solve the issues and avoid double-displays (since there is one main VCS and the other is there as a backup, alternative interface or backend).

Note we don't remove the standalone modules, to still allow for the many-vcs use case.

#### Motivation and Context

Closes #5932 

#### Checklist:


davidkna:
I think we should keep `jj` modes and a general `vcs` module as separate PRs.

For `git` in particular, repo discovery is also a bit more complex (e.g. it can discover bare repos), and a final version of this should support `GIT_CEILING_DIRECTORIES` for better performance.

docs/config/README.md
```diff
@@ -4684,6 +4684,128 @@ By default the module will be shown if any of the following conditions are met:
...
+| `order`     | `["jujutsu"]` | The order in which to search VCSes. |
+| `jujutsu`   | _see below_   | The Jujutsu configuration.          |
+| `mercurial` | _see below_   | The Mercurial configuration.        |
+| `disabled`  | `true`        | Disables the `vcs` module.          |
+
+Note that an empty `order` has the same effect as disabling the module.
+
+VCS will be searched for following `order`. The first one that returns valid informations will be the one printed,
+e.g. in a repository with both `.git` and `.jj` present, `order = ["jujutsu", "git"]` will always print Jujutsu's output.
+
```
KingMob, line 4705: As a user, I'd be confused as to why there's no `git` listed in the Options table. If git can be configured here, that would be ideal, but if not, the README should tell you where to go instead.

Also, I thought the plan was not to make `["jujutsu"]` the default ordering. Should that not be `["git"]`, instead?

AltGr:
Very much interested in `jj` support; for what it's worth, here are the info that I'd like to see in my prompt:
- The latest bookmark (`jj log -r 'heads(bookmarks()::@ & bookmarks())::@' --no-graph -T bookmarks`)
- + the current change id (maybe only if I'm not exactly on the tag) (`jj log -r '@ & ~bookmarks()' --no-graph -T 'commit_id.shortest()'`)

poliorcetics:
To give an update on this: I've actively restarted work on it this weekend, I've not been happy at all with what I had previously, I'm finishing the initial implementation and will update the MR soon.

My new approach is as follows:

```toml
format = "$vcs"

[git_branch]
# config

[pijul_channel]
# config

[vcs]
order = ["git", "pijul"]
git_modules = "$git_branch$git_state"
pijul_modules = "$pijul_channel"
# Same for other VCSes:
# hg_modules = "..."
# fossil_modules = "..."
```

It keeps individual VCSes modules accessible and configurable at the top level, but when using `$vcs` in `format`, it will directly replace `$vcs` by the modules in `vcs.<found vcs>_modules`. That's it. Nothing more. Any `$<name>` is replaced by the `<name>` module (the default configuration is all modules for that VCS, but I'm not going to limit it that way, it's both easier not to and it allows for custom modules to be added in)

There is no complicated formatting or stuff like this for example, I'm not even going to add any smarts like "not rediscovering repos for each VCS module" for now, I want a very simple implementation to make it easier to review and test for (they can easily be added in a follow up MR).

In the same vein, I'm not going to add JJ support in this MR since it's currently not in starship, but I'll make an MR to build on top of this one quickly.

poliorcetics:
One thing I would like recommendations/help on from more experience Starship contributors is testing. This new `vcs` module essentially directly forward to its underlying `<vcs>_modules` so it has extremely little logic by itself (and the little bit about discovering the repo's root I made sure to reuse elsewhere to be sure it was covered by existing tests).

I don't want to copy-paste all existing VCSes tests so I'm not sure what would be a good way to test this MR.

poliorcetics:
Took the time to write test that cover everything (except for the one windows-specific line) so this MR is ready to go 🎉

KingMob:
@poliorcetics What's the current plan with respect to jj? 

I see this comment above: 
> In the same vein, I'm not going to add JJ support in this MR since it's currently not in starship, but I'll make an MR to build on top of this one quickly

but then there's also your comment https://github.com/starship/starship/issues/5772#issuecomment-2781110535:

> As per discord messages with the author, consider this MR abandoned. I'm working on https://github.com/starship/starship/pull/6388 and then I'll add jj_* modules on top of it in their own MRs

I'm confused. What's the plan?

poliorcetics:
> I'm confused. What's the plan? Do you need a contributor to add stand-alone jj support?

The plan is to wait for this MR to be merged before I do anything more, because it's already 6 months old.

After that I'll see about doing one MR per JJ module, following the pattern for git modules.

I could also do one big module that does everything but that's not the starship way I feel?

DD5HT:
@davidkna  could you take a look here?
Anyway I can help?

docs/config/README.md
```diff
@@ -4815,6 +4806,44 @@ By default the module will be shown if any of the following conditions are met:
 format = 'via [V $version](blue bold) '
 ```
 
+## VCS
+
+The `vcs` module displays the current active Version Control System (VCS).
+The module will be shown only if a configured VCS is currently in use.
```
joshka, line 4918: As a new starship user, I think it would be helpful here to see some links in this section to the other modules and some links from those modules to this section to help several use cases:
- the user that finds older configurations with explicit modules in the config
- the user that setup their system sometime before this module existed and wants to understand how to upgrade their config

Add docs something like:
> This module replaces the existing fossil, git, hg, and pijul modules that were previously included in $all with a module that finds the first matching VCS by performing the appropriate search for folders (or whatever fossil does differently), in the parent directories.
>
> The configuration of the modules is done with their configuration. See ...

joshka (approved):
This seems like it logically should do exactly what it seems to do on the description. It's well laid out, and has simple maintainable code, and has the appropriate tests for each VCS. I'd expect this would open up a path to adding jj to starship.

A few of the functions could do with adding doc comments to capture intent, but in general the names are intention revealing, so that's a fairly minor nitpick.

KingMob:
What's the next step? Is there more code to be written, or is it ready for review?

poliorcetics:
The MR has been ready for review since https://github.com/starship/starship/pull/6388#issuecomment-2781106020

KingMob:
Great! Does the review committee know? The linked "ready" comment was 3 months ago. 

Are they busy, or has the PR fallen through the cracks somehow? I see CONTRIBUTING.md says "No need to assign the pull request to anyone, we'll review it when we can". Should we just wait?

benthillerkus:
I don't think there's a review committee, the main maintainers seem to have taken a step back and there are 186 open pull requests -- I don't think this will be seen without pinging someone.

matchai (approved):
This looks good to me as it is 👍
@starship/mission-control for another pair of eyes since this is a pretty consequential change.

Note, this would need to be part of a v2 release as a breaking change.

penryu:
So excited to see movement on this!

It's working brilliantly for me, and I see that the the `order` option can gracefully ignore any unsupported modules (tested with both `jj` and a garbage value).

I saw that you want to keep `jj` work separate, and that makes sense! Is that work being tracked in a public branch by any chance?

davidkna:
@matchai In general the changes LGTM. The main issue I see with the general approach in the PR is that independent styling of vcs segments is no longer possible (e.g. moving just branch names into `right_format`).

If the changes to the root format string are reverted, I feel like this could also be merged without having to consider it as a breaking change to allow for more experimentation prior to an eventual v2 (that said this PR would be another good opportunity to do a v2).

 For a v2-style change any disabled-by-default vcs modules should become enabled-by-default to avoid confusion.

joshka:
> @matchai In general the changes LGTM. The main issue I see with the general approach in the PR is that independent styling of vcs segments is no longer possible (e.g. moving just branch names into `right_format`).

I wonder if it might be possible to solve this problem orthogonal to this PR by introducing named override versions of each module. I.e. by making the modules support being configured as `[vcs.foo]`, `[vcs.bar]` etc. with a similar approach to how `custom.foo` works.

matchai:
@davidkna I hadn't considered how it would impact the ability to separate modules contained within VCS. That definitely is a configuration trade-off as-is.

As @joshka suggests, the way that `custom` works can be a fix. By default, `$vcs` is in the format string, but it can be broken up and used independently as `${vcs.branch} ${vcs.state}`.

I'm open to keeping vcs disabled-by-default before a v2 release, as we iron this out.

poliorcetics:
Reverted the default config changes and added a documentation note to say why the module is not part of the default config as well as note it may change in the future.

I don't use right-aligned prompt at all and the current module (once I work on a Jujutsu follow-up) will handle my needs entirely but if merged and someone wants to take up the work, feel free to ask me questions :)

KingMob:
Is this ready to be merged? Or is there still more to be done?

Looking forward to this landing!

davidkna (approved):

SalTor:
Would have been great if the JJ vcs module could have been introduced on its own and then this larger initiative of unifying all of them came second. Either way, though, great work

nicolaric:
are there any updates when this is planned to be merged?

johnstegeman:
Commenting here to add my support. I'm currently using a custom module for jujutsu (and for git so that i can ensure only one set of modules actually shows anything in the case that I'm in a colocated repo). Works just fine until i want the vcs stuff on the right side of a 2-line prompt (using the fill module), as [fill doesn't play nice with custom modules I've found ](https://github.com/starship/starship/discussions/7007)

djamaile:
Is there a particular reason this isn't being merged?

KKKZOZ:
any updates?

davidkna:
Thanks for the contribution @poliorcetics, and thanks for the reviews @matchai, @joshka and @KingMob!

PR #6476, opened 2025-01-06
perf: use `gitoxide` for `git_status` and `git_metrics` modules

Byron:
I have been waiting for a long time to be able to start working on this, but here we are.

The promise of doing so is that `status` will run only once, instead of twice, and using `gix` means it's 35% (on Linux Kernel) to 200% (on WebKit) faster as measured on MacOS.

This means that in theory, less then half the work will be performed which should be good for very noticable performance improvements.

#### Tasks

* [x] ability to share state between `git_status` and `git_metrics`
* [x] fetch `status` information just once between `git_metrics` and `git_status`.
    - [x] ahead/behind via CLI invocation
* [x] assure sparse repositories don't fail (but use the previous implementation)
* [x] assure timeout implementation doesn't yield partial results.
    - [x] make iterator check interrupts by itself
* [x] fix CI on Windows
    - it was `core.autocrlf=true` being picked up, and also an incorrectly configured filter
* [x] update `gix` to the most recent released version (due on the 22nd of January)
* [x] special-case for `bare` repositories (now it shows everything as missing)

#### Performance Impact

* **2x** on WebKit
* **2x** on the Linux kernel
* **up to 6.75x** faster in the Rust repository with lots of huge submodules
    - The trick is that `gitoxide` also has dirty-checking so it can abort early if it finds something
* **3x** on the LLVM repository with all files staged as added, 33.5 million line additions
* **1.75x** on the LLVM repository with just one line change
* **1.6x** on GitLab

#### Notes for the Reviewer

* Ahead/Behind information isn't implemented in `gix` yet and I don't want to rush it, so that's still done with a `git` CLI invocation.
  It's expected to be fast though, no matter how big the repository is, and I consider running it in the background.
* Can we validate or assure that a basic call to `get_static_repo_status()` will always be suitable as basis for `git_metrics`? Maybe it's just logical that `gix_status` is the foundation for metrics?
* The WSL handling is still not applied to `git_metrics` even though it seems it could benefit just like `git_status`. However, if `git_metrics` runs before `git_status`, `git_status` can now re-use the status part if it needs it.
* As a notable change that isn't captured in tests right now: **submodules aren't shown as changed 'lines' in `git_metrics` anymore**. I think that's a feature. However, if it is not, it's easy to add it back.
* Related issue: https://github.com/GitoxideLabs/gitoxide/issues/1771 - thus far I didn't spend time on figuring out what the problem is.

#### Motivation and Context

Performance - or quite practically, the ceiling at which one will get timeouts will go up, so larger repositories can yield useful information with the default timeouts.

n/a

#### Checklist:

- [ ] I have updated the tests accordingly. *n/a*
 
#### CI Failure research

Configuration on the CI system:

```
[src\modules\git_status.rs:251:5] gix_repo.config_snapshot() = [diff "astextplain"]
	textconv = astextplain
[filter "lfs"]
	clean = git-lfs clean -- %f
	smudge = git-lfs smudge -- %f
	process = git-lfs filter-process
	required = true
[http]
	sslBackend = openssl
	sslCAInfo = C:/Program Files/Git/mingw64/etc/ssl/certs/ca-bundle.crt
[core]
	autocrlf = true
	fscache = true
	symlinks = true
[pull]
	rebase = false
[credential]
	helper = manager
[credential "https://dev.azure.com/"]
	useHttpPath = true
[init]
	defaultBranch = master
[safe]
	directory = *
[core]
	repositoryformatversion = 0
	filemode = false
	bare = false
	logallrefupdates = true
	ignorecase = true
[user]
	email = starship@example.com
	name = starship
```

I could reproduce this locally, and there were two problems:

* `gix_status` failed because `gix` didn't read all the configuration, so it didn't know that files can only be compared after doing some processing
* `gix_metrics` failed as it `gix` would convert the object in the index to one with CRLF as it picks up the configuration, but git instead prints: `warning: in the working copy of 'the_file', LF will be replaced by CRLF the next time Git touches it`, while showing the correct result.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -294,12 +294,13 @@ impl<'a> Context<'a> {
                 let mut git_open_opts_map =
                     git_sec::trust::Mapping::<gix::open::Options>::default();
 
-                // don't use the global git configs
+                // Load all the configuration as it affects aspects of the
+                // `git_status` and `git_metrics` modules.
                 let config = gix::open::permissions::Config {
-                    git_binary: false,
-                    system: false,
-                    git: false,
-                    user: false,
+                    git_binary: true,
+                    system: true,
+                    git: true,
+                    user: true,
                     env: true,
                     includes: true,
                 };
--- a/src/context.rs
+++ b/src/context.rs
@@ -671,7 +671,7 @@ impl Repo {
     pub fn exec_git<T: AsRef<OsStr> + Debug>(
         &self,
         context: &Context,
-        git_args: &[T],
+        git_args: impl IntoIterator<Item = T>,
     ) -> Option<CommandOutput> {
         let mut command = create_command("git").ok()?;
 
--- a/src/context.rs
+++ b/src/context.rs
@@ -653,7 +653,7 @@ pub struct Repo {
 
     /// Contains `true` if the value of `core.fsmonitor` is set to `true`.
     /// If not `true`, `fsmonitor` is explicitly disabled in git commands.
-    fs_monitor_value_is_true: bool,
+    pub(crate) fs_monitor_value_is_true: bool,
 
     // Kind of repository, work tree or bare
     pub kind: Kind,
--- a/src/modules/git_metrics.rs
+++ b/src/modules/git_metrics.rs
@@ -50,7 +50,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         }
         let status_module = context.new_module("git_status");
         let status_config = GitStatusConfig::try_load(status_module.config);
-        let status = super::git_status::get_static_repo_status(context, &repo, &status_config)?;
+        let status = super::git_status::get_static_repo_status(context, repo, &status_config)?;
         let gix_repo = gix_repo.with_object_memory();
         gix_repo.write_blob([]).ok()?; /* create empty blob */
         let tree_index_cache = prevent_external_diff(
@@ -274,6 +274,7 @@ fn prevent_external_diff(mut cache: gix::diff::blob::Platform) -> gix::diff::blo
     cache
 }
 
+#[allow(clippy::too_many_arguments)]
 fn diff_two_opt(
     lhs_location: &BStr,
     lhs_id: gix::ObjectId,
--- a/src/modules/git_metrics.rs
+++ b/src/modules/git_metrics.rs
@@ -1,11 +1,15 @@
+use gix::bstr::{BStr, ByteSlice};
+use gix::diff::blob::pipeline::WorktreeRoots;
+use gix::diff::blob::ResourceKind;
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 use regex::Regex;
 
+use super::Context;
+use crate::configs::git_status::GitStatusConfig;
 use crate::{
     config::ModuleConfig, configs::git_metrics::GitMetricsConfig, formatter::StringFormatter,
     formatter::string_formatter::StringFormatterError, module::Module,
 };
-
-use super::Context;
 
 /// Creates a module with the current added/deleted lines in the git repository at the
 /// current directory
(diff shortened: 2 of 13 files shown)


jankatins:
Just FYI: a few years back starship was changed from the git2 library to using git directly because git2 had(/has?) abyssal runtime characteristics in case of some large git repos (see https://github.com/starship/starship/pull/2465 and all the related issues around that).

Is that not a problem anymore with gitoxide?

Byron:
Thanks for asking, that's fair!

Here is the preliminary results using the WebKit repository with 407k files. Note that only the `git_status` module is implemented right now, and when done, the performance of `git_metrics` will improve greatly to the point where it probably is free.

```
WebKit ( main) +1 [!?] via △ took 4s
❯ time /Users/byron/dev/github.com/starship/starship/target/release/starship timings

 git_status   -  2396ms  -   "[!?] "
 git_metrics  -  2173ms  -   "+1 "
 directory    -     9ms  -   "WebKit "
 pulumi       -     4ms  -   ""
 username     -     1ms  -   ""
 git_branch   -    <1ms  -   "( main) "
 cmake        -    <1ms  -   "via △ "
 line_break   -    <1ms  -   "\n"
 character    -    <1ms  -   "❯ "
/Users/byron/dev/github.com/starship/starship/target/release/starship timings  0.97s user 22.49s system 507% cpu 4.621 total

WebKit ( main) +1 [!?] via △ took 4s
❯ time starship timings

 git_status   -  3099ms  -   "[!?] "
 git_metrics  -  2087ms  -   "+1 "
 directory    -     2ms  -   "WebKit "
 pulumi       -     1ms  -   ""
 git_branch   -    <1ms  -   "( main) "
 cmake        -    <1ms  -   "via △ "
 line_break   -    <1ms  -   "\n"
 character    -    <1ms  -   "❯ "
starship timings  0.61s user 32.63s system 638% cpu 5.205 total
```

Here we have a 29% speedup compared to Git (Apple M1 Pro). On more powerful machines, the speedup will be more pronounced - I have measured a 2.8x speedup on an M4 Pro Max with maxed out CPU.

It's going to be good :).

davidkna:
Thanks for the PR!
For the Windows issue, it looks like something in the global config which is currently not read is causing the issue. This PR with https://github.com/davidkna/starship/commit/e71e8f2c376c0a768eebbea6043df92183f91845 passes CI.

Byron:
You are very welcome - I was so looking forward to this one (also as I can't wait to use it myself :D).

Thanks so much for figuring out the Windows issue, I was puzzled (and still am), but this narrows the possibilities.

Byron:
OMG! 5x!

```
❯ time /Users/byron/dev/github.com/starship/starship/target/release/starship timings

 git_metrics  -  1094ms  -   "+1 -154 "
 directory    -     7ms  -   "WebKit "
 username     -     1ms  -   ""
 git_branch   -     1ms  -   "( main) "
 git_status   -    <1ms  -   "[✘»!?] "
 cmake        -    <1ms  -   "via △ "
 line_break   -    <1ms  -   "\n"
 character    -    <1ms  -   "❯ "
/Users/byron/dev/github.com/starship/starship/target/release/starship timings  0.58s user 2.83s system 298% cpu 1.143 total

WebKit ( main) +1 -154 [✘»!?] via △
❯ time starship timings

 git_status   -  3878ms  -   "[✘»!?] "
 git_metrics  -  1958ms  -   "+1 -154 "
 directory    -    <1ms  -   "WebKit "
 git_branch   -    <1ms  -   "( main) "
 cmake        -    <1ms  -   "via △ "
 line_break   -    <1ms  -   "\n"
 character    -    <1ms  -   "❯ "
starship timings  0.66s user 50.77s system 879% cpu 5.848 total
```


With `hyperfine` it's just 2x.

```
❯ hyperfine -N -w1 '/Users/byron/dev/github.com/starship/starship/target/release/starship timings' 'starship timings'
Benchmark 1: /Users/byron/dev/github.com/starship/starship/target/release/starship timings
  Time (mean ± σ):      1.530 s ±  0.045 s    [User: 0.579 s, System: 4.075 s]
  Range (min … max):    1.438 s …  1.589 s    10 runs

Benchmark 2: starship timings
  Time (mean ± σ):      3.140 s ±  0.077 s    [User: 0.558 s, System: 13.973 s]
  Range (min … max):    3.044 s …  3.296 s    10 runs

Summary
  /Users/byron/dev/github.com/starship/starship/target/release/starship timings ran
    2.05 ± 0.08 times faster than starship timings
```

Byron:
@davidkna Tomorrow I'd hope I can figure out CI, it's probably something about line endings and/or transformations. With the spellcheck failure I am a bit puzzled on how to configure it.

Otherwise I am very, very happy with the performance impact of this PR - the difference can be felt and it's snappier. Even WebKit is now within the realm of usable at 1.6s, which on Linux could probably be much faster.

@pascalkuthe maybe this PR is something you'd want to try out - now `starship` is reaching light-speed also thanks to your tech 🚀.!

Byron:
I could reproduce the issue locally now. The problem is with `core.autocrlf` and index-worktree content checks seemingly going for a simple size comparison to consider it changed, even though it has to consider certain config flags and potentially do a full content check after running it through a conversion pipeline. This sounds expensive, so it will take a moment until that is figured out, but at least it seems to be a legitimate bug in `gitoxide`.

Byron:
And finally, Windows works! The issues were very legitimate, and could all be fixed through minor tweaks here:

* read all configuration, just like Git would.
    - Note that this also invokes the Git executable (if present) to get its 'hidden' configuration location
* use the correct filter mode when producing line-stats.
    - Git obviously converts *to Git* for diffs, whereas I thought I'd convert to a diffable worktree version. This is fine if the worktree version already has the correct filters applied, but if not one would compare the properly converted version from Git to one that isn't correct in the worktree.

@davidkna Could you help me with the spellchecker?

davidkna:
@Byron I've pushed a commit that should fix the spellchecker issue.

Byron:
Thanks so much! In my sleepiness I accidentally force-pushed over your commit, but I recreated it with "Co-authored-by" hoping that's OK.

With a little bit of luck nightly won't fail anymore with an ICE so CI can be green for the first time. Besides changes due to the review, the PR would then be ready but has to wait until the 22nd or so when a new version of the `gix` crate will be released.

Byron:
@davidkna One more note regarding a notable, but untested change: submodule changes aren't displayed by `git_metrics` anymore, as it only shows line changes now, not "one line per submodule change" like before. I think that's a feature, but since it's a change compared to before I thought I'd point it out so you can decide what it should do. Restoring the old behaviour is no problem.

Byron:
I am now setting this PR as ready for review as it is ready technically.

While the inconsistency in performance isn't fixed, it might be considerably slower sometimes, but also should generally be faster if `git_status` and `git_metrics` are used together. When there are submodules with worktree changes, it will be considerably faster.

Something worth considering would be a feature toggle or some setting that makes `gitoxide` opt-out initially to get coverage. Should problems mount, a patch release could be published to make it opt-in.
Depending on the stance, the initial setting of such a flag could of course be swapped as well

davidkna:
This PR has mostly been an improvement for repos I actively work on, so I think that changing the default would be fine.

Since the original implementation is being kept either way for now, I think that adding an opt-out is warranted (with a warning that it could be removed in the future). Beyond that given the performance regressions I think that `core.fsmonitor` should also trigger an automatic fallback to git.

Byron:
That is a great call and I am planning to implement the `core.fsmonitor` auto-fallback as well, along with an opt-out. Progress might be slower than I would want, but each weekend is a new chance to finally get to it.

Byron:
@davidkna I have finally gotten around to implementing the `core.fsmonitor` handling. The opt-out is also there but I wouldn't know where to document it (i.e. the warning that it could be removed in the future).

davidkna (commented):
Thanks for the update. The new module option and the warning should be added to the module documentation in `docs/config/README.md`. I have also made two suggestions to make use of `repo.fs_monitor_value_is_true` (sorry for not mentioning that in my earlier comment).

Please also update the config schema via `cargo run --features config-schema -- config-schema > .github/config-schema.json` if possible.

Byron:
Sorry for the long silence, just as a quick update: this PR isn't forgotten, I just have a very limited amount of time each day and most of it is taken merely by catching up on the more numerous emails. But I will get to it - apologies for the inconvenience.

src/modules/git_metrics.rs
```diff
@@ -20,14 +24,224 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     };
 
     let repo = context.get_repo().ok()?;
-    let mut git_args = vec!["diff", "--shortstat"];
-    if config.ignore_submodules {
-        git_args.push("--ignore-submodules");
-    }
+    let gix_repo = repo.open();
+    // TODO: remove this special case once `gitoxide` can handle sparse indices for tree-index comparisons.
+    let stats = if gix_repo.index_or_empty().ok()?.is_sparse() {
```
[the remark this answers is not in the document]
Byron: A great catch!

src/context.rs
```diff
@@ -652,7 +653,7 @@ pub struct Repo {
 
     /// Contains `true` if the value of `core.fsmonitor` is set to `true`.
     /// If not `true`, `fsmonitor` is explicitly disabled in git commands.
-    fs_monitor_value_is_true: bool,
+    pub(crate) fs_monitor_value_is_true: bool,
```
Byron, line 656: This was a quick-fix to allow using the field, I hope it's alright.

Byron (commented):
@davidkna If CI passes, this PR should be ready for another review.
Even though I rebased and fixed merge conflicts, the changes pertaining the review are in the most recent commit.
Apologies again for the long delay.

src/modules/git_status.rs
```diff
@@ -219,40 +245,224 @@ fn get_repo_status(
...
+                Submodule::AsConfigured { check_dirty }
+            })
+            .index_worktree_options_mut(|opts| {
+                // TODO: figure out good defaults for other platforms, maybe make it configurable.
+                //       Git uses everything (if repo-size permits), but that's not the best choice for MacOS.
+                opts.thread_limit = if cfg!(target_os = "macos") {
+                    Some(3)
+                } else {
+                    None
+                };
```
davidkna, line 325: I don't have my Windows system at hand right now, but I assume that Windows will also benefit from the lower thread limit. For other operating systems, this should probably be limited to [8 or `STARSHIP_NUM_THREADS`](https://github.com/starship/starship/blob/master/src/main.rs#L284-L290) like the global thread pool to avoid issues on systems with high core counts.

src/context.rs
davidkna: It's completely fine.

davidkna (commented):
Thanks for the update! 
The PR mostly looks good to me now.

Byron:
Thanks a lot for the review, and sorry for the sloppy markdown table. A typical copy-paste error.

The thread-count is now pinned to whatever amount of threads rayon would use at most, even though I kept the MacOS specific hardcoded value if the user didn't override it.

davidkna (approved):

Byron:
I'd appreciate if this PR could be merged after #6700 was merged.

Byron:
@matchai I have added a commit with an upgrade to the latest (now working) release of `gix`, and hope this PR can be merged with it.

matchai:
Thank you so much for all your work getting gitoxide into Starship and @davidkna for the review! 🎉
Really looking forward to seeing these performance gains!

slang25:
This PR looks really cool. I've updated to the latest version of starship, however I still see:
> [WARN] - (starship::utils): Executing command "/opt/homebrew/bin/git" timed out.

I thought maybe `git` is still being used for `git_branch`, but when I switch that off I still see this. Is my expectation wrong?

Byron:
Thanks for bringing this up!

There are a couple of conditions that would make it use the binary:

* it's enabled via configuration
* the index is sparse
* the git filesystem monitor is configured

https://github.com/starship/starship/blob/daf8d93d27fdc70b110a4a56799d7dc3a10c3810/src/modules/git_status.rs#L252-L254

Maybe one of these conditions is true for you?

A good test is to run `starship timings` - if the Git binary is used, the `git_metrics` and `git_status` module will consume about the same time, whereas otherwise one of them will basically be free.

PR #6502, opened 2025-01-13
feat(package): support JSR package version

canac:
#### Description

Add support for extracting JSR package version from `jsr.json` or `deno.json`. The [docs](https://jsr.io/docs/publishing-packages#package-config-file) indicate that `jsr.json` is the preferred location of the `version` string, but `deno.json` is also supported for deno users.

I added test cases for all new code and tested the prompt in a local JSR project.

#### Checklist:


canac:
I realized that `deno.json` can also be a `deno.jsonc` file. If it is acceptable to add [`jsonc-parser`](https://crates.io/crates/jsonc-parser) as a new dependency, I can add support for extracting the version from that file as well.

src/modules/package.rs
```diff
@@ -68,6 +68,23 @@ fn get_node_package_version(context: &Context, config: &PackageConfig) -> Option
...
+        .read_file_from_pwd("jsr.json")
+        .and_then(|contents| json::from_str(&contents).unwrap_or_default())
+        .or_else(|| json::from_str(&context.read_file_from_pwd("deno.json")?).unwrap_or_default())
+        .or_else(|| {
+            jsonc_parser::parse_to_serde_value(
+                &context.read_file_from_pwd("deno.jsonc")?,
+                &Default::default(),
+            )
+            .unwrap_or_default()
+        })?;
```
davidkna, line 82: This should probably cover `jsr.jsonc`. I would also prefer if this code use`.find()` with an iterator over a list of the filenames.

docs/config/README.md
```diff
@@ -3426,6 +3426,7 @@ package, and shows its current version. The module currently supports `npm`, `ni
 
 - [**npm**](https://docs.npmjs.com/cli/commands/npm) – The `npm` package version is extracted from the `package.json` present
   in the current directory
+- [**JSR**](https://jsr.io/) – The `jsr` package version is extracted from the `jsr.json` or `deno.json` present in the current directory
```
davidkna, line 3429: This should mention the jsonc files.

src/modules/package.rs
```diff
@@ -68,6 +68,22 @@ fn get_node_package_version(context: &Context, config: &PackageConfig) -> Option
     Some(formatted_version)
 }
 
+fn get_jsr_package_version(context: &Context, config: &PackageConfig) -> Option<String> {
+    ["jsr.json", "jsr.jsonc", "deno.json", "deno.jsonc"]
+        .iter()
+        .find_map(|file| {
+            let contents = context.read_file_from_pwd(file)?;
```
davidkna, line 75: It's likely best to stop after finding the first file, whether it parses correctly or not:
```suggestion
        .find_map(|file| context.read_file_from_pwd(file))
        .and_then(|contents| {
```

canac: Done. In quick testing, JSR seems to proiritize `deno.json(c)` files over `jsr.json(c)` files, so I changed it to look for deno config files before JSR config files.

davidkna (approved):

davidkna:
Thanks for the contribution @canac!

PR #6523, opened 2025-01-26
feat(python): add `uv` binary support for version

RoloEdits:
#### Description

Adds the ability to get the python version via `uv`. As this uses `uv`, it also follows its fallbacks when funding a python executable to run: https://docs.astral.sh/uv/concepts/python-versions/#finding-a-python-executable

#### Motivation and Context

If you have no "system" install of python, and are only using `uv` to manage installations and virtual environments, the current way to lookup the version wont work.

image

#### Checklist:

I'm not sure how this should be tested, as its a bit different to a normal python installation and management. Also not sure how to reword the documentation for the same reason.


src/configs/python.rs
```diff
@@ -29,7 +29,7 @@ impl Default for PythonConfig<'_> {
         PythonConfig {
             pyenv_version_name: false,
             pyenv_prefix: "pyenv ",
-            python_binary: VecOr(vec!["python", "python3", "python2"]),
+            python_binary: VecOr(vec!["python", "python3", "python2", "uv"]),
```
jankatins, line 32: I really dislike this change: there is a difference between using python and using a package manager like uv to run python programms. At least I would be very surprised if I would get uvs python shown when I have not activated the venv where uv put's it stuff.

Maybe it would make more sense to add this in the virtualenv part or add a new uv module. In any case, I guess doing the same for poetry also makes sense, but on the other hand it might already be too late for poetry...

RoloEdits: > At least I would be very surprised if I would get uvs python shown when I have not activated the venv where uv put's it stuff.

Even if not activated, does it really matter if it shows a version there or not? Not sure I would find it quite so surprising. What are the worries here?

> there is a difference between using python and using a package manager like uv to run python programs.

As far as I can tell `python_binary` is only used to get the version. Nothing else. It also says as such in the documentation:
image

davidkna: I am also concerned that this would run any potentially untrusted binaries without interaction which are located in `.venv/bin/python`.

RoloEdits: Let me ask the question in the `uv` project on how it handles potentially untrusted `python` binaries. As far as I know, running `uv run python --version` runs whatever python executable it can find first, with a `.venv\Scripts\python.exe`(on windows) being fairly early on.

This is all I managed to find in the docs:
> In some cases, uv allows using a Python version from a virtual environment. In this case, the virtual environment's interpreter will be checked for compatibility with the request before searching for an installation as described above. See the [pip-compatible virtual environment discovery](https://docs.astral.sh/uv/concepts/python-versions/#discovery-of-python-versions../../pip/environments/#discovery-of-python-environments) documentation for details.
> 
> When performing discovery, non-executable files will be ignored. Each discovered executable is queried for metadata to ensure it meets the [requested Python version](https://docs.astral.sh/uv/concepts/python-versions/#discovery-of-python-versions#requesting-a-version). If the query fails, the executable will be skipped. If the executable satisfies the request, it is used without inspecting additional executables.

RoloEdits: Given that `.venv` should not be checked into source control, any binary that should be called should be one that the user themselves installed?

RoloEdits: But `uv` isn't just limited to just the `.venv`, you can run it just like a system wide installation of python. `uv run python --version` would then return that version.

davidkna: I would prefer to neither encourage nor default to the use of `uv` here, for the reasons outlined. How about changing the `python_binary` type to `VecOr<VecOr<&'a str>>` to allow listing binaries with additional arguments instead?

RoloEdits: > I would prefer to neither encourage nor default to the use of uv here, for the reasons outlined.

I'm still not sure what the exact issues are. `starship` is something that runs without user input already, for example. Of course it would not take user interaction. And `uv` is more than just a package manger, it can also be interfaced to run python. So there are different contexts that it could be run in, but as we only are trying to get the version, I don't think there needs to be a worry about the semantics here. And if the user had a malicious binary that's unknown to them and tries to run it( agsin with not knowing), the result would be the same if `starship` ran it for them anyways. The attack vector didnt change any. And if they do know, then I would imagine they would be a professional and know to sandbox the environment.

> How about changing the `python_binary` type to `VecOr<VecOr<&'a str>>` to allow listing binaries with additional arguments instead?

I can do this, but the binary is only used to get the version, so not sure what exposing that solves here that would be different from hard coding it?

RoloEdits: The other option was as @jankatins suggested and only running `uv` if there is a `uv.lock` in the root. In my use case this would fit all instances where all projects are managed by `uv`, thus having a `uv.lock`.

davidkna: @RoloEdits Without this option, no binary from the cwd is run without input, unless a virtual environment is explicitly activated. Whether an `uv.lock` exists is irrelevant to the issue at hand, given the lack of any protections in uv itself. `VecOr<VecOr<&'a str>>` would also have the advantage of not being specific to uv. E.g. on Windows `C:/` is generally writable by all users and may contain a uv.lock file and a hostile virtual environment, and there is also the issue of untrusted git repository downloads or clones.

RoloEdits: Just so I am clear are you looking for:
```toml
python_binary = ["python", "python3", "python2", ["uv", "run", "python",  "--version"]]
```

With only the current three pythons being included in the default, and the user can add `uv` if they wish?

And also, perhaps a safer way, rather than offloading any risk to the user and washing hands of it here, what about querying a `.python-version` file, if one exists, instead? This would fit all use cases I have as well, and if someone is using `uv` explicitly, is included in the default `init`.

So, cycle through the python binaries as usual, and then if none are found, read the `.python-version` file if one exists:
```text
3.13
```

And if there is nothing, then just fallback to current behavior with no version displayed?

davidkna: I would prefer `python_binary = ["python", "python3", "python2", ["uv", "run", "python", "--no-python-downloads"]]`. Any binary that is launched should still match the expected interface of a python binary.

As for parsing `.python-version`, I am not sure how well-specified that file is, and it is likely to lead to less specific version strings.

RoloEdits: I should be well specified. `pyenv` uses it too.

RoloEdits: > I would prefer` python_binary = ["python", "python3", "python2", ["uv", "run", "python", "--no-python-downloads"]]`.

Also just to be clear, this would still run the `venv` binary. Making it configurable still leaves the same concerns as you had before. Is it just that the user would have to manually undertake the risk that makes this better?

Although a `venv` should always be created and never reused, even when just moving its location on disk(and never checked into source control), should it be documented that adding this can leave you more suspectable to a malicious `venv`?

RoloEdits: As for the `.python-version` files sometimes less specific version, perhaps adding a config option for it could be in order, with documentation that it can sometimes exclude the patch version? That its more of a "show something rather than nothing"  best effort kind of fallback? It would be "safer" if you still want something shown but dont like auto running an environment you didnt setup yourself. I can also just remove this functionality and just go forward with the binary idea. 

The binary idea would also allow other python management users, like for `poetry`, the ability to add version support.

RoloEdits: Also, instead of the nested `[]` we could also have "uv run --no-python-downloads python" and then in the code split on the first space. Which way would you prefer?

RoloEdits: I removed the fallback for the `.python-version` file. This is simple enough to add if its wanted later, but for now its just the ability to provide arguments to a binary of which `--version` will be passed.

RoloEdits:
I will say, I did find `VecOr` a bit confusing as the expectation is the struct would have been more like:

```rust
enum VecOr<T> {
    Vec(Vec<T>),
    Or(T),
}
```

and not a:
```rust
struct VecOr<T>(pub Vec<T>)
```

Just a note for potential documentation changes on `VecOr` as initially I was confused with no way to create a `VecOr` with just a single instance.

src/modules/python.rs
```diff
@@ -93,14 +93,24 @@ fn get_pyenv_version(context: &Context) -> Option<String> {
...
+    config
         .python_binary
         .0
         .iter()
-        .find_map(|binary| context.exec_cmd(binary, &["--version"]))
-        .map(get_command_string_output)?;
-
-    parse_python_version(&version)
+        .find_map(|binary| {
+            let command = binary.0[0];
```
davidkna, line 101:
```suggestion
            let command = binary.0.get(0)?;
```

davidkna: Please also update the config schema via `cargo run --features config-schema -- config-schema > .github/config-schema.json`.

davidkna (approved):
LGTM

davidkna:
Thanks for the contribution @RoloEdits! By the way, I noticed that uv puts version information into pyenv.cfg that could be used without having to execute a binary for displaying venv-version information.

PR #6526, opened 2025-01-27
feat(python): add option to use the parent directory as venv name

RFCreate:
#### Description

This PR adds the option of using the parent directory name when venv is activated instead of using the name of the venv folder (when prompt is empty).

#### Motivation and Context

Most of the time the venv folder name is `.venv` or `venv` which when activated does not give valuable information when displaying the `virtualenv` name because all of the virtual environments will have the same name regardless of the project. 

venv

Creating virtual environments with a standard name (.venv), activating them, and seeing all display the same information.

#### Checklist:

If I [understand correctly](https://github.com/starship/starship/blob/master/CONTRIBUTING.md#documentation), locale docs should not be modified.


.github/config-schema.json
```diff
@@ -1452,6 +1452,7 @@
         ],
         "style": "yellow bold",
         "symbol": "🐍 ",
+        "use_venv_parent_name": false,
```
jankatins, line 1455: It would be nice if that is clearer if it replaces the logic or is a fallback. So either "prefer_venv_parent_name" or "fallback_to_venv_parent_name" or so...

jankatins:
I like this as a fallback when the venv name is not set. At least with a now common `<repo-root>/.venv/`  placement, it makes more sense to got for the repo root. I wonder if it would make even more sense to not have it a config but to say "if venv name is `.venv` or `venv`, get the parent name. That would cover both "venv is in a central location and the name is important" and ".venv is in repo root and parent name is important". and it wouldn't even need a config :-)

RFCreate:
How about doing an array like "override_venv_names": ['venv, '.venv'] _(empty by default)_ ?

So that the names can be customizable.

andytom:
Looks good to me but need to fix the conflicts.

RFCreate:
Is this ready? @andytom @jankatins

PR #6528, opened 2025-01-27
feat(c++): add module for C++ language

siddharthkumarjha:
feat(cpp): add module for cpp language.

#### Description

copy c module's impl and make necessary cosmetic changes to support cpp instead.

#### Motivation and Context

Closes #

I didn't really receive any replies here https://github.com/starship/starship/issues/6525 
maybe I was too long winded?

To detect cpp files in a dir.
since `c` is not `cpp`.

- [ x] I have tested using **Linux**

changed the binary in my install dir with the one containing these changes. noticed both c and c++ files being nicely detected. 
Even starship.toml is properly formatted.

#### Checklist:

- [x ] I have updated the tests accordingly.

I have updated the tests for the `cpp` module as per the `c` module.
I just have issue with 

```
cargo test
```
stalling at git tests.
```
...
test utils::tests::exec_slow_command ... ok
test modules::git_commit::tests::test_latest_tag_shown_with_tag_enabled has been running for over 60 seconds
test modules::git_commit::tests::test_latest_tag_shown_with_tag_enabled_lightweight has been running for over 60 seconds
```

Want to know what exactly am I supposed to do to move these test cases?

Stuck here both with and without my changes.

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -241,6 +241,53 @@
         }
       ]
     },
+    "cpp": {
+      "default": {
+        "commands": [
+          [
+            "c++",
+            "--version"
+          ],
+          [
+            "g++",
+            "--version"
+          ],
+          [
+            "clang++",
+            "--version"
+          ]
+        ],
+        "detect_extensions": [
+          "cc",
+          "cp",
+          "cxx",
+          "cpp",
+          "CPP",
+          "c++",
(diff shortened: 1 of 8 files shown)


siddharthkumarjha:
This is obviously not complete yet. Since test cases keep stalling at git tests on my machine.
Need help with that.

PR #6552, opened 2025-02-09
Fix: Issue #6549

kiramidru:
Fix: Issue #6549 - changing space in the OS icons after the format, and removing the spaces from the table

Removing the space in the OS icons allows users to just set format = '[$symbol]($style)' and not touch the table.


davidkna:
closing in favor of #6499

PR #6570, opened 2025-02-16
feat(c++): adding c++ module

### LINKED ISSUE #4583: Add a module for C++

siddharthkumarjha:
hello, @davidkna 
Take a look at this PR.
This is for #4583  Implementing a module for C++.

Sorry it took so long.

#### Description

Implement a c and c++ modules with a common implementation.

#### Motivation and Context

Closes #4583

2025-02-16T13:52:57,430623543+05:30

#### Checklist:

### DIFF
--- a/src/configs/cc.rs
+++ b/src/configs/cc.rs
@@ -1,0 +1,24 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Clone)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields),
+    schemars(bound = "CcConfig<'a, T>: std::default::Default, T: schemars::JsonSchema")
+)]
+#[serde(default)]
+pub struct CcConfig<'a, T> {
+    #[serde(skip)]
+    pub marker: std::marker::PhantomData<T>,
+
+    pub format: &'a str,
+    pub version_format: &'a str,
+    pub style: &'a str,
+    pub symbol: &'a str,
+    pub disabled: bool,
+    pub detect_extensions: Vec<&'a str>,
+    pub detect_files: Vec<&'a str>,
+    pub detect_folders: Vec<&'a str>,
+    pub commands: Vec<Vec<&'a str>>,
+}
--- a/src/configs/cpp.rs
+++ b/src/configs/cpp.rs
@@ -1,0 +1,34 @@
+use crate::configs::cc::CcConfig;
+use serde::{Deserialize, Serialize};
+
+#[derive(Default, Clone, Deserialize, Serialize)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+pub struct CppConfigMarker;
+
+pub type CppConfig<'a> = CcConfig<'a, CppConfigMarker>;
+
+impl Default for CppConfig<'_> {
+    fn default() -> Self {
+        CcConfig {
+            marker: std::marker::PhantomData::<CppConfigMarker>,
+
+            format: "via [$symbol($version(-$name) )]($style)",
+            version_format: "v${raw}",
+            style: "149 bold",
+            symbol: "C++ ",
+            disabled: true,
+            detect_extensions: vec!["cpp", "cc", "cxx", "c++", "hpp", "hh", "hxx", "h++", "tcc"],
+            detect_files: vec![],
+            detect_folders: vec![],
+            commands: vec![
+                vec!["c++", "--version"],
+                vec!["g++", "--version"],
+                vec!["clang++", "--version"],
+            ],
+        }
+    }
+}
--- a/src/modules/cc.rs
+++ b/src/modules/cc.rs
@@ -1,0 +1,130 @@
+use super::{Context, Module, ModuleConfig};
+
+use crate::configs::c::{CConfig, CConfigMarker};
+use crate::configs::cc::CcConfig;
+use crate::configs::cpp::{CppConfig, CppConfigMarker};
+use crate::formatter::StringFormatter;
+use crate::formatter::VersionFormatter;
+
+use crate::segment::Segment;
+
+use semver::Version;
+use std::borrow::Cow;
+use std::ops::Deref;
+use std::sync::LazyLock;
+
+pub enum Lang {
+    C,
+    Cpp,
+}
+
+fn is_cc_project<T>(config: &CcConfig<T>, context: &Context) -> Option<bool> {
+    Some(
+        context
+            .try_begin_scan()?
+            .set_extensions(&config.detect_extensions)
+            .set_files(&config.detect_files)
+            .set_folders(&config.detect_folders)
+            .is_match(),
+    )
+}
+
+fn parse_module<T>(
+    context: &Context,
+    module: &mut Module,
+    config: CcConfig<T>,
+    compilers: [(&str, &str); 2],
+) -> Result<Vec<Segment>, crate::formatter::string_formatter::StringFormatterError> {
+    StringFormatter::new(config.format).and_then(|formatter| {
+        let cc_compiler_info = LazyLock::new(|| context.exec_cmds_return_first(config.commands));
+
+        formatter
+            .map_meta(|var, _| match var {
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "name" => {
+                    let cc_compiler_info = &cc_compiler_info.deref().as_ref()?.stdout;
+
+                    let cc_compiler =
+                        compilers
+                            .iter()
+                            .find_map(|(compiler_name, compiler_hint)| {
+                                cc_compiler_info
+                                    .contains(compiler_hint)
+                                    .then_some(*compiler_name)
+                            })?;
+                    Some(Ok(Cow::Borrowed(cc_compiler)))
+                }
+                "version" => {
+                    let cc_compiler_info = &cc_compiler_info.deref().as_ref()?.stdout;
+
+                    VersionFormatter::format_module_version(
+                        module.get_name(),
+                        cc_compiler_info
+                            .split_whitespace()
+                            .find(|word| Version::parse(word).is_ok())?,
+                        config.version_format,
+                    )
+                    .map(Cow::Owned)
+                    .map(Ok)
+                }
+                _ => None,
+            })
+            .parse(None, Some(context))
+    })
+}
+
+fn create_module<'a, T>(
+    context: &'a Context,
+    lang: &str,
+    compilers: [(&str, &str); 2],
+    mut module: Module<'a>,
+    config: CcConfig<T>,
+) -> Option<Module<'a>> {
+    if config.disabled {
+        return None;
+    }
+
+    if !is_cc_project(&config, context)? {
+        return None;
+    }
+
+    let parsed = parse_module(context, &mut module, config, compilers);
+
+    module.set_segments(match parsed {
+        Ok(segments) => segments,
+        Err(error) => {
+            log::warn!("Error in module `cc` for `lang: {}` :\n{}", lang, error);
+            return None;
+        }
+    });
+
+    Some(module)
+}
+
+pub fn module<'a>(context: &'a Context, lang: Lang) -> Option<Module<'a>> {
+    match lang {
+        Lang::C => {
+            let lang = "c";
+            let compilers = [("clang", "clang"), ("gcc", "Free Software Foundation")];
+            let module = context.new_module(lang);
+            let config = CConfig::try_load(module.config);
+
+            create_module::<CConfigMarker>(context, lang, compilers, module, config)
+        }
+        Lang::Cpp => {
+            let lang = "cpp";
+            let compilers = [("clang++", "clang"), ("g++", "Free Software Foundation")];
+            let module = context.new_module(lang);
+            let config = CppConfig::try_load(module.config);
+
+            create_module::<CppConfigMarker>(context, lang, compilers, module, config)
+        }
+    }
+}
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -134,7 +134,7 @@
       },
       "allOf": [
         {
-          "$ref": "#/definitions/CConfig"
+          "$ref": "#/definitions/CcConfig_for_CConfigMarker"
         }
       ]
     },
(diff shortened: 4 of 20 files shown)


src/configs/cpp.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub format: &'a str,
+    pub version_format: &'a str,
+    pub style: &'a str,
+    pub symbol: &'a str,
+    pub disabled: bool,
+    pub detect_extensions: Vec<&'a str>,
+    pub detect_files: Vec<&'a str>,
+    pub detect_folders: Vec<&'a str>,
+    pub commands: Vec<Vec<&'a str>>,
+}
```
davidkna, line 20: What do you think about handling it like that?
```suggestion
#[serde(transparent)]
pub struct CppConfig<'a>(CConfig<'a>)
```

siddharthkumarjha: I tried implementing it the way you suggested but I am facing lifetime related issues in deserialize,

impl:

```rust
use serde::{Deserialize, Serialize};
use crate::configs::c::CConfig;

#[derive(Clone, Deserialize, Serialize)]
#[cfg_attr(
    feature = "config-schema",
    derive(schemars::JsonSchema),
    schemars(deny_unknown_fields)
)]
#[serde(transparent)]
pub struct CppConfig<'a>(CConfig<'a>);

impl Default for CppConfig<'_> {
    fn default() -> Self {
        CppConfig(CConfig {
            format: "via [$symbol($version(-$name) )]($style)",
            version_format: "v${raw}",
            style: "149 bold",
            symbol: "C++ ",
            disabled: false,
            detect_extensions: vec![
                "cc", "cp", "cxx", "cpp", "CPP", "c++", "C", "hh", "H", "hp", "hxx", "hpp", "HPP",
                "h++", "tcc",
            ],
            detect_files: vec![],
            detect_folders: vec![],
            commands: vec![
                vec!["c++", "--version"],
                vec!["g++", "--version"],
                vec!["clang++", "--version"],
            ],
        })
    }
}
```

error:

```console
error: lifetime may not live long enough
  --> src/configs/cpp.rs:4:17
   |
4  | #[derive(Clone, Deserialize, Serialize)]
   |                 ^^^^^^^^^^^
   |                 |
   |                 lifetime `'de` defined here
   |                 associated function was supposed to return data with lifetime `'a` but it is returning data with lifetime `'de`
...
11 | pub struct CppConfig<'a>(CConfig<'a>);
   |                      -- lifetime `'a` defined here
   |
   = help: consider adding the following bound: `'de: 'a`
   = note: this error originates in the derive macro `Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I tried doing as compiler suggested. 
```rust
#[serde(transparent)]
pub struct CppConfig<'a, 'de : 'a>(CConfig<'a>);
```
but then I get the error 
```console
cannot deserialize when there is a lifetime parameter called 'de [macro-error]
rustc: cannot deserialize when there is a lifetime parameter called 'de
```

using,
```console
cargo expand configs::cpp --lib
```
I can see the error likely happens due to,
```rust
    #[doc(hidden)]
    #[allow(non_upper_case_globals, unused_attributes, unused_qualifications)]
    const _: () = {
        #[allow(unused_extern_crates, clippy::useless_attribute)]
        extern crate serde as _serde;
        #[automatically_derived]
        impl<'de, 'a> _serde::Deserialize<'de> for CppConfig<'a> {
            fn deserialize<__D>(
                __deserializer: __D,
            ) -> _serde::__private::Result<Self, __D::Error>
            where
                __D: _serde::Deserializer<'de>,
            {
                _serde::__private::Result::map(
                    _serde::Deserialize::deserialize(__deserializer),
                    |__transparent| CppConfig { 0: __transparent },
                )
            }
        }
    };
```

any suggestions? I am kind of stuck here...

siddharthkumarjha: I think I might have resolved it?

```rust
use serde::{Deserialize, Serialize, Deserializer};
use crate::configs::c::CConfig;

#[derive(Clone, Serialize)]
#[cfg_attr(
    feature = "config-schema",
    derive(schemars::JsonSchema),
    schemars(deny_unknown_fields)
)]
#[serde(transparent)]
pub struct CppConfig<'a>(CConfig<'a>);

impl<'de> Deserialize<'de> for CppConfig<'de>
{
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
            Result::map(
                Deserialize::deserialize(deserializer),
                |transparent| CppConfig { 0: transparent },
            )
    }
}

impl Default for CppConfig<'_> {
    fn default() -> Self {
        CppConfig(CConfig {
            format: "via [$symbol($version(-$name) )]($style)",
            version_format: "v${raw}",
            style: "149 bold",
            symbol: "C++ ",
            disabled: false,
            detect_extensions: vec![
                "cc", "cp", "cxx", "cpp", "CPP", "c++", "C", "hh", "H", "hp", "hxx", "hpp", "HPP",
                "h++", "tcc",
            ],
            detect_files: vec![],
            detect_folders: vec![],
            commands: vec![
                vec!["c++", "--version"],
                vec!["g++", "--version"],
                vec!["clang++", "--version"],
            ],
        })
    }
}
```

siddharthkumarjha: never mind it came back for 
```rust
    #[serde(borrow)]
    cpp: cpp::CppConfig<'a>,
```
in FullConfig.

siddharthkumarjha: finally figured out...
```
#[derive(Clone, Deserialize, Serialize)]
#[cfg_attr(
    feature = "config-schema",
    derive(schemars::JsonSchema),
    schemars(deny_unknown_fields)
)]
#[serde(transparent)]
pub struct CppConfig<'a>(#[serde(borrow)] CConfig<'a>);
```

just had to do serde borrow.
sorry for spamming... very new to rust.

siddharthkumarjha: @davidkna 

please take a look again. I have made the suggested changes.

siddharthkumarjha: @davidkna 
I have also updated the docs in the meantime... Pls take a look at them as well.

siddharthkumarjha:
@davidkna 
can you please take a look at my PR? I have impl the suggested changes + updated the docs. Pls tell me if there is something else I should change or if my approach is not aligned with what you were expecting for this module?

src/configs/cc.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn symbol(&self) -> &'a str;
+    fn disabled(&self) -> bool;
+    fn detect_extensions(&self) -> &Vec<&'a str>;
+    fn detect_files(&self) -> &Vec<&'a str>;
+    fn detect_folders(&self) -> &Vec<&'a str>;
+    fn commands(&self) -> &Vec<Vec<&'a str>>;
+}
+
+#[macro_export]
+macro_rules! impl_lang_config {
```
davidkna, line 14: This enum is no longer necessary with a shared base config.

src/modules/cc.rs
```diff
@@ -0,0 +1,133 @@
...
+use std::borrow::Cow;
+use std::ops::Deref;
+use std::sync::LazyLock;
+
+pub enum Lang {
+    C,
+    Cpp,
+}
+
+trait CommonConfig<'a, E>: ModuleConfig<'a, E> + LangConfig<'a>
```
davidkna, line 23: Given the common config struct, the module can use the regular `module` function approach.

src/configs/cpp.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn default() -> Self {
+        CppConfig(CConfig {
+            format: "via [$symbol($version(-$name) )]($style)",
+            version_format: "v${raw}",
+            style: "149 bold",
+            symbol: "C++ ",
+            disabled: false,
+            detect_extensions: vec![
+                "cc", "cp", "cxx", "cpp", "CPP", "c++", "C", "hh", "H", "hp", "hxx", "hpp", "HPP",
+                "h++", "tcc",
```
davidkna, line 34: Some of these extensions appear to be very unusual.

src/configs/cpp.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+impl Default for CppConfig<'_> {
+    fn default() -> Self {
+        CppConfig(CConfig {
+            format: "via [$symbol($version(-$name) )]($style)",
+            version_format: "v${raw}",
+            style: "149 bold",
+            symbol: "C++ ",
+            disabled: false,
```
davidkna, line 31: Given the potential overlaps, this module should be disabled-by-default or ensure (`!`-patterns) that they are mutually exclusive.

src/configs/cpp.rs
siddharthkumarjha: Regarding these extensions they are simply all the recognised c++ extensions as per gcc's manual. 

Should I trim this down to just the widely used extensions?

src/configs/cc.rs
siddharthkumarjha: removed

src/modules/cc.rs
siddharthkumarjha: I didn't properly understand this point.

are you asking me to make a CommonConfig struct out of this trait and use it somehow??

I have removed it and updated the code right now.

src/modules/cc.rs
```diff
@@ -0,0 +1,125 @@
...
+                    let cc_compiler_info = &cc_compiler_info.deref().as_ref()?.stdout;
+
+                    let cc_compiler = 'search: {
+                        for (compiler_name, compiler_hint) in compilers {
+                            if cc_compiler_info.contains(compiler_hint) {
+                                break 'search Some(compiler_name);
+                            }
+                        }
+                        None
+                    };
```
davidkna, line 60: I would prefer `find_map` or a combination of `find` and `map` here.
E.g. (untested):
```suggestion
                    let cc_compiler = compilers.iter().find_map(|(compiler_name, compiler_hint)| {
                        cc_compiler_info.contains(compiler_hint).then_some(compiler_name)
                    };
```

src/configs/cpp.rs
davidkna: I think that's a good idea since this amount could break the table in the docs, but if you feel strongly about leaving it as-is should be fine.

src/modules/cc.rs
davidkna: I am satisfied with how it turned out now.

src/configs/cpp.rs
siddharthkumarjha: I have updated this to most commonly used c++ extensions

src/modules/cc.rs
siddharthkumarjha: done

siddharthkumarjha: done

src/configs/cpp.rs
siddharthkumarjha: made it disabled by default

siddharthkumarjha:
@davidkna 
I have a pretty weird thing happening to me.

When I try to do

```
    #[test]
    fn folder_without_cpp_files() -> io::Result<()> {

        let actual = ModuleRenderer::new("cpp")
            .config(toml::toml! {
                [cpp]
                    disabled = true
            })
            .path(dir.path()).collect();

        let expected = None;
        assert_eq!(expected, actual);
        dir.close()
    }
```

I get the following config in my cc.rs module, (I added a print there)

```
[DEBUG] - (starship::config): Config found for "cpp": Some(Table({"disabled": Boolean(true)}))
{
  "format": "via [$symbol($version(-$name) )]($style)",
  "version_format": "v${raw}",
  "style": "149 bold",
  "symbol": "C ",
  "disabled": true,
  "detect_extensions": [
    "c",
    "h"
  ],
  "detect_files": [],
  "detect_folders": [],
  "commands": [
    [
      "cc",
      "--version"
    ],
    [
      "gcc",
      "--version"
    ],
    [
      "clang",
      "--version"
    ]
  ]
}
```
But when I change this to 
```
[c]
disabled = false
```
This changes the default cpp config instead...

And this bizzare thing happens only with unit tests not with actual toml file...

Do you have any idea why this kind of swapping between configs might be happening??

davidkna:
It's possible that serde has issues handling nested struct defaults. Furthermore, disabled-by-default modules require explicit `if config.disable { return None; }`.

siddharthkumarjha:
> It's possible that serde has issues handling nested struct defaults. Furthermore, disabled-by-default modules require explicit `if config.disable { return None; }`.

yes I saw that in other modules. was trying to implement the same but this config parsing issue popped up for unit tests...

siddharthkumarjha:
@davidkna  I am ready for another round of review :) 
- I have been able to fix the toml cmd not working in test cases. It was indeed serde.
- Added the disable check in module.

Just a smol problem the fix I came up with meant manually writing JsonSchema impl. Any suggestions on this?

siddharthkumarjha:
@davidkna whenever you have time, please review the changes. 
I ended up going the PhantomData route because of the following issue:

```rust
#[serde(transparent)]
pub struct CConfig<'a>(#[serde(borrow)] pub CcConfig<'a, CConfig<'a>>);
```

The goal was to create a "common config" that works for both C and C++.

However, `#[serde(default)]` became problematic:

Initially, I applied `#[serde(default)]` on the `CcConfig` struct, but this meant Serde will always load the default of the inner struct (CcConfig) during deserialization.

I figured it might help to also apply `#[serde(default)]` on the outer struct (`CConfig` and `CppConfig`). This led to a strange behavior where:

- Serde loaded the default for the inner struct.
- Then it applied the test case values.
- Finally, it again applied the outer struct’s default, overwriting the test case values.
Atleast, I think it did so?

I then tried removing the default from the inner struct altogether — but bizarrely, this still didn’t work. Serde correctly used the outer default, but didn’t override the fields with the test values (e.g., `disabled = false`).

At this point, I thought: why not just let the inner struct load the defaults (since that part worked)? So, I introduced the PhantomData trick to make Rust treat `CcConfig<CConfig>` and `CcConfig<CppConfig>` as two different types.

This approach worked. But now `schemars::JsonSchema` is unhappy — because there’s no “default” generic version of `CcConfig`. The two specialized implementations (`CcConfig<'a, crate::configs::c::CConfig<'a>>` and `CcConfig<'a, crate::configs::c::CppConfig<'a>>`) conflict when deriving the schema.

So, I manually implemented `JsonSchema` for just these two specializations.

Wanted your thoughts on this ?

davidkna:
The general approach seems fine, but I managed to avoid the manual schema trait implementation via 04a580e4723dc0421c8e99b1d4ee515334cf51fd, and to further reduce boilerplate the newtypes can be replaced with a type alias (b9ac704ab1e4fd203daa79f834df29aad610763c).

siddharthkumarjha:
@davidkna ,
Thanks a lot :). I have integrated your changes as is without modifications. Didn't really have anything to add from my side.

Any other comments for my patch @davidkna ? Anything I should fix?

davidkna (commented):
Please remove `disable = false` from the presets, but keep the sections themselves, otherwise LGTM.

siddharthkumarjha:
@davidkna done. removed them from presets

davidkna (approved):

siddharthkumarjha:
Hello @davidkna ,
Wanted to ask how big is the review period for the project? 

I know projects leave pr open for a while for everyone to review it before it gets merged (tock handles this with "last-call" label). 

Wanted to ask what's the window for starship?

davidkna:
I usually merge PRs after around a week.

siddharthkumarjha:
> I usually merge PRs after around a week.

Then would I be a candidate for consideration now. Since its been 2 weeks since last review. 🙂

Let me know if I need to do something like rebase my changes on top of master or similar.

davidkna:
Thanks for the contribution @siddharthkumarjha!

siddharthkumarjha:
> Thanks for the contribution @siddharthkumarjha!

Thankyou for helping me with the PR @davidkna. Thankyou for so actively maintaining this awesome project as well.

PR #6578, opened 2025-02-18
feat: Change java model default symbol to nerd-font

cncsl:
The original symbol is not nerd-font icon, therefore the display of the Java module is different from that of other modules. To avoid this problem, we can replace the symbol with nerd-font icon.

#### Description

Change model `java`'s default config, replace the original symbol with nerd-font. Change doc and test causes.

#### Motivation and Context

Many other model has a default nerd-font symbol, they looks beautiful, so I think java model can do the same thing.

image

#### Checklist:


davidkna:
Thanks for the contribution, but starship is moving away from using nerd font symbols by default (https://github.com/starship/starship/issues/2563, https://github.com/starship/starship/pull/3544).

Thus, I am closing this. If you prefer to use nerd fonts, you might want to use the [nerd font preset](https://starship.rs/presets/nerd-font).

cncsl:
> nerd font preset

Sorry for unnecessary PR, waiting for the moving process to complete🫡
And for someone who want to use nerd-font symbol in `java` model early, this configuration maybe help:

```toml
[java]
symbol = " "
```

PR #5170, opened 2023-05-05
feat(aws): support aws sso with automatic authentication refresh

kensasongko:
#### Description
Currently, starship only supports AWS SSO with [legacy non-refreshable configuration](https://docs.aws.amazon.com/cli/latest/userguide/sso-configure-profile-legacy.html).

This PR adds a support for the newer and recommended configuration for AWS CLI to authenticate AWS SSO with [automatic authentication refresh](https://docs.aws.amazon.com/cli/latest/userguide/sso-configure-profile-token.html).

Example of the config file with refreshable token (taken from their [documentation](https://docs.aws.amazon.com/cli/latest/userguide/sso-configure-profile-token.html#sso-configure-profile-token-auto-sso)):
```
[profile my-dev-profile]
sso_session = my-sso
sso_account_id = 123456789011
sso_role_name = readOnly
region = us-west-2
output = json

[sso-session my-sso]
sso_region = us-east-1
sso_start_url = https://my-sso-portal.awsapps.com/start
sso_registration_scopes = sso:account:access
```
This PR also renames the test function name for the old SSO configuration to `sso_legacy_set()`

#### How Has This Been Tested?
- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:
- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #6614, opened 2025-03-08
feat(aws): support duration for aws sso

raylu:
#### Description

show time until expiration of credentials from `aws sso`

#### Motivation and Context

#5170 added support for showing profiles from `aws sso`
but `duration` still only works for credentials in `~/.aws/credentials`

`starship.toml`:
```toml
[aws]
format = '[($profile )($duration)]'
```
`~/.aws/config`:
```ini
[default]
sso_start_url = https://[redacted].awsapps.com/start
sso_region = us-east-1
```
`~/.aws/sso/cache/[redacted].json`:
```json
{"startUrl": "https://[redacted].awsapps.com/start", "region": "us-east-1", "accessToken": "[redacted]", "expiresAt": "2025-03-08T05:32:14Z", "clientId": "[redacted]", "clientSecret": "[redacted]", "registrationExpiresAt": "2025-05-29T17:37:08Z"}
```
```
$ target/debug/starship module aws
44m14s
```

#### Checklist:

I don't know if there is any documentation to update. Let me know if you'd like anything reworded
Some tests assume the default config with `$duration` in the format and started failing on my machine with active SSO creds, so I made them stop showing the duration.
I didn't add a test because that would mean writing into `~/.aws/sso/cache`, which seems not ideal. Happy to do it if you prefer.

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -1,10 +1,12 @@
+use std::cell::OnceCell;
 use std::collections::HashMap;
 use std::path::PathBuf;
 use std::str::FromStr;
 
 use chrono::DateTime;
 use ini::Ini;
-use std::cell::OnceCell;
+use serde_json as json;
+use sha1::{Digest, Sha1};
 
 use super::{Context, Module, ModuleConfig};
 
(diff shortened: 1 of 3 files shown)


davidkna:
@raylu The changes look fine, but I would prefer having a test for this. You should be able to add a test by setting `HOME` via `ModuleRenderer::env` like in the pulumi module. You could also use this to avoid changing the other tests, in general helper function that returns a `ModuleRenderer` with `HOME` set to a temporary diretory or `/dev/null` / `nul` could be useful.

raylu:
is that what you had in mind?

davidkna (approved):
LGTM

PR #6639, opened 2025-03-20
fix(nodejs): improve Node.js version extraction from polluted output

liby:
#### Description

This PR improves the Node.js version extraction logic to handle version manager installation logs that pollute the output of `node --version`. Specifically:

1. Added a robust version extraction algorithm that:
   - Prioritizes the last line of output (standard location for version info)
   - Falls back to searching lines from bottom to top for version numbers
   - Handles various edge cases like mixed logs with multiple version numbers

2. Added comprehensive test cases for various scenarios:
   - Clean version output
   - Polluted installation logs (e.g., from Proto version manager)
   - Multiple version numbers in output
   - Complex cases with version numbers not in the last line

#### Motivation and Context

Closes #6632

When using Node.js version managers like [proto](https://moonrepo.dev/proto) with auto-install enabled, running `node --version` can output installation logs along with the version number. This causes Starship to display these logs in the prompt instead of just the version number, resulting in a cluttered and incorrect prompt.

Added new test cases that verify:
1. Standard version output handling
2. Handling of installation logs from Proto and similar tools
3. Complex scenarios with multiple version numbers in the output
4. Version extraction when the version isn't in the last line

All tests pass without breaking existing functionality.

#### Checklist:


davidkna:
This seems fairly complex and I am not sure if all node versions conform wholly to this regex, taking a whole line that starts with the expected format may be simpler.
Please also open a feature request for proto to use stderr which seems more appropriate in this case.

liby:
> This seems fairly complex and I am not sure if all node versions conform wholly to this regex, taking a whole line that starts with the expected format may be simpler. 

Sorry, I almost forgot about this PR. I think I should close it now because the author of the proto has helped to handle it on the proto side, so there's no need to do anything on the Starship side.

> Please also open a feature request for proto to use stderr which seems more appropriate in this case.

Sounds good, current proto handling method hides this information when passing `--help` or `--version`.

PR #6649, opened 2025-03-23
fix: remove double spaces in default values

acid-bong:
Command run: `for file in $(rg --hidden '".  "' -l); do sed -i '/".  "/s/  ",/ ",/' $file; done`

#### Motivation and Context

Double spaces between blocks look weird and inconsistent (cuz everything else is separated by single spaces).

**Before**:
- Emacs (Vterm)
изображение

- Alacritty
изображение

**After**:
- Emacs
изображение

- Alacritty
изображение

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1232,7 +1232,7 @@
         "disabled": false,
         "format": "via [$symbol($version )]($style)",
         "style": "bold blue",
-        "symbol": "🪖  ",
+        "symbol": "🪖 ",
         "version_format": "v${raw}"
       },
       "allOf": [
@@ -1307,7 +1307,7 @@
           "Ultramarine": "🔷 ",
           "Unknown": "❓ ",
           "Uos": "🐲 ",
-          "Void": "  ",
+          "Void": " ",
           "Windows": "🪟 ",
           "openEuler": "🦉 ",
           "openSUSE": "🦎 "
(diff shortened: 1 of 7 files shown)


davidkna (approved):

acid-bong:
Ight, one little problem here: while this does help in Emacs, Alacritty and Kitty, in st it's not so good:
изображение

davidkna:
As far as I can tell from the diff the nix symbol wasn't changed in this PR. Multi-character emoji use double spaces for this reason.

acid-bong:
So double spaces are desired?

davidkna:
The changes in this particular PR seem fine.

acid-bong:
Yup, these one were indeed kinda off
изображение
VVV
изображение

Will undraft then

anotherhadi:
I'm not sure if it's related, but I sometimes have a one-character issue with my prompt. My cursor is shifted one character to the left when I'm in a Nix shell (so if I delete one char, it doesn't delete the one on the left of the cursor but the one on the right)
It doesn't happen every time

injust:
@acid-bong Any idea why the `aws` section was changed in `nerd-font-symbols.toml` but not the `gcloud` section?

acid-bong:
Hmm, `sed` ignored it for some reason 🤔

PR #6656, opened 2025-03-25
feat(package): pyproject hatchling dynamic version

lucascolley:
#### Description
TODO

#### Motivation and Context
https://github.com/starship/starship/issues/3949#issuecomment-2734804195

#### Checklist:


lucascolley:
I have left a few TO-DOs in the code, and we could definitely add more extensive tests, but I would like a maintainer's opinion on the changes so far as this covers my use-case!

src/modules/package.rs
```diff
@@ -87,25 +88,71 @@ fn get_jsr_package_version(context: &Context, config: &PackageConfig) -> Option<
...
+fn parse_file_version(context: &Context, path: &str) -> Option<String> {
+    let file_contents = read_file(context.current_dir.join(path)).ok()?;
+    let re = Regex::new(r#"(__version__|VERSION)\s*=\s*["']([^"']+)["']"#).ok()?;
+    Some(
+        re.captures(&file_contents)
+            .and_then(|cap| cap.get(2))?
+            .as_str()
+            .to_owned(),
+    )
+}
```
lucascolley, line 109: based on my limited Rust knowledge, it seems that we need to return a `String` rather than an `&str` here otherwise there would be a dangling reference to the file contents which is local to this function. The other nearby functions are modified to return `Option<String>` rather than `Option<&str>` accordingly.

src/modules/package.rs
```diff
@@ -87,25 +88,71 @@ fn get_jsr_package_version(context: &Context, config: &PackageConfig) -> Option<
...
+
+fn parse_hatchling_dynamic_version(context: &Context, pyproject: &toml::Table) -> Option<String> {
+    let version_path = pyproject
+        .get("tool")?
+        .get("hatch")?
+        .get("version")?
+        .get("path")?
+        .as_str()?;
+    // TODO: custom regex pattern
```
lucascolley, line 118: https://hatch.pypa.io/1.9/version/#configuration

src/modules/package.rs
```diff
@@ -87,25 +88,71 @@ fn get_jsr_package_version(context: &Context, config: &PackageConfig) -> Option<
...
         .get("version")?
         .as_str()
+        .map(|s| s.to_owned())
 }
 
-fn get_pep621_version(pyproject: &toml::Table) -> Option<&str> {
-    pyproject.get("project")?.get("version")?.as_str()
+fn parse_file_version(context: &Context, path: &str) -> Option<String> {
+    let file_contents = read_file(context.current_dir.join(path)).ok()?;
+    let re = Regex::new(r#"(__version__|VERSION)\s*=\s*["']([^"']+)["']"#).ok()?;
```
lucascolley, line 102: https://hatch.pypa.io/1.9/version/#configuration

src/modules/package.rs
```diff
@@ -87,25 +88,71 @@ fn get_jsr_package_version(context: &Context, config: &PackageConfig) -> Option<
...
+        .get("version")?
+        .get("path")?
+        .as_str()?;
+    // TODO: custom regex pattern
+    parse_file_version(context, version_path)
+}
+
+fn parse_pep621_dynamic_version(context: &Context, pyproject: &toml::Table) -> Option<String> {
+    // TODO: Flit, PDM, Setuptools
+    parse_hatchling_dynamic_version(context, pyproject)
```
lucascolley, line 124: https://packaging.python.org/en/latest/discussions/single-source-version/#build-system-version-handling

src/modules/package.rs
```diff
@@ -87,25 +88,71 @@ fn get_jsr_package_version(context: &Context, config: &PackageConfig) -> Option<
...
+fn get_pep621_version(context: &Context, pyproject: &toml::Table) -> Option<String> {
+    get_pep621_static_version(pyproject).or_else(|| get_pep621_dynamic_version(context, pyproject))
 }
 
 fn get_pyproject_version(context: &Context, config: &PackageConfig) -> Option<String> {
     let file_contents = context.read_file_from_pwd("pyproject.toml")?;
     let pyproject_toml: toml::Table = toml::from_str(&file_contents).ok()?;
 
-    get_pep621_version(&pyproject_toml)
+    get_pep621_version(context, &pyproject_toml)
```
lucascolley, line 153: passing through the context allows us to use `context.current_dir` to access files referenced by `project.toml` above

lucascolley, line 153: passing through the context allows us to use `context.current_dir` to access files referenced by `pyproject.toml` above

src/modules/package.rs
```diff
@@ -1528,10 +1619,16 @@ Title: Starship
...
-        file_name: &str,
+        file_path: &str,
         contents: Option<&str>,
     ) -> io::Result<()> {
-        let mut file = File::create(project_dir.path().join(file_name))?;
+        let mut file = {
+            let path = project_dir.path().join(file_path);
+            if let Some(parent) = path.parent() {
+                fs::create_dir_all(parent)?;
+            }
```
lucascolley, line 1629: the codecov warning seems erroneous, these lines are used where we pass `"src/__init__.py"` to `file_path` in `test_extract_pep621_dynamic_version_hatchling_1`, as can be observed by commenting out these lines to make the test fail.

lucascolley (commented):
self-review to aid reviewers

lucascolley:
hey @davidkna @andytom ! Is there anything I can do to push this forwards?

davidkna:
There are still some TODO comments, I gather that you already consider this reviewable?

lucascolley:
> There are still some TODO comments, I gather that you already consider this reviewable?

Yes. However, if you require the TODO comments to go before a review, I can accept that also :)

src/modules/package.rs
lucascolley: this could be a follow-up PR as the regex option seems rather niche

src/modules/package.rs
lucascolley: other build backends could also be separate PRs

lucascolley:
but tbh I think the TODOs are best left to separate PRs

src/modules/package.rs
```diff
@@ -87,25 +88,71 @@ fn get_jsr_package_version(context: &Context, config: &PackageConfig) -> Option<
...
         .get("tool")?
         .get("poetry")?
         .get("version")?
         .as_str()
+        .map(|s| s.to_owned())
 }
 
-fn get_pep621_version(pyproject: &toml::Table) -> Option<&str> {
-    pyproject.get("project")?.get("version")?.as_str()
+fn parse_file_version(context: &Context, path: &str) -> Option<String> {
```
davidkna: This module is not python-specific so the function naming should reflect that.

src/modules/package.rs
davidkna: Please reference relevant documentation in comments as appropriate .

src/modules/package.rs
davidkna: Since this is parsing Python files please validate any returned value via `semver::Version::parse`

lucascolley: I opted for https://docs.rs/pep440_rs/latest/pep440_rs/index.html instead as the `semver` crate seemed undesirably restrictive on allowed syntax. Let me know if that is okay!

lucascolley:
thanks for the review @davidkna ! I checked and this is working with my use-case:

```console
array-api-extra on 🎋 main via 🐍 
❯ /Users/lucascolley/ghq/github.com/starship/starship/target/debug/starship prompt

array-api-extra on 🎋 main is 📦 v0.9.0.dev0 via 🐍 
❯ 
```

lucascolley:
CI failure looks unrelated to this PR!

src/modules/package.rs
davidkna: Wait I picked the wrong crate from our dependencies, `versions::Version::new` (or `versions::Mess::new`) should be flexible enough to handle everything.

lucascolley: > Wait I picked the wrong crate from our dependencies, `versions::Version::new` (or `versions::Mess::new`) should be flexible enough to handle everything.

changed to `versions::Version::new`!

davidkna: You changed it to `Versioning` which may end up falling back to `Mess`; I would prefer avoiding going through multiple fallbacks during validation (e.g. just using `Mess` straight away). Since `Mess` looks very loose, I feel like `Version` is a sensible choice to ensure a higher level of sanity, though I am not sure how messy python version strings get in practice.

lucascolley: let's give `Version` a go!

src/modules/package.rs
```diff
@@ -88,25 +90,75 @@ fn get_jsr_package_version(context: &Context, config: &PackageConfig) -> Option<
...
+    let version_path = pyproject
+        .get("tool")?
+        .get("hatch")?
+        .get("version")?
+        .get("path")?
+        .as_str()?;
+
+    parse_file_version_for_hatchling(context, version_path)
+        .and_then(|s| Version::new(s.as_str()))
+        .map(|v| v.to_string())
```
davidkna: Apart from that LGTM if tests pass once `master` is fixed.

```suggestion
        .filter(|s| Version::new(s.as_str()).is_some())
```

lucascolley: Don't we still need to convert from `Version` to `String` here?

davidkna: The `filter` maps the value to `None` if the it can't parse the `Version`, while not changing the original value if it can. Thus the original verified string can be returned because it is still there.

lucascolley: Aha thanks, TIL

lucascolley:
@davidkna I think this is ready!

lucascolley:
hey @davidkna ! Anything left to do here?

lucascolley:
gentle bump

davidkna:
Thanks for the contribution @lucascolley!

lucascolley:
thanks for the review!!

PR #6676, opened 2025-04-08
docs(preset/nerd-font): Add symbol for status module

real-or-random:

#### Checklist:

### DIFF
--- a/docs/public/presets/toml/nerd-font-symbols.toml
+++ b/docs/public/presets/toml/nerd-font-symbols.toml
@@ -171,6 +171,9 @@ symbol = "󱘗 "
 [scala]
 symbol = " "
 
+[status]
+symbol = " "
+
 [swift]
 symbol = " "
 


docs/public/presets/toml/nerd-font-symbols.toml
```diff
@@ -171,6 +171,9 @@ symbol = "󱘗 "
 [scala]
 symbol = " "
 
+[status]
+symbol = " "
```
davidkna, line 175: The default emoji does not have a space:
```suggestion
symbol = ""
```

davidkna: I would prefer not to use a double-width nerd font char, since it will likely render inconsistently (which renders one of the  advantages of nf symbols moot).

real-or-random: Currently, literally every symbol in https://github.com/starship/starship/blob/master/docs/public/presets/toml/nerd-font-symbols.toml has a trailing space. So this change is consistent with existing practice.

Switching to single-width chars is a different endeavor.

davidkna: The base preset also uses a space after most symbols apart from this module.

real-or-random: I've verified that every symbol currently in https://github.com/starship/starship/blob/master/docs/public/presets/toml/nerd-font-symbols.toml is a double-width char with a trailing space. My conclusion is that this PR is in line with the existing file, and it's ready for merge.


> The base preset also uses a space after most symbols apart from this module.

Okay, but I don't see how that affects this PR. I'm confused. What do you want me to do?

davidkna: Well if all the symbols are double-width I suppose this PR is fine as-is.

davidkna:
Thanks for the addition @real-or-random!

PR #6684, opened 2025-04-15
feat(nu): Add job support for nushell

Bahex:
#### Description
Add `--jobs` support for nushell.

#### Motivation and Context
As of `0.103.0`, nushell has job support.

#### Checklist:


andytom:
What would happen if the user is using an earlier version of nu shell? Do we need to bump the minimum supported version or add some guards here?

Bahex:
This should work without raising the minimum version.

davidkna:
Please also update the warning in the jobs module description (`docs/config/README.md`).

Bahex:
Which one should I go with?

```diff
- This module is not supported on tcsh and nu.
+ This module is not supported on tcsh.
```

```diff
- This module is not supported on tcsh and nu.
+ This module is not supported on tcsh and nu(before 0.103.0).
```

davidkna:
I think the mention of nu should be removed completely since all nu versions that have job-support will support the module.

davidkna (approved):

andytom:
Thank you for your contribution @Bahex and thanks for reviewing @davidkna

PR #6685, opened 2025-04-15
feat: Fortran module

### LINKED ISSUE #6680: Support Fortran
I am working on some [Fortran](https://fortran-lang.org/) code and noticed starship doesn't have a module for it!

Add a Fortran module

I wouldn't mind writing this myself but would need some pointers for all the things that would need to happen:
- [ ] write a config/Fortran.rs file, similar to all the others (i.e. base it off something I know such as c, Julia, or R).
- [ ] decide on default file extensions to detect, i.e. `["f90", "F90", ..?]`
- [ ] add a section to the config docs

minomy13:
#### Description
Added module for Fortran. Displays version of `gfortran`; decided against supporting multiple compilers because most alternatives are proprietary software.

#### Motivation and Context
There was no support for Fortran yet.
Closes #6680

Screenshot 2025-04-15 at 4 30 18 PM

Tested on macOS by executing `starship prompt` in a directory containing Fortran files. Prompt looked correct. Tested on linux by setting up Docker container and enabling Starship in bash. Tested multiple scenarios; prompt looked correct. Also created unit tests which are all passing.

#### Checklist:

### DIFF
--- a/src/modules/fortran.rs
+++ b/src/modules/fortran.rs
@@ -120,4 +120,18 @@ mod tests {
 
         dir.close()
     }
+
+    #[test]
+    fn folder_with_fpm_config_file() -> io::Result<()> {
+        let dir = tempfile::tempdir()?;
+        File::create(dir.path().join("fpm.toml"))?.sync_all()?;
+
+        let actual = ModuleRenderer::new("fortran").path(dir.path()).collect();
+
+        let expected = Some(format!("via {}", Color::Purple.bold().paint("󱈚 14.2.0 ")));
+
+        assert_eq!(expected, actual);
+
+        dir.close()
+    }
 }
--- a/src/modules/fortran.rs
+++ b/src/modules/fortran.rs
@@ -33,7 +33,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             })
             .map(|variable| match variable {
                 "version" => {
-                    let fortran_version = get_version(context, &config)?;
+                    let fortran_version = get_version(context)?;
                     VersionFormatter::format_module_version(
                         &module.get_name(),
                         &fortran_version,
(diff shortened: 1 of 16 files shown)


src/modules/fortran.rs
```diff
@@ -0,0 +1,289 @@
...
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "name" => {
+                    let compiler_info = &compiler_info.deref().as_ref()?.stdout;
+
+                    let compiler = if compiler_info.contains("GNU") {
+                        "gfortran"
+                    } else if compiler_info.contains("flang-new") || compiler_info.contains("flang")
```
davidkna, line 46:
```suggestion
                    } else if compiler_info.contains("flang")
```

davidkna (commented):
Thanks for the PR! I have made some suggestions. If you do not mind, this might benefit from using the common C/C++ module base that was added recently, but this versions should work too.

minomy13:
> If you do not mind, this might benefit from using the common C/C++ module base that was added recently

I took a look into the codebase and I definitely agree with you that this would work perfectly. That said, I was wondering, wouldn't it be semantically misleading to use the C module base for the Fortran module? Id really appreciate your perspective on this.

minomy13:
After replacing the symbol in my tests' `expected` strings, I somehow had to add two whitespaces after it for one whitespace to render. Do you have any Idea why this is and if it's intended to be that way?

minomy13:
Hey @davidkna just wanted to check in — have you had a chance to look at the PR again? Let me know if you need anything from my side. Thanks!

davidkna:
> I took a look into the codebase and I definitely agree with you that this would work perfectly. That said, I was wondering - wouldn't it be semantically misleading to use the C module base for the Fortran module? I'd really appreciate your perspective on this.

Considering that fortran and c toolchain are often coupled it's probably fine, and this could be revised in the future.

The PR looks good to me, switching to a c module base could also be handled later if you would prefer not to delay this any further. That said please rebase the PR and regenerate the config-schema, because the schema generator lib has been upgraded.

src/modules/fortran.rs
davidkna: `flang` should match both `flang-new` and `flang`, but keeping should be fine because there is no functional difference.

lstngr:
Hi both! And thank you @minomy13 for the nice PR, looking forward to see this merged!

May I suggest using the `nf-dev-fortran` icon  (U+e7de) which is visually closer to the C/C++ ones due to the “cutout” letter?

minomy13:
I've made the requested changes :)

davidkna (approved):

davidkna:
Thanks for the contribution, and for persisting despite all the delays @minomy13!

PR #6689, opened 2025-04-20
fix(preset/pure): only enable python module if venv is active

pantheraleo-7:
#### Description

~This PR, inspired by the `no-empty-icons` preset, wraps the venv name and the trailing whitespace in a *conditional* format string. So if the venv name is not available (i.e empty, due to inactive venv), the trailing whitespace will not be either.~

This PR makes it so that the python module is enabled only when a venv is active. All file [& extension] based detection has been disabled.

#### Motivation and Context

When a python module is detected (due to the presence of `.py` or `.ipynb` file(s)) but a venv is not active, only the trailing whitespace is displayed and since, in pure, it is placed before the prompt character, this leads to indentation of the prompt character by a single whitespace.

### DIFF
--- a/docs/public/presets/toml/pure-preset.toml
+++ b/docs/public/presets/toml/pure-preset.toml
@@ -44,3 +44,5 @@ style = "yellow"
 [python]
 format = "[$virtualenv]($style) "
 style = "bright-black"
+detect_extensions = []
+detect_files = []


davidkna (approved):

davidkna:
Thanks for the contribution @pantheraleo-7!

PR #6706, opened 2025-04-28
feat(install): Skip installation if Starship is already present and up-to-date

### LINKED ISSUE #6696: Skip installation if Starship is already present and up-to-date

I'm running Fedora, and since Starship isn't available in my system's package repository, I use the `install.sh` script to manage installation and updates, [as prescribed in the docs](https://github.com/starship/starship/blob/fa1fabf777b6be5e507eeee449dec62d0ccbdbc6/docs/README.md?plain=1#L50-L60).

However, I’ve noticed that the script proceeds with downloading and installing Starship even if it's already installed and up-to-date. This presents an opportunity for optimization.

A check should be added to ensure that download and installation only occur when necessary.

As a workaround, I've created a wrapper function in Bash for my own use:

```bash
update_starship() {
  # Skip installation if Starship is already present and up-to-date
  if command -v starship &>/dev/null; then
    local curr="$(starship -V)"
    curr="v${curr#starship }"
    local latest="$(curl -fsSL -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/starship/starship/releases/latest | jq -r '.tag_name')"
    if [[ "$curr" == "$latest" ]]; then
      echo "Starship is already up-to-date ($curr)."
      return 0
    fi
  fi
  # The platform needs to be overriden because on Linux,
  # a binary with statically linked musl is used by default.
  # See: https://github.com/starship/starship/issues/5361
  curl -fsSL https://starship.rs/install.sh | sh -s -- -V -f -p unknown-linux-gnu
}
```

This works fine, but it would be better if the check were to be incorporated into `install.sh` itself, as this is a functionality many users would benefit from. For this to happen, my script would need to be adapted to be more portable and less reliant on external utilities (i.e `jq`).

supatipunnon:
#### Description

Update install.sh to skip installation if it's already installed and up to date.

#### Motivation and Context

Closes #6696

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -302,6 +302,20 @@ check_bin_dir() {
   fi
 }
 
+check_upto_date(){
+
+  if [ "${VERSION}" = "latest" ] && [ -e "${BIN_DIR}/starship" ]; then
+    curr="$(${BIN_DIR}/starship -V)"
+    curr="v$(echo "$curr" | sed 's/^starship //')"
+    latest="$(curl -fsSL -H "X-GitHub-Api-Version: 2022-11-28" https://api.github.com/repos/starship/starship/releases/latest | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')"
+    if [ "$curr" = "$latest" ]; then
+      echo "Starship is already up-to-date ($curr)."
+      exit 0
+    fi
+  fi
+
+}
+
 print_install() {
   # if the shell does not fit the default case change the config file
   # and or the config cmd variable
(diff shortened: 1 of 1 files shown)


xty:
👋 Hi @supatipunnon! OP of #6696 here. I've just realized you created this PR. Thank you for taking the initiative to address the issue.

I've looked at your proposed changes, and they appear to be largely based on the code snippet I shared in the original issue. That snippet was intended as a temporary workaround and wasn’t particularly well thought out.

Since then, I’ve opened #6713, which proposes a much improved solution. I urge you to join forces with me there. Any feedback is welcome!

davidkna:
Closing in favor or https://github.com/starship/starship/pull/6713

PR #6708, opened 2025-04-28
docs(pixi): Fix typo

pavelzw:
Follow-up of #6335 

cc @davidkna


davidkna (approved):

davidkna:
Thanks for the fix @pavelzw!

PR #6712, opened 2025-04-29
Fix when starship would dissapear while having installed bash-preexec and bashrc was reloaded (and PS1 was set beforehand)

excited-bore:
#### Description

In certain scenarios, when PS1 is being set and bash-preexec is installed, when ~/.bashrc is reloaded starship prompt would dissapear. This was because starship.bash did not check whether starship_preexec_all / starship_precmd where already loaded by bash-preexec and would constantly rerun the block of code where it would add said functions to `preexec_functions / precmd_functions`

Issue reference: [#6702](https://github.com/starship/starship/issues/6702)

I have put the output of `starship init bash --print-full-init` in a file and sourced it before and after sourcing [bash-preexec.sh](https://github.com/rcaloras/bash-preexec) while having set PS1 beforehand.

Before this edit - starship dissapeared when resourcing - `source ~/.bashrc`
After this edit: issue gone

Testing environment referenced in issue.

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -92,8 +92,7 @@ if [[ ${BLE_VERSION-} && _ble_version -ge 400 ]]; then
     blehook PRECMD!='starship_precmd'
 # If the user appears to be using https://github.com/rcaloras/bash-preexec,
 # then hook our functions into their framework.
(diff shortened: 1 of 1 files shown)


excited-bore:
starship needs to be sourced **after** bash-preexec.sh though

excited-bore:
On rerunning this modification I noticed I missed that there was an entire unnesecary `if/else` for when the prompt would actually run if bash-preexec.sh/ble.sh was installed. Will have to retest new fix with both ble.sh and bash-preexec to make sure, then will make new pull request.

PR #5728, opened 2024-01-25
feat(install): Add version option to install script

luismgsantos:
Added version support so users can choose a specific version to install, instead of it always being the latest.

#### Description

Added support for users to specify the option `-v` or `--version` followed by the desired version.
Applied the suggestions on #5668 

#### Motivation and Context

Closes #5668

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #6713, opened 2025-04-29
feat: skip installation opportunistically based on installed version

### LINKED ISSUE #6696: Skip installation if Starship is already present and up-to-date

xty:
#### Description
This PR aims to optimize and streamline the installation/update process carried out by `install.sh`. Specifically, the proposed changes allow installation to be skipped entirely under one of the following conditions:
1) `install.sh` is run to install or update to the latest version, but it is already installed.
2) `install.sh` is run with the [`--version`](https://github.com/starship/starship/pull/5728) option to install a specific version, but that version is already installed.

The first scenario entails resolving the `latest` release to an actual release tag, which is done cleanly by querying the relevant public API endpoint, and in a way that is compatible with not only GitHub, but other code forges alike, which plays well with the existing `--base-url` design.

Extra care has been taken to ensure the introduced changes are robust, well-documented, consistent with the existing coding style, and cross-platform compatible.

#### Motivation and Context
Closes #6696

- ShellCheck has been used for static analysis, and no errors or warnings were reported.
- Manual testing has also been conducted thoroughly, covering a wide range of use cases.

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -379,6 +379,46 @@ print_install() {
   printf "\n"
 }
 
+# Test if the wanted version is already installed
+is_installed() {
+  wanted="$1"
+  wants_latest="$([ "${wanted}" = "latest" ] && printf true || printf false)"
+
+  if has starship; then
+    # Use ERE syntax for better clarity
+    tag_pattern="v[[:digit:]]+(\.[[:digit:]]+){2}"
+    # The currently installed version
+    current="v$(starship -V | sed 's/^starship //')"
+    printf '%s' "${current}" | grep -E -q "^${tag_pattern}" || return 1
+    if "${wants_latest}"; then
+      # Extract the latest release tag from the `location` header in a 3xx response
+      # See: https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases#linking-to-the-latest-release
+      # This method is also compatible with other forges like Codeberg (🇪🇺) and Gitee (🇨🇳).
+      has curl || return 1
+      location_header="$(
+        # Header picking using `--write-out '%header{location}'` (introduced in curl 7.84.0) could
+        # prove to be more straightforward.
+        # See: https://daniel.haxx.se/blog/2022/03/24/easier-header-picking-with-curl/
+        # Consider switching when adoption becomes more widespread
+        curl -fsS -I "${BASE_URL}/latest" |
+        grep -E -i -m 1 "^location:.*/releases/tag/${tag_pattern}" |
+        # Remove trailing whitespace (needed because each HTTP header is terminated by a CRLF)
+        sed 's/[[:space:]]*$//'
+      )"
+      wanted="${location_header##*/tag/}"
+    fi
+    if [ "${current}" = "${wanted}" ]; then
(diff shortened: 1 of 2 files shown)


xty:
It's ready for review. @andytom Would you care to take a look? 😊

flyutothemoon (approved):

install/install.sh
```diff
@@ -379,6 +379,46 @@ print_install() {
...
+is_installed() {
+  wanted="$1"
+  wants_latest="$([ "${wanted}" = "latest" ] && printf true || printf false)"
+
+  if has starship; then
+    # Use ERE syntax for better clarity
+    tag_pattern="v[[:digit:]]+(\.[[:digit:]]+){2}"
+    # The currently installed version
+    current="v$(starship -V | sed 's/^starship //')"
+    printf '%s' "${current}" | grep -E -q "^${tag_pattern}" || return 1
```
reneleonhardt, line 392: Wouldn't it be better to print a warning if current can't be determined from `starship -V`?
And why not just use the tag_pattern match instead of everything after the "starship " prefix when matching anyway (which would allow any suffix like " darwin/arm64")? 🤔

xty: Thanks for the input. Unfortunately, at this point, it’s unclear whether any substantial PRs are still being considered, so I’m not sure discussions around minor details are worthwhile.

reneleonhardt: Sorry to hear that, I know the feeling, hope the project stays great :(

davidkna:
The checks need to use the starship binary in the target installation directory, not in path. I am also not wholly convinced this should be the default behaviour and either way skipping should be avoidable.

PR #6735, opened 2025-05-13
style: use 'Self' to refer to own type (clippy::use_self)

danieleades:
#### Description

applies the `clippy::use_self` lint

#### Motivation and Context

1. makes the code slightly less verbose
2. marginally easier to refactor object names
3. makes 'constructor' methods more obvious

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,3 +134,6 @@ strip = false
 [[bin]]
 name = "starship"
 path = "src/main.rs"
+
+[lints.clippy]
+use_self = "warn"
--- a/src/configs/aws.rs
+++ b/src/configs/aws.rs
@@ -51,7 +51,7 @@ pub struct AwsConfig<'a> {
 
 impl Default for AwsConfig<'_> {
     fn default() -> Self {
-        AwsConfig {
+        Self {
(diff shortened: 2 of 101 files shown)


danieleades:
this could be enforced by adding it to the lints table in cargo.toml

davidkna:
> this could be enforced by adding it to the lints table in cargo.toml

Please do, additionally please also move the rules from `clippy.toml` there.

danieleades:
> > this could be enforced by adding it to the lints table in cargo.toml
> 
> Please do, additionally please also move the rules from `clippy.toml` there.

i've added this section. looks like there aren't any rules in `clippy.toml` so I haven't touched this

davidkna:
@danieleades It does contain [a few `disallowed-methods` definitions](https://github.com/starship/starship/blob/master/clippy.toml#L1-L8).

davidkna:
In that case let's just use `clippy.toml` to keep all clippy rules in one place.

danieleades:
> In that case let's just use `clippy.toml` to keep all clippy rules in one place.

you can't use `clippy.toml` to enable/disable lints, only to configure a subset of lints

davidkna (approved):

davidkna:
Thanks for the contribution @danieleades!

PR #6737, opened 2025-05-13
style: tiny refactor of the Module constructor

danieleades:
#### Motivation and Context

1. removes two instances where a string is converted to a reference and then back to a string

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -127,11 +127,15 @@ pub struct Module<'a> {
 
 impl<'a> Module<'a> {
     /// Creates a module with no segments.
-    pub fn new(name: &str, desc: &str, config: Option<&'a toml::Value>) -> Self {
+    pub fn new(
+        name: impl Into<String>,
+        desc: impl Into<String>,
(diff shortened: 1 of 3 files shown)


davidkna (approved):
LGTM.

davidkna:
> the other option would be to have the constructor accept `String` and then add a `to_string` everywhere it's currently called with a static string.

These changes should work fine. Another option would be to use `Cow<str, 'static>` as the internal type, which should avoid the owned variant in most cases.

danieleades:
> > the other option would be to have the constructor accept `String` and then add a `to_string` everywhere it's currently called with a static string.
> 
> These changes should work fine. Another option would be to use `Cow<str, 'static>` as the internal type, which should avoid the owned variant in most cases.

yeah i thought about that. that would reduce memory usage, i guess? can't see it having an impact on latency though, which i'm assuming is the main driver for this repo. Just didn't seem worth it in this case, but i don't know enough about this codebase to say, and there's no benchmarks (some benchmarks would be nice)

PR #6744, opened 2025-05-17
Option to cache os info

junzh0u:
Allow caching OS info in OS module.

#### Description
This PR adds an option to cache OS info in OS module.
It's off by default and can be turned on via:

```
[os]
cache = true
```

#### Motivation and Context
According to `starship explain`, the OS module takes 10~20ms on my Macbook (Macbook Pro M1 Max).
Given that the OS info shouldn't change within same terminal session, it makes sense to cache it.
The cache file is put in "$STARSHIP_CACHE" with `.log` extension, to reuse the old logs cleanup mechanism.

Before change:
```
$ starship explain

...
 " " (12ms)               -  The current operating system
...
```

After change:
```
$ starship explain

...
 " " (<1ms)               -  The current operating system
...
```

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,14 +10,14 @@ checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "ahash"
-version = "0.8.11"
+version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


davidkna:
Thanks for the contribution!

On macOS, patching `os_info` to avoid external commands improves performance to <1ms without caching (M1). I will try to get these changes merged first before looking at this further, since this would remove the need for caching in this module.

junzh0u:
Thanks!
I see your PR: https://github.com/stanislav-tkach/os_info/pull/406
Yes it makes a lot of sense if we can make the upstream faster instead.

PR #6747, opened 2025-05-20
feat(python): add support for `pyenv.cfg` version

RoloEdits:
#### Description

As mentioned [`here`](https://github.com/starship/starship/pull/6523#issuecomment-2708925618), `.venv/pyvenv.cfg` has a `version_info` field:
```
home = ...
implementation = CPython
uv = 0.7.4
version_info = 3.11.12
include-system-site-packages = false
prompt = test
```

#### Motivation and Context

This can be used to read a version number without having to run any binaries, potentially reducing the surface area for an attack, as was also discussed.

Closes #

image

With a config of:
```toml
pyenv_version_name = true
python_binary = []
```

I copied the binary to run in my running shell.

#### Checklist:

### DIFF
--- a/src/configs/python.rs
+++ b/src/configs/python.rs
@@ -11,6 +11,8 @@ use serde::{Deserialize, Serialize};
 #[serde(default)]
 pub struct PythonConfig<'a> {
     pub pyenv_version_name: bool,
+    pub venv_pyenvcfg_version: bool,
+    pub force_venv_pyenvcfg_version: bool,
     pub pyenv_prefix: &'a str,
     pub python_binary: VecOr<VecOr<&'a str>>,
     pub format: &'a str,
@@ -28,6 +30,8 @@ impl Default for PythonConfig<'_> {
     fn default() -> Self {
         PythonConfig {
             pyenv_version_name: false,
+            venv_pyenvcfg_version: false,
+            force_venv_pyenvcfg_version: false,
             pyenv_prefix: "pyenv ",
             python_binary: VecOr(vec![
                 VecOr(vec!["python"]),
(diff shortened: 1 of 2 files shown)


RoloEdits:
@davidkna Want to get some feedback on the design of this before going forward, as you were the one who mentioned it before. Currently I just have it tied to the `pyenv_version_name` option, where when its true, it will try to parse the `pyvenv.cfg` first before any binary is invoked. 

Not sure if you had any specifics in mind when you suggested this. This currently doesn't require any activation of the venv, but this would also be my preference, to see the version even before activation, and it also works well for `uv` as most projects don't really need activation, as when you run, it will resolve whats needed from the venv automatically.

davidkna:
This looks good to me. I think that this should use some sort of opt-in (if no venv is activated) in the starship config to avoid confusion, perhaps a list of search directories like #6526.
Also I am not sure how well-formed any given `pyenv.cfg` will be, so I would be more comfortable if this checked if `semver:Version::parse(cand_version).is_ok()`.

RoloEdits:
So doing some more digging, the `pyenv.cfg` file is separate from `pyenv`. So now what I am doing is to make it a separate version getter (just like the `pyenv` one) which takes priority when `venv_pyenvcfg_version` is true and the virtual environment is activated. 

There is then another option, `force_venv_pyenvcfg_version`, that will force to get the version from the `pyenv.cfg` file even if not activated. This (hopefully) follows more inline with expectations?

I did add a fallout for the force option, where when the file is not found, it will still try to get the version via `pyenv` or the binaries.

If this looks good I will add the needed tests and update the docs.

PR #6754, opened 2025-05-27
feat: add different time formats to `cmd_duration`

### LINKED ISSUE #6642: More transparent command duration output
## Feature Request

Would be great ore transparent command duration output, like this:
```

format	      0.001s	2.1s	    3m2.1s	   4h3m 2.1s

example        1ms      2.1s	    3m 2.1s	   4h 3m 2.1s
example2       1ms     2s 100ms	   3m 2s 100ms	   4h 3m 2s 100ms
```

sehraf:
#### Description

This PR introduces different ways to format the duration displayed by the `cmd_duration` module. 

#### Motivation and Context

I've recently switched from `oh-my-posh` to `starship.rs` and was missing a readable time duration format. I've implemented [all formats](https://ohmyposh.dev/docs/segments/system/executiontime#style) from `oh-my-posh` and copied [all their tests](https://github.com/JanDeDobbeleer/oh-my-posh/blob/fb6d8fc9965170959af8087e0dec800eb7bdc97f/src/segments/executiontime_test.go#L86-L347) (tests cover both `utils.rs` and `cmd_duration.rs`).

Closes #6642 

#### (Possible) Open Points
- I'm making heavy use of `format!`, no idea if this has a relevant impact on performance.
- I've failed to directly deserialize the new format enum from `toml` and therefore went the `From<&str>` route. The side-effect of this is that the config struct takes a `&str` instead of the enum itself. Looking at existing modules, every config option seems to be a primitiv type (or vector) :thinking: I'm open to suggestions.
- This could be seen as a breaking change, though the current behavior is kept when no format is specified. The `show_milliseconds` switch is obsolete (but kept to keep the previous behavior when the format is `None`). Any suggestion on how to deal with this?

image
image

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -183,7 +183,8 @@
         "min_time_to_notify": 45000,
         "show_milliseconds": false,
         "show_notifications": false,
-        "style": "yellow bold"
+        "style": "yellow bold",
+        "time_format": "None"
       },
       "allOf": [
         {
@@ -2559,6 +2560,10 @@
           "default": 45000,
           "type": "integer",
           "format": "int64"
+        },
+        "time_format": {
+          "default": "None",
+          "type": "string"
         },
         "notification_timeout": {
           "type": [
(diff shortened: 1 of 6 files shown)


acuteaangle:
> [all formats](https://ohmyposh.dev/docs/segments/system/executiontime#style) from oh-my-posh

As of https://github.com/JanDeDobbeleer/oh-my-posh/pull/7298, this is technically missing `iso8601` and `iso8601ms`.

Interestingly, Oh My Posh decided to omit days. I do not think copying that behaviour makes sense, as we already have days available and use it in other formats.

1 day, 2 hours, 3 minutes, 4 seconds, 567 miliseconds:
P1DT2H3M4.567S

PR #6755, opened 2025-05-27
fix(git_status): Do not show up-to-date when untracked

### LINKED ISSUE #6736: git_status: incorrectly using `up_to_date` when branch is untracked
#### Current Behavior

`git_status` module reports branch to be `up_to_date` when untracked

#### Expected Behavior

None of `up_to_date`, `ahead`, `diverged` and `behind` should be displayed if the current branch is untracked.
This is a regression, introduced by the switch to gitoxide. Setting use_git_executable = true reverts back to the previous correct behavior.

Bonus points: consider adding an `untracked` configuration option

#### Possible Solution

Looking at [the code](https://github.com/starship/starship/blob/master/src/modules/git_status.rs#L68-L88), it looks like the `info.get_ahead_behind()` call just fails (or something) when using the git executable, making the block do nothing, while gitoxide happily gives always something.

I'd be happy to put together a PR if this is confirmed (and in that case I'm taking all advices!)

#### Environment
- Starship version: 1.23.0
- bash version: GNU bash, version 5.2.37(1)-release (aarch64-apple-darwin23.4.0)
Copyright (C) 2022 Free Software Foundation, Inc.

- Operating system: Mac OS 15.4.1
- Terminal emulator: vscode 1.100.0
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.86.0 (05f9846f8 2025-03-31) (Homebrew)
- Rust channel:  release
- Build Time: 2025-04-27 15:52:14 +00:00

```bash
#!/usr/bin/env bash

# shellcheck source=bash_profile
source ~/.bash_profile
```

#### Starship Configuration

```toml

format = '$time$all$shlvl$character'

[git_status]
format = '($all_status$ahead_behind )'
untracked = "[%](red)"
modified = "[*](red)"
staged = "[+](green)"
stashed = '[\$](cyan bold)'
typechanged = "[~](red)"
renamed = '[»](green)'
deleted = '[✘](red)'
conflicted = "[@](red)"

up_to_date = "|u="
ahead = '|u⇡${count}'
diverged = '|u⇡${ahead_count}⇣${behind_count}'
behind = '|u⇣${count}'

```

serl:
#### Description

Added an extra bit to the `git for-each-ref` call to make sure we have a remote tracking branch.

#### Motivation and Context

Closes #6736

#### Checklist:

### DIFF
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -570,32 +570,27 @@ impl RepoStatus {
     }
 
     fn set_ahead_behind_for_each_ref(&mut self, mut s: &str) {
-        let re = Regex::new(r"^([^ ]+) (\[(.*)\])?$").unwrap();
-        let Some(caps) = re.captures(s) else {
+        if s == " " || s.ends_with(" [gone]") {
             self.ahead = None;
             self.behind = None;
             return;
-        };
-        let ahead_behind_match = caps.get(3);
-        if let Some(ahead_behind_match) = ahead_behind_match {
-            s = ahead_behind_match.as_str();
+        }
 
-            if s == "gone" {
-                self.ahead = None;
-                self.behind = None;
-                return;
-            }
+        s = s
+            .split_once(' ')
+            .unwrap()
+            .1
+            .trim_matches(|c| c == '[' || c == ']');
 
-            for pair in s.split(',') {
-                let mut tokens = pair.trim().splitn(2, ' ');
-                if let (Some(name), Some(number)) = (tokens.next(), tokens.next()) {
-                    let storage = match name {
-                        "ahead" => &mut self.ahead,
-                        "behind" => &mut self.behind,
-                        _ => return,
-                    };
-                    *storage = number.parse().ok();
-                }
+        for pair in s.split(',') {
+            let mut tokens = pair.trim().splitn(2, ' ');
+            if let (Some(name), Some(number)) = (tokens.next(), tokens.next()) {
+                let storage = match name {
+                    "ahead" => &mut self.ahead,
+                    "behind" => &mut self.behind,
+                    _ => return,
+                };
+                *storage = number.parse().ok();
             }
         }
         for field in [&mut self.ahead, &mut self.behind] {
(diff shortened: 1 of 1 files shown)


serl:
(This PR is in draft since I forgot about the `gone` case. I'm adding something for that before calling that "ready")

src/modules/git_status.rs
```diff
@@ -570,23 +570,38 @@ impl RepoStatus {
...
-            let mut tokens = pair.trim().splitn(2, ' ');
-            if let (Some(name), Some(number)) = (tokens.next(), tokens.next()) {
-                let storage = match name {
-                    "ahead" => &mut self.ahead,
-                    "behind" => &mut self.behind,
-                    _ => return,
-                };
-                *storage = number.parse().ok();
+        let re = Regex::new(r"^([^ ]+) (\[(.*)\])?$").unwrap();
+        if let Some(caps) = re.captures(s) {
```
davidkna:
```suggestion
        let Some(caps) = re.captures(s) else {
            self.ahead = None;
            self.behind = None;
            return
        };
```

src/modules/git_status.rs
```diff
@@ -570,23 +570,38 @@ impl RepoStatus {
...
-        for pair in s.split(',') {
-            let mut tokens = pair.trim().splitn(2, ' ');
-            if let (Some(name), Some(number)) = (tokens.next(), tokens.next()) {
-                let storage = match name {
-                    "ahead" => &mut self.ahead,
-                    "behind" => &mut self.behind,
-                    _ => return,
-                };
-                *storage = number.parse().ok();
+        let re = Regex::new(r"^([^ ]+) (\[(.*)\])?$").unwrap();
```
davidkna: This regex should not be recreated for every invocation, please try to avoid regexes or if that is not possible cache the regex.

serl: Done. And funnily enough, the diff is even shorter now 😁 

Let me know if there's anything else to do

src/modules/git_status.rs
serl: Not relevant anymore after removing the regex

davidkna:
Thanks for the fix @serl!

PR #6770, opened 2025-06-03
chore: remove redundant config in cargo.toml

danieleades:
this config block is not required, since this is the default behaviour when `src/main.rs` is present


davidkna:
@danieleades Please fix the conflicts.

davidkna:
Thanks for the cleanup @danieleades, and thanks for the review @cappyzawa!

PR #6771, opened 2025-06-03
style: pedantic refactoring

danieleades:


davidkna:
Thanks for the cleanup @danieleades!

PR #6795, opened 2025-06-14
feat(haskell): Improve `haskell` module

mmhat:
This commit refactors the Haskell module to provide a cleaner discovery for Haskell setups. The following setups are tried in order:

 1. Check if we are in a directory tree below a Stack project.
 2. Check if we are in a directory tree below a Cabal project.
 3. If neither is the case, look for Haskell files in the current directory.

For Stack and Cabal the compiler used by the project is now determined correctly using the respective `stack` or `cabal` tool: Before, the output was simply the version of GHC compiler found in PATH, which may not be the one used by the project setup.

#### Description

  * There are new configuration options `stack_extensions`, `stack_files`, `stack_folders` we use to check if we are below a Stack project.
    _Note:_ We also look in ancestor directories for `stack_files` and `stack_folders` now.

  * Similarly, there are new configuration options `cabal_extensions`, `cabal_files`, `cabal_folders` we use to check if we are below a Cabal project.
    _Note:_ We also look in ancestor directories for `cabal_files` and `cabal_folders` now.

  * The configuration options `detect_extensions`, `detect_files` and `detect_folders` are now used to detect a "plain" Haskell setup, i.e. one that is neither a Cabal nor a Stack project.

  * A new configuration option `default_compiler` was added. This can be used to override the executable name that is queried for version information if we are neither in Stack nor in a Cabal project. The executable is expected to understand the flag `--numeric-version`.

  * New output fields `compiler_id`, `compiler_name` and `compiler_version` were added.
    The already existing field `ghc_version` is now an alias for `compiler_version`, where the latter has the benefit that its name does not suggest a particular Haskell compiler implementation.

#### Motivation and Context

The module only scanned the _current_ for Cabal- and Stack project files. So when you were working in a source directory of a project (i.e. in a directory below the project root) Starship did not pick that up and displayed the wrong information (or none at all).

In addition to that, I found it irritating that the module displays the installed instead of the used compiler:
Both Cabal and Stack allow you to specify the version (or path) of the used compiler in their project files, and when working with those the default compiler in `PATH` may not be used.

I updated the tests and added mock-commands for `cabal`, `ghc`, `mhs` and `stack`.

#### Checklist:

### DIFF
--- a/src/modules/haskell.rs
+++ b/src/modules/haskell.rs
@@ -2,22 +2,23 @@ use super::{Context, Module, ModuleConfig};
 
 use crate::configs::haskell::HaskellConfig;
 use crate::formatter::StringFormatter;
+use crate::utils::read_file;
+use std::path::PathBuf;
+
+struct HaskellResult {
+    compiler_name: Option<String>,
+    compiler_version: Option<String>,
+    snapshot: Option<String>,
+}
 
 /// Creates a module with the current Haskell version
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("haskell");
     let config = HaskellConfig::try_load(module.config);
 
-    let is_hs_project = context
-        .try_begin_scan()?
-        .set_files(&config.detect_files)
-        .set_extensions(&config.detect_extensions)
-        .set_folders(&config.detect_folders)
-        .is_match();
-
-    if !is_hs_project {
-        return None;
-    }
+    let haskell_result: HaskellResult = stack_result(context, &config)
+        .or_else(|| cabal_result(context, &config))
+        .or_else(|| plain_haskell_result(context, &config))?;
 
     let parsed = StringFormatter::new(config.format).and_then(|formatter| {
         formatter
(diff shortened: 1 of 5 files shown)


mmhat:
Ok, I think this one is ready for a review.

src/modules/haskell.rs
```diff
@@ -2,22 +2,23 @@ use super::{Context, Module, ModuleConfig};
...
-        .set_extensions(&config.detect_extensions)
-        .set_folders(&config.detect_folders)
-        .is_match();
-
-    if !is_hs_project {
-        return None;
-    }
+    let haskell_result: HaskellResult = stack_result(context, &config)
+        .or(cabal_result(context, &config))
+        .or(plain_haskell_result(context, &config))?;
```
ksqsf: I think you meant to use `or_else`. Rust has no lazy evaluation :)

mmhat: Done.

mmhat:
Is there anything I can do to get this merged?

davidkna:
It looks like https://github.com/starship/starship/pull/6795#discussion_r2164286494 has not actually been addressed and this PR is s still showing warnings related to clippy in CI.

mmhat:
> It looks like [#6795 (comment)](https://github.com/starship/starship/pull/6795#discussion_r2164286494) has not actually been addressed and this PR is s still showing warnings related to clippy in CI.

@davidkna Oh, you're absolutely right! I forgot to push that change!
I applied the suggestion (really, this time).

mmhat:
Gentle ping; Is there anything I should change or can we merge it?

PR #6797, opened 2025-06-15
<!--- Provide a general summary of your changes in the Title above --> <!--- To help with semantic versioning the PR title should start with one of the conventional commit types. --> <!--- The conventional commit types for Semantic PR are: feat, fix, docs, style, refactor, perf, test, build, ci, chore, revert -->

cfergeau:
#### Description

fix(preset): Add missing modules to no-runtime-versions

A few modules are missing from the no-runtime-versions preset.
This also ensures the file is alphabetically ordered.
mojo and buf use "with" rather than "via", the preset now uses "with"
for these.

I’ve built starship and checked that starship preset no-runtime-versions hasthe expected content.

#### Checklist:

I don’t think a documentation update is needed, the preset is already documented.
Not sure there are tests for this?


docs/public/presets/toml/no-runtime-versions.toml
```diff
@@ -2,14 +2,20 @@
 format = "via [$symbol]($style)"
 
 [buf]
-format = "via [$symbol]($style)"
+format = "with [$symbol]($style)"
+
+[c]
+format = "via [$symbol($name)]($style)",
```
davidkna: Some of the TOML syntax is invalid

cfergeau: Oops, I knew this would happen /o\. Fixed now.

davidkna (approved):

davidkna:
Thanks for handling the missing modules @cfergeau!

PR #6806, opened 2025-06-19
feat(cmd_duration): add `notifications_transient` config to suppress persistent notifications

guillemglez:
Adds a new `notifications_transient` boolean option to the `cmd_duration` module. When enabled, the module will request that desktop notifications are not stored in the notification center/tray via Hint::Transient(true).

#### Description

This change introduces an optional `notifications_transient` setting in the `cmd_duration` module config.

When `true`, Starship will set the `Transient` hint on the notification, signaling that it should not be kept in the system's notification history/tray. Note this only works on Linux.

#### Motivation and Context

The cmd_duration module displays how long the last command took to run, and optionally shows a desktop notification if the command exceeded a configured threshold. These notifications help the user in knowing long-running tasks are over even if the terminal isn’t in focus. Without transient behavior, repeated "command finished" notifications can accumulate and clutter the notification tray. Adding a transient hint will make brief notifications that disappear after being shown.

Enabling notifications_transient helps mitigate this by requesting that the notification be displayed without being retained. This is useful in (Linux) environments where Hint::Transient(true) is supported (like GNOME, KDE, Sway, XFCE...).

If the notification daemon does not support the transient hint, enabling this option will have no effect: the notification will behave as it normally would.

The default is set to `false` to preserve current behavior, but would agree in considering changing it to `true`, which imo feels like a better default UX.

- [ x ] I have tested using **Linux**

#### Checklist:

- [ x ] I have updated the documentation accordingly.

Note: not updated tests since it does not look applicable to this feature.

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -278,6 +278,7 @@
         "show_milliseconds": false,
         "disabled": false,
         "show_notifications": false,
+        "notifications_transient": false,
         "min_time_to_notify": 45000
       }
     },
@@ -2534,6 +2535,10 @@
           "default": false
         },
         "show_notifications": {
+          "type": "boolean",
+          "default": false
+        },
+        "notifications_transient": {
           "type": "boolean",
           "default": false
         },
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -914,16 +914,17 @@ running `eval $(starship init $0)`, and then proceed as normal.
 
 ### Options
 
-| Option                 | Default                       | Description                                                                                                                                                       |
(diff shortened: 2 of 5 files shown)


guillemglez:
This is ready for review, perhaps can take a look, @davidkna ? 

(I don't think testing can really be applied to this feature, but if you have suggestions, happy to increase the diff coverage)

guillemglez:
Worth a review after this rebase, @davidkna ? Coverage shows as an issue, but this is very GNOME/DBus-specific... runtime behavior is best validated manually on those systems... Don't know if you have any suggestion about that.

PR #6694, opened 2025-04-22
fix: Respect timeout in from_path_with_timeout in a NAS context

SpookyYomo:
In the case of an auto-mountable directory, the fetching may stall. 
This commit pushes the fetching into its own thread, with results sent across a channel. 
The original thread will then monitor for timing out, ensuring that the timeout is enforced.

#### Description

We create a potentially dangling join_handle which runs the enumeration of `fs::read_dir`. It should terminate the moment it fetchs a single file and realises the channel is closed. The fs entry is then sent over a channel to the parent thread for processing.

I am not sure if having a separate thread + channel may be too heavy-handed.

#### Motivation and Context

Closes #5740

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [X] I have tested using **Linux**
- [ ] I have tested using **Windows**
I ran starship with my fstab containing my NAS and without my NAS. In the case with the NAS, the logging shows timeout was hit, and without the NAS, nothing noticeably different occured.

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #6818, opened 2025-06-29
test: Add stalling tests

SpookyYomo:
#### Description

libfuser is a means of creating a userspace filesystem, which allows for us to mock
an unresponsive filesystem. This is however Unix-only. For now, we reserve the tests 
strictly for Linux before seeing if CI thinks Mac can handles this too.

#### Motivation and Context

#6694 apparently needs more testing.

#### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3294,6 +3294,7 @@ dependencies = [
  "home",
  "indexmap",
  "jsonc-parser",
+ "libc",
  "log",
  "mockall",
  "nix",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -924,6 +924,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "28dd6caf6059519a65843af8fe2a3ae298b14b80179855aeb4adc2c1934ee619"
 
 [[package]]
+name = "fuser"
+version = "0.15.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "53274f494609e77794b627b1a3cddfe45d675a6b2e9ba9c0fdc8d8eee2184369"
+dependencies = [
+ "libc",
+ "log",
+ "memchr",
+ "nix",
+ "page_size",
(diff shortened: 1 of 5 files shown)


SpookyYomo:
If maintainers like this idea I'll make the unit tests.
However I am not certain if the file locations as I have placed them at are sensible.

davidkna:
Should this be closed or do you still want to add tests for this with the other PR merged?

SpookyYomo:
> Should this be closed or do you still want to add tests for this with the other PR merged?

Sorry I'll get round to finishing this end within the next 2 weeks, its just mangled pieces of uncommited code atm (I'm in the midst of testing that FUSE unmounts with the relevant struct dropping/thread closing) which is quite irritating to test and hence the general lack of my cleaning up of my WIP for pushing/review.

SpookyYomo:
Rebased against a newer commit that included #6694 that didn't have build failures.
I don't think I've written the test case wrongly...? The test I written shows that timeout is not being respected.

edit:
Diagnosed the issue.
```rust
            let msg = if cfg!(test) {
                // recv() errors out only when the corresponding sender closes.
                // See mpsc::RecvError.
                rx.recv().map_err(|_| mpsc::RecvTimeoutError::Disconnected)
            } else {
                rx.recv_timeout(remaining_time)
            };
```
It's still blocking on these lines until the worker thread does any amount of `tx.send(...)`.
I think the only real solution to this problem might be through
1. async in some form (details need to be examined).
2. a busy try_recv() loop.

This is needed because of the semantic of `recv_timeout` in contrast to `recv_deadline`.

SpookyYomo:
@davidkna Let me know if you'd prefer I mark the test as ignore first, and the desired strategy between async (I'll have to see if async_std crate is suitable) and busy looping is preferred in a separate PR.

SpookyYomo:
@davidkna I saw #7120 and am just bumping you on this which shows why it even occurs.

PR #6822, opened 2025-06-30
fix(preset/nerd-font): remove double spaces in `gcloud` default value and use Google Cloud-specific icon

injust:
#### Description

https://github.com/starship/starship/pull/6649 appears to have missed a double space in the `gcloud` section.

#### Motivation and Context

Follow-up of https://github.com/starship/starship/pull/6649

#### Checklist:

### DIFF
--- a/docs/public/presets/toml/nerd-font-symbols.toml
+++ b/docs/public/presets/toml/nerd-font-symbols.toml
@@ -47,7 +47,7 @@ symbol = " "
 symbol = " "
 
 [gcloud]
-symbol = " "
+symbol = " "
 
 [git_branch]
 symbol = " "


docs/public/presets/toml/nerd-font-symbols.toml
```diff
@@ -47,7 +47,7 @@ symbol = " "
 symbol = " "
 
 [gcloud]
-symbol = "  "
+symbol = " "
```
davidkna: I looked at this, and some terminals actually render this one as double-width (requiring two spaces), let's replace this with `\ue7f1` (`nf-dev-googlecloud`) which appears to work better in this matter and is specific to gcloud.

davidkna (approved):

davidkna:
Thanks for the fix @injust!

PR #6825, opened 2025-06-30
docs(shlvl): fix repeat example

injust:
#### Description

To get a prompt like `❯❯❯`, `$shlvl` needs to be removed from `format`.

#### Checklist:

### DIFF
--- a/.dprint.json
+++ b/.dprint.json
@@ -25,7 +25,7 @@
     "target/"
   ],
   "plugins": [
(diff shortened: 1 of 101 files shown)


injust:
`threshold = 0` in the example is also redundant and should be removed.

The example sets `repeat_offset = 1`, which makes the behaviour of `threshold = 0/1/2` identical. It's simpler to leave `threshold` to the default value of `2`.

davidkna:
Thanks for the fix @injust!

PR #6828, opened 2025-06-30
docs(c, cpp): remove incorrect note about default `format` string

injust:
#### Description

The default `format` string contains `$version`, so the note is incorrect and should be removed.

#### Checklist:


davidkna (approved):

davidkna:
Thanks for the contribution @injust!

PR #6830, opened 2025-06-30
docs: add tip for all disabled-by-default modules

injust:
#### Description

Add missing tips to all disabled-by-default modules.

#### Motivation and Context

It's annoyingly difficult to find a list of modules that are disabled by default. I had to use regex to search the documentation source to do this.

#### Checklist:


davidkna (approved):

davidkna:
Thanks for the contribution @injust!

PR #6834, opened 2025-07-01
feat(zsh): add wide character glitch sequences to prevent cursor positioning issues

iainlane:
#### Description

zsh doesn't always have the same idea of character width as terminal emulators do. Wide Unicode characters can cause prompt rendering issues in `zsh:` the cursor overlaps the prompt or `RPROMPT` can push the cursor to the next line inadvertently . This change wraps wide characters in zsh's [glitch prompt sequence] (`%{<char>%<width>G%}`) to explicitly tell zsh the actual display width of the character being drawn.

[glitch prompt sequence]: https://zsh.sourceforge.io/Doc/Release/Prompt-Expansion.html#Visual-effects

#### Screenshots

before, notice the cursor is positioned over the left prompt

image

after

image

#### Checklist

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -20,7 +20,7 @@ use crate::module::Module;
 use crate::modules;
 use crate::segment::Segment;
 use crate::shadow;
-use crate::utils::wrap_colorseq_for_shell;
+use crate::utils::{wrap_colorseq_for_shell, wrap_widechar_for_zsh};
 
 pub struct Grapheme<'a>(pub &'a str);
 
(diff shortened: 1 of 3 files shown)


iainlane:
One thing I don't know is: could this cause problems for people where the terminal emulator reports a width of 1 for these characters currently, meaning they're already aligned with zsh? I'm not sure, because Starship will be reporting 2 thanks to `unicode_width`, so perhaps it's fine that we more explicitly align with that.

PR #6835, opened 2025-07-01
test(print): prevent environment leakage in prompt tests

iainlane:
#### Description

The `prompt_with_all` and `rprompt_with_all` tests were failing here because they accessed real environment data instead of using isolated test environments. This was observed with `gcloud` and caused tests to fail with output like:

```console
thread 'print::test::prompt_with_all' panicked at src/print.rs:576:9:
  left: ">"
 right: "on \u{1b}[1;34m☁\u{fe0f}  user@example.com\u{1b}[0m >"
```

Fix by setting HOME="/nonexistent" in these tests to prevent modules from accessing real configuration files.


src/print.rs
```diff
@@ -568,6 +568,7 @@ mod test {
                 [character]
                 format=">"
         });
+        context.env.insert("HOME", "/nonexistent".to_string());
```
davidkna: I would prefer `/dev/null` on -nix and `NUL` on Windows if that works.

iainlane: `/dev/null` works, at least for these tests. Updated!

I don't think `${HOME}` is usually used on Windows, is it?

davidkna: On Windows starship also uses `HOME` for mocking the homedir in the tests.

iainlane: OK! I see the Windows tests did pass here. Is that enough or should I update this?

davidkna: I would prefer an update for correctness since `/dev/null` could exist on Windows (I think it should resolve to `c:\dev\null`).

iainlane: OK thanks @davidkna. Updated!

davidkna:
Thanks for the fix @iainlane!

PR #6861, opened 2025-07-24
fix(explain): Filter out non-printing ANSI escape codes

KaiSforza:
#### Description

The string `\x1b[35;49mkai` is, on the screen, only 3 characters long, however the `width_graphemes()` function would say that it is 10 characters long. This is only really an issue for `explain`, as it makes the width of the left column incorrect for the padding. Instead of the following:

```console
$ starship explain
"/" (<1ms)    - <description>
"kai" (<1ms)  - <description>
```

it would print out like this for any parts that don't use ansi escapes internally:

```console
"/" (<1ms)           - <description>
"kai" (<1ms)  - <description>
```

This should cause no problems with any of the unstyled lengths that are generated normally, but any pre-styled strings in `custom` calls will cause this issue.

Tested with my config which uses `starship-jj`.

#### Checklist:

### DIFF
--- a/src/print.rs
+++ b/src/print.rs
@@ -1,10 +1,12 @@
 use clap::{ValueEnum, builder::PossibleValue};
 use nu_ansi_term::AnsiStrings;
 use rayon::prelude::*;
+use regex::Regex;
 use std::collections::BTreeSet;
 use std::fmt::{Debug, Write as FmtWrite};
 use std::io::{self, Write};
 use std::path::PathBuf;
+use std::sync::OnceLock;
 use std::time::Duration;
 use terminal_size::terminal_size;
 use unicode_segmentation::UnicodeSegmentation;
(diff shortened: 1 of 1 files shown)


cappyzawa (commented):
Great catch! This fixes the ANSI escape code width calculation issue in `starship explain`.

The fix works correctly, but I noticed a performance impact - `width_graphemes()` now compiles the regex on every call. Consider using `std::sync::OnceLock` to cache the compiled regex:

```rust
use std::sync::OnceLock;

static ANSI_REGEX: OnceLock<Regex> = OnceLock::new();

fn get_ansi_regex() -> &'static Regex {
    ANSI_REGEX.get_or_init(|| Regex::new(r"\x1B\[[0-9;]*m").unwrap())
}

// Then in width_graphemes():
fn width_graphemes(&self) -> usize {
    get_ansi_regex()
        .replace_all(self.as_ref(), "")
        .graphemes(true)
        .map(Grapheme)
        .map(|g| g.width())
        .sum()
}
```

Otherwise LGTM! 👍

KaiSforza:
Great catch, I've been trying to make my rust more compact, sometimes without any concern for performance. Coming from other languages, it's not something I have spent a lot of time debugging or fixing!

Thank you and I'll submit an updated commit soon!

davidkna:
Thanks for the contribution @KaiSforza, and thanks for reviewing @cappyzawa!

PR #6887, opened 2025-08-07
fix(test/git_commit): avoid spawning an editor lead to timeout if tag signing is enabled

siddharthkumarjha:
#### Description
The 
```bash
cargo test
```
command was stalling indefinitely for me in `test_latest_tag_shown_with_tag_enabled_lightweight` and `test_latest_tag_shown_with_tag_enabled` tests. The reason being git was launching my editor for me to add a message for the tags in background which wasn't visible to me as a user for commands like 
```bash
git tag l0
```
when launched from cargo tests.

#### Motivation and Context
The changes essentially enable proper testing of git modules by simply providing the tag message inline instead of waiting for user to provide input to an invisible editor.

**issue before my changes:**
image showing these tests stalling.
2025-08-07T23:16:38,610269214+05:30

**the same test with my changes.**
- the full test suite runs properly.
2025-08-07T23:01:18,146610449+05:30
- the two tests properly working.
2025-08-07T23:01:53,084848085+05:30

#### Checklist:

### DIFF
--- a/src/modules/git_commit.rs
+++ b/src/modules/git_commit.rs
@@ -348,26 +348,26 @@ mod tests {
         let commit_output = str::from_utf8(&git_commit).unwrap().trim();
 
         create_command("git")?
-            .args(["tag", "v2", "-m", "Testing tags v2"])
+            .args(["tag", "--no-sign", "v2", "-m", "Testing tags v2"])
             .env("GIT_COMMITTER_DATE", "2022-01-01 00:00:00 +0000")
             .current_dir(repo_dir.path())
             .output()?;
 
         create_command("git")?
-            .args(["tag", "v0", "-m", "Testing tags v0", "HEAD~1"])
+            .args(["tag", "--no-sign", "v0", "-m", "Testing tags v0", "HEAD~1"])
             .env("GIT_COMMITTER_DATE", "2022-01-01 00:00:01 +0000")
             .current_dir(repo_dir.path())
             .output()?;
 
         create_command("git")?
-            .args(["tag", "v1", "-m", "Testing tags v1"])
+            .args(["tag", "--no-sign", "v1", "-m", "Testing tags v1"])
             .env("GIT_COMMITTER_DATE", "2022-01-01 00:00:01 +0000")
             .current_dir(repo_dir.path())
             .output()?;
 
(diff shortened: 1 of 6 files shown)


src/modules/git_commit.rs
```diff
@@ -367,7 +367,7 @@ mod tests {
 
         // Annotated tags are preferred over lightweight tags
         create_command("git")?
-            .args(["tag", "l0"])
+            .args(["tag", "l0", "-m", "Testing tags l0"])
```
davidkna: This is supposed to create a lightweight tag, as opposed to a an annotated tag (with a message) which are handled quite differently internally by git. Please consult your git config if there is anything requiring tags to be annotated. The command should be changed to disable that setting.

siddharthkumarjha: ok, btw I also noticed starship is not compiling (on CI pipeline, due to warnings) with rust 1.89, the latest stable release and included the warning fixes in the PR. is that okay? or should I create a separate PR for that, or leave it to maintainers?

davidkna: Keeping the fix should be fine in this case.

siddharthkumarjha: done, @davidkna it seems (at least for me) the issue was, 
```gitconfig
[tag]
    gpgsign = true
```
param of my `.gitconfig`

siddharthkumarjha: here is my full `.gitconfig` just in case you wish to repro and test on your side,

```gitconfig
[user]
    name = <NAME>
    email = <EMAIL_ADDR>
    signingkey = <HOME>/.ssh/id_ed25519.pub
[core]
    editor = nvim
[color]
    status = auto
    branch = auto
    interactive = auto
    diff = auto
[gpg]
    format = ssh
[commit]
    gpgsign = true
[gpg "ssh"]
    allowedSignersFile = <HOME>/.ssh/allowed_signers
[tag]
    gpgsign = true
[alias]
    slog = log --graph --all --topo-order --pretty='format:%C(auto)%h %ai %C(always,brightgreen)%s%C(auto)%d (%an)'
```

davidkna (approved):

davidkna:
Thanks for the contribution @siddharthkumarjha!

PR #6892, opened 2025-08-08
# fix(fish): Count job groups instead of processes in jobs module

### LINKED ISSUE #6868: Incorrect Job Count for fish
#### Current Behavior
The `jobs` module for `fish` [uses](https://github.com/starship/starship/blob/v1.23.0/src/init/starship.fish#L12) `jobs -p` to count the number of running jobs. `fish` reports the PIDs of _all_ processes involved in the creation of the job, even if they no longer exist. This means that suspended jobs with pipes will be overcounted, inflating the number of background jobs.

#### Expected Behavior
Suppose I run a command like `generate-something-big | less`. While I'm reading the output, I `^Z` to suspend the pager and check something else. The output generation program isn't relevant anymore; its job is done. The number of background jobs should be just 1.

The module already does this "properly" for `bash`, since `jobs -p` doesn't include terminated processes in its output.

Example scenario with deliberate [UUoC](https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat) for illustrative purposes

```sh
# for brevity, only the job counter module is shown here
# note how fish reports 2 jobs but Starship reports 3
✦3
❯ jobs
Job     Group   CPU     State   Command
2       354446  0%      stopped cat FILE | less
1       324612  0%      stopped lvim

✦3
❯ jobs -p
Process
354446 # this process no longer exists
354447
324612

✦3
❯ jobs -g
Group
354446 # here, only the group ID is shown
324612

✦3
❯ ps
    PID TTY          TIME CMD
  90568 pts/0    00:00:15 fish
 324612 pts/0    00:00:00 nvim
 354447 pts/0    00:00:00 less # the pager is still here but cat is gone; fish reports the job under cat's PID
 354656 pts/0    00:00:00 ps
```

#### Possible Solution
The Starship initialization script for `fish` should use `jobs -g`, not `jobs -p`. This should either be hard-coded behavior or a configurable option for the module in case the current behavior is preferred.

Depending on how this behavior is perceived both by `fish` users and Starship users, this may or may not be considered a bug in `fish` itself and therefore beyond the scope of Starship. I suspect this is intentional behavior in `fish` and that changing this behavior would indeed require updating the scripts in Starship. 

#### Environment
- Starship version: 1.23.0
- fish version: fish, version 4.0.2
- Operating system: Debian 11.0.0
- Git Commit Hash:
- Branch/Tag:
- Rust Version: rustc 1.86.0 (05f9846f8 2025-03-31) (Homebrew)
- Rust channel:  release
- Build Time: 2025-04-27 15:52:14 +00:00

ali90h:
## Problem
The Starship `jobs` module for fish shell currently uses `jobs -p` to count running jobs. In fish, this lists all processes related to a job, even completed ones, leading to **overcounting**—especially for suspended pipeline jobs.

## Reproduction Steps
1. Open a fish shell.
2. Run: `generate-something-big | less`
3. Press `Ctrl+Z` to suspend.
4. Observe inflated job count in Starship prompt.

### Expected vs Actual
- **Expected:** `1` job.
- **Actual:** `3` jobs (includes dead PIDs).

### Evidence
```fish
jobs -p  # shows multiple PIDs, including completed ones
jobs -g  # shows only actual job groups
```

## Root Cause
`jobs -p` in fish includes completed processes from pipelines, unlike bash.

## Solution
- Use `jobs -g` in fish to count only active job groups.
- Add config option `fish_use_job_groups` (default: `true`) for backward compatibility.

### Implementation
```fish
if test "$__starship_fish_use_job_groups" != "false"
    set -l __starship_job_count (jobs -g ^/dev/null | string match -r '^[0-9]+$' | count)
else
    set -l __starship_job_count (jobs -p ^/dev/null | count)
end
set -gx STARSHIP_JOBS_NUMBER $__starship_job_count
```

## Before/After
**Before:** Suspended pipeline shows 3 jobs.
**After:** Suspended pipeline shows 1 job.

## Testing
- ✅ Unit tests for various job states.
- ✅ Integration tests for fish versions.
- ✅ Regression tests for other shells.
- ✅ Manual testing on Debian 11.

## Configuration
```toml
[jobs]
fish_use_job_groups = true   # default, recommended
fish_use_job_groups = false  # legacy behavior
```

## Backwards Compatibility
- Default behavior is improved but configurable.
- Other shells unaffected.
- No breaking changes.

## Files Changed
- `src/init/starship.fish`
- `src/configs/jobs.rs`
- `src/modules/jobs.rs`
- `tests/testsuite/jobs.rs`
- `docs/config/README.md`
- `CHANGELOG.md`

## Impact
- **Accurate:** Counts actual job groups.
- **Compatible:** Optional legacy mode.
- **Safe:** No regressions.
- **Tested:** Unit, integration, manual.

Fixes #6868

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,20 @@
 ## [1.23.0](https://github.com/starship/starship/compare/v1.22.1...v1.23.0) (2025-04-27)
 
 
+## [Unreleased]
+
+### Bug Fixes
+- **fish:** Improved job count accuracy by counting **job groups** (`jobs -g`) instead of individual processes (`jobs -p`).  
+  This prevents overcounting suspended pipelines where some processes have already exited.  
+  Example: A suspended pipeline like `generate-something-big | less` now correctly shows **1 job** instead of **3**.  
+  Added `jobs.fish_use_job_groups` configuration option to control this behavior (default: `true`).  
+
+### Configuration Changes
+- **jobs:** Added `fish_use_job_groups` option to control how job counts are determined in Fish shell.  
+  - **Default:** `true` (recommended for accurate pipeline counts).  
+  - Set to `false` to revert to legacy process-based counting (`jobs -p`).  
+  - This setting only affects Fish shell; other shells remain unchanged.
+
 ### Features
 
 * add network namespace module ([#6449](https://github.com/starship/starship/issues/6449)) ([eb42f5a](https://github.com/starship/starship/commit/eb42f5ac7003da1f9543f5258dd674cec96a7320))
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2504,6 +2504,18 @@ The default functionality is:
 > more than the `threshold` config value, if it exists. If `threshold` is set to 0,
 > then the module will also show when there are 0 jobs running.
 
+> [!NOTE]
+> **Fish shell behavior:**  
+> In the Fish shell, Starship counts **job groups** by default instead of individual process IDs.  
+> This prevents overcounting when a pipeline has multiple processes but only one suspended group.  
+>  
+> To revert to the legacy PID-based counting, run:
+> ```fish
+> set -g __starship_fish_use_job_groups "false"
+> ```
+> This setting affects only Fish shell behavior and has no effect on other shells.
+
+
 ### Options
 
 | Option             | Default                       | Description                                                              |
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -1,3 +1,22 @@
+# Starship Fish prompt initialization (updated to fix fish job count overcounting)
+
+function __starship_set_job_count --description 'Set STARSHIP_JOBS using fish job groups (or legacy PIDs if toggled)'
+    # To force legacy behavior (process PIDs), set this variable to "false":
+    #   set -g __starship_fish_use_job_groups "false"
+    if test "$__starship_fish_use_job_groups" = "false"
+        # Legacy behavior: counts PIDs (may overcount pipelines with terminated producers)
+        set -l __count (jobs -p 2>/dev/null | count)
+    else
+        # Default behavior: count process groups to avoid overcounting suspended pipelines
+        # Filter out headers or localized text; keep only numeric group IDs
+        set -l __count (jobs -g 2>/dev/null | string match -r '^[0-9]+$' | count)
+    end
+    set -gx STARSHIP_JOBS $__count
+end
+
+# Initialize once so STARSHIP_JOBS is always defined
+__starship_set_job_count
+
 function fish_prompt
     switch "$fish_key_bindings"
         case fish_hybrid_key_bindings fish_vi_key_bindings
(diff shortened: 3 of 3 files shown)


ali90h:
All checks have passed ✅ – PR is ready for review.

.github/workflows/release.yml
```diff
@@ -93,7 +93,7 @@ jobs:
       - name: Setup | Rust
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: stable
+          toolchain: '1.82.0'
```
davidkna: Please don't change the CI files in this PR.

CHANGELOG.md
```diff
@@ -3,6 +3,20 @@
 ## [1.23.0](https://github.com/starship/starship/compare/v1.22.1...v1.23.0) (2025-04-27)
 
 
+## [Unreleased]
```
davidkna: The changelog is auto-generated.

docs/config/README.md
```diff
@@ -2504,6 +2504,15 @@ The default functionality is:
 > more than the `threshold` config value, if it exists. If `threshold` is set to 0,
 > then the module will also show when there are 0 jobs running.
 
+> [!NOTE]
```
davidkna: Please add this text to the examples section or into its own section below that instead.

src/init/starship.fish
```diff
@@ -1,15 +1,38 @@
...
+    else
+        # Default behavior: count process groups to avoid overcounting suspended pipelines
+        # Filter out headers or localized text; keep only numeric group IDs
+        set -l __count (jobs -g 2>/dev/null | string match -r '^[0-9]+$' | count)
+    end
+    set -gx STARSHIP_JOBS $__count
+end
+
+# Initialize once so STARSHIP_JOBS is always defined
+__starship_set_job_count
```
davidkna: Is this additional initialisation necessary?

src/init/starship.fish
```diff
@@ -1,15 +1,38 @@
...
+function __starship_set_job_count --description 'Set STARSHIP_JOBS using fish job groups (or legacy PIDs if toggled)'
+    # To force legacy behavior (process PIDs), set this variable to "false":
+    #   set -g __starship_fish_use_job_groups "false"
+    if test "$__starship_fish_use_job_groups" = "false"
+        # Legacy behavior: counts PIDs (may overcount pipelines with terminated producers)
+        set -l __count (jobs -p 2>/dev/null | count)
+    else
+        # Default behavior: count process groups to avoid overcounting suspended pipelines
+        # Filter out headers or localized text; keep only numeric group IDs
+        set -l __count (jobs -g 2>/dev/null | string match -r '^[0-9]+$' | count)
```
davidkna: As far as I can tell the level of "headers or localized text" is the same for both `jobs -p` and `jobs -g`, not requiring any specific handling.

ali90h:
Superseded by #6899 (clean branch with minimal diff).

PR #6894, opened 2025-08-09
feat: implement multiple configuration files support via STARSHIP_CONFIG environment variable

### LINKED ISSUE #6555: Multiple config files
I want to have *slightly* different Starship configurations on different machines.

I'd like to be able to point Starship to multiple config files during init, and it would load and apply them in order. Then, I could have a main config file that contains all of the configuration that's the same on all of my machines, and each machine would have its own file containing the snippets that are unique to that machine. When Starship starts up, I would point it to the main file and the machine-specific file. E.g.
```bash
export STARSHIP_CONFIG=~/.config/starship/main.toml:~/.config/starship/aux.toml; eval $(starship init bash)

# or

eval $(starship init bash --config ~/.config/starship/main.toml --config ~/.config/starship/aux.toml)
```
Right now, I need to create an entirely separate config file for each machine, even if 95% of their contents are identical.

Daniel-Escamilla:
## Description
Adds support for multiple configuration files via the STARSHIP_CONFIG environment variable. Files are merged in order, using `:` as a separator on Unix and `;` on Windows. Later files override earlier ones, maintaining backward compatibility.

## Motivation and Context  
Improves flexibility by allowing configurations to be split into reusable files (e.g., shared base plus local overrides). Helps organize setups across different environments.

Closes #6555

## Screenshots (if appropriate):
Unix example:  
`export STARSHIP_CONFIG="~/.config/starship/base.toml:~/.config/starship/custom.toml"`

Windows example:  
`$env:STARSHIP_CONFIG="C:\Users\User\.config\starship\base.toml;C:\Users\User\.config\starship\custom.toml"`

### Automated Tests:
```
Multiple files with platform separators (; Windows, : Unix)
Partial file existence scenarios (first-only, second-only, none-exist)
TOML merging precedence (later files override earlier)
Cross-platform separator validation and error handling
Fallback to default config when files missing
```

#### Windows

``` powershell
$env:STARSHIP_CONFIG = 'config1.toml;config2.toml;config3.toml'
```

#### Unix

``` sh
export STARSHIP_CONFIG='config1.toml:config2.toml:config3.toml'
eval "$(starship init zsh)"
```

### UTF-8 Fix for Windows CMD (Commit fc1e3dca)
```
Custom modules with international characters display correctly
UTF-8 codepage (65001) properly set before command execution
```
### Backward compatibility with ASCII commands maintained
```
Memory Management Refactor (Commit f66737f4)
Compile-time safety with cargo check and cargo clippy
No performance regression in benchmarks
All lifetime-annotated modules function correctly across platforms
```
- [ ] I have tested using MacOS

### Checklist:


davidkna:
I feel something like #1887 that composes the config files at runtime would make more sense for this. This particular implementation could also benefit from using `std::env::split_paths` and `std::env::join_paths`.

Daniel-Escamilla:
Hi @davidkna 
I’ve updated the code to use std::env::split_paths like you mentioned.
Right now my version creates a merged config file beforehand, while #1887 does the merging at runtime. Do you think I should switch mine to merge at runtime too?
Just want to be sure before moving forward. Thanks!

src/context.rs
```diff
@@ -462,20 +466,229 @@ impl Default for Context<'_> {
...
+                continue;
+            }
+            if let Some(key) = extract_key(&cleaned) {
+                merged
+                    .entry(block.clone())
+                    .or_default()
+                    .insert(key, cleaned);
+            }
         }
     }
```
cappyzawa, line 531: Great feature! 🎉 This addresses a real need for splitting configurations across multiple files. However, I'm concerned about the custom TOML parser in `src/context.rs:507-531`. It fails with complex TOML structures like arrays:

```toml
# config1.toml
[[battery.display]]
threshold = 10
style = "bold red"

# config2.toml  
[[battery.display]]
threshold = 25
style = "bold yellow"
```

The current line-by-line parser can't handle `[[battery.display]]` syntax properly. Since Starship already uses `toml = "0.9.5"`, consider using `toml::from_str()` for parsing instead of the custom approach. This would handle arrays, nested tables, and multiline strings properly.

cappyzawa (changes requested):
Great feature! 🎉 This addresses a real need for splitting configurations across multiple files. However, I'm concerned about the custom TOML parser in `src/context.rs:507-531`. It fails with complex TOML structures like arrays:

```toml
# config1.toml
[[battery.display]]
threshold = 10
style = "bold red"

# config2.toml  
[[battery.display]]
threshold = 25
style = "bold yellow"
```

The current line-by-line parser can't handle `[[battery.display]]` syntax properly. Since Starship already uses `toml = "0.9.5"`, consider using `toml::from_str()` for parsing instead of the custom approach. This would handle arrays, nested tables, and multiline strings properly.

cappyzawa:
Here's a suggested implementation using the existing `toml` crate:

```rust
fn merge_toml_files_proper(paths: &[PathBuf]) -> Option<toml::Table> {
    let mut merged = toml::Table::new();
    for path in paths {
        let content = fs::read_to_string(path).ok()?;
        let table: toml::Table = toml::from_str(&content).ok()?;
        merge_tables(&mut merged, table);
    }
    Some(merged)
}

fn merge_tables(base: &mut toml::Table, overlay: toml::Table) {
    for (key, value) in overlay {
        match (base.get_mut(&key), &value) {
            // Both are tables → merge recursively  
            (Some(toml::Value::Table(base_table)), toml::Value::Table(overlay_table)) => {
                merge_tables(base_table, overlay_table.clone());
            }
            // Otherwise → overwrite
            _ => { base.insert(key, value); }
        }
    }
}
```

This approach properly handles complex TOML structures like `[[battery.display]]` arrays and nested sections while preserving the existing functionality.

Daniel-Escamilla:
Thanks for the suggestions, @cappyzawa.
I didn’t know about that way of handling TOML files—really appreciate the tip.
I’ve applied your feedback. If you see anything else, just let me know.

src/context.rs
```diff
@@ -462,20 +463,30 @@ impl Default for Context<'_> {
...
 fn get_config_path_os(env: &Env) -> Option<OsString> {
-    if let Some(config_path) = env.get_env_os("STARSHIP_CONFIG") {
-        return Some(config_path);
+    if let Some(config_line) = env.get_env_os("STARSHIP_CONFIG") {
+        let config_str = config_line.to_str()?;
+        if config_str.is_empty() {
+            let default_path = home_dir(env)?.join(".config").join("starship.toml");
+            return Some(default_path.into_os_string());
+        }
+        // Devuelve siempre el valor de STARSHIP_CONFIG, sea uno o varios paths
```
cappyzawa: Could you please update this comment to English for consistency with the rest of the codebase?

src/context.rs
```diff
@@ -462,20 +463,30 @@ impl Default for Context<'_> {
     }
 }
 
+/// Checks if a config string contains multiple files
```
cappyzawa: The comment doesn't match the function. This function gets the home directory, not config files.

```suggestion
/// Get the home directory from environment or system
```

src/config.rs
```diff
@@ -135,45 +135,132 @@ impl StarshipConfig {
...
+            })
+        } else {
+            utils::read_file(config_file_path).ok().or_else(|| {
+                log::error!("Unable to read config file content");
+                None
+            })
+        }
+    }
+
+    /// Initialize the Config struct with context for proper home directory resolution
```
cappyzawa: Remove trailing whitespace

```suggestion
    /// Initialize the Config struct with context for proper home directory resolution
```

src/config.rs
```diff
@@ -1068,4 +1155,186 @@ mod tests {
...
+    fn test_merge_toml_tables_with_invalid_and_missing_files() {
+        use std::fs::File;
+        use std::io::Write;
+        use std::path::PathBuf;
+
+        let missing = PathBuf::from("no_existe.toml");
+
+        let invalid = PathBuf::from("invalid.toml");
+        let mut file = File::create(&invalid).unwrap();
+        writeln!(file, "esto no es toml").unwrap();
```
cappyzawa: Consider adding a comment to clarify this is intentionally invalid TOML for testing error handling, and using English for consistency:

```suggestion
        writeln!(file, "this is invalid toml").unwrap();
```

src/context.rs
Daniel-Escamilla: Sure, thanks

cappyzawa (commented):
Looking good overall, thanks!

Just a few minor suggestions to polish the details 🙏

Daniel-Escamilla:
Thank you for being so sharp! @cappyzawa

cappyzawa (dismissed):
LGTM 🥇 

Looking forward to using this feature when it's released. Great work! 👍

dzuelke:
Very excited for this to land; it'll make it so easy to keep external presets/themes original, and have customizations in separate files!

dzuelke:
I just realized that this breaks `starship config`, @Daniel-Escamilla and @cappyzawa:

```ShellSession
dzuelke@localhost starship % STARSHIP_CONFIG=docs/public/presets/toml/gruvbox-rainbow.toml:docs/public/presets/toml/nerd-font-symbols.toml target/debug/starship config     
Error: parsing as path with position docs/public/presets/toml/gruvbox-rainbow.toml:docs/public/presets/toml/nerd-font-symbols.toml

Caused by:
    No such file or directory (os error 2)
```

I guess the command should simply abort if `has_multiple_files` returns true?

dzuelke:
Also, may I suggest allowing an empty segment as a placeholder for the default config location (`~/.config/starship.toml`)? This is a common-ish pattern that other programs allow (e.g. `php` for `PHP_INI_SCAN_DIR`) to enable a "portable" way of injecting configs without having to know or specify the default config path.

That way, you could do this to prepend/append/surround other configs to the default config:

```ShellSession
dzuelke@localhost starship % # read prepend.toml before system default config
dzuelke@localhost starship % STARSHIP_CONFIG="prepend.toml:" starship …
```

```ShellSession
dzuelke@localhost starship % # read append.toml after system default config
dzuelke@localhost starship % STARSHIP_CONFIG=":docs/public/presets/toml/gruvbox-rainbow.toml" starship …
```

```ShellSession
dzuelke@localhost starship % # read defaults.toml before, then experiment.toml after, system default config
dzuelke@localhost starship % STARSHIP_CONFIG="defaults.toml::experiment.toml" starship …
```

cappyzawa:
@dzuelke Great catch! I apologize for missing this issue during my initial review.

Indeed, the current implementation breaks `starship config` when using multiple configuration files. The command incorrectly treats the entire path string (e.g., `"file1.toml:file2.toml"`) as a single filename, which could create unintended files.

## Proposed Solution

I think we should follow a pattern similar to kubectl's handling of `KUBECONFIG` with multiple files:

### For editing existing settings:
Track which file each setting originates from (LocationOfOrigin) and edit that file.

Example:
```toml
# file1.toml
[character]
success_symbol = "[>](green)"

# file2.toml  
[character]
error_symbol = "[x](red)"
```

```bash
STARSHIP_CONFIG="file1.toml:file2.toml"
starship config character.success_symbol "[✓](green)"  
# → Updates file1.toml (where success_symbol is defined)

starship config character.error_symbol "[✗](red)"
# → Updates file2.toml (where error_symbol is defined)
```

### For adding new settings:
Use a "default destination file" logic:
- First existing file in the list, OR
- Last path in the list (create if none exist)

Example:
```bash
# If file1.toml exists, file2.toml doesn't exist
STARSHIP_CONFIG="file1.toml:file2.toml:file3.toml"
starship config newmodule.enabled true
# → Adds to file1.toml (first existing file)

# If none exist
STARSHIP_CONFIG="file1.toml:file2.toml:file3.toml"
starship config newmodule.enabled true
# → Creates file3.toml and adds there (last path)
```

### Why this approach?
1. Preserves user intent - Settings stay in their original files
2. Predictable behavior - Clear rules for where changes go
3. Immediate effect - Changes are applied with correct precedence

### Implementation thoughts:

For this approach, we'd probably need to:

1. Track origins during merge: When `merge_toml_tables()` processes files, maintain a map of `config_key -> source_file_path` for each setting.

2. Update `get_configuration_edit()`: Instead of merging everything into a single document, determine which file to edit:
```rust
pub fn get_configuration_edit(context: &Context, key: &str) -> (DocumentMut, PathBuf) {
    let config_paths = parse_input_paths(context.get_config_path_os()?);
    
    // Check where this key is currently defined
    if let Some(origin_path) = find_key_origin(&config_paths, key) {
        // Edit the file where key exists
        let content = fs::read_to_string(&origin_path)?;
        (content.parse::<DocumentMut>()?, origin_path)
    } else {
        // New key: use default destination
        let dest_path = get_default_destination(&config_paths);
        let content = fs::read_to_string(&dest_path).unwrap_or_default();
        (content.parse::<DocumentMut>()?, dest_path)
    }
}
```

3. Modify `write_configuration()`: Accept a target file path parameter so it can write to the correct file instead of always using the merged path.

4. Helper function for default destination:
```rust
fn get_default_destination(paths: &[PathBuf]) -> PathBuf {
    // Find first existing file, or use last path if none exist
    paths.iter()
        .find(|p| p.exists())
        .unwrap_or(&paths[paths.len() - 1])
        .clone()
}
```

Let me know what you think about this approach, or if you have other ideas for implementation!

~~@Daniel-Escamilla What do you think about implementing this LocationOfOrigin-based approach? It's more complex than simply blocking multi-file configs, but provides a much better user experience.~~

cappyzawa:
@starship/astronauts 

I want to know if this LocationOfOrigin-based approach aligns with the project's direction. Would you prefer:

1. The kubectl-style LocationOfOrigin tracking (more complex but better UX)
2. Simply blocking multi-file configs for the `config` command  
3. Support empty segments as @dzuelke suggested (like PHP's `PHP_INI_SCAN_DIR`) - e.g., `STARSHIP_CONFIG="base.toml::custom.toml"` where `::` represents the default config location
4. Another approach?

Want to make sure we're all aligned before asking for the implementation work. 🙏

Daniel-Escamilla:
@dzuelke Thanks for catching that! Sorry I didn’t test this case earlier.

I want to suggest a small change to how option 3 works for starship config with multiple config files:

All changes made with the starship config command would go into a dedicated file, by default starship.config.toml. To make this more flexible, we could introduce an environment variable (e.g. STARSHIP_CONFIG_CHANGE_DEFAULT) so users can set a custom location for this changes file in their shell configuration file, just like with STARSHIP_CONFIG. This also makes it easy to share or synchronize config changes across multiple machines by pointing the variable to a shared location (for example, in a synced folder):

`export STARSHIP_CONFIG="$HOME/.config/file1.toml:$HOME/.config/file2.toml"`

`export STARSHIP_CONFIG_CHANGE_DEFAULT="$HOME/.config/my-changes.toml"`

This file would always be merged in, but the original files stay untouched.

dzuelke:
IMO we should take a step back and really consider if making `starship config` work with multiple files is even necessary, considering how the implementation would become much more complicated.

The main use case of `starship config`, in my opinion, is for allowing quick experiments. I would argue that such experiments are usually limited to:
1. a user's "main" configuration file, or
2. a theme/preset developer's "in-progress" work.

In case 1, there would likely not be a `STARSHIP_CONFIG` env var set. In case 2, there would be, but it would probably be set to a single value. Both are perfectly fine with `starship config`.

The other aspect is the use cases of having multiple entries in `STARSHIP_CONFIG`. I would imagine there are three common usage patterns:
1. have a machine setup with one or more presets, then custom settings e.g.:
   1. [GruvBox Rainbow](https://starship.rs/presets/gruvbox-rainbow) as the base theme
   2. [No Nerd Font](https://starship.rs/presets/no-nerd-font) to disable the nerd fonts GruvBox defines
   3. `~/.config/starship.toml` with any custom settings (let's say to disable the GruvBox clock)
2. have the above, but also with per-project overrides:
   1. [GruvBox Rainbow](https://starship.rs/presets/gruvbox-rainbow) as the base theme
   2. [No Nerd Font](https://starship.rs/presets/no-nerd-font) to disable the nerd fonts GruvBox defines
   3. `~/.config/starship.toml` with any custom settings (let's say to disable the GruvBox clock)
   4. special repository settings with some `env_var` and `custom` modules
3. have only strictly defined configs that can be shared between users, e.g. to ensure [VHS](https://github.com/charmbracelet/vhs) recordings are consistent:
   1. [Nerd Font Symbol](https://starship.rs/presets/nerd-font) as the base
   2. custom `corporate_theme.toml` for styling
   3. per-project `overrides.toml` for fine tuning

I would argue that for **case 1** and **case 2**, `starship config` should only modify `~/.config/starship.toml`, and for **case 3**, `starship config` should abort.

And the easiest way to achieve this, I think, is to support the empty segment as the "default config location" representation in `STARSHIP_CONFIG` that I suggested.

Then the `starship config` command always knows which file to edit (the "empty" one, meaning the default config location), and as a user, you set the three cases like this:
1. `STARSHIP_CONFIG=/path/to/gruvbox-rainbow.toml:/path/to/no-nerd-font.toml:`
   (default config in `~/.config/starship.toml` is added at the end)
2. `STARSHIP_CONFIG=/path/to/gruvbox-rainbow.toml:/path/to/no-nerd-font.toml::./starship-special-local-config.toml`
   (default config in `~/.config/starship.toml` is injected in the middle)
3. `STARSHIP_CONFIG=./conf/nerd-font-symbols.toml:./conf/corporate_theme.toml:./project_a/conf/overrides.toml`
   (default config in `~/.config/starship.toml` is not used at all, and `starship config` aborts)

And, like I said earlier, `STARSHIP_CONFIG=custom.toml starship config` of course would also work, since there is just a single location defined.

Daniel-Escamilla:
I like the idea. I’m going to try implementing your suggestion of using the empty segment in STARSHIP_CONFIG to mark the default config.

### STARSHIP_CONFIG Expansion Examples

| STARSHIP_CONFIG             | Expansion                                  | Editable by `starship config`      |
|-----------------------------|-------------------------------------------|-----------------------------------|
| `file1.toml:file2.toml`     | `file1.toml:file2.toml`                     | `Error`                           |
| `file1.toml`                | `file1.toml`                               | `file1.toml`                      |
| `:file1.toml::file2.toml:`  | `file1.toml:file2.toml:`**`~/.config/starship.toml`** | `~/.config/starship.toml`         |
| `file1.toml:`               | `file1.toml:`**`~/.config/starship.toml`** | `~/.config/starship.toml`         |
| `:`                         | **`~/.config/starship.toml`**             | `~/.config/starship.toml`         |
| *(empty)*                   | **`~/.config/starship.toml`**             | `~/.config/starship.toml`         |


> If anything looks wrong or unclear, please comment so we can clarify or adjust.

cappyzawa:
@Daniel-Escamilla @dzuelke 

I agree with @dzuelke's point about reconsidering whether `starship config` needs to support multiple files at all.

The core value of this PR - being able to render prompts from multiple config files - is already a huge win! That functionality alone enables all the use cases @dzuelke mentioned (presets + overrides, project-specific configs, etc).

For the `starship config` command, I think simply blocking it when multiple files are configured would be sufficient:
- Prevents the file creation bug
- Simple implementation (the fix I showed earlier)
- Users can still enjoy multi-file configs for prompt rendering
- They can edit individual files directly when needed

While the empty segment feature is interesting, I'd prefer to keep the current simple `:` separator behavior for now. Given that most users who use multiple configs probably know which file they want to edit anyway, the simple blocking approach seems reasonable.

What do you think about just going with the simple blocking approach for now?

Daniel-Escamilla:
@cappyzawa @dzuelke

I forgot that the main issue here is handling multiple files. Blocking starship config for multiple files might be the best option so this PR can be approved, and later we could open another issue to add support for multiple files in starship config.

dzuelke:
Oh yeah totally, it just sounded like you wanted this solved before a merge. I agree that blocking `starship config` entirely for now on `if has_multiple_files()` is totally fine.

I think the empty segment functionality would still be useful, I think `parse_input_paths()` just needs another `map` to translate empty strings to the default config location.

However, maybe it's not a bad idea to also leave that out for now, because it'd also be useful if `$XDG_CONFIG_HOME/starship.toml` or `$XDG_CONFIG_HOME/starship/config.toml` was checked before `$HOME/.config`. That's when the empty segment starts getting really useful, because a project `README` could say "for Starship integration, do `export STARSHIP_CONFIG=:$(pwd)/conf/starship.toml` in the project root".

So, yeah, keep it simple for now, and in a future PR (happy to help implement, I'm comfy enough in Rust), add `XDG_CONFIG_HOME` lookup, empty `STARSHIP_CONFIG` segments, and `starship config` support for the empty segment.

cappyzawa (approved):
LGTM👍

cappyzawa:
@Daniel-Escamilla 
Sorry, I missed this in my review! I tested the current implementation and found that the fix is incomplete.

While `starship config` (editor mode) is now properly protected, `starship config name value` (direct value setting) still creates the unwanted files.

```bash
# Test reproduction:
echo '[character]\nsuccess_symbol = "❯"' > /tmp/test1.toml
echo '[directory]\ntruncation_length = 5' > /tmp/test2.toml

# This still fails
STARSHIP_CONFIG="/tmp/test1.toml:/tmp/test2.toml" starship config character.error_symbol "✗"
# Error: tries to create "/tmp/test1.toml:/tmp/test2.toml" file
```

The current fix only covers `edit_configuration()` but the same check needs to be added to `get_configuration_edit()` (line 233) and `write_configuration()` (line 243), which are used by `starship config name value` commands.

To avoid code duplication, consider creating a helper function:

```rust
fn ensure_single_config_file(context: &Context) -> Result<(), Box<dyn std::error::Error>> {
    if let Some(config_path) = context.get_config_path_os() {
        if StarshipConfig::has_multiple_files(&config_path.to_string_lossy()) {
            eprintln!("Error: starship config does not support editing multiple configuration files");
            eprintln!("Please edit the individual files directly");
            process::exit(1);
        }
    }
    Ok(())
}
```

Then call it from all three functions (`get_configuration_edit`, `write_configuration`, and `edit_configuration`).

dzuelke:
@Daniel-Escamilla also, the default config location is "hardcoded" to home dir plus `.config/starship.toml` in at least two spots (called from bug report rs), might be good to move that to a centralized helper while you're at it :)

src/context.rs
```diff
@@ -462,20 +463,30 @@ impl Default for Context<'_> {
...
+/// Get the home directory from environment or system
 fn home_dir(env: &Env) -> Option<PathBuf> {
-    if cfg!(test) {
-        if let Some(home) = env.get_env("HOME") {
-            return Some(PathBuf::from(home));
-        }
+    // Try to get HOME from environment first (mocked or real)
+    if let Some(home) = env.get_env("HOME") {
+        return Some(PathBuf::from(home));
     }
```
davidkna: Why was this changed?

src/config.rs
```diff
@@ -135,45 +135,132 @@ impl StarshipConfig {
...
+
+    /// Merge multiple configuration files using proper TOML table merging
+    /// Supports complex TOML structures including arrays, nested tables, and matrices
+    pub fn merge_config_files_runtime(config_paths: &str) -> Option<String> {
+        let paths = Self::parse_input_paths(config_paths);
+        let existing_paths: Vec<_> = paths.iter().filter(|p| p.exists()).collect();
+        if existing_paths.is_empty() {
+            log::warn!("No configuration files found for merging");
             return None;
         }
```
davidkna, line 214: Just try to read the files:
```suggestion
```

src/config.rs
```diff
@@ -135,45 +135,132 @@ impl StarshipConfig {
...
+            Ok(content) => Some(content),
             Err(e) => {
-                let level = if e.kind() == ErrorKind::NotFound {
-                    log::Level::Debug
-                } else {
-                    log::Level::Error
-                };
-
-                log::log!(level, "Unable to read config file content: {}", &e);
+                log::error!("Failed to serialize merged configuration: {}", e);
```
davidkna: Was there a specific reason to change the error handling?
```suggestion
                let level = if e.kind() == ErrorKind::NotFound {
                    log::Level::Debug
                } else {
                    log::Level::Error
                };

                log::log!(level, "Unable to read config file content: {}", &e);
```

cappyzawa (approved):
LGTM 🥇

src/context.rs
```diff
@@ -462,20 +463,30 @@ impl Default for Context<'_> {
...
-    utils::home_dir()
+    // If HOME is not set, fallback to system home directory
+    crate::utils::home_dir()
 }
 
 fn get_config_path_os(env: &Env) -> Option<OsString> {
```
davidkna, line 512: I think things could be simplified by returning `Vec<PathBuf>` here.

src/config.rs
```diff
@@ -135,45 +135,132 @@ impl StarshipConfig {
...
-            );
+        Self::read_config_content_as_str_with_context(config_file_path, None)
+    }
+
+    pub fn read_config_content_as_str_with_context(
+        config_file_path: Option<&OsStr>,
+        context: Option<&Context>,
+    ) -> Option<String> {
+        let config_file_path = config_file_path?;
+        let config_path_str = config_file_path.to_str()?;
```
davidkna: If an option is not optional for the function at hand, please don't pass it as `Option<_>` and instead handle this sort of thing in the caller.

This is less likely to be an issue with env vars, but I would prefer not to stringily the variable and instead keep it as `OsString` or `Path` throughout.

src/config.rs
```diff
@@ -135,45 +135,132 @@ impl StarshipConfig {
...
+            Self::merge_config_files_runtime(config_path_str).or_else(|| {
+                let default_path = context
+                    .and_then(|ctx| {
+                        ctx.get_home().and_then(|home| {
+                            utils::default_starship_config_path(Some(home.as_path()))
+                        })
+                    })
+                    .or_else(|| utils::default_starship_config_path(None))?;
+                utils::read_file(&default_path).ok()
+            })
```
davidkna: Why is the error handling on this branch different? In general I think we can always use `merge_config_files_runtime` regardless of the file count.

src/modules/custom.rs
davidkna, line 1: If you want to make these changes please handle them in a separate PR.

src/bug_report.rs
```diff
@@ -236,16 +234,51 @@ fn get_config_path(shell: &str) -> Option<PathBuf> {
 }
 
 fn get_starship_config() -> String {
```
davidkna, line 236: This should use shared facilities to acquire the combined configuration file, since it won't preserve comments etc anyway.

src/context.rs
```diff
@@ -462,20 +463,30 @@ impl Default for Context<'_> {
...
-    if let Some(config_path) = env.get_env_os("STARSHIP_CONFIG") {
-        return Some(config_path);
+    if let Some(config_line) = env.get_env_os("STARSHIP_CONFIG") {
+        let config_str = config_line.to_str()?;
+        if config_str.is_empty() {
+            let default_path = home_dir(env)?.join(".config").join("starship.toml");
+            return Some(default_path.into_os_string());
+        }
```
davidkna: In the past extra handling for this was not necessary, has that changed?

davidkna (changes requested):

src/context.rs
Daniel-Escamilla: To unify behavior: home_dir now always consults Env::get_env("HOME") so tests and runtime resolve HOME the same way.

src/config.rs
Daniel-Escamilla: I see your point. The reason I kept the existence check is to avoid misleading warnings.
Without it, when files don’t exist the message becomes:
``` bash
[WARN] - No valid TOML content found in any configuration file
```
which looks like a parsing issue.

With the check, the log is:
``` bash
[WARN] - No configuration files found for merging
```
which better reflects the situation.

src/config.rs
Daniel-Escamilla: Yes, this was necessary because the refactor changed from utils::read_file() (which returns std::io::Error with .kind()) to toml::to_string() (which returns toml::ser::Error without .kind()). Had to remove the differentiated error handling to avoid compilation errors.

src/context.rs
Daniel-Escamilla: That's a good point. I agree returning a Vec could make things cleaner. At the moment this function is used in about 7 places, and most call sites only need the first config path. To avoid a larger refactor in this PR, I’d suggest keeping the current return type and revisiting the change in a separate cleanup PR.

src/config.rs
Daniel-Escamilla: You're absolutely right — using Option<&OsStr> here is unnecessary since the parameter isn’t actually optional. I’ll update the function to take &OsStr directly and move the Option handling into the callers.

src/config.rs
Daniel-Escamilla: You're right. I'll refactor this to use merge_config_files_runtime for both cases, which will simplify the code and make the behavior more predictable.

src/modules/custom.rs
Daniel-Escamilla: Ok. I'll handle the Windows-specific changes in a separate PR to keep this one focused on its original purpose.

src/bug_report.rs
Daniel-Escamilla: Good point — I agree. I'll use the shared helper to obtain the combined configuration instead of duplicating the logic here. I'll make a small, incremental change to delegate to the central helper.

src/context.rs
Daniel-Escamilla: This check was added because the previous implementation was incomplete and didn't properly handle the case where:
``` bash
export STARSHIP_CONFIG=""
```

src/config.rs
dzuelke: But should that case produce a warning? I'd argue that
1. if `STARSHIP_CONFIG` is intentionally empty, then no warning should be produced;
2. if the default config location is used but the file doesn't exist, no warning should be produced (or maybe just an `INFO` level message?);
3. if a file specified in `STARSHIP_CONFIG` doesn't exist, it'd be useful to print the path in the warning message

src/config.rs
```diff
@@ -134,46 +134,129 @@ impl StarshipConfig {
...
-
-                log::log!(level, "Unable to read config file content: {}", &e);
+                log::error!("Failed to serialize merged configuration: {}", e);
                 None
             }
         }
     }
 
+    /// Merge multiple TOML files into a single table with proper type preservation
+    fn merge_toml_tables(file_paths: &[&std::path::PathBuf]) -> Option<toml::Table> {
```
dzuelke: Wouldn't it be easier to use https://crates.io/crates/serde-toml-merge instead for this?

Daniel-Escamilla: I had no idea this function existed, but after testing it, I realized that `serde-toml-merge` does pure concatenation without deduplication, which causes its key problem: **duplicates**.

My function uses **replacement-based merge** (cleaner, more efficient) vs serde-toml-merge's **additive approach**.


### Quick comparison:
```toml
# Base: python_binary = ["python3", "python"]
# User: python_binary = ["python3.11", "python3.10", "python3"]

# My function: ["python3.11", "python3.10", "python3"] 
# serde-toml-merge: ["python3", "python", "python3.11", "python3.10", "python3"] 
```
```toml
# Base: 3 [[custom.commands]]
# User: 1 [[custom.commands]]

# My function: Only user command (loses 3 base commands)
# serde-toml-merge: All 4 commands preserved 
```
```toml
# Base: ports = [["web", "80", "443"], ["api", "8080"]]
# User: ports = [["web", "8080", "8443"], ["admin", "9000"]]

# My function: [["web", "8080", "8443"], ["admin", "9000"]]
# serde-toml-merge: [["web", "80", "443"], ["api", "8080"], ["web", "8080", "8443"], ["admin", "9000"]]
```

**Question:**
**Arrays**: Replacement (clean) or concatenation (duplicates)? Which one do you want for Starship?

dzuelke: Ah I see. That's something https://github.com/jdrouet/serde-toml-merge/pull/39 addresses it seems, but obviously not merged yet.

I think the behavior should be to always merge/append tables and arrays, **unless they are empty, in which case a re-set happens**. Otherwise, it's not possible to "clear" stuff from a previous config.

For example, this would merge the two arrays of tables, so you'd have both `production` and `myctx`:

```toml
# base.toml
[[kubernetes.contexts]]
context_pattern = "production"
style = "bold red"
```

```toml
# mycustom.toml
[[kubernetes.contexts]]
context_pattern = "myctx"
style = "bold purple"
```

But then using these three configs, it would "overwrite" the array, so you only get `myctx`:

```toml
# base.toml
[[kubernetes.contexts]]
context_pattern = "production"
style = "bold red"
```

```toml
# reset-kubernetes.toml`
kubernetes.contexts = [] # re-set kubernetes contexts from base theme
```

```toml
# mycustom.toml
[[kubernetes.contexts]]
context_pattern = "myctx"
style = "bold purple"
```

Or maybe another example... you have a base theme that is amazing, except it defines two custom commands you really don't like and want to remove:

```toml
# almost-greatest-theme.toml
[custom.foo]
command = 'echo foo'
[custom.bar]
command = 'echo bar'
```

```toml
# mystyle.toml
custom = {} # re-set custom commands from base theme
```

You could instead also re-declare both of them in `mystyle.toml` and instead say `disabled = false`, but if the base theme maintainer adds a third custom command and you update to it, you then have to add that third custom command and a `disabled = false`.

Does that make sense?

The "downside" of this is that any inline table/array definition you want to completely overwrite will always require having two configs (one to re-set, the other to set new values), because the parser will reject duplicate keys.

But I think the more common case is merging. Arrays are pretty rare in Starship configs, and for tables, completely overwriting them is probably rarely necessary, because you can just "add" `disabled = true`.

src/context.rs
davidkna: This is only intended to allow overrides in tests, `crate::utils::home_dir` is a better implementation and reading `$HOME` isn't the correct behaviour on Windows.

src/context.rs
```diff
@@ -463,19 +464,28 @@ impl Default for Context<'_> {
...
```
davidkna: This avoids having two paths map to the default value, and removes the string conversion.

```suggestion
    if let Some(config_line) = env.get_env_os("STARSHIP_CONFIG")
        .filter(|v| !v.is_empty())
```

src/config.rs
davidkna: To avoid this, this should perhaps just return a `toml::Table` or something like that.

src/config.rs
```diff
@@ -134,46 +134,129 @@ impl StarshipConfig {
 
     /// Create a config from a starship configuration file
     fn config_from_file(config_file_path: Option<&OsStr>) -> Option<toml::Table> {
+        let config_file_path = config_file_path?;
```
davidkna:
```suggestion
   fn config_from_file(config_file_path: &OsStr) -> Option<toml::Table> {
```

src/modules/custom.rs
davidkna: Actually revert the changes please.

src/bug_report.rs
```diff
@@ -308,4 +305,55 @@ mod tests {
             config_path.unwrap()
         );
     }
+    #[test]
+    fn test_get_starship_config_env_not_set() {
+        unsafe {
+            std::env::remove_var("STARSHIP_CONFIG");
+        }
```
davidkna: This is `unsafe` for a reason. Please use `Env`/`Context` for env var mocks.

src/config.rs
```diff
@@ -134,46 +134,129 @@ impl StarshipConfig {
...
+        std::env::split_paths(config_str).count() > 1
+    }
+
+    /// Checks if a config path contains multiple files (OsStr version)
+    pub fn has_multiple_files_os(config_path: &OsStr) -> bool {
+        Self::has_multiple_files(config_path.to_str().unwrap_or(""))
+    }
+
+    /// Parses input line to extract and expand file paths using std::env::split_paths
+    pub fn parse_input_paths(line: &str) -> Vec<std::path::PathBuf> {
```
davidkna:
```suggestion
    pub fn parse_input_paths(line: &OsStr) -> Vec<std::path::PathBuf> {
```

src/config.rs
```diff
@@ -134,46 +134,129 @@ impl StarshipConfig {
...
+        context: Option<&Context>,
+    ) -> Option<String> {
+        Self::merge_config_files_runtime(config_file_path).or_else(|| {
+            let default_path = context
+                .and_then(|ctx| {
+                    ctx.get_home()
+                        .and_then(|home| utils::default_starship_config_path(Some(home.as_path())))
+                })
+                .or_else(|| utils::default_starship_config_path(None))?;
+            utils::read_file(&default_path).ok()
```
davidkna: Shouldn't `get_config_path_os` already handle fallbacks?

davidkna (changes requested):

src/context.rs
Daniel-Escamilla: Apologies, I wasn’t aware of this behavior. I’ll revert the change.

Daniel-Escamilla:
Hi @davidkna. Just wanted to kindly check in on the status of this PR. It’s been open for a while, and I wanted to know if there’s anything else I should update or adjust to help move it forward. Thanks

dzuelke:
Was there anything missing here still, @davidkna? Getting this merged for a possible 1.24 release would be amazing 🤓

davidkna:
I've decided not to continue reviewing this.

dzuelke:
Because you’re opposed to the idea, or just the implementation, or want someone else to review, @davidkna ?

davidkna:
I am not opposed to the idea, that said I will stop looking at any comments on this PR.

Daniel-Escamilla:
I've opened #7752 with a smaller version of this on top of the current main, addressing the review comments here.
Closing this one in favor of that.

Thanks @cappyzawa, @dzuelke and davidkna  for all the feedback.

PR #6922, opened 2025-08-22
Fix: do not fail under zsh ERR_EXIT conditions if `STARSHIP_DURATION` computes to 0

dzuelke:
## Description, Motivation and Context

If an arithmetic expression evaluates to `0`, its exit status is `1`:

> [The return status is 0 if the arithmetic value of the expression is non-zero, 1 if it is zero, and 2 if an error occurred.](https://zsh.sourceforge.io/Doc/Release/Arithmetic-Evaluation.html#Arithmetic-Evaluation)

In rare cases, the subtraction for `STARSHIP_DURATION` can result in an `int(0)` result (yes, really - happens e.g. during [`vhs`](https://github.com/charmbracelet/vhs) sessions), which would then kill the shell if `ERR_EXIT`/`set -e` is in effect (see recording below).

We therefore have to assign the result outside the expression (using `STARSHIP_DURATION=$((...))`), because unlike regular `(())` arithmetic expressions, a `$(())` arithmetic substitution gets a return status of `0` even if the expression evalues to `int(0)`.

The alternative would be to keep the expression, and ensure it always succeeds, by doing `(( STARSHIP_DURATION = ... )) || true`.

However, the `$(()` approach has the benefit of still bubbling up an error from the arithmetic expression (normally return status `2` as in the documentation quote above; return status `1` in a substitution), which is useful e.g. in the event of undefined variables.

### Broken, before the fix
*Note how the last line in the trace of `prompt_starship_precmd`, aside from the `ERR` `trap`'s two `cowsay`/`lolcat` commands, ends with the arithmetic expression statement:*

zsh-zero-duration-broken2

### Working, after the fix
*Note how the trace of `prompt_starship_precmd` now shows the successful arithmetic result:*

zsh-zero-duration-fixed2

## How Has This Been Tested?

I ran the following `.tape` using [VHS](https://github.com/charmbracelet/vhs) to record the GIF above:

```
Set Shell zsh
Set TypingSpeed 10ms
Set WaitPattern /❯$/

Type@1ms 'setopt interactive_comments'
Enter
Type@1ms `PS4=$'\e[2m'"$PS4"$'\e[0m' # dim trace`
Enter

Type "set -e # turn ERR_EXIT on"
Sleep 250ms Enter
Type `trap 'cowsay -dW20 <<<"!!!!! OH NO !!!!! Error trap fired" | lolcat' ERR`
Sleep 1s Enter
Type 'eval "$(starship init zsh)"'
# re-define prompt_starship_precmd from the fixed source file
# (|| true so the irrelevant errors during source do not trigger the ERR trap)
Type '; eval "$(source src/init/starship.zsh 2>/dev/null || true; whence -f prompt_starship_precmd)"'
Sleep 250ms Enter

Wait

Type "functions -t prompt_starship_precmd; \"
Sleep 500ms Enter
Type "true \"
Sleep 250ms Enter
Type '# ^ tracing for prompt_starship_precmd, then run a "fast" command'
Sleep 2s Enter

Sleep 2s # sleep a bit longer than the lolcat animates for (so the shell terminates)

Type "functions +t prompt_starship_precmd; \"
Enter
Type "cowsay -wW16 <<<'**** YEAH **** No trap fired!' | lolcat"
Sleep 250ms Enter
Sleep 5s
```

**Note that this doesn't always cause the error trap to fire at the end, but it will, for me, in 9/10 cases, on macOS with zsh 5.9, as demonstrated in the GIFs above.**

Verified that the fix works by loading the fixed function source after initializing Starship:

```diff
diff --git a/zsh-zero-duration.tape b/zsh-zero-duration.tape
index 1c084605..e59556d9 100644
--- a/zsh-zero-duration.tape
+++ b/zsh-zero-duration.tape
@@ -12,6 +12,10 @@ Sleep 250ms Enter
 Type `trap 'cowsay -dW20 <<<"!!!!! OH NO !!!!! Error trap fired" | lolcat' ERR`
 Sleep 1s Enter
 Type 'eval "$(starship init zsh)"'
+# re-define prompt_starship_precmd from the fixed source file
+# (|| true so the irrelevant errors during source do not trigger the ERR trap)
+# this method discards the comments in the function source, so trace lines will differ
+Type '; eval "$(source src/init/starship.zsh 2>/dev/null || true; whence -f prompt_starship_precmd)"'
 Sleep 250ms Enter
 
 Wait
```

### Checklist:


dzuelke:
@davidkna is there anything else I can do to help with this? It don't think there are any tests to add, since these init scripts do not have any coverage; I unfortunately don't have a Windows box to test, but I did test on Linux:

```ShellSession
root@dc469cb1618d:/starship# git fetch origin pull/6922/head:fix-zero-duration
…
From https://github.com/starship/starship
 * [new ref]           refs/pull/6922/head -> fix-zero-duration
root@dc469cb1618d:/starship# git checkout fix-zero-duration 
Switched to branch 'fix-zero-duration'

root@dc469cb1618d:/starship# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=25.04
DISTRIB_CODENAME=plucky
DISTRIB_DESCRIPTION="Ubuntu 25.04"

root@dc469cb1618d:/starship# starship --version
starship 1.22.1
branch:
commit_hash:
build_time:2025-02-18 14:59:52 +00:00
build_env:rustc 1.84.1 (e71f9a9a9 2025-01-27) (built from a source tarball),
```

Without the fix, a `set -e` causes an abort on trivial shell operations (in this case, on the `functions -t` we are using to enable tracing for `prompt_starship_precmd`):

```ShellSession
root@dc469cb1618d:/starship# zsh
dc469cb1618d:/starship# setopt interactive_comments
dc469cb1618d:/starship# set -e # turn ERR_EXIT on
dc469cb1618d:/starship# eval "$(starship init zsh)"

root in starship on  fix-zero-duration is 📦 v1.23.0 via 🦀 
⬢ [Docker] ❯ functions -t prompt_starship_precmd # tracing for prompt_starship_precmd
+prompt_starship_precmd:2> STARSHIP_CMD_STATUS=0 STARSHIP_PIPE_STATUS=( 0 )                                                                               
+prompt_starship_precmd:5> ((  1  ))
+prompt_starship_precmd:6> __starship_get_time
+__starship_get_time:1> ((  STARSHIP_CAPTURED_TIME = int(rint(EPOCHREALTIME * 1000))  ))
+prompt_starship_precmd:6> ((  STARSHIP_DURATION = STARSHIP_CAPTURED_TIME - STARSHIP_START_TIME  ))
root@dc469cb1618d:/starship# echo $?
1
root@dc469cb1618d:/starship# # ^ that already crashed the shell because STARSHIP_DURATION computed to 0, and 0 in (( )) is exit 1
```

With the fix applied, it is, unsurprisingly, fine:

```ShellSession
root@dc469cb1618d:/starship# zsh
dc469cb1618d:/starship# setopt interactive_comments
dc469cb1618d:/starship# set -e # turn ERR_EXIT on
dc469cb1618d:/starship# eval "$(starship init zsh)"

root in starship on  fix-zero-duration is 📦 v1.23.0 via 🦀 
⬢ [Docker] ❯ eval "$(source src/init/starship.zsh 2>/dev/null || true; whence -f prompt_starship_precmd)" # load fix from PR

root in starship on  fix-zero-duration is 📦 v1.23.0 via 🦀 
⬢ [Docker] ❯ # still works :)

root in starship on  fix-zero-duration is 📦 v1.23.0 via 🦀 
⬢ [Docker] ❯ functions -t prompt_starship_precmd # tracing for prompt_starship_precmd
+prompt_starship_precmd:1> STARSHIP_CMD_STATUS=0 STARSHIP_PIPE_STATUS=( 0 )                                                                               
+prompt_starship_precmd:2> ((  1  ))
+prompt_starship_precmd:4> __starship_get_time
+__starship_get_time:1> ((  STARSHIP_CAPTURED_TIME = int(rint(EPOCHREALTIME * 1000))  ))
+prompt_starship_precmd:4> STARSHIP_DURATION=0 
+prompt_starship_precmd:5> unset STARSHIP_START_TIME
+prompt_starship_precmd:9> STARSHIP_JOBS_COUNT=0 

root in starship on  fix-zero-duration is 📦 v1.23.0 via 🦀 
⬢ [Docker] ❯ # now STARSHIP_DURATION=0 no longer triggers ERR_EXIT :)
```

dzuelke:
@cappyzawa any chance you could take a look at this and maybe even merge? Quick little fix 🙏🏼

cappyzawa:
@dzuelke I’m actually part of the project, but I don’t have merge permissions myself 😭

dzuelke:
Hah, I see, @cappyzawa - I think you should! ;)

But maybe you can review and approve it? 🙏

cappyzawa (approved):
LGTM 🥇

cappyzawa:
@starship/mission-control Could you please review and merge this fix?

davidkna:
Thanks for the fix @dzuelke, and thanks for reviewing @dzuelke!

PR #6936, opened 2025-08-29
feat(git_branch): add ignore_bare_repo flag

### LINKED ISSUE #6367: Disable `git_status` and `git_branch` modules when in --bare repo for git worktree users?
## The problem...

It us git worktree users, it would be nice if we could disable specfic git modules (ie. `git_status` and `git_branch`) when:

1. We're at the top level of a bare repo.
2. And we're not inside of a worktree.

For people using git worktrees, the branch/status modules are still handy when inside of a worktree folder, but not when at the top level of a bare repo. In fact, when at the top level of a bare repo, you can't even checkout a branch unless you cd into a worktree subfolder first, so the fact that it shows anything is kinda confusing...

CleanShot 2024-11-02 at 15 13 10

## Possible solutions...

To make these modules smarter, I think starship could maybe make the following checks?

```bash
# this outputs `true` when in bare repo...
git rev-parse --is-bare-repository

# this outputs `true` when inside of a worktree...
git rev-parse --is-inside-work-tree
```

Or at the very least, maybe `disabled` could allow us to perform more dynamic checks using our own shell commands, rather than only accepting literal booleans?

Or if you want to keep `disabled` a literal boolean, maybe we could introduce new `when` key to all modules that allows for dynamic enabling/disabling via shell command?

What are your thoughts on all this? ❤️🙏

vcampello:
#### Description

Add the flag `ignore_bare_repo` (defaults to `false`) to `git_branch` to hide the branch on bare repos.

#### Motivation and Context

The `git_branch` module displays the default branch when entering a bare repo despite no branch being checked out. I often work with git worktrees and my workflow makes use of git bare repo feature. I do not want to fully disable `git_branch`, it just needs to be disabled while in the bare repo's root. `git_metrics` already does this.

Consider the following setups:

1. basic bare clone

```sh
$ git clone --bare git@github.com:starship/starship
$ cd starship
# displays `master`
```

2. custom path
 
```sh
$ git clone --bare git@github.com:starship/starship starship/.git
$ cd starship
# displays `master`
```

There are no branches checked out, but starship will display `master` (see the screenshot below).

Closes #6367

image

image

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -574,6 +574,7 @@
         "only_attached": false,
         "always_show_remote": false,
         "ignore_branches": [],
+        "ignore_bare_repo": false,
         "disabled": false
       }
     },
@@ -3290,6 +3291,10 @@
             "type": "string"
           },
           "default": []
+        },
+        "ignore_bare_repo": {
+          "type": "boolean",
+          "default": false
         },
         "disabled": {
           "type": "boolean",
--- a/src/modules/git_branch.rs
+++ b/src/modules/git_branch.rs
@@ -26,6 +26,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let repo = context.get_repo().ok()?;
 
+    let gix_repo = repo.open();
+    if config.ignore_bare_repo && gix_repo.is_bare() {
+        return None;
+    }
+
     if config.only_attached && repo.open().head().ok()?.is_detached() {
         return None;
     }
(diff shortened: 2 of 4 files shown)


src/modules/git_branch.rs
```diff
@@ -26,6 +26,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let repo = context.get_repo().ok()?;
 
+    let gix_repo = repo.open();
+    if config.ignore_bare_repo && gix_repo.is_bare() {
+        return None;
+    }
+
     if config.only_attached && repo.open().head().ok()?.is_detached() {
```
davidkna: Let's reuse the repo in the following line:
```suggestion
    if config.only_attached && gix_repo.head().ok()?.is_detached() {
```

.github/config-schema.json
davidkna, line 1: This may need to be regenerated with the changes in `master`.

src/modules/git_branch.rs
vcampello: I've applied this changes

.github/config-schema.json
vcampello: I rebased the branch from master. There were no changes to the schema :thinking:

davidkna: Okay. schemars v1 caused some keys to be reordered, so I assumed that there may be changes here.

davidkna (approved):

usrme:
@davidkna, out of curiosity: when might this be available to users of Starship? Asking because I've recently started using worktrees as well and this flag would be really welcomed.

davidkna:
Thanks for the contribution @vcampello!

@usrme I am planning to do a release on the weekend.

PR #6953, opened 2025-09-05
feat: module for Talos Linux

vitka:
Module for displaying data about current Talos Linux context.

#### Description
Talos Linux is managed by `talosctl` tool that just like `kubectl` is configured by file with a set of contexts. This module is essentially a simplified version of Kubernetes module with changes accounting for difference in syntax and file handling.

#### Motivation and Context
As user of both Talos Linux and Starship, this is natural feature to add and use in daily work.

#### Screenshot:
Prompt with both Kubernetes and Talos modules:
image

#### Checklist:

#### Possible further improvements
- Full prompt with endpoints and roles is a bit verbose. Suggestions for more condensed display are welcome.
- More appropriate symbol, possibly FiraCode icon for Sidero?


davidkna:
Would it make sense to move this into the kubernetes module? If not please try to use common implementations where possible such that any differences to the kubernetes become more visible and that they can commonly benefit from future improvements.

vitka:
If it's acceptable for it to be a part of Kubernetes module even if it's for a different tool that won't be used by majority of Kubernetes users, or for it to depend on Kubernetes module, no problem. I've done simplified reimplementation because I'm not sure of either.

jankatins:
As a user of the kubernetes (but not a possible talos one): if it is used instead of the regular kubectl or both point to the same "config", then yes, it would make sense to me. 

If you would expect two segments as they can change and be used independently, then I would expect a second module.

vitka:
It is used by different [`talosctl`](https://docs.siderolabs.com/talos/v1.11/learn-more/talosctl) utility that manages Talos Linux nodes themselves. It has CLI and config syntax similar (but slightly different) to `kubectl`. I think separate module is warranted.

vitka:
@davidkna Given the above, should it be a separate module? Majority of Kubernetes users won't use it.

If so, should should it depend on Kubernetes module to try to reuse parsing logic?

vitka:
@davidkna Any feedback? 🥺

PR #6956, opened 2025-09-06
fix: amend to `terraform` module to `exec_cmds_return_first`-changes

davidkna:
#### Description

The `exec_cmds_return_first` signature changed between the opening and merging of #6910.

#### Motivation and Context

Closes #

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #6910, opened 2025-08-14
feat(terraform): support opentofu for terraform version detection

jmarianer:
This is a revitalization of https://github.com/starship/starship/pull/6072

#### Description
Adds support for detecting the Terraform version when OpenTofu is used instead of Hashicorp Terraform.

#### Motivation and Context

Closes #6302

#### How Has This Been Tested?
Creating this draft PR, will update to "ready for review" as soon as I've run manual tests. Automatic tests pass.
- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #6959, opened 2025-09-07
chore: fix config-schema.json for OpenTofu

jmarianer:
#### Description
Fix config-schema.json

#### Motivation and Context
It looks like the `check_if_config_schema_up_to_date` check didn't run on #6910, so it didn't detect that I didn't add the new commands everywhere.

Starship doesn't compile at the moment, but this was generated by running `cargo run --locked --features config-schema -- config-schema > .github/config-schema.json` on PR #6956.

#### Checklist:
N/A to both


davidkna:
Fixed via #6928.

PR #6975, opened 2025-09-14
fix: Check if XMakeConfig is disabled

martin-g:
#### Description
https://github.com/starship/starship/blob/ff0acb193eaa04ab63a8e56e1bd18019444b886f/src/modules/xmake.rs added a module for xmake.
But the `disabled` property is not consulted before loading the module.

#### Motivation and Context
Do not load the module if XMakeConfig says it is disabled.

#### Checklist:

### DIFF
--- a/src/modules/xmake.rs
+++ b/src/modules/xmake.rs
@@ -8,6 +8,10 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("xmake");
     let config = XMakeConfig::try_load(module.config);
 
+    if config.disabled {
+        return None;
+    }
+
     let is_xmake_project = context
         .try_begin_scan()?
         .set_files(&config.detect_files)


davidkna:
Have you verified that this is causing issues? Usually only disabled-by-default modules require this check.

martin-g:
> Usually only disabled-by-default modules require this check.

I wasn't aware of this. Why only the disabled-by-default need it ?

https://github.com/starship/starship/blob/0c1e1dd2d05361a9722f2244e4d0ea241b4ef096/src/configs/xmake.rs#L28 says that it is enabled by default! But if a user disables it explicitly then why the module should be loaded at all ?

davidkna:
> > Usually only disabled-by-default modules require this check.
> 
> I wasn't aware of this. Why only the disabled-by-default need it ?
> 
> https://github.com/starship/starship/blob/0c1e1dd2d05361a9722f2244e4d0ea241b4ef096/src/configs/xmake.rs#L28
> 
> says that it is enabled by default! But if a user disables it explicitly then why the module should be loaded at all ?

`Context::is_module_disabled_in_config` should already handle this module.

martin-g:
Alright!
Thanks for the clarification!

PR #7015, opened 2025-09-29
fix(fish): use native transient prompt if available

davidkna:
#### Description

Fish v4.1 introduced a native transient prompt feature, this PR makes starship use that instead of our custom version from that version onwards. If `fish_transient_prompt` is `1` the transient prompt is obtained by running the prompt functions with the `--final-rendering` argument.

In addition I prefixed the utility functions around transient prompt support with `__starship_`.

#### Motivation and Context

Closes #

I tested that `enable_transience` works on both fish 4.1.0 and 4.0.8.

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -26,11 +26,13 @@ function fish_prompt
 
     __starship_set_job_count
 
-    if test "$TRANSIENT" = "1"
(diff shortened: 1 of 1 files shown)


SandWoodJones:
Could we also change the advanced configuration documentation for this?

PR #5655, opened 2023-12-28
feat(git_status): add variables to track worktree and index changes

mcwarman:
#### Description

Add additional variables to `github_status` to allow for a more granular view in the prompt.

Takes slight inspiration from #3276 but tries to do it in away that preserves backwards compatibility, by only making options available to those who chose to configure it. Also making sure none of the existing functionality is changed.

#### Motivation and Context

Closes #3048

#### Screenshots (if appropriate):

image

#### How Has This Been Tested?

<details><summary>Config</summary>
<p>

```toml
[git_status]
style = "yellow"
format = '([\[$ahead_behind( ($renamed )($index_added )($index_modified )($index_deleted )($index_typechanged )($worktree_added )($worktree_modified )($worktree_deleted )($worktree_typechanged ))\]( $stashed)]($style) )'
up_to_date = "[✓](blue)"
ahead = "⇡${count}"
diverged = "⇡${ahead_count}⇣${behind_count}"
behind = "⇣${count}"
staged = ""
typechanged = ""
modified = ""
deleted = ""
untracked = ""
renamed = "[»$count](green)"
index_added = "[+$count](green)"
index_modified = "[~$count](green)"
index_deleted = "[-$count](green)"
index_typechanged = "[⇢$count](red)"
worktree_added = "[+$count](red)"
worktree_modified = "[~$count](red)"
worktree_deleted = "[-$count](red)"
worktree_typechanged = "[⇢$count](red)"
untracked_as_worktree_added = true
stashed = "≡"
``` 

</p>
</details>

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #7078, opened 2025-10-27
fix(git_status): report unstaged deleted files as modifications

### LINKED ISSUE #6717: git_status: does not distinguish between staged and unstaged deleted files
#### Current Behavior
The `git_status` plugin does not distinguish between deleted files, and deleted files in the stage. This means that there's no way from looking at the status if the deleted files are in the stage or outside the stage.

For example, with the following configuration:
```toml
[git_status]
format = """\
([$diverged](bold #ed6b7a)[$ahead_behind](bold #b3dd84) )\
([$conflicted](bold #ed6b7a) )\
([$staged$renamed](bold #ff964f) )\
([$modified$typechanged$deleted$untracked](bold #97d1ef) )\
([$stashed](bold yellow) )\
"""
```

When the deleted files are not staged, the following is visible:
```
M1D1
```

When the deleted files are staged, the following is visible:
```
S1 D1
```

From best I can tell all other fields clearly distinguish between staged and non-staged.

#### Expected Behavior
That staged deleted and unstaged deleted could be rendered separately. That it would be clear when a change is staged.

#### Possible Solution

That there would be a different field, such as `$stageddeleted` for rendering staged deleted differently to unstaged deleted.

#### Environment
- Starship version: `starship 1.22.1`
- Shell type: `zsh`
- Shell version: `zsh 5.9 (arm64-apple-darwin24.0)`
- Shell plugin manager: `oh-my-zsh` `d41ca84af1271e8bfbe26f581cebe3b86521d0db`
- Terminal emulator: `ghostty`
- Operating system: `macOS 15.3.2`

https://github.com/leighmcculloch/dotfiles/blob/835cd284512ccc0f223ebd19cf598e7dc0ac3ed0/files/zshrc

#### Starship Configuration
https://github.com/leighmcculloch/dotfiles/blob/835cd284512ccc0f223ebd19cf598e7dc0ac3ed0/files/config/starship.toml

serl:
#### Description

Treat **unstaged** deletions as modifications.

#### Motivation and Context

Closes #6717.

As stated in the issue, the docs talks about deletion being "staged deletion", so unstaged deletions should not count. Just removing them gives a blind spot, so I counted them as modifications, as `__git_ps1` would do.

NOTE: The edit applies only to the gitoxide implementation. When `use_git_executable = true` is set, the behavior is the "incorrect" one. If that should be tackled as well, let me know.

--

#### Checklist:

### DIFF
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -398,7 +398,7 @@ fn get_repo_status(
                             status: EntryStatus::Change(Change::Removed),
                             ..
                         } => {
-                            repo_status.deleted += 1;
+                            repo_status.modified += 1;
                         }
                         Item::Modification {
                             status:
@@ -1365,7 +1365,7 @@ pub(crate) mod tests {
     }
 
     #[test]
-    fn shows_deleted_file() -> io::Result<()> {
+    fn shows_deleted_file_as_modified() -> io::Result<()> {
         let repo_dir = fixture_repo(FixtureProvider::Git)?;
 
         create_deleted(repo_dir.path())?;
@@ -1373,7 +1373,27 @@ pub(crate) mod tests {
         let actual = ModuleRenderer::new("git_status")
             .path(repo_dir.path())
             .collect();
-        let expected = format_output("✘");
+        let expected = format_output("!");
+
+        assert_eq!(expected, actual);
+        repo_dir.close()
+    }
+
+    #[test]
+    fn shows_deleted_as_modified_with_count() -> io::Result<()> {
+        let repo_dir = fixture_repo(FixtureProvider::Git)?;
+
+        create_deleted(repo_dir.path())?;
+
+        let actual = ModuleRenderer::new("git_status")
+            .config(toml::toml! {
+                [git_status]
+                modified = "!$count"
+                ahead = ""
+            })
+            .path(repo_dir.path())
+            .collect();
+        let expected = format_output("!1");
 
         assert_eq!(expected, actual);
         repo_dir.close()
@@ -1398,7 +1418,7 @@ pub(crate) mod tests {
     fn shows_deleted_file_with_count() -> io::Result<()> {
         let repo_dir = fixture_repo(FixtureProvider::Git)?;
 
-        create_deleted(repo_dir.path())?;
+        create_deleted_in_index(repo_dir.path())?;
 
         let actual = ModuleRenderer::new("git_status")
             .config(toml::toml! {
@@ -1447,7 +1467,7 @@ pub(crate) mod tests {
         let actual = ModuleRenderer::new("git_status")
             .path(repo_dir.path())
             .collect();
-        let expected = format_output("✘?");
+        let expected = format_output("!?");
 
         assert_eq!(expected, actual);
         worktree_dir.close()?;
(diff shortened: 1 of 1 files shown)


serl:
Failing test in CI (only on Windows) seems unrelated (didn't touch that module):

```
thread 'modules::git_metrics::tests::shows_deleted_lines_of_entire_file' panicked at src\modules\git_metrics.rs:545:9:
  left: Some("\u{1b}[1;31m-3\u{1b}[0m ")
 right: None
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library\std\src\panicking.rs:697
   1: core::panicking::panic_fmt
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library\core\src\panicking.rs:75
   2: core::panicking::assert_failed_inner
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library\core\src\panicking.rs:443
   3: core::panicking::assert_failed<enum2$<core::option::Option<alloc::string::String> >,enum2$<core::option::Option<alloc::string::String> > >
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588\library\core\src\panicking.rs:403
   4: starship::modules::git_metrics::tests::shows_deleted_lines_of_entire_file
             at .\src\modules\git_metrics.rs:545
   5: starship::modules::git_metrics::tests::shows_deleted_lines_of_entire_file::closure$0
             at .\src\modules\git_metrics.rs:535
   6: core::ops::function::FnOnce::call_once<starship::modules::git_metrics::tests::shows_deleted_lines_of_entire_file::closure_env$0,tuple$<> >
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588\library\core\src\ops\function.rs:253
   7: core::ops::function::FnOnce::call_once
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library\core\src\ops\function.rs:253
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    modules::git_metrics::tests::shows_deleted_lines_of_entire_file
```

davidkna:
Since this is apparently a long-standing behaviour I would prefer updating the documentation instead, or waiting for an eventual v2.

serl:
> Since this is apparently a long-standing behaviour I would prefer updating the documentation instead, or waiting for an eventual v2.

I get the point, still that doesn't seem as breaking change to me, since both `modified` and `deleted` have a symbol in the default configuration, so the "not-reading-the-doc" users shouldn't be impacted that much. The others should have opened an issue before 😁

If that still didn't convince you, I have other ideas:
- have two "deleted" counts: "unstaged deleted" and "staged deleted" (not my favorite. Probably the most confusing)
- have a flag to choose if "unstaged deleted" means "deleted" or "modified", by default set to "deleted" <- this seems like a good compromise to me. Would be happy to update the PR if we agree on the details

mcwarman:
@serl I believe your request is covered in: https://github.com/starship/starship/pull/5655

serl:
> @serl I believe your request is covered in: #5655

Indeed, if I read correctly, it's like the `two "deleted" counts` but for all possible states, right? Yeah I think it's the way to go.

I'm closing this, hoping to see that merged soon 😊

PR #7084, opened 2025-10-29
fix: show nothing on empty git status

cscnk52:
#### Description

if i have below config:

```toml
[git_status]
format = 'there should not render when git status is empty ([$all_status$ahead_behind]($style))'
```

This is what i actually get:
image

when git status is empty, then whole part should not render.

This PR fix this question.

#### Checklist:

### DIFF
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -818,7 +818,7 @@ pub(crate) mod tests {
         let actual = ModuleRenderer::new("git_status")
             .config(toml::toml! {
                 [git_status]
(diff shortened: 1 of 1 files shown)


pantheraleo-7:
You can just include that in the [conditional format string](https://starship.rs/config/#conditional-format-strings):
```
[git_status]
format = '(there should not render when git status is empty [$all_status$ahead_behind]($style))'
```

cscnk52:
Thanks @pantheraleo-7 , i will close this PR.

PR #7091, opened 2025-10-31
style(config/default): better default symbols for `conflicted` & `deleted` git status options

pantheraleo-7:
#### Description

- changed default of `modules::git_status::GitStatusConfig.deleted`
- changed default of `modules::git_status::GitStatusConfig.conflicted`
 
#### Motivation and Context

Yes, "better" is subjective but I think I can enumerate some objective reasons for this change. So please hear me out:
- the current `deleted` symbol (✘) looks out of place when compared to other git status symbols like `modified`, `staged`, `untracked`, etc. They use simple punctuation marks and have symmetrical, boxy look; while `deleted` looks "emoji-like" and is unsymmetrical.
- the current `conflicted` symbol is a simple equal sign, but a conflict is when the present and incoming changes are **unequal**?

#### Screenshots:
Before/After
Screenshot 2025-10-31 at 11 20 39 PM
Screenshot 2025-11-01 at 8 04 50 AM

#### Checklist:


davidkna:
Thanks for the PR! I am going to decline this as changing default symbols is a breaking change that affects all users. If you want to change the defaults in an upcoming major version please create an issue so any potential candidate symbols can be discussed by the community.

PR #7092, opened 2025-11-01
fix: correctly use platform-specific cache directories; respect `$XDG_CACHE_HOME` on Linux

### LINKED ISSUE #6672: Starship does not respect to XDG_CACHE_HOME environment variable.
#### Current Behavior
While all other programs does, NeoVim and Fish shell included; Starship does not follow XDG_CACHE_HOME environment variable and creates it's cache directory on $HOME/.cache even when it's configured for some other location.

#### Expected Behavior
It should follow XDG_CACHE_HOME environment variable when set and fallback to $HOME/.cache if not set like all other programs does.

#### Possible Solution
Firstly, the program should check for if $XDG_CACHE_HOME environment variable is set or not;
If it's set, then Starship should follow it to decide its location for caches folder;
If it's not set, then Starship should fallback to $HOME/.cache as it's location to put caches.

#### Environment
- Starship version: 1.22.1

pantheraleo-7:
#### Description

The default config cache home `~/.cache` can now be overriden by `XDG_CACHE_HOME`.

#### Motivation and Context

While the current cache file location is standard in some platform, it is not objectively cross-platform. Adding support for `XDG_CACHE_HOME` solves this issue _without_ introducing any complexity of different locations.

Closes #6672

#### Checklist:

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -21,15 +21,21 @@ pub struct StarshipLogger {
 
 /// Returns the path to the log directory.
 pub fn get_log_dir() -> PathBuf {
-    env::var_os("STARSHIP_CACHE")
+    let log_dir = env::var_os("STARSHIP_CACHE")
         .map(PathBuf::from)
-        .unwrap_or_else(|| {
-            utils::home_dir()
-                .map(|home| home.join(".cache"))
-                .or_else(dirs::cache_dir)
-                .unwrap_or_else(std::env::temp_dir)
-                .join("starship")
-        })
+        .or_else(dirs::cache_dir)
+        .unwrap_or_else(std::env::temp_dir)
+        .join("starship");
+
+    if !log_dir.exists()
+        && let Some(home_dir) = utils::home_dir()
+    {
+        let old_log_dir = home_dir.join(".cache/starship");
+        let _ = fs::create_dir_all(&log_dir);
+        let _ = fs::rename(&old_log_dir, &log_dir);
+    }
+
+    log_dir
 }
 
 /// Deletes all log files in the log directory that were modified more than 24 hours ago.
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -1777,24 +1777,20 @@
       }
     },
     "vcs": {
+      "$ref": "#/$defs/VcsConfig",
       "default": {
-        "disabled": false,
-        "fossil_modules": "$fossil_branch$fossil_metrics",
-        "git_modules": "$git_branch$git_commit$git_state$git_metrics$git_status",
-        "hg_modules": "$hg_branch$hg_state",
         "order": [
           "git",
           "hg",
           "pijul",
           "fossil"
         ],
+        "disabled": false,
+        "fossil_modules": "$fossil_branch$fossil_metrics",
+        "git_modules": "$git_branch$git_commit$git_state$git_metrics$git_status",
+        "hg_modules": "$hg_branch$hg_state",
         "pijul_modules": "$pijul_channel"
-      },
-      "allOf": [
-        {
-          "$ref": "#/definitions/VcsConfig"
-        }
-      ]
+      }
     },
     "vcsh": {
       "$ref": "#/$defs/VcshConfig",
@@ -6696,54 +6692,45 @@
       "type": "object",
       "properties": {
         "order": {
-          "description": "Order in which to discover VCSes. The first one found is the one used.",
+          "description": "Order in which to discover VCSes.\nThe first one found is the one used.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
           "default": [
             "git",
             "hg",
             "pijul",
             "fossil"
-          ],
-          "type": "array",
-          "items": {
-            "$ref": "#/definitions/Vcs"
-          }
+          ]
         },
         "disabled": {
           "description": "Disables the VCS module.",
-          "default": false,
-          "type": "boolean"
+          "type": "boolean",
+          "default": false
         },
         "fossil_modules": {
           "description": "Modules to use when Fossil is matched.\n\nThey are configured separately at the top level.",
-          "default": "$fossil_branch$fossil_metrics",
-          "type": "string"
+          "type": "string",
+          "default": "$fossil_branch$fossil_metrics"
         },
         "git_modules": {
           "description": "Modules to use when Git is matched.\n\nThey are configured separately at the top level.",
-          "default": "$git_branch$git_commit$git_state$git_metrics$git_status",
-          "type": "string"
+          "type": "string",
+          "default": "$git_branch$git_commit$git_state$git_metrics$git_status"
         },
         "hg_modules": {
           "description": "Modules to use when Mercurial is matched.\n\nThey are configured separately at the top level.",
-          "default": "$hg_branch$hg_state",
-          "type": "string"
+          "type": "string",
+          "default": "$hg_branch$hg_state"
         },
         "pijul_modules": {
           "description": "Modules to use when Pijul is matched.\n\nThey are configured separately at the top level.",
-          "default": "$pijul_channel",
-          "type": "string"
+          "type": "string",
+          "default": "$pijul_channel"
         }
       },
       "additionalProperties": false
-    },
-    "Vcs": {
-      "type": "string",
-      "enum": [
-        "fossil",
-        "git",
-        "hg",
-        "pijul"
-      ]
     },
     "VcshConfig": {
       "type": "object",
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -46,7 +46,14 @@ os.setenv('STARSHIP_CONFIG', 'C:\\Users\\user\\example\\non\\default\\path\\star
 
 ### Logging
 
-By default starship logs warnings and errors into a file named `~/.cache/starship/session_${STARSHIP_SESSION_KEY}.log`, where the session key is corresponding to an instance of your terminal.
+Starship logs warnings and errors into files named `session_${STARSHIP_SESSION_KEY}.log`, where `STARSHIP_SESSION_KEY` corresponds to an instance of your terminal.
+
+These files are stored in platform-specific cache directories:
+
+- Linux: `XDG_CACHE_HOME` (if set) or `~/.cache/starship` (default)
+- macOS: `~/Library/Caches/starship`
+- Windows: `~/AppData/Local/starship`
+
 This, however can be changed using the `STARSHIP_CACHE` environment variable:
 
 ```sh


src/logger.rs
```diff
@@ -19,13 +19,20 @@ pub struct StarshipLogger {
...
 }
 
+fn cache_home_dir() -> Option<PathBuf> {
+    if let Some(xdg_cache_home) = env::var_os("XDG_CACHE_HOME") {
+        Some(PathBuf::from(xdg_cache_home))
+    } else {
+        Some(utils::home_dir()?.join(".cache"))
+    }
+}
+
```
davidkna:
```suggestion
```

davidkna:
I am somewhat conflicted because this will make discoverability of the cache dir more difficult (making it dependant on the user environment), also I feel like an os-native option as has been provided by `dirs::cache_dir()` may be a better option. Migration may also be necessary to ensure log file cleanup.

pantheraleo-7:
> this will make discoverability of the cache dir more difficult (making it dependant on the user environment)

I would argue that the same goes for `STARSHIP_CACHE`. But the point of these env vars is to _override_ locations. They got nothing to do with what _starship_ is ought to use. The default (`~/.cache`) and the fallbacks (`dirs::cache_dir` & `std::env::temp_dir`) remain the same.

> I feel like an os-native option as has been provided by dirs::cache_dir() may be a better option.

First of all, `dirs::cache_dir` is a (rare) fallback and not something we intend to use. 

Since you mentioned it, I took a closer look, and it turns out we aren't even using `dirs::cache_dir` correctly. `dirs::cache_dir` supports `XDG_CACHE_HOME` on Linux, but we override this by our default (`~/.cache`). On macOS (and possibly Windows), we are almost never going to use platform-specific directories, since `~/.cache` directory will exist if the user has at least one other terminal program as almost all terminal programs use `~` or XDG spec directories (we are an outlier in this). Basically, our choices are _not_ deterministic, it depends on user state.

Also, as I have mentioned before, these platform-specific directories are not user-facing and are meant for [GUI] **apps**. We are a user-facing terminal program. But log files are not user-facing. So in this case, we're okay.

> Migration may also be necessary to ensure log file cleanup.

Will do.

---

BTW, I just noticed. Why are we storing **logs** files in cache directory??? This was only required on Linux until [XDG spec v0.8](https://specifications.freedesktop.org/basedir/0.8/#:~:text=actions%20history%20) standardised it. macOS (and probably Windows?) has platform-specific log directories from the start.

pantheraleo-7:
We should decide between the following options:
1. since we force all platforms to use `~/.config` for the `config.toml`, we might as well force all to use `~/.cache` for logs (here we will also need to support `XDG_CACHE_HOME`)
2. we recognise that logs files are not user-facing (unlike `config.toml`), and hence use `dirs::cache_dir()` _only_
3. we use platform-specific **log** directories (see [directories-rs#81](https://codeberg.org/dirs/directories-rs/issues/81))

Choosing any will bring about consistency, and most importantly, we can be sure that on a given platform the log files will be located in one and only one possible location (given that the directory exists of course, else we fall back to `std::env::temp_dir`). The current changes this PR makes is a mix of option 1 & 2.

pantheraleo-7:
@davidkna please review the latest commit - which implements option 1 and migration

davidkna:
1. I fail see how starship specific env overrides for the cache dir (or config location) are an issue.
2. XDG on all platforms feels less applicable to cache dirs.
3. Log files are only saved to avoid printing warnings multiple times per session, so the cache dir feels somewhat appropriate and `dirs` doesn't even have a convenience wrapper for log dirs.

pantheraleo-7:
> I fail see how starship specific env overrides for the cache dir (or config location) are an issue.

No, no. You are misunderstanding. Nobody has problem with starship specific environment variables, not even with the default cache location (but just with the config file location because its placement is odd and wrong).

I only removed `STARSHIP_CACHE` because once `XDG_CACHE_HOME` is supported, it is redundant. Because this env var overrides the cache **directory** - same as `XDG_CACHE_HOME` (unlike `STARSHIP_CONFIG` which overrides the file location). But if you want, I can reinstate it. On second thought, it may even be necessary for users who don't use XDG and at the same time don't want the default location `~/.cache` either (are there such users though? lol).

> XDG on all platforms feels less applicable to cache dirs.

Either we (fully) respect the XDG spec or we don't.

> Log files are only saved to avoid printing warnings multiple times per session, so the cache dir feels somewhat appropriate and dirs doesn't even have a convenience wrapper for log dirs.

So option 2?

(Before you say we already support platform-specific locations, I'd like to remind that the current implementation is wrong. In a bid to do everything, we are all over the place. We are neither properly using platform-specific directories (especially on macOS), nor properly supporting the `~/.cache`/`XDG_CACHE_HOME` convention.)

pantheraleo-7:
As per @davidkna 's suggestion, we are sticking to platform-specific directories for the log files.

This PR (now) fixes the current implementation instead of changing the current (intended) behaviour. Most notably:
- Linux: we now respect `XDG_CACHE_HOME` (via `dirs::cache_dir()`)
- macOS: we now (correctly) use `~/Library/caches/starship` as the log directory.

src/logger.rs
```diff
@@ -19,17 +19,37 @@ pub struct StarshipLogger {
     log_level: Level,
 }
 
+fn cleanup_old_log_dir(log_dir: &Path) -> io::Result<()> {
```
davidkna: This should be tested.

src/logger.rs
```diff
@@ -19,17 +19,37 @@ pub struct StarshipLogger {
...
+    {
+        let _ = fs::File::create(log_dir.join(".migrated"))?;
+        #[cfg(target_os = "linux")]
+        {
+            if let Some(xdg_cache_home) = env::var_os("XDG_CACHE_HOME").map(PathBuf::from)
+                && xdg_cache_home == home_dir.join(".cache")
+            {
+                return Ok(());
+            }
+        }
```
davidkna: Linux should not need special handling here.

src/logger.rs
```diff
@@ -19,17 +19,37 @@ pub struct StarshipLogger {
...
+        fs::remove_dir_all(home_dir.join(".cache/starship"))?;
```
davidkna: This could lead to issue with nushell set-ups based on previous instructions that put the init file there. I suppose having the log file cleaner looking at both paths could work.

src/logger.rs
```diff
@@ -19,17 +19,37 @@ pub struct StarshipLogger {
...
-    env::var_os("STARSHIP_CACHE")
+    let log_dir = env::var_os("STARSHIP_CACHE")
         .map(PathBuf::from)
-        .unwrap_or_else(|| {
-            utils::home_dir()
-                .map(|home| home.join(".cache"))
-                .or_else(dirs::cache_dir)
-                .unwrap_or_else(std::env::temp_dir)
-                .join("starship")
-        })
```
davidkna, line 32: Might as well keep this in the fallback list?

src/logger.rs
```diff
@@ -250,6 +270,20 @@ mod test {
...
 
+    #[test]
+    fn test_log_file_path() -> () {
+        assert_eq!(get_log_dir(), utils::cache_dir().unwrap().join("starship"));
+
+        #[cfg(target_os = "linux")]
+        {
+            let xdg_cache_home = std::env::temp_dir();
+            unsafe {
+                env::set_var("XDG_CACHE_HOME", xdg_cache_home.to_str().unwrap());
```
davidkna: Do not set env vars in tests.

src/utils.rs
```diff
@@ -739,6 +739,10 @@ pub fn home_dir() -> Option<PathBuf> {
     dirs::home_dir()
 }
 
+pub fn cache_dir() -> Option<PathBuf> {
+    dirs::cache_dir()
+}
+
```
davidkna: This does not seem necessary, the `home_dir` helper mostly exists due to historical reasons.
```suggestion
```

davidkna (changes requested):

src/logger.rs
pantheraleo-7: Ok, but what else is there to test even haha

src/logger.rs
pantheraleo-7: Seems unnecessary. Because `~/Library/Caches` and `~/AppData/Local` are practically guaranteed to exist.

src/logger.rs
pantheraleo-7: I have no experience with nushell, could you elaborate more? Are you saying that there might be a ~init~ `starship.nu` file in `~/.cache/starship`?

src/logger.rs
davidkna: Some other tests also use env file mocking that does not rely on `Context`

src/logger.rs
davidkna: Nushell does not have `eval`, so past instructions lead to a first nu init-file (env.nu?) creating a `starship.nu` in the cache directory that was sourced in the second init-file (init.nu?). Sourcing and creating the same file isn't possible with nushell in the same invocation.

src/logger.rs
pantheraleo-7: The old and new log dir is the same on Linux (if `XDG_CACHE_HOME` is not set to a different location) and since we rely on `new_log_dir/.migrated`, we can't unconditionally remove the old log dir.

Though I'm trying a different approach, so we don't have to do this. Let's see...

src/logger.rs
```diff
@@ -21,15 +21,28 @@ pub struct StarshipLogger {
...
+        let _ = fs::File::create(log_dir.join(".migrated"));
+        let old_log_dir = home_dir.join(".cache/starship");
+        cleanup_log_files(&old_log_dir);
+        if let Err(err) = fs::remove_dir(&old_log_dir) {
+            eprintln!(
+                "Failed to cleanup old log dir {}: {}!",
+                old_log_dir.display(),
+                err
+            );
+        }
```
davidkna: Let's avoid printing an error if it's non-empty.
```suggestion
        let _ = fs::remove_dir(&old_log_dir);
```

src/logger.rs
```diff
@@ -21,15 +21,28 @@ pub struct StarshipLogger {
...
+        .or_else(dirs::cache_dir)
+        .unwrap_or_else(std::env::temp_dir)
+        .join("starship");
+
+    if !log_dir.join(".migrated").exists()
+        && let Some(home_dir) = utils::home_dir()
+    {
+        let _ = fs::File::create(log_dir.join(".migrated"));
+        let old_log_dir = home_dir.join(".cache/starship");
+        cleanup_log_files(&old_log_dir);
```
davidkna: I would prefer handling this together with the other call to `clean_log_files`.

src/logger.rs
```diff
@@ -21,15 +21,28 @@ pub struct StarshipLogger {
...
-        })
+        .or_else(dirs::cache_dir)
+        .unwrap_or_else(std::env::temp_dir)
+        .join("starship");
+
+    if !log_dir.join(".migrated").exists()
+        && let Some(home_dir) = utils::home_dir()
+    {
+        let _ = fs::File::create(log_dir.join(".migrated"));
+        let old_log_dir = home_dir.join(".cache/starship");
```
davidkna, line 33: It looks like this may actually delete the new location on Linux?

davidkna (commented):
I suppose you could do something like this:
```
cleanup_log_files(&new_log_dir);
if new_log_dir != old_log_dir {
    cleanup_log_files(&old_log_dir);
    let _ = fs::remove_dir(&old_log_dir);
}
```

davidkna:
@pantheraleo-7 Are you still working at this?

src/logger.rs
pantheraleo-7: I did this for 2 reasons:
- logs newer than 24 hours won't be cleaned up, so alerting users to do it manually seems like a good idea
- IIRC [some] nushell users may have non-log files?

src/logger.rs
pantheraleo-7: does it matter? wouldn't it get recreated?

src/logger.rs
pantheraleo-7: what do you mean? there's only one call to `cleanup_log_files()`...

src/logger.rs
davidkna: `fs::remove_dir` is not recursive.

src/logger.rs
davidkna: Performance is the first thing that comes to mind.

src/logger.rs
davidkna: That's a problem in itself.

src/logger.rs
pantheraleo-7: yes, exactly. The intended behaviour is to fail if there are non-log files (one known issue is that log files newer than 24 hours will cause failure too).

src/logger.rs
pantheraleo-7: This isn't a helpful response, is it?

src/logger.rs
pantheraleo-7: well I was initially special-casing Linux for precisely this reason, but you said no to.

davidkna:
I'm done with reviewing this.

pantheraleo-7:
This PR is ready.

kuchta:
I think that XDG_\<ROLE\>_HOME should be supported on all platforms, since just by setting those env variables the user is signaling he wants the programs to follow that spec.

pantheraleo-7:
the maintainer got offended and has refused to review this, after I pointed out that he didn't answer my clarifying question to his review comment. I don't think it is possible to construe that as being rude, and most importantly, I didn't mean it to be.

PR #7100, opened 2025-11-05
fix(aws): support `$duration` for `sso_session`

### LINKED ISSUE #7099: Fix `$duration` for AWS `sso_session`
### Is your feature request related to a problem? Please describe.

PR #6614 added support for displaying `$duration` for AWS SSO sessions, but only for the legacy format in which `sso_session` is not configured.

As seen in the related test, the AWS module is expected to _not_ display a duration when `sso_session` is present.

https://github.com/starship/starship/blob/c0206cd0f53841190ec1f3c177272135f2284557/src/modules/aws.rs#L1110-L1113

If not providing a session name, `aws configure sso` raises a warning.

```
WARNING: Configuring using legacy format (e.g. without an SSO session).
Consider re-running "configure sso" command and providing a session name.
```

### Describe the solution you'd like

It would be helpful to support SSO sessions because this is the format recommended by AWS.

### Describe alternatives you've considered

Related:

- https://github.com/starship/starship/issues/4697
- https://github.com/starship/starship/pull/5170
- https://github.com/starship/starship/pull/6614

br3ndonland:
## Motivation and Context

Closes #7099

PR #6614 added support for displaying `$duration` for AWS SSO sessions, but only for the legacy format in which `sso_session` is not configured.

If not providing a session name, `aws configure sso` raises a warning.

```
WARNING: Configuring using legacy format (e.g. without an SSO session).
Consider re-running "configure sso" command and providing a session name.
```

It would be helpful to support SSO sessions because this is the format recommended by AWS.

## Description

This PR will add support for `$duration` when `sso_session` is present in the configuration for the selected AWS profile.

The implementation is based on the linked code from `botocore`.

https://github.com/starship/starship/blob/c0206cd0f53841190ec1f3c177272135f2284557/src/modules/aws.rs#L158-L159

The reason this doesn't work currently is that, if `sso_session` is present, the AWS CLI uses the _session name_ to compute the cache key, not the start URL. This PR will fix the implementation and update tests accordingly.

## How Has This Been Tested?

- [x] I have tested using **MacOS** (unit tests)
- [x] I have tested using **Linux** (unit tests)

## Checklist:

### DIFF
--- a/src/modules/aws.rs
+++ b/src/modules/aws.rs
@@ -124,6 +124,15 @@ fn get_aws_profile_and_region(
     }
 }
 
+// Get the SSO cache key input from a profile section, handling both direct sso_start_url
+// and sso_session (which references a separate [sso-session <name>] section).
+fn get_sso_cache_key_input(profile_section: &ini::Properties) -> Option<String> {
+    profile_section
+        .get("sso_session")
+        .map(|s| s.to_string())
+        .or_else(|| profile_section.get("sso_start_url").map(|s| s.to_string()))
+}
+
 fn get_credentials_duration(
     context: &Context,
     aws_profile: Option<&Profile>,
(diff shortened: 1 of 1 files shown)


davidkna:
Thanks for the contribution @br3ndonland!

PR #7112, opened 2025-11-10
feat(python): add option to replace generic venv-names with parent dir

### LINKED ISSUE #5889: Allow option to display the $virtualenv differently, with more of $VIRTUAL_ENV as context
The $virtualenv tag in python projects pretty much always says (.venv) since the .venv dir is always called that.  The problem is that switching directories and things you can get into a situation where you are using the wrong venv.  And various IDEs automatically set venv (Im looking at you vscode) so the shells that get opened may not be what you expect.

``` bash
~/g/website-support-scripts feat/WEB-8765-migrate-script… 2?  is 📦 v0.0.2 [ v16.19.0][🐍 v3.12.2(.venv)]
10:23:18 ❯ echo $VIRTUAL_ENV
/Users/marvingreenberg/git/tws-stack-deployment/.venv
```
Here, my shell is using the venv from some other project...  Not what is wanted. 

A setting to modify how $virtualenv behaves, instead of just displaying .venv (or whatever the .venv dir is?)

* Add option virtualenv_qualified_dir : true that would display, e.g., `(tws-stack-deployment/.venv)`
* Truncate to x character, maybe configurable, or have qualified_dir_len: 9 (default 0) e.g., `(tws-…ment/.venv)`
* Only display qualification when the virtual env seems ambiguous so just say (.venv)
   when VIRTUAL_ENV and current dir (or subdir) seems consistent with the PWD
   
Or some combination of one or more of those.  The last one is not immediately obvious exactly how to check
if the VIRTUAL_ENV is "consistent".  With the first and third implemented, implementing the second is much less important and maybe even undesirable?  Obviously more thought to the option name(s) to be consistent with existing styles/practice

I know (I think) I can write a custom tag (sic) to format the appropriate info from the VIRTUAL_ENV env var (although I tried a few iteraions unsuccessfully, so far).  But this _seems_ to be a reasonable useful feature to have builtin.

RFCreate:
#### Description
This PR adds a new list of venv names as a customizable option, where if the resolve virtual venv path matches one of the names in the list, it will be overwritten with the parent directory name.

#### Motivation and Context
Every time I create a virtualenv, I always name it .venv or venv which when activated does not give me information of the project I am working on since it always displays the same.

Closes #5889

venv

Create a virtual environment with a standard name like .venv or venv, activate it, and the parent directory will be displayed instead of the name of the virtualenv folder.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3982,13 +3982,13 @@ By default, the module will be shown if any of the following conditions are met:
 
 ### Variables
 
-| Variable     | Example         | Description                                |
-| ------------ | --------------- | ------------------------------------------ |
-| version      | `'v3.8.1'`      | The version of `python`                    |
-| symbol       | `'🐍 '`         | Mirrors the value of option `symbol`       |
-| style        | `'yellow bold'` | Mirrors the value of option `style`        |
-| pyenv_prefix | `'pyenv '`      | Mirrors the value of option `pyenv_prefix` |
-| virtualenv   | `'venv'`        | The current `virtualenv` name              |
+| Variable     | Example         | Description                                                                 |
+| ------------ | --------------- | --------------------------------------------------------------------------- |
+| version      | `'v3.8.1'`      | The version of `python`                                                     |
+| symbol       | `'🐍 '`         | Mirrors the value of option `symbol`                                        |
+| style        | `'yellow bold'` | Mirrors the value of option `style`                                         |
+| pyenv_prefix | `'pyenv '`      | Mirrors the value of option `pyenv_prefix`                                  |
+| virtualenv   | `'venv'`        | The current `virtualenv` name or the parent if matches `generic_venv_names` |
 
 ### Example
 


RFCreate:
If you got time, this is a rework of #6526 @jankatins @andytom

jankatins:
I can't say much regarding the code, but I'm a bit surprised that the code gets only applied when the `prompt` is not given. I would have expected the generic venv names would overwrite that as well  `python -m venv .venv --prompt "whatever"` -> if whatever is in the generic venv name list, it gets replaced. reason: sometimes I have more control over the starship config than over how my venv is getting constructed.

E.g. I would expect the following outcome (with prompt from `pyvenv.cfg`):

| prompt | folder | generic_venv_names | displayed name |
| --- | --- | --- | --- |
| `<unset>` / set to the same as the .venv folder name  | `.venv` | `.venv`  | `<project-folder>` |
| `abc`  | `.venv` | `.venv`  | `abc` |
| `abc`  | `.venv` | `.venv`, `abc`  | `<project-folder>` |

In any case, even if the latter stays this way: a testcase to document that design decision which value takes precedence would be nice.

RFCreate:
Great suggestion, i have implemented it so that the `generic_venv_names` list takes precedence over the prompt value in `pyvenv.cfg`.

jankatins:
To me the logic looks good (judging by the tests). I would have expected a sentence in the docs which states the precedence, but as that wasn't in the docs up to now, I consider this only a "would be nice" thingy.

I can make no clear recommendation regarding the rust code itself.

RFCreate:
It is now written into the documentation!

davidkna (commented):
LGTM, but please change the default config value to an empty list, because this config option is likely to lead to show the same string as the directory module.

RFCreate:
Changes are ready!

davidkna (approved):

davidkna:
Thanks for the contribution @RFCreate!

PR #7082, opened 2025-10-28
fix(fish): fix job counting

davidkna:
#### Description

This PR moves `__count` from local/block-scope to the proper function-scope, to make it available after the `if`-block.

#### Motivation and Context

Closes #7081

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #7124, opened 2025-11-17
fix(zsh): fix job counting

TomJo2000:
#### Description

The current `${#jobstates}` expansion incorrectly uses the length of the first array entry for the job count in Zsh.
This can be fixed by explicitly operating on the whole array using `${jobstates[*]]`.
#### Motivation and Context

- This acts as a followup to #7082

While looking through the changelog for 1.24.1 I saw the fix for the jobcount in the Fish shell.
I have previously noticed issues with the jobcount in Zsh, though I had previously assumed this was an issue with my Zsh config.
After looking into it I can however confirm this is just a misbehaving length expansion on the `jobstates` array.

*Technically* this is caused by the `KSH_ARRAYS` shell option, which makes Zsh arrays behave identically to Ksh/Bash arrays[^1].
[^1]: https://man.archlinux.org/man/zshparam.1#Array_Subscripts:~:text=When%20an%20array%20parameter%20is%20referenced%20as%20%60%24name%27%20(with%20no%20subscript)%20it%20evaluates%20to%20%60%24name%5B*%5D%27%2C%20unless%20the%20KSH_ARRAYS%20option%20is%20set

But I decided to make a PR for it anyway since the presence of that shell option should not impact the results of the job count even if it is not set by default.

image

  - [x] I have tested using **Android/Termux**

#### Checklist:

- [x] Not applicable.

### DIFF
--- a/src/init/starship.zsh
+++ b/src/init/starship.zsh
@@ -50,7 +50,7 @@ prompt_starship_precmd() {
 
     # Use length of jobstates array as number of jobs. Expansion fails inside
     # quotes so we set it here and then use the value later on.
-    STARSHIP_JOBS_COUNT=${#jobstates}
+    STARSHIP_JOBS_COUNT="${#jobstates[*]}"
 }
 
 # Runs after the user submits the command line, but before it is executed and


davidkna:
Thanks for the contribution @TomJo2000!

PR #7131, opened 2025-11-19
fix(mise): align documentation and implementation

nagromc:
#### Description

* Fixed the documentation to include the missing `detect_*` properties as implemented
* Fixed the implementation to include the `on ` prefix as described in the documentation

`on ` is maybe not the best prefix. Should we change it to `via ` like the other ones?

#### Motivation and Context

Consistency.

```console
$ cd target/debug 
$ ./starship prompt

%{%}…/starship/target/debug%{%} on %{%} mise-docs-impl%{%} 
%{%}❯%{%} %                                                                                                                                                                                 
$ touch mise.toml  
$ ./starship prompt

%{%}…/starship/target/debug%{%} on %{%} mise-docs-impl%{%} on %{%}mise healthy%{%}
%{%}❯%{%} %                                                                                                                                                                                  
```

#### Checklist:

### DIFF
--- a/src/modules/mise.rs
+++ b/src/modules/mise.rs
@@ -88,7 +88,10 @@ mod tests {
                 }),
             );
 
-        let expected = Some(format!("on {} ", Color::Purple.bold().paint("mise healthy")));
+        let expected = Some(format!(
+            "on {} ",
+            Color::Purple.bold().paint("mise healthy")
+        ));
         assert_eq!(expected, renderer.collect());
 
         dir.close()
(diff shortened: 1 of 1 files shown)


davidkna (approved):

nagromc:
> `on ` is maybe not the best prefix. Should we change it to `via ` like the other ones?

What do you think?

davidkna:
Some of the modules related to environments use `on`, so using it makes sense to me here.

PR #6866, opened 2025-07-25
feat(install.sh): handle installation directory prompt responses case-insensitively

Daniel-Escamilla:
#### Description

Improve input handling in `install/install.sh` for confirmation prompts. Specifically:

- Interprets user responses case-insensitively to correctly handle both affirmative and negative answers.
- Adds clearer error messages when the user aborts input with `Ctrl+C` or `Ctrl+D`.
- Ensures that if the user declines installation in the default folder, the prompt exits gracefully with appropriate messaging.

#### Motivation and Context

This change ensures a more robust and user-friendly install experience, particularly when users provide lowercase or uppercase variants of "no" (`n`, `N`, `no`, `NO`) or interrupt input unexpectedly. Previously, such responses could lead to unclear errors or unintended behaviour.

Closes #5608 

#### How Has This Been Tested?

Manually tested the following cases in Starship:

- User inputs `y`, `Y`, `yes`, `YES` → script proceeds as expected.
- User inputs `n`, `N`, `no`, `NO` → script prompts again for an alternative installation path.
- User presses `Ctrl+C` → script exits cleanly.
- User presses `Ctrl+D` → script exits with error message.

#### Checklist:

- [x] I have tested using **Linux**
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #3691, opened 2022-03-06
fix: use shell-compatible curl install

matchai:
#### Description

Update curl installation method to not use command expansion.

#### Motivation and Context

Closes #3664

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #7138, opened 2025-11-23
Revert "feat(install.sh): handle installation directory prompt respon…

### LINKED ISSUE #7133: Interactive install fails when running "curl | sh"; documentation should update recommended command.
### Current Behavior

When following the installation command shown in the documentation:

``` sh
curl -sS https://starship.rs/install.sh | sh
```

the interactive installer fails when the user chooses a custom installation path.
Because the script is executed through a pipe, stdin is no longer a TTY.
As a result, the interactive read prompts in install.sh cannot read user input and the installer exits with:

```
"Error reading from prompt (please re-run with '--yes' or provide path)"
```

> Note: This happens because the script expects a real TTY for interactive prompts, but piping via curl | sh removes the TTY.

Image

> Linux terminal

Image

> Apple terminal

### Expected Behavior

The recommended installation command should allow the installer to read user input normally and complete the interactive process without errors.

Image

### Possible Solution

Update the documentation to recommend a TTY-preserving installation command such as:

``` sh
sh <(curl -sS https://starship.rs/install.sh)
```

### Environment

Shell type:
- zsh

Shell version:
- zsh 5.9 (Linux and macOS)

Bash version:
- GNU bash 5.2.21 (Linux)

Terminal emulator:
- Linux terminal / Apple Terminal

Operating system:
- Linux (x86_64-ubuntu-linux-gnu)
- macOS (arm64-apple-darwin25.0)

davidkna:
…ses case-insensitively (#6866)"

This reverts commit bc48b5fcb07b3ea806127bc6ab13616028e92642.

#### Description

With the changes made in #6866, the suggested installation instructions no longer work as expected (#7133).

#### Motivation and Context

Closes #7133

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -257,18 +257,12 @@ detect_target() {
   printf '%s' "${target}"
 }
 
-set_interrupt_handler() {
-  trap 'echo; error "Interrupted (please re-run with the '--yes' option)"; exit 130' INT
-}
 
 confirm() {
   if [ -z "${FORCE-}" ]; then
-    set_interrupt_handler
     printf "%s " "${MAGENTA}?${NO_COLOR} $* ${BOLD}[y/N]${NO_COLOR}"
     set +e
     read -r yn </dev/tty
-    trap - INT
-    yn=$(echo "$yn" | tr '[:upper:]' '[:lower:]')
     rc=$?
     set -e
     if [ $rc -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


cappyzawa (approved):
This makes sense — #7133 breaks the recommended `curl | sh` install path, so reverting is the safest option.

Also, I should have tested the pipe / non-TTY execution path during the original review. Thanks for catching this and following through.

Approved 👍

davidkna:
Thanks for the review @cappyzawa!

Daniel-Escamilla:
Hi @davidkna @cappyzawa 

I noticed that my previous changes improving input handling in `install/install.sh` have been reverted. Could you please clarify the reason behind this decision?

The update I proposed made the confirmation logic more robust and user-friendly. In addition to handling case-insensitive inputs and exiting gracefully on `Ctrl+C` and `Ctrl+D`, it also allowed users who answered "no" to the default installation path to specify a custom directory instead of aborting immediately.

If the reversion was due to the `curl | sh` installation path losing its TTY (as mentioned in #7133), wouldn’t it be preferable to update the documentation instead? For example, recommending a TTY-preserving installation command like:

sh <(curl -sS https://starship.rs/install.sh)

This approach would keep the improved input handling while avoiding the TTY issue for users following the documented installation path.

Thanks.

davidkna:
The reason `curl | sh` was eventually settled is that other approaches lead to incompatibilities with the fish shell (#3691 #3664). New install approaches would need to be evaluated carefully, and this PR has provided a simple fix that did not require more elaborate testing.

PR #7143, opened 2025-11-25
feat: implement referencing next segment's style in conf file

christophehenry:
#### Description
Solves [#6558-2888265896](https://github.com/starship/starship/issues/6558#issuecomment-2888265896). Add ability to reference next modules' foreground and background. Useful to make a spacer seperating 2 modules, merging the style of both.

#### Motivation and Context
Starship is missing the ability to create a spacer sharing colors from surrounding modules. Some presets like [Gruvbox](https://starship.rs/presets/gruvbox-rainbow) partly implement this design. But since color needs to be hard-coded, it gets broken as soon as one module is conditionnaly rendered. This PR allows to define colors that will always match the following module no matter what it is.

#### Checklist:

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -356,7 +356,8 @@ Style strings are a list of words, separated by whitespace. The words are not ca
 - `none`
 
 where `<color>` is a color specifier (discussed below). `fg:<color>` and `<color>` currently do the same thing, though this may change in the future.
-`<color>` can also be set to `prev_fg` or `prev_bg` which evaluates to the previous item's foreground or background color respectively if available or `none` otherwise.
+`<color>` can also be set to `prev_fg`, `prev_bg`, `next_fg` or `next_fg`, which evaluates to, respectively, the previous and next item's foreground or background color if available or `none` otherwise.
+Please also note that if the previous or next module also references a color with, for instance, `bg:next_fg`, the reference won't be followed and the default style will apply.
 `inverted` swaps the background and foreground colors. The order of words in the string does not matter.
 
 The `none` token overrides all other tokens in a string if it is not part of a `bg:` specifier, so that e.g. `fg:red none fg:blue` will still create a string with no styling. `bg:none` sets the background to the default color so `fg:red bg:none` is equivalent to `red` or `fg:red` and `bg:green fg:red bg:none` is also equivalent to `fg:red` or `red`. It may become an error to use `none` in conjunction with other tokens in the future.
--- a/src/module.rs
+++ b/src/module.rs
@@ -1,6 +1,6 @@
 use crate::segment;
 use crate::segment::{FillSegment, Segment};
-use nu_ansi_term::{AnsiString, AnsiStrings, Style as AnsiStyle};
+use nu_ansi_term::{AnsiString, AnsiStrings};
 use std::fmt;
 use std::time::Duration;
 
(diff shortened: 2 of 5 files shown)


christophehenry:
Any chance for a feedback on this one?

src/module.rs
```diff
@@ -202,27 +202,42 @@ where
...
         match segment {
             Segment::Fill(fs) => {
                 chunks.push((current, fs));
                 current = Vec::new();
-                prev_style = None;
             }
             _ => {
                 used += segment.width_graphemes();
-                let current_segment_string = segment.ansi_string(prev_style.as_ref());
+                let prev = if idx <= 0 {
```
davidkna, line 222: Please make sure CI passes and fix any clippy warnings, and prefer using safe integer manipulation in cases like this.

src/module.rs
```diff
@@ -202,27 +202,42 @@ where
...
+    let mut collected = Vec::<&'a Segment>::new();
     for segment in segments {
+        collected.push(segment);
+        if matches!(segment, Segment::LineTerm) {
+            break;
+        }
+    }
+
+    for (idx, &segment) in collected.iter().enumerate() {
         match segment {
```
davidkna, line 215: I would prefer not to change this to indexing, please use `peek()` to reference upcoming segments. AFAICT the current approach will **break** transitive backward colour references. Please also ensure that:
- transitive backward colour references are tested.
- corner cases involving this are tested.

docs/advanced-config/README.md
```diff
@@ -356,7 +356,8 @@ Style strings are a list of words, separated by whitespace. The words are not ca
 - `none`
 
 where `<color>` is a color specifier (discussed below). `fg:<color>` and `<color>` currently do the same thing, though this may change in the future.
-`<color>` can also be set to `prev_fg` or `prev_bg` which evaluates to the previous item's foreground or background color respectively if available or `none` otherwise.
+`<color>` can also be set to `prev_fg`, `prev_bg`, `next_fg` or `next_fg`, which evaluates to, respectively, the previous and next item's foreground or background color if available or `none` otherwise.
```
davidkna, line 359:
```suggestion
`<color>` can also be set to `prev_fg`, `prev_bg`, `next_fg` or `next_bg`, which evaluates to, respectively, the previous and next item's foreground or background color if available or `none` otherwise.
```

docs/advanced-config/README.md
```diff
@@ -356,7 +356,8 @@ Style strings are a list of words, separated by whitespace. The words are not ca
...
+Please also note that if the previous or next module also references a color with, for instance, `bg:next_fg`, the reference won't be followed and the default style will apply.
```
davidkna, line 360: This was not true for backward colour references.

src/segment.rs
```diff
@@ -9,17 +9,17 @@ use unicode_segmentation::UnicodeSegmentation;
...
 
     /// The string value of the current segment.
-    value: String,
+    pub(crate) value: String,
 }
 
 impl TextSegment {
     // Returns the AnsiString of the segment value
-    fn ansi_string(&self, prev: Option<&AnsiStyle>) -> AnsiString<'_> {
+    fn ansi_string(&self, prev: Option<&AnsiStyle>, next: Option<&AnsiStyle>) -> AnsiString<'_> {
```
davidkna, line 20: Suggestion: Switch to an optional struct that holds both next and prev.

docs/advanced-config/README.md
christophehenry: Hmm… True. Now, reflecting on this I see that `prev_bg` was transitive because we go through the sequences in order so any reference to the previous style is known because we keep a reference to the previous style. Were I to do the same for forward colour references, I would have to first go through the sequences in reverse order, resolve the forward references and then process the sequences like it is currently done. But TBH, I don't really know how to to this in Rust. should I just generate a new `segments` where I would `segment.copy()` with partially resolved styles?

davidkna: You can store the segment iterator in a variable and use `peek()` (may require `peekable()`) to reference upcoming iterator items.

christophehenry: Ah I see that I can use `Peekable::nth` to look further that the next element so I can do transitive. However it feels suboptimal. Worst case scenario is all the segments are references to the very last then I traverse the vector *n×(n+1)/2*. It should be fine since I expect the segments vector size to be less than 10 and the number of references to be no more than 1 or 2 segments away. But still. It bothers me a bit.

davidkna:
Closing as stale.

christophehenry:
No, I intend to finish this one. Please leave it opened.

PR #7145, opened 2025-11-26
feat(directory): add support for regexes in substitutions 

### LINKED ISSUE #5572: `directory.substitutions` expect substitutions to be in order, when TOML spec does not guarantee order
Starship currently expects directory substitutions to be laid out in order of when they're processed, like so:
```toml
[directory.substitutions]
"/a/b" = "b"
"/a" = "a"
```

But relying on that order is actually not well-defined in [TOML, whose tables are *specifically* unordered](https://toml.io/en/v1.0.0#table):
> Under that, and until the next header or EOF, are the key/values of that table. **Key/value pairs within tables are not guaranteed to be in any specific order.** 

This causes other tools, like Nix, to generate incorrect configs despite being following the spec. As shown by nix-community/home-manager#2519, since Nix always sorts tables (or "attrsets") in alphabetical order, this ruins the order of substitutions — which is again, allowed by spec.

Make `directory.substitutions` an array of tables, who do have a defined order. TOML even has a neat syntax for this:
```toml
[[directory.substitutions]]
from = "/a/b"
to = "b"

[[directory.substitutions]]
from = "/a"
to = "a"
```
Sure, it's a bit more typing, but I really doubt people would have dozens of substitutions at a time in their configs anyway.

This issue could also be fixed on Nix's/Home Manager's side, but honestly I think given this is nonstandard behavior, any other tool can run into this despite being compliant with how TOML is specified.

christophehenry:
**Edit** : this PR has involved into supporting regexes in substitutions. I'm keeping the original description for history. 

The directory module allows to substitute `/` for another character. But then the `/` representing the filesystem root is also stripped away with no replacement, even though it doesn't represent a directory seperator in this context. This commit adds an exception to the substitution of root in the path to avoid this behavior.

#### Description

If the substitued string is `"/"`, the initial slash is replaced by `root/` before substitution and then put back in place after so that the root character is not stripped away.

#### Motivation and Context
Closes #5572

If you decide to substitute the directory separator character for something more visual, Starship end up display a path without the initial `/` which is disturbing. I consider this a UX bug but I get this would be controversial. In the future, this could be replaced by something more integrated like a `separator_char` option.

Windows is probably unaffected by this so this PR concentrates on Unixes.

Before:

After:

#### Checklist:
- [ ] I have updated the documentation accordingly. (Does it need documentation?)

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -367,7 +367,7 @@
       "default": {
         "truncation_length": 3,
         "truncate_to_repo": true,
-        "substitutions": {},
+        "substitutions": [],
         "fish_style_pwd_dir_length": 0,
         "use_logical_path": true,
         "format": "[$path]($style)[$read_only]($read_only_style) ",
@@ -2616,32 +2616,7 @@
           "default": true
         },
         "substitutions": {
-          "anyOf": [
-            {
-              "type": "array",
-              "items": {
-                "type": "object",
-                "properties": {
-                  "from": {
-                    "type": "string"
-                  },
-                  "to": {
-                    "type": "string"
-                  },
-                  "regex": {
-                    "type": "string"
-                  }
-                },
-                "required": [
-                  "from",
-                  "to"
-                ]
-              }
-            },
-            {
-              "type": "object"
-            }
-          ],
+          "$ref": "#/$defs/Either",
           "default": []
         },
         "fish_style_pwd_dir_length": {
@@ -2705,6 +2680,42 @@
         }
       },
       "additionalProperties": false
+    },
+    "Either": {
+      "anyOf": [
+        {
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/SubstitutionConfig"
+          }
+        },
+        {
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
+        }
+      ]
+    },
+    "SubstitutionConfig": {
+      "type": "object",
+      "properties": {
+        "from": {
+          "type": "string"
+        },
+        "to": {
+          "type": "string"
+        },
+        "regex": {
+          "type": "boolean",
+          "default": false
+        }
+      },
+      "additionalProperties": false,
+      "required": [
+        "from",
+        "to"
+      ]
     },
     "DirenvConfig": {
       "type": "object",
@@ -5290,7 +5301,7 @@
       "type": "object",
       "properties": {
         "pixi_binary": {
-          "$ref": "#/$defs/Either",
+          "$ref": "#/$defs/Either2",
           "default": [
             "pixi"
           ]
@@ -5346,7 +5357,7 @@
       },
       "additionalProperties": false
     },
-    "Either": {
+    "Either2": {
       "anyOf": [
         {
           "type": "string"
@@ -5454,7 +5465,7 @@
           "default": "pyenv "
         },
         "python_binary": {
-          "$ref": "#/$defs/Either",
+          "$ref": "#/$defs/Either2",
           "default": [
             [
               "python"
@@ -6064,7 +6075,7 @@
           "default": "S "
         },
         "compiler": {
-          "$ref": "#/$defs/Either",
+          "$ref": "#/$defs/Either2",
           "default": [
             "solc"
           ]
@@ -6717,7 +6728,7 @@
           "default": ""
         },
         "when": {
-          "$ref": "#/$defs/Either2",
+          "$ref": "#/$defs/Either3",
           "default": false
         },
         "require_repo": {
@@ -6725,7 +6736,7 @@
           "default": false
         },
         "shell": {
-          "$ref": "#/$defs/Either",
+          "$ref": "#/$defs/Either2",
           "default": []
         },
         "description": {
@@ -6784,7 +6795,7 @@
       },
       "additionalProperties": false
     },
-    "Either2": {
+    "Either3": {
       "anyOf": [
         {
           "type": "boolean"


src/modules/directory.rs
```diff
@@ -292,9 +293,17 @@ fn real_path<P: AsRef<Path>>(path: P) -> PathBuf {
 /// Given a list of (from, to) pairs, this will perform the string
 /// substitutions, in order, on the path. Any non-pair of strings is ignored.
 fn substitute_path(dir_string: String, substitutions: &IndexMap<String, &str>) -> String {
+    let root_regex = Regex::new(r"^/").unwrap();
+    let root_regex_final = Regex::new(r"^root/?").unwrap();
```
christophehenry: Can this be a `const` defined outside the function?

kkoniuszy: A `Regex` can't be `const`, but you can make it `static` using a static cell such as  `std::sync::OnceLock`.

You can do this:
```rust
fn root_regex() -> &'static Regex {
    static ROOT_REGEX: OnceLock<Regex> = OnceLock::new();

    ROOT_REGEX.get_or_init(|| {
        Regex::new(r"^/").unwrap()
    })
}

fn root_regex_final() -> &'static Regex {
    static ROOT_REGEX_FINAL: OnceLock<Regex> = OnceLock::new();

    ROOT_REGEX_FINAL.get_or_init(|| {
        Regex::new(r"^root/?").unwrap()
    })
}
```

This way, the regex will be initialized just once and calling `root_regex()` will give you the same ref every time. The static `OnceLock`s are defined inside functions to hide them from the rest of the module.

christophehenry: Oh nice trick. Is it preferable performance-wise or should I keep the code as it is?

kkoniuszy: `Regex::new` parses and compiles the regex without caching, so it's preferable to do it once and store the Regex in a static. 

I just learned that there's a cleaner way to do this in Rust >= 1.80.0, and it's already used in starship's code.

```rust
static ROOT_REGEX: LazyLock<Regex> = LazyLock::new(|| { Regex::new(r"^/").unwrap() });
```

`LazyLock` is `Deref`, so you can call methods from Regex on it normally, something like `ROOT_REGEX.replace(…)` should still work.

christophehenry: Ok, done. Thank you.

davidkna:
I’m a bit hesitant about supporting this as-is. It could be considered a breaking change, and to me both the current behavior and your proposed changes can make sense depending on the situation. That is, I also wouldn’t necessarily consider the root directory `/` not to contain a separator.

christophehenry:
That seems legit and I was expecting some concerns. So what would be the appropriate way to handle that situation?

davidkna:
Regex support should give enough capabilities for supporting this sort of ting. But since regex-sensitive character-sequences are more common in directory substitutions, it should probably be opt-in (either global or per-substitution).

christophehenry:
Alright I think I can implement something per-substitution. [Borg backup has prefixes](https://borgbackup.readthedocs.io/en/stable/usage/help.html#borg-patterns) how a path should be interpreted like `sh:` for shell-like glob pattern or `re:` for Python regex. I can code something similar where `re:` will be interpreted as a regex substitution. WDYT?

davidkna:
I would prefer to switch substitutions to also allow array-of-objects/tables since the current table-of-substitution was causing issues with nix-configs anyway (in a backward compatible-manner).

For example:
```toml
[directory]
substitutions = [{ from = "^./", to = "/root", regex = true } ]
```

christophehenry:
Alright, I edited my commit with your suggestion. I still need to document the new syntax but in the meantime, WDYT of the code?

src/modules/directory.rs
```diff
@@ -291,10 +292,36 @@ fn real_path<P: AsRef<Path>>(path: P) -> PathBuf {
...
+
+    for substitution in substitutions {
+        if substitution.regex.unwrap_or(false) {
+            if let Ok(re) = Regex::new(substitution.from.as_str()) {
+                substituted_dir = re
+                    .replace(substituted_dir.as_str(), substitution.to)
+                    .to_string()
+            } else {
+                continue;
+            }
```
christophehenry: Maybe this would be controversial. In case of a bad regex, we silently don't do anything and move on to the next. Also, this probably needs a test?

src/configs/directory.rs
```diff
@@ -1,7 +1,29 @@
...
+#[derive(Clone, Deserialize, Serialize)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+pub struct SubstitutionConfig<'a> {
+    pub from: String,
+    pub to: &'a str,
+    pub regex: Option<bool>,
```
davidkna:
```suggestion
    #[serde(default)]
    pub regex: bool,
```

src/modules/directory.rs
davidkna: This could use `log::warn!`. For better testability returning an `Err` here would be fine too (+ `log::warn!` in upper scope).

christophehenry:
Ok, done. Thank you for your review. I have a few questions left. Also, where can I document the new synthax?

src/modules/directory.rs
```diff
@@ -74,7 +75,13 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let dir_string = remove_extended_path_prefix(dir_string);
 
     // Apply path substitutions
-    let dir_string = substitute_path(dir_string, &config.substitutions);
+    let dir_string = match substitute_path(dir_string.clone(), &config.substitutions) {
+        Ok(result) => result,
+        Err(err) => {
+            log::warn!("{err}");
```
christophehenry: Is it ok to just print the error here or should I put something more descriptive?

src/modules/directory.rs
```diff
@@ -74,7 +75,13 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let dir_string = remove_extended_path_prefix(dir_string);
 
     // Apply path substitutions
-    let dir_string = substitute_path(dir_string, &config.substitutions);
+    let dir_string = match substitute_path(dir_string.clone(), &config.substitutions) {
```
christophehenry, line 78: Is it ok to clone here? Otherwise, the value is moved and I can't move it later (if I understood correctly).

src/modules/directory.rs
```diff
@@ -291,12 +298,35 @@ fn real_path<P: AsRef<Path>>(path: P) -> PathBuf {
 ///
 /// Given a list of (from, to) pairs, this will perform the string
 /// substitutions, in order, on the path. Any non-pair of strings is ignored.
-fn substitute_path(dir_string: String, substitutions: &IndexMap<String, &str>) -> String {
+fn substitute_path(
```
davidkna, line 301: Please add a test with an invalid regex.

src/modules/directory.rs
davidkna: Please add something more descriptive, e.g. "Invalid regex in directory substitutions: {err}".

src/modules/directory.rs
davidkna: Cloning should be fine, alternatively you could have to return the `String` along with the `Error`.

davidkna (changes requested):

davidkna:
Please also update the documentation.

src/modules/directory.rs
christophehenry: Should I also test the error was logged too? How can I test that? Is there a way to mock a function in Rust?

davidkna: In this case I would just test `substitute_path` itself (this is why I suggested using `Result` over logging directly in the first place).

christophehenry: Ok, so I added the test. The documentation is failing but I suppose this is because translations don't match anymore. Is there something I can do for that?

docs/config/README.md
```diff
@@ -1205,12 +1205,36 @@ it would have been `nixpkgs/pkgs`.
 
 | Advanced Option             | Default | Description                                                                                                                                                            |
 | --------------------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `substitutions`             |         | A table of substitutions to be made to the path.                                                                                                                       |
+| `substitutions`             |         | Array of substitutions to be made to the path.                                                                                                                         |
```
davidkna:
```suggestion
| `substitutions`             |         | An Array or table of substitutions to be made to the path.                                                                                                                         |
```

docs/config/README.md
```diff
@@ -1205,12 +1205,36 @@ it would have been `nixpkgs/pkgs`.
...
 | `fish_style_pwd_dir_length` | `0`     | The number of characters to use when applying fish shell pwd path logic.                                                                                               |
 | `use_logical_path`          | `true`  | If `true` render the logical path sourced from the shell via `PWD` or `--logical-path`. If `false` instead render the physical filesystem path with symlinks resolved. |
 
+**Warning**: the syntax for `subscriptions` changed.
```
davidkna:
```suggestion
```

christophehenry:
@davidkna I integrated your changes.

davidkna (approved):

davidkna:
Thanks for the contribution @christophehenry!

PR #7151, opened 2025-11-28
fix(fish): wrap prompt to terminal width to prevent truncation by shell

### LINKED ISSUE #6663: Truncation issue with small terminal width in Fish shell
pkg_version:1.22.1
branch:master
commit_hash:d605196
build_time:2025-01-11 17:06:48 +00:00
build_env:rustc 1.84.0 (9fc6b4312 2025-01-07),stable-aarch64-apple-darwin

Generated bug report:
#### Current Behavior
When the fish shell is used in a terminal with a small width, the prompt is truncated instead of wrapping to a new line. This makes it difficult to see the full prompt information.

I don't want to increase width because I put Terminal on the right-side panel in Visual Code

Image
Image

#### Expected Behavior
It should wrap the prompt without truncation. For example in ZSH:

Image

#### Environment
- Starship version: 1.22.1
- fish version: fish, version 4.0.1-g67b6aff
- Operating system: Mac OS 15.3.2
- Terminal emulator: vscode 1.99.0-insider
- Git Commit Hash: d605196
- Branch/Tag: master
- Rust Version: rustc 1.84.0 (9fc6b4312 2025-01-07)
- Rust channel: stable-aarch64-apple-darwin release
- Build Time: 2025-01-11 17:06:48 +00:00

kkoniuszy:
#### Description
- Wrap prompts longer than the terminal width
- Remove a no longer needed workaround for grapheme width. This also fixes width calculation for emoji variants of text symbols   
- Make a workaround for fish <3.2 conditional on the fish version variable 

This PR wraps the main prompt on `fish` by inserting line breaks, which makes it look the same as on other shells. The wrapping is based on grapheme clusters, and the ANSI escape sequences are left untouched.

An existing workaround for calculating width was removed, as the current version of `unicode_width` appears to compute the widths correctly and the test for the past bug no longer fails with `UnicodeWidthStr`. This was needed because the `UnicodeWidthChar` way did not correctly determine the width of symbols presented as emojis using `\u{FE0F}`, such as the double-width `❄️`, instead returning 1 which would be the width of the regular `❄`. 

While implementing the wrapping, I noticed that a workaround for `fish` < 3.2 added in #865 is still in place, and a [previous PR](https://github.com/starship/starship/pull/5500) removing this was rejected because the affected `fish` versions are still distributed. I added a minor change that limits this workaround to the affected fish versions. 

#### Motivation and Context
Fish truncates prompts that are longer than the terminal width, and it starts from the beginning, which can easily make the current directory name unreadable with the default starship configuration. In https://github.com/fish-shell/fish-shell/issues/11341, it was stated that this behaviour is [intended](https://github.com/fish-shell/fish-shell/blob/9814bae727ff891c581a7d9a9149afd2b7e77615/src/screen.rs#L1522) and the prompt should handle wrapping or splitting text wider than `$COLUMNS`. 

Closes #6663

image

#### Checklist:

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -11,6 +11,13 @@ function __starship_set_job_count --description 'Set STARSHIP_JOBS using fish jo
     set -g STARSHIP_JOBS $__count
 end
 
+function __starship_set_fish_needs_clear --description 'Set STARSHIP_FISH_NEEDS_CLEAR on fish <= 3.2.x'
+    set vs (string split . $version)
+    if test \( $vs[1] -eq 3 -a $vs[2] -le 2 \) -o \( $vs[1] -lt 3 \)
+        set -gx STARSHIP_FISH_NEEDS_CLEAR 1
+    end
+end
+
 function fish_prompt
     switch "$fish_key_bindings"
         case fish_hybrid_key_bindings fish_vi_key_bindings fish_helix_key_bindings
@@ -25,6 +32,7 @@ function fish_prompt
     set STARSHIP_DURATION "$CMD_DURATION$cmd_duration"
 
     __starship_set_job_count
+    __starship_set_fish_needs_clear
 
     if test "$TRANSIENT" = "1"
         set -g TRANSIENT 0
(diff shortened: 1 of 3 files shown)


davidkna:
I don't think this is a good idea. Please request an option in fish to disable this.

PR #7154, opened 2025-11-30
fix(git): Basic Reftable compatibility and future-proofing

### LINKED ISSUE #6086: [Git 2.45] Reftable storage backend is not supported
pkg_version:1.19.0
branch:
commit_hash:
build_time:1980-01-01 00:00:00 +00:00
build_env:rustc 1.78.0 (9b00956e5 2024-04-29) (built from a source tarball),

#### Current Behavior

Git 2.45 is needed to reproduce this bug, since it makes use of the new reference storage backend "[reftable](https://github.blog/2024-04-29-highlights-from-git-2-45/#preliminary-reftable-support)".

Just clone a git repository with reftable used:

```bash
git clone --ref-format=reftable git@github.com:starship/starship.git
```

Then go to the newly cloned directory. Git information are not displayed properly:

```bash
 …/starship [🌱 HEAD](+0/-0)
```

#### Expected Behavior

The bit branch and git commit should be shown:

```bash
 …/starship [🌱 master](27cc4bc)(+0/-0)
```

#### Environment
- Starship version: 1.19.0
- zsh version: zsh 5.9 (aarch64-apple-darwin23.5.0)
- Operating system: Mac OS 14.5.0
- Terminal emulator: Apple_Terminal 453
- Git Commit Hash:
- Branch/Tag:
- Rust Version: rustc 1.78.0 (9b00956e5 2024-04-29) (built from a source tarball)
- Rust channel:  release

#### Starship Configuration

```toml
[git_branch]
format = "\\[[$symbol$branch]($style)\\]"
style = "bold yellow"
symbol = "🌱 "

[git_commit]
commit_hash_length = 7
format = "[\\($hash\\)]($style)( [\\($tag\\)]($style))"
only_detached = false
style = "bold white"
tag_disabled = false
tag_symbol = "🔖"

[git_metrics]
added_style = "bold dimmed green"
deleted_style = "bold dimmed red"
disabled = false
format = "\\([+$added]($added_style)/[-$deleted]($deleted_style)\\)"
only_nonzero_diffs = false

[git_state]
format = "[\\($state( $progress_current of $progress_total)\\)]($style) "

[git_status]
ahead = "🏎💨×$count"
behind = "🐢×$count"
conflicted = "⚔"
deleted = "🗑×$count"
diverged = "🏎💨×$ahead_count 🐢×$behind_count"
format = "([\\[$all_status$ahead_behind\\]]($style))"
modified = "📝×$count"
renamed = "📛×$count"
staged = "🗃×$count"
stashed = "📦"
style = "bold white"
untracked = "🛤"
```

Byron:
#### Description

With `gitoxide` as main driver, we may have certain benefits, but these come at a code: compatibility. Git, despite slow moving, keeps adding new features and at least for now `gitoxide` will be late for the party.

This PR introduces the machinery to automatically fallback to using the Git executable when unsupported features are detected.
For now, this is only implemented for reftable support, but follow-ups are possible for instance for SHA256.

#### Motivation and Context

Originally I wanted to support reftables, but realised that it's very straightforward to detect reftables and fallback automatically. So instead of just for `git_branch`, this was also implemented for `git_status`.

Since branches are quite straightforward as a feature, I refrained from adding a `use_cli_executable` option there as well, and we can auto-detect the need for the CLI with certainty.

Closes #6086.

- [x] I have tested using **Linux** via https://github.com/Byron/starship/pull/1
- [x] I have tested using **Windows** via https://github.com/Byron/starship/pull/1

#### Checklist:

- [ ] ~~I have updated the documentation accordingly.~~ - I didn't know what to update.
    - I went for coverage and found that stashes aren't actually handled correctly in the CLI implementation. But that regression already exited and I didn't try to fix it. Otherwise, it's now clear that the CLI implementation is equal to the `gitoxide` one, at least for what's tested here.

### DIFF
--- a/src/modules/git_branch.rs
+++ b/src/modules/git_branch.rs
@@ -38,16 +38,14 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let (branch_name, remote_branch, remote_name) = if uses_reftables(&gix_repo) {
         // Use git executable for branch information
         match get_branch_info_from_git(context, repo) {
-            Some((branch, upstream)) => {
+            Some((branch, remote_branch)) => {
                 // Successfully got branch name, parse upstream into remote_name and remote_branch
-                let (remote_name, remote_branch) = upstream
-                    .map(|upstream| {
-                        (
-                            find_longest_matching_remote_name(
-                                upstream.as_ref(),
-                                &gix_repo.remote_names(),
-                            ),
-                            Some(upstream),
+                let remote_name = remote_branch
+                    .as_ref()
+                    .map(|remote_branch| {
+                        find_longest_matching_remote_name(
+                            remote_branch.as_ref(),
+                            &gix_repo.remote_names(),
                         )
                     })
                     .unwrap_or_default();
--- a/src/modules/git_status.rs
+++ b/src/modules/git_status.rs
@@ -249,8 +249,8 @@ fn get_repo_status(
     let has_untracked = !config.untracked.is_empty();
     let git_config = gix_repo.config_snapshot();
     if config.use_git_executable
-        || uses_reftables(&repo.repo.to_thread_local())
         || repo.fs_monitor_value_is_true
+        || uses_reftables(&repo.repo.to_thread_local())
         || gix_repo.index_or_empty().ok()?.is_sparse()
     {
         let mut args = vec!["status", "--porcelain=2"];
--- a/src/modules/git_metrics.rs
+++ b/src/modules/git_metrics.rs
@@ -6,6 +6,7 @@ use regex::Regex;
 
 use super::Context;
 use crate::configs::git_status::GitStatusConfig;
+use crate::modules::git_status::uses_reftables;
 use crate::{
     config::ModuleConfig, configs::git_metrics::GitMetricsConfig, formatter::StringFormatter,
     formatter::string_formatter::StringFormatterError, module::Module,
@@ -33,6 +34,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     // TODO: remove this special case once `gitoxide` can handle sparse indices for tree-index comparisons.
     let stats = if repo.fs_monitor_value_is_true
         || status_config.use_git_executable
+        || uses_reftables(&repo.repo.to_thread_local())
         || gix_repo.index_or_empty().ok()?.is_sparse()
     {
         let mut git_args = vec!["diff", "--shortstat"];
@@ -412,6 +414,11 @@ mod tests {
     use crate::test::{FixtureProvider, ModuleRenderer, fixture_repo};
     use nu_ansi_term::Color;
 
+    const NORMAL_AND_REFTABLES: [FixtureProvider; 2] =
+        [FixtureProvider::Git, FixtureProvider::GitReftable];
+    const BARE_AND_REFTABLE: [FixtureProvider; 2] =
+        [FixtureProvider::GitBare, FixtureProvider::GitBareReftable];
+
     #[test]
     fn shows_nothing_on_empty_dir() -> io::Result<()> {
         let repo_dir = tempfile::tempdir()?;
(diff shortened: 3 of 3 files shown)


Byron:
Please note that this is best reviewed with "Hide whitespace" enabled.

Screenshot 2025-11-30 at 14 56 55

akkadaya:
can't wait for this PR to be merged, thanks @Byron

akkadaya:
does it need approval?

src/modules/git_branch.rs
```diff
@@ -31,12 +34,64 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                    .map(|upstream| {
+                        (
+                            find_longest_matching_remote_name(
+                                upstream.as_ref(),
+                                &gix_repo.remote_names(),
+                            ),
+                            Some(upstream),
+                        )
+                    })
+                    .unwrap_or_default();
```
davidkna: I would prefer aliasing `upstream` to `remote_branch` and handling `remote_name` separately to make the control flow more clear.

src/modules/git_status.rs
```diff
@@ -243,6 +249,7 @@ fn get_repo_status(
     let has_untracked = !config.untracked.is_empty();
     let git_config = gix_repo.config_snapshot();
     if config.use_git_executable
+        || uses_reftables(&repo.repo.to_thread_local())
         || repo.fs_monitor_value_is_true
```
davidkna: It's probably a bit faster to check `repo.fs_monitor_value_is_true` first:

```suggestion
        || repo.fs_monitor_value_is_true
        || uses_reftables(&repo.repo.to_thread_local())
```

In addition, please also add the `uses_reftables` check to the `git_metrics` module.

davidkna (commented):
Thanks for the PR! I have raised a few minor concerns but overall lgtm.

src/modules/git_status.rs
Byron: Also a great catch!
And `git_metrics` has the conditional now as well.

src/modules/git_branch.rs
Byron: Such a great catch, `Some(upstream)` definitely isn't needed there, and it does look much cleaner now.

Byron:
Thanks a lot for the review!

I am still working on the `git-metrics` tests of which 4 regress when using Git, but I will see if this is something with a quick fix, or something to accept for now as prior issue that can be fixed in a follow-up.

Byron:
Thanks for the approval, I hoe to get CI green tomorrow.

Byron:
This should go through now. Please note that while considering to fix the "staged changes don't show with Git" issue, I noticed that this would require another call via `git diff --cached` (or something like that), which seems to have been a long(er) standing issue that maybe nobody complained about yet.
Thus, it's probably not too important in practice and I adjusted the test expectations with TODOs on how it *can* be fixed.

Besides, when running clippy, I am presented with this.

```
error: wildcard matches known variants and will also match future added variants
   --> src/modules/os.rs:384:13
    |
384 |             _ => "",
    |             ^ help: try: `Type::ALTLinux | Type::Elementary | Type::Ios | Type::Zorin | _`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#wildcard_enum_match_arm
note: the lint level is defined here
   --> src/modules/os.rs:325:16
    |
325 |         #[deny(clippy::wildcard_enum_match_arm)]
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Maybe that's an opportunity to support even more diverse platforms with a dedicated icon. While tempted, I held back doing that as I really don't know what's desirable there 😅. So another thing that could be done is to allow that clippy lint here, while probably rightfully assuming that there might be contributions by users of these platforms when there is demand for an icon mapping.

akkadaya:
Ready to go I guess, can't wait for it to be released!

davidkna:
Thanks for handling this @Byron!

PR #7159, opened 2025-12-03
ci: add install script workflow

cappyzawa:
#### Description

Adds a new GitHub Actions workflow to test the install script (`install/install.sh`).

The workflow includes:
- **shellcheck job**: Lints the install script using `reviewdog/action-shellcheck`
  - Temporarily excludes SC3045 (POSIX compatibility issue with `read -p`)
  - Provides inline PR comments for any issues found
- **test_install_script job**: Tests the installation method documented on https://starship.rs/
  - Runs on both Linux and macOS
  - Simulates `curl -sS https://starship.rs/install.sh | sh` using local file
  - Verifies the installed binary works

#### Motivation and Context

Issue #7133 reported that the installation method documented on https://starship.rs/ stopped working after changes. This CI workflow will catch such regressions before they reach users.

> [!NOTE]
> After #7138 is merged, we should add test cases for interactive mode (handling "no" responses without `--yes` flag) to ensure proper error handling when stdin is not a TTY. This will be addressed in a separate PR.

N/A

#### Checklist:


.github/workflows/install-script.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v6
+
+      - name: Lint | Run shellcheck
+        uses: reviewdog/action-shellcheck@v1
+        with:
+          path: "install/install.sh"
+          severity: "warning"
+          # TODO: Remove exclude after fixing POSIX compatibility
+          # SC3045: read -p is not POSIX compliant
+          exclude: "SC3045"
```
davidkna: Maybe add this to `install.sh` to also trigger the workflow.

.github/workflows/install-script.yml
```diff
@@ -0,0 +1,47 @@
...
+    permissions:
+      checks: write
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Setup | Checkout
+        uses: actions/checkout@v6
+
+      - name: Lint | Run shellcheck
+        uses: reviewdog/action-shellcheck@v1
```
davidkna: I think I would prefer just running the CLI-tool here, it should already be installed on the ubuntu runner. Something like `shfmt` would also be interesting.

cappyzawa: Addressed in bbd1328e (switch to shellcheck CLI) and ce74e2b3 (add shfmt check and format scripts).

.github/workflows/install-script.yml
cappyzawa: I added `# shellcheck disable=SC3045` before the `read -p` line in bbd1328e. Is this what you had in mind?

davidkna:
LGTM, but please fix the conflicts.

cappyzawa:
@davidkna Thanks for your review!
I've fixed this 👍

PR #7171, opened 2025-12-11
fix: Try to create config directory before modifying config

### LINKED ISSUE #7129: unable to edit config file -- unable to find config file?
#### Current Behavior
I'm trying to figure out where my starship configuration file is.   The documentation [here](https://starship.rs/config/) lists `~/.config/starship.toml` as the default path, but it's not at all clear if this is the path on all platforms, or only on linux-like platforms.  I'm on Windows.

Running `starship config` brings up a Notepad window that just says `The system cannot find the path specified` but there is no hint about what the path is.  There's nothing printed out to stdout/stderr.

Running `starship config foo bar` prints this (I have `RUST_BACKTRACE=full`):

```
thread 'main' (58072) panicked at src\configure.rs:252:10:
Error writing starship config: Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }
stack backtrace:
   0:     0x7ff76e8f4460 - <unknown>
   1:     0x7ff76e1fd5ca - <unknown>
   2:     0x7ff76e8f27a4 - <unknown>
   3:     0x7ff76e8f4264 - <unknown>
   4:     0x7ff76e8f23cf - <unknown>
   5:     0x7ff76e8f988a - <unknown>
   6:     0x7ff76e8f97df - <unknown>
   7:     0x7ff76e8f97ce - <unknown>
   8:     0x7ff76e9b6241 - <unknown>
   9:     0x7ff76e9b69b0 - <unknown>
  10:     0x7ff76e8eae70 - <unknown>
  11:     0x7ff76e8efd55 - <unknown>
  12:     0x7ff76e135e4a - <unknown>
  13:     0x7ff76e127e46 - <unknown>
  14:     0x7ff76e146d4b - <unknown>
  15:     0x7ff76e98cf1c - <unknown>
  16:     0x7ffb5a44259d - BaseThreadInitThunk
  17:     0x7ffb5b0eaf78 - RtlUserThreadStart
```

Again, there is no hint about what path starship is trying to use

#### Expected Behavior

I would like `starship config` to tell me the path it is configured to use for its config file.  And I would like `starship config foo bar` to not panic

I do not have any starship envvars set.  `STARSHIP_CONFIG` is not set

#### Environment
- Starship version: 1.24.1
- Git Commit Hash: 378f3e4
- Branch/Tag: master
- Rust Version: rustc 1.91.1 (ed61e7d7e 2025-11-07)
- Build Time: 2025-11-16 11:53:38 +00:00

spenserblack:
#### Description

Before calling an editor or creating the configuration file, this will first try to create the folder that contains the config file, if there's a folder in the config file's path.

#### Motivation and Context

It helps starship "work out of the box" a little bit more, especially on Windows where there's a decent chance that `~/.config` hasn't been created yet.

Closes #7129

I haven't written any test cases for this, just manually checked the presence of folders after running the executable. But I could look into making the new function a bit more testable if necessary. I put a call to this function in test code, though, so it does get implicitly tested.

#### Checklist:


eminence:
Thanks for the patch, this would definitely help my experience that I reported in https://github.com/starship/starship/issues/7129

However it doesn't really solve the problem of telling the user what paths to check when things go wrong.  For example, I still get this unhelpful error when testing your commit:

```
>cargo run -- config foo bar
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.61s
     Running `target\debug\starship.exe config foo bar`

thread 'main' panicked at src\configure.rs:250:37:
Error creating configuration directory: Os { code: 183, kind: AlreadyExists, message: "Cannot create a file when that file already exists." }
error: process didn't exit successfully: `target\debug\starship.exe config foo bar` (exit code: 101)

```

spenserblack:
> `kind: AlreadyExists`

~~Oh, ouch, looks like I created a new error with this patch. I assumed that `create_dir_all` functioned much like `mkdir -p`, and would still be OK if the directory existed.~~

Edit: Nevermind, just tested on both Linux and Windows and it seems to work fine. I can't reproduce that error on Windows.

:thinking: Do you have a *file* with the same name as the configuration directory?

eminence:
> Do you have a file with the same name as the configuration directory?

Yes, exactly.  But a similar type of error would happen if the directory couldn't be created for other reasons (permission errors, for example).  So all I'm looking for here is a better error message that tells the user the full path of the config file

spenserblack:
> all I'm looking for here is a better error message that tells the user the full path of the config file

I'm going to leave that up to the reviewer(s) to decide the best course of action regarding that. Personally, I would unconditionally print `"attempting to edit {config_path}"`, so that the user is is *always* alerted to the current config file path when calling `starship config`, regardless of error status. But the maintainers may prefer something else.

PR #7189, opened 2025-12-27
feat: add Maven module

### LINKED ISSUE #7188: Add a `maven` module
## Feature Request

The [Gradle module](https://starship.rs/config/#gradle) exposes the [Gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) version. Maven also has wrapper support, and I'd like an equivalent Starship module for Maven. I suspect that this has not been done yet because Maven wrappers are not as widely used as Gradle wrappers, probably because the exact Maven version doesn't matter as much (Maven is much more stable than Gradle), but I expect that to change with the upcoming release of [Maven 4](https://maven.apache.org/whatsnewinmaven4.html).

rschmitt:
#### Description
This adds a `maven` module that exposes the presence of a Maven project in the current directory, as well as the Maven wrapper version if present. In terms of behavior and configuration, this feature is essentially identical to the `gradle` module.

#### Motivation and Context

This provides feature parity for Maven projects, given the existing support for Gradle. Maven 4 is currently in release candidate; as this is the first new major version of Maven in over fifteen years, I expect that developers will start paying more attention to the specific Maven version used by a particular project, and I also expect that use of the Maven wrapper will become more common.

Closes #7188.

In addition to the unit tests, I tested this manually in various projects and folders and with various config options (e.g. `version_format`, `recursive`). I tested various Maven wrapper versions, including 3.9.12 and 4.0.0-rc-5.

#### Checklist:

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -305,6 +305,7 @@ $julia\
 $kotlin\
 $gradle\
 $lua\
+$maven\
 $nim\
 $nodejs\
 $ocaml\
@@ -2914,6 +2915,41 @@ By default the module will be shown if any of the following conditions are met:
 [lua]
 format = 'via [🌕 $version](bold blue) '
 ```
+
+## Maven
+
+The `maven` module indicates the presence of a Maven project in the current directory. If the [Maven Wrapper](https://maven.apache.org/wrapper/) is enabled, the Maven version will be parsed from `.mvn/wrapper/maven-wrapper.properties` and shown.
+
+By default the module will be shown if any of the following conditions are met:
+
+- The current directory contains a `pom.xml` file.
+- The current directory contains a `.mvn/wrapper/maven-wrapper.properties` file.
+
+If you use an alternate POM syntax (for example `pom.hocon`), add its filename to `detect_files`.
+
+### Options
+
+| Option              | Default                              | Description                                                               |
+| ------------------- | ------------------------------------ | ------------------------------------------------------------------------- |
+| `format`            | `'via [$symbol($version )]($style)'` | The format for the module.                                                |
+| `version_format`    | `'v${raw}'`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
+| `symbol`            | `'🅼 '`                               | A format string representing the symbol of Maven.                         |
+| `detect_extensions` | `[]`                                 | Which extensions should trigger this module.                              |
+| `detect_files`      | `['pom.xml']`                        | Which filenames should trigger this module.                               |
+| `detect_folders`    | `['.mvn']`                           | Which folders should trigger this module.                                 |
+| `style`             | `'bold bright-cyan'`                 | The style for the module.                                                 |
+| `disabled`          | `false`                              | Disables the `maven` module.                                              |
+| `recursive`         | `false`                              | Enables recursive finding for the `.mvn` directory.                       |
+
+### Variables
+
+| Variable | Example  | Description                          |
+| -------- | -------- | ------------------------------------ |
+| version  | `v3.2.0` | The version of `maven`               |
+| symbol   |          | Mirrors the value of option `symbol` |
+| style*   |          | Mirrors the value of option `style`  |
+
+*: This variable can only be used as a part of a style string
 
 ## Memory Usage
 
--- a/src/modules/maven.rs
+++ b/src/modules/maven.rs
@@ -1,0 +1,226 @@
+use std::path::Path;
+
+use crate::{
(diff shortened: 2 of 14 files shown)


docs/config/README.md
```diff
@@ -2899,6 +2900,40 @@ By default the module will be shown if any of the following conditions are met:
...
+
+### Options
+
+| Option           | Default                              | Description                                                               |
+| ---------------- | ------------------------------------ | ------------------------------------------------------------------------- |
+| `format`         | `'via [$symbol($version )]($style)'` | The format for the module.                                                |
+| `version_format` | `'v${raw}'`                          | The version format. Available vars are `raw`, `major`, `minor`, & `patch` |
+| `symbol`         | `'🅼 '`                               | A format string representing the symbol of Maven.                         |
+| `detect_files`   | `['pom.xml']`                        | Which filenames should trigger this module.                               |
+| `detect_folders` | `['.mvn']`                           | Which folders should trigger this module.                                 |
```
davidkna: Please also handle `detect_extensions`

src/modules/maven.rs
```diff
@@ -0,0 +1,227 @@
...
+    context::Context,
+    formatter::{StringFormatter, VersionFormatter},
+    module::Module,
+    utils,
+};
+
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("maven");
+    let config = MavenConfig::try_load(module.config);
+    let wrapper_properties = get_wrapper_properties_file(context, config.recursive);
```
davidkna: Please handle `is_maven_project` first for performance. Running this early also diverges from the `gradle` behaviour. If you are planning to keep it this way please add a `TODO` comment to the gradle module to streamline them in the future.

Another potential alternative would be to conditionally handle `is_maven_project` with `begin_ancestor_scan` depending on the value of `upwards` (only using the config options), giving users more control.

src/modules/maven.rs
```diff
@@ -0,0 +1,227 @@
...
+        distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.12/apache-maven-3.9.12-bin.zip
+    */
+    let version = wrapper_properties
+        .lines()
+        .find(|line| line.starts_with("distributionUrl="))?
+        .rsplit_once('/')?
+        .1
+        .strip_prefix("apache-maven-")?
+        .rsplit_once('-')?
+        .0;
```
davidkna, line 78: I think it would be more reliable to use the `/3.9.12/` part of the url.

src/modules/maven.rs
```diff
@@ -0,0 +1,227 @@
...
+    Some(version.to_string())
+}
+
+/// Tries to find the maven-wrapper.properties file.
+fn get_wrapper_properties_file(context: &Context, recursive: bool) -> Option<String> {
```
davidkna: Please explore sharing this function with the `gradle` module.

src/modules/maven.rs
```diff
@@ -0,0 +1,227 @@
...
+        return Some(properties);
+    }
+
+    // Try parent directories if recursive
+    if recursive {
+        for dir in context.current_dir.ancestors().skip(1) {
+            if let Some(properties) = read_wrapper_properties(dir) {
+                return Some(properties);
+            }
+        }
```
davidkna: Please make use of `context.begin_ancestor_scan()`

rschmitt: OK, but note that this doesn't appear to work:

```rs
            .begin_ancestor_scan()
            .set_files(&[".mvn/wrapper/maven-wrapper.properties"])
            .scan()
```

I'll just look for the `.mvn` folder instead, but it would be handy if I could scan for a full relative path.

src/modules/maven.rs
rschmitt: Any reason? A regex would have to account for various suffixes like `-SNAPSHOT` and whatever else might be possible. On the Gradle side, this approach has worked well for me, including for oddball use cases like custom Gradle distributions with non-standard version numbers.

rschmitt:
I updated the files in the `toml/` directory by following the existing configuration for the `gradle` module where present. When I regenerated `config-schema.json`, there were some unrelated changes due to things like JSON field ordering, so I cherry-picked the relevant ones. Let me know if I should change anything in this approach.

davidkna (approved):

rschmitt:
I rebased the change. Let me know if you need anything else.

davidkna:
Thanks for the contribution @rschmitt!

PR #7193, opened 2025-12-30
fix(helm): remove deprecated --client flag from helm version command

### LINKED ISSUE #7161: Helm version not included in Starship output
#### Current Behavior
Helm version is missing from Starship output, whilst symbol, etc remain present

#### Expected Behavior
Helm version printed in Starship when in a Helm context

It looks as though the `--client` flag was [recently removed](https://github.com/helm/helm/pull/31301) from the `helm version` command

#### Possible Solution
Removing the `--client` flag from [`src/modules/helm.rs#L37`](https://github.com/starship/starship/blob/61a76a26658fe7ac7eb802e32bf2c7e1fd1f3f66/src/modules/helm.rs#L37), though this may require updating the implementation for parsing version output from Helm v2

#### Environment
- Starship version: 1.23.0
- Shell type: bash
- Shell version: 5.2.15(1)-release
- Shell plugin manager: N/A
- Operating system: Debian 13

darwin808:
## Summary
- Remove the deprecated `--client` flag from `helm version` command
- The `--client` flag was removed in a recent Helm version (see [helm/helm#31301](https://github.com/helm/helm/pull/31301))
- This was causing the helm module to silently fail when detecting the version

Fixes #7161

## Test plan
- [x] Existing helm module tests pass
- [x] Verified the fix works with the updated mock command

### DIFF
--- a/src/modules/helm.rs
+++ b/src/modules/helm.rs
@@ -61,16 +61,11 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 fn parse_helm_version(helm_stdout: &str) -> Option<String> {
     // `helm version --short` output looks like this:
     // v3.1.1+gafe7058
-    // For Helm 2, output looks like this:
-    // Client: v2.16.9+g8ad7037
     let version = helm_stdout
-        // split into ("v3.1.1","gafe7058") or ("Client: v3.1.1","gafe7058")
+        // split into ("v3.1.1","gafe7058")
         .split_once('+')
-        // return "v3.1.1" or "Client: v3.1.1"
-        .map_or(helm_stdout, |x| x.0)
-        // return "v3.1.1" or " v3.1.1"
-        .trim_start_matches("Client: ")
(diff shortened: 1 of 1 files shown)


cappyzawa (commented):
One suggestion: Consider removing the Helm 2 comment. Helm 2 was EOL in November 2020 (https://helm.sh/blog/helm-2-becomes-unsupported/), and without the `--client` flag, `helm version --short` attempts to connect to Tiller - if that connection fails, the command returns a non-zero exit code and `exec_cmd` returns `None`, so the parsing code is never reached anyway.

I agree that starship doesn't need to support Helm 2 at this point.

darwin808:
@cappyzawa 
Good point, done. Also removed the Client: parsing code and test since they're effectively dead code now.

cappyzawa (approved):
LGTM!

davidkna:
Thanks for the fix @darwin808, and thanks for reviewing again @cappyzawa!

PR #7210, opened 2026-01-04
perf: add get_repo_workdir

juliesaia:
#### Description

Adds `get_repo_workdir` to `context.rs`, similar to `get_repo` but skips logic unnecessary for just getting the workdir.

In a few modules (directory and dotnet), only the workdir from get_repo() is used. In my setup `directory` was taking ~10ms to run. The new `get_repo_workdir` function is identical for this use case and runs instantaneously in comparison.

#### Motivation and Context

Perf improvement for directory and dotenv modules

Existing test suite passes

#### Checklist:

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -54,9 +54,6 @@ pub struct Context<'a> {
     /// Private field to store Git information for modules who need it
     repo: OnceLock<Result<Repo, Box<gix::discover::Error>>>,
 
-    /// Cached workdir from lightweight git discovery (avoids full repo open)
-    repo_workdir: OnceLock<Option<PathBuf>>,
-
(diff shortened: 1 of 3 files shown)


davidkna:
If possible I would prefer moving stuff outside `get_repo` into the appropraite modules, or switching to a on-demand lazy evaluation approach.

juliesaia:
> If possible I would prefer moving stuff outside `get_repo` into the appropraite modules, or switching to a on-demand lazy evaluation approach.

done! at least if i interpreted that correctly 😅 moved everything into individual gix::discover calls in the 2 modules its used

ty for the feedback!

davidkna:
No, I meant making `get_repo()` fast instead.

PR #7211, opened 2026-01-05
fix: enable std feature for whoami crate

gary-kim:
Without the "std" feature enabled, whoami falls back to the stub implementation and Starship shows anonymous@localhost regardless of the actual user or hostname.

#### Description
Enable the `std` feature for the `whoami` crate to include the `unix.rs` implementation.

#### Motivation and Context
The prompt currently shows `anonymous@localhost` from the `whoami` stub implementation.

```
$ ./target/release/starship explain

 " my-username" (<1ms)                  -  The active user's username
 "@my-hostname " (<1ms)         -  The system hostname
...
```

This previously would report ` anonymous` for the username and `localhost` for the hostname.

#### Checklist:

Pure bug fix so I don't think these are necessary?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3878,6 +3878,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "web-sys"
+version = "0.3.83"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


davidkna:
Thanks for the quick fix @gary-kim!

PR #7217, opened 2026-01-09
style(config/default): Update gradle module icon

IThundxr:
#### Description

Switch to `nf-dev-gradle` which is provided by nerdfonts for the gradle module's icon

#### Motivation and Context

Currently the Gradle module's icon is "🅶" which isn't really isn't the best icon considering starship depends on nerdfonts which already provides `nf-dev-gradle`

image

N/A - Icon update

#### Checklist:


davidkna:
Thanks for the contribution, but starship is moving away from using nerd font symbols by default (https://github.com/starship/starship/issues/2563, https://github.com/starship/starship/pull/3544).

Thus, I am closing this. If you prefer to use nerd fonts, you might want to use the [nerd font preset](https://starship.rs/presets/nerd-font).

PR #7224, opened 2026-01-12
WIP(vcs): add support for Jujutsu VCS

### LINKED ISSUE #6076: [FR] Support Jujutsu VCS (git backed)

RodEsp:
#### Description

Uses the new VCS module to add support for [Jujutsu](https://github.com/jj-vcs/jj).

The work here is based on the structure of the `git_` modules, so that Jujutsu info can be configured in whatever manner the user prefers in their starship's `vcs` module config.

e.g.
```toml
[vcs]
order=["jujutsu", "git"]
jujutsu_modules="$jujutsu_state$jujutsu_change$jujutsu_commit$jujutsu_bookmark"
```

The individual `jujutsu_` modules themselves also offer configuration for the different symbols uses to denote status & their color.

#### Motivation and Context

Closes #6076

Here is an example of different jj states shown in starship's prompt with all the jj module defaults.
image

#### Checklist:


src/modules/jujutsu_bookmark.rs
johnstegeman, line 1: I think a more common workflow is when bookmarks dont point to the current change. Many people (myself included) move the bookmark just before pushing, so the bookmark is often on @-

src/modules/jujutsu_status.rs
```diff
@@ -0,0 +1,578 @@
...
+        7 // default to 7
+    } else {
+        length
+    }
+}
+
+const TRACKED_BOOKMARKS_TEMPLATE: &str = r#"if(remote, name ++ "\x1f" ++ tracking_ahead_count.lower() ++ "\x1f" ++ tracking_behind_count.lower() ++ "\n", "")"#;
+
+fn jujutsu_log_template(change_id_length: isize, commit_hash_length: isize) -> String {
+    format!(
```
bryceberger:
```
join("\n",
    change_id.short({}),
    ...
)
```

src/modules/jujutsu_status.rs
```diff
@@ -0,0 +1,578 @@
...
+        log::warn!("{label} should be a positive value, found {length}");
+        7 // default to 7
+    } else {
+        length
+    }
+}
+
+const TRACKED_BOOKMARKS_TEMPLATE: &str = r#"if(remote, name ++ "\x1f" ++ tracking_ahead_count.lower() ++ "\x1f" ++ tracking_behind_count.lower() ++ "\n", "")"#;
+
+fn jujutsu_log_template(change_id_length: isize, commit_hash_length: isize) -> String {
```
bryceberger: `isize` seems like the wrong parameter for these. Negative values don't make sense.

Separately, you _could_ skip allocating this string by using the `config()` jj template function:

```
jj --config custom.starship.change_id_length=... log -T "config('custom.starship.change_id_length').as_integer()"
```

src/modules/jujutsu_status.rs
```diff
@@ -0,0 +1,578 @@
...
+                "-T",
+                &jujutsu_log_template(change_id_length, commit_hash_length),
+            ],
+        )?
+        .stdout;
+
+    let mut lines = output.lines();
+
+    let change_id = lines.next()?.to_string();
+    let bookmarks_str = lines.next().unwrap_or("");
```
bryceberger: These should never be empty, based on the template given. You exit early if the `change_id` is unavailable, why not these?

src/modules/jujutsu_status.rs
```diff
@@ -0,0 +1,578 @@
...
+                TRACKED_BOOKMARKS_TEMPLATE,
+            ],
+        )
+        .map(|output| output.stdout)
+        .unwrap_or_default();
+    merge_tracked_bookmarks(
+        &mut tracked_bookmarks,
+        parse_tracked_bookmarks(&tracked_bookmarks_output),
+    );
+    let bookmarks = if bookmarks_str.is_empty() {
```
bryceberger: You don't need this condition: `"".split('\x1e').collect()` will result in an empty vec already.

src/modules/jujutsu_status.rs
davidkna, line 1: I'd prefer moving the shared implementation to `src/modules/utils/`

src/modules/jujutsu_status.rs
```diff
@@ -0,0 +1,578 @@
...
+fn jujutsu_log_template(change_id_length: isize, commit_hash_length: isize) -> String {
+    format!(
+        r#"change_id.short({}) ++ "\n"
+        ++ local_bookmarks.map(|b| b.name()).join("\x1e") ++ "\n"
+        ++ remote_bookmarks.filter(|b| b.tracked()).map(|b| b.name() ++ "\x1f" ++ b.tracking_ahead_count().lower() ++ "\x1f" ++ b.tracking_behind_count().lower()).join("\x1e") ++ "\n"
+        ++ commit_id.short({}) ++ "\n"
+        ++ conflict ++ "\n"
+        ++ divergent ++ "\n"
+        ++ hidden ++ "\n"
+        ++ immutable ++ "\n"
```
davidkna: Please consider using JSON output for more robust parsing.

bryceberger: Would JSON be more robust? Perhaps less code, if you can `#[derive(Deserialize)]` it, but RS is an illegal character for all the fields given, so will never be output, except from its use in the template.

src/modules/jujutsu_status.rs
RodEsp: Do you mean [slice.join()](https://doc.rust-lang.org/std/primitive.slice.html#method.join), or something else?

I'm not sure `slice.join()` ends up being any cleaner.

src/modules/jujutsu_status.rs
RodEsp: 🤦🏼‍♂️ good catch, thanks.

src/modules/jujutsu_status.rs
bryceberger: The template function in jj. See `jj help -k templates`

src/modules/jujutsu_status.rs
davidkna: It's robust against re-ordering and I fee like some of the items could also contain newlines.

src/modules/jujutsu_bookmark.rs
RodEsp: I'm a little divided about this, I feel like this is not the right way of thinking when working with `jj`. Branches work differently for `git`, you're always on a branch at any given point, but that's not how bookmarks work in `jj` and I'd like the prompt to reflect the current state of the `jj` repo.

But I do sympathize with working on `git` forges, like GitHub, and how you do always kind of want to show what "branch" you're on with `jj`.

Maybe someone could add a different module that does what you're saying and finds the closest bookmark for display, maybe called `jujutsu_branch`?
Then you could customize your config to use that
e.g.
```toml
[vcs]
# jujutsu_branch instead of jujutsu_bookmark at the end
jujutsu_modules="$jujutsu_state$jujutsu_change$jujutsu_commit$jujutsu_branch" 
```

src/modules/jujutsu_status.rs
RodEsp: Ok, I've switched this to output JSON that is then deserialized. Thanks for the suggestion!

src/modules/jujutsu_status.rs
RodEsp: I've moved the shared code to `src/modules/utils/jujutsu.rs` and got rid of the `jujutsu_status` module all together since it wasn't really necessary.

nicolaric:
let's gooooo! been waiting for this to come for so long now.

RodEsp:
FYI @nicolaric, not sure how soon this might land. I'm still trying to figure out how to do this without calling `jj log` for every individual `jujutsu_` module. Might have to do a significant refactor for that.

RodEsp:
Ok, this is starting to get to a spot that I like. I just incorporated `jj-lib`, would you mind taking a look @bryceberger?

After this I'm going to work on tests and docs, and then hopefully we should be good to go 🤞🏼 

I will move to "Ready for Review" and out of draft mode after tests & docs.

src/configs/vcs.rs
```diff
@@ -45,11 +49,12 @@ pub struct VcsConfig<'a> {
 impl Default for VcsConfig<'_> {
     fn default() -> Self {
         VcsConfig {
-            order: vec!["git", "hg", "pijul", "fossil"],
+            order: vec!["git", "hg", "pijul", "fossil", "jujutsu"],
```
bryceberger, line 52: `jujutsu` should have higher priority than `git` here. Due to colocation, it's likely that there will be both a `.jj` and a `.git`. Since the user is explicitly using `jj`, they would probably prefer to see information from `jj`.

bryceberger: Similarly, (not on this line of code but in general), at least one of the jujutsu modules should be enabled by default. As it is, if the user happens to get the jj codepath, no information is displayed because all the configs are disabled by default.

RodEsp: I was going to leave that stuff up to the maintainers of Starship, since it's easy enough for users to modify their own configs to reorder VCSs and enable whatever they want. But I get your point. I'll change it before I mark this ready for review and see what the Starship peeps say.

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,413 @@
...
+        && let Err(error) = config.load_file(ConfigSource::Repo, path)
+    {
+        log::debug!("Failed to read JJ repo config file: {error}");
+    }
+
+    UserSettings::from_config(config).ok()
+}
+
+fn repo_config_path(workspace_root: &Path) -> Option<PathBuf> {
+    let repo_config = workspace_root.join(".jj").join("repo").join("config.toml");
```
bryceberger, line 361: This is incorrect as of https://github.com/jj-vcs/jj/pull/8222

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,413 @@
...
+        .0;
+
+    Ok(Some((ahead, behind)))
+}
+
+fn commit_bookmarks(
+    repo: &ReadonlyRepo,
+    commit_id: &CommitId,
+    tracked_bookmarks: &HashMap<String, (usize, usize)>,
+) -> (Vec<BookmarkInfo>, bool) {
```
bryceberger, line 242: Why is this multiple bookmarks, but the conflicted info only corresponds to one of them?

RodEsp: Because I screwed that up and haven't written any tests 😅

Good catch, thanks!

nicolaric:
any updates on that?

RodEsp:
No, and I don't know when I'll be able to work on it again, but feel free to pick it up yourself.

PR #7234, opened 2026-01-20
feat: add statusline subcommand for Claude Code integration

davidkna:
#### Description

This PR implements a `starship statusline claude-code` subcommand for use in the [claude code statusline](https://code.claude.com/docs/en/statusline). `starship statusline` could be extended in the future to support tmux or other statuslines. See `docs/advanced-config/README.md` for setup docs.

Context data is read from stdin and stored in `Context` (with `Box<>` to reduce impact on regular prompt rendering), the following modules were added (not added to default `PROMPT_ORDER` to avoid them being used in main prompt, but shouldn't break anything if someone included them):
- `claude_model` - Displays the current Claude model name with support for custom aliases    
- `claude_context` - Shows context window usage as a percentage with a visual gauge and threshold-based styling
- `claude_cost` - Displays session cost  with threshold-based styling and optional statistics (duration, lines changed)

The main status string for the statusline defaults to `$claude_model$git_branch$claude_context$claude_cost`, which is read from `profiles.claude_code` by default.

🤖 co-authored via claude code

#### Motivation and Context

Closes #
[context](https://discord.com/channels/567163873606500352/804807190539665488/1454886345355956234)

#### Checklist:

### DIFF
--- a/src/modules/claude_context.rs
+++ b/src/modules/claude_context.rs
@@ -15,10 +15,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let claude_data = context.claude_code_data.as_ref()?;
 
     let total_tokens = claude_data.context_window.context_window_size;
-    let percentage_float = claude_data
-        .context_window
-        .used_percentage
-        .clamp(0.0, 100.0);
+    let percentage_float = claude_data.context_window.used_percentage.clamp(0.0, 100.0);
     let percentage = percentage_float.round() as u8;
 
     // Determine style based on percentage
--- a/src/print.rs
+++ b/src/print.rs
@@ -82,7 +82,8 @@ pub fn prompt_with_claude_code(args: Properties, target: Target) {
         .inspect_err(|e| log::error!("Failed to read Claude Code JSON from stdin: {e}"))
         .unwrap_or_default();
 
-    let context = Context::new(args, target).with_claude_code_data(claude_data);
+    let mut context = Context::new(args, target).with_claude_code_data(claude_data);
+    context.shell = Shell::Unknown;
     let stdout = io::stdout();
     let mut handle = stdout.lock();
 


Congee:
Can you please increase test coverage to pass the CI check? This PR looks promising

NickSdot:
Would love to see this! 🚀

src/print.rs
```diff
@@ -73,6 +73,19 @@ pub fn prompt(args: Properties, target: Target) {
...
+
+    write!(handle, "{}", get_prompt(&context)).unwrap();
+}
+
+pub fn prompt_with_claude_code(args: Properties, target: Target) {
+    let claude_data = serde_json::from_reader(io::stdin())
+        .inspect_err(|e| log::error!("Failed to read Claude Code JSON from stdin: {e}"))
+        .unwrap_or_default();
+
+    let context = Context::new(args, target).with_claude_code_data(claude_data);
```
cappyzawa: When running under zsh (or bash), `starship init zsh` exports `STARSHIP_SHELL=zsh`, which Claude Code inherits as a child process. This causes `wrap_colorseq_for_shell` to wrap ANSI color codes with `%{...%}` (zsh) or `\[...\]` (bash).

Since the statusline outputs directly to the terminal — not through a shell prompt — these escapes are unnecessary and break the output. Fish is unaffected because it falls through to `_ => return ansi` in `wrap_seq_for_shell`.

Setting `context.shell = Shell::Unknown` here would fix this:

```rust
let mut context = Context::new(args, target).with_claude_code_data(claude_data);
context.shell = Shell::Unknown;
```

src/modules/claude_context.rs
```diff
@@ -0,0 +1,429 @@
...
+    let current_tokens = claude_data.context_window.total_input_tokens
+        + claude_data.context_window.total_output_tokens;
+    let total_tokens = claude_data.context_window.context_window_size;
+
+    let percentage_float = if total_tokens > 0 {
+        (current_tokens as f32 / total_tokens as f32) * 100.0
+    } else {
+        0.0
+    };
+    let percentage = percentage_float.round().clamp(0.0, 100.0) as u8;
```
cappyzawa: Claude Code's JSON includes `context_window.used_percentage` with the correct value, but the `ContextWindow` struct doesn't have this field so it's silently ignored due to `#[serde(default)]`.

This self-calculation using `total_input_tokens + total_output_tokens` double-counts cache-hit tokens. In practice I observed 109k + 117k = 226k > 200k (context_window_size), resulting in 113% clamped to 100% — while `used_percentage` reported the actual usage as 42%.

Adding `used_percentage: f64` to `ContextWindow` and using it directly here would give the correct value.

o6uoq:
All checks have passed, is this PR ready to merge now ? 🤞🏼

davidkna:
Thanks for the review @cappyzawa!

o6uoq:
What version will this be released in ?

PR #7236, opened 2026-01-21
docs: add missing "style" option on "Enviroment Variable"

### LINKED ISSUE #7209: docs: Environment Variable module is missing the `style` option
#### Current Behavior

The docs for the [Environment Variable module](https://starship.rs/config/#environment-variable) is missing the `style` option.

lbellomo:
#### Description

I added the `style` option on the table for the "Enviroment Variable" module. Delete the value `example` for the "Variables" table because the other modules don't have a value there either. And I'm not sure if I have to put it in any particular order in the “style” table and if you prefer " or ' or nothing for this kind of example (like: `"black bold dimmed"` vs `'black bold dimmed'` vs `black bold dimmed`).

#### Motivation and Context

Closes #7209

I don't test this change. I just check that is the correct color (from [`src/configs/env_var.rs`](https://github.com/starship/starship/blob/master/src/configs/env_var.rs#L26))

#### Checklist:


davidkna (approved):

davidkna:
Thanks for the fix @lbellomo

PR #7241, opened 2026-01-24
feat(lean): Introduce the Lean module

eneoli:
#### Description

Introduces the `lean` module to display the current version of the [Lean Language and Theorem Prover](https://lean-lang.org).

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -957,6 +957,28 @@
         "contexts": []
       }
     },
+    "lean": {
+      "$ref": "#/$defs/LeanConfig",
+      "default": {
+        "format": "via [$symbol($version )]($style)",
+        "version_format": "v${raw}",
+        "symbol": " L∃∀N ",
+        "style": "blue",
+        "disabled": false,
+        "detect_extensions": [
+          "lean"
+        ],
+        "detect_files": [
+          "lake-manifest.json",
+          "lakefile.toml",
+          "lakefile.lean",
+          "lean-toolchain"
+        ],
+        "detect_folders": [
+          ".lake"
+        ]
+      }
+    },
     "line_break": {
       "$ref": "#/$defs/LineBreakConfig",
       "default": {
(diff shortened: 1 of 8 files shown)


.github/config-schema.json
```diff
@@ -957,6 +957,28 @@
         "contexts": []
       }
     },
+    "lean": {
+      "$ref": "#/$defs/LeanConfig",
+      "default": {
+        "format": "via [$symbol($version )]($style)",
+        "version_format": "v${raw}",
+        "symbol": " L∃∀N ",
```
davidkna, line 965: By default I would prefer a short symbol, most other symbols are single-letter or emoji.

src/modules/lean.rs
```diff
@@ -0,0 +1,149 @@
...
+fn get_toolchain(context: &Context) -> Option<String> {
+    if !has_lean_toolchain_file(context) {
+        return None;
+    }
+
+    let file_contents = context.read_file_from_pwd("lean-toolchain")?;
+    let version = file_contents.split_once(":");
+    match version {
+        Some(("leanprover/lean4", version)) => Some(version[1..].trim().to_string()),
+        Some((_, version)) => Some(format!("custom: {}", version.trim().to_string())),
```
davidkna, line 71: Please fix the clippy warnings and add a unit test for this case.

Does `v` also need to be trimmed in this case?

src/modules/lean.rs
```diff
@@ -0,0 +1,149 @@
...
+
+fn get_toolchain(context: &Context) -> Option<String> {
+    if !has_lean_toolchain_file(context) {
+        return None;
+    }
+
+    let file_contents = context.read_file_from_pwd("lean-toolchain")?;
+    let version = file_contents.split_once(":");
+    match version {
+        Some(("leanprover/lean4", version)) => Some(version[1..].trim().to_string()),
```
davidkna, line 70: I would prefer `strip_prefix("v")` or at least `.get(1..)` here.

davidkna (changes requested):

davidkna:
Closing as stale.

PR #7275, opened 2026-02-14
feat(fish): permit overriding job count/status in Fish

keysmashes:
#### Description

If `$STARSHIP_CMD_PIPESTATUS`, `$STARSHIP_CMD_STATUS`, and/or `$STARSHIP_JOBS` are set when Starship's `fish_prompt` is invoked, use their values directly rather than actually checking the exit status or number of jobs. (To prevent this breaking when they aren't set by something else, we now also unset those variables at the end of each `fish_prompt`.)

#### Motivation and Context

When using plugins like [fish-async-prompt][1], the automatic detection of exit status and running jobs can be flaky. To facilitate workarounds, let preexisting values for the job count and exit status variables take precedence over the values determined by Starship.

For example, with this commit (and `$async_prompt_inherit_variables` set to `all`), adding the following snippet to `config.fish` fixes my own problems with fish-async-prompt and Starship:

```fish
function __update_starship_job_count --on-event fish_postexec
  set -g STARSHIP_CMD_PIPESTATUS $pipestatus
  set -g STARSHIP_CMD_STATUS $status
  set -g STARSHIP_JOBS (jobs -g 2>/dev/null | count)
end
```

[1]: https://github.com/acomagu/fish-async-prompt

With the `__update_starship_job_count` function, this fixes my problems with job count/exit status not being detected; without that function, the behaviour doesn't change, as far as I can tell.

#### Checklist:

I'm not sure where this should be documented, it's a pretty niche problem. I also can't see any existing tests for the shell-specific bits of Starship?

### DIFF
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -18,12 +18,14 @@ function fish_prompt
             set STARSHIP_KEYMAP insert
     end
 
-    set STARSHIP_CMD_PIPESTATUS $pipestatus
-    set STARSHIP_CMD_STATUS $status
+    set -l _pipestatus $pipestatus
+    set -l _status $status
+    set -q STARSHIP_CMD_PIPESTATUS; or set STARSHIP_CMD_PIPESTATUS $_pipestatus
+    set -q STARSHIP_CMD_STATUS; or set STARSHIP_CMD_STATUS $_status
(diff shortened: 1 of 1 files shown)


keysmashes:
Closing in favour of #7276.

PR #7276, opened 2026-02-14
fix(fish): correctly fetch exit status/job count with fish-async-prompt

keysmashes:
#### Description

Rather than populating `$STARSHIP_CMD_PIPESTATUS` etc. in `fish_prompt` and `fish_right_prompt`, we now set them in a `fish_postexec` event handler. This avoids some code duplication and also improves Starship's behaviour in situations where `fish_prompt` is executed asynchronously in the background.

(This is an alternative to #7275, but doesn't require any explicit configuration by the user besides making sure the `$STARSHIP_*` variables get passed to the environment that `fish_prompt` runs in.)

#### Motivation and Context

When using plugins like [fish-async-prompt][1], the automatic detection of exit status and running jobs can be flaky, because `fish_prompt` gets executed in a different process and fish-async-prompt isn't always able to exactly emulate the environment it would normally run in. To improve the situation, let's retrieve the exit status and job count as soon as possible after a command exits, since this will always happen in the Fish process no matter what.

[1]: https://github.com/acomagu/fish-async-prompt

This fixes my problems with job count/exit status not being detected!

#### Checklist:

I can't see any existing tests for the shell-specific bits of Starship, so I'm not sure how to write a test for this change.

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -377,3 +377,11 @@ Not every style string will be displayed correctly by every terminal. In particu
 - Many terminals disable support for `blink` by default.
 - `hidden` is [not supported on iTerm](https://gitlab.com/gnachman/iterm2/-/issues/4564).
 - `strikethrough` is not supported by the default macOS Terminal.app.
+
+## Async prompt in Fish
+
+If using [fish-async-prompt][fish-async-prompt] to render the prompt asynchronously, you'll need to make sure some of Starship's internal variables are correctly propagated to the `fish_prompt` function.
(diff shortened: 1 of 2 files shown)


davidkna:
I am not very enthused with changing the init to this degree to support querying basic shell state when this plugin is in use.

Is the expectation that every prompt needs to integrate with this plugin for it to work correctly, or is something special about starship's behaviour?

keysmashes:
I think this is really fish-async-prompt's problem, not starship's, and probably the same thing happens with every prompt. In the past it's worked fine for me, but at some point (either due to an update of the plugin or an update to Fish itself) it stopped being able to propagate the exit status, and despite a lot of effort I couldn't work out why, but I _could_ figure out how to modify starship to work around the problem.

There's some related discussion at https://github.com/fish-shell/fish-shell/issues/7830 about introducing new `$interactive_{,pipe}status` variables which I *think* would also be a fine solution, but given that has been stagnating for the last half a decade, I'm not expecting any movement there...

The only downside I can see to this PR is that, if the prompt is redrawn after a job exits (e.g. by a `repaint` keybinding) without executing a new command, the job count will no longer be updated immediately: it'll only change after another command is executed.

PR #7262, opened 2026-02-08
feat(env_var): include $symbol in default format

MiguelRegueiro:
#### Description
This PR adds the missing `$symbol` variable to the default formatting string of the `env_var` module.

#### Motivation and Context
Currently, the `env_var` module does not include a symbol by default, which is inconsistent with almost every other module in Starship. This change ensures a uniform visual experience for users who want to display environment variables with icons.

Closes #7208

#### How Has This Been Tested?

I have added a new unit test in `src/modules/env_var.rs` titled `default_format_includes_symbol` which verifies that a custom symbol is correctly rendered when using the default format.
- [ ] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #7298, opened 2026-02-28
fix(docs): Re-add `style` option to `env_var` module

injust:
Re-land #7236, which was partially undone in #7262

cc @MiguelRegueiro


MiguelRegueiro:
Thanks for catching this and restoring it. I checked the diff and it looks good to me

PR #7300, opened 2026-03-01
feat: add support to overide all the configuration in kubernetes.contexts

0x416d616e:
#### Description

Extends `[[kubernetes.contexts]]` entries to support all configuration properties from the parent `[kubernetes]` module, not just `style`, `symbol`, `context_alias`, and `user_alias`.

New per-context overridable fields:
- `format` -- override the format string for a matched context
- `disabled` -- hide the module entirely for a matched context
- `detect_extensions` -- override or clear parent's extension detection
- `detect_files` -- override or clear parent's file detection
- `detect_folders` -- override or clear parent's folder detection
- `detect_env_vars` -- override or clear parent's env var detection

#### Motivation and Context

Previously, `[[kubernetes.contexts]]` only allowed overriding `style`, `symbol`, `context_alias`, and `user_alias`. Properties like `format`, `disabled`, and detection rules (`detect_files`, `detect_folders`, `detect_extensions`, `detect_env_vars`) were always inherited from the parent `[kubernetes]` block and could not vary per context.

This made it impossible to:
- Use a different format string for production vs dev contexts
- Hide the module for local/minikube contexts
- Require specific files to be present only for certain contexts
- Show the module unconditionally for some contexts while keeping file detection for others

Closes #

5 new tests added following existing `test_config_context_*` patterns:
- `test_config_context_overwrites_format` -- verifies per-context format override
- `test_config_context_disabled` -- verifies per-context disable returns `None`
- `test_config_context_detect_files_override` -- verifies per-context file detection (shows when file present, hidden when absent)
- `test_config_context_detect_folders_clear` -- verifies `detect_folders = []` clears parent's detection (module shows without the folder)
- `test_config_context_inherits_parent_detect` -- verifies omitted `detect_files` inherits parent's detection

All 38 kubernetes tests pass (33 existing + 5 new).

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -4467,6 +4467,13 @@
           ],
           "default": null
         },
+        "format": {
+          "type": [
+            "string",
+            "null"
+          ],
+          "default": null
+        },
         "context_alias": {
           "type": [
             "string",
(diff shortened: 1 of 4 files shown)


0x416d616e:
Bump

PR #7302, opened 2026-03-03
Disambiguate winget package

skalahonza:
#### Motivation and Context

When installing starhip via winget. Winget cli notified me that there are two packages with the same name.
```
# winget install starship
Multiple packages found matching input criteria. Please refine the input.
Name                Id                Source
---------------------------------------------
Starship! Starship! 9N7JDF8Q4QLC      msstore
starship            Starship.Starship winget
```

This is documentation only.

#### Checklist:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -70,7 +70,7 @@ onMounted(() => {
    With [Winget](https://github.com/microsoft/winget-cli):
 
    ```powershell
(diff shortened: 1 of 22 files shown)


skalahonza:
I know that editing translation files directly is not ideal. But here I only modified the link which is not translated.

PR #7310, opened 2026-03-09
feat: Add module for godot projects

BenBergman:
#### Description

This PR adds a module for Godot projects similar to what exists for other languages. It grabs the version of godot currently installed for use in the printout (stripping the tail end of it as it broke version parsing - see the included test for behaviour).

#### Motivation and Context

I work in a lot of languages and it has been great to have the language of my projects detected. I started using Godot recently and missed having that.

image

I have run all unit tests and have the same number of failures as exist already on main. I added a unit test with my module as well.

I have manually tested my module with `starship module godot` in both Godot and non-Godot directories with both Linux and macOS (after adding the Godot.app binary to my path). I have confirmed with a friend that the version output on Windows is the same as macOS.

#### Checklist:

- [x] I have updated the tests accordingly. (to the best of my knowledge)


BenBergman:
I've fixed all the formatting issues in my PR, but I'm not sure what to do about the coverage test. I think it is complaining about the mock_cmd entry not being tested, but I can't seem to find where all the other entries in that list are being tested in order to update it...?

BenBergman:
> I've fixed all the formatting issues in my PR, but I'm not sure what to do about the coverage test. I think it is complaining about the mock_cmd entry not being tested, but I can't seem to find where all the other entries in that list are being tested in order to update it...?

Never mind, I figured it out. I added a few more tests of the full module with different triggers to make sure it is always picked up (I even caught an error in my default config!) and now it is passing. Anything else I should change?

src/modules/godot.rs
```diff
@@ -0,0 +1,193 @@
...
+fn get_godot_version(command: &str, context: &Context) -> Option<String> {
+    context
+        .exec_cmd(command, &["--version"])?
+        .stdout
+        .split_whitespace()
+        .next()
+        .map(str::to_owned)
+}
+
+fn format_godot_version(version: &str, version_format: &str) -> Option<String> {
```
davidkna, line 74: Please consider using the regular `VersionFormatter`

src/modules/godot.rs
```diff
@@ -0,0 +1,193 @@
...
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "version" => {
+                    format_godot_version(raw_godot_version.as_str(), &config.version_format).map(Ok)
```
davidkna: Please ensure that clippy warnings are addressed.

davidkna (changes requested):
Thank you for the PR. This is not a full review yet, but I found two issues that should be addressed.

src/modules/godot.rs
BenBergman: I am using the regular `VersionFormatter` within this function. The raw version number produced by godot has a dot and text following the major.minor.patch numbers which breaks `VersionFormatter`, so I had to wrap it to remove that component. Is there a different way I should have fixed that?

src/modules/godot.rs
BenBergman: Change pushed. Thanks for catching this one!

src/modules/godot.rs
```diff
@@ -0,0 +1,192 @@
...
+        .set_files(&config.detect_files)
+        .set_extensions(&config.detect_extensions)
+        .set_folders(&config.detect_folders)
+        .is_match();
+
+    if !is_gd_project {
+        return None;
+    }
+
+    let raw_godot_version = get_godot_version(config.godot_command, context)?;
```
davidkna:
```suggestion
    let raw_godot_version = LazyLock::new(|| get_godot_version(config.godot_command, context));
```

src/modules/godot.rs
davidkna: For the `Regex` I am a bit concerned about the current error handling approach and the fairly strict regex forcing a 3-number version. (I assume it's still parsable as `Complex`/`Mess`, so adding support for that into the formatter may work. Reformatting into a `General`/`Version`-style version string with just the version number and hash might be a good idea).

In general, for a prompt the full version feels too verbose by default so I would also prefer defaulting to `$numver` or a different `version_format`.

src/configs/godot.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub detect_folders: Vec<&'a str>,
+    pub godot_command: &'a str,
+}
+
+impl Default for GodotConfig<'_> {
+    fn default() -> Self {
+        Self {
+            format: "via [$symbol($version )]($style)",
+            version_format: "v${raw}",
+            symbol: " ",
```
davidkna: Please don't default to nerd font icons by default, please replace it with an appropriate emoji.

davidkna (changes requested):

src/modules/godot.rs
BenBergman: After messing around with it a bunch, I can get it to pass through the unmodified `VersionFormatter` after removing underscores and separating the `numver` portion from the rest with a `-` instead of a `.`, for example `4.6.1-stable.arch.linux.14d19694e`. I don't love that as my modifiers likely won't be sufficiently robust for all future formats, so I also modified the `VersionFormatter` to accept a `Complex`/`Mess`, but I'm not sure if that will have knock on effects (all tests pass on my machine though).

I think I prefer the updated `VersionFormatter`, so I'm pushing that, but let me know if you'd prefer I go back to altering the string before passing it along.

For posterity, this is what I ended up with that worked when there were no underscores:

```
    let re = Regex::new(r"(^(\d|\.)+)\.([a-zA-Z].+)").unwrap();
    let ver = match re.captures(version) {
        Some(caps) => caps[1].to_string() + "." + &caps[3].to_string(),
        _ => version.to_string(),
    };
```

src/modules/godot.rs
```diff
@@ -0,0 +1,208 @@
...
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "version" => format_godot_version(
+                    raw_godot_version.as_ref().unwrap().as_str(),
```
davidkna, line 38: Please don't unwrap here, instead handle errors by defaulting to `None`

PR #7312, opened 2026-03-09
fix(hostname): detect SSH sessions using SSH_CLIENT and SSH_TTY

### LINKED ISSUE #7207: Hostname module only checks $SSH_CONNECTION (different from Username module)
#### Current Behavior

The Username module checks $SSH_CONNECTION, $SSH_CLIENT, and $SSH_TTY to determine if there is an SSH connection.

However, the Hostname module only checks $SSH_CONNECTION. This is not documented and is not user-configurable.

#### Expected Behavior

Both modules (Hostname and Username) should use the same criteria to check for an SSH connection, and it should be documented at https://starship.rs/config/#hostname.

#### Environment

- Starship version: 1.24.2

Xavrir:
## Summary

The hostname module only checked `SSH_CONNECTION` to detect SSH sessions, while the username module checks `SSH_CONNECTION`, `SSH_CLIENT`, and `SSH_TTY`. This inconsistency caused hostname to not display in valid SSH session scenarios where only `SSH_CLIENT` or `SSH_TTY` were set.

## Changes

- Added `is_ssh_session()` helper function that checks all three SSH environment variables
- Updated hostname module to use the same SSH detection criteria as username module
- Added tests for `SSH_CLIENT` and `SSH_TTY` detection
- Updated hostname config documentation to reflect expanded SSH detection

## Testing

All existing tests pass, including new tests for `SSH_CLIENT` and `SSH_TTY` scenarios. Full test suite verified with `cargo test`.

Fixes #7207

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2468,11 +2468,16 @@ format = 'via [⎈ $version](bold white) '
 
 The `hostname` module shows the system hostname.
 
+> [!TIP]
+> SSH connection is detected by checking environment variables
+> `SSH_CONNECTION`, `SSH_CLIENT`, and `SSH_TTY`. If your SSH host does not set up
+> these variables, one workaround is to set one of them with a dummy value.
+
 ### Options
 
 | Option            | Default                                | Description                                                                                                                           |
 | ----------------- | -------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
-| `ssh_only`        | `true`                                 | Only show hostname when connected to an SSH session (`SSH_CONNECTION`, `SSH_CLIENT`, or `SSH_TTY`).                                   |
+| `ssh_only`        | `true`                                 | Only show hostname when connected to an SSH session.                                                                                  |
 | `ssh_symbol`      | `'🌐 '`                                | A format string representing the symbol when connected to SSH session.                                                                |
 | `trim_at`         | `'.'`                                  | String that the hostname is cut off at, after the first match. `'.'` will stop after the first dot. `''` will disable any truncation. |
 | `detect_env_vars` | `[]`                                   | Which environment variable(s) should trigger this module.                                                                             |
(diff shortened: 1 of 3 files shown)


src/modules/hostname.rs
```diff
@@ -74,6 +66,12 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     Some(module)
 }
 
+fn is_ssh_session(context: &Context) -> bool {
+    ["SSH_CONNECTION", "SSH_CLIENT", "SSH_TTY"]
+        .iter()
+        .any(|env| context.get_env_os(env).is_some())
+}
+
```
davidkna: Please export `is_ssh_session` from `username` instead and use it directly, or move it to `utils/`

docs/config/README.md
```diff
@@ -2465,7 +2465,7 @@ The `hostname` module shows the system hostname.
 
 | Option            | Default                                | Description                                                                                                                           |
 | ----------------- | -------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
-| `ssh_only`        | `true`                                 | Only show hostname when connected to an SSH session.                                                                                  |
+| `ssh_only`        | `true`                                 | Only show hostname when connected to an SSH session (`SSH_CONNECTION`, `SSH_CLIENT`, or `SSH_TTY`).                                   |
```
davidkna: As not to overload the table, I would prefer putting it into a `> [!TIP]` like with the username module.

Xavrir:
Done. I made `is_ssh_session` `pub(crate)` in the username module and hostname now imports it directly, so the duplicate helper is gone. I also moved the SSH detection note out of the options table into a `> [!TIP]` block that matches the username docs. Tests pass.

StantonMatt (commented):
I checked this locally at `832ec9d` and the shared `is_ssh_session` path covers both the `ssh_only` gate and the `$ssh_symbol` rendering, so this matches the username module behavior rather than only changing one side.

Local checks passed here:

- `cargo test modules::hostname::tests --lib` (14 passed)
- `cargo test modules::username::tests --lib` (13 passed, 2 ignored)
- `cargo fmt --check`

I also rechecked the current PR checks, and every visible job is green.

davidkna:
Closing in favour of https://github.com/starship/starship/pull/7611

PR #7319, opened 2026-03-13
feat: add Vault module with token expiration support

### LINKED ISSUE #2195: hashicorp vault support (show token expiration time)

I forget to renew by vault token before it expires.

Show the token expiration time. Maybe only if it's less than a config value (ex: < 4 days).

foxfromworld:
#### Description

Adds a new vault module to Starship. This module displays HashiCorp Vault context and shows token expiration when the token is close to expiring.

#### Motivation and Context

Many developers use Vault for secrets management. Having token expiration visible in the prompt helps prevent unexpected authentication failures and improves workflow awareness.

Closes #2195

image

#### Checklist:


docs/config/README.md
```diff
@@ -4948,6 +4949,44 @@ By default the module will be shown if any of the following conditions are met:
...
+## Vault
+
+The `vault` module shows your current [HashiCorp Vault](https://www.hashicorp.com/en/products/vault) context in the prompt.\
+The module only displays token expiration (in days) when the token is close to expiring.
+
+### Options
+
+| Option             | Default                                           | Description                                                           |
+| ------------------ | ------------------------------------------------- | --------------------------------------------------------------------- |
+| `symbol`           | `'⚠️'`                                             | The symbol shown when displaying Vault token expiration.              |
```
davidkna: I would prefer a default symbol that is more vault-related.

docs/config/README.md
```diff
@@ -4948,6 +4949,44 @@ By default the module will be shown if any of the following conditions are met:
...
+| `style`            | `'bold red'`                                      | The style for the module output.                                      |
+| `format`           | `'[$symbol Token expires: $expire_time]($style)'` | The format string used to display the expiration time.                |
```
davidkna: The default format is a bit too verbose.

src/configs/vault.rs
```diff
@@ -0,0 +1,29 @@
...
+
+#[derive(Clone, Deserialize, Serialize)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+#[serde(default)]
+
+pub struct VaultConfig<'a> {
```
davidkna, line 10: Please fix any clippy warnings.

src/modules/vault.rs
davidkna: Please use `context.exec_cmd`

davidkna: Please simplify this and the following code, and reduce nesting, e.g.:
```rust
let Some(expire_time) = json["data"]["expire_time"].as_str() else {
    return None
}
```

src/modules/vault.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    if expire_time != "null" {
+        if let Ok(expire_dt) = DateTime::parse_from_rfc3339(expire_time) {
+            let now = Local::now();
+            let expire_local = expire_dt.with_timezone(&Local);
+            let duration = expire_local.signed_duration_since(now);
+
+            if duration.num_days() <= config.show_within_days as i64 {
+                let dt: DateTime<FixedOffset> = expire_time.parse().ok()?;
+                let formatted = dt.format("%Y-%m-%d").to_string();
```
davidkna: Please simplify e.g. using the `let Ok() else`-pattern. In addition, the format as used here is not consistent with existing expiration time handling, as in the AWS module, please add an option or expose both in separate variables.

In addition, please handle any per-variable logic as part of the variable mappers.

src/modules/vault.rs
```diff
@@ -0,0 +1,333 @@
...
+
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("vault");
+    let config: VaultConfig = VaultConfig::try_load(module.config);
+
+    if config.disabled {
+        return None;
+    }
+
+    // Run vault CLI to get token info in JSON format
```
davidkna: Please consider checking for `VAULT_TOKEN || ~/.vault-token` before running `vault` commands for better performance.

src/configs/starship_root.rs
```diff
@@ -103,6 +103,7 @@ pub const PROMPT_ORDER: &[&str] = &[
     "typst",
     "vlang",
     "vagrant",
+    "vault",
```
davidkna: Please put the module along other cloud environment modules like `aws`, and also add it to the default format in the `docs/config/README.md`.

davidkna (changes requested):

src/modules/vault.rs
```diff
@@ -0,0 +1,333 @@
...
+                        .map_meta(|variable, _| match variable {
+                            "symbol" => Some(config.symbol),
+                            _ => None,
+                        })
+                        .map_style(|variable| match variable {
+                            "style" => Some(Ok(config.style)),
+                            _ => None,
+                        })
+                        .map(|variable| match variable {
+                            "expire_time" => Some(Ok(&formatted)),
```
davidkna: Please also consider exposing more information such as the namespace

src/modules/vault.rs
```diff
@@ -0,0 +1,333 @@
...
+        let expire_dt = DateTime::parse_from_rfc3339(&expire_time)
+            .unwrap()
+            .with_timezone(&Local);
+        let now = Local::now();
+        let duration = expire_dt.signed_duration_since(now);
+
+        // 驗證超過 show_within_days
+        assert!(duration.num_days() > VaultConfig::default().show_within_days as i64);
+
+        // 模擬 module 邏輯：因為超過 show_within_days，應該 return None
```
davidkna: Please keep any code comments English.

src/modules/vault.rs
```diff
@@ -0,0 +1,333 @@
...
+                });
+                return Some(module);
+            }
+        }
+    }
+    None
+}
+
+#[cfg(test)]
+mod tests {
```
davidkna: Please add more meaningful tests, including using `ModuleRenderer`.

docs/config/README.md
foxfromworld: Thanks for pointing that out! 🔒 makes more sense, I’ll push the change in the upcoming update.

docs/config/README.md
foxfromworld: Appreciate the feedback — I see your point that the default format is a bit verbose compared to what’s really needed. I’ll simplify it in the next update so it’s cleaner but still informative.

src/configs/vault.rs
foxfromworld: Thanks for catching that — I’ve fixed the Clippy warnings.

src/modules/vault.rs
foxfromworld: Thanks for the suggestion — I’ll switch to using context.exec_cmd in the next commit.

foxfromworld: Thanks for the suggestion. I first tried let Some(expire_time) = … else { return None; }, but Clippy warned to use the ? operator instead. So I updated it to: let expire_time = json["data"]["expire_time"].as_str()?;

src/modules/vault.rs
foxfromworld: Thanks for the reminder — I’ll make sure to keep all code comments in English in the next commit

src/modules/vault.rs
foxfromworld: Thanks for the review!
I’ll simplify parsing with let Ok(...) else, expose both expire_time and expire_date for consistency, and move all per‑variable logic into the mappers. These changes will be included in the next commit.

src/modules/vault.rs
foxfromworld: Thanks for the suggestion!
I’ll add a check for VAULT_TOKEN and ~/.vault-token before running vault commands to improve performance. This will also be included in the next commit.

src/configs/starship_root.rs
foxfromworld: Thanks for the suggestion!
I’ll move the Vault module alongside other cloud environment modules (like AWS) and add it to the default format in docs/config/README.md in the next commit.

src/modules/vault.rs
foxfromworld: Thanks for the suggestion! In my current environment, the output from vault token lookup --format=json does not include a namespace field, so there isn’t any value to expose. I’ll keep the implementation as-is for now, and if namespaces become available in the future, support can be added then.

src/modules/vault.rs
foxfromworld: Thanks for the feedback! I’ve noted your suggestion and will add more meaningful tests using ModuleRenderer in my next commit.

foxfromworld:
feat(vault): add Vault module with tests and improvements

- Set default symbol to 🔒 for a more vault-related icon
- Simplify default format to reduce verbosity
- Fix clippy warnings (cargo clippy passes cleanly)
- Switch to using context.exec_cmd for command execution
- Refactor expire_time parsing with proper Option handling
- Expose two formats of expire_time via variable mappers
- Add checks for VAULT_TOKEN and ~/.vault-token before running vault commands
- Move Vault module alongside other cloud environment modules (e.g. AWS)
- Update docs/config/README.md to include Vault in default format
- Namespace exposure deferred for now
- Rewrite tests to use ModuleRenderer for consistency and meaningful coverage

foxfromworld:
Hi @davidkna, 
I’ve rebased and resolved conflicts. Please let me know if there’s anything else I should adjust. Thank you.

PR #7328, opened 2026-03-19
chore: automate copyright year updates with Renovate

lumilla:
#### Description

- Updates copyright year references to the current year
- (ab)uses a Renovate custom datasource to, ideally, never need to do that again
- to do that, adds regex and miscellaneous changes to the Renovate configuration
- semantic commit type chore for current-year updates.

#### Motivation and Context
- Removes manual copyright updates, which were done previously
- Should keep legal text in sync with minimal maintenance.

Ran` npx renovate` locally. No Rust code has been touched, so it should not be affected.

#### Checklist:

Checklist not filled, as I do not think it is applicable. The Renovate config is commented roughly on what it does.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -59,8 +59,22 @@
             "groupName": "npm",
             "matchDatasources": ["npm"],
             "extends": ["schedule:weekly"]
+        },
+        {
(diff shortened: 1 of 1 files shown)


davidkna:
Closing as stale.

PR #7329, opened 2026-03-20
fix(hg_state): avoid false-positive MERGING state

### LINKED ISSUE #7322: hg_state always shows MERGING
#### Current Behavior
in any mercurial repository, the hg_state module always shows MERGING, even when that's not the case.

#### Expected Behavior
MERGING should not be indicated when it's not happening.

#### Possible Solution
the current condition is `else if is_merge_state(hg_root).is_ok()`, and should probably be something like `else if is_merge_state(hg_root).is_ok_and(|b| b == true)`

#### Environment
- Starship version: 1.24.2
- pwsh version: PowerShell 7.5.5
- Operating system: Windows 10.0.26200
- Git Commit Hash: 33f7077
- Branch/Tag: master
- Rust Version: rustc 1.92.0 (ded5c06cf 2025-12-08)
- Build Time: 2025-12-30 19:16:37 +00:00

```pwsh
<unknown config>
```

#### Starship Configuration

```toml
[hg_branch]
disabled = false

[hg_state]
disabled = false
```

mvanhorn:
## Summary

`is_merge_state()` returns `io::Result<bool>`, but line 83 checked `.is_ok()` instead of inspecting the inner `bool`. This means `Ok(false)` (a non-merge repo) passed the check, causing every hg repo with a readable `.hg/dirstate` to display the MERGING state.

**Fix:** Replace `.is_ok()` with `.is_ok_and(|is_merge| is_merge)` so only `Ok(true)` triggers the merge label.

`is_ok_and` is stable since Rust 1.70 - no new dependencies needed.

Fixes #7322

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo clippy` passes
- [ ] `cargo test` passes (existing `test_hg_merging` validates the positive case)
- [ ] Verify non-merge hg repos no longer show MERGING

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/src/modules/hg_state.rs
+++ b/src/modules/hg_state.rs
@@ -80,7 +80,7 @@ fn get_state_description<'a>(
         Some(StateDescription {
             label: config.histedit,
         })
-    } else if is_merge_state(hg_root).is_ok_and(|is_merge| is_merge) {
+    } else if is_merge_state(hg_root).unwrap_or(false) {
         Some(StateDescription {
             label: config.merge,
         })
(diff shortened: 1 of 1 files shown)


src/modules/hg_state.rs
```diff
@@ -80,7 +80,7 @@ fn get_state_description<'a>(
         Some(StateDescription {
             label: config.histedit,
         })
-    } else if is_merge_state(hg_root).is_ok() {
+    } else if is_merge_state(hg_root).is_ok_and(|is_merge| is_merge) {
```
davidkna:
```suggestion
    } else if is_merge_state(hg_root).unwrap_or(false) {
```

If possible also add a test.

mvanhorn: Applied the suggestion and added unit tests for both the non-merge case (p2 all zeros) and merge case (p2 non-zero) in eb9390c.

davidkna:
Thanks for the contribution @mvanhorn!

mvanhorn:
Thanks for merging! Happy to have contributed to starship.

mvanhorn:
Thanks for merging. The hg_state false positives were tricky since they only showed up with certain working directory states - good to have it resolved.

mvanhorn:
Thanks for the review and merge!

mvanhorn:
Thanks for picking this up, @davidkna.

PR #7337, opened 2026-03-22
ci: add riscv64 to release build matrix

### LINKED ISSUE #6547: Need Starhip for BeagleV Ahead (Ubuntu 24.04, RISC-V 64 bits)
Getting this error message trying to install Starship on a BeagleV Ahead running Ubuntu 24.04 LTS:

```
echo "Installing Starship"
export BIN_DIR=$HOME/.local/bin
curl -sS https://starship.rs/install.sh | sh

...

x riscv64 builds for unknown-linux-musl are not yet available for Starship

> please create an issue requesting a build for riscv64-unknown-linux-musl:
> https://github.com/starship/starship/issues/new/
```

gounthar:
#### Description

Adds `riscv64gc-unknown-linux-gnu` to the release build matrix. Uses native RISE riscv64 runners (`ubuntu-24.04-riscv`) instead of cross-compilation, so it builds with `cargo build` directly (same path as macOS/Windows targets).

#### Motivation and Context

There have been several requests for riscv64 builds (#6547, #4394, #5623). RISC-V hardware is becoming more common for development and edge use cases; the install script currently tells riscv64 users the build is "not yet available."

Native riscv64 GitHub Actions runners are provided by the [RISE Project](https://riseproject.dev/), free for open source. To enable: install the [RISE runners GitHub App](https://github.com/apps/rise-risc-v-runners). Without the app, the job stays queued (never blocks other builds).

Closes #6547

- [x] I have tested using **Linux** (native riscv64, BananaPi F3 + RISE runner)

Build validated on RISE runner: https://github.com/gounthar/starship/actions/runs/23410203251

```
$ starship --version
starship 1.24.2
```

#### Checklist:

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -19,7 +19,7 @@ SUPPORTED_TARGETS="x86_64-unknown-linux-gnu x86_64-unknown-linux-musl \
                   aarch64-apple-darwin x86_64-pc-windows-msvc \
                   i686-pc-windows-msvc aarch64-pc-windows-msvc \
                   x86_64-unknown-freebsd \
-                  riscv64gc-unknown-linux-gnu"
+                  riscv64gc-unknown-linux-musl"
 
 info() {
 	printf '%s\n' "${BOLD}${GREY}>${NO_COLOR} $*"


gounthar:
One thing to note: the release workflow only runs on tags, so the riscv64 build won't actually execute until a release is cut. Before merging, the RISE runners GitHub App would need to be installed on the starship org, otherwise the riscv64 job will stay queued with no runner available.

If you'd prefer to not block on that, I can add `continue-on-error: true` to the riscv64 matrix entry so it doesn't hold up releases if the runner isn't available.

gounthar:
@davidkna Makes sense. cross-rs does have riscv64gc support (both gnu and musl Dockerfiles). I'll rework to use cross on ubuntu-latest instead of native runners, and update the install script.

gounthar:
Reworked to use cross-rs on ubuntu-latest + updated the install script with riscv64 -> riscv64gc arch mapping. Ready for another look whenever you feel like it. 🙏

.github/workflows/release.yml
```diff
@@ -55,6 +55,10 @@ jobs:
             os: ubuntu-latest
             name: starship-arm-unknown-linux-musleabihf.tar.gz
 
+          - target: riscv64gc-unknown-linux-gnu
+            os: ubuntu-latest
+            name: starship-riscv64gc-unknown-linux-gnu.tar.gz
```
davidkna: I would prefer only publishing the musl variant

gounthar: Makes sense, will switch to riscv64gc-unknown-linux-musl to stay consistent with the other Linux targets.

davidkna:
Thanks for the contribution @gounthar!

gounthar:
Thanks for the review and merge, @davidkna!

PR #7340, opened 2026-03-23
docs(python): update python_binary tip to reflect argument support

### LINKED ISSUE #7279: Updated documentation and schema for `python_binary` arguments
#### Current Behavior

GH-6523 added support for `python_binary` arguments.
But it's not reflected in [documentation](https://starship.rs/config/#options-77:~:text=The%20python%5Fbinary,used) and [config-schema.json](https://starship.rs/config-schema.json).

The [TIP](https://starship.rs/config/#options-77:~:text=The%20python%5Fbinary,used) is even saying the opposite:

> [...] Starship will try executing each binary until it gets a result. Note you can only change the binary that Starship executes to get the version of Python not the arguments that are used.

#### Expected Behavior

Up to date documentation.

Image

#### Environment
- Starship version: 1.24.2

sedat4ras:
## Description

The `python_binary` option gained support for passing arguments to binaries in #6523 (e.g. `['uv', 'run', 'python', '--version']`), but the documentation tip was never updated and still says:

> Note you can only change the binary that Starship executes to get the version of Python not the arguments that are used.

This is now incorrect and misleading.

## Changes

- Updated the `[!TIP]` block to accurately describe the supported formats: string, list of strings, or list of lists of strings
- Updated the `python_binary` option description in the options table
- Added an example showing how to use `uv run python` to get the Python version

## Related

Fixes #7279

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -4017,7 +4017,7 @@ By default, the module will be shown if any of the following conditions are met:
 | `style`              | `'yellow bold'`                                                                                              | The style for the module.                                                             |
 | `pyenv_version_name` | `false`                                                                                                      | Use pyenv to get Python version                                                       |
 | `pyenv_prefix`       | `'pyenv'`                                                                                                    | Prefix before pyenv version display, only used if pyenv is used                       |
-| `python_binary`      | `['python', 'python3', 'python2']`                                                                           | Configures the python binaries that Starship should execute when getting the version. |
+| `python_binary`      | `['python', 'python3', 'python2']`                                                                           | Configures the python binaries (or binaries with arguments) that Starship should execute when getting the version. |
 | `detect_extensions`  | `['py', 'ipynb']`                                                                                            | Which extensions should trigger this module                                           |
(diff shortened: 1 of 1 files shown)


davidkna (changes requested):
I am fine with documenting this, but please don't add examples with `uv` since in that case it can lead to unintended code execution in malicious directories.

sedat4ras:
Thanks for the feedback\! The `uv` example has been removed — the tip now uses `['python3.11', '--version']` as the example instead. Let me know if anything else needs adjusting.

martimlobao:
@sedat4ras why was this pr closed? i can still reproduce the bug on the latest version.

I also think this needs adjusting as well: https://starship.rs/config-schema.json

sedat4ras:
Hey @martimlobao — closed it during a cleanup pass since it had 'changes requested' status and I wasn't sure the maintainers wanted to move forward with it. If the bug is still reproducible on the latest version, I'd be happy to reopen and update the fix. Let me know!

PR #7343, opened 2026-03-24
fix(explain): show jobs module by reading STARSHIPS_JOBS_COUNT env var

### LINKED ISSUE #3264: Missing "jobs" description in staship explain
#### Current Behavior
`starship explain` does not show the signification of the "+" sign added by the `jobs` module

```
phgiraud in mypc in ~ ❯ top &
phgiraud in mypc in ~ + ❯ starship explain

 "phgiraud in " (<1ms)      -  The active user's username
 "mypc in " (<1ms)       -  The system hostname
 "~ " (<1ms)                -  The current working directory
 "❯ " (<1ms)                -  A character (usually an arrow) beside where the text is entered in your terminal

```

#### Expected Behavior
A line should be explaining this symbol, something like
`"+ " (<1ms)                - The number of background running jobs`

Additionnaly, I could be great if the name of the module was explicitly stated (even though here its quite obvious), like ending the line with `from the module [jobs]`

#### Environment
- Starship version: 0.58.0
- bash version: GNU bash, version 5.1.4(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2020 Free Software Foundation, Inc.
Licence GPLv3+ : GNU GPL version 3 ou ultérieure <http://gnu.org/licenses/gpl.html>

Ceci est un logiciel libre ; vous être libre de le modifier et de le redistribuer.
AUCUNE GARANTIE n'est fournie, dans les limites permises par la loi.
- Operating system: Debian 11
- Git Commit Hash: 0a60d1e9
- Branch/Tag: v0.58.0
- Rust Version: rustc 1.55.0 (c8dfcfe04 2021-09-06)
- Rust channel:  release
- Build Time: 2021-09-21 13:43:49 +00:00

```bash
# for examples

case $- in
    *i*) ;;
      *) return;;
esac

HISTCONTROL=ignoreboth

shopt -s histappend

HISTSIZE=1000
HISTFILESIZE=2000

shopt -s checkwinsize

#shopt -s globstar

#[ -x /usr/bin/lesspipe ] && eval "$(SHELL=/bin/sh lesspipe)"

fi

case "$TERM" in
esac

#force_color_prompt=yes

	color_prompt=yes
    else
	color_prompt=
    fi
fi

else
fi

case "$TERM" in
xterm*|rxvt*)
    ;;
*)
    ;;
esac

    alias ls='ls --color=auto'
    #alias dir='dir --color=auto'

    #alias grep='grep --color=auto'
    #alias fgrep='fgrep --color=auto'
    #alias egrep='egrep --color=auto'
fi

# some more ls aliases
alias ll='ls -l'
alias la='ls -A'
alias l='ls -CF'

# Alias definitions.

    . ~/.bash_aliases
fi

# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
    . /etc/bash_completion
  fi
fi

# Add autocompletion for please + aliases
if [ -f ~/cs/tools/shell/bashrc ]; then
  source ~/cs/tools/shell/bashrc 
fi

[ -f ~/.fzf.bash ] && source ~/.fzf.bash

#Charge le prompt starship
eval "$(starship init bash)"
```

#### Starship Configuration

```toml

add_newline = false

# Don't break prompt in two lines
[line_break]
disabled = true

[username]
show_always = true

[hostname]
ssh_only = false

[directory]
truncation_length = 0
truncate_to_repo = false

[git_branch]
symbol = ""

[python]
#disabled = true

[git_status]
modified.enabled = true
untracked.enabled = true
stashed = "" # hide stashed

[rust]
disabled = true

[package]
disabled = true

[battery]
full_symbol = "🔋 "
charging_symbol = "⚡️ "
discharging_symbol = "💀 "

[[battery.display]]
threshold = 10
style = "regular"

[jobs]
symbol_threshold = 1
symbol = "+"

[status]
disabled = true

[time]
disabled = true



```

maxmynter:
This PR introduces a fix to the `jobs` module in `starship explain` when there are active background jobs. 
This was previously not happening because `explain` creates a fresh context with `--jobs` defaulting to `0`, causing the jobs module to return `None` and be filtered out of the output.

#### Description
Two changes:

1. **`src/init/starship.zsh`** (and `bash`, `fish`): export `STARSHIP_JOBS_COUNT` as an environment variable so it is visible to subprocesses. Previously it was only a shell-local variable used to build the prompt string.

2. **`src/print.rs`**: in `explain`, if `--jobs` was not explicitly passed (i.e. it is `0`), fall back to reading `STARSHIP_JOBS_COUNT` from the environment. Explicit `--jobs N` still takes priority.

#### Considered Alternatives

Chosen alternative: the resolved jobs value is written back into `args` before passing it to
  `Context::new`:

```rust
pub fn explain(mut args: Properties) {
      if args.jobs == 0 {
          if let Ok(val) = std::env::var("STARSHIP_JOBS_COUNT") {
              args.jobs = val.parse().unwrap_or(0);
          }
      }
      let context = Context::new(args, Target::Main);
```
This makes `explain` take `mut args` which did compile, but changes the interface.
But it is now essentially what the function does and invisible to callers. 

An alternative without mut was considered using struct update syntax:

```rust
let context = Context::new(Properties { jobs, ..args }, Target::Main);
```
This would need changes to the `Context` as its fields `terminal_width`, `path`, and `logical_path` are private and inaccessible from outside `context.rs`. 
A workaround could have been to add a `with_jobs` method on Properties:

```rust
impl Properties {
      pub fn with_jobs(self, jobs: i64) -> Self {
          Self { jobs, ..self }
      }
  }
```
I found it preferable to keep the changes localized. But let me know what you think.

#### Motivation and Context
I came across this because I was wondering about the star symbol in my prompt, found out what it was via the issue below and thought I could provide a fix following the suggestion in the first comment. 
Closes https://github.com/starship/starship/issues/3264

Screenshot for the difference with release version installed on my system and the one off this branch.
Screenshot 2026-03-24 at 01 09 16

See screenshot above
- [x] I have tested using **MacOS** (M4 Tahoe 26.2)

I'm unsure how to provide tests but happy to do so with guidance. 

#### Checklist:

### DIFF
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -57,7 +57,7 @@ starship_precmd() {
     # like z/autojump, which background certain jobs, do not cause spurious background jobs
     # to be displayed by starship. Also avoids forking to run `wc`, slightly improving perf.
     for job in $(jobs -p); do [[ $job ]] && ((NUM_JOBS++)); done
-    export STARSHIP_JOBS_COUNT="${NUM_JOBS}"
+    export STARSHIP_JOB_COUNT="${NUM_JOBS}"
 
     # Run the bash precmd function, if it's set. If not set, evaluates to no-op
     "${starship_precmd_user_func-:}"
--- a/src/init/starship.bash
+++ b/src/init/starship.bash
@@ -57,6 +57,7 @@ starship_precmd() {
     # like z/autojump, which background certain jobs, do not cause spurious background jobs
     # to be displayed by starship. Also avoids forking to run `wc`, slightly improving perf.
     for job in $(jobs -p); do [[ $job ]] && ((NUM_JOBS++)); done
+    export STARSHIP_JOBS_COUNT="${NUM_JOBS}"
 
     # Run the bash precmd function, if it's set. If not set, evaluates to no-op
     "${starship_precmd_user_func-:}"
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -7,7 +7,8 @@ function __starship_set_job_count --description 'Set STARSHIP_JOBS using fish jo
     else
         # Default behavior: count job groups
         set -g STARSHIP_JOBS (jobs -g 2>/dev/null | count)
-    end    
+    end
+    set -gx STARSHIP_JOBS_COUNT $STARSHIP_JOBS
 end
 
 function fish_prompt
--- a/src/init/starship.fish
+++ b/src/init/starship.fish
@@ -1,14 +1,13 @@
-function __starship_set_job_count --description 'Set STARSHIP_JOBS using fish job groups (or legacy PIDs if toggled)'
+function __starship_set_job_count --description 'Set STARSHIP_JOB_COUNT using fish job groups (or legacy PIDs if toggled)'
     # To force legacy behavior (process PIDs), set this variable to "false":
     #   set -g __starship_fish_use_job_groups "false"
     if test "$__starship_fish_use_job_groups" = "false"
         # Legacy behavior: counts PIDs (may overcount pipelines with terminated producers)
-        set -g STARSHIP_JOBS (jobs -p 2>/dev/null | count)
+        set -gx STARSHIP_JOB_COUNT (jobs -p 2>/dev/null | count)
     else
         # Default behavior: count job groups
-        set -g STARSHIP_JOBS (jobs -g 2>/dev/null | count)
+        set -gx STARSHIP_JOB_COUNT (jobs -g 2>/dev/null | count)
     end
-    set -gx STARSHIP_JOBS_COUNT $STARSHIP_JOBS
 end
 
 function fish_prompt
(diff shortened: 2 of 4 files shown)


src/print.rs
```diff
@@ -210,7 +210,12 @@ pub fn timings(args: Properties) {
     }
 }
 
-pub fn explain(args: Properties) {
+pub fn explain(mut args: Properties) {
```
davidkna, line 213: clippy has a feature for this that could be used.

src/init/starship.zsh
```diff
@@ -50,7 +50,7 @@ prompt_starship_precmd() {
 
     # Use length of jobstates array as number of jobs. Expansion fails inside
     # quotes so we set it here and then use the value later on.
-    STARSHIP_JOBS_COUNT="${#jobstates[*]}"
+    export STARSHIP_JOBS_COUNT="${#jobstates[*]}"
```
davidkna: I would prefer a name of `STARSHIP_JOBS` or `STARSHIP_JOB_COUNT`

src/init/starship.fish
```diff
@@ -7,7 +7,8 @@ function __starship_set_job_count --description 'Set STARSHIP_JOBS using fish jo
     else
         # Default behavior: count job groups
         set -g STARSHIP_JOBS (jobs -g 2>/dev/null | count)
-    end    
+    end
+    set -gx STARSHIP_JOBS_COUNT $STARSHIP_JOBS
```
davidkna: Please just change the variable name.

src/init/starship.bash
```diff
@@ -57,6 +57,7 @@ starship_precmd() {
     # like z/autojump, which background certain jobs, do not cause spurious background jobs
     # to be displayed by starship. Also avoids forking to run `wc`, slightly improving perf.
     for job in $(jobs -p); do [[ $job ]] && ((NUM_JOBS++)); done
+    export STARSHIP_JOBS_COUNT="${NUM_JOBS}"
```
davidkna: Please also change the variable name here.

davidkna (changes requested):

src/print.rs
maxmynter: You mean just running `cargo clippy --fix` before commit ? Thanks. My rust is rusty atm :/ 

If you meant something else, I'm not sure I understand ...

davidkna: Sorry for the typo I meant the clap env feature.

maxmynter:
Does the change look ok, @davidkna ?

PR #7345, opened 2026-03-24
fix: skip default-disabled modules when no user config exists

### LINKED ISSUE #7232: `pijul_channel` module triggered
#### Current Behavior

The `pijul_channel` module seems to be triggered even though it is said to be disabled by default. I checked the `print-config` output and it clearly is disabled.

```
 custom.jj          -  1063ms  -   "🥋 mynv | (empty) (no description set) "
 custom.git_status  -   479ms  -   ""
 directory          -   473ms  -   "pds "
 custom.git_branch  -   456ms  -   ""
 custom.git_commit  -   411ms  -   ""
 pijul_channel      -    14ms  -   ""
 battery            -     8ms  -   ""
 username           -     1ms  -   ""
 pulumi             -     1ms  -   ""
 line_break         -    <1ms  -   "\n"
 character          -    <1ms  -   "❯ "
```

Note: `pulumi` and `gcloud` also seem to be enabled.

#### Expected Behavior

The `disabled` property should be correctly handled.

#### Possible Solution

I don't see any.

#### Environment
- Starship version: 1.24.1
- nu version: 0.109.1
- Operating system: Windows 10.0.19045
- Git Commit Hash: 378f3e4
- Branch/Tag: master
- Rust Version: rustc 1.91.1 (ed61e7d7e 2025-11-07)
- Build Time: 2025-11-16 11:53:38 +00:00

```bash
# config.nu
#
# Installed by:
# version = "0.105.1"
#
# This file is used to override default Nushell settings, define
# (or import) custom commands, or run any other startup tasks.
# See https://www.nushell.sh/book/configuration.html
#
# This file is loaded after env.nu and before login.nu
#
# You can open this file in your default editor using:
# config nu
#
# See `help config nu` for more options
#
# You can remove these comments if you want or leave
# them for future reference.

$env.config.buffer_editor = "hx"

mkdir ($nu.data-dir | path join "vendor/autoload")
starship init nu | save -f ($nu.data-dir | path join "vendor/autoload/starship.nu")

use std "path add"
fnm env --json | from json | load-env
path add $env.FNM_MULTISHELL_PATH

zoxide init nushell | save -f ~/.zoxide.nu
source ~/.zoxide.nu

# jj util completion nushell | save completions-jj.nu
use completions-jj.nu * # Or
# source completions-jj.nu

def jjstar [] {
    jj log --revisions @ --no-graph --ignore-working-copy --color always --limit 1 --template '
        separate(" ",
            change_id.shortest(4),
            bookmarks,
            "|",
            concat(
            if(conflict, "💥"),
            if(divergent, "🚧"),
            if(hidden, "👻"),
            if(immutable, "🔒"),
            ),
            raw_escape_sequence("\x1b[1;32m") ++ if(empty, "(empty)"),
            raw_escape_sequence("\x1b[1;32m") ++ coalesce(
            truncate_end(29, description.first_line(), "…"),
            "(no description set)",
            ) ++ raw_escape_sequence("\x1b[0m"),
        )
    '
}
```

#### Starship Configuration

```toml
command_timeout = 1500

# custom module for jj status
[custom.jj]
description = "The current jj status"
when = "jj --ignore-working-copy root"
ignore_timeout = true
symbol = "🥋 "
command = '''
jjstar
'''

# optionally disable git modules when JJ is found.
# note that you'll need to add ${custom.git_branch}, ${custom.git_commit} etc
# into format: https://starship.rs/config/#default-prompt-format
[git_status]
disabled = true

[custom.git_status]
when = '''
! jj --ignore-working-copy root
'''
command = "starship module git_status"
style = "" # This disables the default "(bold green)" style
description = "Only show git_status if we're not in a jj repo"

[git_commit]
disabled = true

[custom.git_commit]
when = '''
! jj --ignore-working-copy root
'''
command = "starship module git_commit"
style = ""
description = "Only show git_commit if we're not in a jj repo"

[git_branch]
disabled = true

[custom.git_branch]
when = '''
! jj --ignore-working-copy root
'''
command = "starship module git_branch"
description = "Only show git_branch if we're not in a jj repo"
style = ""
shell = ["cmd", "/C"]
```

majiayu000:
#### Description

When a module like `pijul_channel` has no section in the user's TOML config, `is_module_disabled_in_config()` returns `false` because `get_module_config()` returns `None` and the disabled check defaults to not-disabled. This causes all 21 default-disabled modules to be invoked on every prompt render, adding unnecessary overhead (~14ms on Windows).

This PR adds a `DEFAULT_DISABLED_MODULES` constant and modifies `is_module_disabled_in_config()` to fall back to it when no explicit `disabled` field exists in user config:

- explicit `disabled = true` → skipped (unchanged)
- explicit `disabled = false` → executed (unchanged)  
- no config + default disabled → skipped (NEW)
- no config + default enabled → executed (unchanged)

#### Motivation and Context

Closes #7232

Default-disabled modules (pijul_channel, kubernetes, azure, etc.) are invoked on every prompt render even though they should be skipped, adding ~14ms overhead per prompt on Windows.

- Added unit tests verifying `is_module_disabled_in_config` returns `true` for default-disabled modules when no user config exists
- Added test verifying explicit `disabled = false` in user config overrides the default
- Added consistency test that deserializes each module's config from an empty TOML table and asserts the `disabled` field matches the `DEFAULT_DISABLED_MODULES` list
- `cargo test` — all 1173 tests pass
- `cargo clippy --all-targets --all-features` — no new warnings
- `cargo fmt --all -- --check` — clean

#### Checklist:

### DIFF
--- a/src/configs/mod.rs
+++ b/src/configs/mod.rs
@@ -334,7 +334,7 @@ pub struct FullConfig<'a> {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::module::{ALL_MODULES, DEFAULT_DISABLED_MODULES};
+    use crate::module::ALL_MODULES;
     use toml::value::Value;
 
     #[test]
@@ -343,41 +343,6 @@ mod test {
         let cfg_table = full_cfg.as_table().unwrap();
         for module in ALL_MODULES {
             assert!(cfg_table.contains_key(*module));
-        }
-    }
-
-    #[test]
-    fn test_default_disabled_modules_matches_config_defaults() {
-        let full_cfg = Value::try_from(FullConfig::default()).unwrap();
-        let cfg_table = full_cfg.as_table().unwrap();
-
-        // Verify every entry in DEFAULT_DISABLED_MODULES is a known module
-        for module in DEFAULT_DISABLED_MODULES {
-            assert!(
-                ALL_MODULES.contains(module),
-                "DEFAULT_DISABLED_MODULES contains '{module}' which is not in ALL_MODULES"
-            );
-        }
-
-        // Verify DEFAULT_DISABLED_MODULES matches actual config defaults
-        for module in ALL_MODULES {
-            let disabled_in_config = cfg_table
-                .get(*module)
-                .and_then(|v| v.as_table())
-                .and_then(|t| t.get("disabled"))
-                .and_then(|v| v.as_bool())
-                .unwrap_or(false);
-
-            let in_default_disabled = DEFAULT_DISABLED_MODULES.contains(module);
-
-            assert_eq!(
-                disabled_in_config,
-                in_default_disabled,
-                "Module '{module}': config default disabled={disabled_in_config}, \
-                 but {is}in DEFAULT_DISABLED_MODULES. Please update DEFAULT_DISABLED_MODULES \
-                 in src/module.rs.",
-                is = if in_default_disabled { "" } else { "not " }
-            );
         }
     }
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,5 +1,5 @@
 use crate::config::{ModuleConfig, StarshipConfig};
-use crate::configs::StarshipRootConfig;
+use crate::configs::{FullConfig, StarshipRootConfig};
 use crate::context_env::Env;
 use crate::module::Module;
 use crate::utils::{CommandOutput, PathExt, create_command, exec_timeout, read_file};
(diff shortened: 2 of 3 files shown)


davidkna:
Dues to this all default-disabled modules have this check:
```rs
if config.disabled {
    return None;
}
```

If this is behaving unexpectedly, please prefer improving the existing handling of default-disabled modules.

majiayu000:
Updated this based on the feedback in `4b1682d4`.\n\nI removed the separate `DEFAULT_DISABLED_MODULES` list and changed `is_module_disabled_in_config()` to fall back to the existing module defaults from `FullConfig::default()` (cached with `OnceLock`) when the user config does not explicitly set `disabled`.\n\nThis keeps the default value source in the config structs themselves while still skipping default-disabled modules before they are invoked. I also added a regression test for the case where a module section exists but `disabled` is not explicitly set.

majiayu000:
Fair enough — I'll put together a repro showing the overhead from invoking the default-disabled modules (timing with `starship timings`). If the numbers don't show anything meaningful I'll close this out.

On the approach: would checking the module's own `StarshipConfig` default be acceptable, or would you rather the fix live inside the module invocation path itself? Want to make sure I'm headed in the right direction before writing more code.

davidkna:
@majiayu000 For all purposes disabled modules are already skipped (`if module.disabled {return None}`), the issue is really whether this is not triggering in some cases or why it was taking that long in the issue that was raised.

majiayu000:
Thanks for @davidkna , I will work on the true reason

PR #7346, opened 2026-03-24
feat(pixi): expose PIXI_PROJECT_NAME as format placeholder

### LINKED ISSUE #7239: Pixi: expose `$PIXI_PROJECT_NAME`
When in a `pixi shell`, I'd like to see the name of the Pixi project in my prompt.

I'd like the Pixi module to expose the environment variable `PIXI_PROJECT_NAME` as a placeholder that I can include in `format`. Currently, only `PIXI_ENVIRONMENT_NAME` is exposed, which is just "default" unless you create more environments in that Pixi project.

Pixi itself modifies the prompt to include the name of the project once inside `pixi shell`, but that messes up the Starship prompt, of course, and can't be integrated and customized. I don't see an alternative within Starship itself - as far as I can tell, arbitrary env vars can't be used.

majiayu000:
#### Description

Expose `$PIXI_PROJECT_NAME` as a `project_name` format placeholder in the pixi module, alongside the existing `environment` variable.

#### Motivation and Context

Closes #7239

The pixi module already exposes `PIXI_ENVIRONMENT_NAME` but not `PIXI_PROJECT_NAME`. Users need this to distinguish between different pixi projects in their prompt.

- Added unit test `project_name_set`: sets `PIXI_PROJECT_NAME=my-project` and verifies it renders correctly
- Added unit test `project_name_not_set`: verifies the variable is absent when the env var is not set  
- All existing pixi tests continue to pass

#### Checklist:


src/modules/pixi.rs
```diff
@@ -6,13 +6,14 @@ use crate::utils::get_command_string_output;
 
 /// Creates a module with the current Pixi environment
 ///
-/// Will display the Pixi environment iff `$PIXI_ENVIRONMENT_NAME` is set.
+/// Will display the Pixi environment and project name iff `$PIXI_ENVIRONMENT_NAME` or `$PIXI_PROJECT_NAME` is set.
```
davidkna: This does not hold true as implemented.

src/modules/pixi.rs
```diff
@@ -44,6 +45,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             })
             .map(|variable| match variable {
                 "environment" => pixi_environment_name.clone().map(Ok),
+                "project_name" => pixi_project_name.clone().map(Ok),
```
davidkna: Please prefer initialising values lazily if possible 
```suggestion
                "project_name" => context.get_env("PIXI_PROJECT_NAME").map(Ok),
```

davidkna (changes requested):

src/modules/pixi.rs
majiayu000: Fixed — the `is_pixi_project` activation check now includes `context.get_env("PIXI_PROJECT_NAME").is_some()`, and the doc comment has been updated to match the actual behaviour.

src/modules/pixi.rs
majiayu000: Done — removed the eagerly-evaluated `pixi_project_name` variable and replaced it with `context.get_env("PIXI_PROJECT_NAME").map(Ok)` inline in the formatter closure, as suggested.

src/modules/pixi.rs
```diff
@@ -6,14 +6,15 @@ use crate::utils::get_command_string_output;
...
-/// Will display the Pixi version iff pixi files are detected or `$PIXI_ENVIRONMENT_NAME` is set.
+/// Will display the Pixi environment and project name iff `$PIXI_ENVIRONMENT_NAME` or `$PIXI_PROJECT_NAME` is set.
+/// Will display the Pixi version iff pixi files are detected, `$PIXI_ENVIRONMENT_NAME` is set, or `$PIXI_PROJECT_NAME` is set.
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("pixi");
     let config: PixiConfig = PixiConfig::try_load(module.config);
 
     let pixi_environment_name = context.get_env("PIXI_ENVIRONMENT_NAME");
     let is_pixi_project = pixi_environment_name.is_some()
+        || context.get_env("PIXI_PROJECT_NAME").is_some()
```
davidkna: Please just store it in a variable to avoid refetching in `map` or update the comment and docs to avoid enabling in case of `PIXI_PROJECT_NAME` (I assume that if `PIXI_ENVIRONMENT_NAME` is set there is no need to check `PIXI_PROJECT_NAME`)

majiayu000: Stored it in a variable now, same pattern as pixi_environment_name. Also fixed the doc comment to not mention project_name.

src/modules/pixi.rs
majiayu000: Updated — removed project_name from the doc comment since it doesn't depend on PIXI_ENVIRONMENT_NAME.

majiayu000: Updated — removed project_name from the doc comment since it doesn't depend on PIXI_ENVIRONMENT_NAME.

src/modules/pixi.rs
```diff
@@ -32,6 +32,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         pixi_environment_name
     };
 
+    let pixi_project_name = context.get_env("PIXI_PROJECT_NAME");
+
```
davidkna:
```suggestion
```

davidkna (approved):

PR #7367, opened 2026-03-31
feat(aws): introduce support for AWS to support alias patterns

jacobbednarz:
Much like the Kubernetes module, it is often nice to have a regex pattern for aliases or regions which can map to more usable or friendly names.

Examples

```toml
[[aws.profiles]]
# "bold red" style + default symbol when AWS profile equals "production" *and* the current region
# matches "us-east-*"
profile_pattern = "production"
region_pattern = "us-east-.*"
style = "bold red"
profile_alias = "prod"

[[aws.profiles]]
# Using capture groups
# Profiles with a company prefix like "company-prod-team" can be shortened to "team (prod)":
profile_pattern = "company-(?P<env>[\\w-]+)-(?P<team>[\\w-]+)"
profile_alias = "$team ($env)"
```

As part of this change, I've left the previous AWS region + profile aliases in place but deprecated to be removed in a much later version.

Closes #6099 

#### Screenshots (if appropriate):

n/a

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [x] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.

PR #7368, opened 2026-03-31
fix(ci): dprint wants `plugin@checksum` syntax

jacobbednarz:
Updates the dprint version file to include the checksum version which is causing failures on other PRs (like #7367).

Error in CI: https://github.com/starship/starship/actions/runs/23777843801/job/69283523808?pr=7367#step:3:13

No dependencies were altered or changed as part of this, just fixing the lockfile syntax.

### DIFF
--- a/.dprint.json
+++ b/.dprint.json
@@ -25,9 +25,9 @@
     "target/"
   ],
   "plugins": [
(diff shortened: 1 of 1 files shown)


davidkna:
This appears to have been fixed with the most recent release of dprint.

jacobbednarz:
thanks for the heads up @davidkna 🙇

PR #7381, opened 2026-04-05
feat: implement referencing next segment's style in conf file

christophehenry:
Ref: #7143

#### Description
Solves [#6558-2888265896](https://github.com/starship/starship/issues/6558#issuecomment-2888265896). Add ability to reference next modules' foreground and background. Useful to make a spacer seperating 2 modules, merging the style of both.

#### Motivation and Context
Starship is missing the ability to create a spacer sharing colors from surrounding modules. Some presets like [Gruvbox](https://starship.rs/presets/gruvbox-rainbow) partly implement this design. But since color needs to be hard-coded, it gets broken as soon as one module is conditionnaly rendered. This PR allows to define colors that will always match the following module no matter what it is.

#### Checklist:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -268,42 +268,24 @@ where
 }
 
 #[derive(Clone, Copy, Debug, PartialEq)]
-enum PrevColor {
-    Fg,
-    Bg,
+pub(crate) enum RefColor {
+    PrevFg,
+    PrevBg,
+    NextFg,
+    NextBg,
 }
 
 #[derive(Clone, Copy, Debug, Default, PartialEq)]
 /// Wrapper for `nu_ansi_term::Style` that supports referencing the previous style's foreground/background color.
 pub struct Style {
-    style: nu_ansi_term::Style,
-    bg: Option<PrevColor>,
-    fg: Option<PrevColor>,
+    pub(crate) style: nu_ansi_term::Style,
+    pub(crate) bg: Option<RefColor>,
+    pub(crate) fg: Option<RefColor>,
 }
 
 impl Style {
-    pub fn to_ansi_style(&self, prev: Option<&nu_ansi_term::Style>) -> nu_ansi_term::Style {
-        let Some(prev_style) = prev else {
-            return self.style;
-        };
-
-        let mut current = self.style;
-
-        if let Some(prev_color) = self.bg {
-            match prev_color {
-                PrevColor::Fg => current.background = prev_style.foreground,
-                PrevColor::Bg => current.background = prev_style.background,
-            }
-        }
-
-        if let Some(prev_color) = self.fg {
-            match prev_color {
-                PrevColor::Fg => current.foreground = prev_style.foreground,
-                PrevColor::Bg => current.foreground = prev_style.background,
-            }
-        }
-
-        current
+    pub fn to_ansi_style(&self) -> nu_ansi_term::Style {
+        self.style
     }
 
     fn map_style<F>(&self, f: F) -> Self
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -316,19 +298,127 @@ impl Style {
...
+        };
+
+        if let Some(prev_style) = maybe_previous {
+            if let Some(ref_color) = self.bg {
+                match ref_color {
+                    RefColor::PrevFg => {
+                        if prev_style.fg.is_some() {
+                            panic!(
+                                "Circular reference: segment background references back to a segment itself referencing forward"
+                            )
```
christophehenry, line 337: Should it `panic!()` or return `Result` or default to `Style::default().background` with a log message? Also, being able to point where in the line this happens would be nice to debug but I don't know hos to do so.

src/formatter/string_formatter.rs
```diff
@@ -748,23 +749,42 @@ mod tests {
...
+
+        assert_eq!(result.len(), 2);
+
+        // First segment: should inherit bg from second
+        let first_ansi = result.get(0).unwrap();
+        assert_eq!(
+            first_ansi.style_ref().background,
+            Some(nu_ansi_term::Color::Rgb(154, 52, 142))
+        );
+    }
```
christophehenry, line 786: Shouldn't these tests be moved elsewhere since style resolving is entirely happening in `module.rs` and `config.rs` now? Plus they feel lik they're duplicates of several tests in`config.rs`, for instance `table_get_styles_previous`.

src/config.rs
```diff
@@ -928,42 +1035,132 @@ mod tests {
...
+        let bg_prev_bg = <StyleWrapper>::from_config(&Value::from("bg:next_bg"))
+            .unwrap()
+            .0;
+        assert_eq!(
+            bg_prev_bg
+                .resolve(Some(&prev_style), Some(&next_style))
+                .to_ansi_style(),
+            AnsiStyle::new().on(Color::Cyan)
+        );
+    }
```
christophehenry, line 1155: Should these tests be more extensive? For instance, I am not sure how styles are transmitted to `Segment::LineTerm` is tested somewhere.

src/config.rs
```diff
@@ -316,19 +298,127 @@ impl Style {
...
+                    }
+                    RefColor::PrevBg => {
+                        if prev_style.bg.is_some() {
+                            panic!(
+                                "Circular reference: segment foreground references back to a segment itself referencing forward"
+                            )
+                        }
+                        result.style.foreground = prev_style.style.background
+                    }
+                    _ => { /* Referencing forward not resolved here */ }
```
christophehenry, line 370:
```suggestion
                    _ => unreachable!()
```
?

src/config.rs
davidkna: `panic!()` is not acceptable for a shell prompt.

davidkna (commented):
I would prefer avoiding much of the complexity by limiting next-segement references to a single step, as I have mentioned on your previous PR.

christophehenry:
Yes. Sorry I tried that solution but I couldn't resolve to using it. This means that for each segment, I need to go through each following segment until I resolved both foreground and background colors. And I know it is unlikely but worst case scenario has a complexity of O(n!). The alternative would be to make the segments mutable so I only need to go through the segments once. I also tried that and I don't exactly remember why segments couldn't be made mutable. I'm still very new to Rust so it is highly possible I missed something though.

christophehenry:
@davidkna Can you tell me if this is a firm no-go or if we can make this work so I can progress?

davidkna:
@christophehenry Please check if #7482 works for you.

christophehenry:
Since `next_fg` and `next_bg` are not transitive, it causes a slight issue when one optional module is skipped and 2 or more separators are rendered in a succession. In the following example, the `directory` is followed by a separator then `git_status` and then a separator; the separators are defined as `[](bg:next_bg fg:prev_bg)`:



Since in this case, I'm not in a git directory, both separators follow in a row so the second one is unable to take the style of the first one.

Since I was always planning on adding a special separator module that would be handled so that successive separators are eliminated, I would argue it's not an issue here and #7482 is ok until a valid use-case for transitive `next_fg` and `next_bg` arise.

The rest of the shell prompt works as expected so I will close this PR in favor of #7482.

PR #7394, opened 2026-04-12
feat(nix-shell): Add level variable to show nix shell depth

### LINKED ISSUE #7369: nix-shell: Add support for NIX_SHELL_LEVEL
## Feature Request

Recently lix (one of the nix implementations) started supporting `NIX_SHELL_LEVEL`. See [the release notes](https://lix.systems/blog/2026-03-25-lix-2.95-release/) It represents how many levels deep you are in nested nix shells. I would really like to display this as part of the prompt so I know how many shells I do have to exit to get rid of all the shells.

The problem is that it is hard to know how many shells you have to exit out of until you are back to the base shell. So usually the workaround is to close too many - until the terminal closes because you have closed the base shell. This is usually only slightly annoying, but can become majorly annoying if getting the base shell to that location was a lot of work.

I would like to be able to reference the variable as part of the format string. Something like this:

```toml
[nix_shell]
heuristic = true
impure_msg = '🧪'
pure_msg = '✨'
unknown_msg = '[unknown shell](bold yellow)'
format = '[$symbol $state $level $name](bold blue) '
symbol = "❄️"
```

I tried using the env_vars module to get this working but couldn't with he provided documentation.

maddymakesgames:
#### Description

Adds a variable `$level` that is mirrors the environment variable `$NIX_SHELL_LEVEL` added by Lix.
#### Motivation and Context

Closes #7369 

In the 2.95 release of Lix, a fork of Nix, they added `$NIX_SHELL_LEVEL`  which is similar to `$SHLVL` but only increments on `nix shell` and `nix-shell`.
This allows prompts to now show how many nested `nix shell`s they are currently in, which is nice to know for people who use it extensively. 

This could be implemented just using the environment variable module but having it as part of the `nix_shell` module is nicer since many people might not know it exists and requires less setup for users.

Wrote a unit test to ensure that `$NIX_SHELL_LEVEL` maps to `$level`.
Also tested `starship prompt` in both fish and zsh while on NixOS.
I don't have Mac or Windows Lix environments set up and so I can't test there, but I can't imagine there would be much difference.
Changes are pretty minimal overall and so I don't expect much room for bugs.

#### Checklist:

### DIFF
--- a/src/modules/nix_shell.rs
+++ b/src/modules/nix_shell.rs
@@ -45,6 +45,7 @@ impl NixShellType {
 /// The module will use the `$IN_NIX_SHELL` and `$name` environment variable to
 /// determine if it's inside a nix-shell and the name of it.
 ///
+///
 /// The following options are available:
 ///     - `impure_msg` (string)  // change the impure msg
 ///     - `pure_msg` (string)    // change the pure msg
@@ -57,6 +58,8 @@ impl NixShellType {
 ///     - impure         // $name == "" in an impure nix-shell
 ///     - unknown (name) // $name == "name" in an unknown nix-shell
 ///     - unknown        // $name == "" in an unknown nix-shell
+///
+///  When using Lix 2.95+, will also have $level from `$NIX_SHELL_LEVEL`
 pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
     let mut module = context.new_module("nix_shell");
     let config: NixShellConfig = NixShellConfig::try_load(module.config);
@@ -68,6 +71,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         NixShellType::Impure => config.impure_msg,
         NixShellType::Unknown => config.unknown_msg,
     };
+    let shell_level = context.get_env("NIX_SHELL_LEVEL");
 
     let parsed = StringFormatter::new(config.format).and_then(|formatter| {
         formatter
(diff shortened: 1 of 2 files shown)


leana8959:
Hello, would it be possible to configure how one wants to show the count? I personally really like to repeat the symbol n times (for n nested level) instead of just showing the number.

image

maddymakesgames:
thats definitely possible to implement but I wanted this PR to be as simple as possible, which would be just echoing the environment variable.

src/modules/nix_shell.rs
```diff
@@ -82,6 +86,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
             })
             .map(|variable| match variable {
                 "name" => shell_name.as_ref().map(Ok),
+                "level" => shell_level.as_ref().map(Ok),
```
davidkna, line 89: I would prefer only reading the variable on demand:
```suggestion
                "level" => context.get_env("NIX_SHELL_LEVEL").as_ref().map(Ok),
```

davidkna (approved):

src/modules/nix_shell.rs
maddymakesgames: Just doing this change creates a temporary so the new commit no longer returns references & also reads name on demand too.

davidkna:
Thanks for the contribution @maddymakesgames!

dwt:
Also from me! Thanks!

PR #7395, opened 2026-04-12
feat(yadm): add native module (#3935)

### LINKED ISSUE #3935: Module to track YADM status
## Feature Request

[YADM](https://yadm.io/) is a dotfiles manager. It:

* stores your dotfiles in a bare Git repository (for easy versioning and syncing)
* overlays that repo onto `$HOME`
* doesn't consider untracked files (so your files, other projects, etc. don't get checked into your YADM repo)

Basically you get a Git repo where you can put your `.config/`, `.profile`, `.(z|bash|.*)rc` files and sync them between hosts.

It would be great to have a module that checks if that repo has uncommitted or unsync'ed changes. The existing `git*` plugins would likely be a good place to look for inspiration.

Vassili-REGNIER:
#### Description
This PR adds a native `yadm` module to track the status of local dotfiles. 
It automatically detects the `yadm` bare repository (typically located at `~/.local/share/yadm/repo.git` or `~/.yadm/repo.git` and checks its status.

To keep the module simple and optimized for typical small dotfiles repositories:
- It relies exclusively on a standard git status --porcelain=2 --branch subprocess.
- It verifies if the repository is dirty (uncommitted changes) or ahead of upstream (unpushed commits) so users know if their local dotfiles are out of sync.
- It seamlessly integrates with the command_timeout configuration.

#### Motivation and Context
Users currently have to rely on complex custom shell scripts to track their YADM dotfiles. This native module provides built-in, fast, and reliable support right out of the box.

Closes #3935

yadm-module-demo

Tested locally on WSL (Ubuntu).
- Initialized a local yadm repo and added test configuration files.
- Verified the module correctly triggers and displays the symbol on uncommitted changes (dirty index/worktree).
- Verified the module correctly triggers on unpushed commits (ahead of upstream).
- Ensured the module disappears when the working tree is clean and synced.
- Ran `cargo test` to ensure the simulated git environment tests pass.

#### Checklist:


eliottreich:
Flagging: there's a $25 bounty for new Starship prompt modules on [Code.TaskBounty](https://code.task-bounty.com/task/starship-add-new-shell-prompt-module). New module requests like this are exactly the type of issue that benefits from competing implementations.

https://code.task-bounty.com/task/starship-add-new-shell-prompt-module

davidkna:
@eliottreich Please don't, at the current stage I will close any competing PR without further notice. In addition, if you are working on a bounty, please always mention this in the PR description.

src/modules/yadm.rs
davidkna, line 1: This is too complicated, please simplify this.

davidkna (changes requested):

src/modules/yadm.rs
Vassili-REGNIER: Thanks for the review @davidkna! 

You were right, the initial version was definitely overkill. I just pushed a refactored version. 

I removed the `gix` and multithreading logic entirely. The module now just relies on a standard `git status` subprocess. I tested it locally and the execution time remains exactly the same (around 0.13s).

Let me know if this looks better!

davidkna: This is better, but I would prefer sharing some of the utilities with `git_status`, and it would be nice if this would use `exec_git`.

And even the gitoxide version could have been much simpler, if it e.g. used `is_dirty()`

Vassili-REGNIER: Thanks for the review @davidkna! I just pushed a refactor:

- Reused `exec_git`: Added a new `Repo::open_bare` helper. Since YADM uses `$HOME` as a work-tree instead of the `current_dir`, this lets us build a standalone repo and fully reuse `exec_git` without duplicating the timeout/fsmonitor logic.

- Shared utilities: Extracted a `is_dirty_or_ahead` helper in `git_status.rs` that reuses the existing `RepoStatus` porcelain v2 parser. No more duplicated parsing logic.

- Regarding `is_dirty()`: I kept the single `git` subprocess path for maximum simplicity right now to avoid logic drift. I'm happy to add a `gix` fast path in a follow up PR if you'd prefer.

Let me know if this looks good to you.

PR #7103, opened 2025-11-07
Revert "fix: consider text-only meta variables in conditionals set if non-empty (#6338)"

davidkna:
This reverts commit 286b0f2be30db81bc948579c137f0cc646fcde65.

#### Description

The #6338 PR some fallout, so I would suggest reverting it. E.g. the no-empty-icon preset was broken with #6338 applied.

#### Motivation and Context

Closes #7077

#### How Has This Been Tested?

- [x] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #7401, opened 2026-04-17
fix(status): mirror symbols inside conditional formats

### LINKED ISSUE #5471: Status symbol not mirrored in a conditional format string
#### Current Behavior
The prompt is `a` when a command exits successfully

#### Expected Behavior
The prompt should be `aa` instead

image

#### Environment
- Starship version: 1.16.0
- fish version: fish, version 3.6.1
- Operating system: NixOS 23.11.0
- Terminal emulator: foot 1.15.3
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.72.0 (5680fa18f 2023-08-23) (built from a source tarball)
- Rust channel:  release

```fish
if status is-interactive
    starship init fish | source
end
```

#### Starship Configuration

```toml
format = "$status"

[status]
disabled = false
format = "$symbol($symbol) "
map_symbol = true
success_symbol = 'a'
```

lawrence3699:
#### Description
Fix mirrored `status` symbols inside conditional format segments.

#### Motivation and Context
Closes #5471.

Before this change, a status config such as:

```toml
[status]
disabled = false
map_symbol = true
success_symbol = 'a'
format = '$symbol($symbol) '
```

rendered `a ` instead of `aa `.

The root cause was that the formatter treated meta variables with literal output as empty when deciding whether to render a conditional segment. This patch checks the rendered output of meta variables in that code path, which fixes the mirrored `$symbol` case without changing `status` symbol format-string semantics.

Validation:
- `cargo test success_symbol_is_mirrored_in_conditional_format -- --nocapture`
- `cargo test test_conditional_meta_variable -- --nocapture`
- `cargo test test_conditional_meta_variable_literal_text -- --nocapture`
- `cargo test test_conditional_meta_variable_styled_text -- --nocapture`
- `cargo test test_nested_conditional -- --nocapture`
- `cargo test modules::status::tests::special_symbols -- --nocapture`
- `cargo test modules::status::tests::pipeline_uses_pipestatus_format -- --nocapture`
- `cargo test modules::status::tests::failure_plaintext_status -- --nocapture`
- `cargo test modules::status::tests::failure_hex_status -- --nocapture`
- `cargo fmt --check`

Unrelated note: `cargo test modules::status::tests::pipestatus_width -- --nocapture` also fails on clean `master` in this environment because `TERM=dumb`, so I treated that as baseline noise rather than a regression from this patch.

### DIFF
--- a/src/formatter/string_formatter.rs
+++ b/src/formatter/string_formatter.rs
@@ -346,6 +346,8 @@ impl<'a> StringFormatter<'a> {
                             fn should_show_elements<'a>(
                                 format_elements: &[FormatElement],
                                 variables: &'a VariableMapType<'a>,
+                                style_variables: &'a StyleVariableMapType<'a>,
+                                context: Option<&Context>,
                             ) -> bool {
                                 format_elements.get_variables().iter().any(|var| {
                                     variables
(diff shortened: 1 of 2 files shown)


davidkna:
Are you aware of https://github.com/starship/starship/pull/7103?

PR #7402, opened 2026-04-17
fix: preserve explicit starship path in init stubs

### LINKED ISSUE #6775: `starship init bash` generates init code which calls a different starship executable
#### Current Behavior
When I build starship from the repository and then run:
```
./target/debug/starship init bash
```
I get:
```
eval -- "$(/home/komar/.nix-profile/bin/starship init bash --print-full-init)"
```
which installs a different instance of starship (the one installed in my system) in my current bash session.
 
#### Expected Behavior
starship should install the exact version which generated the install expression - this way I could switch between instances for testing purposes or even have multiple versions of starship in the system.

#### Possible Solution
The use of `which` was introduced here, but no rationale was provided: https://github.com/starship/starship/commit/cc2c8c4a5450f2811612129abfbdc1aba12def91.
Reverting this commit would directly fix the problem (but I don't know what kind of usage scenarios it would then break)

#### Environment
- Starship version: 
   ```
   starship 1.23.0
   branch:master
   commit_hash:c4100d9c
   build_time:2025-06-03 20:08:47 +02:00
   build_env:rustc 1.85.0 (4d91de4e4 2025-02-17),stable-x86_64-unknown-linux-gnu
- Shell type: bash
- Shell version: `GNU bash, version 5.2.37(1)-release (x86_64-pc-linux-gnu)`
- Shell plugin manager: none
- Operating system: NixOS 25.05

seems irrelevant

#### Starship Configuration
the bug is starship-configuration-agnostic

lawrence3699:
#### Description
Use the currently running starship binary when generating init stubs from an explicit path invocation.

#### Motivation and Context
Closes #6775.

`./target/debug/starship init bash` currently re-resolves `starship` on `PATH`, which can swap in a different binary than the one that generated the init code. This keeps the existing `PATH` lookup for bare `starship` invocations, while preserving the current executable when init was invoked via an explicit path like `./target/debug/starship` or `/path/to/starship`.

Commands run:

```sh
cargo test init::tests --lib
cargo fmt --check
cargo clippy --all-targets --all-features -- -D warnings
```

I also reproduced the bug locally by prepending a fake `starship` executable to `PATH` and verifying that `./target/debug/starship init bash` now keeps the debug binary instead of switching to the `PATH` entry.

#### Checklist:

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -26,11 +26,37 @@ struct StarshipPath {
 impl StarshipPath {
     fn init() -> io::Result<Self> {
         let exe_name = option_env!("CARGO_PKG_NAME").unwrap_or("starship");
-
-        let native_path = which(exe_name).or_else(|_| env::current_exe())?;
+        let native_path = Self::resolve_native_path(
+            env::args_os().next().as_deref(),
+            || which(exe_name).map_err(io::Error::other),
+            env::current_exe,
+        )?;
 
         Ok(Self { native_path })
     }
+
+    fn resolve_native_path<LookupPath, CurrentExe>(
+        argv0: Option<&OsStr>,
+        lookup_path: LookupPath,
+        current_exe: CurrentExe,
+    ) -> io::Result<PathBuf>
+    where
+        LookupPath: FnOnce() -> io::Result<PathBuf>,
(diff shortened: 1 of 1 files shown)


lawrence3699:
Closing this to avoid multiple concurrent PRs from my account in the same repo.

PR #7413, opened 2026-04-21
fix(git): render modules inside attached worktrees of bare repositories

KKamJi98:
#### Description

Replace the early bare-repository guards in `git_status`, `git_branch`, and
`git_metrics` so attached worktrees of bare repositories are treated as
worktrees instead of true bare repositories.

This switches the guard from `repo.open().is_bare()` / `gix_repo.is_bare()` to
`repo.workdir.is_none()` and adds regression tests for all three modules using
the `git clone --bare` + `git worktree add` layout.

#### Motivation and Context

`git_status` regressed in v1.25.0 after #5655, and the same bare-vs-worktree
distinction also affected `git_branch` and `git_metrics`.

In an attached worktree of a bare-cloned repository, the shared repo config
still has `core.bare = true`, so `gix::Repository::is_bare()` returns `true`
even though the current context has a writable working tree. Starship already
discovers that working tree and stores it in `context::Repo.workdir`, so using
that signal is both smaller and more accurate for these module guards.

`git_status.use_git_executable = true` does not mitigate the `git_status` part
of the bug because the early return happens before either status backend runs.

Related: #6367, #6417, #6725

N/A

- `git diff --check`
- Added regression tests for attached worktrees of bare repositories in:
  - `git_status`
  - `git_branch`
  - `git_metrics`
- I could not run `cargo fmt`, `cargo test`, or `cargo clippy` in this
  environment because the Rust toolchain is not installed locally

#### Checklist:

### DIFF
--- a/src/modules/git_metrics.rs
+++ b/src/modules/git_metrics.rs
@@ -26,7 +26,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let repo = context.get_repo().ok()?;
     let gix_repo = repo.open();
-    if gix_repo.is_bare() {
+    if repo.workdir.is_none() {
         return None;
     }
     let status_module = context.new_module("git_status");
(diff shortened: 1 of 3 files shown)


davidkna (commented):
Thanks for the PR! Please also update `git_metrics` and `git_branch`

KKamJi98:
Updated the PR to also handle git_branch and git_metrics alongside git_status, with matching regression tests for attached worktrees of bare repositories. Latest commit: b8955d88.

src/modules/git_metrics.rs
```diff
@@ -26,7 +26,7 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
 
     let repo = context.get_repo().ok()?;
     let gix_repo = repo.open();
-    if gix_repo.is_bare() {
+    if repo.workdir.is_none() {
```
davidkna, line 29: Please fix the clippy warning:
```
29 | /     if gix_repo.workdir().is_none() {
30 | |         return None;
31 | |     }
   | |_____^ help: replace it with: `gix_repo.workdir()?;`
   |
```

davidkna:
Closed in favor of #7421

PR #7390, opened 2026-04-11
ci: Pin actions digests, remove unnecessary actions and add minimumReleaseAge

davidkna:
#### Description

With supply-chain attacks on the rise, this PR aims to harden the project's CI configuration:

Changes:
- pin actions digests via `pinact`
- remove unnecessary actions
  -  `dtolnay/rust-toolchain` -> plain `rustup`
  - `giraffate/clippy-action` (unmaintained) -> plain `clippy`
  - `reviewdog/action-suggester` -> manual error annotation
  - `softprops/action-gh-release` -> gh cli
- Improve renovate config:
  - Add `minimumReleaseAge`
  - Instead of picking items from `config:base`, use `ignoredPresets` to exclude undesirable presets
- `taiki-e/install-action`: version installed tools and add renovate regex manager for updates

#### Motivation and Context

Closes #

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #7415, opened 2026-04-21
fix(python): align `python_binary` schema with argument support

martimlobao:
#### Description
Fixes config schema generation for nested `VecOr` types and regenerates `.github/config-schema.json`.

This updates the Rust schema source (`VecOr<T>: JsonSchema`) so nested forms get unique schema identities, avoiding `$defs` collisions. With that fix, `python_binary` now resolves to `VecOr<VecOr<string>>` (`VecOr_VecOr_string`) in the generated schema and supports:
- `string`
- `string[]`
- `string[][]`

Adds a targeted regression test (`vec_or_schema_ids_distinguish_nesting` in `src/config.rs`, behind `config-schema`) that asserts nested and non-nested `VecOr` types have distinct `schema_id`/`schema_name`, preventing future schema identity collisions.

Also updates Python config docs to with example usage.

#### Motivation and Context
The Python module has supported argument-bearing `python_binary` commands since [#6523](https://github.com/starship/starship/pull/6523), but schema generation could previously collapse different `VecOr` nestings to the same schemars identity and `$defs` entry.

This change addresses that root cause in Rust generation logic, so schema validation stays aligned with actual parser/runtime behavior.  
The regenerated schema now uses `VecOr_*` definitions for `VecOr`-based shapes, and `Either2` now corresponds to the bool/string union used by `custom.when`.

- [x] Added and ran a regression test asserting nested `VecOr` schema identities are distinct:
  - `cargo test --locked --features config-schema vec_or_schema_ids_distinguish_nesting`
- [x] Regenerated schema and verified committed file matches generator output:
  - `cargo run --locked --features config-schema -- config-schema > .github/config-schema.json`
- [x] Confirmed `python_binary` now references the nested generated `VecOr` schema form
- [x] Tested on **macOS**
- [ ] Tested on **Linux**
- [ ] Tested on **Windows**

#### Checklist:

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -5832,7 +5832,7 @@
           "default": "pyenv "
         },
         "python_binary": {
-          "$ref": "#/$defs/Either2",
+          "$ref": "#/$defs/Either4",
           "default": [
             [
               "python"
@@ -7220,6 +7220,28 @@
         },
         {
           "type": "string"
+        }
+      ]
+    },
+    "Either4": {
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        {
+          "type": "array",
+          "items": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          }
         }
       ]
     }
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -5668,7 +5668,7 @@
       "type": "object",
       "properties": {
         "pixi_binary": {
-          "$ref": "#/$defs/Either2",
+          "$ref": "#/$defs/VecOr_string",
           "default": [
             "pixi"
           ]
@@ -5724,7 +5724,7 @@
       },
       "additionalProperties": false
     },
-    "Either2": {
+    "VecOr_string": {
       "anyOf": [
         {
           "type": "string"
@@ -5832,7 +5832,7 @@
           "default": "pyenv "
         },
         "python_binary": {
-          "$ref": "#/$defs/Either4",
+          "$ref": "#/$defs/VecOr_VecOr_string",
           "default": [
             [
               "python"
@@ -5915,6 +5915,19 @@
         }
       },
       "additionalProperties": false
+    },
+    "VecOr_VecOr_string": {
+      "anyOf": [
+        {
+          "$ref": "#/$defs/VecOr_string"
+        },
+        {
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/VecOr_string"
+          }
+        }
+      ]
     },
     "QuartoConfig": {
       "type": "object",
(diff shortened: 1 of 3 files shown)


docs/config/README.md
```diff
@@ -4031,10 +4031,16 @@ By default, the module will be shown if any of the following conditions are met:
...
-> The `python_binary` variable accepts either a string or a list of strings.
-> Starship will try executing each binary until it gets a result. Note you can
-> only change the binary that Starship executes to get the version of Python not
-> the arguments that are used.
+> The `python_binary` variable accepts either:
+>
+> - a string (e.g. `'python3'`),
+> - a list of strings (e.g. `['python', 'python3']`)
+> - a list of lists of strings, representing commands with optional arguments (e.g.
+>   `[['uv', 'run', 'python'], 'python3']`)
```
martimlobao:
```suggestion
>   `[['uv', 'run', 'python'], ['python3']]`)
```

.github/config-schema.json
davidkna, line 1: This file is generated, any changes should be made via the source rust code.

docs/config/README.md
```diff
@@ -4031,10 +4031,16 @@ By default, the module will be shown if any of the following conditions are met:
 | `disabled`           | `false`                                                                                                      | Disables the `python` module.                                                         |
 
 > [!TIP]
-> The `python_binary` variable accepts either a string or a list of strings.
-> Starship will try executing each binary until it gets a result. Note you can
-> only change the binary that Starship executes to get the version of Python not
-> the arguments that are used.
+> The `python_binary` variable accepts either:
```
davidkna, line 4034: `uv` specially was not documented on purpose due to that not being secure. Any tip for it should mention that it can lead to unexpected untrusted binary execution.

The primary example could e.g. use a package manager with a trust system like mise.

martimlobao: @davidkna what do you mean about it not being secure? Just that running a command with arbitrary arguments is insecure in general or do you mean something specific to do with `uv`?

I'm not familiar with `mise`, do you have an example you'd like to see?

davidkna: If you `cd python_project` it will more or less run any binary located at `<project root>/.venv/bin/python` without interaction.

For mise an example could be this: `['mise', 'exec', '--', 'python']`.

.github/config-schema.json
martimlobao: @davidkna i just noticed there's an existing CI schema check that's supposed to catch this sort of drift but it didn't fire here. i think that's because it overwrites the file and relies on the reviewdog suggester path (which can be a no-op / not fail hard depending on PR token/permissions). mind if i include a change here to make it a plain generated-vs-committed diff check so CI always flags it?

davidkna: @martimlobao #7390 should fix that issue.

martimlobao: @davidkna ah sweet, i'll leave that part out of this PR then

davidkna (approved):

davidkna:
Thanks for the contribution @martimlobao!

PR #7432, opened 2026-04-28
chore(deps): update gix to 0.83 as 0.82 was yanked

Byron:
It's just that, sorry for the hassle.
That particular error source also shouldn't be causing this again.


davidkna (approved):

davidkna:
Thanks for the fix!

PR #7436, opened 2026-04-28
fix: wrap all CSI sequences in wrap_colorseq_for_shell, not just SGR

### LINKED ISSUE #7435: `wrap_colorseq_for_shell` breaks prompt for non-SGR escape sequences
#### Current Behavior

`wrap_colorseq_for_shell` only handles SGR sequences (those ending in `m`). Any other CSI sequence in the prompt, like `\x1b[K` (erase to end of line), corrupts the shell wrapper output and breaks readline's width calculation.

This matters because many terminal emulators (xterm, konsole, foot, wezterm, ghostty, rio, ...) have a rendering bug where background colors bleed across soft-wrapped lines when using powerline-style presets. A possible workaround is adding a custom module that emits `\x1b[49m\x1b[K` at the end of the prompt:

```toml
format = """
...\
${custom.erase_line}\
$line_break$character"""

[custom.erase_line]
description = "Resets background color and erases to end of line"
command = "printf '\\e[49m\\e[K'"
when = true
format = '$output'
shell = ['sh']
```

This fixes the color bleed, but introduces a second bug: typing long input lines no longer wraps to the next line. Instead, the cursor jumps back to column 0 and visually overwrites the prompt.

#### Expected Behavior

Non-SGR CSI sequences like `\x1b[K` should be wrapped in `\[...\]` (bash) / `%{...%}` (zsh) just like SGR sequences, so that readline computes the prompt width correctly. This would allow the `custom.erase_line` workaround (and any other non-SGR sequences) to work without breaking line wrapping.

**Bug 1: color bleed on soft-wrapped lines without the workaround:**

Image

**Bug 2: input overwrites prompt when using the workaround:**

Image

The root cause is in `wrap_colorseq_for_shell` in `src/utils/mod.rs`. The function opens a shell wrapper (`\[` for bash) when it sees `\x1b`, then closes it (`\]`) when it sees `m`. But `[` (the CSI introducer that appears in every `\x1b[` sequence) has byte value 0x5B ( which falls in the `@`..=`~` range (0x40-0x7E). The code has no special handling for it.

For `\x1b[K`, the sequence is parsed as`\[\x1b[\]K`. Readline counts `K` as a visible character and gets the width wrong. The erase sequence itself also runs at the wrong point.

This can be verified by inspecting the prompt output:

```
$ starship prompt --status=0 | od -c | grep K
0000400   9   m   \   [ 033   [   \   ]   K
```

The `K` is outside the `\[...\]` wrapper.

#### Possible Solution

Replace the boolean escape tracking with a loop that consumes the full escape sequence when `\x1b` is encountered. Recognize `[` after `\x1b` as the CSI introducer instead of a final byte, and keep the wrapper open until the actual final byte (any character in `@`..=`~` that is not `[`).

This would also correctly handle two-character escapes like `\x1bM` (reverse line feed), where the second character is the final byte.

#### Environment
- Starship version: 1.25.0
- bash version: GNU bash, version 5.3.9(1)-release (x86_64-pc-linux-gnu)

- Operating system: NixOS 26.5.0
- Git Commit Hash: b426db92
- Branch/Tag: master
- Rust Version: rustc 1.94.1 (e408947bf 2026-03-25) (built from a source tarball)
- Rust channel:  release

```bash
# Commands that should be applied only for interactive shells.
[[ $- == *i* ]] || return

HISTFILESIZE=100000
HISTSIZE=10000

shopt -s histappend
shopt -s extglob
shopt -s globstar
shopt -s checkjobs



if [[ ! -v BASH_COMPLETION_VERSINFO ]]; then
  . "/nix/store/8lpwn6d9k2j1dwdgkidhjqjy3imfw49f-bash-completion-2.17.0/etc/profile.d/bash_completion.sh"
fi

if [[ -n "${GHOSTTY_RESOURCES_DIR}" ]]; then
  builtin source "${GHOSTTY_RESOURCES_DIR}/shell-integration/bash/ghostty.bash"
fi

GPG_TTY="$(tty)"
export GPG_TTY
/nix/store/k4d44dcjlqp3xr849x985myp0slfhp9n-gnupg-2.4.9/bin/gpg-connect-agent --quiet updatestartuptty /bye > /dev/null

eval "$(/nix/store/0k10ldsna7n046589r067579z9j5651k-direnv-2.37.1/bin/direnv hook bash)"
```
#### Starship Configuration

This is the [Tokyo Night preset from the docs](https://starship.rs/presets/tokyo-night).

```toml

format = """
[░▒▓](#a3aed2)\
[  ](bg:#a3aed2 fg:#090c0c)\
[](bg:#769ff0 fg:#a3aed2)\
$directory\
[](fg:#769ff0 bg:#394260)\
$git_branch\
$git_status\
[](fg:#394260 bg:#212736)\
$nodejs\
$rust\
$golang\
$php\
[](fg:#212736 bg:#1d2230)\
$time\
[ ](fg:#1d2230)\
\n$character"""

[directory]
style = "fg:#e3e5e5 bg:#769ff0"
format = "[ $path ]($style)"
truncation_length = 3
truncation_symbol = "…/"

[directory.substitutions]
"Documents" = "󰈙 "
"Downloads" = " "
"Music" = " "
"Pictures" = " "

[git_branch]
symbol = ""
style = "bg:#394260"
format = '[[ $symbol $branch ](fg:#769ff0 bg:#394260)]($style)'

[git_status]
style = "bg:#394260"
format = '[[($all_status$ahead_behind )](fg:#769ff0 bg:#394260)]($style)'

[nodejs]
symbol = ""
style = "bg:#212736"
format = '[[ $symbol ($version) ](fg:#769ff0 bg:#212736)]($style)'

[rust]
symbol = ""
style = "bg:#212736"
format = '[[ $symbol ($version) ](fg:#769ff0 bg:#212736)]($style)'

[golang]
symbol = ""
style = "bg:#212736"
format = '[[ $symbol ($version) ](fg:#769ff0 bg:#212736)]($style)'

[php]
symbol = ""
style = "bg:#212736"
format = '[[ $symbol ($version) ](fg:#769ff0 bg:#212736)]($style)'

[time]
disabled = false
time_format = "%R" # Hour:Minute Format
style = "bg:#1d2230"
format = '[[  $time ](fg:#a0a9cb bg:#1d2230)]($style)'
```

zmberber:
#### Description

`wrap_colorseq_for_shell` now correctly wraps all CSI escape sequences in shell-specific markers (`\[...\]` for bash, `%{...%}` for zsh), not just SGR sequences ending in `m`.

The old code opened a wrapper on `\x1b` and closed it on `m`. But `[` (the CSI introducer in every `\x1b[` sequence) has byte value 0x5B, which falls in `@`..=`~` -- the range of CSI final bytes. The code had no special handling for it, so for something like `\x1b[K`:

1. `\x1b` -> open wrapper
2. `[` -> mistaken for final byte, close wrapper
3. `K` -> emitted as literal outside the wrapper

The fix replaces the boolean escape tracking with a loop that consumes the full sequence, recognizing `[` after `\x1b` as the CSI introducer.

#### Motivation and Context

Closes #7435

Many terminal emulators have a bug where background colors bleed on soft-wrapped lines in powerline-style prompts. The standard workaround is a custom module emitting `\x1b[49m\x1b[K`, but because `wrap_colorseq_for_shell` didn't handle `K`-terminated sequences, this broke readline's width calculation and caused input to overwrite the prompt instead of wrapping.

See #7435 for screenshots of both bugs.

Verified with `starship prompt | od -c` that `\x1b[K` is now properly wrapped inside `\[...\]`. Tested that long input lines wrap correctly and that the erase_line workaround no longer breaks the prompt.

#### Checklist:

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -581,22 +581,12 @@ CMake suite maintained and supported by Kitware (kitware.com/cmake).\n",
     Some(out)
 }
 
-/// Wraps ANSI color escape sequences in the shell-appropriate wrappers.
+/// Many shells cannot deal with raw unprintable characters and miscompute the
+/// cursor position, leading to strange visual bugs like duplicated/missing
+/// chars. This function wraps ANSI escape sequences in shell-specific markers
+/// to avoid these problems. Handles all CSI sequences (\x1b[...X), not just
+/// SGR (\x1b[...m).
 pub fn wrap_colorseq_for_shell(ansi: String, shell: Shell) -> String {
-    const ESCAPE_BEGIN: char = '\u{1b}';
-    const ESCAPE_END: char = 'm';
-    wrap_seq_for_shell(ansi, shell, ESCAPE_BEGIN, ESCAPE_END)
-}
-
-/// Many shells cannot deal with raw unprintable characters and miscompute the cursor position,
-/// leading to strange visual bugs like duplicated/missing chars. This function wraps a specified
-/// sequence in shell-specific escapes to avoid these problems.
-pub fn wrap_seq_for_shell(
-    ansi: String,
-    shell: Shell,
-    escape_begin: char,
-    escape_end: char,
-) -> String {
     let (beg, end) = match shell {
         // \[ and \]
         Shell::Bash => ("\u{5c}\u{5b}", "\u{5c}\u{5d}"),
@@ -605,24 +595,28 @@ pub fn wrap_seq_for_shell(
         _ => return ansi,
     };
 
-    // ANSI escape codes cannot be nested, so we can keep track of whether we're
-    // in an escape or not with a single boolean variable
-    let mut escaped = false;
-    let final_string: String = ansi
-        .chars()
-        .map(|x| {
-            if x == escape_begin && !escaped {
-                escaped = true;
-                format!("{beg}{escape_begin}")
-            } else if x == escape_end && escaped {
-                escaped = false;
-                format!("{escape_end}{end}")
-            } else {
-                x.to_string()
+    // ANSI escape codes cannot be nested, so when we hit \x1b we can just
+    // consume characters until the sequence ends.
+    let mut result = String::new();
+    let mut chars = ansi.chars().peekable();
+    while let Some(c) = chars.next() {
+        if c == '\x1b' {
+            result.push_str(beg);
+            result.push('\x1b');
+            while let Some(&next) = chars.peek() {
+                chars.next();
+                result.push(next);
+                // '[' is the CSI introducer, not a final byte
+                if next != '[' && ('@'..='~').contains(&next) {
+                    break;
+                }
             }
-        })
-        .collect();
-    final_string
+            result.push_str(end);
+        } else {
+            result.push(c);
+        }
+    }
+    result
 }
 
 fn internal_exec_cmd<T: AsRef<OsStr> + Debug, U: AsRef<OsStr> + Debug>(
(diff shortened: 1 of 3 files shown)


src/utils/mod.rs
```diff
@@ -605,24 +595,28 @@ pub fn wrap_seq_for_shell(
...
+    let mut chars = ansi.chars().peekable();
+    while let Some(c) = chars.next() {
+        if c == '\x1b' {
+            result.push_str(beg);
+            result.push('\x1b');
+            while let Some(&next) = chars.peek() {
+                chars.next();
+                result.push(next);
+                // '[' is the CSI introducer, not a final byte
+                if next != '[' && ('@'..='~').contains(&next) {
```
davidkna: If you're going this far maybe also handle OSC sequences.
I would also consider moving to a proper state machine to avoid the nested loops.

zmberber: ah i actually first wrote a state machine, i should have chosen that.

any way, yeah we could also handle OSC sequences.  should i do a PR that includes this?  it might take a while though, i am not experienced with this stuff (let alone rust).

src/utils/mod.rs
```diff
@@ -581,22 +581,17 @@ CMake suite maintained and supported by Kitware (kitware.com/cmake).\n",
...
+/// cursor position, leading to strange visual bugs like duplicated/missing
+/// chars. This function wraps ANSI escape sequences in shell-specific markers
+/// to avoid these problems.
+///
+/// Handles:
+/// - CSI sequences: `\x1b[...<final>` (terminated by final byte 0x40–0x7E)
+/// - String-terminated sequences (OSC/DCS/SOS/PM/APC): `\x1b]...<term>` etc.,
+///   terminated by BEL (0x07) or the two-byte String Terminator ST (ESC \)
+/// - Other two-byte ESC sequences: `\x1b<X>` (single char after ESC)
 pub fn wrap_colorseq_for_shell(ansi: String, shell: Shell) -> String {
```
davidkna, line 594: The updated state machine looks reasonable, but please ensure that tests still pass.

PR #7438, opened 2026-04-29
feat(cmd_duration): add variables for custom formatting duration

### LINKED ISSUE #5832: `cmd_duration`: Introduce Variables
There are different possible preferences regarding the formatting of a duration, which are currently not supported:

- Seconds with a decimal points
- Raw seconds
- Only the biggest one or two duration components
- ....

Other modules solve this by having variables that can be used in a format string. The `cmd_duration` module only offers a single pre-formatted variable though. I propose to introduce variables for the different possible components like:

- Raw days / hours / minutes / seconds  (with thousand separators)
- "Modulo" days / hours / minutes / seconds including `0`s
- "Modulo" days / hours / minutes / seconds with `0`s being empty strings

This would allow a lot of reasonable configs for different preferences.

I tried adding more single config options to support those cases, but it is hard to do that without being convoluted. Config could still be useful for setting things like decimal digits, thousand separators, including nulls etc for the variables mentioned above.

I would be happy to implement this.

mkrasnitski:
#### Description
Introduces new variables `duration_in_{milliseconds,seconds,minutes,hours,days}` for truncating the total execution time to a desired precision level, as well as `duration_modulo_{milliseconds,seconds,minutes,hours}` variables which contain non-overflowing values that allow for easy custom formatting.

Additionally, introduces a `show_zeros` config option which controls whether `duration_*` variables containing leading zeros should be left blank or not.

#### Motivation and Context

Closes #5832

#### Checklist:


davidkna:
I would rather not handle this that way.

mkrasnitski:
Do you have any suggestions or preferences for an alternative? My own personal use case for this is to delimit the formatted duration with spaces (e.g. `2h 31m 50s`); would an option to control this be something you'd approve, or are you looking for a better more general formatting solution?

PR #7449, opened 2026-05-01
fix: use cargo-zigbuild for riscv64gc-unknown-linux-musl release builds

gounthar:
#### Description

Switch the `riscv64gc-unknown-linux-musl` release target from `cross` to `cargo-zigbuild`.

`cross` 0.2.5's Docker image for `riscv64gc-unknown-linux-musl` does not set up the musl cross-linker correctly: when the build runs, `/usr/bin/ld` (the x86_64 host linker) is invoked instead of an riscv64 musl linker, producing a `file in wrong format` linker error. The cross job exits with code 0 because the error goes to stderr and the build step doesn't catch it, so releases are published without the riscv64 artifact.

`cargo-zigbuild` uses Zig's built-in riscv64 musl toolchain as a drop-in cross-linker. No custom Docker image or linker script needed.

Changes:
- Add `use_zigbuild: true` flag to the `riscv64gc-unknown-linux-musl` matrix entry
- Conditionalise the existing `cross` setup/build steps so they only run when `use_zigbuild` is false
- Add Zig setup (`mlugg/setup-zig@v2.2.1`) and `cargo-zigbuild` install steps gated on `use_zigbuild`
- Add a `Build [Zigbuild]` step for the riscv64 target

All other targets are unaffected.

#### Motivation and Context

Closes #

riscv64 binaries have been silently missing from releases since v1.25.0 (after the riscv64 target was added in #7337). This fixes the linker setup so the artifact actually gets built and uploaded.

N/A — CI build artifact and release asset list.

Pushed the branch to my fork and verified the `riscv64gc-unknown-linux-musl` job completes via the fork's CI.

#### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -114,9 +114,7 @@ jobs:
 
       - name: Setup | Install Zig [riscv64]
         if: matrix.use_zigbuild
-        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
-        with:
-          version: 0.13.0
+        run: pip install ziglang==0.13.0
 
       - name: Setup | Install cargo-zigbuild [riscv64]
         if: matrix.use_zigbuild
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,6 +58,7 @@ jobs:
           - target: riscv64gc-unknown-linux-musl
             os: ubuntu-latest
             name: starship-riscv64gc-unknown-linux-musl.tar.gz
+            use_zigbuild: true
 
           - target: x86_64-apple-darwin
             os: macos-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -106,19 +107,35 @@ jobs:
...
+        if: matrix.os == 'ubuntu-latest' && !matrix.use_zigbuild
         uses: taiki-e/install-action@cca35edeb1d01366c2843b68fc3ca441446d73d3 # v2.77.1
         with:
           tool: cross@0.2.5
 
+      - name: Setup | Install Zig [riscv64]
+        if: matrix.use_zigbuild
+        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
+        with:
+          version: 0.13.0
```
davidkna: I would prefer installing zig via pip. Please also use a `--version` flag and add a custom `renovate`-manager to handle updates.

davidkna: I feel that it's also worth investigating fixing the cross-rs image.

gounthar: @davidkna Done. Swapped the setup-zig action for the `ziglang` PyPI package pinned at 0.13.0 (`pip install ziglang==0.13.0`); cargo-zigbuild picks it up via `python3 -m ziglang`. Added a Renovate custom manager in `.github/renovate.json5` that tracks that pin against PyPI, matching the existing workflow-tool managers.

On the `--version` flag: I read that as pinning via `ziglang==0.13.0`. Did you mean that, or pinning cargo-zigbuild too, or a `python -m ziglang version` check? I'll adjust to match.

Re cross-rs: agreed it's worth a look. That's really why this PR exists though, since the riscv64 cross image is the blocker and fixing it upstream is a bigger separate effort. I can open a follow-up issue to track that so it doesn't get lost, and keep zigbuild as the unblock here.

davidkna (approved):

davidkna:
Thank you for the fix @gounthar!

PR #7481, opened 2026-05-16
chore: add ai policy

davidkna:
#### Description

Given recent PRs by OpenClaw-style agents, I feel like an AI policy for the project is in order.

#### Motivation and Context

Closes #

#### How Has This Been Tested?

- [ ] I have tested using **MacOS**
- [ ] I have tested using **Linux**
- [ ] I have tested using **Windows**

#### Checklist:

- [ ] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly.

PR #7483, opened 2026-05-17
feat(prompt): add OSC 133 semantic prompt support

### LINKED ISSUE #5463: Add support for semantic prompts specification
It would be great if starship could send OSC 133 escape codes.

Full specification can be found here: https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md

Modern terminal emulators support this, such as popular terminal emulators such as:

- [iTerm2](https://iterm2.com/documentation-escape-codes.html)
- [WezTerm](https://github.com/wez/wezterm/issues/115)
- [Alacritty](https://github.com/alacritty/alacritty/issues/5850)
- [Kitty](https://sw.kovidgoyal.net/kitty/shell-integration/#how-it-works)
- [VS Code](https://code.visualstudio.com/docs/terminal/shell-integration)

Add configuration option/module that when enabled sends corresponding request codes.

Manually hacking the `PS1` and `PS2` to include relevant escape codes.

----

I found that it could be done as follows (though I was unable to confirm that it works in my environment):

```zsh
    function __prompt_precmd() {
        local ret="$?"
        if test "$_prompt_executing" != "0"
        then
          _PROMPT_SAVE_PS1="$PS1"
          _PROMPT_SAVE_PS2="$PS2"
          PS1=$'%{\e]133;P;k=i\a%}'$PS1$'%{\e]133;B\a\e]122;> \a%}'
          PS2=$'%{\e]133;P;k=s\a%}'$PS2$'%{\e]133;B\a%}'
        fi
        if test "$_prompt_executing" != ""
        then
           printf "\033]133;D;%s;aid=%s\007" "$ret" "$$"
        fi
        printf "\033]133;A;cl=m;aid=%s\007" "$$"
        _prompt_executing=0
    }
    function __prompt_preexec() {
        PS1="$_PROMPT_SAVE_PS1"
        PS2="$_PROMPT_SAVE_PS2"
        printf "\033]133;C;\007"
        _prompt_executing=1
    }
    preexec_functions+=(__prompt_preexec)
    precmd_functions+=(__prompt_precmd)
```
https://gitlab.freedesktop.org/Per_Bothner/specifications/-/blob/master/proposals/prompts-data/shell-integration.zsh

ghostbuster91:
#### Description

 Adds OSC 133 (semantic prompts) emission, gated behind a new top-level `osc_133` config option (default `false`).

When enabled, `starship prompt` emits:
 - `OSC 133;D;<exit>` - previous command finished, with exit status (when known)
 - `OSC 133;A` - prompt start (`A;k=s` for continuation prompts)
 - `OSC 133;B` - end of prompt / start of user input

~and the bash/zsh/fish init scripts emit `OSC 133;C` from their preexec hooks to mark the start of command output. The bash 4.4+ PS0 fallback path (which bypasses `starship_preexec`) injects the C marker directly into `PS0`.~ (dropped as requested)

Sequences are wrapped in `\[\]` / `%{%}` for bash / zsh / tcsh so width calculation stays correct (the issue that broke #3018 and #5381). `Target::Right` is skipped since the right prompt isn't the input cursor position. The `osc_133` flag is read at `starship init <shell>` time, so users need to re-source their shell after flipping it.

Other shells (powershell, nu, elvish, ion, tcsh, xonsh, cmd) don't get the C marker in this PR - their preexec story varies and is left for a follow-up. They still get A/B/D from `starship prompt`.

Per @WhyNotHugo's suggestion in the issue, the A/B/D logic lives in the prompt renderer, not in shell init scripts.

#### Motivation and Context

Modern terminals (VS Code, iTerm2, kitty, WezTerm, Windows Terminal, foot, KDE Konsole, ...) use OSC 133 to enable prompt-to-prompt navigation, click-to-position cursor in the input area, exit-status indicators in the gutter, and selecting command output as a unit.

Closes #5463

#### Checklist:

#### AI-Assistance

Per the AI Policy proposed in #7481.

- [x] Yes
- [ ] No

Scope: Claude Code (Opus) was used for implementation drafting, test authoring, and explaining the OSC 133 spec / shell-escape semantics. The design (where to emit each marker, scope of C in this PR, deferral of non-preexec shells) and the structural feedback were human-directed. I have read every line of the diff and can explain each escape sequence, the shell wrapping (`\[\]` / `%{%}`), and the bash PS0 prompt-escape form (`\[\e]133;C\a\]`).

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -35,6 +35,10 @@
       "type": "boolean",
       "default": true
     },
+    "osc_133": {
+      "type": "boolean",
+      "default": false
+    },
     "follow_symlinks": {
       "type": "boolean",
       "default": true
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -210,20 +210,26 @@ This is the list of prompt-wide configuration options.
 
 ### Options
 
-| Option            | Default                        | Description                                                                                                                                                                        |
-| ----------------- | ------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `format`          | [link](#default-prompt-format) | Configure the format of the prompt.                                                                                                                                                |
-| `right_format`    | `''`                           | See [Enable Right Prompt](../advanced-config/#enable-right-prompt)                                                                                                                 |
-| `scan_timeout`    | `30`                           | Timeout for starship to scan files (in milliseconds).                                                                                                                              |
-| `command_timeout` | `500`                          | Timeout for commands executed by starship (in milliseconds).                                                                                                                       |
-| `add_newline`     | `true`                         | Inserts blank line between shell prompts.                                                                                                                                          |
-| `palette`         | `''`                           | Sets which color palette from `palettes` to use.                                                                                                                                   |
-| `palettes`        | `{}`                           | Collection of color palettes that assign [colors](../advanced-config/#style-strings) to user-defined names. Note that color palettes cannot reference their own color definitions. |
-| `follow_symlinks` | `true`                         | Follows symlinks to check if they're directories; used in modules such as git.                                                                                                     |
+| Option            | Default                        | Description                                                                                                                                                                                                      |
+| ----------------- | ------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `format`          | [link](#default-prompt-format) | Configure the format of the prompt.                                                                                                                                                                              |
+| `right_format`    | `''`                           | See [Enable Right Prompt](../advanced-config/#enable-right-prompt)                                                                                                                                               |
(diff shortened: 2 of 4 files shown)


davidkna (changes requested):
I would prefer not to add additional code to the init script at all.

ghostbuster91:
Sure thing. Unfortunately `;C` marks "start of command output", which fires between the user pressing Enter and the command running - starship itself never sees that moment, so a shell preexec hook is the only mechanism available.

The `;A/;B/;D` markers don't have this constraint - they're emitted entirely from starship prompt with no init-script changes. Should I just drop the `;C` mark support from the PR?

ghostbuster91:
I updated the PR by dropping the support for `;C` mark.

PR #7493, opened 2026-05-21
fix(bash): clear status after empty command

### LINKED ISSUE #7450: [Bash] Exit status does not clear on empty command (for the `status` module and for success/failure symbol from `character` module)

puneetdixit200:
#### Description
This clears the status and pipestatus values passed to `starship prompt` when Bash renders a prompt after an empty command line. That makes pressing Enter on an empty prompt behave like a successful command for the `status` and `character` modules, matching the behavior already used to skip `cmd_duration`.

For the older DEBUG-trap path, this also avoids treating `starship_precmd` itself as a preexec event. Without that guard, Bash 3.2 immediately started a new timer while drawing the prompt, so the empty prompt still received the previous failed status.

#### Motivation and Context
Closes #7450

- Reproduced the issue with an interactive Bash harness before the fix: after `false`, an empty command still sent `--status=1` to `starship prompt`.
- Verified the same harness after the fix with `/bin/bash` 3.2.57 and Homebrew Bash 5.3.9: the empty command sends empty `--status` and `--pipestatus`.
- `CARGO_HOME="$PWD/.cargo-home" CARGO_TARGET_DIR="$PWD/target" cargo fmt --check`
- `CARGO_HOME="$PWD/.cargo-home" CARGO_TARGET_DIR="$PWD/target" cargo test init::tests::bash_empty_command_clears_status_args`
- `CARGO_HOME="$PWD/.cargo-home" CARGO_TARGET_DIR="$PWD/target" cargo test init::`
- `CARGO_HOME="$PWD/.cargo-home" CARGO_TARGET_DIR="$PWD/target" cargo clippy --all-targets --all-features -- -D warnings`
- `PATH="/opt/homebrew/bin:$PATH" TERM=xterm-256color CARGO_HOME="$PWD/.cargo-home" CARGO_TARGET_DIR="$PWD/target" cargo test`

#### Checklist:

### DIFF
--- a/src/init/mod.rs
+++ b/src/init/mod.rs
@@ -282,6 +282,9 @@ const CMDEXE_INIT: &str = include_str!("starship.lua");
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::io::Write;
+    use std::process::Stdio;
+
     #[test]
     fn escape_pwsh() -> io::Result<()> {
         let starship_path = StarshipPath {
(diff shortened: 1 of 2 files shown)


src/init/mod.rs
davidkna, line 1: Please remove this.

PR #7495, opened 2026-05-22
fix(format): include parse error details when format string fails to parse

### LINKED ISSUE #7491: Improve error message: "Error parsing format string" should suggest escaping $
## Summary

When a `$` literal appears in a format string without escaping (e.g. `stashed = " ${count}$"`), the error message is:

```
[WARN] - (starship::modules::git_status): Error parsing format string `git_status.stashed`
```

This is technically correct but not actionable — it doesn't tell the user *why* parsing failed or *how* to fix it.

## Suggested improvement

Include the parse error reason and a hint, e.g.:

```
[WARN] - (starship::modules::git_status): Error parsing format string `git_status.stashed`:
  unexpected end of input after `$` at position N.
  Hint: use `\$` to display a literal `$` character.
```

## Reproduction

```toml
[git_status]
stashed = " ${count}$"   # broken: bare $ at end
stashed = " ${count}\$"  # correct: escaped $
```

starship 1.25.1 on Windows (git bash / PowerShell).

## Why

This is a common mistake — `$` is the variable prefix in starship format strings, but the current error gives no clue about the root cause. A one-line hint would save users a web search.

Thanks for the great tool!

gaurav0107:
#### Description

Two log sites in `src/modules/git_status.rs` (`format_text`) and
`src/modules/status.rs` (the `status.format` branch of the `module`
function) previously dropped the underlying `StringFormatterError` when
`StringFormatter::new` failed:

- `src/modules/git_status.rs:format_text` used
  `if let Ok(formatter) = StringFormatter::new(format_str) { ... } else {
  log::warn!("Error parsing format string \`{}\`", config_path); ... }`,
  with no binding for the error.
- `src/modules/status.rs` for `status.format` used `Err(_error)` and
  emitted only `"Error parsing format string in \`status.format\`"`.

A third site (`status.rs:84`, the `pipestatus_segment_format` branch)
already included the error via `{e:?}`. This change aligns all three
sites to use `Display` formatting (`{e}` / `{error}`), so the log
message includes pest's own multi-line renderer with a position
indicator and the rules it expected to see.

For a config like:

```toml
[git_status]
stashed = " ${count}$"   # bare $ at end — common mistake
```

the warn line goes from:

```
[WARN] - (starship::modules::git_status): Error parsing format string `git_status.stashed`
```

to:

```
[WARN] - (starship::modules::git_status): Error parsing format string `git_status.stashed`:
 --> 1:9
  |
1 |  ${count}$
  |          ^---
  |
  = expected variable_name or variable_scope
```

which makes it obvious that the trailing `$` is the problem and that the
fix is to escape it (`\$`).

A new test, `test_parse_error_display_includes_context`, asserts that the
`Display` rendering of a parse error names one of the `variable_*` rules,
so log readers can always rely on the diagnostic carrying actionable
context.

#### Motivation and Context

Closes #7491.

The original report describes a config like `stashed = " ${count}$"`
producing only the bare `Error parsing format string` line, with no clue
about the offending position or the cause. The pest renderer attached to
`StringFormatterError::Parse` already produces a position-aware
diagnostic; we just weren't surfacing it. This is a small, low-risk
diagnostic improvement.

N/A — log output only.

`cargo fmt --all -- --check`, `cargo clippy --workspace --locked --
-D warnings`, `cargo check --workspace --locked`, and
`cargo test --workspace --locked formatter::string_formatter` all pass
locally on macOS (aarch64). Linux/Windows coverage relies on CI.

#### Checklist:

- [x] I have updated the documentation accordingly. (No user-facing docs
  affected — the change is in internal log messages.)

### DIFF
--- a/src/formatter/string_formatter.rs
+++ b/src/formatter/string_formatter.rs
@@ -839,6 +839,26 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_error_display_includes_context() {
+        // Bare `$` at end of format — common user mistake (cf. issue #7491).
+        // The Display impl forwards to pest's renderer, which should mention
+        // both a position indicator and the rules it expected to see. We
+        // don't lock the exact rendering (pest may evolve), but any
+        // diagnostic worth showing the user must name a `variable_*` rule.
+        // (We avoid `unwrap_err`/`expect_err` because they require `Debug`
+        // on `StringFormatter`, which is not derived.)
+        let rendered = match StringFormatter::new(" ${count}$") {
+            Ok(_) => panic!("bare trailing `$` should fail to parse"),
+            Err(error) => format!("{error}"),
+        };
+        assert!(!rendered.is_empty(), "error display should not be empty");
+        assert!(
+            rendered.contains("variable_name") || rendered.contains("variable_scope"),
+            "expected pest error to name an expected rule, got:\n{rendered}"
+        );
+    }
+
+    #[test]
     fn test_variable_error() {
         const FORMAT_STR: &str = "$never$some";
         let never_error = StringFormatterError::Custom("NEVER".to_owned());
--- a/src/formatter/string_formatter.rs
+++ b/src/formatter/string_formatter.rs
@@ -839,34 +839,6 @@ mod tests {
     }
 
     #[test]
-    fn test_parse_error_display_includes_context() {
-        // Bare `$` at end of format — common user mistake (cf. issue #7491).
-        // The Display impl forwards to pest's renderer, which should mention
-        // both a position indicator and the rules it expected to see. We
-        // don't lock the exact rendering (pest may evolve), but any
-        // diagnostic worth showing the user must name a `variable_*` rule.
-        // (We avoid `unwrap_err`/`expect_err` because they require `Debug`
-        // on `StringFormatter`, which is not derived.)
-        let rendered = match StringFormatter::new(" ${count}$") {
-            Ok(_) => panic!("bare trailing `$` should fail to parse"),
-            Err(error) => format!("{error}"),
-        };
-        assert!(!rendered.is_empty(), "error display should not be empty");
-        assert!(
-            rendered.contains("variable_name") || rendered.contains("variable_scope"),
-            "expected pest error to name an expected rule, got:\n{rendered}"
-        );
-    }
-
-    #[test]
-    fn test_custom_error_display_passes_message_through() {
-        // The other variant log sites might encounter: a Custom error built
-        // from a String. Display should round-trip the message verbatim.
-        let error = StringFormatterError::Custom("boom".to_owned());
-        assert_eq!(format!("{error}"), "boom");
-    }
-
-    #[test]
     fn test_variable_error() {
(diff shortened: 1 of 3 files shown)


src/formatter/string_formatter.rs
```diff
@@ -836,6 +836,34 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_custom_error_display_passes_message_through() {
+        // The other variant log sites might encounter: a Custom error built
+        // from a String. Display should round-trip the message verbatim.
+        let error = StringFormatterError::Custom("boom".to_owned());
+        assert_eq!(format!("{error}"), "boom");
+    }
+
```
davidkna: These tests are superfluous:

```suggestion
```

gaurav0107: Good call — removed in ffabe1b1. Both tests were really exercising upstream behavior (pest's renderer and the derived `Display` on `Custom`) rather than starship code, so dropping them keeps the test surface focused on the actual log-site change.

Heads up: `codecov/patch` was already failing on this PR (71.42% vs 93.68% target) because the new `Err` arms in `git_status.rs` / `status.rs` aren't hit by existing module tests. Removing these doesn't change that — happy to add a small integration-style test that drives a malformed `git_status.stashed` through the module if you'd prefer that route, otherwise I'll leave the patch coverage as-is.

gaurav0107:
Rebased onto the latest `main` and added the patch coverage that `codecov/patch` was flagging (it was at ~71%, below the 93.68% target, because the new error arms weren't exercised).

Two small module tests now drive the parse-error paths this PR introduces:

- `git_status`: a malformed count sub-format (`staged = "${"`) is dropped and the module renders nothing instead of aborting — covers the new `Err` arm in `format_text`.
- `status`: a malformed top-level `format` logs the parse error and returns `None` — covers the new `Err` arm in `module`.

Both exercise the actual log sites and the graceful-degradation behavior rather than pest's renderer or a derived `Display` impl, so they restore regression coverage for the parsing-error handling without re-adding the upstream-behavior tests removed earlier. The `pipestatus_segment_format` arm was already covered by the existing `pipestatus_segment_format_err` test.

`cargo fmt --check` and `cargo clippy --workspace --locked -- -D warnings` are clean locally.

gaurav0107:
The cubic P2 flag ("Display diagnostic tests removed, reducing protection") was raised against an earlier commit (`ffabe1b1`) and is addressed on the current head (`797731c2`):

- The two removed tests only asserted **upstream** behavior — pest's error renderer and the derived `Display` on the `Custom` variant — and were dropped at @davidkna's explicit request ("these tests are superfluous"), not starship code.
- Replacement regression coverage was added in `797731c2` that drives the **actual starship code this PR changes** — the `StringFormatter::new` error arms in `git_status::format_text` and `status::module`:
  - `modules::git_status::tests::parse_error_in_count_format_is_skipped`
  - `modules::status::tests::format_parse_error_renders_nothing`

These exercise the graceful-degradation + parse-error-logging paths directly, so the protection cubic was concerned about is restored (and `codecov/patch` is green on the current head). A cubic re-review on `797731c2` should clear this.

PR #7528, opened 2026-05-27
fix(elixir): try --short-version first to avoid slow BEAM VM startup

### LINKED ISSUE #3867: Allow Elixir module to use `elixir --short-version` because `elixir --version` is very slow

Requesting the Elixir version can take upwards of 700ms, because `elxiir --version` starts the entire VM.  This can cause the prompt to frequently timeout or seem very slow.

Since Elixir v 1.13.0 there is a `--short-version` switch.  This doesn't start the VM and returns results in less than 10ms, but it can only show the full version of Elixir and not OTP.  It would be nice if there was a way to make use of this.

The most complex but exhaustive option would be:
 ```
   if format configuration does NOT contain "$otp_version"
   and the elixir value from the mix.exs file (e.g. elixir: "~>1.13") and that version 1.13 or greater
   then use the `--short-version` command
```

Alternatively simply providing a configuration switch that uses `elxiir --short-version`, but that would have to come with caveats that it will fail on versions older than 1.13 and will not provide the OTP version.

### PR DESCRIPTION
**Pull request by dawitlabs**
## Summary

Closes #3867.

`elixir --version` starts the full Erlang VM (~370 ms). Since Elixir 1.13, the `elixir` shell script handles `--short-version` entirely before launching the VM, printing just the bare version string (e.g. `1.13.4`) in ~9 ms.

## Changes

**`src/modules/elixir.rs`**
- `get_elixir_version` now tries `elixir --short-version` first. On success it returns the Elixir version with `otp_version = None`.
- Falls back to `elixir --version` if the flag is absent or returns unexpected output (Elixir < 1.13 or unusual installs), restoring full OTP + version info.
- New `parse_elixir_short_version` validates the output starts with a digit (guards against the pre-1.13 behaviour where unrecognised flags still boot the VM and print the full banner).
- New `test_with_mix_file_version_fallback` test exercises the fallback path by disabling the `--short-version` mock.
- New `test_parse_elixir_short_version` unit tests.

**`src/configs/elixir.rs`**
- Default format changed to place each segment in its own optional group:
  ```
  before: [$symbol($version \(OTP $otp_version\) )]($style)
  after:  [$symbol($version )(\(OTP $otp_version\) )]($style)
  ```
  Visually identical when OTP is present; the OTP group is independently suppressed on the fast path.

**`src/utils/mod.rs`**
- Added `"elixir --short-version"` mock returning `"1.13.4\n"` so the existing `test_with_mix_file` exercises the fast path by default.

## Test results

```
test modules::elixir::tests::test_parse_elixir_short_version ... ok
test modules::elixir::tests::test_parse_elixir_version ... ok
test modules::elixir::tests::test_without_mix_file ... ok
test modules::elixir::tests::test_with_mix_file ... ok
test modules::elixir::tests::test_with_mix_file_version_fallback ... ok

test result: ok. 5 passed; 0 failed
```

### DIFF
--- a/src/modules/elixir.rs
+++ b/src/modules/elixir.rs
@@ -51,8 +51,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "otp_version" => versions
                     .deref()
                     .as_ref()
-                    .map(|(otp_version, _)| otp_version.to_string())
-                    .map(Ok),
+                    .and_then(|(otp_version, _)| otp_version.as_deref())
+                    .map(|otp| Ok(otp.to_string())),
                 _ => None,
             })
             .parse(None, Some(context))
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/modules/elixir.rs [hunk 1]
```diff
@@ -51,8 +51,8 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                 "otp_version" => versions
                     .deref()
                     .as_ref()
-                    .map(|(otp_version, _)| otp_version.to_string())
-                    .map(Ok),
+                    .and_then(|(otp_version, _)| otp_version.as_deref())
```
davidkna, line 54: Please remove `$otp_version` from the default format string and ensure that `--version` still runs if the format string contains `$otp_version`.

PR #7547, opened 2026-06-11
fix!: replace PUA default symbols with standard Unicode characters

### LINKED ISSUE #2563: Make Nerd Fonts icons the default

zmberber:
This replaces all Private Use Area (PUA) default symbols in Starship with standard Unicode equivalents. PUA glyphs (Powerline , Devicons , Nerd Fonts 󰠅, etc.) only render with patched fonts and show up as tofu for everyone else.

This is a possibly controversial change, since it is technically breaking.

Though everything is customizable, in my opinion, things should "just work" out-of-the-box.  We should strive for sensible (I know this is highly subjective) and consistent defaults.

I see the following arguments:
- Though more common these days, PUA symbols have no standardized meaning, and therfore cross some some sort of line.  A terminal cannot fall back to another font for them.
- If we do assume nerdfonts as a requirement, which starship actually does at the moment, it seems like a rather strong requirement for just a handful (see below) of PUA symbols.
- Mixing PUA symbols like nerdfonts with emojis is visually inconsistent, since the former can be themed (change color etc.), and the latter not.
  - If we assume nerdfonts as a requirement, then we might as well use way more PUA symbols, like the actual OS symbols or actual programming language symbols. These are currently just emojis, presumably for less breakage, which doesn't make sense with the nerdfonts requirement.

The git_branch change is the biggest visible one. git_branch is enabled by default, so every git user immediately sees ⎇ instead of .

See also #3544, where a request by a starship organization member has been made to have that PR be fixed so that it can be merged.  I am basically doing this here.

---

Changes:

- git_branch, fossil_branch, hg_branch, pijul_channel:  -> ⎇
- nodejs:  -> ⬢
- erlang:  -> e
- pulumi:  -> ⠿
- azure: 󰠅 -> 🅰
- battery: 󰁹/󰂄/󰂃/󰂑/󰂎 -> •/⇡/⇣/?/!
- os (Void Linux):  -> ⊚

Also updates all module tests, the EN config docs, FAQ, README prerequisites, the generated config-schema.json, and deprecates the no-nerd-font preset (it is now redundant since defaults match it).

Fixes #2563


src/configs/hg_branch.rs
```diff
@@ -19,7 +19,7 @@ pub struct HgBranchConfig<'a> {
 impl Default for HgBranchConfig<'_> {
     fn default() -> Self {
         Self {
-            symbol: " ",
+            symbol: "⎇ ",
```
davidkna, line 22: I think in prior discussion we settled on keeping these, as e.g. currently many terminal provide the necessary symbols.

src/configs/battery.rs
```diff
@@ -22,11 +22,11 @@ pub struct BatteryConfig<'a> {
...
-            full_symbol: "󰁹 ",
-            charging_symbol: "󰂄 ",
-            discharging_symbol: "󰂃 ",
-            unknown_symbol: "󰂑 ",
-            empty_symbol: "󰂎 ",
+            full_symbol: "• ",
+            charging_symbol: "⇡ ",
+            discharging_symbol: "⇣ ",
+            unknown_symbol: "? ",
+            empty_symbol: "! ",
```
davidkna, line 29: Please prefer battery emoji. They can be less expressive but they have now been around for a bit and previously there was a lot of confusion and issues about what the module does without battery related symbols.

src/configs/azure.rs
```diff
@@ -20,7 +20,7 @@ impl Default for AzureConfig<'_> {
     fn default() -> Self {
         Self {
             format: "on [$symbol($subscription)]($style) ",
-            symbol: "󰠅 ",
+            symbol: "🅰 ",
```
davidkna, line 23: I think I would prefer the cloud emoji here (like aws/gcloud).

docs/presets/no-nerd-font.md
```diff
@@ -2,12 +2,13 @@
 
 # No Nerd Fonts Preset
 
-This preset restricts the use of symbols to those from emoji and powerline sets.
+::: warning DEPRECATED
```
davidkna, line 5: Please prefer GitHub-style admonitions:

 ```
> [!WARNING]
```

The `:::` syntax often caused issues with crowdin translations.
If we do end up keeping e.g. the powerline git symbol this could stay around to avoid powerline.

File: docs/faq/README.md [hunk 5]
```diff
@@ -112,17 +112,19 @@ particular do not come with font support out-of-the-box. You need to ensure that
...
 
 ```sh
 echo -e "\xf0\x9f\x90\x8d"
-echo -e "\xee\x82\xa0"
+echo -e "\xe2\x8e\x87"
 ```
 
 The first line should produce a [snake emoji](https://emojipedia.org/snake/),
-while the second should produce a [powerline branch symbol (e0a0)](https://github.com/ryanoasis/powerline-extra-symbols#glyphs).
+while the second should produce a [branch symbol (2387)](https://unicode-table.com/en/2387/).
```
davidkna, line 127: Please keep this as-is, as whether the powerline symbols is visible is more useful for debugging, though arguably it should be a "proper" nf symbol (even if not required many presets do use nf symbols).

src/configs/battery.rs
zmberber: That is a fair change, I can change that.

src/configs/azure.rs
zmberber: Sure, a cloud seems like a good choice, I can change that.

src/configs/hg_branch.rs
zmberber: Even if it is widely adopted, literally right here in my browser, also on mobile, and in my Email client, everywhere, the currently used symbol is broken, and my suggestion for a symbol is working.  So just for copy and pasting prompts elsewhere, like for notetaking or in forums or whatnot, it already is annoying.

With terminals, in my experience, whenever this symbol works, all other nerdfonts also work anyway.

I argue by principle, not by how widely they are adopted:  PUA symbols are inherently non-standard.  I repeat my argument:  Either we assume nerdfonts and PUA symbols, in which case we might as well use them for all other sybols as well, or we don't, and we stay consistent.  Right now we both have presets for "no nerdfonts" as well as "nerdfonts", which reflects my personal setup as well, which I don't think is right.

If we agree to all other changes, this would be the only PUA symbol left, which would be an odd single exception.

Sorry about arguing so strongly about such a detail.

docs/presets/no-nerd-font.md
zmberber: Nerdfonts include powerline, so the premise of the preset is already kind of slightly wrong in my opinion.  The info also suggests that there are multiple powerline symbols used all around starship, when in reality, there is only a single powerline character in use, being the git branch symbol.

zmberber: but of course, this is just the same discussion as above

src/configs/hg_branch.rs
zmberber: also, if you are referring to a prior discussion in my PR for the no-nerdfonts preset, I was convinced that since it was powerline, it was not nerdfont, I didn't check if what was said as a response to me was correct.  Also of course, the info mentioned powerline as being assumed with this preset, so it makes sense there.  But powerline does belong to the nerdfonts pack, and most importantly, the bigger point is that it is all PUA.

davidkna: Nerdfont is a collection of symbols from different sources, powerline is one of the font sets bundled in nerdfont patches.

Nerdfont patches have a significant file size impact for fonts, whereas powerline is fairly common in prompts etc. so recent fonts targeting programming environments include that - which is why I am currently still leaning towards keeping that symbol.

IIRC the preliminary decision in https://github.com/starship/starship/pull/3544 was to still include powerline symbols (the branch icon).

PR #7575, opened 2026-07-02
fix(fill): don't hang when the fill symbol is zero-width

chatman-media:
`FillSegment::ansi_string` cycles through the fill symbol's graphemes and stops once the accumulated width reaches the target. If every grapheme in the symbol has width 0 (a bare combining mark, a lone ZWJ, a variation selector...), the accumulated length never advances, so `.cycle()` spins forever and the whole prompt just hangs. `symbol = "‍"` in a `[fill]` config reproduces it directly.

Fix is to filter out zero-width graphemes before cycling — they can't contribute visible width anyway. As a side benefit this also stops symbols that mix zero-width and visible characters from leaking stray invisible bytes into the rendered fill.

Added a couple of unit tests around `FillSegment::ansi_string` plus an end-to-end one through `get_prompt`, all of which hang without the fix and pass with it.

<a href="https://cubic.dev/pr/starship/starship/pull/7575?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/segment.rs
+++ b/src/segment.rs
@@ -42,6 +42,12 @@ impl FillSegment {
             Some(w) => self
                 .value
                 .graphemes(true)
+                // Zero-width graphemes (e.g. a bare combining mark or ZWJ used as the fill
+                // symbol) never advance `len` below, so cycling over them would spin
+                // forever without ever reaching `w`. Drop them before cycling: they can't
+                // contribute visible width anyway, and this guarantees termination even
+                // if every grapheme in the symbol turns out to be zero-width.
+                .filter(|g| Grapheme(g).width() > 0)
                 .cycle()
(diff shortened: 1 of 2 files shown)


src/segment.rs
```diff
@@ -42,6 +42,12 @@ impl FillSegment {
             Some(w) => self
```
davidkna, line 42: I would prefer simply verifying that the whole string length is != 0 here.

PR #7612, opened 2026-07-21
feat: Add Jujutsu support to the `vcs` module

poliorcetics:
#### Description

[Jujutsu](https://docs.jj-vcs.dev/latest) is a VCS like Git, Mercurial and others.

This merge request adds support for it in the form of four (4) new modules and one modified one.

It was split into 5 MRs to be reviewable:

- `jj_bookmark`: finds the closest JJ bookmark amongst parent of the current change, filtering via some options, analogous to `git_branch`, see https://github.com/starship/starship/pull/7643
- `jj_change`: prints the current JJ change ID, and optionally the underlying commit ID, analogous to `git_commit`, see https://github.com/starship/starship/pull/7691
- `jj_metrics`: prints the current totals of added and removed lines, analogous to `git_metrics`, see https://github.com/starship/starship/pull/7720
- `jj_status`: prints the current change status, analogous to `git_status`, see https://github.com/starship/starship/pull/7729
- modifies the `vcs` module to add JJ support, this MR

The `vcs` module is modified to handle a `"jj"` VCS and adds the `jj_modules` option.
Note we **don't** put `jj` before `git` here because I expect the vast majority of Starship users are using Git and not JJ and I don't want them to pay the failed discovery every time.
Actual JJ users can use `order = ["jj", "git", "..."]` to look for JJ first.

#### Motivation and Context

#6388 was merged at the start of the year and the `vcs` module it added was explicitly introduced as solution to Git and JJ coexisting in the same space ([colocation](https://docs.jj-vcs.dev/latest/git-compatibility/#colocated-jujutsugit-repos) in JJ parlance), I just never got around to adding the JJ support I initially created it for.

- Closes #6968
- Closes #6957 (because the request was made to support JJ)
- Closes #6076
- Closes #5919 (because the request was made to support JJ)

#### Design notes

The `JJRepo` type uses a `OnceLock` internally because running a `jj` command is slow (10s of ms) so we only do it on demand.

I have optimized the command (notably the revision set and template) as much as I can (see the screenshot below) so that only one `jj` call is necessary and it still runs as fast as possible, with parsing made very easy by the template passed to it.

This leads to the last three modules in the screenshot being basically instant (< 1 ms) thanks to the `OnceLock` caching.

#### Screenshots (if appropriate):

[...]

PR #7613, opened 2026-07-21
feat: add jujutsu modules and vcs support

### LINKED ISSUE #6076: [FR] Support Jujutsu VCS (git backed)

101100:
#### Description

This adds support for Jujutsu to the `vcs` module along with 4 new modules:

- `jujutsu_bookmark` - For showing bookmarks on or before the current revision along with remote tracking information.
- `jujutsu_change` - For showing the current change ID and/or description.
- `jujutsu_commit` - For showing the current git commit
- `jujutsu_state` - For showing several on/off Jujutsu states: conflicted, empty, divergent, hidden, immutable and missing description.

This was built on #7224, with the first commit being the squashed and rebased changes from that PR. I pulled in code from https://github.com/dmmulroy/jj-starship to simplify some of the code and tidy it up. I've tried to add support to make most example configurations that I've seen possible.

#### Motivation and Context
Closes #6076 (the 5th time's the charm?)

Default config:
default config

Kitchen sink (enabling commit and bookmarks with defaults):
kitchen sink

#### AI-Assistance

- [ ] Yes
- [X] No

N/A

#### Checklist:
- [X] I understand and have read the code I contribute and can answer questions about it.

---

I will update the documentation and add reasonable defaults once I confirm that no other configuration values will change.

I am also willing to add some tests as recommended.


src/modules/jujutsu_change.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    // Only run in jj repositories
+    vcs::discover_repo_root(context, vcs::Vcs::Jujutsu)?;
+
+    let JujutsuChangeInfo {
+        change_id,
+        prefix_len,
+        description,
+    } = get_jujutsu_change_id(context)?;
+    let remaining_len = config.change_id_length.saturating_sub(prefix_len);
```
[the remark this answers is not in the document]
101100: If a longer ID is needed, then it should be printed, even if it is long, so I'll leave this unless a human agrees with you.

src/modules/jujutsu_change.rs
```diff
@@ -0,0 +1,86 @@
...
+                    .chars()
+                    .take(config.change_id_length)
+                    .collect::<String>())),
+                "change_prefix" => Some(Ok(change_id.chars().take(prefix_len).collect::<String>())),
+                "change_suffix" => Some(Ok(change_id
+                    .chars()
+                    .skip(prefix_len)
+                    .take(remaining_len)
+                    .collect::<String>())),
+                "description" if !description.is_empty() && config.description_limit > 0 => {
```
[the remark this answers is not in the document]
101100: I updated the utility method to trim the description.

src/modules/jujutsu_bookmark.rs
```diff
@@ -0,0 +1,155 @@
...
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(|variable| match variable {
+                "bookmark_name" => {
+                    let chars = bookmark.name.chars().collect::<Vec<_>>();
+                    if chars.len() > config.max_length {
```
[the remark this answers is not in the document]
101100: Added a check to ensure the length is at least 2.

Cargo.toml
```diff
@@ -35,17 +35,20 @@ config-schema = ["schemars"]
 notify = ["notify-rust"]
 
 [dependencies]
+chrono = "0.4"
 clap = { version = "4.6.2", features = ["derive", "cargo", "unicode"] }
 clap_complete = "4.6.7"
 clap_complete_nushell = "4.6.0"
 dirs = "6.0.0"
 dunce = "1.0.5"
+futures = "0.3"
```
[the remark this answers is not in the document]
101100: This was from an interim version; removed.

src/modules/jujutsu_commit.rs
```diff
@@ -0,0 +1,62 @@
...
+use crate::modules::vcs;
+
+/// Creates a module with the Jujutsu commit hash in the current directory
+///
+/// Will display the commit hash if the current directory is a jj repo
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("jujutsu_commit");
+    let config: JujutsuCommitConfig = JujutsuCommitConfig::try_load(module.config);
+
+    // We default to disabled=true, so we have to check after loading our config module.
```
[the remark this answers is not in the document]
101100: Removed these comments.

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,458 @@
...
+
+            for parent_id in commit.parent_ids() {
+                queue.push_back((parent_id.clone(), depth + 1));
+            }
+        }
+    }
+
+    // Convert to vec and sort by distance
+    let mut result = bookmarks_with_details.into_values().collect::<Vec<_>>();
+    result.sort_by_key(|details| details.distance);
```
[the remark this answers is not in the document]
101100: Good idea; added secondary sort by name.

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,458 @@
...
+        .resolve_change_id(commit.change_id())
+        .ok()
+        .and_then(|targets| targets.map(|targets| targets.is_divergent()))
+        .unwrap_or(false);
+    let empty = commit
+        .is_empty(repo.repo().as_ref())
+        .block_on()
+        .unwrap_or(false);
+    let immutable_heads = find_immutable_heads(repo.repo().view());
+    let immutable = immutable_heads.contains(&commit_id);
```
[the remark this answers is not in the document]
101100: Fixed by reverting to original implementation.

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,458 @@
...
+    workspace_root: PathBuf,
+    fileset_aliases: FilesetAliasesMap,
+    revset_aliases: RevsetAliasesMap,
+    revset_extensions: RevsetExtensions,
+}
+
+impl JujutsuRepo {
+    pub fn load(context: &Context<'_>) -> Option<Self> {
+        let workspace_root = vcs::discover_repo_root(context, vcs::Vcs::Jujutsu)?;
+        let settings = create_minimal_settings()?;
```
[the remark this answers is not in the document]
101100: I've updated the configuration loading to load system, user, repo and workspace settings.

101100:
Addressed initial automated review comments other than the need for tests. Looks like I need a minimal number of tests to pass the build, so I'll do that tomorrow.

src/modules/jujutsu_commit.rs
```diff
@@ -0,0 +1,57 @@
...
+    let mut module = context.new_module("jujutsu_commit");
+    let config: JujutsuCommitConfig = JujutsuCommitConfig::try_load(module.config);
+
+    if config.disabled {
+        return None;
+    }
+
+    let (commit_id, prefix_len) = get_jujutsu_commit_id(context)?;
+    let remaining_len = config.commit_hash_length.saturating_sub(prefix_len);
+
```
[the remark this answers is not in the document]
101100: Like with your [previous comment](https://github.com/starship/starship/pull/7613#discussion_r3625246043), if we need more characters to disambiguate, then I think we should print them, but I'm open to hearing a human argument for this change and then I'd apply it to both the commit hash and change ID.

src/modules/jujutsu_bookmark.rs
```diff
@@ -0,0 +1,162 @@
...
+            config.max_length,
+        );
+        default_value
+    } else {
+        config.max_length
+    };
+
+    let bookmarks = if jujutsu_bookmarks.is_empty() {
+        return None;
+    } else {
```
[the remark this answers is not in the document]
101100: I'll think about this one.

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,610 @@
...
+use jj_lib::revset;
+use jj_lib::revset::{
+    ResolvedRevsetExpression, RevsetAliasesMap, RevsetDiagnostics, RevsetExtensions,
+    RevsetParseContext, RevsetWorkspaceContext, SymbolResolver, UserRevsetExpression,
+};
+use jj_lib::secure_config::SecureConfig;
+use jj_lib::settings::UserSettings;
+use jj_lib::str_util::StringMatcher;
+use jj_lib::view::View;
+use jj_lib::workspace::{Workspace, default_working_copy_factories};
```
poliorcetics, line 25: JJ as a library means the user is locked in on compiled-in backends at starship build time, which e.g. will preclude all users at google from simply using starship without modifications

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,610 @@
...
+        let name = symbol.name.as_str();
+        let remote = symbol.remote.as_str();
+
+        if remote == "git" {
+            continue;
+        }
+
+        // trunk: main/master/trunk on origin/upstream
+        let is_trunk =
+            matches!(remote, "origin" | "upstream") && matches!(name, "main" | "master" | "trunk");
```
poliorcetics, line 308: This is making assumptions that I have seen not hold in various company repos

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,610 @@
...
+  remote_bookmarks(exact:"main", exact:"upstream") |
+  remote_bookmarks(exact:"master", exact:"upstream") |
+  remote_bookmarks(exact:"trunk", exact:"upstream") |
+  root()
+)
+'''
+'builtin_immutable_heads()' = 'trunk() | tags() | untracked_remote_bookmarks()'
+'immutable_heads()' = 'builtin_immutable_heads()'
+'immutable()' = '::(immutable_heads() | root())'
+'mutable()' = '~immutable()'
```
poliorcetics, line 432: By redefining the defaults here, you miss the user's config setting e.g `mutable()` to be different from the defaults

src/configs/jujutsu_bookmark.rs
```diff
@@ -0,0 +1,44 @@
...
+
+impl Default for JujutsuBookmarkConfig<'_> {
+    fn default() -> Self {
+        Self {
+            disabled: true,
+            format: "[$symbol$bookmarks]($style) ",
+            bookmark_format: "$bookmark_name(~$distance)$ahead_behind",
+            ahead: " ⇡$count",
+            behind: " ⇣$count",
+            diverged: " ⇡$ahead_count⇣$behind_count",
```
poliorcetics, line 34: Jujutsu bookmarks and Git branches are not the same thing and I think pretending they are is counterproductive

Notably, JJ encourages local merges a lot more and bookmarks don't auto-advance (unless the user actively modifies the config), so it will happen a lot more often that there will be several bookmarks at `@-` or `@--` for example, which will easily become super verbose

In #7612 I made the choice to print only the first bookmark found, that's it.

Additionally, since JJ bookmarks don't auto-advance, I don't think printing ahead/behind/divergence information makes a lot of sense: printing it in the format of `name(@remote)?\*?` like JJ does in `jj log` seems more in line with how JJ is used by default

(Note I keep relying on the defaults, because research has shown time and time again most people don't modify the default settings of stuff they use, be it software or hardware)

src/modules/utils/jujutsu.rs
```diff
@@ -0,0 +1,610 @@
+use std::collections::{HashMap, HashSet};
```
poliorcetics, line 1: Having to reimplement lots of JJ CLI functions is not amazing IMO, I could easily see drift coming in and it will be very hard to test and mock output compared to calling the CLI

For example, config resolution is likely to evolve in the future when trust is implemented or git-attributes support (e.g. LFS) is added, in which case starship would lag behind at best, produce incorrect output, or crash at worse

src/context.rs
```diff
@@ -56,6 +57,8 @@ pub struct Context<'a> {
 
     /// Private field to store Git information for modules who need it
     repo: OnceLock<Result<Repo, Box<gix::discover::Error>>>,
```
poliorcetics, line 59:
```suggestion
    git_repo: OnceLock<Result<Repo, Box<gix::discover::Error>>>,
```

For this and every method that was simply called `*repo*`, to make it easier to navigate the code

src/configs/jujutsu_state.rs
```diff
@@ -0,0 +1,48 @@
+use serde::{Deserialize, Serialize};
```
poliorcetics, line 1: This module is more in line with `git_status` so it should be `jujutsu_status`, there is not really a `git_state` equivalent since JJ doesn't force the user to handle rebase, merge and conflicts immediately

src/modules/jujutsu_commit.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::configs::jujutsu_commit::JujutsuCommitConfig;
+use crate::formatter::StringFormatter;
+use crate::modules::utils::jujutsu::get_jujutsu_commit_id;
+
+/// Creates a module with the Jujutsu commit hash in the current directory
+///
+/// Will display the commit hash if the current directory is a jj repo
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("jujutsu_commit");
+    let config: JujutsuCommitConfig = JujutsuCommitConfig::try_load(module.config);
```
poliorcetics, line 12:
```suggestion
    let config = JujutsuCommitConfig::try_load(module.config);
```

no need to double specify the type here and the other modules

poliorcetics (commented):
A quick review to look over what you added that I could adapt in my own MR and see how you implemented features

Overall I'm not a fan of using `jj_lib` at all, and the fact your MR is almost as big as mine without any docs and tests yet makes it hard for reviewers

Left some comments but don't take them as gospel :)

We took different approaches in both styling and available information and I obviously am biased in favor of mine but I don't have precise arguments here excepts for those below

---

I adapted the `immutable` and `hidden` booleans for my `jj_status` module

I voluntarily ignored `empty` and `missing_description`, I don't feel like they bring anything valuable enough: the first just means no display in my module, which communicates it well IMO and the second would just be symbol noise since `jj_status` is for the _current_ commit so it will often happen that the user is just waiting to do `jj commit` to write a description, no need to make their prompt more verbose for this

I'm iffy on `divergent`, I think it's too specific a state to be in to have a symbol for it but I could be convinced otherwise if shown a use case. Easily added in a future MR if people report missing it so it's not a deal break either

src/modules/utils/jujutsu.rs
101100: Can you expand on this? They use a special version of `jj`? That is a good argument for running the command separately.

src/modules/utils/jujutsu.rs
101100: My understanding of Jujutsu internals is limited, so this was taken from the popular https://github.com/dmmulroy/jj-starship plugin.

src/configs/jujutsu_bookmark.rs
101100: I like to be aware if I'm doing a history re-write (signified by the down arrow) or just additions (signified by the up arrow) when pushing especially because Jujutsu hides this so well (which I like) and others using git or reviewing PRs may get less useful information if I re-write history during the review process. It isn't a hill I'd die on, though.

101100:
Closing in favour of #7612.

PR #7619, opened 2026-07-24
fix(package): support build.gradle.kts in gradle version detection

### LINKED ISSUE #7561: Add support for build.gradle.kts files in package module
## Feature Request
Support build.gradle.kts in package module

Package version numbers can be extracted from build.gradle files, but not from the (recommended) Kotlin script variant (build.gradle.kts). 

The same logic as for build.gradle files could be applied to extract the version number from build.gradle.kts (I actually tested this by simply renaming my build.gradle.kts to build.gradle, which did the trick).

The only alternative would be to remain with support just for the more dated build.gradle files.

arimu1:
## Summary

`get_gradle_version` in the `package` module only looked for a file named `build.gradle`, so projects using the Kotlin DSL variant (`build.gradle.kts`) never had their package version detected — the module simply reported nothing for those projects.

Fixes #7561.

## Root cause

The existing version regex (`^version( |\s*=\s*)['"]...`) already matches Kotlin DSL syntax like `version = "1.0.0"` — the issue reporter confirmed this themselves by renaming `build.gradle.kts` to `build.gradle`, which "did the trick". The only actual gap was that the code never looked for the `build.gradle.kts` filename at all.

## Fix

Fall back to reading `build.gradle.kts` when `build.gradle` isn't present, reusing the same regex-based extraction logic.

## Test plan

- [x] Added `test_extract_gradle_version_kotlin_dsl`, mirroring the existing `build.gradle` tests, using Kotlin DSL syntax (`id("java")`, `version = "0.1.0"`)
- [x] `cargo test --lib modules::package` — all 67 tests pass, including the new one
- [x] `cargo fmt --check`

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -3712,7 +3712,7 @@ package, and shows its current version. The module currently supports `npm`, `ni
 - [**Python**](https://www.python.org) - The `python` package version is extracted from a [PEP 621](https://peps.python.org/pep-0621/) compliant `pyproject.toml` or a `setup.cfg` present in the current directory
 - [**Composer**](https://getcomposer.org/) – The `composer` package version is extracted from the `composer.json` present
   in the current directory
-- [**Gradle**](https://gradle.org/) – The `gradle` package version is extracted from the `build.gradle` present in the current directory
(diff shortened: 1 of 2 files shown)


davidkna:
Please use the PR template

arimu1:
Re-opened with the PR template filled in: https://github.com/starship/starship/pull/7627

PR #7637, opened 2026-07-27
docs(presets): add Clean Text preset

AbhinayMaurya:
### Description
Adds a new preset called **Clean Text** to the Starship presets documentation.

### Reason for Change
Many environments (stock Windows Terminal, remote SSH sessions, minimal Linux containers, CI/CD runners) do not have Nerd Fonts or custom patched fonts installed. 

This two-line preset keeps full developer runtime visibility (.NET, Node, Python, Go, Rust, Java, Docker) and Git status tracking using clean, standard ASCII text labels instead of special unicode icons.


davidkna:
What is the difference to the plain text preset for the proposed purpose?

PR #7626, opened 2026-07-25
refactor: Extract `context::git_repo` to prepare for the addition of Jujutsu

poliorcetics:
**See https://github.com/starship/starship/pull/7612 for the full MR with every changes, this MR has been extracted from it to make review easier.**

- Extract `src/context/{mod,git_repo}.rs` to make the coming addition of `src/context/jj_repo.rs` less painful
- Rename every `*repo*` method and struct member to `*git_repo*`
- Update Git module descriptions to mention they're for Git, JJ modules will in the same way mention they're for Jujutsu

Each commit can be reviewed on its own, they all pass tests and formatting.

PR #7643, opened 2026-07-30
feat: Add `jj_bookmark` module

poliorcetics:
**See https://github.com/starship/starship/pull/7612 for the full MR with every changes, this MR has been extracted from it to make review easier.**

- Following the refactor in #7626, add `src/context/jj_repo.rs` with two commits, the second being the main of it
- Add `jj_bookmark` in another two commits, one for the module implementation and one for the rendering tests

Each commit can be reviewed on its own, they all pass tests and formatting.

Some of the test infrastructure in `src/test/jj_tester.rs` will only make sense once later JJ module come up, it reduces duplicated lines a lot and allows tests to be short and to the point with little boilerplate.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2564,6 +2564,50 @@ By default the module will be shown if any of the following conditions are met:
 symbol = '🌟 '
 ```
 
+## JJ Bookmark
+
+The `jj_bookmark` module shows the [Jujutsu](https://docs.jj-vcs.dev/) bookmark when the current directory is in a Jujutsu repository.
+
+It looks at `@ | @-` to find bookmarks and will prioritize displaying those of `@` (after filtering has been applied).
+
+### Options
+
+| Option              | Default                                                                               | Description                                                                              |
+| ------------------- | ------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- |
+| `format`            | `"on [$symbol$bookmark(@$remote)$diverged( \\(+$overflow_count others\\))]($style) "` | The format for the module.                                                               |
+| `symbol`            | `" "`                                                                                | The symbol used in the `$symbol` variable.                                               |
+| `style`             | `"bold purple"`                                                                       | The style for the module.                                                                |
+| `truncation_length` | `2^16-1`                                                                              | Truncates the bookmark's name and remote to `N` graphemes.                               |
+| `truncation_symbol` | `"…"`                                                                                 | The symbol used to indicate a branch name was truncated. You can use `''` for no symbol. |
+| `diverged_symbol`   | `"*"`                                                                                 | Symbol used in the `$diverged` variable in the local bookmark diverged from its remote.  |
+| `ignore_names`      | `[]`                                                                                  | A list of bookmark names to avoid displaying. Useful for `'master'` or `'main'`.         |
+| `ignore_remotes`    | `[]`                                                                                  | A list of bookmark remotes to avoid displaying. Useful for `'upstream'` or `'fork'`.     |
+| `disabled`          | `false`                                                                               | Disables the `jj_bookmark` module.                                                       |
+
+### Variables
+
+| Variable         | Example  | Description                                                    |
+| ---------------- | -------- | -------------------------------------------------------------- |
+| bookmark         | `main`   | The bookmark's name                                            |
+| remote           | `origin` | The bookmark's remote, if any                                  |
+| diverged\*       | `*`      | Set when the bookmark is tracked and divergent from its remote |
+| overflow_count\* | `3`      | How many other bookmarks were found (and not ignored)          |
+| symbol           |          | Mirrors the value of option `symbol`                           |
+| style\*\*        |          | Mirrors the value of option `style`                            |
+
+- *: These variables are only set if there is cause: a tracked and divergent bookmark / a non-zero overflow count
+- **: This variable can only be used as a part of a style string
+
+### Example
+
+```toml
+# ~/.config/starship.toml
+
+[jj_bookmark]
+ignore_names = ["main", "master"]
+diverged_symbol = "⇕"
+```
+
 ## Jobs
 
 The `jobs` module shows the current number of jobs running.
--- a/docs/public/presets/toml/pure-preset.toml
+++ b/docs/public/presets/toml/pure-preset.toml
@@ -43,6 +43,10 @@ style = "bright-black"
 format = "[$duration]($style) "
 style = "yellow"
 
+[jj_bookmark]
+format = "[$bookmark(@$remote)$diverged( \\(+$overflow_count others\\))]($style)"
+style = "bright-black"
+
 [python]
 format = "[$virtualenv]($style) "
 style = "bright-black"
--- a/src/context/jj_repo.rs
+++ b/src/context/jj_repo.rs
@@ -1,0 +1,16 @@
+use std::path::PathBuf;
+
+use super::Context;
+
+pub struct JJRepo {
+    #[expect(unused)] // will go away in follow up commits
+    root: PathBuf,
+}
+
+impl JJRepo {
+    pub fn discover(context: &Context) -> Option<Self> {
+        let root = context.begin_ancestor_scan().set_folders(&[".jj"]).scan()?;
+
+        Some(Self { root })
+    }
+}
--- a/src/context/jj_repo.rs
+++ b/src/context/jj_repo.rs
@@ -1,16 +1,343 @@
 use std::path::PathBuf;
+use std::sync::OnceLock;
 
 use super::Context;
 
+/// A Jujutsu (JJ) repository's data.
 pub struct JJRepo {
-    #[expect(unused)] // will go away in follow up commits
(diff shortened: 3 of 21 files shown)


src/context/jj_repo.rs
```diff
@@ -0,0 +1,501 @@
...
+                            diff.stat().total_removed(),
+                            self
+                                .diff(".")
+                                .files()
+                                .filter(|file| !conflict || conflicted_files.all(|c| c.path().absolute() != file.path().absolute()))
+                                .map(|file| file.status_char())
+                                .join(""),
+                            "",
+                        ),
+                        "conflict_before|",
```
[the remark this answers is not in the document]
poliorcetics: That is entirely wrong. Trying out the JJ command in a conflicted repo will show it. Resolving.

src/context/jj_repo.rs
```diff
@@ -0,0 +1,501 @@
...
+                            hidden,
+                            immutable,
+                            divergent,
+                            change_offset,
+                            change_id,
+                            change_id.shortest().prefix().len(),
+                            commit_id,
+                            commit_id.shortest().prefix().len(),
+                            bookmarks,
+                            parents.map(|p| p.bookmarks()),
```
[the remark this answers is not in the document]
poliorcetics: Bookmarks are already printed separated with a space (because the character is not allowed in bookmark names or remote names), that is fine.

docs/public/presets/toml/pure-preset.toml
```diff
@@ -43,6 +43,10 @@ style = "bright-black"
 format = "[$duration]($style) "
```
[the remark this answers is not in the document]
poliorcetics: That is expected, `jj_bookmark` should never be in the default set since it would conflict with git (and be extra computation for something that the extremely vast majority of starship users do no use), that is documented in `docs/config/README.md` in https://github.com/starship/starship/pull/7612 when support for JJ in the VCS module is added

src/modules/jj_bookmark.rs
```diff
@@ -0,0 +1,319 @@
...
+pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
+    let mut module = context.new_module("jj_bookmark");
+    let config = JJBookmarkConfig::try_load(module.config);
+
+    if config.disabled {
+        return None;
+    }
+
+    let current_change = context.get_jj_repo()?.current_change(context)?;
+    let bookmarks = current_change.bookmarks.as_deref()?;
```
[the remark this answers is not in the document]
poliorcetics: Bookmarks from parent changes are put in `.bookmarks` at parse time with line `bookmarks: parse_bookmark_lines(lines.next()?, lines.next()?),`, _as is tested in the rendering tests_.

poliorcetics:
Same as https://github.com/starship/starship/pull/7612#issuecomment-5221887728, tests are failing for reasons I cannot control

src/context/jj_repo.rs
```diff
@@ -0,0 +1,501 @@
...
+
+                        flags
+                    },
+                })
+            })
+            .as_ref()
+    }
+}
+
+impl CurrentChange {
```
davidkna, line 205: This may not be necessary in the initial version, but for git module feature parity consider keeping track of the counts.

docs/config/README.md
```diff
@@ -2564,6 +2564,48 @@ By default the module will be shown if any of the following conditions are met:
...
+
+### Options
+
+| Option              | Default                                                            | Description                                                                              |
+| ------------------- | ------------------------------------------------------------------ | ---------------------------------------------------------------------------------------- |
+| `format`            | `"on [$symbol$bookmark( \\(+$overflow_count others\\))]($style) "` | The format for the module.                                                               |
+| `symbol`            | `" "`                                                             | The symbol used in the `$symbol` variable.                                               |
+| `style`             | `"bold purple"`                                                    | The style for the module.                                                                |
+| `truncation_length` | `u16::MAX`                                                         | Truncates the bookmark's name and remote to `N` graphemes.                               |
```
davidkna: I would prefer writing the definition here:
```suggestion
| `truncation_length` | `2^16-1`                                                         | Truncates the bookmark's name and remote to `N` graphemes.                               |
```

src/context/jj_repo.rs
```diff
@@ -0,0 +1,501 @@
...
+                // If the conflict is _before_ '@', the output is
+                //
+                //     conflict_before|false
+                //     <...>
+                //
+                // If the conflict is _at_ '@', the output is
+                //
+                //     true
+                //     <...>
+                let first_line = lines.next()?;
```
davidkna, line 130: I don't feel too strongly about this, but I would somewhat prefer jj JSON output, to avoid potential parsing issues.

src/modules/jj_bookmark.rs
```diff
@@ -0,0 +1,319 @@
...
+        result.push('@');
+        result.push_str(&truncate_text(
+            remote,
+            config.truncation_length.into(),
+            config.truncation_symbol,
+        ));
+    }
+    if diverged {
+        result.push_str(config.diverged_symbol);
+    }
```
davidkna: Prefer handling this via format strings/variables to allow individual styling of the different parts.

src/modules/jj_bookmark.rs
```diff
@@ -0,0 +1,319 @@
...
+pub mod tests {
+    use nu_ansi_term::Color;
+    use toml::{Table, Value, toml};
+
+    use crate::context::JJRepo;
+    use crate::test::ModuleRenderer;
+
+    /// Helper to test JJ modules
+    #[must_use]
+    pub struct Tester {
```
davidkna: I assume this will be shared between modules, so consider putting these helpers into `src/test`.

davidkna (changes requested):
I've made a few suggestions, but this mostly looks good already.

src/modules/jj_bookmark.rs
poliorcetics: done

src/context/jj_repo.rs
poliorcetics: JSON is quite annoying here because there are multiple situations:

1. No conflicts anywhere
2. Conflict in a parent, not in current
3. Conflict in current
4. Conflict in both current and parent

1, 3 and 4 produce the same output format but 2 does not and makes parsing json output super annoying since it would be two different "types".

From local testing, using JSON is also slower because of the contorsions it requires me to do in the templating (JJ does more work and we also do more work and allocations in starship), easily slowing down the modules by 10% even on a repo like starship with only one changed file. I'm wary of what that would look like on bigger repos or with lots of changes

Even with 0 changes in a repo and comparing only the jj commands with `hyperfine`, forcing JSON formatting is 3% slower at best, consistently, in several of my local repos

Unless we actively start having problems with parsing being brittle (it should not be, I checked allowed values for everything), I would recommend against JSON for now.

docs/config/README.md
poliorcetics: done

src/context/jj_repo.rs
poliorcetics: Could you mention this again when it's the turn of `jj_change` since that's the module that would need this ? that would make it easier for me 😅

src/modules/jj_bookmark.rs
poliorcetics: done

src/modules/jj_bookmark.rs
```diff
@@ -0,0 +1,218 @@
...
+        config.truncation_symbol,
+    );
+    if let Some(remote) = remote {
+        result.push('@');
+        result.push_str(&truncate_text(
+            remote,
+            config.truncation_length.into(),
+            config.truncation_symbol,
+        ));
+    }
```
davidkna: I would also prefer adding `$remote` as a variable - the format string could become the following:
`on [$symbol$bookmark(@$remote)$diverged( \\(+$overflow_count others\\))]($style) `

src/modules/jj_bookmark.rs
```diff
@@ -0,0 +1,218 @@
...
+                "symbol" => Some(config.symbol),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map(move |variable| match variable {
+                "bookmark" => Some(Ok(format_bookmark(&config, name, remote))),
+                "diverged" => diverged.then(|| Ok(config.diverged_symbol.to_string())),
```
davidkna: I think this should also go into `map_meta`.

src/modules/jj_bookmark.rs
poliorcetics: done

src/modules/jj_bookmark.rs
poliorcetics: done

davidkna (approved):

davidkna:
Thanks for all the jj contributions @poliorcetics!

poliorcetics:
@davidkna thank you for reviewing them, don't hesitate to ping me if issues crop up regarding JJ support

PR #7690, opened 2026-08-21
feat(directory): add `use_osc8_url` option

### LINKED ISSUE #3036: Clickable directory paths

It's not related to a problem. But it'd certainly be a QoL addition :)

Most modern terminals support the OSC 8 terminal escape sequence, which gives terminals the ability to display clickable links. Starship can leverage those sequences. Whether the path displayed in the directory module is clickable, can be controlled by a config flag `clickable_path` (true or false).

OhMyPosh has this feature builtin, so I got the idea from there.

#### Proposed Implementation Details

Consider the following PowerShell command:
```powershell
"`e]8;;$PWD`e`\`e[1;36m$(Split-Path -Path $PWD -Leaf)`e[0m`e]8;;`e`\"
```

If you run this, your terminal will display the current folder name in bold cyan color, and that folder name would be clickable. Clicking the folder name would open that folder in Explorer.

There are two parts to this:
1. The first is `$PWD`, which can be replaced by starship internally with the fully qualified path that it evaluates for the current directory.
2. The second is 
```
`e[1;36m$(Split-Path -Path $PWD -Leaf)`e[0m
```
which can be replaced by the decorated and/or truncated path that starship evaluates according to the user's config.

#### Additional context
I tried testing my hypothesis by specifying the above text in my directory config in starship.toml, but it seems that starship does not allow escape sequences in the toml file.

### LINKED ISSUE #4685: Make directory a file:// hyperlink
## Feature Request
It would be great if I could click on the directory of my prompt to open my file browser to the current location.

I usually type `nautilus .` in my terminal to open the current directory in a file browser but with hyperlinked text I could just click on the directory and not have to worry about what is the current default file browser.
Changing `~` to be linked to `file://$HOME` or whatever full path I'm currently in would be a great feature and time saver.

A write up of how to do this with PS1 prompt is here https://purpleidea.com/blog/2018/06/29/hyperlinks-in-gnome-terminal/

TomJo2000:
#### Description

This PR adds a setting to the `[directory]` module to allow optionally turning the output of the [directory] module into an interactable hyperlink in terminals
supporting the OSC 8 sequence.
For spec details and implementation status see:
https://github.com/Alhadis/OSC8-Adoption

#### Motivation and Context

- Resolves #3036
- Resolves #4685
- Partially addresses #7296

<details><summary>I have also been toying around with this as an ad-hoc feature via an [env] module for a while now.<br>
But due to the prompt length calculation issues I never put it into permanent use in my config.</summary>
<p>

```toml
# example config snippet of ad-hoc OSC 8 support
format = """\
    ${env_var.pwd}\
    $directory\
    [\u001B\\]8;;\u001B\\\\]()\
"""


# OSC 8 file:// URI embedding
[env_var.pwd]
disabled = true
variable = "PWD"
format = """\
    [\u001B\\]8;;]()\
    [file://$env_value]()\
    [\u001B\\\\]()\
"""
```

</p>
</details>

  - I am aware of the currently unaccounted for prompt-length discrepancy issue,
as described in [`#7296`](https://github.com/starship/starship/issues/7296).
I haven't been able to track down where that would need to be accounted for yet.

#### AI-Assistance

- [ ] Yes
- [x] No

N/A

#### Checklist:

  - I'm not entirely sure what tests would make sense for this feature,
  so I wanted to get feedback on the general approach I have taken to implement it first.
  At the very least we *should* test that `right_format` behaves correctly.
  e.g. that the OSC 8 sequence is correctly removed from the prompt length calculation.
  As also described in [`#7296`](https://github.com/starship/starship/issues/7296).
  And related to that probably also checks against the `[fill]` module,
  and some of the more esoteric `[directory]` options I didn't use in my local testing.
  I don't yet know the codebase well enough yet from the development side to tell where special attention may be needed with testing this feature.

<h1><em></em></h1> 

It may make sense to extend this feature to additional modules such as the `git_branch` or `git_commit` for example, to allow them to be interacted with to open the branch/tag/commit they point at on the remote in the browser.
If this makes sense for inclusion in other modules it would probably be best to implement "`OSC 8`-ification" as general a formatting function that modules can `use`.

### DIFF
--- a/.github/config-schema.json
+++ b/.github/config-schema.json
@@ -451,7 +451,8 @@
         "read_only_style": "red",
         "truncation_symbol": "",
         "home_symbol": "~",
-        "use_os_path_sep": true
+        "use_os_path_sep": true,
+        "use_osc8_url": false
       }
     },
     "direnv": {
@@ -2970,6 +2971,10 @@
         "use_os_path_sep": {
           "type": "boolean",
           "default": true
+        },
+        "use_osc8_url": {
+          "type": "boolean",
+          "default": false
         }
       },
       "additionalProperties": false
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -774,6 +774,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "displaydoc"
+version = "0.2.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c6232dd377dcc64799954cbd3a9bb882e9cdc1308ccd87b1c098f1fb2eaf82a8"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 3.0.3",
+]
+
+[[package]]
 name = "dlv-list"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -953,6 +964,15 @@ name = "foldhash"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
+
+[[package]]
+name = "form_urlencoded"
+version = "1.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cb4cb245038516f5f85277875cdaa4f7d2c9a0fa0468de06ed190163b1581fcf"
+dependencies = [
+ "percent-encoding",
+]
 
 [[package]]
 name = "fragile"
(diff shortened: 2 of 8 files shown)


TomJo2000:
I've now changed the `file://` URI encoding to use `Url::from_directory_path` which takes care of the percent encoding and deals with Windows paths.

I've also added `OSC` sequence stripping to `print.rs`.
To not make the regex pattern overly complex I've split the separate parts of the pattern up into a `patterns` array that gets joined with `|` to create the full pattern.
Since we are no longer exclusively matching ANSI SGR sequences I've renamed the function to `strip_escapes` and the `ANSI_REGEX` to `ESCAPE_REGEX`.

davidkna (commented):
This should use something like what is implemented in #4208  `src/segment.rs` to avoid overcounting width in `$fill`. For this particular feature, I think enhancing style strings to support `url:` for linking and a style variable `$dir_url` for that could also work.

As for your changes to the escaping, I would prefer something that aligned more with #7436.

TomJo2000:
Hadn't even looked at `$fill` yet.
Thanks for the tip.

Edit: Ah I see you added a `Control` segment that's always width 0, that's a good way to do it.
But I'll need to figure out how that fits in here.

TomJo2000:
I'm currently stuck trying to fit a `format!("\x1B]8;;{url}\x1B\\")` shaped peg,
into a `&'static str` shaped hole with this.

I'm not sure if I just backed myself into a corner because I made a wrong assumption about how lifetimes work, but as far as I can tell `&'static str` must be known at compile time.
Which is a problem because the `physical_dir` is a runtime value.
So I don't think it's possible to include runtime data in a Control segment as it is currently implemented in #4208, this may just be inexperience or me backing myself into a corner with my current approach though.

I think generally speaking adding a `url` field to style strings is probably a more general approach that would make it easier to reuse this in other modules.
But I doubt I'd have much more luck with the lifetimes there.

<h1><em></em></h1> 

Edit: I solved the lifetimes issue.
The assumption in #4208 seems to have been that `Segment::Control` would never need runtime data.
That's evidently wrong, so I've implemented a `pub struct ControlSegment` to hold the value like `TextSegment` and `FillSegment` already use.
The implementation details needed some slight changes too but follow the same format as `TextSegment`/`FillSegment` or `LineTerm` as appropriate.

davidkna:
@TomJo2000 `Box<str>` (or Cow) would have likely worked also worked for the record. It's also possible to get a static string at runtime via `Box::leak`.

TomJo2000:
I can push up what I have so far but it isn't plumbed up in the directory module yet.
I'll also still looking through #7436's changes to see where they fit into this PR.

I talked through the `&'static str` problem with a friend and they also mentioned `Box::leak` as a possibility, but we both agreed that probably wasn't a good idea.

TomJo2000:
I have rebased the branch against the latest commit on `main` and squashed the two earlier fixup commits into the OSC 8 implementation in `directory.rs`.

The `Segment::Control()` changes based on #4208 are kept as a separate commit with Co-author tags for you and @tstein who helped me reason through the changes made compared to the implementation in your earlier PR.

PR #7713, opened 2026-08-31
Support ANSI Notifications

schrieveslaach:
#### Description

This changes brings support for using ANSI notifications instead of direct desktop notifications which provides following benefits:

- The undistract-me feature will work via SSH shells
- Grouping of notifications in certain desktop environment will be done via the terminal emulator.

#### Motivation and Context

The main motivation for adjusting the feature was that starship notification became a lengthy list if there are not grouped (see GNOME shell screenshot below at the bottom, showing notifications of the latest starship release):

image

In contrast, the notification at the top show the notification via ANSI. There are nicely grouped and also I can click on the notification and I will be brought to the application in which this notification was raised (also useful if there are multiple terminal emulators open and maybe one doing a long running task via SSH when I share the same dotfiles on both machines). 

#### AI-Assistance

- [ ] Yes
- [x] No

N/A

#### Checklist:


src/configs/cmd_duration.rs
```diff
@@ -1,4 +1,77 @@
...
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(
+    feature = "config-schema",
+    derive(schemars::JsonSchema),
+    schemars(deny_unknown_fields)
+)]
+pub enum NotificationMode {
+    None,
+    Desktop,
+    Osc9,
```
schrieveslaach, line 15: Not sure about the name yet because there a different kind of ANSI codes that could be used.

src/configs/cmd_duration.rs
```diff
@@ -1,4 +1,77 @@
...
+}
+
+struct NotificationModeVisitor;
+
+impl<'de> Visitor<'de> for NotificationModeVisitor {
```
schrieveslaach, line 20: FYI: I implemented the visitor pattern here to make sure that configs with bools won't break. But that makes me wonder would I should do about the `schemars` of the enum.

src/modules/cmd_duration.rs
```diff
@@ -63,50 +63,66 @@ fn undistract_me<'a>(
...
+                    .body(&body)
+                    .icon("utilities-terminal")
+                    .timeout(timeout);
+
+                if let Err(err) = notification.show() {
+                    log::trace!("Cannot show notification: {err}");
+                }
+            }
+            crate::configs::cmd_duration::NotificationMode::Osc9 => {
+                module.segments.extend(crate::segment::Segment::from_text(
```
schrieveslaach, line 120: That part could be supported even if it is compiled without the `notify` feature.

schrieveslaach (commented):
I marked this PR as draft because it is currently a proof of concept and the code is not yet finished (also conventional commit pattern is missing). There are some remarks from my end that I would like to hammer out with the maintainers.

davidkna (changes requested):
I also made this PoC a while ago, which has some features that should be implemented here, please reference it: https://github.com/starship/starship/compare/main...davidkna:starship:native-notify

schrieveslaach:
@davidkna, thanks for pointing out this commit. I'll have a look.

src/configs/cmd_duration.rs
schrieveslaach: Found the Either struct and started to use this one.

schrieveslaach:
FYI: I incorporated some aspects of the changes. I'll have a look at Segment::Control later.

PR #7715, opened 2026-09-02
fix(directory): render symlinks pointing into a repo subdirectory

### LINKED ISSUE #7714: Symlink into subdirectory of repo doesn't show path in prompt
#### Current Behavior

If you have a symlink that points sideways into a subdirectory of a repo, `cd` ing into results in no path being shown in the starship prompt at all.

Example:
1. `~/code/me` is the repo
2. `~/code/me/private` is a directory in that repo
3. `~/code/example` symlinks into `~/code/me/private`
4. If `~/code/example` is your CWD you will see no preview directory

#### Expected Behavior

In example step 4 I would expect `~/code/example` to be shown as my path in the prompt

Image

#### Possible Solution

From what I can see, it looks like https://github.com/starship/starship/blob/e939a19ab777d64be2003c6f9c4c683ab1292945/src/modules/directory.rs#L115 hits the None case in my example? We could just default to `prefix + dir_string.as_str()` like some of the other fallback branches.

#### Environment
- Starship version: 1.26.0
- fish version: fish, version 4.8.1
- Operating system: Mac OS 15.7.9
- Git Commit Hash: 
- Branch/Tag: 
- Rust Version: rustc 1.97.0 (2d8144b78 2026-07-07) (built from a source tarball)
- Rust channel:  release

#### Starship Configuration

https://github.com/PatrickShaw/patrickshaw/blob/main/personal/dotfiles/starship/starship.toml

qatcod:
Fixes #7714.

A symlink that points sideways into a subdirectory of a repository resolves into that repository, so the repo is found, but the logical path shares no prefix with the repo root and therefore has no contracted form. `contract_repo_path` returns `None`, and the `?` on that call propagated out of `module()` itself, so nothing was rendered at all.

Reproducing with `~/code/example` symlinked to `~/code/me/private`, where `~/code/me` is the repo:

```
                              no repo_root_style      with repo_root_style
  real subdir  (me/private)   me/private              me/private
  via symlink  (example)      code/example            <nothing>
```

Only the last cell is wrong, and it needs `repo_root_style` to be set, since that is what selects this branch of the match. That is why the existing `symlinked_subdirectory_git_repo_out_of_tree` test does not catch it: without `repo_root_style` the path goes through the `_` arm, which has no `?`.

The fix is the one suggested on the issue: fall back to the plain path, matching what the neighbouring arms of the same match already do for the cases they cannot contract. The symlink then renders as `code/example`, and repo-root highlighting is unaffected for paths that genuinely sit under the root.

The new test is `#[ignore]`d like the other symlink tests, so it still runs under the coverage job, which passes `--include-ignored`. Reverting the `directory.rs` change with the test in place fails with `right: None`, which is the reported symptom exactly.

`cargo fmt` and `cargo clippy --all-targets` are clean, and all 65 directory tests pass. The only failures in the full run are the `hg_branch` and `hg_state` suites, which fail the same way on a pristine checkout here because Mercurial is not installed.

### DIFF
--- a/src/modules/directory.rs
+++ b/src/modules/directory.rs
@@ -110,13 +110,15 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
         String::new()
     };
 
+    // the whole path in one segment, with no repo root to highlight
+    let unsplit = |prefix: String| [String::new(), String::new(), prefix + dir_string.as_str()];
+
     let path_vec = match &repo.and_then(|r| r.workdir.as_ref()) {
         Some(repo_root) if config.repo_root_style.is_some() => {
             // The logical path need not lie under the repo root even though the
             // repo was found: a symlink pointing sideways into a subdirectory of
             // the repo resolves into it without sharing a prefix with it, and so
-            // has no contracted form. Render it like any other path instead of
-            // giving up, which used to blank the whole module.
+            // has no contracted form, which is the None case here.
             match contract_repo_path(display_dir, repo_root) {
                 Some(contracted_path) => {
                     let repo_path_vec: Vec<&str> = contracted_path.split('/').collect();
@@ -130,13 +132,13 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
                         let before = before_root_dir(&dir_string, &contracted_path);
                         [prefix + before, root.to_string(), after_repo_root]
                     } else {
-                        [String::new(), String::new(), prefix + dir_string.as_str()]
+                        unsplit(prefix)
                     }
                 }
-                None => [String::new(), String::new(), prefix + dir_string.as_str()],
+                None => unsplit(prefix),
             }
         }
-        _ => [String::new(), String::new(), prefix + dir_string.as_str()],
+        _ => unsplit(prefix),
     };
 
     let path_vec = if config.use_os_path_sep {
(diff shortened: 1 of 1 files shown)


src/modules/directory.rs
```diff
@@ -112,19 +112,28 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                        || ((num_segments_after_root - 1) as i64) < config.truncation_length
+                    {
+                        let root = repo_path_vec[0];
+                        let before = before_root_dir(&dir_string, &contracted_path);
+                        [prefix + before, root.to_string(), after_repo_root]
+                    } else {
+                        [String::new(), String::new(), prefix + dir_string.as_str()]
+                    }
+                }
+                None => [String::new(), String::new(), prefix + dir_string.as_str()],
```
PatrickShaw: There's a triple-up of fallback return cases as a consequence of this PR.

Not a fan person - Rule of 3s tells me this should probably be refactored

qatcod:
Good catch on the triple-up, that was a consequence of adding the third case. Pulled it into one closure:

```rust
// the whole path in one segment, with no repo root to highlight
let unsplit = |prefix: String| [String::new(), String::new(), prefix + dir_string.as_str()];
```

It takes `prefix` as a parameter rather than capturing it, since the contracting branch still needs it. All three arms call `unsplit(prefix)` now.

Also moved the regression test into the unix-only module, next to `symlinked_subdirectory_git_repo_out_of_tree`, which it mirrors. That is what the Windows failures were: the test asserts a `~`-contracted path through `HOME`, which does not hold on Windows, and its neighbour is in that module for the same reason.

Behaviour is unchanged either way round: the symlink renders as `code/example`, a real subdirectory still gets the repo root highlighted, and all 65 directory tests pass with `cargo fmt` and `cargo clippy --all-targets` clean.

src/modules/directory.rs
```diff
@@ -110,24 +110,35 @@ pub fn module<'a>(context: &'a Context) -> Option<Module<'a>> {
...
+                    let num_segments_after_root = after_repo_root.split('/').count();
+
+                    if config.truncation_length == 0
+                        || ((num_segments_after_root - 1) as i64) < config.truncation_length
+                    {
+                        let root = repo_path_vec[0];
+                        let before = before_root_dir(&dir_string, &contracted_path);
+                        [prefix + before, root.to_string(), after_repo_root]
+                    } else {
+                        unsplit(prefix)
```
PatrickShaw, line 135: Nit: Still technically a tripple-up , just a smaller one xD

i.e. You can still run into the case where 1 fallback is changed but the other 2 don't get the same fixes.

Certainly better though

PatrickShaw:
To my eyes, it looks like it'll work.

If we need me to confirm it works: I won't be around to test this for another 3 weeks, just a heads up.

Thank you for the patch regardless :)

qatcod:
Fair point, three copies is three chances to fix one and miss the others. Collapsed it: the split is now an `Option` built up from the repo root, and a single match either uses it or falls back to the whole path in one segment. The closure is gone.

Behaviour is the same, the symlink renders as `code/example` and a real subdirectory still gets the repo root highlighted. `fmt` and `clippy --all-targets` are clean and the 59 directory tests pass.

No rush on the verification, and thanks for the reviews.

davidkna (approved):

PR #7729, opened 2026-09-12
feat: add `jj_status` module

poliorcetics:
**See https://github.com/starship/starship/pull/7612 for the full MR with every changes, this MR has been extracted from it to make review easier.**

- First commit prepares the addition of `jj_status` by extracting functions from `git_status` to `utils::format`
- Second keeps track of counts for added/copied/deleted/modified/renamed files to address https://github.com/starship/starship/pull/7643#discussion_r3740771615
- Third and fourth commit add `jj_status`, starting with the module implementation and then the rendering tests

Each commit can be reviewed on its own, they all pass tests and formatting.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -2690,6 +2690,85 @@ added_style = 'bold blue'
 format = '[+$added]($added_style)/[-$deleted]($deleted_style) '
 ```
 
+## JJ Status
+
+The `jj_status` module shows symbols representing the state of the [Jujutsu](https://docs.jj-vcs.dev/) repo in your current directory.
+
+### Options
+
+| Option                | Default                   | Description                                                                |
+| --------------------- | ------------------------- | -------------------------------------------------------------------------- |
+| `format`              | `'([\[$all\]]($style) )'` | The default format for `jj_status`.                                        |
+| `conflicted`          | `'!'`                     | The format shown when the current change or a parent has merge conflicts.  |
+| `description_empty`   | `'◌'`                     | The format shown when the current change has no description.               |
+| `description_present` | `''`                      | The format shown when the current change has a description.                |
+| `hidden`              | `''`                      | The format shown when the current change is hidden.                        |
+| `immutable`           | `'◆'`                     | The format shown when the current change is immutable.                     |
+| `added`               | `'+'`                     | The format shown when a new file has been added in the working directory.  |
+| `copied`              | `'='`                     | The format shown when a new file has been copied in the working directory. |
+| `deleted`             | `'✘'`                     | The format shown when a file has been deleted in the working directory.    |
+| `modified`            | `'~'`                     | The format shown when a file has been modified in the working directory.   |
+| `renamed`             | `'»'`                     | The format shown when a file has been renamed in the working directory.    |
+| `style`               | `'bold red'`              | The style for the module.                                                  |
+| `disabled`            | `false`                   | Disables the `jj_status` module.                                           |
+
+### Variables
+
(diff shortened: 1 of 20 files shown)


docs/public/presets/toml/pure-preset.toml
```diff
@@ -47,6 +47,20 @@ style = "yellow"
 format = "[$bookmark(@$remote)$diverged( \\(+$overflow_count others\\))]($style)"
 style = "bright-black"
 
+[jj_status]
+format = "[[(*$all)](218)]($style)"
```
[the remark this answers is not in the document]
poliorcetics: I see presets as additional documentation so I don't think that's bad

poliorcetics:
@davidkna I addressed https://github.com/starship/starship/pull/7643#discussion_r3740771615 where you asked to support `count` for added/copied/deleted/modified/renamed, this is has been done in this MR

src/modules/jj_status.rs
```diff
@@ -0,0 +1,294 @@
...
+                "all" => Some("$conflicted$description$hidden$immutable$added$copied$deleted$modified$renamed"),
+                _ => None,
+            })
+            .map_style(|variable| match variable {
+                "style" => Some(Ok(config.style)),
+                _ => None,
+            })
+            .map_variables_to_segments(|variable| {
+                let segments = match variable {
+                "added" => format::count(config.added, "jj_status.added", context, current_change.status.added),
```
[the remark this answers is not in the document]
poliorcetics: I wish I could be as confidently wrong as a bot (`cargo fmt` doesn't format anything here, I had to do it manually)

src/context/jj_repo.rs
```diff
@@ -38,8 +38,24 @@ pub struct CurrentChange {
...
+}
+
+#[derive(Debug, Default)]
+pub struct Status {
     /// See `impl CurrentChange` below
-    flags: u32,
+    flags: u8,
+
+    /// Count of added files
+    pub added: usize,
```
davidkna, line 50: It probably doesn't matter, but how about `std::num::Saturating<u32>`?

docs/config/README.md
```diff
@@ -2690,6 +2690,85 @@ added_style = 'bold blue'
...
+
+```toml
+# ~/.config/starship.toml
+
+[jj_status]
+added = "+($count)"
+copied = "=($count)"
+deleted = "✘($count)"
+modified = "~($count)"
+renamed = "»($count)"
```
davidkna: `$count` will always be set, so optionals could give the wrong impression:
```suggestion
added = "+$count"
copied = "=$count"
deleted = "✘$count"
modified = "~$count"
renamed = "»$count"
```
```suggestion
added = "+${count}"
copied = "=${count}"
deleted = "✘${count}"
modified = "~${count}"
renamed = "»${count}"
```

davidkna (approved):
I made some minor comments, but in general LGTM.

docs/config/README.md
poliorcetics: done

src/context/jj_repo.rs
poliorcetics: I initially used `u32` yes but since I'm sharing the `utils::format::count` function with the `git_status` module I have to either use `usize` or make `count` generic and that second option seemed overkill for such a small utility

`usize` doesn't have a `From<u32>` impl either so using `u32` would mean a manual cast, I didn't want to do that

I can do it if you want though ?

davidkna:
Thanks again!

PR #7611, opened 2026-07-21
fix(hostname): check SSH_CLIENT and SSH_TTY in addition to SSH_CONNECTION

itzArka1407:
#### Description
Fixes the hostname module only checking `$SSH_CONNECTION` to detect
an SSH session, missing `SSH_CLIENT` and `SSH_TTY`.

#### Motivation and Context
The `username` module already checks all three SSH-related
environment variables (`SSH_CONNECTION`, `SSH_CLIENT`, `SSH_TTY`),
but `hostname` only checked `SSH_CONNECTION`. Depending on SSH
server/client configuration, only one of these three may be set,
so `hostname` could incorrectly report "not SSH" in sessions where
`username` correctly detected one.

Closes #7207

#### How Has This Been Tested?
Added three unit tests (`ssh_connection`, `ssh_client`, `ssh_tty`)
covering each of the three SSH-related environment variables.
Ran `cargo test --lib hostname::` — all 14 tests pass. Also ran
`cargo fmt` and `cargo clippy --all-targets --all-features` with
no issues.

#### AI-Assistance
- [x] Yes

I used AI assistance (Claude) to understand the codebase — comparing
`hostname.rs` and `username.rs` to identify the inconsistency, and to
learn the git/GitHub contribution workflow, since this is my first
open-source contribution. The actual fix (checking all three SSH env
vars) and the three test functions were written by me. I reviewed,
tested, and understand every part of this change and can answer
questions about it.

#### Checklist:
- [x] I have updated the documentation accordingly.
- [x] I have updated the tests accordingly.
- [x] I understand and have read the code I contribute and can answer questions about it.

PR #7735, opened 2026-09-15
fix(hostname): also detect SSH via SSH_CLIENT and SSH_TTY

### LINKED ISSUE #7207: Hostname module only checks $SSH_CONNECTION (different from Username module)

angelinaaziz:
hostname only looks at `SSH_CONNECTION` to decide whether you're in an SSH session. username already checks `SSH_CLIENT` and `SSH_TTY` as well, so the two modules disagree. On a host that sets `SSH_CLIENT` or `SSH_TTY` but not `SSH_CONNECTION`, username shows up but hostname stays hidden under the default `ssh_only = true` (and the `ssh_symbol` doesn't render). That's #7207.

I moved the check username was already doing into a `Context::is_ssh_session()` helper and used it in both modules, so they can't drift apart again. I also documented it on the hostname options — the note just mirrors the one username already has — and added tests for the `SSH_CLIENT` and `SSH_TTY` cases.

Tested locally:
- `cargo fmt --check`
- `cargo clippy --lib --all-features -- -D warnings`
- `cargo test` (hostname + username)

Closes #7207


davidkna:
Closing as this was already implemented in #7611.

PR #7736, opened 2026-09-16
fix(pwsh): clear leftover ListView prediction rows on transient prompt

### LINKED ISSUE #6396: Transient Prompt Doesn't Clear Correctly
pkg_version:1.21.1
branch:master
commit_hash:47ccc36
build_time:2024-10-18 16:32:39 +00:00
build_env:rustc 1.82.0 (f6e511eec 2024-10-15),stable-x86_64-pc-windows-msvc

Generated bug report:
#### Current Behavior

When using a transient prompt in PowerShell (`Enable-TransientPrompt`), the PowerShell history is not cleared correctly.
(See GIF)

#### Expected Behavior

The history should be cleared correctly, as it is when transient prompt is not enabled (`Disable-TransientPrompt`).
(See GIF)

TransientPromptError
**NOTE**: Notice the left-over `[History]` text?

#### Environment
- Starship version: 1.21.1
- pwsh version: PowerShell 7.4.6
- Operating system: Windows 10.0.19045
- Git Commit Hash: 47ccc36
- Branch/Tag: master
- Rust Version: rustc 1.82.0 (f6e511eec 2024-10-15)
- Build Time: 2024-10-18 16:32:39 +00:00

```pwsh
Set-PSReadLineOption -HistorySearchCursorMovesToEnd
Set-PSReadLineKeyHandler -Key UpArrow -Function HistorySearchBackward
Set-PSReadLineKeyHandler -Key DownArrow -Function HistorySearchForward

Import-Module F7History

# https://youtu.be/Q11sSltuTE0 - John Savill's Technical Training (@NTFAQGuy) - Transforming PowerShell experience with PSReadLine
#
# Sometimes you enter a command but realise youy forgot to do something else first.
# This binding will let you save that command in the history so you can recall it,
# but it doesn't actually execute. It also clears the line with `RevertLine` so the
# undo stack is reset - though redo will still reconstruct the command line.
Set-PSReadLineKeyHandler -Key Alt+w `
                         -BriefDescription SaveInHistory `
                         -LongDescription "Save current line in history but do not execute" `
                         -ScriptBlock {
                            param($key, $arg)

                            $line = $null
                            $cursor = $null

                            [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$line, [ref]$cursor)
                            [Microsfot.PowerShell.PSConsoleReadLine]::AddToHistory($line)
                            [Microsfot.PowerShell.PSConsoleReadLine]::RevertLine()
                         }
```

#### Starship Configuration

```toml
format = '$all$shlvl$character'

# Get editor completions based on the config schema

# Inserts a blank line between shell prompts
add_newline = true

[[battery.display]] # "bold red" style when capacity is between 0% and 10%
threshold = 10
style = "bold red"

[[battery.display]] # "bold yellow" style when capacity is between 10% and 30%
threshold = 30
style = "bold yellow"

[[battery.display]] # "bold green" style when capacity is between 30% and 100%
threshold = 100
style = "bold green"

[git_commit]
tag_disabled = false
tag_symbol = " 🏷 "

[git_metrics]
disabled = false

[hostname]
ssh_only = false
style = 'bold green'

[localip]
ssh_only = false
disabled = false

[memory_usage]
disabled = false

[os]
disabled = false

[os.symbols]
Windows = " "
Macos = " "
Linux = " "
#Raspbian = " "

[shell]
bash_indicator = "bash"
powershell_indicator = "pwsh"
unknown_indicator = "unknown shell"
style = "cyan bold"
disabled = false

[shlvl]
# To make SHLVL work, ensure that something like the code below is included in your PowerShell profile:
# $env:SHLVL = [int] $env:SHLVL + 1
disabled = false
symbol = "❯" # "SHLVL" # "↕️"
repeat = true
repeat_offset = 1
format = '[$symbol]($style)'

[status]
disabled = false
format = "[$symbol $status]($style) "

[username]
style_root = 'bold bright-red inverted blink'

[sudo]
allow_windows = true
disabled = false

[time]
time_format = "%H:%M:%S%.3f"
disabled = false

#[custom.yazi]
#description = "Indicate the shell was launched by `yazi`"
#symbol = "󰇥 "
#when = "[int]$env:YAZI_LEVEL -ge 1"
#shell = "pwsh"
#style = "bold yellow"

```

mpostument:
#### Description
`Enable-TransientPrompt`'s Enter handler tries to erase PSReadLine's `ListView` prediction rows before accepting the line, by inserting newlines and undoing them:

```pwsh
[Microsoft.PowerShell.PSConsoleReadLine]::Insert("`n" * [math]::Min($Host.UI.RawUI.WindowSize.Height - $Host.UI.RawUI.CursorPosition.Y - 1, 12))
[Microsoft.PowerShell.PSConsoleReadLine]::Undo()
```

`CursorPosition.Y` is the cursor's row in the whole scroll buffer, while `WindowSize.Height` is only the visible window's row count. The subtraction goes deeply negative as soon as the buffer has scrolled past one window height (i.e. almost immediately in normal use), so the newline count clamps to ~0 and nothing gets cleared.

This PR drops that math entirely and always inserts the already-documented safe constant (12 lines, covering the list's 10-item max), keeping the same `Insert()`/`Undo()` mechanism so PSReadLine's own internal render-line bookkeeping (`-ExtraPromptLineCount`, used to compute cursor position on the *next* render) stays consistent. (An earlier version of this PR tried writing `ESC[0J` directly to the console instead - no coordinate math needed at all - but that bypasses PSReadLine's own bookkeeping and corrupts the *next* prompt render instead of fixing the current one. See the PR comments for details; reverted back to `Insert()`/`Undo()` after live testing showed the raw-ANSI approach didn't actually work.)

#### Motivation and Context
Closes #6396 — leftover `[History]`/ListView prediction rows stay on screen after Enter when `Enable-TransientPrompt` is combined with `Set-PSReadLineOption -PredictionViewStyle ListView`.

Manually, interactively, in Windows Terminal + PowerShell 7 (built this branch locally and swapped it into a throwaway PSReadLine+`Enable-TransientPrompt` session): scrolled the buffer past one window height, typed a partial command to bring up the `ListView` dropdown, confirmed the rows are fully cleared after Enter - including with the cursor moved left off the end of the line before hitting Enter (the caret-position case reviewers flagged). Also compared side-by-side against a build of unmodified `upstream/main` to confirm the bug reproduces there and not on this branch.

Not tested on macOS or Linux (this code path is PowerShell/Windows-only).

#### AI-Assistance

- [x] Yes
- [ ] No

Used Claude Code (Anthropic) throughout: diagnosing the bug, writing/committing/iterating on the fix (including an approach that turned out not to work under live testing, and was reverted), and building the branch locally to test. I (the PR author) ran the actual interactive verification myself in my own terminal and reviewed/directed each iteration of the fix, including rejecting the first approach once real-world testing showed it was broken.

#### Checklist:


mpostument:
Thanks for the review — addressed the cursor-position issue, and a note on the other two:

**1. ESC[0J erasing visible command text (confirmed, fixed in 125d66fd)**

Correct catch. `ESC[0J` erases from wherever the cursor currently is, not "everything below the input" — if the edit cursor was left mid-line (e.g. the user moved left to fix a typo before hitting Enter), the erase would wipe part of the visible command text instead of just the stale `ListView` rows below it. Added `[Microsoft.PowerShell.PSConsoleReadLine]::EndOfLine()` right before the erase so the cursor is always at the true end of the buffer first.

**2. Missing interactive test coverage**

Agreed this is a gap, but it's a pre-existing one — nothing in the current test suite exercises this file's PSReadLine rendering behavior at all (it needs a live console/PTY to observe cursor and screen state, which is a bigger infra addition than this fix). Open to pointers if there's a preferred approach already used elsewhere in the repo for this kind of thing; happy to add it as a follow-up.

**3. `$script:DoesUseLists` / VT capability**

This whole prompt already depends unconditionally on VT/ANSI processing for every color code it emits (the character module, right-prompt escapes, etc.), so `ESC[0J` isn't introducing a new terminal requirement beyond what's already assumed throughout `starship.ps1`. Didn't add separate capability detection for that reason, but can if you'd rather have it explicit/defensive here specifically.

mpostument:
Re: interactive regression reproduction for scrolled ListView + caret-before-end (P2)

I looked for existing precedent before responding: there is no PTY/interactive test infrastructure anywhere in this crate today. The only dev-dependency is `mockall`; existing tests are Rust unit tests and module-output snapshots, none of which drive a real terminal or observe cursor/screen state. Building this properly would mean adding a ConPTY-based harness plus a VT100 screen-buffer emulator to parse escape sequences into an actual screen state, simulating real keystrokes (arrow-left, Enter) against a live `pwsh`+PSReadLine session — and since it's exercising `ConPTY`/PSReadLine specifically, it'd only cover Windows, in a crate that's otherwise cross-platform.

That's a meaningful new testing capability, not an incremental addition to this fix, so I don't think it belongs unprompted in a small bugfix PR. Happy to open it as a separate issue/discussion for maintainers to weigh whether that investment is worth it — let me know if you'd like that, or if there's a lighter-weight approach already used elsewhere in the repo that I missed.

mpostument:
Update: reverted the ESC[0J approach in d81225c, back to the original Insert()/Undo() mechanism.

I initially thought `ESC[0J` was the cleaner fix (no buffer/window coordinate math needed), and it looked correct in review. But live interactive testing (Windows Terminal, PowerShell 7) showed it doesn't actually work: writing the escape sequence directly to the console bypasses PSReadLine's own internal tracking of how many lines its output occupies (`Set-PSReadLineOption -ExtraPromptLineCount` a few lines below is exactly this bookkeeping). That desync corrupted the *next* prompt render instead of fixing the current one — visible as prior prompt lines getting overwritten/spliced with new ones.

d81225c goes back to the original `Insert()`/`Undo()` approach (which stays entirely inside PSReadLine's own render pipeline, so its internal state remains consistent), and only removes the buggy `WindowSize.Height - CursorPosition.Y` math in favor of the already-documented safe constant (12 lines). This also makes the P3 finding about `EndOfLine()` on the parse-error/multi-line-continuation path moot, since `EndOfLine()` no longer exists in the code at all.

Manually verified interactively this time (not just parsed/compiled): scrolled the buffer past one window height, typed a partial command to bring up the `ListView` dropdown, confirmed it's fully cleared after Enter, including with the cursor moved off the end of the line before hitting Enter (the P2 caret-position scenario) - full command text stays intact in that case too.

mpostument:
Closing this - the underlying code issue is real and I stand by the fix and the live testing that confirmed it, but the original problem report that prompted this PR turned out not to be reproducible anymore on the reporter's machine (likely unrelated - a terminal/app restart or update in between). Since I can no longer confirm this actually matches a real-world symptom someone is hitting, rather than just a bug I found by reading the code, I'd rather not have it sit open as noise for maintainers to triage. Happy to reopen or have someone else pick it up if #6396 (or something matching this) resurfaces for someone.

